PR #188, opened 2018-11-12
Adds request charge header and makes max continuation token size configurable

johnstairs:
The RUs consumed during a request will now be included in the response headers. 
Also allowing the maximmum continuation token size to be set via config. By default, Cosmos DB's tokens can grow up to 16KB, which could be problematic for some clients since the token needs to be a URI query parameter. It can now be capped at 1KB per request. We are also doing away with the approach of storing the continuation token in a separate document in the database.

Also consolidating the two codegened IDocumentClient classes into one.


johnstairs:
#190 
#191

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/CosmosSearchService.cs
```diff
@@ -66,18 +62,11 @@ public class CosmosSearchService : SearchService
             SearchOptions searchOptions,
             CancellationToken cancellationToken)
         {
-            string ct = null;
```
jackliums: I don't think it's a problem but this is not backwards compatible. (e.g., if there is a search happening while upgrade is going on, then the continuation token would become invalid and the client would receive error). Consider we don't have real load right now that might be okay but just wanted to call out.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/FhirDocumentClient.cs
```diff
@@ -13,22 +13,26 @@
...
-namespace Microsoft.Health.Fhir.CosmosDb.Features.Consistency
+namespace Microsoft.Health.Fhir.CosmosDb.Features.Storage
 {
     /// <summary>
-    /// A <see cref="IDocumentClient"/> wrapper that sets the consistency level and, if applicable, the session token
-    /// for each request.
+    /// A <see cref="IDocumentClient"/> wrapper that, for each request:
+    /// (1) sets the consistency level and, if applicable, the session token
+    /// (2) Sets <see cref="FeedOptions.ResponseContinuationTokenLimitInKb"/>
+    /// (3) Sets the <see cref="CosmosDbHeaders.RequestCharge"/> response header.
```
jackliums: should we also add a line about converting exception?

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/CosmosSearchService.cs
johnstairs: That's true.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/FhirRequestContextDocumentClientExtensions.cs
```diff
@@ -0,0 +1,77 @@
...
+    {
+        /// <summary>
+        /// Adds request charge to the response headers and throws a <see cref="RequestRateExceededException"/>
+        /// if the status code is 429.
+        /// </summary>
+        /// <param name="requestContext">The request context</param>
+        /// <param name="ex">The exception</param>
+        public static void ProcessException(this IFhirRequestContext requestContext, Exception ex)
+        {
+            if (requestContext == null)
```
jackliums: should this be EnsureArg.IsNotNull(requestContext, nameof(requestContext))? It works but semantically calling requestContext.ProcessException where requestContext is null is a little weird :)

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/FhirRequestContextDocumentClientExtensions.cs
```diff
@@ -0,0 +1,77 @@
...
+            {
+                requestContext.ResponseHeaders[CosmosDbHeaders.SessionToken] = sessionToken;
+            }
+
+            requestContext.AddRequestChargeToResponseHeaders(responseRequestCharge);
+        }
+
+        private static void AddRequestChargeToResponseHeaders(this IFhirRequestContext requestContext, double responseRequestCharge)
+        {
+            if (requestContext.ResponseHeaders.TryGetValue(CosmosDbHeaders.RequestCharge, out StringValues existingValues) &&
```
jackliums: Instead of storing the value in the header and parsing header back to double, we could store the double value in requestContext itself and then when the value is updated just calculates the new double value and at the same time write to the header?

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/FhirDocumentClient.cs
```diff
@@ -13,22 +13,26 @@
...
     /// </summary>
-    internal sealed partial class DocumentClientWithConsistencyLevelFromContext
+    internal sealed partial class FhirDocumentClient
```
jackliums: What about the request charged when SDK retry option is not set to 0? When the SDK retries, does it return accumulated request charges or just last operation?

jackliums: What about the request charged when SDK retry option is not set to 0? When the SDK retries, does it return accumulated request charges or just last operation?

johnstairs: Accumulated.

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/CosmosSearchService.cs
```diff
@@ -106,17 +95,7 @@ public class CosmosSearchService : SearchService
...
-                // TODO: Eventually, we will need to take a snapshot of the search and manage the continuation
-                // tokens ourselves since there might be multiple continuation token involved depending on
-                // the search.
-                if (!string.IsNullOrEmpty(fetchedResults.ResponseContinuation))
-                {
-                    continuationTokenId = await _continuationTokenCache.SaveContinuationTokenAsync(fetchedResults.ResponseContinuation, cancellationToken);
-                }
-
-                return new SearchResult(wrappers, continuationTokenId);
+                return new SearchResult(wrappers, fetchedResults.ResponseContinuation);
```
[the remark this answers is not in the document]
johnstairs: The value ends up URL encoded, so we are fine. We could base64 it, but that consumes resources, so it's not a clear win in my opinion.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/FhirRequestContextDocumentClientExtensions.cs
johnstairs: Added a comment explaining the rationale.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/FhirRequestContextDocumentClientExtensions.cs
johnstairs: Taking this approach instead of having to do null checks at all the call sites. Agree it is a bit weird.

PR #194, opened 2018-11-13
Updates package 'Microsoft.AspNetCore.App' to version '2.1.6'

MicrosoftHealthService:


hansenms:
Can we hold on this. For purposes of MAG compatibility, we should not update this version unless we have good reasons to do so.

hansenms:
This should be mergeable now. Azure Gov has .NET Core 2.1.6

Issue #51, opened 2018-09-20
Support compartment search


jackliums:
Because compartment is now indexed, I should be able to do a search based on compartment. 

https://www.hl7.org/fhir/compartmentdefinition.html#use

Acceptance:

- Verify GET [base]/Patient/[id]/Observation returns all observations for the patient
- Verify the logic works for all compartments

Issue #149, opened 2018-10-26
Index Compartments for Search


garamamo:
Load compartments from a compartment definition file.
Create search lookups for the compartment search params.
Add a compartment property to the search index for the reference search param that represents the compartment.

PR #198, opened 2018-11-13
Enable Compartment Search

garamamo:
After the work done for indexing documents with compartment ids in https://github.com/Microsoft/fhir-server/issues/149, this pr enables end to end compartment search.

It includes
a) Controller and Routing changes for searching on compartments as per spec (https://www.hl7.org/fhir/compartmentdefinition.html) as GET.
b) Core changes to support compartment search including resource specific and wildcard searches.
c) CosmosDB querybuilder changes to support compartment search as per spec.
d) Unit and Integration Tests.


src/Microsoft.Health.Fhir.Api/Features/Routing/CompartmentResourceTypesRouteConstraint.cs
```diff
@@ -0,0 +1,39 @@
...
+        public bool Match(HttpContext httpContext, IRouter route, string routeKey, RouteValueDictionary values, RouteDirection routeDirection)
+        {
+            EnsureArg.IsNotNull(httpContext, nameof(httpContext));
+            EnsureArg.IsNotNull(route, nameof(route));
+            EnsureArg.IsNotNullOrEmpty(routeKey, nameof(routeKey));
+            EnsureArg.IsNotNull(values, nameof(values));
+
+            if (values.TryGetValue(KnownActionParameterNames.ResourceType, out var resourceTypeObj) && resourceTypeObj is string resourceType && !string.IsNullOrEmpty(resourceType))
+            {
+                // If wildcard (*) is specified for ResourceType, then set the value to null so that
```
jackliums: We should process this logic inside the Core layer. Otherwise, anyone who needs to implementing their own hosting layer have to implement this logic separately.

src/Microsoft.Health.Fhir.Api/Features/Routing/CompartmentTypesRouteConstraint.cs
```diff
@@ -0,0 +1,31 @@
...
+    public class CompartmentTypesRouteConstraint : IRouteConstraint
+    {
+        public bool Match(HttpContext httpContext, IRouter route, string routeKey, RouteValueDictionary values, RouteDirection routeDirection)
+        {
+            EnsureArg.IsNotNull(httpContext, nameof(httpContext));
+            EnsureArg.IsNotNull(route, nameof(route));
+            EnsureArg.IsNotNullOrEmpty(routeKey, nameof(routeKey));
+            EnsureArg.IsNotNull(values, nameof(values));
+
+            if (values.TryGetValue(KnownActionParameterNames.CompartmentType, out var compartmentTypeObj) && compartmentTypeObj is string compartmentType && !string.IsNullOrEmpty(compartmentType))
```
jackliums: nit: wrap the lines for readability

src/Microsoft.Health.Fhir.Api/Features/Routing/CompartmentTypesRouteConstraint.cs
```diff
@@ -0,0 +1,31 @@
...
+            {
+                return Enum.TryParse<CompartmentType>(compartmentType, false, out var compartmentTypeEnumValue);
```
jackliums: Could just be `return Enum.IsDefined(typeof(CompartmentType), compartmentType);` since it doesn't care about the parsed value?

src/Microsoft.Health.Fhir.Api/Features/Routing/KnownActionParameterNames.cs
```diff
@@ -11,5 +11,6 @@ internal class KnownActionParameterNames
         public const string Resource = "resource";
         public const string Id = "id";
         public const string Vid = "vid";
+        public const string CompartmentType = "compartmenttype";
```
jackliums: nit: compartmentType?

src/Microsoft.Health.Fhir.Core/Features/Search/SearchService.cs
```diff
@@ -56,6 +56,17 @@ protected SearchService(ISearchOptionsFactory searchOptionsFactory, IBundleFacto
             return _bundleFactory.CreateSearchBundle(resourceType, searchOptions.UnsupportedSearchParams, result);
         }
 
+        /// <inheritdoc />
```
jackliums: Do we need new method, handler and Request/Response? Maybe it's time to refactor the code so that search will simply take SearchOptions and move the SearchOptionsFactory to layer above and simplify the code.

src/Microsoft.Health.Fhir.Core/Messages/Search/CompartmentResourceRequest.cs
```diff
@@ -0,0 +1,30 @@
...
+
+using System;
+using System.Collections.Generic;
+using MediatR;
+
+namespace Microsoft.Health.Fhir.Core.Messages.Search
+{
+    public class CompartmentResourceRequest : IRequest<CompartmentResourceResponse>
+    {
+        public CompartmentResourceRequest(string compartmentType, string compartmentId, string resourceType, IReadOnlyList<Tuple<string, string>> queries)
```
jackliums: can these be null?

src/Microsoft.Health.Fhir.Core/Resources.resx
```diff
@@ -157,6 +157,12 @@
   <data name="CompartmentDefinitionIsDupe" xml:space="preserve">
     <value>bundle.entry[{0}].resource has duplicate compartment definitions.</value>
   </data>
+  <data name="CompartmentIdIsInvalid" xml:space="preserve">
+    <value>The compartment id is null or empty.</value>
```
jackliums: I think the convention is to eliminate "the"

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/Queries/QueryBuilder.cs
```diff
@@ -71,6 +90,11 @@ public SqlQuerySpec BuildSqlQuerySpec(SearchOptions searchOptions)
                 return query;
             }
 
+            private static string GetCompartmentIndicesParamName(CompartmentType compartmentType)
+            {
+                return $"{KnownResourceWrapperProperties.CompartmentIndices}.{s_compartmentTypeToParamName[compartmentType]}";
```
jackliums: nit: can we add debug check or something to make sure all of the compartment type has a corresponding name entry? Better than failing with generic ArgumentException.

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/Queries/QueryBuilder.cs
```diff
@@ -134,6 +158,30 @@ private void AppendFilterCondition(string name, object value)
                         .AppendLine(_queryParameterManager.AddOrGetParameterMapping(value));
             }
 
+            private void AppendArrayContainsFilter(string logicalOperator, params (string, string)[] conditions)
```
jackliums: don't really need this method since there is only one condition anyways

src/Microsoft.Health.Fhir.ValueSets/AuditEventSubType.cs
```diff
@@ -35,5 +35,7 @@ public static class AuditEventSubType
         public const string SearchSystem = "search-system";
 
         public const string Capabilities = "capabilities";
+
+        public const string ReadCompartmentByResourceType = "read-compartment";
```
jackliums: this is not a valid value i don't think: https://www.hl7.org/fhir/valueset-audit-event-sub-type.html. I think compartment falls under "search".

src/Microsoft.Health.Fhir.Api/Controllers/FhirController.cs
```diff
@@ -402,6 +408,28 @@ void AddItemsIfNotNull(IEnumerable<KeyValuePair<string, StringValues>> source)
...
+        /// <summary>
+        /// Reads the specified resource.
+        /// </summary>
+        /// <param name="compartmentType">The compartment type.</param>
+        /// <param name="id">The identifier.</param>
+        /// <param name="type">The resource type.</param>
+        [HttpGet]
+        [Route(KnownRoutes.CompartmentTypeByResourceType, Name = RouteNames.ReadCompartmentByResourceType)]
+        [AuditEventSubType(AuditEventSubType.ReadCompartmentByResourceType)]
+        public async Task<IActionResult> ReadCompartmentByResourceType(string compartmentType, string id, string type)
```
jackliums: I think compartment falls under search not read since it returns multiple documents.

test/Microsoft.Health.Fhir.Tests.E2E/Rest/CompartmentTests.cs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            // Delete all relevant resources before starting the test.
+            Client.DeleteAllResources(ResourceType.Device, "_id=d1").Wait();
+            Client.DeleteAllResources(ResourceType.DeviceComponent, "_id=example").Wait();
+            Client.DeleteAllResources(ResourceType.Patient, "_id=f001").Wait();
+            Client.DeleteAllResources(ResourceType.Observation, "_id=f001").Wait();
+            Client.DeleteAllResources(ResourceType.Encounter, "_id=f002").Wait();
+            Client.DeleteAllResources(ResourceType.Condition, "_id=f001").Wait();
+
+            // Create various resources.
```
jackliums: Shoudl use IAsyncLifetime instead for async initializations. I have this fixed on all other classes and I am going to send PR soon.

test/Microsoft.Health.Fhir.Tests.E2E/Rest/CompartmentTests.cs
```diff
@@ -0,0 +1,94 @@
...
+        private readonly Encounter _encounter;
+        private readonly Condition _condition;
+
+        public CompartmentTests(HttpIntegrationTestFixture<Startup> fixture)
+            : base(fixture)
+        {
+            // Delete all relevant resources before starting the test.
+            Client.DeleteAllResources(ResourceType.Device, "_id=d1").Wait();
+            Client.DeleteAllResources(ResourceType.DeviceComponent, "_id=example").Wait();
+            Client.DeleteAllResources(ResourceType.Patient, "_id=f001").Wait();
```
jackliums: All the other tests are using the id generated by the server. Is there any particular reason these needs to be client id? Also, it will be much easier if you just use tags since you could just delete everything matching the tag as opposed to specifying the id and deleting individual items.

src/Microsoft.Health.Fhir.Api/Features/Routing/CompartmentResourceTypesRouteConstraint.cs
```diff
@@ -0,0 +1,39 @@
...
+    public class CompartmentResourceTypesRouteConstraint : IRouteConstraint
+    {
+        public bool Match(HttpContext httpContext, IRouter route, string routeKey, RouteValueDictionary values, RouteDirection routeDirection)
+        {
+            EnsureArg.IsNotNull(httpContext, nameof(httpContext));
+            EnsureArg.IsNotNull(route, nameof(route));
+            EnsureArg.IsNotNullOrEmpty(routeKey, nameof(routeKey));
+            EnsureArg.IsNotNull(values, nameof(values));
+
+            if (values.TryGetValue(KnownActionParameterNames.ResourceType, out var resourceTypeObj) && resourceTypeObj is string resourceType && !string.IsNullOrEmpty(resourceType))
```
jackliums: nit: wrap the lines for easier readability

src/Microsoft.Health.Fhir.Core/Features/Search/SearchService.cs
garamamo: imho, it is cleaner this way, having compartment search its own handlers. Opened an issue to consider this refactor. https://github.com/Microsoft/fhir-server/issues/201.

src/Microsoft.Health.Fhir.Core/Messages/Search/CompartmentResourceRequest.cs
garamamo: The route constraints validate for proper compartment and resource types. The core layer is resilient for these values to be null.

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/Queries/QueryBuilder.cs
garamamo: I added this so that if we have multiple array property values in the index, then we can use this method, sure currently we have only one. I plan to keep it this way unless you feel strongly.

src/Microsoft.Health.Fhir.Api/Features/Routing/CompartmentResourceTypesRouteConstraint.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        public bool Match(HttpContext httpContext, IRouter route, string routeKey, RouteValueDictionary values, RouteDirection routeDirection)
+        {
+            EnsureArg.IsNotNull(httpContext, nameof(httpContext));
+            EnsureArg.IsNotNull(route, nameof(route));
+            EnsureArg.IsNotNullOrEmpty(routeKey, nameof(routeKey));
+            EnsureArg.IsNotNull(values, nameof(values));
+
+            if (values.TryGetValue(KnownActionParameterNames.ResourceType, out var resourceTypeObj)
+                    && resourceTypeObj is string resourceType
```
jackliums: All existing code have convention of && at the end of the line not at the beginning.

src/Microsoft.Health.Fhir.Core/Features/Search/SearchOptionsFactory.cs
```diff
@@ -151,6 +161,29 @@ public SearchOptions Create(string resourceType, IReadOnlyList<Tuple<string, str
 
             options.UnsupportedSearchParams = unsupportedSearchParameters;
 
+            options.CompartmentType = null;
+
+            // Initialize enum to undefined value
+            var parsedCompartmentType = (CompartmentType)(-1);
```
jackliums: you could use nullable here

src/Microsoft.Health.Fhir.Core/Features/Search/SearchOptionsFactory.cs
```diff
@@ -151,6 +161,29 @@ public SearchOptions Create(string resourceType, IReadOnlyList<Tuple<string, str
...
+
+            if (!string.IsNullOrWhiteSpace(compartmentType) &&
+    !Enum.TryParse(compartmentType, out parsedCompartmentType))
+            {
+                throw new NotSupportedException(string.Format(Core.Resources.CompartmentTypeIsInvalid, compartmentType));
```
jackliums: You don't want to throw NotSupportedException since it's not derived from FhirException and which will result in 500. You probably want use InvalidSearchOperationException or a new exception type InvalidCompartmentSearchOperationException?

src/Microsoft.Health.Fhir.Core/Features/Search/SearchOptionsFactory.cs
```diff
@@ -151,6 +161,29 @@ public SearchOptions Create(string resourceType, IReadOnlyList<Tuple<string, str
 
             options.UnsupportedSearchParams = unsupportedSearchParameters;
 
+            options.CompartmentType = null;
+
+            // Initialize enum to undefined value
+            var parsedCompartmentType = (CompartmentType)(-1);
+
+            if (!string.IsNullOrWhiteSpace(compartmentType) &&
+    !Enum.TryParse(compartmentType, out parsedCompartmentType))
```
jackliums: indentation

src/Microsoft.Health.Fhir.Core/Features/Search/SearchOptionsFactory.cs
jackliums: actually you dont really need to declare this since you could just do `Enum.TryParse(compartmentType, out CompartmentType parsedCompartmentType)` below

src/Microsoft.Health.Fhir.Core/Features/Search/SearchOptionsFactory.cs
```diff
@@ -151,6 +161,29 @@ public SearchOptions Create(string resourceType, IReadOnlyList<Tuple<string, str
...
+            {
+                throw new NotSupportedException(string.Format(Core.Resources.CompartmentTypeIsInvalid, compartmentType));
+            }
+
+            if (Enum.IsDefined(typeof(CompartmentType), parsedCompartmentType))
```
jackliums: You don't really need this either since if it was successfully parsed by TryParse above, prasedCompartmentType will have a valid value here.

src/Microsoft.Health.Fhir.Api/Controllers/FhirController.cs
```diff
@@ -402,6 +408,28 @@ void AddItemsIfNotNull(IEnumerable<KeyValuePair<string, StringValues>> source)
...
 
+        /// <summary>
+        /// Reads the specified resource.
+        /// </summary>
+        /// <param name="compartmentType">The compartment type.</param>
+        /// <param name="id">The identifier.</param>
+        /// <param name="type">The resource type.</param>
+        [HttpGet]
+        [Route(KnownRoutes.CompartmentTypeByResourceType, Name = RouteNames.ReadCompartmentByResourceType)]
+        [AuditEventSubType(AuditEventSubType.Search)]
```
jackliums: Additional test in AuditLogginerFilterTest?

src/Microsoft.Health.Fhir.Api/Features/Routing/CompartmentResourceTypesRouteConstraint.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        public bool Match(HttpContext httpContext, IRouter route, string routeKey, RouteValueDictionary values, RouteDirection routeDirection)
+        {
+            EnsureArg.IsNotNull(httpContext, nameof(httpContext));
+            EnsureArg.IsNotNull(route, nameof(route));
+            EnsureArg.IsNotNullOrEmpty(routeKey, nameof(routeKey));
+            EnsureArg.IsNotNull(values, nameof(values));
+
+            if (values.TryGetValue(KnownActionParameterNames.ResourceType, out var resourceTypeObj) &&
+                    resourceTypeObj is string resourceType &&
```
jackliums: nit: inconsistent indentation between this and next file

src/Microsoft.Health.Fhir.Api/Controllers/FhirController.cs
```diff
@@ -402,6 +408,28 @@ void AddItemsIfNotNull(IEnumerable<KeyValuePair<string, StringValues>> source)
...
         }
 
+        /// <summary>
+        /// Reads the specified resource.
+        /// </summary>
+        /// <param name="compartmentType">The compartment type.</param>
+        /// <param name="id">The identifier.</param>
+        /// <param name="type">The resource type.</param>
+        [HttpGet]
+        [Route(KnownRoutes.CompartmentTypeByResourceType, Name = RouteNames.ReadCompartmentByResourceType)]
```
jackliums: this should also change to Search instead of Read right?

src/Microsoft.Health.Fhir.Core/Features/Search/SearchOptionsFactory.cs
```diff
@@ -43,6 +43,16 @@ public SearchOptions Create(IReadOnlyList<Tuple<string, string>> queryParameters
...
+        {
+            return Create(null, null, resourceType, queryParameters);
+        }
+
+        public SearchOptions Create(string compartmentType, string compartmentId, IReadOnlyList<Tuple<string, string>> queryParameters)
+        {
+            return Create(compartmentType, compartmentId, null, queryParameters);
+        }
+
+        public SearchOptions Create(string compartmentType, string compartmentId, string resourceType, IReadOnlyList<Tuple<string, string>> queryParameters)
```
jackliums: Do we need this? It's not called by anything and it can be called with the below method with resourceType set to null.

test/Microsoft.Health.Fhir.Tests.E2E/Rest/CompartmentTests.cs
```diff
@@ -0,0 +1,115 @@
...
+        {
+            _tagValue = $"CompartmentTests_{Guid.NewGuid().ToString()}";
+            _tagQueryString = $"{TagSearchParameterName}={_tagValue}";
+        }
+
+        [Fact]
+        [Trait(Traits.Priority, Priority.One)]
+        public async Task GivenADeviceCompartment_WhenRetrievingDeviceComponents_ThenOnlyResourcesMatchingCompartmentShouldBeReturned()
+        {
+            Bundle bundle = await Client.SearchAsync("Device/d1/DeviceComponent");
```
jackliums: I don't think these test works. If you are creating a new resource now, you will get new server generated id but these tests still uses hard-coded client id?

garamamo: These tests still work because while creating reference search indices we do not check the existence of the reference resource. I will change to using PUT instead of POST to avoid the confusion.

cunninghamjc:
Resolves #51

PR #204, opened 2018-11-14
Improves read and vread performance

johnstairs:
Addresses some low-hanging fruit in read (e.g. `GET /Observation/id`) and vread (e.g. `GET Observation/id/_history/1`) performance.

Average times in ms for one-box, measured from the client:

|       | Before | After |
|-------|--------|-------|
| __Read__  | 8.1    | 3.1   |
| __VRead__ | 11.5   | 5.5   |


johnstairs:
#206

PR #207, opened 2018-11-15
Create user story template

cunninghamjc:
Created a custom issue template for user stories.

### DIFF
--- a/.github/ISSUE_TEMPLATE/user-story.md
+++ b/.github/ISSUE_TEMPLATE/user-story.md
@@ -1,0 +1,12 @@
+---
+name: User story
+about: Tell us something new you want to do
+labels: 
+
+---
+
+**User story**
+As a [type of user], I want [some functionality] so that [benefit].
+
+**Acceptance criteria**
+1. When I do [A], then [B] happens.


.github/ISSUE_TEMPLATE/user-story.md
```diff
@@ -0,0 +1,12 @@
+---
+name: User story
```
jackliums: No bug template?

cunninghamjc: I already submitted that: #205

.github/ISSUE_TEMPLATE/user-story.md
```diff
@@ -0,0 +1,12 @@
...
+about: Tell us something new you want to do
+labels:
```
jackliums: Do we want user to enter type labels here?

cunninghamjc: No, the user doesn't actually see or edit this part; it seems to be metadata used by github on the "new issue" page.  I haven't tried adding labels to the template to see what it does, but I would guess that it adds these labels automatically to the issue.  I don't think we'd do it for user stories, but we could try it on bugs.

PR #223, opened 2018-11-28
Handle NullReferenceException in Fhir Server

poadhika:
Bug fix for the following bug:
Bug 66743: NullReferenceException in Fhir server
Link: https://microsofthealth.visualstudio.com/Health/_workitems/edit/66743?src=WorkItemMention&src-action=artifact_link


src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/DocumentClientWithExceptionHandler.Generated.cs
```diff
@@ -0,0 +1,2291 @@
+//------------------------------------------------------------------------------
```
johnstairs: This file should be deleted.

src/Microsoft.Health.Fhir.Api/Features/Filters/ValidateContentTypeFilterAttribute.cs
```diff
@@ -44,21 +44,28 @@ public override async Task OnActionExecutionAsync(ActionExecutingContext context
...
+                            {
+                                case RouteNames.SearchResourcesPost:
+                                case RouteNames.SearchAllResourcesPost:
+                                    break;
+                                default:
+                                    throw new UnsupportedMediaTypeException(Resources.UnsupportedContentTypeHeader);
+                            }
                         }
                     }
+                    catch (NullReferenceException)
```
johnstairs: We generally don't want to be catching this kind of exception, but instead should use a conditional expression before dereferencing.

src/Microsoft.Health.Fhir.Api.UnitTests/Features/Filters/ValidateContentTypeFilterAttributeTests.cs
```diff
@@ -130,6 +130,23 @@ public async Task GivenARequestWithNotAllowedContentTypeHeader_WhenValidatingThe
             await Assert.ThrowsAsync<UnsupportedMediaTypeException>(async () => await filter.OnActionExecutionAsync(context, actionExecutedDelegate));
         }
 
+        [Theory]
+        [InlineData("\"application/fhir+json")]
+        [InlineData("]application/fhir+json")]
+        [InlineData("application\blah")]
```
brendankowitz: If the test logic is the same, perhaps just add these new cases to the test above. Maybe the test method name could begin with something like `GivenARequestWithUnsupportedContentTypeHeader`

poadhika: Sure. I will add the new  cases to already existing test. Thank you.

poadhika:
Abandoning this pull request as the change has been made in the Fhir API already

Issue #228, opened 2018-12-04
Remove IS_DEFINED from Cosmos query


jackliums:
**Describe the bug**

IS_DEFINED does not utilize index and therefore will slow down query.

PR #234, opened 2018-12-04
Removing IS_DEFINED clause

poadhika:
This change is related to the bug:  https://github.com/Microsoft/fhir-server/issues/228

### DIFF
--- a/src/Microsoft.Health.Fhir.CosmosDb/Features/Search/Queries/QueryBuilder.cs
+++ b/src/Microsoft.Health.Fhir.CosmosDb/Features/Search/Queries/QueryBuilder.cs
@@ -3,11 +3,8 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
-using System.Collections.Generic;
-using System.Diagnostics;
 using System.Text;
 using EnsureThat;
-using Hl7.Fhir.Model;
 using Microsoft.Azure.Documents;
 using Microsoft.Health.Fhir.Core.Features.Persistence;
 using Microsoft.Health.Fhir.Core.Features.Search;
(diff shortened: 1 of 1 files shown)


src/Microsoft.Health.Fhir.CosmosDb/Features/Search/Queries/QueryBuilder.cs
```diff
@@ -148,9 +145,6 @@ private void AppendSystemDataFilter(string keyword = null)
 
                 _queryBuilder
                     .Append("(")
```
jackliums: We don't need this extra parenthese anymore since there is only a single condition. Also, i would remove the keyword parameter above and just directly put "WHERE" instead here since it's no longer needed to have different keyword.

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/Queries/QueryBuilder.cs
```diff
@@ -137,23 +134,12 @@ private void AppendFilterCondition(string name, object value)
...
                 _queryBuilder
-                    .Append("(")
-                    .Append("IS_DEFINED(").Append(SearchValueConstants.RootAliasName).Append(".isSystem)")
-                    .Append(" = ").Append(_queryParameterManager.AddOrGetParameterMapping(false))
-                    .Append(" OR ")
+                    .Append(" WHERE ")
                     .Append(SearchValueConstants.RootAliasName).Append(".isSystem")
-                    .Append(" = ").Append(_queryParameterManager.AddOrGetParameterMapping(false))
-                    .AppendLine(")");
+                    .Append(" = ").Append(_queryParameterManager.AddOrGetParameterMapping(false));
```
jackliums: you should make the send `Append` into AppendLine.

poadhika: Sure. I will update the PR.

Issue #21, opened 2018-09-18
Invalid prefix in quantity search results in 500 error


johnstairs:
http://localhost:53727/Observation?component-value-quantity=lt109 => 200
http://localhost:53727/Observation?component-value-quantity=ll109 => 500

The second should return a 400

PR #264, opened 2018-12-07
Throw BadRequest Exception in QuantitySearchValue on malformed input. Fixes #196, #21.

### LINKED ISSUE #196: Search should return 400 instead of 500 when there is FormatException or other exceptions that causes the parsing to fail
If the parsing of the search parameter value fails because of formatting issue or other issue, the search should return 400 instead of 500.

JeffreyTaylor:
## Description
When `parse` fails in QuantitySearchValue, throw a `BadRequestException` exception. This results in an OperationOutcome & 400 status code returned to the user.

## Related issues
#196, #21

## Testing
Updated tests that test this piece of code to now expect a `BadRequestException`.

## Notes
I notice `BadRequestException` is in the `Persistence` namespace. I think it probably should be moved into the `Microsoft.Health.Fhir.Core.Exceptions` namespace.


src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/QuantitySearchValue.cs
```diff
@@ -63,14 +64,21 @@ public static QuantitySearchValue Parse(string s)
...
-            string system = parts.Count > 1 ? parts[1] : string.Empty;
-            string code = parts.Count > 2 ? parts[2] : string.Empty;
-
-            return new QuantitySearchValue(
-                system.UnescapeSearchParameterValue(),
-                code.UnescapeSearchParameterValue(),
-                quantity);
+            try
+            {
+                decimal quantity = decimal.Parse(parts[0], NumberStyles.Number, CultureInfo.InvariantCulture);
```
brandonpollett: If the goal is to catch errors with `Parse`, could we use `TryParse` to catch these issues?

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/QuantitySearchValue.cs
```diff
@@ -63,14 +64,21 @@ public static QuantitySearchValue Parse(string s)
...
+                string system = parts.Count > 1 ? parts[1] : string.Empty;
+                string code = parts.Count > 2 ? parts[2] : string.Empty;
+
+                return new QuantitySearchValue(
+                    system.UnescapeSearchParameterValue(),
+                    code.UnescapeSearchParameterValue(),
+                    quantity);
+            }
+            catch
```
johnstairs: We should avoid catch-all blocks

brandonpollett:
This is related to #196.

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/QuantitySearchValue.cs
```diff
@@ -63,14 +64,21 @@ public static QuantitySearchValue Parse(string s)
...
+            {
+                throw new BadRequestException($"Malformed search value \"{parts[0]}\"");
```
brandonpollett: This text should be in the corresponding resource file.

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/QuantitySearchValue.cs
JeffreyTaylor: Good call. I updated this to now use `TryParse`.

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/QuantitySearchValue.cs
JeffreyTaylor: Yep, that makes sense. Using `TryParse` solved this issue. This catch-all block has been removed.

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/QuantitySearchValue.cs
JeffreyTaylor: Oh - I missed that. This has been fixed. Now using `Core.Resources.MalformedSearchValue`

src/Microsoft.Health.Fhir.Core/Resources.resx
```diff
@@ -326,4 +326,7 @@
   <data name="WeakETagFormatRequired" xml:space="preserve">
     <value>WeakETag must be in the weak ETag format.</value>
   </data>
+    <data name="MalformedSearchValue" xml:space="preserve">
```
brandonpollett: nit: This file is alphabetized by name to make future diffs easier.

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/QuantitySearchValue.cs
```diff
@@ -63,7 +64,12 @@ public static QuantitySearchValue Parse(string s)
                 throw new FormatException(Core.Resources.MoreThanTwoTokenSeparatorSpecified);
             }
 
-            decimal quantity = decimal.Parse(parts[0], NumberStyles.Number, CultureInfo.InvariantCulture);
+            decimal quantity;
+            if (decimal.TryParse(parts[0], NumberStyles.Number, CultureInfo.InvariantCulture, out quantity) == false)
```
brandonpollett: Elsewhere in the code, we generally use the `!` operator when testing for false on `TryParse`.  
```suggestion
            if (!decimal.TryParse(parts[0], NumberStyles.Number, CultureInfo.InvariantCulture, out quantity))
```

PR #267, opened 2018-12-08
Fix TestEnvironmentFilePath to resolve across platforms. Fixes #266.

### LINKED ISSUE #266: Identity Server 404 on OSX
**Describe the bug**
When running on OSX you receive a 404 when going to /connect/token

**To Reproduce**
1. Run on mac
2. Go to /connect/token
3. Receive a 404

**Expected behavior**
Ability to a token

**Actual behavior**
404 is returned.

JeffreyTaylor:
## Description
Updates paths in `launchSettings.json` to ensure file resolves on both windows and macosx.

## Related issues
Addresses #266.

## Testing
Tested by navigating to `/connect/token` on both a windows computer and a macbook. Both returned correct responses.

### DIFF
--- a/src/Microsoft.Health.Fhir.Web/DevelopmentIdentityProviderRegistrationExtensions.cs
+++ b/src/Microsoft.Health.Fhir.Web/DevelopmentIdentityProviderRegistrationExtensions.cs
@@ -122,6 +122,7 @@ namespace Microsoft.Health.Fhir.Web
             }
 
             string normalizedPath = testEnvironmentFilePath.Replace('\\', Path.DirectorySeparatorChar);
+            testEnvironmentFilePath = Path.GetFullPath(normalizedPath);
             if (!File.Exists(testEnvironmentFilePath))
             {
                 return configurationBuilder;
--- a/src/Microsoft.Health.Fhir.Web/DevelopmentIdentityProviderRegistrationExtensions.cs
+++ b/src/Microsoft.Health.Fhir.Web/DevelopmentIdentityProviderRegistrationExtensions.cs
@@ -121,7 +121,8 @@ namespace Microsoft.Health.Fhir.Web
                 return configurationBuilder;
             }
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Fhir.Web/DevelopmentIdentityProviderRegistrationExtensions.cs
```diff
@@ -121,7 +121,8 @@ public static IConfigurationBuilder AddDevelopmentAuthEnvironment(this IConfigur
                 return configurationBuilder;
             }
 
-            testEnvironmentFilePath = Path.GetFullPath(testEnvironmentFilePath);
+            string normalizedPath = testEnvironmentFilePath.Replace('\\', Path.DirectorySeparatorChar);
```
brandonpollett: There's not a huge need for a new variable here, especially since we're already transforming it below. We should be able to do something like the following instead. 

```suggestion
            testEnvironmentFilePath = testEnvironmentFilePath.Replace('\\', Path.DirectorySeparatorChar);
            testEnvironmentFilePath = Path.GetFullPath(testEnvironmentFilePath);
```

Also, could probably just be done on one line, but readability may suffer.

JeffreyTaylor: Thanks for the input. Good points - I agree. I removed the extra variable and ended up one-lining this.

brandonpollett:
In thinking about this some more, I think we can probably just update the path in the `launchsettings.json` to be "..//..//testauthenvironment.json" and it will work on both.

JeffreyTaylor:
> In thinking about this some more, I think we can probably just update the path in the `launchsettings.json` to be "..//..//testauthenvironment.json" and it will work on both.

Yep - you're 100% right. I was not previously aware that `..//..//` would behave the same as `../../` on a mac, but sure enough it does. Good call - I learned something, and we arrived at a great solution. :)

Issue #222, opened 2018-11-27
Azure Active Directory Proxy for SMART on FHIR


hansenms:
Create a proxy for Azure AD that enable SMART on FHIR Apps to work seamlessly with the Microsoft FHIR Server for Azure

PR #268, opened 2018-12-08
Personal/mihansen/aad proxy

hansenms:
## Description
Introduces SMART on FHIR AAD Proxy

## Related issues
Addresses #222 

## Testing
E2E Tests using Selenium Added.


src/Microsoft.Health.Fhir.Api/Controllers/AadProxyController.cs
```diff
@@ -0,0 +1,363 @@
...
+        {
+            EnsureArg.IsNotNull(grantType, nameof(grantType));
+            EnsureArg.IsNotNull(clientId, nameof(clientId));
+
+            var client = _httpClientFactory.CreateClient();
+
+            // TODO: Add handling of 'aud' -> 'resource'
+            if (grantType != "authorization_code")
+            {
+                List<KeyValuePair<string, string>> fields = new List<KeyValuePair<string, string>>();
```
jmandel: Quick question: am I right in understanding this is a guaranteed failure condition?

If so, why not just return an "unsupported grant type" error? (And if not, which other grant types do we expect here?)

samples/apps/SmartLauncher/appsettings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "FhirServerUrl": "https://localhost:44348",
+    "ClientId": "APP-ID",
+    "SmartAppUrl": "/sampleapp/launch.html"
```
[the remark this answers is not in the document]
hansenms: Good point. I have changed it to `DefaultSmartAppUrl`, I was thinking of providing more of an array of possible apps and a dropdown box, but that would sort of limit it to what was defined in the app. Maybe room for some improvements down the line.

src/Microsoft.Health.Fhir.Api/Controllers/AadProxyController.cs
hansenms: Actually, that is not a guaranteed failure. AAD supports client_credentials, etc. and we don't want to "disable" those features just because we enable SMART proxy. It would actually cause us a lot of problems if we did because many of the E2E tests obtain tokens that way. So this is sort of a bypass; if it is not authorization code, just pass on to AAD>

hansenms:
I think it is time to let the review process start on this. Some comments:

The AAD proxy works by:
1. Proxying the request to AADs `/authorize` endpoint. In that process, the `aud=` parameter is renamed to `resource=` and the `launch=` context parameter is appended to the `state=` parameter (which means it will be returned with authorization code. 
2. The Reply URL is changed from `https://myapp/reply` to `https://fhir-server/AadProxy/callback/<Base64(https://myapp/reply)>`
3. The authorization code is returned to the AAD proxy where the launch context is unpacked from the state parameter and a compound authorization code is created. This compound code contains both the authorization code and the launch context.
4. The AAD proxy sends the compund code to the SMART app reply URL, `https://myapp/reply` 
5. The SMART App sends the compound code to `https://fhir-server/AadProxy/token`.
6. The proxy unpacks the compound code and send sthe original authorization code to AAD, where it obtains a token.
7. Selected fields from the launch context are added to the token response and returned to the SMART App.

There will be a follow up PR to once all of this is merged. This PR, will: 
1. remove the default CORS settings (`https://localhost:6001`) from the template and add the CORS settings as template deployment parameter in the build system.
2. Disable the AAD proxy as a default setting and add it as a deployment parameter in the build systemt.

@johnstairs The launch context is now passed as part of the `state=` parameter in the authorization request. This removes the nasty wildcard reply URL which was previously required. I also experimented with URL encoding (instead of base64) the app reply URL in the proxy reply URL, but that doesn't work. However, current scheme *should* be B2C compatible, albeit with some readability and possible case sensitivity issues. Room for improvement.

src/Microsoft.Health.Fhir.Api/Controllers/AadProxyController.cs
```diff
@@ -0,0 +1,355 @@
...
+        /// <summary>
+        /// Callback function for receiving code from AAD
+        /// </summary>
+        /// <param name="encodedRedirect">Base64 encoded redirect URL on the app.</param>
+        /// <param name="code">Authorization code.</param>
+        /// <param name="state">state URL parameter.</param>
+        /// <param name="sessionState">session_state URL parameter.</param>
+        /// <param name="error">error URL parameter.</param>
+        /// <param name="errorDescription">error_description URL parameter.</param>
+        [HttpGet("callback/{encodedRedirect}")]
```
johnstairs: I'm not sure this will work properly if the base64 encoding contains a /.  It might be safer to adopt a url safe base64 encoding.

src/Microsoft.Health.Fhir.Api/Controllers/AadProxyController.cs
```diff
@@ -0,0 +1,355 @@
...
+            var openIdConfigurationUrl = $"{_securityConfiguration.Authentication.Authority}/.well-known/openid-configuration";
+
+            HttpResponseMessage openIdConfigurationResponse;
+            using (var httpClient = httpClientFactory.CreateClient())
+            {
+                try
+                {
+                    openIdConfigurationResponse = httpClient.GetAsync(new Uri(openIdConfigurationUrl)).GetAwaiter().GetResult();
+                }
+                catch (Exception ex)
```
johnstairs: We should avoid catching all exceptions types.

src/Microsoft.Health.Fhir.Api/Controllers/AadProxyController.cs
```diff
@@ -0,0 +1,355 @@
...
+                {
+                    logger.LogWarning(ex, $"There was an exception while attempting to read the OpenId Configuration from \"{openIdConfigurationUrl}\".");
+                    throw new OpenIdConfigurationException();
+                }
+            }
+
+            if (openIdConfigurationResponse.IsSuccessStatusCode)
```
johnstairs: And if we don't get success? Seems like we should throw as we do above.

johnstairs: We could also wrap this in a Polly retry policy...

src/Microsoft.Health.Fhir.Api/Controllers/AadProxyController.cs
johnstairs: Also, since they often end with =, this could be problematic when used in query strings.

src/Microsoft.Health.Fhir.Api/Controllers/AadProxyController.cs
```diff
@@ -0,0 +1,355 @@
...
+                    }
+                }
+
+                newScopes = newScopes.TrimEnd(' ');
+                newQueryString += $"&scope={Uri.EscapeDataString(newScopes)}";
+            }
+
+            newQueryString += $"&state={newState}";
+
+            return RedirectPermanent($"{_aadAuthorizeEndpoint}?{newQueryString}");
```
johnstairs: Is this supposed to be a permanent redirect?

src/Microsoft.Health.Fhir.Api/Controllers/AadProxyController.cs
```diff
@@ -0,0 +1,355 @@
...
+                newQueryString += $"&resource={aud}";
+            }
+            else
+            {
+                // Azure AD v2.0 uses fully qualified scopes and does not allow '/' (slash)
+                // We add qualification to scopes and replace '/' -> '$'
+
+                EnsureArg.IsNotNull(scope, nameof(scope));
+                var scopes = scope.Split(' ');
+                string newScopes = string.Empty;
```
johnstairs: Should use a stringbuilder instead of repeated string concatenation

src/Microsoft.Health.Fhir.Api/Controllers/AadProxyController.cs
johnstairs: I think we should change the tests to use an environment variable instead and make this case an error as Josh suggests. Not in this PR though...

hansenms: @johnstairs, I agree, I will put a TODO note in the code to remove this. I think we have will some code changes related to multiple IDPs that will affect this too. So I will mark the code for now.

src/Microsoft.Health.Fhir.Api/Controllers/AadProxyController.cs
hansenms: I have switched to using `Base64UrlEncoder` throughout the code and the `New-FhirServerSmartClientReplyUrl` has been updated accordingly.

src/Microsoft.Health.Fhir.Api/Controllers/AadProxyController.cs
hansenms: I adopted this pattern from `CapabilityStatementExtensions.cs`. There could be a few different things to catch. Do you propose we walk through all of them?

src/Microsoft.Health.Fhir.Api/Controllers/AadProxyController.cs
hansenms: Switched to `Redirect`

hansenms:
@johnstairs, I switched to `Base64UrlEncoder` and addressed most other issues. What are your thoughts on the `catch (Exception ec) {}` when trying to get the OpenIdConfiguration statement. What should we be catching?

hansenms:
@johnstairs, Also added a feature filter to deny access to the controller unless the feature is switched on. It could have been done simpler but it should do the trick. Let me know what you think.

samples/apps/SmartLauncher/.gitignore
```diff
@@ -0,0 +1,2 @@
+bin/
+obj/
```
johnstairs: Is this file necessary? We already have this in the repo root.

samples/apps/SmartLauncher/Startup.cs
```diff
@@ -0,0 +1,58 @@
...
+        // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
+        public void Configure(IApplicationBuilder app, IHostingEnvironment env)
+        {
+            if (env.IsDevelopment())
+            {
+                app.UseDeveloperExceptionPage();
+            }
+
+            app.UseDefaultFiles();
+            app.UseStaticFiles(new StaticFileOptions { FileProvider = new EmbeddedFileProvider(Assembly.GetExecutingAssembly(), "SmartLauncher.wwwroot") });
```
johnstairs: You don't really need the embedded file provider here. It could work just as well with files that are copied to the working directory, which is the default.

johnstairs: Or maybe i'm missing something.

hansenms: If we don't use the embedded files, we need to a) publish the SmartLauncher when we build and b) explicitly set to root folder when using the Startup in the E2E test. This approach means that we don't need to do any of those.

johnstairs: I see, we use it from the E2E tests. Might not be so obvious that something under samples is used by the tests. 
Anyway, it's fine.

hansenms: Yeah, the dual use is less than obvious. I could have kept it entirely in the tests, but I thought it might serve as an example as well.

samples/apps/SmartLauncher/wwwroot/index.html
```diff
@@ -0,0 +1,99 @@
...
+        {
+            launchContext.encounter = $('#encounter').val();
+        }
+
+        if ($('#practitioner').val().length)
+        {
+            launchContext.practioner = $('#practitioner').val();
+        }
+
+        var queryString = 'launch=' + encodeURIComponent(btoa(JSON.stringify(launchContext)));
```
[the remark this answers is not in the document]
hansenms: We will do something on this at a later point per offline discussion.

samples/apps/SmartLauncher/wwwroot/index.html
```diff
@@ -0,0 +1,99 @@
...
+            <label for="appurl">Appliction URL:</label>
+            <input type="text" id="appurl" class="form-control" />
+            <label for="fhirurl">FHIR Server URL:</label>
+            <input type="text" id="fhirurl" class="form-control" />
+        </div>
+    </div>
+    <div class="row">
+        <h2>Launch context</h2>
+        <div class="form-group col-xs-8">
+            <textarea class="form-control" rows="5" id="launchContext" >
```
johnstairs: Did you consider making this readonly, since edits to it are ignored?

samples/templates/default-azuredeploy.json
```diff
@@ -40,6 +40,13 @@
             ],
             "defaultValue": "S1"
         },
+        "allowedOrigins": {
+            "type": "array",
+            "defaultValue": [ "https://localhost:6001" ],
```
johnstairs: Reminder to make this empty.

src/Microsoft.Health.Fhir.Api/Controllers/AadProxyController.cs
```diff
@@ -0,0 +1,354 @@
...
+using Microsoft.IdentityModel.Tokens;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Health.Fhir.Api.Controllers
+{
+    /// <summary>
+    /// Controller class enabling Azure Active Directory SMART on FHIR Proxy Capability
+    /// </summary>
+    [TypeFilter(typeof(AadProxyFeatureFilterAttribute))]
+    [Route("/AadProxy")]
```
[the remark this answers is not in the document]
hansenms: We could....and should we store that path segment somewhere so that if and when we want to change it, we will do it in only one place. Do we have such a place?

src/Microsoft.Health.Fhir.Api/Controllers/AadProxyController.cs
johnstairs: HttpRequestException and OperationCanceledException should cover it. Also, I'd just do `openIdConfigurationResponse.EnsureSuccessStatusCode()` instead of the if block below.

src/Microsoft.Health.Fhir.Api/Controllers/AadProxyController.cs
```diff
@@ -0,0 +1,354 @@
...
+                }
+            }
+            else
+            {
+                throw new OpenIdConfigurationException();
+            }
+        }
+
+        /// <summary>
+        /// Proxies a request to the Azure AD authorize endpoint.
```
johnstairs: Might be more accurate to say that it redirects to the AAD authorize endpoint with adjusted parameters.

test/Microsoft.Health.Fhir.Tests.E2E/SmartProxy/SmartProxyTestFixture.cs
```diff
@@ -0,0 +1,74 @@
...
+
+namespace Microsoft.Health.Fhir.Tests.E2E.SmartProxy
+{
+    public class SmartProxyTestFixture : IDisposable
+    {
+        private string _environmentUrl;
+
+        public SmartProxyTestFixture()
+        {
+            _environmentUrl = Environment.GetEnvironmentVariable("TestEnvironmentUrl");
```
johnstairs: This is not defined by default.

test/Microsoft.Health.Fhir.Tests.E2E/SmartProxy/SmartProxyTestFixture.cs
```diff
@@ -0,0 +1,74 @@
...
+                {
+                    config.AddEnvironmentVariables();
+                })
+                .UseStartup<SmartLauncher.Startup>()
+                .UseUrls(baseUrl);
+
+            WebServer = builder.Build();
+            WebServer.Start();
+
+            // _server = new TestServer(builder);
```
johnstairs: remvoe

src/Microsoft.Health.Fhir.Api/Controllers/AadProxyController.cs
```diff
@@ -0,0 +1,354 @@
...
+            string newState;
+            try
+            {
+                JObject launchStateParameters = JObject.Parse(Base64UrlEncoder.Decode(state));
+                JObject launchParameters = JObject.Parse(Base64UrlEncoder.Decode(launchStateParameters["l"].ToString()));
+                launchParameters.Add("code", code);
+                newState = launchStateParameters["s"].ToString();
+                compoundCode = Base64UrlEncoder.Encode(launchParameters.ToString(Newtonsoft.Json.Formatting.None));
+            }
+            catch
```
johnstairs: It would be better to return a 400 in this case

src/Microsoft.Health.Fhir.Api/Controllers/AadProxyController.cs
```diff
@@ -0,0 +1,354 @@
...
+                Content = tokenResponse.ToString(Newtonsoft.Json.Formatting.None),
+                StatusCode = (int)response.StatusCode,
+                ContentType = "application/json",
+            };
+        }
+
+        private static bool IsAbsoluteUrl(string url)
+        {
+            Uri result;
+            return Uri.TryCreate(url, UriKind.Absolute, out result);
```
johnstairs: Syntactic sugar: you could write `return Uri.TryCreate(url, UriKind.Absolute, out _);`

Issue #276, opened 2018-12-10
Store Identity Provider information in datastore


brandonpollett:
**User story**
As an admin, I want to persist identity provider information so that the information can be managed through the runtime.

**Acceptance criteria**
- [ ] Identity provider information is pulled from datastore
- [ ] Service accepts new identity providers
- [ ] Service allows you to update identity providers
- [ ] Service allows you to remove identity providers
- [ ] Appropriate logging is in place

PR #288, opened 2018-12-14
Refactor to have base cosmosdb project to support multiple collections

brandonpollett:
## Description
This PR is set the groundwork for having multiple collections in the same Cosmos DB database to support the persistence of control plane and FHIR data independently. It also includes some basic implementation of the control plane service to demonstrate how different collections would work together. Additionally, it adds configuration options to initialize a database with shared throughput and also the ability to initialize collections with a set throughput. 

## Related issues
Base work for #276, Address #290 

## Testing
Since this is largely a refactor, the testing was primarily that the existing integrations tests do not fail.


src/Microsoft.Health.Fhir.Web/Startup.cs
```diff
@@ -21,9 +23,9 @@ public Startup(IConfiguration configuration)
         // This method gets called by the runtime. Use this method to add services to the container.
         public virtual void ConfigureServices(IServiceCollection services)
         {
-            services.AddFhirServer(Configuration).AddCosmosDb();
+            services.AddCosmosDb().AddCosmosControlPlane(Configuration).AddDevelopmentIdentityProvider(Configuration);
```
brandonpollett: This should be refactored to automatically register the appropriate base objects if either the control plane cosmos db or the fhir server cosmos db is added.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/FhirCollectionInitializer.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace Microsoft.Health.Fhir.CosmosDb.Features.Storage
+{
+    public class FhirCollectionInitializer : CollectionInitializer
+    {
+        public FhirCollectionInitializer(CosmosDataStoreConfiguration cosmosDataStoreConfiguration, FhirCollectionUpgradeManager fhirCollectionUpgradeManager)
+        {
+            EnsureArg.IsNotNull(cosmosDataStoreConfiguration, nameof(cosmosDataStoreConfiguration));
+            EnsureArg.IsNotNull(fhirCollectionUpgradeManager, nameof(fhirCollectionUpgradeManager));
+
+            CollectionId = cosmosDataStoreConfiguration.FhirCollectionId;
```
brandonpollett: Look at using a base constructor here instead of setting the properties.

src/Microsoft.Health.Fhir.Web/appsettings.json
```diff
@@ -25,7 +25,11 @@
         "Host": null,
         "Key": null,
         "DatabaseId": "health",
-        "CollectionId": "fhir",
+        "InitialDatabaseThroughput": null,
```
brandonpollett: The PR release didn't work when this was set to 400. I will need to look into why this failed going to an actual Cosmos DB account, but worked when going to the local emulator.

src/Microsoft.Health.ControlPlane.Core/Features/Rbac/IdentityProvider.cs
```diff
@@ -0,0 +1,31 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+
+namespace Microsoft.Health.ControlPlane.Core.Features.Rbac
+{
+    public class IdentityProvider
+    {
+        public IdentityProvider()
```
jackliums: We should make the class immutable unless there is a need to have default constructor and have all of the properties settable from outside. If this is for the serialization, there is a way to work around this without making the class mutable.

src/Microsoft.Health.ControlPlane.Core/Features/Rbac/IdentityProvider.cs
```diff
@@ -0,0 +1,31 @@
...
+            Name = name;
+            Authority = authority;
+            Audience.AddRange(audience);
+        }
+
+        public string Name { get; set; }
+
+        public string Authority { get; set; }
+
+        public List<string> Audience { get; } = new List<string>();
```
jackliums: Prefer IReadOnlyCollection or IReadOnlyList?

src/Microsoft.Health.ControlPlane.Core/Features/Rbac/RbacService.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using EnsureThat;
+using Microsoft.Health.ControlPlane.Core.Features.Persistence;
+
+namespace Microsoft.Health.ControlPlane.Core.Features.Rbac
+{
+    public class RbacService : IRbacService
+    {
+        private IControlPlaneDataStore _controlPlaneDataStore;
```
jackliums: nit: readonly

src/Microsoft.Health.ControlPlane.Core/Features/Rbac/RbacService.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        public RbacService(IControlPlaneDataStore controlPlaneDataStore)
+        {
+            EnsureArg.IsNotNull(controlPlaneDataStore, nameof(controlPlaneDataStore));
+
+            _controlPlaneDataStore = controlPlaneDataStore;
+        }
+
+        public async Task<IdentityProvider> GetIdentityProviderAsync(string name, CancellationToken cancellationToken)
```
jackliums: Is the idea to have additional validation and/or etc here in the future and if so is that implemented in a different PR?

src/Microsoft.Health.ControlPlane.Core/Microsoft.Health.ControlPlane.Core.csproj
```diff
@@ -0,0 +1,23 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
```
jackliums: nit: maybe change to netstandard so that it can be consumed by not just .net core in the future?

src/Microsoft.Health.CosmosDb/Features/Storage/CollectionInitializer.cs
```diff
@@ -0,0 +1,55 @@
...
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.Azure.Documents;
+using Microsoft.Azure.Documents.Client;
+using Microsoft.Health.CosmosDb.Features.Storage.Versioning;
+
+namespace Microsoft.Health.CosmosDb.Features.Storage
+{
+    public class CollectionInitializer : ICollectionInitializer
```
jackliums: The properties don't need to be publicly settable.

1. Make CollectionInitializer abstract (since this doesn't seem to be used directly).
2. Create a protected constructor that takes CollectionId, RelativeDatabaseUri, and etc.
3. Make the properties public getter and private setter.
4. Make the ControlPlaneCollectionInitialier and FhirCollectionInitializer call the protected constructor.

brandonpollett: Refactored the `CollectionInitializer` and got rid of the public properties since they aren't needed.

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/CosmosIdentityProviderExtensions.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Microsoft.Health.ControlPlane.CosmosDb.Features.Storage
+{
+    public static class CosmosIdentityProviderExtensions
+    {
+        public static IdentityProvider ToIdentityProvider(this CosmosIdentityProvider cosmosIdentityProvider)
+        {
+            return new IdentityProvider(cosmosIdentityProvider.Name, cosmosIdentityProvider.Authority, cosmosIdentityProvider.Audience)
+            {
+                Version = cosmosIdentityProvider.ETag.Trim('"'),
```
jackliums: You could make the Version property virtual in the IdentityProvider class and override it in CosmosIdentityProvider with this logic as oppose to create a new IdentityProvider from CosmosIdentityProvider.

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/ControlPlaneDataStore.cs
```diff
@@ -0,0 +1,136 @@
...
+        }
+
+        public async Task<IdentityProvider> GetIdentityProviderAsync(string name, CancellationToken cancellationToken)
+        {
+            EnsureArg.IsNotNull(name, nameof(name));
+
+            var identityProvider = await GetSystemDocumentByIdAsync<CosmosIdentityProvider>(name, CosmosIdentityProvider.IdentityProviderPartition, cancellationToken);
+            if (identityProvider == null)
+            {
+                throw new Exception("Invalid role name");
```
jackliums: We should be throwing correct exception type not the uber Exception type.

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/Versioning/ControlPlaneCollectionUpgradeManager.cs
```diff
@@ -0,0 +1,75 @@
...
+    {
+        private readonly IEnumerable<ICollectionUpdater> _collectionUpdater;
+        private readonly CosmosDataStoreConfiguration _configuration;
+
+        /// <summary>
+        /// This integer should be incremented when changing any value in the
+        /// UpdateIndexAsync function
+        /// </summary>
+        internal const int CollectionSettingsVersion = 1;
+
```
jackliums: nit: i would move these up with the two entries above.

src/Microsoft.Health.CosmosDb/Exceptions/CosmosDbException.cs
```diff
@@ -0,0 +1,14 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+
+namespace Microsoft.Health.CosmosDb.Exceptions
+{
+    public class CosmosDbException : Exception
+    {
+        public string CustomExceptionMessage { get; protected set; }
```
jackliums: We should reuse the existing Message property instead of introduce a new one (refer MethodNotAllowedException for implementation).

src/Microsoft.Health.CosmosDb/Features/Storage/StoredProcedures/StoredProcedureBase.cs
```diff
@@ -66,7 +66,7 @@ private string GetBody()
             // Assumed convention is the stored proc is in the same directory as the cs file
             var resourceName = $"{GetType().Namespace}.{Name}.js";
 
-            using (Stream resourceStream = Assembly.GetExecutingAssembly().GetManifestResourceStream(resourceName))
+            using (Stream resourceStream = Assembly.GetAssembly(GetType()).GetManifestResourceStream(resourceName))
```
jackliums: nit: this could be GetType().Assembly

src/Microsoft.Health.Fhir.Api/Controllers/ControlPlaneController.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using EnsureThat;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Health.ControlPlane.Core.Features.Rbac;
+
+namespace Microsoft.Health.Fhir.Api.Controllers
+{
+    // TODO: Remove before PR
```
jackliums: Is the comment stil relevant?

src/Microsoft.Health.Fhir.Api/Modules/ControlPlaneModule.cs
```diff
@@ -0,0 +1,33 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Health.ControlPlane.Core.Features.Rbac;
+using Microsoft.Health.Extensions.DependencyInjection;
+using Microsoft.Health.Fhir.Api.Configs;
+
+namespace Microsoft.Health.Fhir.Api.Modules
+{
+    public class ControlPlaneModule : IStartupModule
+    {
+        private FhirServerConfiguration _fhirServerConfiguration;
```
jackliums: nit: readonly

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/FhirCosmosDocumentQueryFactory.cs
```diff
@@ -6,24 +6,25 @@
...
+using Microsoft.Health.CosmosDb.Features.Storage;
 using Microsoft.Health.Fhir.Core.Features.Context;
 
 namespace Microsoft.Health.Fhir.CosmosDb.Features.Storage
 {
     /// <summary>
     /// Factory for creating the <see cref="FhirDocumentQuery{T}"/>.
     /// </summary>
-    public class CosmosDocumentQueryFactory : ICosmosDocumentQueryFactory
+    public class FhirCosmosDocumentQueryFactory : ICosmosDocumentQueryFactory
```
jackliums: There seems to be inconsistency here. This one was changed to FhirCosmosDocumentQueryFacctory but should CosmosHealthCheck, CosmosSearchService, and etc also be renamed to FhirCosmosHealthCheck and so forth?

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/FhirDataStore.cs
```diff
@@ -64,9 +68,9 @@ public sealed class CosmosDataStore : IDataStore, IProvideCapability
             _cosmosDocumentQueryFactory = cosmosDocumentQueryFactory;
             _retryExceptionPolicyFactory = retryExceptionPolicyFactory;
             _logger = logger;
-            _documentClient = documentClient;
+            _documentClient = new FhirDocumentClient(documentClient.Value, fhirRequestContextAccessor, cosmosDataStoreConfiguration.ContinuationTokenSizeLimitInKb);
```
jackliums: I don't think we can do this. We need to maintain the scope.

src/Microsoft.Health.Fhir.Api/Controllers/ControlPlaneController.cs
brandonpollett: Yes, but I guess it should be `remove before PR merge`. The goal is that this API shouldn't be exposed anywhere with this code change. It's mainly here for easier manual verification of the refactor.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Versioning/FhirCollectionSettingsUpdater.cs
```diff
@@ -87,7 +88,7 @@ public async Task ExecuteAsync(IDocumentClient client, DocumentCollection collec
                 // See: https://docs.microsoft.com/en-us/azure/cosmos-db/time-to-live
                 collection.DefaultTimeToLive = -1;
 
-                await client.ReplaceDocumentCollectionAsync(collection);
+                var response = await client.ReplaceDocumentCollectionAsync(collection);
```
jackliums: is this response needed?

src/Microsoft.Health.ControlPlane.Core/Features/Rbac/RbacService.cs
brandonpollett: Yes, the idea is to expand the RBAC service to have validation and any other non datastore RBAC related code in a single place. In this PR, it's mainly used as an example of how different collections can co-exist.

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/CosmosIdentityProviderExtensions.cs
brandonpollett: We still need the extension, but it's been refactored some.

src/Microsoft.Health.ControlPlane.Core/Features/Rbac/IdentityProvider.cs
brandonpollett: Made an IReadOnlyList for now since in the current plan this won't be manipulated.

src/Microsoft.Health.ControlPlane.Core/Features/Rbac/IdentityProvider.cs
brandonpollett: Made immutable.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Versioning/FhirCollectionSettingsUpdater.cs
brandonpollett: Nope, mistakenly added.

src/Microsoft.Health.Fhir.Api/Modules/ControlPlaneModule.cs
brandonpollett: Turns out this wasn't actually being used, removed it.

src/Microsoft.Health.CosmosDb/Features/Storage/StoredProcedures/StoredProcedureBase.cs
brandonpollett: Updated.

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/Versioning/ControlPlaneCollectionUpgradeManager.cs
brandonpollett: Moved.

src/Microsoft.Health.ControlPlane.Core/Features/Rbac/RbacService.cs
brandonpollett: Done.

src/Microsoft.Health.Fhir.Web/appsettings.json
```diff
@@ -25,7 +25,11 @@
         "Host": null,
         "Key": null,
         "DatabaseId": "health",
-        "CollectionId": "fhir",
+        "InitialDatabaseThroughput": 400,
```
jackliums: Should we considerr separating the setting for FHIR and the CP? I think it has a few advantages for doing so:

1. It allows different persistent layer to be used so the right technology/price could be picked.
2. It allows different connection to be used. For example, if I am hosting 2 FHIR servers but want to share the CP, i could use 2 separate Cosmos DB account for the FHIR but share the same Cosmos DB account for managing these two servers so that the roles could be shared between the two.

These would be some redundency in terms of settings but it gives us the flexibility?

src/Microsoft.Health.ControlPlane.Core/Features/Rbac/IdentityProvider.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        public IdentityProvider(string name, string authority, IReadOnlyList<string> audience, string version)
+        {
+            Name = name;
+            Authority = authority;
+            Audience = audience;
+            Version = version;
+        }
+
+        [JsonProperty("Name")]
```
jackliums: The other documents uses JSON format with camel case so we should follow the same.

src/Microsoft.Health.ControlPlane.Core/Microsoft.Health.ControlPlane.Core.csproj
jackliums: (having said that... i am not sure if all the dependencies are ready to converted so maybe we will do it another day all at once)

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/ControlPlaneCollectionInitializer.cs
```diff
@@ -0,0 +1,26 @@
...
+using Microsoft.Extensions.Logging;
+using Microsoft.Health.ControlPlane.CosmosDb.Features.Storage.Versioning;
+using Microsoft.Health.CosmosDb.Configs;
+using Microsoft.Health.CosmosDb.Features.Storage;
+
+namespace Microsoft.Health.ControlPlane.CosmosDb.Features.Storage
+{
+    public class ControlPlaneCollectionInitializer : CollectionInitializer
+    {
+        public ControlPlaneCollectionInitializer(CosmosDataStoreConfiguration cosmosDataStoreConfiguration, ControlPlaneCollectionUpgradeManager controlPlaneCollectionUpgradeManager, ILogger<ControlPlaneCollectionInitializer> logger)
```
jackliums: Since both ControlPaneCollectionInitializer and FhirCollectionInitializer don't actually overwrite any methods and just simply passing the constructor values through, another way that you _could_ do this is do this at the IoC container. Instead of having two derived classes, you could just register CollectionInitializer like:

```
services.Add<CollectionInitializer>(sp =>
{
    var config = sp.GetService<CosmosDataStoreConfiguration>();
    var upgradeManager = sp.GetService<ControlPlaneCollectionUpgradeManager>();
    var loggerFactory = sp.GetService<ILoggerFactory>();

    return new CollectionInitializer(
        config.ControlPlaneCollectionId,
        config.RelativeDatabaseUri,
        config.RelativeControlPlaneCollectionUri,
        config.InitialControlPlaneCollectionThroughput,
        upgradeManager,
        loggerFactory.CreateLogger<CollectionInitializer>());
})
.Singleton()
.AsService<ICollectionInitializer>();
```

Since you are consuming the ICollectionInitializer as IEnumerable<ICollectionInitializer> anyways. But then again, i don't know which one is cleaner 😂

src/Microsoft.Health.ControlPlane.Core/Microsoft.Health.ControlPlane.Core.csproj
brandonpollett: I would vote for doing this later and all at once. A number of our projects target `netcoreapp2.1` which could instead target `netstandard`

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/Rbac/CosmosIdentityProvider.cs
```diff
@@ -0,0 +1,41 @@
...
+namespace Microsoft.Health.ControlPlane.CosmosDb.Features.Storage.Rbac
+{
+    public class CosmosIdentityProvider : IdentityProvider
+    {
+        internal const string IdentityProviderPartition = "_identityProviders";
+
+        public CosmosIdentityProvider(IdentityProvider identityProvider)
+            : base(identityProvider.Name, identityProvider.Authority, identityProvider.Audience, identityProvider.Version)
+        {
+            ETag = $"\"{identityProvider.Version}\"";
```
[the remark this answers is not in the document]
brandonpollett: Fair enough... my initial thought was supporting etag aware updates, but that can be handled other ways.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/FhirCosmosDocumentQueryFactory.cs
brandonpollett: Yes, I think it would make sense to work out the inconsistencies here.

brandonpollett: All of the relevant files in the Fhir.CosmosDb project should be renamed now.

src/Microsoft.Health.Fhir.Web/appsettings.json
brandonpollett: I've refactored to have the collection specific settings in the `FhirServer` section and a new `ControlPlane` section.

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/ControlPlaneCollectionInitializer.cs
brandonpollett: With the separation of the collection settings I've moved to the model suggested above.

src/Microsoft.Health.ControlPlane.Core/Features/Exceptions/ControlPlaneException.cs
```diff
@@ -0,0 +1,17 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+
+namespace Microsoft.Health.ControlPlane.Core.Features.Exceptions
+{
+    public abstract class ControlPlaneException : Exception
```
jackliums: We probably should consider deriving all of our exception from MicrosoftHealthException like this one or FhirException.

src/Microsoft.Health.ControlPlane.Core/Features/Rbac/IdentityProvider.cs
```diff
@@ -0,0 +1,38 @@
...
+namespace Microsoft.Health.ControlPlane.Core.Features.Rbac
+{
+    public class IdentityProvider
+    {
+        [JsonConstructor]
+        protected IdentityProvider()
+        {
+        }
+
+        public IdentityProvider(string name, string authority, IReadOnlyList<string> audience, string version)
```
jackliums: null checks?

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/ControlPlaneDataStore.cs
```diff
@@ -0,0 +1,142 @@
...
+            var context = new CosmosQueryContext(_collectionUri, sqlQuerySpec, feedOptions);
+
+            return _cosmosDocumentQueryFactory.Create<T>(_documentClient.Value, context);
+        }
+
+        private async Task<T> GetSystemDocumentByIdAsync<T>(string id, string partitionKey, CancellationToken cancellationToken)
+        {
+            EnsureArg.IsNotNull(id, nameof(id));
+            EnsureArg.IsNotNull(partitionKey, nameof(partitionKey));
+            var documentQuery = new SqlQuerySpec(
```
jackliums: nit: maybe some new lines for readability

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/ControlPlaneDataStore.cs
```diff
@@ -0,0 +1,142 @@
...
+                    requestOptions,
+                    true,
+                    cancellationToken);
+                _logger.LogInformation("Request charge: {RequestCharge}, latency: {RequestLatency}", response.RequestCharge, response.RequestLatency);
+                return (dynamic)response.Resource;
+            }
+            catch (DocumentClientException dce)
+            {
+                _logger.LogError(dce, "Unhandled Document Client Exception");
+                throw;
```
jackliums: UpsertIdentityProviderAsync is called by RbacService which sits in the Core and since it doesn't know anything about the underlying persistent layer, it doesn't understand DocumentClientException. Should we normalize it into some exception it understands so it can be handled later? Just some food for thought

src/Microsoft.Health.Fhir.Api/Features/Filters/OperationOutcomeExceptionFilterAttribute.cs
```diff
@@ -38,70 +40,110 @@ public override void OnActionExecuted(ActionExecutedContext context)
...
-                        fhirResult.StatusCode = HttpStatusCode.InternalServerError;
-                        break;
-                    case RequestRateExceededException ex:
-                        fhirResult.StatusCode = HttpStatusCode.TooManyRequests;
-
-                        if (ex.RetryAfter != null)
+                return;
+            }
+
+            switch (context?.Exception)
```
jackliums: I feel like nested switch statements are kind of hard to read. Maybe stick with if/else if?

```
if (context?.Exception is FhirException fhirException)
{
    ...
}
else if (context?.Exception is MicrosoftHealthException mshException)
{
    ...
}
```

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/ControlPlaneDataStore.cs
```diff
@@ -0,0 +1,142 @@
...
+            try
+            {
+                var response = await _documentClient.Value.UpsertDocumentAsync(
+                    _collectionUri,
+                    systemObject,
+                    requestOptions,
+                    true,
+                    cancellationToken);
+                _logger.LogInformation("Request charge: {RequestCharge}, latency: {RequestLatency}", response.RequestCharge, response.RequestLatency);
+                return (dynamic)response.Resource;
```
jackliums: nit: (T)(dynamic)response.Resource so that if it fails to cast for whatever reason it fails fast rather than somwhere down the line when property is accessed.

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/ControlPlaneDataStore.cs
```diff
@@ -0,0 +1,142 @@
...
+                new SqlParameterCollection(new[] { new SqlParameter("@id", id) }));
+            var feedOptions = new FeedOptions
+            {
+                PartitionKey = new PartitionKey(partitionKey),
+            };
+            IDocumentQuery<T> cosmosDocumentQuery =
+                CreateDocumentQuery<T>(documentQuery, feedOptions);
+            using (cosmosDocumentQuery)
+            {
+                var response = await cosmosDocumentQuery.ExecuteNextAsync(cancellationToken);
```
jackliums: nit: I think we agreed on using var on if the type is obvious. so documentQuery/feedOptions above are fine but this one should be explicitly typed.

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/Versioning/ControlPlaneCollectionUpgradeManager.cs
```diff
@@ -0,0 +1,79 @@
...
+            EnsureArg.IsNotNull(logger, nameof(logger));
+
+            _collectionUpdater = collectionUpdater;
+            _configuration = configuration;
+            _lockFactory = lockFactory;
+            _collectionConfiguration = namedCosmosCollectionConfigurationAccessor.Get(Constants.CollectionConfigurationName);
+            _logger = logger;
+        }
+
+        public async Task SetupCollectionAsync(IDocumentClient documentClient, DocumentCollection collection)
```
jackliums: I think you can create a base class and move this method there then have this and FhirCollectionUpgradeManager both derived from that class.

jackliums: I think you can create a base class and move this method there then have this and FhirCollectionUpgradeManager both derived from that class.

src/Microsoft.Health.ControlPlane.CosmosDb/Registration/ControlPlaneCosmosDbRegistrationExtensions.cs
```diff
@@ -0,0 +1,60 @@
...
+
+            services.Add<ControlPlaneDataStore>()
+                .Scoped()
+                .AsSelf()
+                .AsImplementedInterfaces();
+
+            services.Add<ControlPlaneCollectionUpgradeManager>()
+                .Singleton()
+                .AsSelf()
+                .AsService<IUpgradeManager>();
```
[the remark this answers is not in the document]
brandonpollett: The `DocumentClientInitializer` actually didn't even use the upgrade manager anymore, so I've removed that from the constructor/class.

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/Versioning/ControlPlaneCollectionUpgradeManager.cs
brandonpollett: refactored

brandonpollett: refactored

src/Microsoft.Health.Fhir.Api/Features/Filters/OperationOutcomeExceptionFilterAttribute.cs
brandonpollett: Works for me.

src/Microsoft.Health.ControlPlane.Core/Features/Exceptions/ControlPlaneException.cs
brandonpollett: That makes sense to me. I'll update the ControlPlaneException to derive from the MicrosoftHealthException and, in the interest of containing this PR, put the FhirException derivation as an issue.

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/ControlPlaneDataStore.cs
brandonpollett: I added a check for the 429 exception, but we should probably come up with something to wrap the persistence specific errors into. I'll open an issue for that portion.

jackliums: We can use the code generator for that to create an implementation of IDocumentClient that checks the common exception like this and also log the RU usage so that it can be used both by CP and Fhir. Let's open the issue for that and we can work on it separately.

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/Rbac/CosmosIdentityProvider.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        [JsonConstructor]
+        protected CosmosIdentityProvider()
+        {
+        }
+
+        [JsonProperty("id")]
+        public string Id => Name;
+
+        [JsonProperty("partitionKey")]
```
jackliums: This can be KnownDocumentProperties.PartitionKey. While you at it, can you also change FhirCosmosResourceWrapper and DocumentClientReadWriteTestProvider as well?

src/Microsoft.Health.ControlPlane.CosmosDb/Registration/ControlPlaneCosmosDbRegistrationExtensions.cs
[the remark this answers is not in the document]
jackliums: Maybe remove AsService<IUpgradeManager> from the registration since it's resolving using the concrete type anyways above? Not big deal though.

src/Microsoft.Health.CosmosDb/Features/Storage/DocumentClientInitializer.cs
```diff
@@ -90,91 +78,63 @@ public IDocumentClient CreateDocumentClient(CosmosDataStoreConfiguration configu
...
-        /// <param name="configuration">The data store config</param>
-        public async Task OpenDocumentClient(IDocumentClient client, CosmosDataStoreConfiguration configuration)
+        /// <inheritdoc />
+        public async Task OpenDocumentClient(IDocumentClient client, CosmosDataStoreConfiguration configuration, CosmosCollectionConfiguration cosmosCollectionConfiguration)
         {
             EnsureArg.IsNotNull(client, nameof(client));
             EnsureArg.IsNotNull(configuration, nameof(configuration));
 
-            _logger.LogInformation("Opening DocumentClient connection to {CollectionUri}", configuration.AbsoluteCollectionUri);
+            _logger.LogInformation("Opening DocumentClient connection to {CollectionUri}", configuration.GetAbsoluteCollectionUri(cosmosCollectionConfiguration.CollectionId));
```
jackliums: nit: should save this to a variable since it's repeatedly used here.

src/Microsoft.Health.CosmosDb/Features/Storage/DocumentClientReadWriteTestProvider.cs
```diff
@@ -22,12 +22,12 @@ public DocumentClientReadWriteTestProvider()
             _partitionKey = new PartitionKey(_document.PartitionKey);
         }
 
-        public async Task PerformTest(IDocumentClient documentClient, CosmosDataStoreConfiguration configuration)
+        public async Task PerformTest(IDocumentClient documentClient, CosmosDataStoreConfiguration configuration, CosmosCollectionConfiguration cosmosCollectionConfiguration)
```
jackliums: How does this one know which CosmosCollectionConfig to use?

src/Microsoft.Health.CosmosDb/Features/Storage/CosmosDocumentQueryFactory.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+        }
+
+        /// <inheritdoc />
+        public IDocumentQuery<T> Create<T>(IDocumentClient documentClient, CosmosQueryContext context)
+        {
+            EnsureArg.IsNotNull(documentClient, nameof(documentClient));
+            EnsureArg.IsNotNull(context, nameof(context));
+
+            return documentClient.CreateDocumentQuery<T>(
```
jackliums: We probably want to move the logging stuff into here so the query/RUs can be logged. Doesn't have to be in this PR. Maybe we can do it together when we refactor the code generator.

jackliums: Took me a sec to realize why this class was here

src/Microsoft.Health.CosmosDb/Features/Storage/Versioning/CollectionUpgradeManager.cs
```diff
@@ -0,0 +1,89 @@
...
+            _configuration = configuration;
+            _collectionConfiguration = cosmosCollectionConfiguration;
+            _lockFactory = lockFactory;
+            _logger = logger;
+        }
+
+        /// <summary>
+        /// This integer should be incremented in the derived instance when changing any configuration in the derived CollectionUpgradeManager
+        /// </summary>
+        public virtual int CollectionSettingsVersion => 0;
```
jackliums: maybe make this abstract so the derived class has to provide its own version?

src/Microsoft.Health.Fhir.Api/Features/Filters/OperationOutcomeExceptionFilterAttribute.cs
```diff
@@ -87,20 +94,53 @@ public override void OnActionExecuted(ActionExecutedContext context)
...
                         }
 
+                        break;
+                    default:
+                        healthExceptionResult = FhirResult.Create(
+                            new OperationOutcome
+                            {
+                                Id = _fhirRequestContextAccessor.FhirRequestContext.CorrelationId,
+                            }, HttpStatusCode.BadRequest);
+                        healthExceptionResult.StatusCode = HttpStatusCode.BadRequest;
```
jackliums: should this be internal server error since it's an exception that we couldn't "handle"

src/Microsoft.Health.ControlPlane.Core.UnitTests/Features/Rbac/RbacServiceTests.cs
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        [Fact]
+        public async void GivenAName_WhenGettingIdentityProvider_ThenDataStoreIsCalled()
+        {
+            var identityProviderName = "aad";
+
+            await _rbacService.GetIdentityProviderAsync(identityProviderName, CancellationToken.None);
+
+            await _controlPlaneDataStore.Received(1).GetIdentityProviderAsync(identityProviderName, Arg.Any<CancellationToken>());
```
jackliums: This test should use assert.Same to verify the returned instance is _identityProvider instead of used Received.

Issue #235, opened 2018-12-05
Querying string search parameter could consumes large RUs under certain circumstance


jackliums:
**Describe the bug**

We have a report from customer that simple query /Patient?given=Bob was not returning any data. Upon some investigation, the query was consuming large number of RUs which causes the query to return empty results.

Investigating further, this is because we have Hash index on string by default but we actually need Range index because Hash index works for equality checks but not for comparison checks and Cosmos DB was doing full scan because the index was not being utilized.

This issue did not come up during testing because we had variety types of resources created within our test. But for this customer, almost all of the resources were Patient and therefore the almost every query resulted in full scan.

poadhika:
Verified that with the change the total number of required RU's  dropped down from 1178.14 to  3.66

PR #289, opened 2018-12-15
Changing to Range Index in Cosmos DB

poadhika:
## Description
This change will change the index type to range index for strings.

## Related issues
Addresses [issue 235]: https://github.com/Microsoft/fhir-server/issues/235.

## Testing
We need to do manual testing to verify the performance  that will be achieved after this change in search.

### DIFF
--- a/src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Versioning/CollectionSettingsUpdater.cs
+++ b/src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Versioning/CollectionSettingsUpdater.cs
@@ -27,7 +27,7 @@ namespace Microsoft.Health.Fhir.CosmosDb.Features.Storage.Versioning
         private readonly CosmosDataStoreConfiguration _configuration;
         private static readonly RangeIndex DefaultStringRangeIndex = new RangeIndex(DataType.String, -1);
 
-        private const int CollectionSettingsVersion = 1;
+        private const int CollectionSettingsVersion = 2;
 
         public CollectionSettingsUpdater(ILogger<CollectionSettingsUpdater> logger, CosmosDataStoreConfiguration configuration)
         {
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Versioning/CollectionSettingsUpdater.cs
```diff
@@ -59,7 +59,7 @@ public async Task ExecuteAsync(IDocumentClient client, DocumentCollection collec
                         Indexes = new Collection<Index>
                         {
                             new RangeIndex(DataType.Number, -1),
-                            new HashIndex(DataType.String, 3),
+                            new RangeIndex(DataType.String, -1),
```
jackliums: Now the range index on string on the root level, some of the policies below is not longer needed and can be removed.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Versioning/CollectionUpgradeManager.cs
```diff
@@ -23,7 +23,7 @@ public class CollectionUpgradeManager : IUpgradeManager
         /// This integer should be incremented when changing any value in the
         /// UpdateIndexAsync function
         /// </summary>
-        internal const int CollectionSettingsVersion = 1;
+        internal const int CollectionSettingsVersion = 2;
```
jackliums: Not your code but can you update the comment above? There is no UpdateIndexAsync function.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Versioning/CollectionSettingsUpdater.cs
```diff
@@ -59,19 +57,9 @@ public async Task ExecuteAsync(IDocumentClient client, DocumentCollection collec
                         Indexes = new Collection<Index>
```
jackliums: the indentation seems to be wrong. Can you fix it since you are at it?

Issue #296, opened 2018-12-19
Fix AspNetCore.App package version build error


hansenms:
**Describe the bug**
AspNetCore.App package reference is versioned (2.1.6), which causes build errors.

**To Reproduce**
Attempt build on Linux agent.

**Expected behavior**
Build success

**Actual behavior**
```
/usr/share/dotnet/sdk/2.2.101/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.DefaultItems.targets(153,5): error NETSDK1071: A PackageReference to 'Microsoft.AspNetCore.App' specified a Version of `2.1.6`. Specifying the version of this package is not recommended. For more information, see https://aka.ms/sdkimplicitrefs [/home/vsts/work/1/s/src/Microsoft.Health.Fhir.Web/Microsoft.Health.Fhir.Web.csproj]
```

PR #297, opened 2018-12-19
Removing AspNetCore.App package version

hansenms:
Adresses #296


Directory.Build.props
```diff
@@ -4,6 +4,7 @@
         <WarningsAsErrors />
         <DebugType>Full</DebugType>
         <LangVersion>7.3</LangVersion>
+        <RuntimeFrameworkVersion>2.1.6</RuntimeFrameworkVersion>
```
jackliums: I thought we were moving away from specifying the framework version?

Issue #309, opened 2019-01-03
Audit log should be created for all operations that requires authentication or authorization.


jackliums:
**User story**
As an administrator, I want authentication/authorization failures to be logged so that I can audit them.

**Acceptance criteria**
1. If a user executes an operation that requires authentication or authorization, an audit log entry should be created with the appropriate result.

PR #310, opened 2019-01-03
Audit the authentication/authorization failures

jackliums:
## Description
- Added support for creating audit log for authentication/authorization failures.
- Refactored some of the classes and moved audit related classes to Audit namespace.

Majority of the time, the audit log is still going to be created by AuditLoggingFilterAttribute. However, because MVC executes authorization filter first before any other filter, if the call is being rejected because the call does not have proper token or the user does not meet the requirement to access the API, then currently it is not logged. Adding the AuditMiddleware to catch such occasion and log them as necessary.

## Related issues
Addresses #309.

## Testing
- Added some unit tests.
- Tried to add E2E tests to cover this scenario but this involves much more work since we need to be able to access the log remotely from the test. For now, I did some manual testing to make sure the audit log is created for normal successful case, unauthenticated case (both valid and invalid token) as well as unauthorized case.


src/Microsoft.Health.Fhir.Api/Features/Audit/AuditMiddleware.cs
```diff
@@ -0,0 +1,70 @@
...
+        {
+            try
+            {
+                await _next(context);
+            }
+            finally
+            {
+                var statusCode = (HttpStatusCode)context.Response.StatusCode;
+
+                // Since authorization filters runs first before any other filters, if the authorization fails,
```
jackliums: Another option would be to remove the OnResultExecuted override from AuditLoggingFilterAttribute and log all of the "executed" audit log here. However, there is a slight problem with that approach. This code relies on MVC framework to have parsed the RouteData already. Because we use the IdentityServer in our testing, when calls are made to the IdentityServer, this middleware is executed but the RouteData is not parsed since the MVC framework is not triggered. We could skip the auditing if "controller" and "action" are not present but we wouldn't know if the middleware is misbehaving. Currently, if the code reaches here and the "controller" or "action" is missing, it will throw AuditException so that we can monitor to make sure the code is behaving the way it should be.

Any thoughts?

src/Microsoft.Health.Fhir.Api/Controllers/FhirController.cs
```diff
@@ -145,7 +145,7 @@ public IActionResult CustomError(int? statusCode = null)
         /// <param name="resource">The resource.</param>
         [HttpPost]
         [Route(KnownRoutes.ResourceType)]
-        [AuditEventSubType(AuditEventSubType.Create)]
+        [AuditEventType(AuditEventSubType.Create)]
```
brandonpollett: Should the enum be changed to `AuditEventType` also?

src/Microsoft.Health.Fhir.Api.UnitTests/Features/Audit/AuditHelperTests.cs
```diff
@@ -0,0 +1,178 @@
...
+        [Fact]
+        public void GivenUnknownActon_WhenLogExecutingIsCalled_ThenAuditExceptionShouldBeThrown()
+        {
+            Assert.Throws<AuditException>(() => _auditHelper.LogExecuting("test", "action"));
+        }
+
+        [Fact]
+        public void GivenAnActionHasAllowAnonymousAttribute_WhenLogExecutedIsCalled_ThenAuditLogEntryShouldNotBeCreated()
+        {
+            _auditHelper.LogExecuting(ControllerName, AnonymousMethodName);
```
brandonpollett: Should this be `LogExecuted`?

src/Microsoft.Health.Fhir.Api.UnitTests/Features/Audit/AuditHelperTests.cs
```diff
@@ -0,0 +1,178 @@
...
+
+            _auditLogger.DidNotReceiveWithAnyArgs().LogAudit(AuditAction.Executed, null, null, null, null, null, null);
+        }
+
+        [Fact]
+        public void GivenAnActionHasAuditEventTypeAttribute_WhenLogExecutedIsCalled_ThenAuditLogEntryShouldNotBeCreated()
```
brandonpollett: `ShouldBeCreated` not `ShouldNotBeCreated`

src/Microsoft.Health.Fhir.Api/Features/Security/Authorization/ResourceActionHandler.cs
```diff
@@ -38,6 +55,17 @@ protected override Task HandleRequirementAsync(AuthorizationHandlerContext conte
             else
             {
                 context.Fail();
+
+                IFhirRequestContext fhirRequestContext = _fhirRequestContextAccessor.FhirRequestContext;
+
+                _auditLogger.LogAudit(
```
brandonpollett: Is this also logged by the middleware?

src/Microsoft.Health.Fhir.Api/Controllers/FhirController.cs
jackliums: The ValueSet itself is defined by http://hl7.org/fhir/STU3/valueset-audit-event-sub-type.html, which is called AuditEventSubType so I left it as is.

src/Microsoft.Health.Fhir.Api/Features/Security/Authorization/ResourceActionHandler.cs
jackliums: I didn't see this code. Remove the logging from here and only logging in the middleware.

src/Microsoft.Health.Fhir.Api.UnitTests/Features/Security/ResourceActionHandlerTests.cs
```diff
@@ -32,8 +31,8 @@ public ResourceActionHandlerTests()
         [InlineData(true)]
         public async void GivenAReadResourceRequest_WhenAuthorizationHandlerHandles_ThenTheAuthorizationHandlerAppropriateStatus(bool authorized)
         {
-            _authorizationPolicy.HasPermission(Arg.Any<ClaimsPrincipal>(), ResourceAction.Read).ReturnsForAnyArgs(authorized);
```
brandonpollett: I think this was inadvertently deleted.

jackliums: Oops I thought I ran the tests before submitting it but apparently I didn't...

src/Microsoft.Health.Fhir.Api/Features/Audit/AuditMiddleware.cs
```diff
@@ -0,0 +1,70 @@
...
+                // the AuditLoggingFilterAttribute, which is where the audit logging would normally happen, will not be executed.
+                // This middleware will log any Unauthorized or Forbidden request if it hasn't been logged yet.
+                if (_fhirRequestContextAccessor.FhirRequestContext.RouteName == null &&
```
garamamo: Just curious, why are we checking for this?

src/Microsoft.Health.Fhir.Api/Features/Audit/AuditMiddleware.cs
```diff
@@ -0,0 +1,70 @@
...
+                {
+                    RouteData routeData = context.GetRouteData();
+
+                    routeData.Values.TryGetValue("controller", out object controllerName);
+                    routeData.Values.TryGetValue("action", out object actionName);
+                    routeData.Values.TryGetValue(KnownActionParameterNames.ResourceType, out object resourceType);
+
+                    _auditHelper.LogExecuted(
+                        controllerName.ToString(),
+                        actionName.ToString(),
```
garamamo: Should we be checking for null?

src/Microsoft.Health.Fhir.Api/Features/Audit/AuditMiddleware.cs
```diff
@@ -0,0 +1,70 @@
...
+using EnsureThat;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing;
+using Microsoft.Health.Fhir.Api.Features.Routing;
+using Microsoft.Health.Fhir.Core.Features.Context;
+
+namespace Microsoft.Health.Fhir.Api.Features.Audit
+{
+    /// <summary>
+    /// A middleware that logs audit entry that cannot be logged by <see cref="AuditLoggingFilterAttribute"/>.
```
brendankowitz:
```suggestion
    /// A middleware that logs audit events that cannot be logged by <see cref="AuditLoggingFilterAttribute"/>.
```

src/Microsoft.Health.Fhir.Api/Resources.resx
```diff
@@ -139,6 +135,10 @@
   <data name="MenuCapabilityStatement" xml:space="preserve">
     <value>Capability Statement</value>
   </data>
+  <data name="MissingAuditInformation" xml:space="preserve">
+    <value>The audit information is missing for Controller: {0} and Action: {1}. This usually means code error where the action is not marked with appropriate attribute.</value>
```
brendankowitz:
```suggestion
    <value>The audit information is missing for Controller: {0} and Action: {1}. This usually means the action is not marked with appropriate attribute.</value>
```

src/Microsoft.Health.Fhir.Api/Features/Audit/AuditMiddleware.cs
jackliums: If the MVC framework is executed, then the AuditLoggingFilterAttribute is triggered and the RouteName would be set. Since the filter already logged the AuditEvent.Executed event, we don't want to log it again here. The comment above describes it but I can add a bit more detail to explain. Also, see the comment above for why they are being logged in 2 different places.

src/Microsoft.Health.Fhir.Api/Features/Audit/AuditMiddleware.cs
jackliums: The code was meant to be controllerName?.ToString() and so forth. I will update it.

Issue #295, opened 2018-12-19
Bootstrap RBAC Info into control plane


brandonpollett:
**User story**
As an admin, I want to be able to bootstrap my RBAC configuration so that on initial load I can manage my service.

**Acceptance criteria**
- [ ] Configuration has a section for entering initial bootstrap of RBAC configuration
- [ ] Identity provider, role, and role membership can be bootstrapped
- [ ] Bootstrapping only occurs once
- [ ] Changes to the bootstrap configuration is not persisted to the datastore

cunninghamjc:
Moving work item tracking to Azure DevOps.

PR #314, opened 2019-01-04
Adds bootstrapping for RBAC in the Control Plane

brandonpollett:
## Description
This PR contains the RBAC bootstrapping code for the initial load of the RBAC data into the datastore. It adds a configuration section under control plane that will hold the information to bootstrap. The bootstrapping will only occur if there is no `bootstrap` document in the control plane datastore or if the bootstrap document has a different hash than what is attempted to be bootstrapped. 

## Related issues
Addresses #295 .

## Testing
This change was tested with a combination of manual testing and unit tests. Additionally this will eventually be used to setup PR environments and CI environments on which automated tests will run.

### DIFF
--- a/src/Microsoft.Health.Abstractions/Cryptography/Hasher.cs
+++ b/src/Microsoft.Health.Abstractions/Cryptography/Hasher.cs
@@ -7,9 +7,9 @@ using System;
 using System.Security.Cryptography;
 using System.Text;
 
-namespace Microsoft.Health.CosmosDb.Features.Storage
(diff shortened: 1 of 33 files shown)


brandonpollett:
Once the rbac support for roles, identity providers, and role memberships are in this PR will need to be updated to bootstrap those values. Additionally, the will need to be supported via the `testAuthEnvironment.json` file so that bootstrapping can be done on both local and PR environments.

brandonpollett:
Abandoning for now

PR #315, opened 2019-01-04
Store Identity Provider information in datastore #276

garamamo:
Addresses issue https://github.com/Microsoft/fhir-server/issues/276.

## Testing
Added unit tests to test this change.

Note: This change has the controller code for identity provider. It is functional but may need additional changes. Currently not sure how ARM/Portal would consume this. I can remove the controller code before merge based on feedback.


src/Microsoft.Health.ControlPlane.CosmosDb.UnitTests/Features/Storage/ControlPlaneDataStoreTests.cs
```diff
@@ -68,7 +67,7 @@ public ControlPlaneDataStoreTests()
                 logger);
         }
 
-        [Fact]
+        [Fact(Skip = "SetupIssue for dynamic")]
```
garamamo: Skipping as the mock setup for underlying azure data structures is complicated for the changes. I have unit tests in the pr that test at a level above. Will also have e2e tests that exercise everything once api changes are in.

src/Microsoft.Health.Fhir.Api/Controllers/ControlPlaneController.cs
```diff
@@ -0,0 +1,84 @@
...
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Health.ControlPlane.Core.Features.Persistence;
+using Microsoft.Health.ControlPlane.Core.Features.Rbac;
+using Microsoft.Net.Http.Headers;
+
+namespace Microsoft.Health.Fhir.Api.Controllers
+{
+    // TODO: Remove before PR
+    [Route("Admin/IdentityProvider/")]
+    public class ControlPlaneController : Controller
```
garamamo: Used this for testing purposes. This is fully functional and can be used to build on for the api story. Based on the feedback, I can either keep this or remove before merge.

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/ControlPlaneDataStore.cs
```diff
@@ -106,41 +137,75 @@ private async Task<T> GetSystemDocumentByIdAsync<T>(string id, string partitionK
...
+
+        private async Task DeleteSystemDocumentsByIdAsync<T>(string id, string partitionKey, string eTag, CancellationToken cancellationToken)
+        {
+            EnsureArg.IsNotNull(id, nameof(id));
+            FeedResponse<Document> response = await GetSystemDocuments<T>(id, partitionKey, cancellationToken);
+            var document = response.SingleOrDefault();
+
+            if (document == null)
+            {
+                throw new IdentityProviderNotFoundException(id);
```
brandonpollett: This should be a more general exception if being thrown in the `DeleteSystemDocumentsByIdAsync` method

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/ControlPlaneDataStore.cs
```diff
@@ -66,14 +67,24 @@ public async Task<IdentityProvider> GetIdentityProviderAsync(string name, Cancel
             return identityProvider.ToIdentityProvider();
         }
 
-        public async Task<IdentityProvider> UpsertIdentityProviderAsync(IdentityProvider identityProvider, CancellationToken cancellationToken)
+        public async Task<IEnumerable<IdentityProvider>> GetAllIdentityProvidersAsync(CancellationToken cancellationToken)
+        {
+            return await GetSystemDocumentsByIdAsync<CosmosIdentityProvider>(null, CosmosIdentityProvider.IdentityProviderPartition, cancellationToken);
```
brandonpollett: Do the response from this look like the cosmos versions of the documents or the ones that get sent in? I've had trouble in the past with the implicit conversion which is why the `ToIdentityProvider()` extension exists.

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/ControlPlaneDataStore.cs
```diff
@@ -106,41 +137,75 @@ private async Task<T> GetSystemDocumentByIdAsync<T>(string id, string partitionK
...
+                    true,
+                    cancellationToken);
+                _logger.LogInformation("Request charge: {RequestCharge}, latency: {RequestLatency}", response.RequestCharge, response.RequestLatency);
+
+                var outcomeType = response.StatusCode == HttpStatusCode.Created ? UpsertOutcomeType.Created : UpsertOutcomeType.Updated;
+                return new UpsertResponse<T>((T)(dynamic)response.Resource, outcomeType, response.Resource.ETag);
+            }
+            catch (DocumentClientException dce)
+            {
+                _logger.LogError(dce, "Unhandled Document Client Exception");
```
brandonpollett: We should go ahead and explicitly catch the 429 error here like we do below.

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/ControlPlaneDataStore.cs
```diff
@@ -90,10 +101,30 @@ public async Task<IdentityProvider> UpsertIdentityProviderAsync(IdentityProvider
...
+            return result;
+        }
+
+        private async Task<IEnumerable<T>> GetSystemDocumentsByIdAsync<T>(string id, string partitionKey, CancellationToken cancellationToken)
+        {
+            FeedResponse<Document> response = await GetSystemDocuments<T>(id, partitionKey, cancellationToken);
+            return response.Select(r => r.GetPropertyValue<T>("r"));
+        }
+
+        private async Task<FeedResponse<Document>> GetSystemDocuments<T>(string id, string partitionKey, CancellationToken cancellationToken)
```
brandonpollett: Let's go one step more in making this generic and allow it to accept a set of key value pairs that can ben appended to the query. That way we can use this also for a basic search/filter that will be needed for the role memberships.

src/Microsoft.Health.ControlPlane.Core/Features/Rbac/RbacService.cs
```diff
@@ -21,18 +22,28 @@ public RbacService(IControlPlaneDataStore controlPlaneDataStore)
...
             return await _controlPlaneDataStore.GetIdentityProviderAsync(name, cancellationToken);
         }
 
-        public async Task<IdentityProvider> UpsertIdentityProviderAsync(IdentityProvider identityProvider, CancellationToken cancellationToken)
+        public async Task<UpsertResponse<IdentityProvider>> UpsertIdentityProviderAsync(IdentityProvider identityProvider, string eTag, CancellationToken cancellationToken)
         {
             EnsureArg.IsNotNull(identityProvider, nameof(identityProvider));
 
-            return await _controlPlaneDataStore.UpsertIdentityProviderAsync(identityProvider, cancellationToken);
+            return await _controlPlaneDataStore.UpsertIdentityProviderAsync(identityProvider, eTag, cancellationToken);
```
brandonpollett: We should validate the identity provider before creating/updating. The validation should be similar to the role validation that currently exists. 

Besides the standard validation we should also verify that we can get to `{authority}/.well-known/openid-configuration` before saving.

src/Microsoft.Health.ControlPlane.Core/Features/Rbac/RbacService.cs
```diff
@@ -21,18 +22,28 @@ public RbacService(IControlPlaneDataStore controlPlaneDataStore)
             _controlPlaneDataStore = controlPlaneDataStore;
         }
 
+        public async Task DeleteIdentityProviderAsync(string name, string eTag, CancellationToken cancellationToken)
+        {
+            await _controlPlaneDataStore.DeleteIdentityProviderAsync(name, eTag, cancellationToken);
```
brandonpollett: We should verify that you're not deleting the last identity provider when deleting. This may be easiest handled by a stored proc because we'll also need to clean up role memberships once they are implemented. Examples of stored procs are in the Fhir control plane.

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/ControlPlaneDataStore.cs
```diff
@@ -106,41 +140,92 @@ private async Task<T> GetSystemDocumentByIdAsync<T>(string id, string partitionK
...
+
             var requestOptions = new RequestOptions
             {
                 PartitionKey = new PartitionKey(partitionKey),
                 AccessCondition = eTagAccessCondition,
             };
+            return requestOptions;
+        }
+
+        private async Task DeleteSystemDocumentsByIdAsync<T>(string id, string partitionKey, string eTag, CancellationToken cancellationToken)
```
brandonpollett: Are we using the generic ever?

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/ControlPlaneDataStore.cs
```diff
@@ -106,41 +140,92 @@ private async Task<T> GetSystemDocumentByIdAsync<T>(string id, string partitionK
...
+                    async () => await _hardDeleteIdentityProvider.Execute(
+                        _documentClient.Value,
+                        _collectionUri,
+                        id,
+                        eTag),
                     cancellationToken);
-                _logger.LogInformation("Request charge: {RequestCharge}, latency: {RequestLatency}", response.RequestCharge, response.RequestLatency);
 
-                return (T)(dynamic)response.Resource;
+                _logger.LogDebug($"Hard-deleted {response.Response.Count} documents, which consumed {response.RequestCharge} RUs. The list of hard-deleted documents: {string.Join(", ", response.Response)}.");
```
brandonpollett: I don't think this function can currently delete more than one document since it is happening by id.

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/Queries/QueryBuilder.cs
```diff
@@ -105,42 +108,22 @@ public SqlQuerySpec GenerateHistorySql(SearchOptions searchOptions)
 
             private void AppendSelectFromRoot(string selectList = SearchValueConstants.RootAliasName)
             {
-                _queryBuilder
-                    .Append("SELECT ")
-                    .Append(selectList)
-                    .Append(" FROM root ")
-                    .AppendLine(SearchValueConstants.RootAliasName);
+                _queryHelper.AppendSelectFromRoot(selectList);
```
brandonpollett: Does it make sense to still have these private methods since they simply call the query helper? Maybe we should just call the query helper directly.

src/Microsoft.Health.Fhir.Api/Controllers/ControlPlaneController.cs
garamamo: Removed this file.

src/Microsoft.Health.ControlPlane.Core.UnitTests/Features/Rbac/RbacServiceTests.cs
```diff
@@ -20,10 +23,10 @@ public class RbacServiceTests
 
         public RbacServiceTests()
         {
-            _identityProvider = new IdentityProvider("aad", "https://login.microsoftonline.com/common", new List<string> { "test" }, "1");
+            _identityProvider = new IdentityProvider("aad", "https://login.microsoftonline.com/microsoft.onmicrosoft.com/", new List<string> { "test" }, "1");
```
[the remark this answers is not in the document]
garamamo: No particular reason. I added this while testing the authority validation code. It should not matter for unit tests.

src/Microsoft.Health.ControlPlane.Core/Features/Exceptions/InvalidDefinitionException.cs
```diff
@@ -0,0 +1,17 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+
+namespace Microsoft.Health.ControlPlane.Core.Features.Exceptions
+{
+    public class InvalidDefinitionException : ControlPlaneException
+    {
+        public InvalidDefinitionException(string name, List<string> issues)
```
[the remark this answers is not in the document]
garamamo: Yes. I renamed the parameter from name to message.

src/Microsoft.Health.ControlPlane.Core/Features/Rbac/RbacService.cs
```diff
@@ -21,18 +24,42 @@ public RbacService(IControlPlaneDataStore controlPlaneDataStore)
...
```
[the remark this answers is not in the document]
garamamo: There is a null/whitespace check in the controlplane datastore method. I added it here as well.

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/ControlPlaneDataStore.cs
```diff
@@ -66,14 +76,24 @@ public async Task<IdentityProvider> GetIdentityProviderAsync(string name, Cancel
...
+        }
+
+        public async Task<UpsertResponse<IdentityProvider>> UpsertIdentityProviderAsync(IdentityProvider identityProvider, string eTag, CancellationToken cancellationToken)
         {
             EnsureArg.IsNotNull(identityProvider, nameof(identityProvider));
 
             var cosmosIdentityProvider = new CosmosIdentityProvider(identityProvider);
-            var resultIdentityProvider = await UpsertSystemObjectAsync(cosmosIdentityProvider, CosmosIdentityProvider.IdentityProviderPartition, cancellationToken);
+            var resultIdentityProvider = await UpsertSystemObjectAsync(cosmosIdentityProvider, CosmosIdentityProvider.IdentityProviderPartition, eTag, cancellationToken);
+            return new UpsertResponse<IdentityProvider>(resultIdentityProvider.ControlPlaneResource.ToIdentityProvider(), resultIdentityProvider.OutcomeType, resultIdentityProvider.ETag);
```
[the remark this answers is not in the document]
garamamo: That's correct.

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/ControlPlaneDataStore.cs
```diff
@@ -87,14 +107,28 @@ public async Task<IdentityProvider> UpsertIdentityProviderAsync(IdentityProvider
             return _cosmosDocumentQueryFactory.Create<T>(_documentClient.Value, context);
         }
 
-        private async Task<T> GetSystemDocumentByIdAsync<T>(string id, string partitionKey, CancellationToken cancellationToken)
+        internal async Task<IEnumerable<T>> GetSystemDocumentsAsync<T>(List<KeyValuePair<string, object>> filterNameValues, string partitionKey, CancellationToken cancellationToken)
```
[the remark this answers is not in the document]
garamamo: This is the generic method you requested. I made it internal so that we can use it from other places.

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/ControlPlaneDataStore.cs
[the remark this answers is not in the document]
garamamo: This is a generic method.

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/ControlPlaneDataStore.cs
```diff
@@ -106,41 +140,92 @@ private async Task<T> GetSystemDocumentByIdAsync<T>(string id, string partitionK
...
             try
             {
-                ResourceResponse<Document> response = await _documentClient.Value.UpsertDocumentAsync(
-                    _collectionUri,
-                    systemObject,
-                    requestOptions,
-                    true,
+                _logger.LogDebug($"Obliterating {id}");
+
+                StoredProcedureResponse<IList<string>> response = await _retryExceptionPolicyFactory.CreateRetryPolicy().ExecuteAsync(
```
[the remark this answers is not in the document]
garamamo: It will throw a bad request with message Cannot delete as only one identity provider exists.

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/ControlPlaneDataStore.cs
garamamo: This is a generic method, I will leave it as is.

src/Microsoft.Health.CosmosDb/Features/Queries/QueryHelper.cs
```diff
@@ -0,0 +1,68 @@
...
+            _queryBuilder
+                    .Append(_rootAliasName).Append(".").Append(name)
+                    .Append(" = ")
+                    .AppendLine(_queryParameterManager.AddOrGetParameterMapping(value));
+        }
+
+        public void AppendSystemDataFilter(bool systemDataValue = false)
+        {
+            _queryBuilder
+                .Append(" WHERE ")
```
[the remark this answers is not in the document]
garamamo: I will leave it as is. We have it the same way in fhir and this is used by search as well.

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/Queries/QueryBuilder.cs
garamamo: The private methods have defaults. We will have to add values everywhere since we will different values for controlplane vs fhir.

src/Microsoft.Health.ControlPlane.Core/Features/Rbac/IdentityProvider.cs
```diff
@@ -40,5 +43,68 @@ public IdentityProvider(string name, string authority, IReadOnlyList<string> aud
...
+
+            foreach (ValidationResult result in ValidateAuthority())
+            {
+                yield return result;
+            }
+        }
+
+        public virtual IEnumerable<ValidationResult> ValidateAuthority()
+        {
+            using (var handler = new HttpClientHandler())
```
brandonpollett: This may not work with the E2E tests with identity server. It won't be available through the standard http handler and has a special one use.

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/Rbac/CosmosIdentityProviderExtensions.cs
```diff
@@ -10,15 +10,15 @@ namespace Microsoft.Health.ControlPlane.CosmosDb.Features.Storage.Rbac
...
+        internal static IdentityProvider ToIdentityProvider(this CosmosIdentityProvider cosmosIdentityProvider)
         {
             EnsureArg.IsNotNull(cosmosIdentityProvider);
 
             return new IdentityProvider(
                 cosmosIdentityProvider.Name,
                 cosmosIdentityProvider.Authority,
                 cosmosIdentityProvider.Audience,
-                cosmosIdentityProvider.ETag.Trim('"'));
+                cosmosIdentityProvider.Version);
```
brandonpollett: Does version ever get set on the cosmosidentityprovider? And if it does, does it auto-increment when you send in a new version of the identity provider?

src/Microsoft.Health.ControlPlane.Core/Features/Rbac/IdentityProvider.cs
```diff
@@ -41,8 +45,8 @@ public IdentityProvider(string name, string authority, IReadOnlyList<string> aud
         [JsonProperty("audience")]
         public virtual IReadOnlyList<string> Audience { get; protected set; }
 
-        [JsonProperty("version")]
-        public virtual string Version { get; protected set; }
+        [JsonProperty("etag")]
+        public virtual string VersionTag { get; protected set; }
```
brandonpollett: Why is there a name mismatch?

src/Microsoft.Health.ControlPlane.Core/AssemblyInfo.cs
```diff
@@ -4,5 +4,7 @@
 // -------------------------------------------------------------------------------------------------
 
 using System.Resources;
+using System.Runtime.CompilerServices;
 
 [assembly: NeutralResourcesLanguage("en-us")]
+[assembly: InternalsVisibleTo("Microsoft.Health.ControlPlane.CosmosDb")]
```
jackliums: Why do we need this?

src/Microsoft.Health.ControlPlane.Core/Features/Exceptions/ControlPlaneException.cs
```diff
@@ -3,13 +3,30 @@
...
 {
     public abstract class ControlPlaneException : Abstractions.Exceptions.MicrosoftHealthException
     {
-        protected ControlPlaneException(string message)
+        protected ControlPlaneException(string message, IEnumerable<string> issues = null)
             : base(message)
         {
+            if (issues != null && issues.Any())
+            {
+                Issues = issues.ToList();
```
jackliums: Issues is already being allocated below. Do Issues.AddRange(issues) instead.

src/Microsoft.Health.ControlPlane.Core/Features/Persistence/UpsertOutcomeType.cs
```diff
@@ -0,0 +1,13 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.ControlPlane.Core.Features.Persistence
+{
+    public enum UpsertOutcomeType
```
jackliums: I think UpsertOutcome is sufficient. Don't need the type.

src/Microsoft.Health.ControlPlane.Core/Features/Persistence/IControlPlaneDataStore.cs
```diff
@@ -13,6 +14,10 @@ public interface IControlPlaneDataStore
     {
         Task<IdentityProvider> GetIdentityProviderAsync(string name, CancellationToken cancellationToken);
 
-        Task<IdentityProvider> UpsertIdentityProviderAsync(IdentityProvider identityProvider, CancellationToken cancellationToken);
+        Task<IEnumerable<IdentityProvider>> GetAllIdentityProvidersAsync(CancellationToken cancellationToken);
+
+        Task<UpsertResponse<IdentityProvider>> UpsertIdentityProviderAsync(IdentityProvider identityProvider, string eTag, CancellationToken cancellationToken);
```
jackliums: I think instead of eTag, we should use something more generic like version.

src/Microsoft.Health.ControlPlane.Core/Features/Persistence/UpsertResponse.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        public UpsertResponse(T controlPlaneResource, UpsertOutcomeType outcomeType, string eTag)
+        {
+            EnsureArg.IsNotNull(controlPlaneResource, nameof(controlPlaneResource));
+            ControlPlaneResource = controlPlaneResource;
+            OutcomeType = outcomeType;
+            ETag = eTag;
+        }
+
+        public T ControlPlaneResource { get; }
```
jackliums: nit: just name the property Resource?

src/Microsoft.Health.ControlPlane.Core/Resources.resx
```diff
@@ -117,7 +117,31 @@
...
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="IdentityProviderAudienceIsNull" xml:space="preserve">
+    <value>Identity provider audience is null.</value>
+  </data>
+  <data name="IdentityProviderAuthorityEmpty" xml:space="preserve">
+    <value>Identity provider authority is null or empty.</value>
+  </data>
+  <data name="IdentityProviderDefinitionIsInvalid" xml:space="preserve">
+    <value>Identity provider definition Is Invalid.</value>
```
jackliums: wrong casing.

src/Microsoft.Health.ControlPlane.Core/Resources.resx
```diff
@@ -117,7 +117,31 @@
...
+  </data>
+  <data name="IdentityProviderInvalidAudience" xml:space="preserve">
+    <value>One or more of audience values is null or empty.</value>
+  </data>
+  <data name="IdentityProviderInvalidMetadataUrl" xml:space="preserve">
+    <value>"Invalid authority '{0}'. Cannot load metadata url '{1}'".</value>
```
jackliums: Why are there double quotes around the text?

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/ControlPlaneDataStore.cs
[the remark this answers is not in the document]
jackliums: If these are generic methods and is meant to be used by other classes, please move it out to another class or extension method.

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/ControlPlaneDataStore.cs
```diff
@@ -66,14 +76,25 @@ public async Task<IdentityProvider> GetIdentityProviderAsync(string name, Cancel
             return identityProvider.ToIdentityProvider();
         }
 
-        public async Task<IdentityProvider> UpsertIdentityProviderAsync(IdentityProvider identityProvider, CancellationToken cancellationToken)
+        public async Task<IEnumerable<IdentityProvider>> GetAllIdentityProvidersAsync(CancellationToken cancellationToken)
+        {
+            var cosmosIdentityProviders = await GetSystemDocumentsAsync<CosmosIdentityProvider>(new List<KeyValuePair<string, object>>(), CosmosIdentityProvider.IdentityProviderPartition, cancellationToken);
```
jackliums: Let's avoid unnecessary allocation ```new List<KeyValuePair<string, object>>()```.

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/ControlPlaneDataStore.cs
```diff
@@ -106,41 +141,102 @@ private async Task<T> GetSystemDocumentByIdAsync<T>(string id, string partitionK
...
-                FeedResponse<dynamic> response = await cosmosDocumentQuery.ExecuteNextAsync(cancellationToken);
-                var result = response.SingleOrDefault();
+                FeedResponse<Document> response = await cosmosDocumentQuery.ExecuteNextAsync<Document>(cancellationToken);
+                return response;
+            }
+        }
+
+        private async Task<T> GetSystemDocumentByIdAsync<T>(string id, string partitionKey, CancellationToken cancellationToken)
+        {
+            EnsureArg.IsNotNull(id, KnownDocumentProperties.Id);
```
jackliums: should this be nameof(id)?

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/ControlPlaneDataStore.cs
```diff
@@ -106,41 +141,102 @@ private async Task<T> GetSystemDocumentByIdAsync<T>(string id, string partitionK
 
             using (cosmosDocumentQuery)
```
jackliums: who is responsible for paging if there are more results?

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/ControlPlaneDataStore.cs
```diff
@@ -106,41 +141,102 @@ private async Task<T> GetSystemDocumentByIdAsync<T>(string id, string partitionK
...
-                var result = response.SingleOrDefault();
+                FeedResponse<Document> response = await cosmosDocumentQuery.ExecuteNextAsync<Document>(cancellationToken);
+                return response;
+            }
+        }
+
+        private async Task<T> GetSystemDocumentByIdAsync<T>(string id, string partitionKey, CancellationToken cancellationToken)
+        {
+            EnsureArg.IsNotNull(id, KnownDocumentProperties.Id);
+            var kvps = new List<KeyValuePair<string, object>> { new KeyValuePair<string, object>(KnownDocumentProperties.Id, id) };
```
jackliums: let's not do this. If it's retrieving document by id, it's much cheaper (in terms of RU) to write specific query for retriving docuemnt by id rather than a query.

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/ControlPlaneDataStore.cs
jackliums: There can only be one document if the id and partition key is specified. Let's not call it "documents".

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/Rbac/CosmosIdentityProvider.cs
```diff
@@ -9,12 +9,12 @@
...
 {
-    public class CosmosIdentityProvider : IdentityProvider
+    internal class CosmosIdentityProvider : IdentityProvider
     {
-        internal const string IdentityProviderPartition = "_identityProviders";
+        public const string IdentityProviderPartition = "_identityProviders";
 
         public CosmosIdentityProvider(IdentityProvider identityProvider)
-            : base(identityProvider.Name, identityProvider.Authority, identityProvider.Audience, identityProvider.Version)
+            : base(identityProvider.Name, identityProvider.Authority, identityProvider.Audience)
```
jackliums: nit: check null before using it?

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/StoredProcedures/HardDelete/hardDeleteIdentityProvider.js
```diff
@@ -0,0 +1,139 @@
...
+    let deletedResourceIdList = new Array();
+
+    getIdentityProviderCountAndDelete(eTag);
+
+    function getIdentityProviderCountAndDelete(eTag) {
+
+        var doc;
+        // Find the resource and all of its history.
+        let query = {
+            query: "SELECT COUNT(r.id) docCount FROM ROOT r"
```
jackliums: I don't think this is a valid syntax.

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/StoredProcedures/HardDelete/hardDeleteIdentityProvider.js
```diff
@@ -0,0 +1,139 @@
...
+            throw new Error(errorMessage.requestEntityTooLarge);
+        }
+
+    }
+
+    function tryQueryAndHardDelete(eTag) {
+        // Find the resource and all of its history.
+        let query = {
+            query: "SELECT * FROM ROOT r WHERE r.id = @id",
+            parameters: [{ name: "@id", value: id }]
```
jackliums: eTag can be directly accessed so it doesn't need to be passed in?

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/StoredProcedures/HardDelete/hardDeleteIdentityProvider.js
```diff
@@ -0,0 +1,139 @@
...
+
+        if (!isQueryAccepted) {
+            // We ran out of time.
+            throw new Error(errorMessage.requestEntityTooLarge);
+        }
+
+    }
+
+    function tryQueryAndHardDelete(eTag) {
+        // Find the resource and all of its history.
```
jackliums: does this comment apply?

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/StoredProcedures/HardDelete/hardDeleteIdentityProvider.js
```diff
@@ -0,0 +1,139 @@
...
+
+        if (!isQueryAccepted) {
+            // We ran out of time.
+            throw new Error(errorMessage.requestEntityTooLarge);
+        }
+    }
+
+    function tryHardDelete(documents, eTag) {
+        if (documents.length > 0) {
+            deletedResourceIdList.push(documents[0].id);
```
jackliums: The query above query by id and therefore we will have either 0 or 1 document. I don't think we need all of this logic to push into array and return?

In fact, since it's just deleting one document, you don't really need stored procedure? Is there additional things that will be deleted later or something?

src/Microsoft.Health.CosmosDb/Features/Queries/QueryHelper.cs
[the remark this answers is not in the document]
jackliums: In FHIR codebase, these methods are not exposed as publich method so the ordering is implementation detail. If this is exposed and can be used generically, then the class should ensure that it generates correct SQL statement.

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/StoredProcedures/HardDelete/hardDeleteIdentityProvider.js
brandonpollett: The idea was to put additional validation on delete within the stored procedure. There are certain cases that we want to protect against such as not deleting the last identity provider.

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/StoredProcedures/HardDelete/hardDeleteIdentityProvider.js
garamamo: It worked fine while I tested. I think it is invalid only of it is a cross partition query I believe?

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/StoredProcedures/HardDelete/hardDeleteIdentityProvider.js
garamamo: I am passing in for every function. Not sure if this is global,

src/Microsoft.Health.ControlPlane.Core/Features/Rbac/IdentityProvider.cs
garamamo: This is because there is already a eTag on the derived class that is deserialied from the document.

src/Microsoft.Health.ControlPlane.CosmosDb/Features/Storage/StoredProcedures/HardDelete/hardDeleteIdentityProvider.js
garamamo: Updated the stored proc.

src/Microsoft.Health.ControlPlane.Core/AssemblyInfo.cs
garamamo: There is a ToIdentityProvider method on the assembly that is used by controlplane.

src/Microsoft.Health.ControlPlane.Core/Features/Rbac/IdentityProvider.cs
garamamo: Brandon and I spoke. We will have to test this when we write integration tests. At that point if there is an issue, we can register one using Dependency Injection.

src/Microsoft.Health.ControlPlane.Core/Features/Persistence/IControlPlaneDataStore.cs
garamamo: The microsoft standards guidelines for apis uses eTag. I plan to leave it as is.

Issue #319, opened 2019-01-08
_summary=count no longer returns the total count


jackliums:
**Describe the bug**
There is a regression bug where _summary=count no longer returns the total count.

**To Reproduce**
1. /Patient?_summary=count

**Expected behavior**
The total number of resources matching the criteria should be returned.

**Actual behavior**
The total count is not returned.

jackliums:
This was fixed by #320

PR #320, opened 2019-01-08
Fixed the regression bug where the total count was not being set.

jackliums:
## Description
Fix the issue where the total count was not returned when _summary=count is specified.

## Related issues
Addresses #319.

## Testing
Added E2E test.

### DIFF
--- a/test/Microsoft.Health.Fhir.Tests.E2E/Rest/Search/BasicSearchTests.cs
+++ b/test/Microsoft.Health.Fhir.Tests.E2E/Rest/Search/BasicSearchTests.cs
@@ -219,10 +219,10 @@ namespace Microsoft.Health.Fhir.Tests.E2E.Rest.Search
 
             var tag = new Coding(string.Empty, Guid.NewGuid().ToString());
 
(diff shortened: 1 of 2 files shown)


test/Microsoft.Health.Fhir.Tests.E2E/Rest/Search/BasicSearchTests.cs
```diff
@@ -210,5 +210,30 @@ public async Task GivenResourceWithHistory_WhenSearchedWithParams_ThenOnlyCurren
...
+        [Trait(Traits.Priority, Priority.One)]
+        public async Task GivenListOfResources_WhenSearchedWithCountSummary_ThenTotalCountShouldBeReturned()
+        {
+            const int numberOfResources = 5;
+
+            var tag = new Coding(string.Empty, Guid.NewGuid().ToString());
+
+            for (int i = 0; i < numberOfResources; i++)
+            {
+                Patient patient = Samples.GetDefaultPatient();
```
brandonpollett: Do we need to create the patient 5 times or can we just create 5 new patients using the same patient object?

PR #332, opened 2019-01-09
Added the integration tests for audit logs.

jackliums:
## Description
Added the integration tests for the audit logs.

## Related issues
Addresses #309.

## Testing
Integration tests.

### DIFF
--- a/src/Microsoft.Health.Fhir.Api/Features/Context/FhirRequestContextMiddleware.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/Context/FhirRequestContextMiddleware.cs
@@ -13,6 +13,8 @@ namespace Microsoft.Health.Fhir.Api.Features.Context
 {
     public class FhirRequestContextMiddleware
     {
+        private const string RequestIdHeaderName = "X-Request-Id";
+
         private readonly RequestDelegate _next;
 
         public FhirRequestContextMiddleware(RequestDelegate next)
@@ -38,14 +40,18 @@ namespace Microsoft.Health.Fhir.Api.Features.Context
                 request.Path,
                 request.QueryString);
 
+            string correlationId = correlationIdProvider.Invoke();
+
             var fhirRequestContext = new FhirRequestContext(
                 method: request.Method,
                 uriString: uriInString,
                 baseUriString: baseUriInString,
                 requestType: ValueSets.AuditEventType.RestFulOperation,
-                correlationId: correlationIdProvider.Invoke(),
+                correlationId: correlationId,
                 requestHeaders: context.Request.Headers,
                 responseHeaders: context.Response.Headers);
+
+            context.Response.Headers[RequestIdHeaderName] = correlationId;
 
             if (context.User != null)
             {
--- a/test/Microsoft.Health.Fhir.Tests.E2E/Rest/Audit/AuditTests.cs
+++ b/test/Microsoft.Health.Fhir.Tests.E2E/Rest/Audit/AuditTests.cs
@@ -65,6 +65,7 @@ namespace Microsoft.Health.Fhir.Tests.E2E.Rest.Audit
         [Fact]
         public async Task GivenANonExistingResource_WhenRead_ThenAuditLogEntriesShouldBeCreated()
         {
+            // TODO: The resource type being logged here is incorrect. The issue is tracked by https://github.com/Microsoft/fhir-server/issues/334.
             await ExecuteAndValidate(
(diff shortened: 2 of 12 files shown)


src/Microsoft.Health.Fhir.Api/Features/Context/FhirRequestContextMiddleware.cs
```diff
@@ -13,6 +13,8 @@ namespace Microsoft.Health.Fhir.Api.Features.Context
 {
     public class FhirRequestContextMiddleware
     {
+        private const string RequestIdHeaderName = "X-Request-Id";
```
jackliums: We have some headers that are in all lower cases (x-ms-session-token, x-ms-consistency-level) and some that are in Pascal casing (X-Request-Id, X-Content-Type-Options). Since the X-Request-Id is already being exposed in the PaaS side, I used that but should we standardize them?

test/Microsoft.Health.Fhir.Tests.E2E/Rest/Audit/AuditTests.cs
```diff
@@ -0,0 +1,377 @@
...
+                    }
+                    catch (FhirException ex)
+                    {
+                        result = ex.Response;
+                    }
+
+                    return result;
+                },
+                "read",
+                ResourceType.OperationOutcome,
```
jackliums: There are some issues with the existing audit event logging logic. For example, when a non-existing resource is read, it logs the resource type as OperationOutcome, which is incorrect. In the delete test below, it logs the resource type as null, which I think is incorrect as well. In both cases, we should log the actual resource type where the operation is being executed on, which is Patient in these two cases. I will file a separate issue to fix that.

test/Microsoft.Health.Fhir.Tests.E2E/Rest/Audit/AuditTests.cs
```diff
@@ -0,0 +1,375 @@
...
+            // Create a new client with no token supplied.
+            var client = new FhirClient(_fixture.CreateHttpClient(), ResourceFormat.Json);
+
+            await clientSetup(client);
+
+            FhirResponse<OperationOutcome> response = null;
+
+            try
+            {
+                await client.ReadAsync<Patient>(url);
```
cunninghamjc: Should we Assert.Fail() after this, in case the call actually succeeds?  The test will still fail later with a null ref exception, but an explicit failure here might be clearer.

test/Microsoft.Health.Fhir.Tests.E2E/Rest/Audit/AuditTests.cs
```diff
@@ -0,0 +1,375 @@
...
+
+        private void ValidateAuditEntry(AuditEntry auditEntry, AuditAction expectedAuditAction, string expectedAction, ResourceType? expectedResourceType, Uri expectedUri, HttpStatusCode? expectedStatusCode, string expectedCorrelationId)
+        {
+            Assert.NotNull(auditEntry);
+            Assert.Equal(expectedAuditAction, auditEntry.AuditAction);
+            Assert.Equal(expectedAction, auditEntry.Action);
+            Assert.Equal(expectedResourceType?.ToString(), auditEntry.ResourceType);
+            Assert.Equal(expectedUri, auditEntry.RequestUri);
+            Assert.Equal(expectedStatusCode, auditEntry.StatusCode);
+            Assert.Equal(expectedCorrelationId, auditEntry.CorrelationId);
```
cunninghamjc: >); [](start = 72, length = 2)

Could we add a comment about why we're not validating the claims here?  Just so we know it's by design and not a mistake.

jackliums: Added the comment. We can probably do this by extending the IdentityServer but that will be separate work. I will file an issue for that.

test/Microsoft.Health.Fhir.Tests.E2E/Rest/Audit/AuditTests.cs
jackliums: Updated the code to make sure it throws exception.

test/Microsoft.Health.Fhir.Tests.E2E/Rest/Audit/AuditTests.cs
jackliums: Filed issue and added comments.

PR #333, opened 2019-01-12
Make it build on linux (Ubuntu 18.10)

jmandel:
## Description
Hacking at the FHIR connectathon and trying to make `fhir-server` build on linux. I made these changes to get `dotnet build` to succeed. I expect they're unsuitable to merge back, since I don't really understand what I'm doing :-)

(Though https://github.com/Microsoft/fhir-server/compare/master...jmandel:fix-linux-build?expand=1#diff-b497308945bef2a867e474987d9fccebR24 perhaps is a real change that's worth making.)

## Testing

I just ran `dotnet build` on my own machine and called it a day when I saw:

```
Build succeeded.
    0 Warning(s)
    0 Error(s)

Time Elapsed 00:00:17.44
```


hansenms:
We do build on Linux but Ubuntu 1604. We should probably switch to 1804, so maybe we should put in an issue for that to track the work? We probably will not be going there in the next few weeks but we should.

This PR breaks the windows build.

brandonpollett:
@jmandel It looks like you probably have the .net core SDK 2.2 installed. Can you install SDK 2.1.403 from https://dotnet.microsoft.com/download/dotnet-core/2.1 and give it a go from master?

jmandel:
Thanks for the suggestion! I can only easily install `2.1.503`, which gives me the same set of errors I saw with `2.2`.

jmandel:
Trying now with `v2.1.403` directly, I still get lots of issues relating to `Microsoft.Build.Tasks.Git`, like:
```
/home/jmandel/.nuget/packages/microsoft.build.tasks.git/1.0.0-beta-63127-02/build/Microsoft.Build.Tasks.Git.targets(20,5): error MSB4018: The "Microsoft.Build.Tasks.Git.LocateRepository" task failed unexpectedly. [/home/jmandel/work/fhir-server/src/Microsoft.Health.Fhir.ValueSets/Microsoft.Health.Fhir.ValueSets.csproj]
```

brandonpollett:
I believe that issue probably has to do with a setting in the `Directory.Build.props`. We have disabled source linking on OSX, but it looks like we probably also need to disable it on Linux. There's a condition for `$([MSBuild]::IsOSPlatform('osx'))` and it needs to be updated to something like `$([MSBuild]::IsOSPlatform('osx')) Or $([MSBuild]::IsOSPlatform('linux'))`. 

Alternatively we could update it to check for only windows and flip the flag we're setting. 

Unfortunately, I don't have a linux box handy to test this on currently.

jmandel:
Thanks @brandonpollett  -- that helps a lot. To make it build I also had to remove the `sdk.version` from `global.json`; unclear if this is expected. I've pushed an update to this branch.

I also see two warnings about over-specified package versions:

```
/usr/share/dotnet/sdk/2.2.102/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.DefaultItems.targets(153,5): warning NETSDK1071: A PackageReference to 'Microsoft.AspNetCore.App' specified a Version of `2.1.6`. Specifying the version of this package is not recommended. For more information, see https://aka.ms/sdkimplicitrefs [/home/jmandel/work/fhir-server/src/Microsoft.Health.Fhir.Web/Microsoft.Health.Fhir.Web.csproj]

/usr/share/dotnet/sdk/2.2.102/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.DefaultItems.targets(153,5): warning NETSDK1071: A PackageReference to 'Microsoft.AspNetCore.App' specified a Version of `2.1.6`. Specifying the version of this package is not recommended. For more information, see https://aka.ms/sdkimplicitrefs [/home/jmandel/work/fhir-server/src/Microsoft.Health.Fhir.Web/Microsoft.Health.Fhir.Web.csproj]
```

brandonpollett:
The `global.json` setting is there so that we can explicitly target a .net sdk version. Without it, the sdk runtime is determined to be the latest and we get issues like the one you referenced (`Specifying the version of this package is not recommended.`) that vary with the .net sdk version installed. 

We haven't upgraded the solution to 2.2 yet, but plan on doing that in the future.

jmandel:
Gotcha -- and there doesn't see to be a way to say `2.1.*`, for example. Thanks!

global.json
```diff
@@ -1,5 +1,4 @@
 {
   "sdk": {
-    "version": "2.1.403"
```
brandonpollett: Instead of removing this, can you update it to `2.1.503`? 

It appears that the easy to follow instructions for the linux dotnet sdk installs that version. My goal is to see how our PR builds behave when this is done.

brandonpollett:
Thanks for the update, we'll take a look at the best way to address the issue that we run into with the change to the global.json.

paule96:
@brandonpollett 
If this is important for you the code in the PR is working now. So you can build and publish by your own in your private marketplace account. (please share it only with your AzureDevOps Account)

I think we are done in a month with this PR if you have time :)

johnstairs:
@jmandel @brandonpollett I'm closing this. It builds out of the box now with the more recent bits. The key is to install the right SDK version. Currently, we are using 2.2.107, so `sudo apt-get install dotnet-sdk-2.2=2.2.107-1` on Ubuntu.
Sorry for not addressing this sooner.

Issue #339, opened 2019-02-12
Add configurable Cors Support


brandonpollett:
**User story**
As a developer, I want configurable cors support so that I can enable clients to make requests from different domains.

I should be able to configure cors with the following options. Each also supports a `*` value which equates to allow all :

| Option | Notes |
| --- | --- |
| Origins | Array of origins that may be returned in `Access-Control-Allow-Origin`. |
| Headers | Array of headers that may be returned `Access-Control-Allow-Headers`. |
| Methods | Array of methods that may be returned `Access-Control-Allow-Methods`. |
| MaxAge | Value to be returned in `Access-Control-Max-Age`. Set in seconds, defaults to 0. |
| AllowCredentials | Boolean to determine if `Access-Control-Allow-Credentials: true` should be returned |

**Acceptance criteria**
- [ ] When I do a request, then I get a response with the appropriate `Access-Control-Allow-Origin` header.
- [ ] When I do a request, then I get a response with the appropriate `Access-Control-Allow-Headers` header.
- [ ] When I do a request, then I get a response with the appropriate `Access-Control-Allow-Methods` header.
- [ ] When I do a request, then I get a response with the appropriate `Access-Control-Max-Age` header.
- [ ] When I do a request, then I get a response with the appropriate `Access-Control-Allow-Credentials` header.
- [ ] When I do an `options` request, then I get an emtpy `204` response with the appropriate headers.

PR #340, opened 2019-02-12
Add configurable cors support

brandonpollett:
## Description
Adds a configuration section to support various cors configurations. 

One thing to note is that the behavior may not be expected while we're on .net 2.1 as there is some filtering that is in place in the `CorsService.cs`. You can see the details here: https://github.com/aspnet/AspNetCore/commit/2db11051d86cd2b209989e039dc32d4b7976e54c

This should change once our code is brought into 2.2.1

## Related issues
Addresses #339.

## Testing
Initial manual testing was carried out. 

A single integration test with a permissive cors configuration is included. Additional tests are challenging due to not being able to easily modify the configuration on a deployed instance.


src/Microsoft.Health.Fhir.Core/Configs/CorsConfiguration.cs
```diff
@@ -0,0 +1,17 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+using Microsoft.Health.Fhir.Core.Features.Cors;
+
+namespace Microsoft.Health.Fhir.Core.Configs
+{
+    public class CorsConfiguration
```
johnstairs: I think we may want this to be more customizable (perhaps looking like `CorsPolicy`). Also, I would have guessed that `SupportsCredentials` would be needed for SMART on FHIR?

brandonpollett: I refactored this to be more configurable.

samples/templates/default-azuredeploy.json
```diff
@@ -147,6 +147,7 @@
             "APPINSIGHTS_SNAPSHOTFEATURE_VERSION": "1.0.0",
             "WEBSITE_NODE_DEFAULT_VERSION": "6.9.4",
             "KeyVault:Endpoint": "[variables('keyvaultEndpoint')]",
+            "FhirServer:Cors:Mode": "All",
```
johnstairs: Need to update this, right?

src/Microsoft.Health.Fhir.Core/Configs/CorsConfiguration.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        public IList<string> Origins { get; } = new List<string>();
+
+        public IList<string> Headers { get; } = new List<string>();
+
+        public IList<string> Methods { get; } = new List<string>();
+
+        public int MaxAge { get; set; }
```
brandonpollett: Should default this to -1 which disables caching.

See - https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Max-Age

test/Microsoft.Health.Fhir.Tests.E2E/Rest/CorsTests.cs
```diff
@@ -0,0 +1,56 @@
...
+    public class CorsTests : IClassFixture<HttpIntegrationTestFixture<Startup>>
+    {
+        private readonly FhirClient _client;
+
+        public CorsTests(HttpIntegrationTestFixture<Startup> fixture)
+        {
+            _client = fixture.FhirClient;
+        }
+
+        [Fact]
```
jackliums: Should we add a few tests to make sure the CORS are set correctly and the service accepts/rejects calls?

brandonpollett: The server won't reject calls based on CORS. There's some checking already in the SMART on FHIR proxy tests and this test just checks that headers come back as specified.

Currently this test is based on a permissive setup, but we could restrict it. If we restrict it we could do some negative checks in addition to a positive check.

brandonpollett: Also, we could add some unit tests to make sure that the CORS policy that is set is correct based on the configuration.

src/Microsoft.Health.Fhir.Core/Configs/CorsConfiguration.cs
brandonpollett: Turns out that .net restricts this to not be negative. I've updated this to be nullable, so that it can just not be specified.

Issue #358, opened 2019-02-28
Add Performance Test project


brendankowitz:
**User story**
As a developer, I want a performance testing project so that I can review the performance impact of changes.

**Acceptance criteria**
1. Create a Performance test project
2. Performance tests are run during the build

PR #360, opened 2019-02-28
Adds perf test project

brendankowitz:
## Description
Adds perftest project

## Related issues
Addresses #358.

### DIFF
--- a/src/Microsoft.Health.Fhir.Core.UnitTests/Performance/ResourceHandlerPerformanceTests.cs
+++ b/src/Microsoft.Health.Fhir.Core.UnitTests/Performance/ResourceHandlerPerformanceTests.cs
@@ -1,0 +1,98 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Linq;
+using System.Net.Http;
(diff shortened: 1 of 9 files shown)


src/Microsoft.Health.Fhir.Core.UnitTests/Performance/ResourceHandlerPerformanceTests.cs
```diff
@@ -0,0 +1,101 @@
...
+
+            ServiceProvider provider = collection.BuildServiceProvider();
+            _mediator = new Mediator(type => provider.GetService(type));
+
+            resource = Samples.GetDefaultObservation();
+            resource.Id = null;
+        }
+
+        [Benchmark]
+        public async Task GivenAFhirMediator_WhenSavingAResourceWithNoId_ThenAnIdShouldBeAssigned()
```
johnstairs: I wonder if this naming formula right for perf tests.

src/Microsoft.Health.Fhir.Core.UnitTests/Performance/ResourceHandlerPerformanceTests.cs
```diff
@@ -0,0 +1,101 @@
...
+        {
+            await _mediator.UpsertResourceAsync(resource);
```
johnstairs: It's not immediately obvious to me what this is measuring. Maybe add an explanation?

johnstairs:
Nice to get a start on this! Have you thought about where we would run these and how we would track changes across builds?

Issue #294, opened 2018-12-18
Migrate to use the built-in ASP.Net Core health check


jackliums:
**User story**
As an administrator of the server, I want to have health check so that I can monitor the health of the service.

We already have the health check implemented but ASP.Net Core 2.2 now has health check functionality baked in with the framework so we should switch to that.

**Acceptance criteria**
1. When the server can connect to Cosmos DB, then health check returns healthy.
2. When the server cannot connect to Cosmos DB, then health check returns unhealthy.

PR #361, opened 2019-03-01
Migrate to built-in ASP.NET Core health checks 

namadabu:
## Description
Currently we have our own custom health checks. Starting from 2.2 .NET Core has introduced a built-in health check feature. This PR migrates our current health check to work with the new infrastructure. Have also added a health check for the control plane cosmos db. Removed some code related to our custom health checks.

## Related issues
Addresses [issue #294 ].

## Testing
Existing unit tests and integration tests 
Tested manually on a local deployment


src/Microsoft.Health.Fhir.Web/Startup.cs
```diff
@@ -24,6 +29,10 @@ public virtual void ConfigureServices(IServiceCollection services)
             services.AddControlPlaneCosmosDb(Configuration).AddDevelopmentIdentityProvider(Configuration);
 
             services.AddFhirServer(Configuration).AddFhirServerCosmosDb(Configuration);
+
+            services.AddHealthChecks()
+                .AddCheck<FhirCosmosHealthCheck>(name: "FhirCosmosDb", failureStatus: HealthStatus.Unhealthy)
```
jackliums: I don't think we should have the name here as that's yet another place we have to maintain the name (By the way the name is 'fhirCosmosDb' and 'controlPlaneCosmosDb').

jackliums: Perhaps one way is that since you already have `FhirCosmosHealthCheck` and `ControlPlaneCosmosHealthCheck`, instead of passing the `name` here, have the name returned by those two classes.

src/Microsoft.Health.Fhir.CosmosDb/Features/Health/FhirCosmosHealthCheck.cs
```diff
@@ -42,39 +32,16 @@ public class FhirCosmosHealthCheck : IHealthCheck
...
-                return HealthCheckResult.Healthy("Successfully connected to the data store.");
-            }
-            catch (Exception ex)
-            {
-                _logger.LogWarning(ex, "Failed to connect to the data store.");
-
-                return HealthCheckResult.Unhealthy("Failed to connect to the data store.");
-            }
-        }
+        public override CosmosCollectionConfiguration CollectionConfiguration { get; }
```
jackliums: The return value is the same. Instead of overriding this, you could simply pass the `CollectionConfigurationName` to the base class instead?

src/Microsoft.Health.Fhir.Web/Startup.cs
namadabu: This name is used in the final healthreport that is generated after the healthcheck. Something like:
`{"overallStatus":"Healthy","details":[{"name":"FhirCosmosDb","status":"Healthy","exception":null},{"name":"ControlPlaneCosmosDb","status":"Healthy","exception":null}]}`

`{"overallStatus":"Unhealthy","details":[{"name":"FhirCosmosDb","status":"Unhealthy","exception":null},{"name":"ControlPlaneCosmosDb","status":"Healthy","exception":null}]}`

Will update to use `nameof(<classname>)`

src/Microsoft.Health.Fhir.Web/Startup.cs
```diff
@@ -32,6 +41,11 @@ public virtual void Configure(IApplicationBuilder app)
             app.UseFhirServer();
 
             app.UseDevelopmentIdentityProvider();
+
+            app.UseHealthChecks("/health/check", new HealthCheckOptions
```
brendankowitz: This should probably be done inside `UseFhirServer`

src/Microsoft.Health.Fhir.Web/Startup.cs
```diff
@@ -24,6 +29,10 @@ public virtual void ConfigureServices(IServiceCollection services)
             services.AddControlPlaneCosmosDb(Configuration).AddDevelopmentIdentityProvider(Configuration);
 
             services.AddFhirServer(Configuration).AddFhirServerCosmosDb(Configuration);
+
+            services.AddHealthChecks()
+                .AddCheck<FhirCosmosHealthCheck>(name: nameof(FhirCosmosHealthCheck), failureStatus: HealthStatus.Unhealthy)
+                .AddCheck<ControlPlaneCosmosHealthCheck>(name: nameof(ControlPlaneCosmosHealthCheck), failureStatus: HealthStatus.Unhealthy);
```
brendankowitz: This should probably move into `AddFhirServerCosmosDb`

src/Microsoft.Health.CosmosDb/Features/Health/ResponseWriter.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+using Newtonsoft.Json;
+
+namespace Microsoft.Health.CosmosDb.Features.Health
+{
+    public static class ResponseWriter
+    {
+        public static async Task HealthResponseWriter(HttpContext httpContext, HealthReport healthReport)
```
brendankowitz: Perhaps just call this method `WriteJson` which is in line with the built-in (HealthCheckResponseWriters.WriteMinimalPlaintext)

src/Microsoft.Health.CosmosDb/Features/Health/CosmosHealthCheck.cs
```diff
@@ -0,0 +1,80 @@
...
+            EnsureArg.IsNotNull(documentClient, nameof(documentClient));
+            EnsureArg.IsNotNull(configuration, nameof(configuration));
+            EnsureArg.IsNotNull(namedCosmosCollectionConfigurationAccessor, nameof(namedCosmosCollectionConfigurationAccessor));
+            EnsureArg.IsNotNullOrWhiteSpace(collectionConfigurationName, nameof(collectionConfigurationName));
+            EnsureArg.IsNotNull(testProvider, nameof(testProvider));
+            EnsureArg.IsNotNull(logger, nameof(logger));
+
+            _documentClient = documentClient;
+
+            EnsureArg.IsNotNull(_documentClient, optsFn: options => options.WithMessage("Factory returned null."));
```
brendankowitz: Isn't this already checked by https://github.com/Microsoft/fhir-server/pull/361/files#diff-8b7f39e471a45c4b5fe6f5db9d6b4662R45?

src/Microsoft.Health.CosmosDb/Features/Health/HealthCheckResponseWriter.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Collections.Generic;
+using System.Net.Mime;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+using Newtonsoft.Json;
+
+namespace Microsoft.Health.CosmosDb.Features.Health
+{
+    public static class HealthCheckResponseWriter
```
brendankowitz: I think we can safely move this class to the API project.
Then remove the reference from API[Cosmos]. The api project shouldn't have a reference to a specific database implementation

namadabu: Agree. I think I will just make it an anonymous method instead of having a separate class/method for the response writer. We can refactor it later if we the case for re-using it comes up.

src/Microsoft.Health.CosmosDb/Features/Health/CosmosHealthCheck.cs
namadabu: I copy pasted this from the previous health check ctor. Will remove it.

src/Microsoft.Health.Fhir.Api/Registration/FhirServerApplicationBuilderExtensions.cs
```diff
@@ -18,6 +25,34 @@ public static IApplicationBuilder UseFhirServer(this IApplicationBuilder app)
...
+                    foreach (var reportEntry in healthReport.Entries)
+                    {
+                        var individualEntity = new
+                        {
+                            name = reportEntry.Key,
+                            status = Enum.GetName(typeof(HealthStatus), reportEntry.Value.Status),
+                        };
+
+                        allEntities.Add(individualEntity);
+                    }
```
brendankowitz: Can be collapsed
```
var allEntities = healthReport.Select(x => new
                        {
                            name = reportEntry.Key,
                            status = Enum.GetName(typeof(HealthStatus), reportEntry.Value.Status),
                        }).ToArray();
```

src/Microsoft.Health.Fhir.Api/Microsoft.Health.Fhir.Api.csproj
```diff
@@ -38,6 +39,7 @@
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.Health.Abstractions\Microsoft.Health.Abstractions.csproj" />
     <ProjectReference Include="..\Microsoft.Health.ControlPlane.Core\Microsoft.Health.ControlPlane.Core.csproj" />
+    <ProjectReference Include="..\Microsoft.Health.CosmosDb\Microsoft.Health.CosmosDb.csproj" />
```
brendankowitz: Should be able to remove this reference now

Issue #373, opened 2019-03-05
Use deterministic runtime versions


johnstairs:
We do not target a specific runtime version, so the version is determined by whatever the latest SDK is installed on the system. We should make that explicit.

PR #377, opened 2019-03-06
Force specific SDK and runtime versions.

johnstairs:
## Description
.NET runtime version 2.2.2 has a [bug](https://github.com/aspnet/AspNetCore/issues/7975) that breaks our end to end tests when run locally. So we want to make sure to use 2.2.1. 

There are two default behaviors that work against this goal:
1. At runtime, the latest patch version of the runtime is used. So if you happen to have 2.2.2 installed on your system, it will use that instead of the requested 2.2.1.
2. If the SDK version in global.json is not installed on the system, the latest installed SDK is used instead.

For (1), in order to prevent the lastest runtime patch from being used, we inject:

``` json
"applyPatches": false
```
into the runtimeconfig.json file

For (2), we want to make sure that only the specified sdk is used. The reason for this is to be able to target the specific runtime version that the SDK ships with. To enforce this, I created an msbuild task that reads the global.json file and compares that version with the actual sdk version used during the build. If they don't match the build fails.

## Related issues
Addresses #373


test/Microsoft.Health.Fhir.Tests.E2E/Microsoft.Health.Fhir.Tests.E2E.csproj
```diff
@@ -50,7 +50,27 @@
     </None>
   </ItemGroup>
   <Import Project="Microsoft.AspNetCore.Mvc.Testing.targets" />
+
+  <Target Name="VerifyExactSdkVersion" BeforeTargets="Build">
```
brendankowitz: Can/should any of this go in Build.props? Is it self-explanatory enough just to throw the error here?

johnstairs: We don't want this to evaluate for every project. One project is enough. This seemed like a good one to me :). What would you add to the error message?

brendankowitz: I was just referring to only one project failing to build. In regards to the message, we could reference the download url maybe (https://dotnet.microsoft.com/download), but I think it's fine otherwise. Folks who download the code will probably see this sooner or later

johnstairs: Updated the message

Issue #368, opened 2019-03-04
App Service and App Insights not working with .NET Core 2.2


hansenms:
**Describe the bug**
After upgrading to .NET Core 2.2, App Insights is no longer receiving telemetry when app is deployed in app service. 

https://docs.microsoft.com/en-us/azure/azure-monitor/app/azure-web-apps

**To Reproduce**
1. Deploy current template 
2. Send requests to server
3. Check App Insights

**Expected behavior**
Request logs, etc. should be sent to app insights.

**Actual behavior**
No logs are sent to app insights.

PR #380, opened 2019-03-07
Application Insights in code

johnstairs:
## Description
Rather than relying on App Service's Application Insights support (which does not yet support asp.net core 2.2), we App Insights directly from code, passing in the instrumentation key as a config setting.

## Related issues
Addresses #368 

## Testing
Verified telemetry is flowing.


hansenms:
I would propose adding this to `Startup.cs` instead:

```
services.AddApplicationInsightsTelemetry(Configuration);
```

Apparently user secrets are not read by `UseApplicationInsights()`, https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/695

Issue #370, opened 2019-03-04
Request with _format but no Accept header returns 500


cunninghamjc:
**Describe the bug**
A request with _format in the query string but no Accept header returns 500.

**To Reproduce**
1. Issue a request with the _format param, e.g. https://<hostname>/Patient?_format=json, and no Accept header.

**Expected behavior**
200 with the results in the specified format

**Actual behavior**
500 Internal Server Error

PR #381, opened 2019-03-07
Handle missing Accept header when _format is provided

cunninghamjc:
## Description
Fixed the format/content-type resolution code to handle the case where _format is specified in the query string but the Accept header is not provided.

## Related issues
Addresses #370 

## Testing
Added unit tests

### DIFF
--- a/src/Microsoft.Health.Fhir.Api/Features/ContentTypes/ContentTypeService.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/ContentTypes/ContentTypeService.cs
@@ -55,7 +55,7 @@ namespace Microsoft.Health.Fhir.Api.Features.ContentTypes
                     throw new UnsupportedMediaTypeException(Resources.UnsupportedFormatParameter);
                 }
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.Health.Fhir.Api/Features/ContentTypes/ContentTypeService.cs
```diff
@@ -55,7 +55,7 @@ public async Task CheckRequestedContentTypeAsync(HttpContext httpContext)
                     throw new UnsupportedMediaTypeException(Resources.UnsupportedFormatParameter);
                 }
 
-                string closestClientMediaType = _outputFormatters.GetClosestClientMediaType(resourceFormat, acceptHeaders.Select(x => x.MediaType.Value));
+                string closestClientMediaType = _outputFormatters.GetClosestClientMediaType(resourceFormat, acceptHeaders?.Select(x => x.MediaType.Value));
```
johnstairs: Any chance MediaType is null?

cunninghamjc: No, it's a StringSegment, which is a struct.  The documentation says it defaults to Empty.

Issue #342, opened 2019-02-23
Support for $export operation API call


jackliums:
**User story**

As an engineer, I want to call $export operation so that I have the endpoint ready for queuing a bulk export job in the future.

**Acceptance criteria**

1. Verify there is a feature flag to manage the API.
1. Verify the HTTP status code for the following endpoints: `[base]/Patient/$export`, `[base]/Group/[id]/$export`, `[base]/$export`
   - Verify `501 Not Implemented` is returned if feature flag is turned on.
   - Verify `400 Bad Request` is returned if feature flag is turned off.
1. Verify `400 Bad Request` is returned if any other endpoint is called with $export.
1. Verify `400 Bad Request` is returned if `Accept` header is not supplied or if supplied and the value is not `application/fhir+json`.
1. Verify `400 Bad Request` is returned if `Prefer` header is not supplied or if supplied and the value is not 'respond-async'.
1. Verify `400 Bad Request` is returned if any other search parameters are provided.
1. Verify in case of failure, `OperationOutcome` is returned with error message describing the issue.

PR #382, opened 2019-03-08
Add skeleton support for $export operation

namadabu:
## Description
This change just adds the initial framework for supporting $export operations.

- Added a separate OperationsController to handle requests related to operations. Currently it just handled export operation request, but might expand in the future depending on what operations we might support.
- Added 3 export routes and the corresponding action methods. Currently they just do some validation and return either `BadRequest` or `NotImplemented`.
- Added a filter attribute to validate the Accept and Prefer headers for export requests. 
- Added a feature flag to enable to disable export operations for the fhir-server

## Related issues
Addresses [issue #342 ].

## Testing
UTs have been added for the controller and filter attribute.
E2E tests


src/Microsoft.Health.Fhir.Api/Controllers/OperationsController.cs
```diff
@@ -0,0 +1,128 @@
...
+using Microsoft.Extensions.Options;
+using Microsoft.Health.Fhir.Api.Features.ActionResults;
+using Microsoft.Health.Fhir.Api.Features.Filters;
+using Microsoft.Health.Fhir.Api.Features.Routing;
+using Microsoft.Health.Fhir.Core.Configs;
+using Microsoft.Health.Fhir.Core.Features.Context;
+
+namespace Microsoft.Health.Fhir.Api.Controllers
+{
+    [ServiceFilter(typeof(OperationOutcomeExceptionFilterAttribute))]
```
namadabu: Need to add the other required filter attributes

namadabu:
I still need to clean up some of the details and add integration tests. But just wanted to make sure my overall implementation/approach is correct.

src/Microsoft.Health.Fhir.Api/Controllers/OperationsController.cs
```diff
@@ -0,0 +1,128 @@
...
+        {
+            HttpStatusCode returnCode;
+            OperationOutcome result;
+
+            if (_operationsConfig.SupportsExport)
+            {
+                result = GenerateOperationOutcome(
+                    OperationOutcome.IssueSeverity.Error,
+                    OperationOutcome.IssueType.NotSupported,
+                    "Export operation not supported");
```
brandonpollett: Should be in resources resx.

src/Microsoft.Health.Fhir.Api/Controllers/OperationsController.cs
```diff
@@ -0,0 +1,128 @@
...
+            {
+                OperationOutcome result = GenerateOperationOutcome(
+                    OperationOutcome.IssueSeverity.Error,
+                    OperationOutcome.IssueType.NotSupported,
+                    $"{type} type not supported for Export by ResourceType operation");
+
+                return FhirResult.Create(result, HttpStatusCode.BadRequest);
+            }
+
+            return Export();
```
brandonpollett: Are we falling back to this just to handle the logic on if it's enabled or not?

src/Microsoft.Health.Fhir.Web/appsettings.json
```diff
@@ -31,6 +31,9 @@
             "Headers": [],
             "MaxAge": null,
             "AllowCredentials": false
+        },
+        "Operations": {
+            "SupportsExport": false
```
brandonpollett: Is there any type of configuration that we're going to need beyond if export is supported or not?

namadabu: I don't think there will be more configurations under export. We will be adding more to Operations as we support more (example: import).

src/Microsoft.Health.Fhir.Api/Controllers/OperationsController.cs
namadabu: Yes. This will change once we add the actual functionality. I will add some comments to the code regarding this.

src/Microsoft.Health.Fhir.Api/Controllers/OperationsController.cs
```diff
@@ -0,0 +1,128 @@
...
+            return FhirResult.Create(result, returnCode);
+        }
+
+        [HttpGet]
+        [Route(KnownRoutes.ExportResourceType)]
+        [ValidateOperationHeadersFilter]
+        public IActionResult ExportResourceType(string type)
+        {
+            // Export by ResourceType is supported only for Patient resource type.
+            if (!string.Equals(type, "Patient", StringComparison.Ordinal))
```
namadabu: Should be using `ResourceType.Patient` instead of a string literal

src/Microsoft.Health.Fhir.Api/Controllers/OperationsController.cs
```diff
@@ -0,0 +1,128 @@
...
+            return Export();
+        }
+
+        [HttpGet]
+        [Route(KnownRoutes.ExportResourceTypeById)]
+        [ValidateOperationHeadersFilter]
+        public IActionResult ExportResourceTypeById(string type, string id)
+        {
+            // Export by ResourceTypeId is supported only for Group resource type.
+            if (!string.Equals(type, "Group", StringComparison.Ordinal) || string.IsNullOrEmpty(id))
```
namadabu: Should be using `ResourceType.Group' instead of string literal

src/Microsoft.Health.Fhir.Web/appsettings.json
namadabu: Seems like I am mistaken. There are other settings we want to add when export is enabled (like concurrency, job timeout, maximum filesize. It seems like a good idea to move the export config into a separate section instead of grouping under "Operations"

brandonpollett: We can either move it to it's own section or could just make each thing under operations an object. Not sure right now which would be the better decision.

namadabu: I just finished renaming everything to be Export specific (rather than Operations) so that things look cleaner and more readable (at least for now). If we add more operations, I thought it would be better to have their own controllers and appropriate filters.

src/Microsoft.Health.Fhir.Api/Features/Filters/ValidateExportHeadersFilterAttribute.cs
```diff
@@ -0,0 +1,54 @@
...
+                    Code = OperationOutcome.IssueType.Invalid,
+                    Diagnostics = Resources.UnsupportedAcceptHeader,
+                };
+
+                throw new ResourceNotValidException(new OperationOutcome.IssueComponent[] { error });
+            }
+
+            if (!context.HttpContext.Request.Headers.TryGetValue("Prefer", out var preferHeaderValue) ||
+                preferHeaderValue.Count != 1 ||
+                !string.Equals(preferHeaderValue[0], "respond-async", StringComparison.Ordinal))
```
brandonpollett: We generally use `const`s instead of inline strings.

src/Microsoft.Health.Fhir.Api/Controllers/ExportController.cs
```diff
@@ -0,0 +1,134 @@
...
+        [Route(KnownRoutes.Export)]
+        [ValidateExportHeadersFilter]
+        public IActionResult Export()
+        {
+            HttpStatusCode returnCode;
+            OperationOutcome result;
+
+            if (_exportConfig.SupportsExport)
+            {
+                result = GenerateOperationOutcome(
```
brandonpollett: We should just be able to throw the appropriate exceptions here instead of newing up fhirresults and outcomes. 

Doing this should take care of setting the appropriate status code and operation outcomes.

exporttestconfiguration.json
```diff
@@ -0,0 +1,7 @@
+{
```
brendankowitz: Can we move these json files to ./test/configuration?

src/Microsoft.Health.Fhir.Api.UnitTests/Controllers/ExportControllerTests.cs
```diff
@@ -0,0 +1,97 @@
...
+            var exportController = GetController(new ExportConfiguration() { SupportsExport = false });
+
+            var result = exportController.Export() as FhirResult;
+
+            Assert.NotNull(result);
+            Assert.Equal(HttpStatusCode.BadRequest, result.StatusCode);
+        }
+
+        [Fact]
+        public void GivenSupportsExportIsEnabled_WhenRequestingExport_ThenNotImplementedResponseShouldBeReturned()
```
brendankowitz: Could be reworded to something like `GivenAnExportRequest_WhenEnabled_`...

src/Microsoft.Health.Fhir.Api/Features/Filters/ValidateExportHeadersFilterAttribute.cs
```diff
@@ -0,0 +1,54 @@
...
+                !string.Equals(acceptHeaderValue[0], "application/fhir+json", StringComparison.Ordinal))
+            {
+                var error = new OperationOutcome.IssueComponent()
+                {
+                    Severity = OperationOutcome.IssueSeverity.Error,
+                    Code = OperationOutcome.IssueType.Invalid,
+                    Diagnostics = Resources.UnsupportedAcceptHeader,
+                };
+
+                throw new ResourceNotValidException(new OperationOutcome.IssueComponent[] { error });
```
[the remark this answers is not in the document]
namadabu: I initially thought it would be okay to bucket this into ResourceNotValid. But now that I think about it more, it seems better to create a new exception type specific to invalid requests.

src/Microsoft.Health.Fhir.Api/Features/Filters/ValidateExportHeadersFilterAttribute.cs
```diff
@@ -0,0 +1,43 @@
...
+
+namespace Microsoft.Health.Fhir.Api.Features.Filters
+{
+    /// <summary>
+    /// A filter that validates the Accept and Prefer headers in the request and short-circuits the pipeline if they are invalid.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method)]
+    internal class ValidateExportHeadersFilterAttribute : ActionFilterAttribute
+    {
+        private const string _preferHeaderName = "Prefer";
```
cunninghamjc: Nit: naming convention for constants is Pascal casing -- PreferHeaderName

exporttestconfiguration.json
namadabu: Since moving the testauthenvironment.json might not be straightforward (might have to update corresponding PS scripts that use the file), it think it would be better to do it in a separate follow-up PR. Thoughts?

src/Microsoft.Health.Fhir.Api.UnitTests/Controllers/ExportControllerTests.cs
```diff
@@ -0,0 +1,89 @@
...
+using Xunit;
+
+namespace Microsoft.Health.Fhir.Api.UnitTests.Controllers
+{
+    public class ExportControllerTests
+    {
+        [Fact]
+        public void GivenAnExportRequest_WhenDisabled_ThenBadRequestResponseShouldBeReturned()
+        {
+            var exportController = GetController(new ExportConfiguration() { Enabled = false });
```
jackliums: You should consider move the creation logic into one location like in constructor and save the controller as an instance variable as opposed to create it in every single test.

If you ever change the signature of the constructor, it's one place to fix as opposed to multiple places.

src/Microsoft.Health.Fhir.Api/Features/Filters/ValidateExportHeadersFilterAttribute.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        public override void OnActionExecuting(ActionExecutingContext context)
+        {
+            EnsureArg.IsNotNull(context, nameof(context));
+
+            if (!context.HttpContext.Request.Headers.TryGetValue(HeaderNames.Accept, out var acceptHeaderValue) ||
+                acceptHeaderValue.Count != 1 ||
+                !string.Equals(acceptHeaderValue[0], ContentType.JSON_CONTENT_HEADER, StringComparison.OrdinalIgnoreCase))
+            {
+                throw new RequestNotValidException(Resources.UnsupportedAcceptHeader);
```
jackliums: We should be able to make the error message generic and pass in the header name instead of have separate message for individual header right?

src/Microsoft.Health.Fhir.Api/Features/Routing/KnownRoutes.cs
```diff
@@ -26,6 +26,10 @@ internal class KnownRoutes
         public const string ResourceTypeByIdHistory = ResourceTypeById + "/" + History;
         public const string ResourceTypeByIdAndVid = ResourceTypeByIdHistory + "/" + VidRouteSegment;
 
+        public const string Export = "$export";
+        public const string ExportResourceType = ResourceType + "/" + Export;
+        public const string ExportResourceTypeById = ResourceTypeById + "/" + Export;
```
jackliums: I don't think this one is supported.

exporttestconfiguration.json
jackliums: Is it possible at least to move the cors and export configuration.json file into ./test/configuration as suggested? That way, we can also update the build script to copy the folder as opposed to individual files. We can then migrate the testauthenvironment.json later?

src/Microsoft.Health.Fhir.Api/Controllers/ExportController.cs
```diff
@@ -0,0 +1,131 @@
...
+        private readonly ExportConfiguration _exportConfig;
+
+        public ExportController(
+            ILogger<ExportController> logger,
+            IFhirRequestContextAccessor fhirRequestContextAccessor,
+            IOptions<ExportConfiguration> exportConfig)
+        {
+            EnsureArg.IsNotNull(logger, nameof(logger));
+            EnsureArg.IsNotNull(fhirRequestContextAccessor, nameof(fhirRequestContextAccessor));
+            EnsureArg.IsNotNull(exportConfig, nameof(exportConfig));
```
jackliums: validate against exportConfig?.Value to make sure exportConfig is not null and its value is also not null.

src/Microsoft.Health.Fhir.Api/Controllers/ExportController.cs
```diff
@@ -0,0 +1,131 @@
...
+        private readonly ILogger<ExportController> _logger;
+        private readonly IFhirRequestContextAccessor _fhirRequestContextAccessor;
+        private readonly ExportConfiguration _exportConfig;
+
+        public ExportController(
+            ILogger<ExportController> logger,
+            IFhirRequestContextAccessor fhirRequestContextAccessor,
+            IOptions<ExportConfiguration> exportConfig)
+        {
+            EnsureArg.IsNotNull(logger, nameof(logger));
```
jackliums: nit: can you move the logger to be the last parameter?

src/Microsoft.Health.Fhir.Api/Controllers/ExportController.cs
```diff
@@ -0,0 +1,131 @@
...
+            else
+            {
+                throw new RequestNotValidException(string.Format(Resources.UnsupportedOperation, "Export"));
+            }
+
+            return FhirResult.Create(result, returnCode);
+        }
+
+        [HttpGet]
+        [Route(KnownRoutes.ExportResourceType)]
```
jackliums: Shouldn't these method also check to see if the config is enabled or not?

test/Microsoft.Health.Fhir.Tests.E2E/Rest/ExportTests.cs
```diff
@@ -0,0 +1,84 @@
...
+        public async Task WhenRequestingExportWithInvalidPreferHeader_GivenExportIsEnabled_TheServerShouldReturnBadRequest(string preferHeaderValue)
+        {
+            HttpRequestMessage request = GenerateExportMessage(preferHeader: preferHeaderValue);
+
+            HttpResponseMessage response = await _client.SendAsync(request);
+
+            Assert.Equal(HttpStatusCode.BadRequest, response.StatusCode);
+        }
+
+        private HttpRequestMessage GenerateExportMessage(
```
jackliums: nit: maybe GenerateExportRequest?

src/Microsoft.Health.Fhir.Web/appsettings.json
```diff
@@ -31,6 +31,9 @@
       "Headers": [],
       "MaxAge": null,
       "AllowCredentials": false
+    },
+    "Export": {
```
jackliums: I wonder if ExportJob is better?

src/Microsoft.Health.Fhir.Api/Features/Filters/ValidateExportHeadersFilterAttribute.cs
namadabu: Ya, makes sense. Changed it.

src/Microsoft.Health.Fhir.Api/Features/Filters/ValidateExportHeadersFilterAttribute.cs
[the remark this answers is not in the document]
namadabu: Created a new exception type for this: `RequestNotValidException`

src/Microsoft.Health.Fhir.Api/Controllers/ExportController.cs
namadabu: These two just validate whether the `type` is expected `ResourceType` and then just call the base `Export()` method, which has the logic to respond based on whether export is enabled or not. I have moved the logic to a separate method and am calling that method from each action.

src/Microsoft.Health.Fhir.Web/appsettings.json
namadabu: As a feature name I feel like either `Export` or `BulkExport` might be better than `ExportJob`. But I don't really have a strong preference and am fine with any one of the above names.

src/Microsoft.Health.Fhir.Api.UnitTests/Controllers/ExportControllerTests.cs
namadabu: I agree with what you said. But in this scenario I need to inject a different `ExportConfiguration` depending on the test. I can't see a way to change the exportconfiguration of an existing controller unless I change the access modifier.

I can create two instances of the controller for these tests (one with config enabled and another with disabled). We would also need to create new instances for each parameter we add to `ExportConfiguration`. But that would restrict future changes to just the `ExportControllerTests()` ctor where we will have one instance for each config we are testing.

src/Microsoft.Health.Fhir.Web/appsettings.json
brandonpollett: Since this is an operation, should we have an operation section that handles the various settings for these?

Something like
```json
"Operations":{
  "Export": {
    "Enabled": true 
  }
}
```

exporttestconfiguration.json
namadabu: Moved it to test/configuration and updated the build pipeline

src/Microsoft.Health.Fhir.Web/appsettings.json
namadabu: Will update the config hierarchy to include Operations

build/package.yml
```diff
@@ -60,7 +60,14 @@ steps:
   - task: PublishBuildArtifacts@1
     displayName: 'publish corstestconfiguration.json'
     inputs:
-      pathToPublish: './corstestconfiguration.json'
+      pathToPublish: './test/configuration/corstestconfiguration.json'
```
brandonpollett: This can now be combined with the other `PublishBuidlArtifact` step below.

src/Microsoft.Health.Fhir.Api/Controllers/ExportController.cs
```diff
@@ -0,0 +1,136 @@
...
+        /// response based on whether export is enabled or not.
+        /// </summary>
+        private FhirResult CheckIfExportIsEnabledAndRespond()
+        {
+            if (!_exportConfig.Enabled)
+            {
+                throw new RequestNotValidException(string.Format(Resources.UnsupportedOperation, "Export"));
+            }
+
+            OperationOutcome result = GenerateOperationOutcome(
```
brandonpollett: We should probably create something like a `NotSupportedException` and let our normal error handling kick in instead of creating an operation outcome here.

src/Microsoft.Health.Fhir.Api/Resources.resx
```diff
@@ -167,15 +167,21 @@
...
-  </data>
   <data name="UnsupportedFormatParameter" xml:space="preserve">
     <value>The requested "_format" parameter is not supported.</value>
   </data>
+  <data name="UnsupportedHeader" xml:space="preserve">
+    <value>The requested "{0}" header is not supported.</value>
+    <comment>{0} is the header name.</comment>
+  </data>
+  <data name="UnsupportedOperation" xml:space="preserve">
+    <value>The requested {0} operation is not supported.</value>
```
brandonpollett: Might want to put quotes around the placeholder like we're doing with `UnsupportedHeader`.

src/Microsoft.Health.Fhir.Web/appsettings.json
```diff
@@ -26,11 +26,16 @@
       "InitialCollectionThroughput": null
     },
     "Cors": {
-      "Origins": [],
-      "Methods": [],
-      "Headers": [],
-      "MaxAge": null,
-      "AllowCredentials": false
+        "Origins": [],
```
brandonpollett: nit: looks like spacing changed.

build/package.yml
jackliums: Can probably use * to copy all files under the folder now right?

brandonpollett: I think you may be able to just give it a folder and it will take everything.

namadabu: I initially went the * route and the build failed ([commit]( https://github.com/Microsoft/fhir-server/pull/382/commits/0b67c56b2ba82d30861d4ab5e1e4a5a0b6af4477)). Then I realized that wildcard characters are not supported ([reference](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/publish-build-artifacts?view=azure-devops)).

I was hesitant to include the whole folder because I thought it would be better to be specific. I'll update it to include the whole configuration folder.

src/Microsoft.Health.Fhir.Api/Controllers/ExportController.cs
namadabu: Created a `OperationNotImplementException` to handle these cases.

jackliums: We should follow the existing pattern (throw exception here and have the filter convert the except into appropriate OperationOutcome.)

src/Microsoft.Health.Fhir.Api/Features/Filters/ValidateExportHeadersFilterAttribute.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        public override void OnActionExecuting(ActionExecutingContext context)
+        {
+            EnsureArg.IsNotNull(context, nameof(context));
+
+            if (!context.HttpContext.Request.Headers.TryGetValue(HeaderNames.Accept, out var acceptHeaderValue) ||
+                acceptHeaderValue.Count != 1 ||
+                !string.Equals(acceptHeaderValue[0], ContentType.JSON_CONTENT_HEADER, StringComparison.OrdinalIgnoreCase))
+            {
+                throw new RequestNotValidException(string.Format(Resources.UnsupportedHeader, HeaderNames.Accept));
```
jackliums: The message might be a bit misleading? It's not that the Accept header is not supported but more like invalid values specified?

jackliums: The message might be a bit misleading? It's not that the Accept header is not supported but more like invalid values specified?

namadabu: Updated this to UnsupportedHeaderValue instead of UnsupportedHeader

PR #386, opened 2019-03-14
Audit log valid claims by default for AAD vs. IdentityServer

cunninghamjc:
## Description
Updated the default LastModifiedClaims setting to oid for AAD, and updated test/local deployments to appid for IdentityServer.

## Related issues
Addresses #387 .

## Testing
Manually verified local deployments have LastModifiedClaims = appid.
Manually verified in App Insights that audit logs from the PR deployment contain oid.

### DIFF
--- a/src/Microsoft.Health.Fhir.Web/DevelopmentIdentityProviderConfiguration.cs
+++ b/src/Microsoft.Health.Fhir.Web/DevelopmentIdentityProviderConfiguration.cs
@@ -10,6 +10,7 @@ namespace Microsoft.Health.Fhir.Web
     public class DevelopmentIdentityProviderConfiguration
     {
         public const string Audience = "fhir-api";
+        public const string LastModifiedClaim = "appid";
 
         public bool Enabled { get; set; }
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.Health.Fhir.Web/DevelopmentIdentityProviderConfiguration.cs
```diff
@@ -10,6 +10,7 @@ namespace Microsoft.Health.Fhir.Web
     public class DevelopmentIdentityProviderConfiguration
     {
         public const string Audience = "fhir-api";
+        public const string LastModifiedClaim = "appid";
```
brandonpollett: Is there any particular reason for changing this from `claim_id` to `appid`?

cunninghamjc: I was originally looking for a claim that was in both AAD and IdentityServer tokens, so I switched client_id to appid, but now it doesn't matter anymore.  Since I already changed the integration tests to look for appid, let's keep this as appid.  I will also revert the change to StartupWithTraceAuditLogger that I made earlier, since we no longer need it.

PR #404, opened 2019-03-28
Packages source for pre-release stu3 versions

brendankowitz:
## Description
Changes to power the following CI builds:

https://microsofthealthoss.visualstudio.com/FhirServer/_build?definitionId=8


build/.vsts-package-ci.yml
```diff
@@ -0,0 +1,35 @@
+# ASP.NET Core (.NET Framework)
```
brandonpollett: Maybe could add comments on the top to say what these are used for.

Issue #409, opened 2019-04-04
SmartProxy E2E test is failing


namadabu:
**Describe the bug**
The SmartProxy E2E test is failing with the following error:
Microsoft.Health.Fhir.Tests.E2E.SmartProxy.SmartProxyTests.GivenPatientIdAndSmartAppUrl_WhenLaunchingApp_LaunchSequenceAndSignIn [FAIL]
Failed  Microsoft.Health.Fhir.Tests.E2E.SmartProxy.SmartProxyTests.GivenPatientIdAndSmartAppUrl_WhenLaunchingApp_LaunchSequenceAndSignIn
Error Message:
OpenQA.Selenium.WebDriverException : unknown error: Element <input type="submit" id="idSIButton9" class="btn btn-block btn-primary" data-bind="

**To Reproduce**
1. Just running the test against a test environment is enough to trigger the failure.

**Expected behavior**
Test should pass

**Actual behavior**
Test fails

hansenms:
@namadabu is this a persistent error or does it repeating fix it. Some of these selenium tests are a bit problematic.

namadabu:
This test started failing recently and in a consistent manner. We (Raja and @brandonpollett) figured out that we were moving through the buttons/actions too quickly. PR out to fix it - https://github.com/Microsoft/fhir-server/pull/410

PR #410, opened 2019-04-04
Fix for SmartProxy E2E test failure

namadabu:
## Description
Fixed the SmartProxy E2E test. The failure was caused because we were advancing to the next action before the previous action completed. Added delay to prevent this.

## Related issues
Addresses [issue #409 ].

## Testing
Existing tests passed.

### DIFF
--- a/test/Microsoft.Health.Fhir.Tests.E2E/SmartProxy/SmartProxyTests.cs
+++ b/test/Microsoft.Health.Fhir.Tests.E2E/SmartProxy/SmartProxyTests.cs
@@ -151,7 +151,7 @@ namespace Microsoft.Health.Fhir.Tests.E2E.SmartProxy
 
         private static IWebElement WaitForAndReturnElement(IWebDriver driver, string elementName, TimeSpan timeout, bool findElementById)
         {
-            // We poll every 100ms to check whether the requested element is available or not.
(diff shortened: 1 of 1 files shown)


test/Microsoft.Health.Fhir.Tests.E2E/SmartProxy/SmartProxyTests.cs
```diff
@@ -111,6 +111,10 @@ void Advance()
                 driver.FindElementByName("loginfmt").SendKeys(testUserName);
                 Advance();
 
+                // We need to add some delay before we enter the password to avoid
+                // clicking the button before the whole text is entered.
+                Thread.Sleep(TimeSpan.FromMilliseconds(1000));
```
brandonpollett: It might be a better approach to wait for the `passwd` to be available before we continue. I think you can use something like 

```suggestion
              WebDriverWait wait = new WebDriverWait(driver, 5);
              wait.until(ExpectedConditions.visibilityOfElementLocated(By.Name("passwd")));
```

namadabu: Seems like the "Commit suggestion" option automatically resolves the comment.

PR #419, opened 2019-04-11
Create export job record and support querying of status

namadabu:
## Description
This change creates a new export job record and stores the metadata in CosmosDb. We will return a url that the client can poll to get status of the export job. This change also adds support for handling these requests.

## Related issues
Addresses [issue [AB#68718](https://microsofthealth.visualstudio.com/web/wi.aspx?pcguid=b2ea084d-b70b-4689-9aad-45fb2945bbc2&id=68718) [AB#68717](https://microsofthealth.visualstudio.com/web/wi.aspx?pcguid=b2ea084d-b70b-4689-9aad-45fb2945bbc2&id=68717)].

## Testing
UTs, E2E tests, manual testing using Postman


src/Microsoft.Health.Fhir.Core.UnitTests/Features/Export/GetExportRequestHandlerTests.cs
```diff
@@ -0,0 +1,83 @@
...
+            jobRecord.JobStatus = OperationStatus.Completed;
+
+            _dataStore.GetExportJobAsync(Arg.Any<string>())
+                .Returns(x => jobRecord);
+
+            var result = await _mediator.GetExportStatusAsync(new Uri(CreateRequestUrl), jobRecord.Id);
+
+            Assert.True(result.JobExists);
+            Assert.Equal(HttpStatusCode.OK, result.StatusCode);
+            Assert.NotNull(result.JobResult);
```
namadabu: Check for required data in JobResult (like output, transactionTime, etc)

src/Microsoft.Health.Fhir.Web/appsettings.json
```diff
@@ -4,7 +4,7 @@
       "UseStrictConformance": true
     },
     "Security": {
-      "Enabled": true,
+      "Enabled": false,
```
namadabu: Revert this change

src/Microsoft.Health.Fhir.Web/appsettings.json
```diff
@@ -34,7 +34,7 @@
     },
     "Operations": {
       "Export": {
-        "Enabled": false
+        "Enabled": true
```
namadabu: Revert this change

src/Microsoft.Health.Fhir.Api/Controllers/ExportController.cs
```diff
@@ -94,6 +128,36 @@ public IActionResult ExportResourceTypeById(string type, string id)
...
+            var getExportResult = await _mediator.GetExportStatusAsync(_fhirRequestContextAccessor.FhirRequestContext.Uri, id);
+
+            if (!getExportResult.JobExists)
+            {
+                throw new JobNotFoundException(string.Format(Resources.JobNotFoundException, id));
+            }
+
+            // If the job is complete, we need to return 200 along the completed data to the client.
+            // Else we need to return 202.
+            OperationResult operationResult;
```
namadabu: The main idea behind using `OperationResult` instead of `FhirResult` (in ExportController) is that we will never be returning a fhir `Resource` as part of the response. So we want the option to send other types of responses. Whenever we throw exceptions here we depend on OperationOutcomeExceptionFilterAttribute to handle it approrpriately. That always uses a `FhirResult` when creating a response. Practically it does not make any difference since we are just returning an error (OperationOutcome) to the client and it shouldn't matter if we use `FhirResult` or `OperationResult` for that. But it does mean `ExportController` will return different types of action results.

src/Microsoft.Health.Fhir.Api/Features/ActionResults/OperationResult.cs
```diff
@@ -0,0 +1,80 @@
...
+        }
+
+        public OperationResult(ExportJobResult jobResult)
+        {
+            EnsureArg.IsNotNull(jobResult, nameof(jobResult));
+
+            JobResult = jobResult;
+        }
+
+        public HttpStatusCode? StatusCode { get; set; }
```
jackliums: Would we ever have situation where the status code won't be set?

src/Microsoft.Health.Fhir.Api/Controllers/ExportController.cs
jackliums: Can we follow the existing pattern and extract these to OperationResult.Accepted() and OperationResult.Create?

src/Microsoft.Health.Fhir.Api/Features/ActionResults/OperationResult.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        public OperationOutcome Error { get; }
+
+        public ExportJobResult JobResult { get; }
+
+        internal IHeaderDictionary Headers { get; set; } = new HeaderDictionary();
```
jackliums: this does not need to be settable since you could just call Headers.Append

src/Microsoft.Health.Fhir.Api/Features/ActionResults/OperationResult.cs
```diff
@@ -0,0 +1,80 @@
...
+        public OperationResult(ExportJobResult jobResult)
+        {
+            EnsureArg.IsNotNull(jobResult, nameof(jobResult));
+
+            JobResult = jobResult;
+        }
+
+        public HttpStatusCode? StatusCode { get; set; }
+
+        public OperationOutcome Error { get; }
```
jackliums: Could this just be passed in as a type parameter? e.g., OperationResult<OperationOutcome> or OperationResult<ExportJobResult>? If it's not being used in an array, it might be cleaner.

src/Microsoft.Health.Fhir.Api/Features/Filters/OperationOutcomeExceptionFilterAttribute.cs
```diff
@@ -66,6 +66,7 @@ public override void OnActionExecuted(ActionExecutedContext context)
 
                         break;
                     case ResourceNotFoundException _:
+                    case JobNotFoundException _:
```
jackliums: In case of JobNotFoundException, shouldn't it return OperationResult instead of FhirResult?

src/Microsoft.Health.Fhir.Api/Features/Headers/OperationHeaders.cs
```diff
@@ -0,0 +1,35 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using EnsureThat;
+using Microsoft.Health.Fhir.Api.Features.ActionResults;
+using Microsoft.Health.Fhir.Core.Features.Routing;
+using Microsoft.Net.Http.Headers;
+
+namespace Microsoft.Health.Fhir.Api.Features.Headers
+{
+    public static class OperationHeaders
```
jackliums: Can you renamed this to be named OperationResultExtensions and rename FhirHeaders to be FhirResultExtensions too?

jackliums: Can you renamed this to be named OperationResultExtensions and rename FhirHeaders to be FhirResultExtensions too?

src/Microsoft.Health.Fhir.Api.UnitTests/Controllers/ExportControllerTests.cs
```diff
@@ -25,17 +27,11 @@ public ExportControllerTests()
         }
 
         [Fact]
-        public void GivenAnExportRequest_WhenDisabled_ThenRequestNotValidExceptionShouldBeThrown()
+        public async void GivenAnExportRequest_WhenDisabled_ThenRequestNotValidExceptionShouldBeThrown()
```
jackliums: should this be `async Task`?

src/Microsoft.Health.Fhir.Api.UnitTests/Controllers/ExportControllerTests.cs
```diff
@@ -73,7 +73,9 @@ private ExportController GetController(ExportConfiguration exportConfig)
             optionsOperationConfiguration.Value.Returns(operationConfig);
 
             return new ExportController(
-                Substitute.For<IFhirRequestContextAccessor>(),
+                mediator ?? Substitute.For<IMediator>(),
```
jackliums: Since the ExportController makes sure these are not null, having null as default value doesn't make sense. I would extract these into member variable and just initialized them as mock in the constructor. If the test needs to mock certain behavior, it can does that or otherwise leave the default as is.

src/Microsoft.Health.Fhir.Api/Features/Filters/OperationOutcomeExceptionFilterAttribute.cs
jackliums: Oh I see what it's doing. It's using the existing code path of FhirResult to serialize the result into OperationOutcome. Might still need some thinking around this one since in one place we are separating the job stuff into OperationResult but here it's tight back to FhirResult.

src/Microsoft.Health.Fhir.Api/Features/Headers/OperationHeaders.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        // Generates the url to be included in the response based on the operation and sets the content location header.
+        public static OperationResult SetContentLocationHeader(this OperationResult operationResult, IUrlResolver urlResolver, string operationName, string id)
```
jackliums: nit: null check

src/Microsoft.Health.Fhir.Api.UnitTests/Features/Headers/OperationHeadersTests.cs
```diff
@@ -0,0 +1,43 @@
...
+using Microsoft.Net.Http.Headers;
+using NSubstitute;
+using Xunit;
+
+namespace Microsoft.Health.Fhir.Api.UnitTests.Features.Headers
+{
+    public class OperationHeadersTests
+    {
+        [Fact]
+        public void WhenSettingAContentLocationHeader_ThenOperationResultHasAContentLocationHeader()
```
jackliums: All the new tests should follow Given_When_Then naming convention.

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Export/CreateExportRequestHandlerTests.cs
```diff
@@ -0,0 +1,58 @@
...
+
+            ServiceProvider provider = collection.BuildServiceProvider();
+            _mediator = new Mediator(type => provider.GetService(type));
+        }
+
+        [Fact]
+        public async void GivenAFhirMediator_WhenSavingAnExportJobSucceeds_ThenResponseShouldBeSuccess()
+        {
+            _dataStore.UpsertExportJobAsync(Arg.Any<ExportJobRecord>())
+                .Returns(x => HttpStatusCode.Created);
```
jackliums: If the mock simply needs to return a value, you can simplify with `.Returns(HttpStatusCode.Created)`.

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Export/GetExportRequestHandlerTests.cs
```diff
@@ -0,0 +1,83 @@
...
+
+            ServiceProvider provider = collection.BuildServiceProvider();
+            _mediator = new Mediator(type => provider.GetService(type));
+        }
+
+        [Fact]
+        public async void GivenAFhirMediator_WhenGettingANonExistingExportJob_ThenHttpResponseShouldBeNotFound()
+        {
+            ExportJobRecord jobRecord = null;
+            _dataStore.GetExportJobAsync(Arg.Any<string>())
```
jackliums: I would update this to `_dataStore.GetExportJobAsync(CreateRequestUrl).Returns(jobRecord);` since you know the incoming value (CreateRequestUrl).

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/CreateExportRequestHandler.cs
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        public async Task<CreateExportResponse> Handle(CreateExportRequest request, CancellationToken cancellationToken)
+        {
+            EnsureArg.IsNotNull(request, nameof(request));
+
+            // TODO: Later we will add some logic here that will check whether a duplicate job already exists
+            // and handle it accordingly. For now we just assume all export jobs are unique and create a new one.
+
+            var jobRecord = new ExportJobRecord(request, 1);
```
jackliums: i don't think the schema version should be passed in.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobRecord.cs
```diff
@@ -0,0 +1,88 @@
...
+        public string Id { get; private set; }
+
+        [JsonProperty(JobRecordProperties.JobHash)]
+        public string JobHash { get; private set; }
+
+        [JsonProperty(JobRecordProperties.JobQueuedTime)]
+        public DateTimeOffset QueuedTime { get; private set; }
+
+        [JsonProperty(JobRecordProperties.PartitonKey)]
+        public string PartitionKey { get; private set; } = OperationsConstants.ExportJobPartitionKey;
```
jackliums: don't need the setter since this is a read-only property.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobResult.cs
```diff
@@ -0,0 +1,54 @@
...
+            Errors = errors;
+        }
+
+        [JsonConstructor]
+        public ExportJobResult()
+        {
+        }
+
+        [JsonProperty("transactionTime")]
+        public Instant TransactionTime { get; }
```
jackliums: I think you need private setter. Otherwise, the deserialization won't be able to set the value correctly.

jackliums: Use the DateTimeOffset instead of Instant here.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportFileInfo.cs
```diff
@@ -0,0 +1,52 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using EnsureThat;
+using Microsoft.Health.Fhir.Core.Features.Operations;
+using Newtonsoft.Json;
+
+namespace Microsoft.Health.Fhir.Core.Features.Export
+{
+    public class ExportFileInfo
```
jackliums: Can you make a models folder and move all of the data models into there?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/GetExportRequestHandler.cs
```diff
@@ -0,0 +1,63 @@
...
+            }
+
+            // We have an existing job. Let us determine the response based on the status of the
+            // export operation.
+            if (result.JobStatus == OperationStatus.Completed)
+            {
+                var jobResult = new ExportJobResult(
+                    new Instant(result.QueuedTime),
+                    result.Request.RequestUri,
+                    false /* requiresAccessToken */,
```
jackliums: nit: use named parameter instead `requiresAccessToken: false`.

src/Microsoft.Health.Fhir.Core/Features/Operations/OperationStatus.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace Microsoft.Health.Fhir.Core.Features.Operations
+{
+    public enum OperationStatus
+    {
+        Cancelled,
+        Completed,
+        Failed,
+        Queued,
+        Running,
+        Unknown,
```
jackliums: Unknown should be the first entry since that would be the default value.

src/Microsoft.Health.Fhir.Core/Messages/Export/CreateExportRequest.cs
```diff
@@ -0,0 +1,31 @@
...
+            RequestUri = requestUri;
+        }
+
+        [JsonConstructor]
+        protected CreateExportRequest()
+        {
+        }
+
+        [JsonProperty(JobRecordProperties.RequestUri)]
+        public Uri RequestUri { get; private set; }
```
jackliums: nit: this can be simplified as `public Uri RequestUri { get; }`.

src/Microsoft.Health.Fhir.Api.UnitTests/Controllers/ExportControllerTests.cs
namadabu: Since I am just asserting that an exception is being thrown and not returning anything, I went with `async void` instead of `async Task`.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobResult.cs
namadabu: I didn't add the private setters here because we will not be de-serializing it on the server side. This will be the final result that we will return to the clients. I missed removing the `[JsonConstructor]` ctor and will remove that.

Regarding the TransactionTime, property the spec - https://github.com/smart-on-fhir/fhir-bulk-data-docs/blob/master/export.md - mentions we have to use a FHIR `Instant` type. Since I was not sure whether I could use `DateTimeOffset`, I went with `Instant`. Will make the change.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobRecord.cs
jackliums: Also, this PartitionKey is Cosmos DB specific need and therefore should not be in the Core. You will need to create a wrapper class just like ResourceWrapper and wrap this class with that.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/FhirDataStore.cs
```diff
@@ -211,6 +213,49 @@ public async Task HardDeleteAsync(ResourceKey key, CancellationToken cancellatio
...
+                disableAutomaticIdGeneration: true,
+                cancellationToken: cancellationToken);
+
+                return result.StatusCode;
+            }
+            catch (DocumentClientException e)
+            {
+                if (e.Error?.Message?.Contains(GetValue(HttpStatusCode.RequestEntityTooLarge), StringComparison.Ordinal) == true)
+                {
+                    throw new ServiceUnavailableException();
```
jackliums: We have `RequestRateExceededException` now. Can we replace this and the one above with the new exception?

src/Microsoft.Health.Fhir.Core/Messages/Export/CreateExportRequest.cs
namadabu: I am storing the whole `CreateExportRequest` object as part of `ExportJobRecord` and hence need the private setter so that deserialization works. But now that I think about it more, I guess I can just store the `RequestUri` instead of the entire request.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobResult.cs
namadabu: Changed `TransactionTime` type to `DateTimeOffset` from `Instant`

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/FhirDataStore.cs
```diff
@@ -211,6 +214,49 @@ public async Task HardDeleteAsync(ResourceKey key, CancellationToken cancellatio
             }
         }
 
+        public async Task<HttpStatusCode> UpsertExportJobAsync(ExportJobRecord jobRecord, CancellationToken cancellationToken = default)
```
jackliums: nit: Remove the default value from cancellationToken. We should do a cleanup when we have chance.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/FhirDataStore.cs
```diff
@@ -211,6 +214,49 @@ public async Task HardDeleteAsync(ResourceKey key, CancellationToken cancellatio
...
+        {
+            EnsureArg.IsNotNull(jobRecord, nameof(jobRecord));
+
+            try
+            {
+                var result = await _documentClient.UpsertDocumentAsync(
+                _collectionUri,
```
jackliums: misaligned

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/Models/ExportJobRecord.cs
```diff
@@ -0,0 +1,86 @@
...
+        public List<ExportFileInfo> Errors { get; private set; } = new List<ExportFileInfo>();
+
+        [JsonProperty(JobRecordProperties.JobStatus)]
+        public OperationStatus JobStatus { get; set; }
+
+        [JsonProperty(JobRecordProperties.LastModified)]
+        public DateTimeOffset LastModifiedTime { get; set; }
+
+        [JsonProperty(JobRecordProperties.JobStartTime)]
+        public DateTimeOffset? JobStartTime { get; set; }
```
jackliums: nit: I think we can remove the word 'Job' from all of these properties.

src/Microsoft.Health.Fhir.Api/Features/ActionResults/OperationResult.cs
namadabu: I guess not. Changed it to be non-nullable

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobRecord.cs
namadabu: Created a new `CosmosExportJobRecordWrapper` class for this.

src/Microsoft.Health.Fhir.Core/Features/Persistence/IDataStore.cs
```diff
@@ -20,5 +22,9 @@ public interface IDataStore
         Task<ResourceWrapper> GetAsync(ResourceKey key, CancellationToken cancellationToken = default(CancellationToken));
 
         Task HardDeleteAsync(ResourceKey key, CancellationToken cancellationToken = default(CancellationToken));
+
+        Task<HttpStatusCode> UpsertExportJobAsync(ExportJobRecord jobRecord, CancellationToken cancellationToken = default);
```
namadabu: Remove default value for cancellationToken (here and below)

src/Microsoft.Health.Fhir.Api/Features/ActionResults/OperationOutcomeActionResult.cs
```diff
@@ -0,0 +1,54 @@
...
+using Microsoft.Extensions.Primitives;
+using Task = System.Threading.Tasks.Task;
+
+namespace Microsoft.Health.Fhir.Api.Features.ActionResults
+{
+    /// <summary>
+    /// This action result is specifically used when we want to return an error
+    /// to the client with the appropriate OperationOutcome.
+    /// </summary>
+    public class OperationOutcomeActionResult : ActionResult
```
namadabu: Rename this to `OperationOutcomeResult` to be in-line with naming convention for other ActionResults. Or would it be better to rename `FhirResult` and `OperationResult` to `FhirActionResult` and `OperationActionResult`?

src/Microsoft.Health.Fhir.Api.UnitTests/Features/Headers/OperationResultExtensionTests.cs
```diff
@@ -0,0 +1,43 @@
...
+using Microsoft.Health.Fhir.Api.Features.Headers;
+using Microsoft.Health.Fhir.Core.Features.Operations;
+using Microsoft.Health.Fhir.Core.Features.Routing;
+using Microsoft.Net.Http.Headers;
+using NSubstitute;
+using Xunit;
+
+namespace Microsoft.Health.Fhir.Api.UnitTests.Features.Headers
+{
+    public class OperationResultExtensionTests
```
jackliums: nit: missing s. OperationResultExtensionsTests.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/CreateExportRequestHandler.cs
```diff
@@ -0,0 +1,50 @@
...
+        public async Task<CreateExportResponse> Handle(CreateExportRequest request, CancellationToken cancellationToken)
+        {
+            EnsureArg.IsNotNull(request, nameof(request));
+
+            // TODO: Later we will add some logic here that will check whether a duplicate job already exists
+            // and handle it accordingly. For now we just assume all export jobs are unique and create a new one.
+
+            var jobRecord = new ExportJobRecord(request.RequestUri);
+
+            var responseCode = await _dataStore.UpsertExportJobAsync(jobRecord, cancellationToken);
```
jackliums: I think `UpsertExportJobAsync` should just throw exception if it fails to create the export job.

src/Microsoft.Health.Fhir.Core/Features/Operations/OperationsConstants.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace Microsoft.Health.Fhir.Core.Features.Operations
+{
+    public static class OperationsConstants
+    {
+        public const string Operations = "_operations";
+
+        public const string Export = "export";
+
+        public const string ExportJobPartitionKey = "ExportJob";
```
jackliums: nit: this should not be in core since it's Cosmos DB only concept.

src/Microsoft.Health.Fhir.Core/Features/Operations/JobRecordProperties.cs
```diff
@@ -0,0 +1,63 @@
...
+namespace Microsoft.Health.Fhir.Core.Features.Operations
+{
+    /// <summary>
+    /// Class for keeping track of the different json properties that will be stored
+    /// for each job record. Some of these will be common across different operations
+    /// and others might be specific.
+    /// </summary>
+    public static class JobRecordProperties
+    {
+        public const string PartitonKey = "partitionKey";
```
jackliums: is "J" supposed to be capitalized?

src/Microsoft.Health.Fhir.Core/Exceptions/JobNotFoundException.cs
```diff
@@ -0,0 +1,26 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using EnsureThat;
+using Hl7.Fhir.Model;
+
+namespace Microsoft.Health.Fhir.Core.Exceptions
+{
+    public class JobNotFoundException : FhirException
```
jackliums: I think these exceptions could go into Operations folder since it's operations specific.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/FhirDataStore.cs
```diff
@@ -211,6 +214,49 @@ public async Task HardDeleteAsync(ResourceKey key, CancellationToken cancellatio
...
+                }
+
+                throw;
+            }
+        }
+
+        public async Task<ExportJobRecord> GetExportJobAsync(string jobId, CancellationToken cancellationToken = default)
+        {
+            EnsureArg.IsNotNullOrEmpty(jobId);
+
```
jackliums: HttpStatus code is a little bit weird here. When we switch to SQL, the calls won't be over HTTP and it will be odd. It should return ExportJobRecord and maybe a wrapper object representing the outcome of the operation.

jackliums: We already have `UpsertOutcome` enum.

jackliums: Actually, maybe split the insert and update? For insert, we will need to check to make sure hash matches and all that stuff but that's not necessary for update.

src/Microsoft.Health.Fhir.Api/Features/Filters/OperationOutcomeExceptionFilterAttribute.cs
namadabu: Added a new ActionResult that will be used for handling all these failure cases. Now ExportController will return a OperationResult, FhirController a FhirResult and the OperationOutcomeFilterAttribute will return a OperationOutcomeResult

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
```diff
@@ -211,6 +214,96 @@ public async Task HardDeleteAsync(ResourceKey key, CancellationToken cancellatio
...
 
+        public async Task<ExportJobOutcome> CreateExportJobAsync(ExportJobRecord jobRecord, CancellationToken cancellationToken)
+        {
+            EnsureArg.IsNotNull(jobRecord, nameof(jobRecord));
+
+            var cosmosExportJob = new CosmosExportJobRecordWrapper(jobRecord);
+
+            try
+            {
+                var result = await _documentClient.UpsertDocumentAsync(
```
namadabu: Fixing this to CreateDocumentAsync

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/FhirDataStore.cs
namadabu: Removed the UpsertExportJobAsync and added CreateExportJobAsync and ReplaceExportJobAsync

src/Microsoft.Health.Fhir.Api/Controllers/ExportController.cs
```diff
@@ -59,9 +75,24 @@ public class ExportController : Controller
...
+        public async Task<IActionResult> Export()
         {
-            return CheckIfExportIsEnabledAndRespond();
+            if (!_exportConfig.Enabled)
+            {
+                throw new RequestNotValidException(string.Format(Resources.UnsupportedOperation, OperationsConstants.Export));
+            }
+
+            CreateExportResponse response = await _mediator.ExportAsync(_fhirRequestContextAccessor.FhirRequestContext.Uri);
+
```
jackliums: This logic should go into core. You will want the logic in controller to be as thin as possible. (If someone wants to take our code and host it using another technology, they should be able to do it with minimum code change.)

src/Microsoft.Health.Fhir.Api/Controllers/ExportController.cs
```diff
@@ -94,6 +125,34 @@ public IActionResult ExportResourceTypeById(string type, string id)
...
 
+        [HttpGet]
+        [Route(KnownRoutes.ExportStatusById, Name = RouteNames.GetExportStatusById)]
+        [AuditEventType(AuditEventSubType.Export)]
+        public async Task<IActionResult> GetExportStatusById(string id)
+        {
+            var getExportResult = await _mediator.GetExportStatusAsync(_fhirRequestContextAccessor.FhirRequestContext.Uri, id);
+
+            if (!getExportResult.JobExists)
+            {
```
jackliums: same here. I think this could go into the core.

src/Microsoft.Health.Fhir.Api/Features/ActionResults/OperationResult.cs
```diff
@@ -0,0 +1,85 @@
...
+        {
+            EnsureArg.IsNotNull(jobResult, nameof(jobResult));
+
+            JobResult = jobResult;
+            StatusCode = statusCode;
+        }
+
+        public HttpStatusCode StatusCode { get; set; }
+
+        public ExportJobResult JobResult { get; }
```
jackliums: The class name seems to indicate this is reusable for operations but the export job result is hard-coded here.

src/Microsoft.Health.Fhir.Api/Features/ActionResults/OperationResult.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// </summary>
+        /// <param name="jobResult">The job result that must be part of the ActionResult.</param>
+        public static OperationResult Ok(ExportJobResult jobResult)
+        {
+            EnsureArg.IsNotNull(jobResult);
+
+            return new OperationResult(jobResult, HttpStatusCode.OK);
+        }
+
+        public override Task ExecuteResultAsync(ActionContext context)
```
jackliums: Some of these logic is repeated here, OperaitonOutcomeResult, and FhirResult. Maybe consider creating a base class?

src/Microsoft.Health.Fhir.Api/Features/ActionResults/OperationResult.cs
namadabu: Changed it to `ExportResult`

src/Microsoft.Health.Fhir.Api/Features/ActionResults/OperationResult.cs
namadabu: Created a `BaseActionResult` class

src/Microsoft.Health.Fhir.Api/Features/ActionResults/BaseActionResult.cs
```diff
@@ -0,0 +1,62 @@
...
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.Health.Fhir.Api.Features.ActionResults
+{
+    public abstract class BaseActionResult<T> : ActionResult
+    {
+        /// <summary>
+        /// Gets the payload associated with this action result.
+        /// </summary>
+        public T Payload { get; protected set; }
```
jackliums: nit: maybe 'Result'? Also, you should make this getter only and then have a `protected BaseActionResult(T result)` as opposed to have a protected setter.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
```diff
@@ -211,6 +215,115 @@ public async Task HardDeleteAsync(ResourceKey key, CancellationToken cancellatio
...
+        {
+            EnsureArg.IsNotNull(jobRecord, nameof(jobRecord));
+
+            var cosmosExportJob = new CosmosExportJobRecordWrapper(jobRecord);
+
+            // Create access condition so that record is replaced only if eTag matches.
+            var condition = new AccessCondition()
+            {
+                Type = AccessConditionType.IfMatch,
+                Condition = eTag.ToString(),
```
jackliums: nit: eTag could be null so AccessCondition is only needed if eTag is provided.

PR #423, opened 2019-04-16
Move csproj boilerplate to Directory.Build.props and create skeletal SqlServer.csproj

johnstairs:
## Description
- Moving common properties to the existing Directory.build.props file
- Introduce Microsoft.Health.Fhir.SqlServer.csproj

### DIFF
--- a/src/Microsoft.Health.ControlPlane.Core.UnitTests/Microsoft.Health.ControlPlane.Core.UnitTests.csproj
+++ b/src/Microsoft.Health.ControlPlane.Core.UnitTests/Microsoft.Health.ControlPlane.Core.UnitTests.csproj
@@ -1,5 +1,9 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.2</TargetFramework>
+  </PropertyGroup>
+  
   <ItemGroup>
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.9.0" />
     <PackageReference Include="NSubstitute" Version="4.0.0" />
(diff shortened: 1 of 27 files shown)


src/Microsoft.Health.ControlPlane.Core.UnitTests/Microsoft.Health.ControlPlane.Core.UnitTests.csproj
```diff
@@ -1,21 +1,10 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
-  <PropertyGroup>
-    <TargetFramework>netcoreapp2.2</TargetFramework>
-    <IsPackable>false</IsPackable>
```
brandonpollett: Do we want to keep this?

src/Microsoft.Health.Extensions.DependencyInjection/Microsoft.Health.Extensions.DependencyInjection.csproj
```diff
@@ -2,19 +2,12 @@
 
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
```
brandonpollett: Does this conflict with having netcoreapp2.2 in the directory props?

johnstairs: It overrides it

src/Microsoft.Health.ControlPlane.Core.UnitTests/Microsoft.Health.ControlPlane.Core.UnitTests.csproj
johnstairs: <IsPackable>false</IsPackable> is set in Directory.Build.props because the filename contains "Test"

src/Microsoft.Health.Extensions.DependencyInjection/Microsoft.Health.Extensions.DependencyInjection.csproj
johnstairs: I had to remove this from the directory props file. VS wasn't happy with it missing from the csproj files.

Issue #426, opened 2019-04-17
Token.text search index entries are sometimes redundant


johnstairs:
**Describe the bug**
When you have a 

**To Reproduce**
Given the following codeable concept in on observation:
``` json
        "code": {
          "coding": [
            {
              "system": "http://loinc.org",
              "code": "55284-4",
              "display": "Blood Pressure"
            }
          ],
          "text": "Blood Pressure"
        }
```

**Expected behavior**
Because the coding's text matches the display of the coding, we don't need a dedicated entry for the codeable concept's text

**Actual behavior**

We end up with two Token.text search index entries with the same text "Blood Pressure", one for the coding, one for the codeable concept's text.

PR #427, opened 2019-04-17
Avoid duplicate Token.text index entries

johnstairs:
## Description
Avoid creating Token search parameter entries for codeableconcept.text where one of the codings has the same display text.

Also changing SearchIndexEntry's constructor to take a SearchParameter instead of just the search parameter's name.

## Related issues
Addresses #426

### DIFF
--- a/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Search/Converters/CodeableConceptToSearchValueTypeConverterTests.cs
+++ b/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Search/Converters/CodeableConceptToSearchValueTypeConverterTests.cs
@@ -74,7 +74,7 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Search.Converters
 
         [Theory]
         [MemberData(nameof(GetMultipleCodingDataSource))]
-        public void GivenACodeableConceptWithCodings_WhenConverted_ThenOneOrMultipleTokenSearchValuesShouldBetCreated(params Token[] tokens)
+        public void GivenACodeableConceptWithCodings_WhenConverted_ThenOneOrMultipleTokenSearchValuesShouldBeCreated(params Token[] tokens)
         {
             Test(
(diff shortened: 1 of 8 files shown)


src/Microsoft.Health.Fhir.Core.UnitTests/Features/Search/Converters/CodeableConceptToSearchValueTypeConverterTests.cs
```diff
@@ -40,7 +74,7 @@ public void GivenACodeableConceptWithNullCoding_WhenConverted_ThenNoSearchValueS
 
         [Theory]
         [MemberData(nameof(GetMultipleCodingDataSource))]
-        public void GivenACodeableConceptWithCodings_WhenConverted_ThenOneOrMultipleTokenSearchValuesShouldBeCreated(params Token[] tokens)
+        public void GivenACodeableConceptWithCodings_WhenConverted_ThenOneOrMultipleTokenSearchValuesShouldBetCreated(params Token[] tokens)
```
jackliums: Extra 't'

src/Microsoft.Health.Fhir.Core/Features/Search/SearchIndexEntry.cs
```diff
@@ -16,21 +17,21 @@ public class SearchIndexEntry
...
+            SearchParameter = searchParameter;
             Value = value;
         }
 
         /// <summary>
-        /// Gets the parameter name.
+        /// Gets the search parameter
         /// </summary>
-        public string ParamName { get; }
+        public SearchParameter SearchParameter { get; }
```
jackliums: What is this change for? It seems like the change in `CodeableConceptToSearchValueTypeConverter` to avoid duplicate doesn't actually require this change and this chagne is not being used anywhere?

johnstairs: It's not required. It's just a refactoring in preparation for the SQL provider

PR #428, opened 2019-04-19
Return authorization URL response on authorize call, return BadReques…

feordin:
Forward errors from authorization URL to caller, and return BadRequest error on missing params

## Description
Updated authorize route in proxy to return the result from the redirect.  For other errors with missing params, return a BadRequest result.

## Related issues
Addresses [issue [AB#68689](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/68689)].

## Testing
Unit test created and run

### DIFF
--- a/src/Microsoft.Health.Fhir.Api.UnitTests/Controllers/AadSmartOnFhirProxyControllerTests.cs
+++ b/src/Microsoft.Health.Fhir.Api.UnitTests/Controllers/AadSmartOnFhirProxyControllerTests.cs
@@ -15,6 +15,7 @@ using Microsoft.Extensions.Logging.Abstractions;
 using Microsoft.Extensions.Options;
 using Microsoft.Health.Fhir.Api.Configs;
 using Microsoft.Health.Fhir.Api.Controllers;
+using Microsoft.Health.Fhir.Api.Features.Exceptions;
 using Microsoft.Health.Fhir.Core.Configs;
 using Microsoft.Health.Fhir.Core.Features.Routing;
 using NSubstitute;
@@ -30,6 +31,7 @@ namespace Microsoft.Health.Fhir.Api.UnitTests.Controllers
         private readonly IUrlResolver _urlResolver = Substitute.For<IUrlResolver>();
         private readonly FeatureConfiguration _featureConfiguration = new FeatureConfiguration();
         private readonly AadSmartOnFhirProxyController _controller;
+        private TestHttpMessageHandler _httpMessageHandler;
 
         public AadSmartOnFhirProxyControllerTests()
         {
@@ -49,16 +51,16 @@ namespace Microsoft.Health.Fhir.Api.UnitTests.Controllers
             using (StreamReader r = new StreamReader(Assembly.GetExecutingAssembly().
                 GetManifestResourceStream("Microsoft.Health.Fhir.Api.UnitTests.Controllers.openid-configuration.json")))
             {
-                    openIdConfiguration = r.ReadToEnd();
+                openIdConfiguration = r.ReadToEnd();
             }
 
-            var httpMessageHandler = new TestHttpMessageHandler(new HttpResponseMessage()
+            _httpMessageHandler = new TestHttpMessageHandler(new HttpResponseMessage()
             {
                 StatusCode = System.Net.HttpStatusCode.OK,
                 Content = new StringContent(openIdConfiguration),
             });
 
-            var httpClient = new HttpClient(httpMessageHandler);
+            var httpClient = new HttpClient(_httpMessageHandler);
 
             _httpClientFactory.CreateClient().Returns(httpClient);
 
@@ -77,8 +79,7 @@ namespace Microsoft.Health.Fhir.Api.UnitTests.Controllers
         [InlineData("code", "clientId", "https://testurl.com", "launch", null, null, null)]
         [InlineData("code", "clientId", "https://testurl.com", "launch", "scope", null, null)]
         [InlineData("code", "clientId", "https://testurl.com", "launch", "scope", "state", null)]
-
-        public void GivenIncompleteQueryPamas_WhenAuthorizeRequestAction_ThenRedirectResultReturned(
+        public void GivenIncompleteQueryParams_WhenAuthorizeRequestAction_ThenRedirectResultReturned(
             string responseType, string clientId, string redirectUri, string launch, string scope, string state, string aud)
         {
             Uri redirect = null;
@@ -91,9 +92,10 @@ namespace Microsoft.Health.Fhir.Api.UnitTests.Controllers
 
             var result = _controller.Authorize(responseType, clientId, redirect, launch, scope, state, aud);
 
-            Assert.IsType<RedirectResult>(result);
+            var redirectResult = result as RedirectResult;
+            Assert.NotNull(redirectResult);
 
-            var uri = new Uri(((RedirectResult)result).Url);
+            var uri = new Uri(redirectResult.Url);
             var queryParams = HttpUtility.ParseQueryString(uri.Query);
 
             Assert.Equal(responseType, queryParams["response_type"]);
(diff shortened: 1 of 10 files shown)


src/Microsoft.Health.Fhir.Api.UnitTests/Controllers/AadSmartOnFhirProxyControllerTests.cs
```diff
@@ -0,0 +1,118 @@
...
+        }
+
+        [Theory]
+        [InlineData(null, null, null, null, null, null, null)]
+        [InlineData("code", null, null, null, null, null, null)]
+        [InlineData("code", "clientId", null, null, null, null, null)]
+        [InlineData("code", "clientId", "https://testurl.com", null, null, null, null)]
+        [InlineData("code", "clientId", "https://testurl.com", "launch", null, null, null)]
+        [InlineData("code", "clientId", "https://testurl.com", "launch", "scope", null, null)]
+        [InlineData("code", "clientId", "https://testurl.com", "launch", "scope", "state", null)]
```
brandonpollett: Do we need a case to test when the `!_isAadV2 && !string.IsNullOrEmpty(aud)` if block is executed?

src/Microsoft.Health.Fhir.Api.UnitTests/Controllers/AadSmartOnFhirProxyControllerTests.cs
```diff
@@ -0,0 +1,118 @@
...
+            if (!string.IsNullOrEmpty(redirectUri))
+            {
+                redirect = new Uri(redirectUri);
+            }
+
+            _urlResolver.ResolveRouteNameUrl(Arg.Any<string>(), Arg.Any<IDictionary<string, object>>()).Returns(redirect);
+
+            var result = _controller.Authorize(responseType, clientId, redirect, launch, scope, state, aud);
+
+            Assert.IsType<RedirectResult>(result);
```
brandonpollett: Could possibly do an `as RedirectResult` and just assert isn't null. That way you don't have to do the cast below.

src/Microsoft.Health.Fhir.Api/Resources.resx
```diff
@@ -189,11 +189,14 @@
   <data name="UrlResourceIdMismatch" xml:space="preserve">
     <value>Id in the URL must match id in the resource.</value>
   </data>
+  <data name="ValueCannotBeNull" xml:space="preserve">
+    <value>Value cannot be null.  Parameter name: {0}.</value>
```
brandonpollett: Can you add a resource comment on what this is supposed to be to help with translation?

src/Microsoft.Health.Fhir.Api/Controllers/AadSmartOnFhirProxyController.cs
```diff
@@ -300,10 +337,10 @@ public AadSmartOnFhirProxyController(IOptions<SecurityConfiguration> securityCon
                 decodedCompoundCode = JObject.Parse(Base64UrlEncoder.Decode(compoundCode));
                 code = decodedCompoundCode["code"].ToString();
             }
-            catch
+            catch (Exception ex)
```
brandonpollett: Is there a particular exception we should be catching here?

src/Microsoft.Health.Fhir.Api/Controllers/AadSmartOnFhirProxyController.cs
```diff
@@ -225,18 +242,23 @@ public AadSmartOnFhirProxyController(IOptions<SecurityConfiguration> securityCon
                 newState = launchStateParameters["s"].ToString();
                 compoundCode = Base64UrlEncoder.Encode(launchParameters.ToString(Newtonsoft.Json.Formatting.None));
             }
-            catch
+            catch (Exception ex)
             {
-                _logger.LogError("Error parsing launch parameters.");
-                return BadRequest("Invalid launch context parameters");
+                _logger.LogError($"Error parsing launch parameters: {ex.Message}");
+                throw new AadSmartOnFhirProxyBadRequestException("Invalid launch context parameters", ex);
```
brandonpollett: If this gets shown to the user it needs to be in resource file.

src/Microsoft.Health.Fhir.Api/Controllers/AadSmartOnFhirProxyController.cs
feordin: There are two different exceptions, one from Newtonsoft and another from the Decode, but I think our response is the same for either.  They have no other common base class beside Exception.

src/Microsoft.Health.Fhir.Api.UnitTests/Controllers/AadSmartOnFhirProxyControllerTests.cs
```diff
@@ -0,0 +1,118 @@
...
+        [Theory]
+        [InlineData(null, null, null, null, null, null, null)]
+        [InlineData("code", null, null, null, null, null, null)]
+        [InlineData("code", "clientId", null, null, null, null, null)]
+        [InlineData("code", "clientId", "https://testurl.com", null, null, null, null)]
+        [InlineData("code", "clientId", "https://testurl.com", "launch", null, null, null)]
+        [InlineData("code", "clientId", "https://testurl.com", "launch", "scope", null, null)]
+        [InlineData("code", "clientId", "https://testurl.com", "launch", "scope", "state", null)]
+
+        public void GivenIncompleteQueryPamas_WhenAuthorizeRequestAction_ThenRedirectResultReturned(
```
hansenms: nit: should it be `GivenIncompleteQueryParams_WhenAuthorizeRequestAction_ThenRedirectResultReturned`

hansenms:
I am not convinced that we need to add `src/Microsoft.Health.Fhir.Api.UnitTests/Controllers/openid-configuration.json`, could we not just in the test provide the `https://login.microsoftonline.com/common` endpoint as authority and the controller would pull the openid-configuration as normal?

brandonpollett:
> I am not convinced that we need to add `src/Microsoft.Health.Fhir.Api.UnitTests/Controllers/openid-configuration.json`, could we not just in the test provide the `https://login.microsoftonline.com/common` endpoint as authority and the controller would pull the openid-configuration as normal?

I think that would make it more of an integration test. We may be able to just mock out the returned configuration instead of adding it.

feordin:
> > I am not convinced that we need to add `src/Microsoft.Health.Fhir.Api.UnitTests/Controllers/openid-configuration.json`, could we not just in the test provide the `https://login.microsoftonline.com/common` endpoint as authority and the controller would pull the openid-configuration as normal?
> 
> I think that would make it more of an integration test. We may be able to just mock out the returned configuration instead of adding it.

As Brandon said, I was trying to avoid having the test call out to any external resources by adding the file.  As it is now, the file is used as the mock response which the HttpClient sends back.  It was just a long enough string it seemed easier to place it in a separate file rather than embed it as a string in the C# file.

src/Microsoft.Health.Fhir.Api/Controllers/AadSmartOnFhirProxyController.cs
```diff
@@ -136,27 +134,37 @@ public AadSmartOnFhirProxyController(IOptions<SecurityConfiguration> securityCon
...
+                queryBuilder.Add("client_id", clientId);
+            }
 
-            if (!_isAadV2)
+            try
+            {
+                var callbackUrl = _urlResolver.ResolveRouteNameUrl(RouteNames.AadSmartOnFhirProxyCallback, new RouteValueDictionary { { "encodedRedirect", Base64UrlEncoder.Encode(redirectUri.ToString()) } });
+                queryBuilder.Add("redirect_uri", callbackUrl.AbsoluteUri);
+            }
+            catch
```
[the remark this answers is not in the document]
feordin: In this instance we want to continue regardless of error and allow AAD to return an error report to the user

src/Microsoft.Health.Fhir.Api.UnitTests/Controllers/AadSmartOnFhirProxyControllerTests.cs
```diff
@@ -0,0 +1,118 @@
...
+
+            _controller = new AadSmartOnFhirProxyController(
+                Options.Create(_securityConfiguration),
+                _httpClientFactory,
+                _urlResolver,
+                _logger);
+        }
+
+        [Theory]
+        [InlineData(null, null, null, null, null, null, null)]
```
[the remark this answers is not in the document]
feordin: Added some additional tests where it made sense to do so and would enable testing of certain parameters in isolation.

PR #432, opened 2019-04-23
Create xUnit extension for testing with combinations of fixture arguments

johnstairs:
## Description
Introduces an xUnit extension that allows reuse of test cases with different fixture combinations. The E2E tests have been updated to use the extension. 

Test classes can now be authored in this way:
``` csharp
[HttpIntegrationFixtureArgumentSets(DataStore.CosmosDb, Format.Json | Format.Xml)]
public class HistoryTests : IClassFixture<HttpIntegrationTestFixture<Startup>>, IDisposable
{
    public HistoryTests(HttpIntegrationTestFixture<Startup> fixture)
    {
        Client = fixture.FhirClient;
    }

    protected FhirClient Client { get; set; }

    [Fact]
    public async Task WhenGettingResourceHistory_GivenAType_TheServerShouldReturnTheAppropriateBundleSuccessfully()
    {
            // ...
    }

    // ...
}
```
The fixture class' constructor takes a `DataStore` and a `Format` parameter. xUnit will generate test cases for combinations of declared fixture constructor arguments:

image

Tests have the option declaring the attribute as well, overriding the class-level attribute.

`HttpIntegrationFixtureArgumentSetsAttribute` is a strongly-typed attribute class that derives from `FixtureArgumentSetsAttribute`

We will use this for reusing test cases for validating Cosmos|Sql, R4|Stu3, and Xml|Json.

This works using `dotnet test`, VS, and VS code. 

**For this to work in VS, you need to have either:**

- "Discover tests in real time from source files" deselected 
image
**OR**
- "Discover tests in real time from source files" and "Additionally discover tests from build assemblies after builds"
image

## Related issues
Addresses [AB#68910](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/68910)


src/Microsoft.Health.Extensions.Xunit/FixtureArgumentSetsXunitTestFrameworkDiscoverer.cs
```diff
@@ -0,0 +1,144 @@
...
+                {
+                    yield break;
+                }
+
+                var enumAsLong = Convert.ToInt64(e);
+
+                foreach (Enum value in Enum.GetValues(e.GetType()))
+                {
+                    var flagAsLong = Convert.ToInt64(value);
+                    if (IsPowerOfTwo(flagAsLong))
```
brandonpollett: Is there ever a case where the `value` won't be a power of 2 here?

src/Microsoft.Health.Extensions.Xunit/FixtureArgumentSetsXunitTestFrameworkExecutor.cs
```diff
@@ -0,0 +1,149 @@
...
+                        new ExceptionAggregator(Aggregator),
+                        CancellationTokenSource,
+                        combinedMappings)
+                    .RunAsync();
+            }
+        }
+
+        /// <summary>
+        /// <see cref="XunitTestCase"/> facts have special optimized serialization. Their deserialization needs to be able to instantiate
+        /// the synthetics classes that we created of the form Namespace.Class(Arg1, Arg2). For these, we want to create a
```
brandonpollett: nit: `synthetic`

src/Microsoft.Health.Extensions.Xunit/FixtureArgumentSetsXunitTestFrameworkExecutor.cs
```diff
@@ -0,0 +1,149 @@
...
+        /// <see cref="TestClassWithFixtureArgumentsTypeInfo"/> with Arg1 and Arg2 as the fixture arguments
+        /// </summary>
+        private class MyAssemblyInfo : IAssemblyInfo
```
brandonpollett: nit: better name?

src/Microsoft.Health.Extensions.Xunit/FixtureArgumentSetsXunitTestFrameworkDiscoverer.cs
johnstairs: Yes, it you have an enum like this
``` csharp
enum 
{
  A = 1,
  B = 2,
  All = A | B
}
```

Issue #346, opened 2019-02-23
Support bulk export


jackliums:
As a customer, I want to be able to bulk export my data.

https://github.com/smart-on-fhir/fhir-bulk-data-docs/blob/master/export.md

CaitlinV39:
@namadabu is there anything else you want to do on this task before I close it?

namadabu:
@CaitlinV39 Nothing to add to this task. We can close it.

PR #434, opened 2019-04-26
Initial export job worker.

jackliums:
## Description
Adding the initial export job worker code. The job worker is hosted as a background service and periodically checks to see if there is a new export job available.

To support allowing multiple workers, the logic to determine how many new jobs are available is done the stored procedure. It also handles resuming stale job in case the job process terminates unexpectedly and did not finish processing the job correctly,

## Related issues
Addresses #346. Work item: [AB#68716](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/68716)

## Testing
Unit tests, integration tests, and manual tests were executed.


src/Microsoft.Health.Fhir.Core/Configs/ExportJobConfiguration.cs
```diff
@@ -3,13 +3,21 @@
...
+    public class ExportJobConfiguration
     {
         /// <summary>
         /// Determines whether export is enabled or not.
         /// </summary>
         public bool Enabled { get; set; }
+
+        public ushort MaximumNumberOfConcurrentJobsAllowed { get; set; } = 1;
+
+        public TimeSpan JobHeartbeatTimeoutThreshold { get; set; } = TimeSpan.FromMinutes(10);
```
brandonpollett: We might want to make this setting something like an `int` representing seconds. I'm not sure how common a string representation of `TimeSpan` is in json.

jackliums: Originally I had these defined as `JobHeartbeatTimeoutThresholdInSeconds` and had the value being a number but decided to make it into `TimeSpan` since I felt it gives better flexibility and is more readable. 

I didn't see any downside of using `TimeSpan` but I am open to change.

src/Microsoft.Health.Fhir.Api/Modules/OperationsModule.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        public void Load(IServiceCollection services)
+        {
+            EnsureArg.IsNotNull(services, nameof(services));
+
+            services.Add<ExportJobTaskFactory>()
+                .Singleton()
+                .AsService<IExportJobTaskFactory>();
+
+            services.Add<ExportJobWorker>()
```
jackliums: @brandonpollett suggested we move this registration out here and allow the user to explicitly register it, which I think is a good idea. Will implement that.

johnstairs:
> Adding the initial export job worker code.

General comment: might want to clarify what you mean by"initial". What is out of scope, incomplete etc. to set expectations.

src/Microsoft.Health.Fhir.Core/Configs/ExportJobConfiguration.cs
johnstairs: My opinion is that a large number representing seconds is not that readable. I would vote for the TimeSpan format, even if is is harder to author.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -0,0 +1,84 @@
...
+                    await UpdateJobStatus(OperationStatus.Running, cancellationToken);
+                }
+                catch (JobConflictException)
+                {
+                    // The job is taken by another process.
+                    _logger.LogWarning("Failed to acquire the job. The job was acquired by another process.");
+                    return;
+                }
+
+                // We have acquired the job, process the export.
```
johnstairs: Dumb question: this is TODO, right?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobWorker.cs
```diff
@@ -0,0 +1,74 @@
...
+                    {
+                        IReadOnlyCollection<ExportJobOutcome> jobs = await _fhirOperationsDataStore.GetAvailableExportJobsAsync(
+                            _exportJobConfiguration.MaximumNumberOfConcurrentJobsAllowed,
+                            _exportJobConfiguration.JobHeartbeatTimeoutThreshold,
+                            cancellationToken);
+
+                        runningTasks.AddRange(jobs.Select(job => _exportJobTaskFactory.Create(job.JobRecord, job.ETag, cancellationToken)));
+                    }
+
+                    await Task.Delay(_exportJobConfiguration.JobPollingFrequency);
```
johnstairs: include the cancellation token

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/Models/ExportJobRecord.cs
```diff
@@ -26,7 +26,6 @@ public ExportJobRecord(Uri exportRequestUri)
             Status = OperationStatus.Queued;
             Id = Guid.NewGuid().ToString();
             QueuedTime = DateTimeOffset.Now;
```
johnstairs: I think we should normalize to UTC.

src/Microsoft.Health.Fhir.Core/Features/Operations/IFhirOperationsDataStore.cs
```diff
@@ -0,0 +1,25 @@
...
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Health.Fhir.Core.Features.Operations.Export.Models;
+using Microsoft.Health.Fhir.Core.Features.Persistence;
+
+namespace Microsoft.Health.Fhir.Core.Features.Operations
+{
+    public interface IFhirOperationsDataStore
```
johnstairs: Maybe `IFhirOperationDataStore`?

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/IFhirDataStoreContext.cs
```diff
@@ -0,0 +1,20 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+
+namespace Microsoft.Health.Fhir.CosmosDb.Features.Storage
+{
+    public interface IFhirDataStoreContext
```
johnstairs: What is the purpose if this vs CosmosDataStoreConfiguration?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/Models/ExportJobRecord.cs
```diff
@@ -26,7 +26,6 @@ public ExportJobRecord(Uri exportRequestUri)
             Status = OperationStatus.Queued;
```
johnstairs: I think we need an `IsSystem => false` property

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Operations/CosmosFhirOperationsDataStore.cs
```diff
@@ -0,0 +1,205 @@
...
+                    cosmosExportJob,
+                    new RequestOptions() { PartitionKey = new PartitionKey(CosmosDbExportConstants.ExportJobPartitionKey) },
+                    disableAutomaticIdGeneration: true,
+                    cancellationToken: cancellationToken);
+
+                return new ExportJobOutcome(jobRecord, WeakETag.FromVersionId(result.Resource.ETag));
+            }
+            catch (DocumentClientException dce)
+            {
+                if (dce.StatusCode == HttpStatusCode.RequestEntityTooLarge)
```
johnstairs: Looks like the wrong error code (below as well)

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/StoredProcedures/GetAvailableExportJobs/getAvailableExportJobs.js
```diff
@@ -0,0 +1,102 @@
...
+                }
+
+                let numberOfRunningJobs = resources[0];
+                
+                tryQueryAvailableJobs(numberOfRunningJobs);
+            });
+
+        if (!isQueryAccepted) {
+            // We ran out of time.
+            throw new Error(errorMessage.requestEntityTooLarge);
```
johnstairs: provide an error code number so you don't have to rely on strings.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/StoredProcedures/GetAvailableExportJobs/getAvailableExportJobs.js
```diff
@@ -0,0 +1,102 @@
...
+* This stored procedure provides the following functionality:
+* - Retrieve list of available export jobs.
+* - The script determines list of available export jobs by looking at the list of running jobs as well as maximumNumberOfConcurrentJobsAllowed.
+* 
+* @constructor
+* @param {string} maximumNumberOfConcurrentJobsAllowedInString - The maximum number of concurrent jobs allowed in string.
+* @param {string} jobHeartbeatTimeoutThresholdInSecondsInString - The number of seconds allowed before the job is considered to be stale in string.
+*/
+
+function getAvailableExportJobs(maximumNumberOfConcurrentJobsAllowedInString, jobHeartbeatTimeoutThresholdInSecondsInString) {
```
johnstairs: Can you explain why this is a stored procedure?

test/Microsoft.Health.Fhir.Tests.Integration/Features/Operations/CosmosFhirOperationsDataStoreTests.cs
```diff
@@ -0,0 +1,199 @@
...
+using Microsoft.Health.Fhir.Core.Features.Operations.Export.Models;
+using Microsoft.Health.Fhir.Core.Messages.Export;
+using Microsoft.Health.Fhir.Tests.Common.FixtureParameters;
+using Microsoft.Health.Fhir.Tests.Integration.Persistence;
+using Xunit;
+
+namespace Microsoft.Health.Fhir.Tests.Integration.Features.Operations
+{
+    [FhirStorageTestsFixtureArgumentSets(DataStore.CosmosDb)]
+    public class CosmosFhirOperationsDataStoreTests : IClassFixture<FhirStorageTestsFixture>, IAsyncLifetime
```
johnstairs: Seems like these are data provider-agnostic. Should probably drop CosmosDb from the test class name

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/StoredProcedures/GetAvailableExportJobs/getAvailableExportJobs.js
jackliums: We need a way to guarantee that maximumNumberOfConcurrentJobs is enforced regardless of how many instances of worker is running so to do that, this logic is implemented in the stored procedure.

Let's say we have 2 worker instances running but the maximumNumberOfConcurrentJobs is set to 1, we still want to make sure at any given point, we only have 1 job running actively. `CosmosDbDistributedLock` was another option but since we have all of the export job in the same partition, I used the stored procedure.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/IFhirDataStoreContext.cs
jackliums: When we split the CosmosDB out for Fhir and CP, we end up making the class needing two pieces of information, `CosmosDataStoreConfiguration` and `IOptionsMonitor<CosmosCollectionConfiguration>`, to be able to use `DocumentClient`. Since this information is needed for both `IFhirDataStore` and `IFhirOperationsDataStore`, I just encapsulated that logic into this `IFhirDataStoreContext` class.

If we merge the code back and remove the `IOptionsMonitor<CosmosCollectionConfiguration>`, the we won't need this class anymore.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
jackliums: Yes. I forgot to link back to the work item, which would have explained the scope of this PR. The work item is here: AB#68716.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/Models/ExportJobRecord.cs
jackliums: I will update it. This logic will change a little bit as well since I am planning to make a stored procedure for inserting the job record. 

The reason I am thinking of using stored procedure is because we want to check the job hash first to see if there is any existing active job with matching hash and return that instead of creating a new record.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/StoredProcedures/GetAvailableExportJobs/getAvailableExportJobs.js
johnstairs: But this sproc is just reading, so how is it preventing two instances from reading and therefore starting the same jobs simultaneously? Optimistic concurrency can solve that, but then the sproc would not be required.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/IFhirDataStoreContext.cs
johnstairs: I think we can consolidate.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/StoredProcedures/GetAvailableExportJobs/getAvailableExportJobs.js
jackliums: The requestEntityTooLarge is defined as `${ErrorCodes.RequestEntityTooLarge}: The request could not be completed.`, which has the error code in the message itself. Since the error object only takes string, that's what we are throwing so that the client can parse the error code out of the message and react accordingly (which I realized is missing so I will add that). Is that what you are referring to?

johnstairs: You can provide an error code: `throw new Error(404, "Document not found");` then you can get the 404 from the x-ms-substatus header. An [example](https://github.com/Azure/azure-cosmos-dotnet-v3/blob/dba4b60bddafe6f6c8ad579c5d8f82c28d4e1969/Microsoft.Azure.Cosmos/src/ChangeFeedProcessor/DocDBErrors/SubStatusHelpers.cs).

johnstairs: (obviously you would not use 404 here :))

src/Microsoft.Health.Fhir.Api/Modules/OperationsModule.cs
jackliums: Separated this out.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/Models/ExportJobRecord.cs
jackliums: Good catch. Added the property.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/StoredProcedures/GetAvailableExportJobs/getAvailableExportJobs.js
jackliums: The stored procedure returns the number of available jobs based on the number of job that's currently running. For example, if there are 3 jobs queued and 1 job running, and the stored procedure is called with `maximumNumberOfConcurrentJobs = 1`, the stored procedure will return 0 available jobs.

Say the there is 2 job queued and 0 job running, with `maximumNumberOfConcurrentJobs = 1`, the stored procedure will return 1 job available. Let's assume there are 2 workers who both called the stored procedure at the same time, both of them get the job, and the first it tries to do is acquire the job by updating the status and optimistic concurrency will only allow one worker to update the job so we ensure that there is only one worker running.

If we don't use stored procedure, then the worker have to get list of running/queued jobs and then figure out which job is available to be picked up. Assume we have 3 instances of workers and the `maximumNumberOfConcurrentJobs = 2`. Each worker makes the query to get list of jobs (let's say job 1, 2, and 3). Let's say both worker A and worker B try to acquire job 1 at the same time. Worker A came in first and therefore were able to update the job record and worker B gets conflict. At this point, worker B can't just acquire job 2 because worker C might have acquired job 2 already so every time a worker gets a conflict trying to acquire the job, it needs to re-run the query to see if there is any job still available or not, which puts additional reads in the partition.

Maybe I am missing something. We can have quick chat about this one when you have time?

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/StoredProcedures/GetAvailableExportJobs/getAvailableExportJobs.js
jackliums: Oh i didn't know that existed. That's better than what we have right now. I will update the code.

src/Microsoft.Health.Fhir.Core/Features/Operations/IFhirOperationsDataStore.cs
jackliums: Updated.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/IFhirDataStoreContext.cs
jackliums: Removed and reverted the code back. There is another work item for tracking merging the Cosmos DB code back.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Operations/CosmosFhirOperationsDataStore.cs
jackliums: Fixed.

test/Microsoft.Health.Fhir.Tests.Integration/Features/Operations/CosmosFhirOperationsDataStoreTests.cs
jackliums: Good call. Updated the name.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobWorker.cs
```diff
@@ -0,0 +1,78 @@
...
+                try
+                {
+                    // Remove all completed tasks.
+                    runningTasks.RemoveAll(task => task.IsCompleted);
+
+                    // Get list of available jobs.
+                    if (runningTasks.Count < _exportJobConfiguration.MaximumNumberOfConcurrentJobsAllowed)
+                    {
+                        IReadOnlyCollection<ExportJobOutcome> jobs = await _fhirOperationDataStore.GetAvailableExportJobsAsync(
+                            _exportJobConfiguration.MaximumNumberOfConcurrentJobsAllowed,
```
namadabu: We should pass in `_exportJobConfiguration.MaximumNumberOfConcurrentJobsAllowed - runningTasks.Count` here. Else we can end up adding more number of jobs than the `MaximumNumberOfConcurrentJobsAllowed` number.

namadabu: We should pass in `_exportJobConfiguration.MaximumNumberOfConcurrentJobsAllowed - runningTasks.Count` here. Else we can end up adding more number of jobs than the `MaximumNumberOfConcurrentJobsAllowed` number.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/StoredProcedures/GetAvailableExportJobs/GetAvailableExportJobs.cs
```diff
@@ -0,0 +1,39 @@
...
+using Microsoft.Health.CosmosDb.Features.Storage.StoredProcedures;
+using Microsoft.Health.Fhir.CosmosDb.Features.Storage.Operations.Export;
+
+namespace Microsoft.Health.Fhir.CosmosDb.Features.Storage.StoredProcedures.GetAvailableExportJobs
+{
+    internal class GetAvailableExportJobs : StoredProcedureBase, IFhirStoredProcedure
+    {
+        public async Task<StoredProcedureResponse<IReadOnlyCollection<CosmosExportJobRecordWrapper>>> ExecuteAsync(
+            IDocumentClient client,
+            Uri collection,
```
namadabu: nit: maybe rename to `collectionUri`?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobWorker.cs
namadabu: I just noticed that the stored procedure takes care of this by querying for the number of running jobs before returning the available jobs. Is there a reason we don't want to have that logic here  instead of in the stored procedure?

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/StoredProcedures/GetAvailableExportJobs/getAvailableExportJobs.js
johnstairs: Did you consider having the stored procedure query and reserve in one go? Kind of like a dequeue operation? (And it could optionally request up to n jobs at once).

jackliums: That makes sense. I will update the code.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobWorker.cs
jackliums: If there are multiple instances of worker running at the same time, you need something (BE in this case) to centrally control that. It would be harder (I think) to enforce global rule from distributed workers.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/StoredProcedures/GetAvailableExportJobs/getAvailableExportJobs.js
jackliums: Updated the stored procedure to throw error with status code.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/StoredProcedures/GetAvailableExportJobs/getAvailableExportJobs.js
jackliums: Updated the stored procedure to query and aquire the job at the same time.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/StoredProcedures/GetAvailableExportJobs/GetAvailableExportJobs.cs
jackliums: Fixed.

jackliums:
With the new iteration of the PR, I have:

- Updated the stored procedure to query and acquire job as suggested.
- Updated the CosmosDbFhirStorageTestsFixture to use IAsyncLifetime as well as introducing a semaphore for collection initialization. With Cosmos DB emulator, it throws error if multiple collections are being initialized at the same time.
- Removed the IExportJobTaskFactory and replaced it with the built-in IoC Factory pattern.

src/Microsoft.Health.Fhir.Api/Registration/FhirServerServiceCollectionExtensions.cs
```diff
@@ -64,6 +67,21 @@ public static class FhirServerServiceCollectionExtensions
...
+        /// Adds the export worker background service.
+        /// </summary>
+        /// <param name="fhirServerBuilder">The FHIR server builder.</param>
+        /// <returns>The builder.</returns>
+        public static IFhirServerBuilder AddExportWorker(
+            this IFhirServerBuilder fhirServerBuilder)
+        {
+            EnsureArg.IsNotNull(fhirServerBuilder, nameof(fhirServerBuilder));
+
+            fhirServerBuilder.Services.AddHostedService<ExportJobWorkerBackgroundService>();
```
namadabu: Should we check that Export is enabled before adding the background service?

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Operations/CosmosFhirOperationDataStore.cs
```diff
@@ -0,0 +1,227 @@
...
+                    cancellationToken);
+
+                return response.Response.Select(wrapper => new ExportJobOutcome(wrapper.JobRecord, WeakETag.FromVersionId(wrapper.ETag))).ToList();
+            }
+            catch (DocumentClientException dce)
+            {
+                string subStatusInString = dce.ResponseHeaders.Get(CosmosDbHeaders.SubStatus);
+
+                if (!string.IsNullOrEmpty(subStatusInString) &&
+                    int.TryParse(subStatusInString, NumberStyles.Integer, CultureInfo.InvariantCulture, out int subStatus))
```
johnstairs: I just opened PR #449 to apply this pattern across out stored procedure calls and created an extension method on DocumentClientException.

jackliums: Merged with that change.

test/Microsoft.Health.Fhir.Tests.Integration/Persistence/FhirStorageTestsFixture.cs
```diff
@@ -31,9 +44,19 @@ public FhirStorageTestsFixture(DataStore dataStore)
...
 
+        public IFhirOperationDataStore OperationDataStore => _scopedFhirOperationDataStore.Value;
+
+        public IFhirStorageTestHelper TestHelper => _scopedTestHelper.Value;
+
         public void Dispose()
         {
             _scopedStore?.Dispose();
+            _scopedFhirOperationDataStore?.Dispose();
+            _scopedTestHelper?.Dispose();
```
johnstairs: Is Dispose() still called when you implement IAsyncLifetime?

test/Microsoft.Health.Fhir.Tests.Integration/Persistence/CosmosDbFhirStorageTestsFixture.cs
```diff
@@ -64,30 +92,52 @@ public CosmosDbFhirStorageTestsFixture()
...
+            {
+                CollectionInitializationSemaphore.Release();
+            }
 
             var cosmosDocumentQueryFactory = new FhirCosmosDocumentQueryFactory(Substitute.For<IFhirRequestContextAccessor>(), NullFhirDocumentQueryLogger.Instance);
-            var fhirRequestContextAccessor = new FhirRequestContextAccessor();
 
-            Value = new CosmosFhirDataStore(
-                new NonDisposingScope(_documentClient),
+            var documentClient = new NonDisposingScope(_documentClient);
```
johnstairs: Ideally we would dispose this

src/Microsoft.Health.Fhir.Api/Registration/FhirServerServiceCollectionExtensions.cs
jackliums: Added the check in the background service.

test/Microsoft.Health.Fhir.Tests.Integration/Persistence/CosmosDbFhirStorageTestsFixture.cs
jackliums: This is disposed in Dispose method above (that was new coded added).

test/Microsoft.Health.Fhir.Tests.Integration/Persistence/FhirStorageTestsFixture.cs
jackliums: Yes Dispose will still be called even if IAsyncLifetime interface is implemented.

PR #437, opened 2019-04-29
Added permission support for export operation.

jackliums:
## Description
Added permission support for the export operation. Created a new policy specifically for export.

_Note_:
This permission model is based on the existing pattern used for controller. Once we move the code to support operations more generally, it's likely that we will have to rethink how the permission would work.

## Related issues
Addresses #346. Work item: [AB#68902](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/68902).

## Testing
Added E2E tests to validate the permission.

### DIFF
--- a/src/Microsoft.Health.Fhir.Api/Controllers/ExportController.cs
+++ b/src/Microsoft.Health.Fhir.Api/Controllers/ExportController.cs
@@ -76,6 +76,7 @@ namespace Microsoft.Health.Fhir.Api.Controllers
         [Route(KnownRoutes.Export)]
         [ValidateExportHeadersFilter]
         [AuditEventType(AuditEventSubType.Export)]
+        [Authorize(PolicyNames.ExportPolicy)]
         public async Task<IActionResult> Export()
         {
             if (!_exportConfig.Enabled)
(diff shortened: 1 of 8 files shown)


test/Microsoft.Health.Fhir.Tests.E2E/Common/FhirClient.cs
```diff
@@ -243,6 +243,22 @@ public async Task<FhirResponse<Bundle>> SearchPostAsync(string resourceType, par
...
+            var message = new HttpRequestMessage(HttpMethod.Get, "$export");
+
+            message.Headers.Add("Accept", "application/fhir+json");
+            message.Headers.Add("Prefer", "respond-async");
+
+            HttpResponseMessage response = await HttpClient.SendAsync(message);
+
+            await EnsureSuccessStatusCodeAsync(response);
+
+            IEnumerable<string> contentLocation = response.Content.Headers.GetValues("Content-Location");
```
namadabu: Maybe we can use `HeaderNames.ContentLocation` here

brandonpollett:
I believe you need to add the new policy to the `ConfigureDefaultPolicy` call so that when security is disabled it still works.

src/Microsoft.Health.Fhir.Api/Controllers/ExportController.cs
```diff
@@ -76,6 +76,7 @@ public class ExportController : Controller
         [Route(KnownRoutes.Export)]
         [ValidateExportHeadersFilter]
         [AuditEventType(AuditEventSubType.Export)]
+        [Authorize(PolicyNames.ExportPolicy)]
```
brandonpollett: I think you can just add this once at the class level.

Issue #440, opened 2019-05-01
Delete operation is not atomic


johnstairs:
The Delete operation is current not atomic. It first does a read, then writes a delete record if a non-deleted resource exists. 
This approach is subject to race conditions and should be done in a transaction.

brandonpollett:
@johnstairs - Can this be closed?

johnstairs:
@brandonpollett closing.

PR #441, opened 2019-05-01
Makes delete operation atomic

johnstairs:
## Description
The delete operation is currently done as a read and then, conditionally a write. This change does that in a transaction.

## Related issues
Addresses #440 

## Testing
Ran and updated integration and unit tests.

### DIFF
--- a/test/Microsoft.Health.Fhir.Tests.Integration/Persistence/FhirStorageTests.cs
+++ b/test/Microsoft.Health.Fhir.Tests.Integration/Persistence/FhirStorageTests.cs
@@ -323,6 +323,34 @@ namespace Microsoft.Health.Fhir.Tests.Integration.Persistence
         }
 
         [Fact]
+        public async Task WhenDeletingAResourceThatNeverExisted_ThenReadingTheResourceReturnsNotFound()
+        {
+            string id = "missingid";
+
+            var deletedResourceKey = await Mediator.DeleteResourceAsync(new ResourceKey("Observation", id), false);
+
+            Assert.Null(deletedResourceKey.ResourceKey.VersionId);
+
+            await Assert.ThrowsAsync<ResourceNotFoundException>(
+                () => Mediator.GetResourceAsync(new ResourceKey<Observation>(id)));
+        }
+
+        [Fact]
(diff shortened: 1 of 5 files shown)


test/Microsoft.Health.Fhir.Tests.Integration/Persistence/FhirStorageTests.cs
```diff
@@ -322,6 +322,34 @@ public async Task WhenDeletingAResource_ThenWeGetResourceGone()
...
+            await Assert.ThrowsAsync<ResourceNotFoundException>(
+                () => Mediator.GetResourceAsync(new ResourceKey<Observation>(id)));
+        }
+
+        [Fact]
+        public async Task WhenDeletingAResourceForASecondTime_ThenWeDoNotGetANewVersion()
+        {
+            var saveResult = await Mediator.UpsertResourceAsync(Samples.GetJsonSample("Weight"));
+
+            await Mediator.DeleteResourceAsync(new ResourceKey("Observation", saveResult.Resource.Id), false);
```
brandonpollett: nit: resourceKey could be a variable since we use the same values twice.

test/Microsoft.Health.Fhir.Tests.Integration/Persistence/FhirStorageTests.cs
```diff
@@ -322,6 +322,36 @@ public async Task WhenDeletingAResource_ThenWeGetResourceGone()
...
+            await Assert.ThrowsAsync<ResourceNotFoundException>(
+                () => Mediator.GetResourceAsync(new ResourceKey<Observation>(id)));
+        }
+
+        [Fact]
+        public async Task WhenDeletingAResourceForASecondTime_ThenWeDoNotGetANewVersion()
+        {
+            var saveResult = await Mediator.UpsertResourceAsync(Samples.GetJsonSample("Weight"));
+
+            var resourceKey = new ResourceKey("Observation", saveResult.Resource.Id);
```
brendankowitz: In tests you should also be able to use the strongly typed ResourceKey
```suggestion
            var resourceKey = new ResourceKey<Observation>(saveResult.Resource.Id);
```

Issue #447, opened 2019-05-02
GivenANonExistingResource_WhenRead_ThenAuditLogEntriesShouldBeCreated failing locally


johnstairs:
[xUnit.net 00:02:04.68]     Microsoft.Health.Fhir.Tests.E2E.Rest.Audit.AuditTests(CosmosDb, Json).GivenANonExistingResource_WhenRead_ThenAuditLogEntriesShouldBeCreated [FAIL]
Failed   Microsoft.Health.Fhir.Tests.E2E.Rest.Audit.AuditTests(CosmosDb, Json).GivenANonExistingResource_WhenRead_ThenAuditLogEntriesShouldBeCreated
Error Message:
 Assert.Collection() Failure
Collection: [AuditEntry { Action = "read", AuditAction = Executing, Claims = [...], CorrelationId = "62ae04d9-494d-4590-a9e4-869f34c05bb0", RequestUri = http://localhost/Patient/123, ... }, AuditEntry { Action = "read", AuditAction = Executed, Claims = [...], CorrelationId = "62ae04d9-494d-4590-a9e4-869f34c05bb0", RequestUri = http://localhost/Patient/123, ... }]
Error during comparison of item at index 1
Inner exception: Assert.Equal() Failure
        Expected: OperationOutcome
        Actual:   (null)
Stack Trace:
   at Microsoft.Health.Fhir.Tests.E2E.Rest.Audit.AuditTests.ValidateAuditEntry(AuditEntry auditEntry, AuditAction expectedAuditAction, String expectedAction, Nullable`1 expectedResourceType, Uri expectedUri, Nullable`1 expectedStatusCode, String expectedCorrelationId, String expectedClaimValue, String expectedClaimKey) in c:\repos\fhir-server\test\Microsoft.Health.Fhir.Tests.E2E\Rest\Audit\AuditTests.cs:line 445
   at Microsoft.Health.Fhir.Tests.E2E.Rest.Audit.AuditTests.ValidateExecutedAuditEntry(AuditEntry auditEntry, String expectedAction, Nullable`1 expectedResourceType, Uri expectedUri, Nullable`1 expectedStatusCode, String expectedCorrelationId, String expectedClaimValue, String expectedClaimKey) in c:\repos\fhir-server\test\Microsoft.Health.Fhir.Tests.E2E\Rest\Audit\AuditTests.cs:line 437
[...]

PR #450, opened 2019-05-06
Fix broken local audit test.

brandonpollett:
## Description
This updates the audit log filter to check for both FhirResults and OperationOutcomeResults to get the resource type.

## Related issues
Addresses #447.

## Testing
The failing local test no longer fails.

--------
[AB#69049](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/69049)

### DIFF
--- a/src/Microsoft.Health.Fhir.Api/Features/Audit/AuditLoggingFilterAttribute.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/Audit/AuditLoggingFilterAttribute.cs
@@ -50,24 +50,13 @@ namespace Microsoft.Health.Fhir.Api.Features.Audit
             Debug.Assert(actionDescriptor != null, "The ActionDescriptor must be ControllerActionDescriptor.");
 
             // The result can either be a FhirResult or an OperationOutcomeResult which both extend BaseActionResult.
-
-            string typeName = null;
-
-            switch (context.Result)
-            {
-                case FhirResult fhirResult:
-                    typeName = fhirResult?.Result?.TypeName;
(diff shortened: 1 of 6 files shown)


src/Microsoft.Health.Fhir.Api/Features/Audit/AuditLoggingFilterAttribute.cs
```diff
@@ -49,13 +49,25 @@ public override void OnResultExecuted(ResultExecutedContext context)
...
             Debug.Assert(actionDescriptor != null, "The ActionDescriptor must be ControllerActionDescriptor.");
 
-            var fhirResult = context.Result as FhirResult;
+            // The result can either be a FhirResult or an OperationOutcomeResult which both extend BaseActionResult.
+
+            string typeName = null;
+
+            switch (context.Result)
+            {
+                case FhirResult fhirResult:
```
jackliums: Both FhirResult and OperationOutcomeResult implement BaseActionResult<T> so maybe have BaseActionResult<T> implement an interface that returns the result type so that if we derive another type from the BaseActionResult in the future, we don't need to update the code again?

jackliums: I think I would change the LogExecuted method and rename the parameter from resourceType to responseObjectType and just have the base class implement an interface that returns the Result?.GetType().Name or something. There isn't any restriction that the logging has to be a FHIR resource.

PR #455, opened 2019-05-08
Add _before parameter to History search

feordin:
## Description
Adds support for a new parameter: _before, for history searches

## Related issues
Addresses [issue [AB#68975](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/68975)].

## Testing
Created several new E2E tests for history covering different values for _since and _before with different number of updates in history

### DIFF
--- a/src/Microsoft.Health.Fhir.Api/Controllers/FhirController.cs
+++ b/src/Microsoft.Health.Fhir.Api/Controllers/FhirController.cs
@@ -211,7 +211,8 @@ namespace Microsoft.Health.Fhir.Api.Controllers
         /// Returns the history of all resources in the system
         /// </summary>
         /// <param name="at">Instant for history to return.</param>
-        /// <param name="since">Starting time for history to return.</param>
+        /// <param name="since">Starting time for history to return (inclusive).</param>
+        /// <param name="before">Ending time for history to return (exclusive)</param>
         /// <param name="count">Number of items to return.</param>
         /// <param name="ct">Continuation token.</param>
         [HttpGet]
@@ -221,10 +222,11 @@ namespace Microsoft.Health.Fhir.Api.Controllers
         public async Task<IActionResult> SystemHistory(
             [FromQuery(Name = KnownQueryParameterNames.At)] PartialDateTime at,
             [FromQuery(Name = KnownQueryParameterNames.Since)] PartialDateTime since,
+            [FromQuery(Name = KnownQueryParameterNames.Before)] PartialDateTime before,
             [FromQuery(Name = KnownQueryParameterNames.Count)] int? count,
             string ct)
         {
-            Bundle response = await _mediator.SearchResourceHistoryAsync(since, at, count, ct, HttpContext.RequestAborted);
+            Bundle response = await _mediator.SearchResourceHistoryAsync(since, before, at, count, ct, HttpContext.RequestAborted);
 
             return FhirResult.Create(response);
         }
@@ -234,7 +236,8 @@ namespace Microsoft.Health.Fhir.Api.Controllers
         /// </summary>
         /// <param name="type">The resource type.</param>
         /// <param name="at">Instant for history to return.</param>
-        /// <param name="since">Starting time for history to return.</param>
+        /// <param name="since">Starting time for history to return (inclusive).</param>
+        /// <param name="before">Ending time for history to return (exclusive).</param>
         /// <param name="count">Number of items to return.</param>
         /// <param name="ct">Continuation token.</param>
         [HttpGet]
@@ -245,10 +248,11 @@ namespace Microsoft.Health.Fhir.Api.Controllers
             string type,
             [FromQuery(Name = KnownQueryParameterNames.At)] PartialDateTime at,
             [FromQuery(Name = KnownQueryParameterNames.Since)] PartialDateTime since,
+            [FromQuery(Name = KnownQueryParameterNames.Since)] PartialDateTime before,
             [FromQuery(Name = KnownQueryParameterNames.Count)] int? count,
             string ct)
         {
-            Bundle response = await _mediator.SearchResourceHistoryAsync(type, since, at, count, ct, HttpContext.RequestAborted);
+            Bundle response = await _mediator.SearchResourceHistoryAsync(type, since, before, at, count, ct, HttpContext.RequestAborted);
 
             return FhirResult.Create(response);
         }
@@ -259,7 +263,8 @@ namespace Microsoft.Health.Fhir.Api.Controllers
         /// <param name="type">The resource type.</param>
         /// <param name="id">The identifier.</param>
         /// <param name="at">Instant for history to return.</param>
-        /// <param name="since">Starting time for history to return.</param>
+        /// <param name="since">Starting time for history to return (inclusive).</param>
+        /// <param name="before">Ending time for hitory to return (exclusive).</param>
         /// <param name="count">Number of items to return.</param>
         /// <param name="ct">Continuation token.</param>
         [HttpGet]
@@ -271,10 +276,11 @@ namespace Microsoft.Health.Fhir.Api.Controllers
             string id,
             [FromQuery(Name = KnownQueryParameterNames.At)] PartialDateTime at,
             [FromQuery(Name = KnownQueryParameterNames.Since)] PartialDateTime since,
+            [FromQuery(Name = KnownQueryParameterNames.Since)] PartialDateTime before,
             [FromQuery(Name = KnownQueryParameterNames.Count)] int? count,
             string ct)
         {
-            Bundle response = await _mediator.SearchResourceHistoryAsync(type, id, since, at, count, ct, HttpContext.RequestAborted);
+            Bundle response = await _mediator.SearchResourceHistoryAsync(type, id, since, before, at, count, ct, HttpContext.RequestAborted);
 
             return FhirResult.Create(response);
         }
--- a/src/Microsoft.Health.Fhir.Api/Controllers/FhirController.cs
+++ b/src/Microsoft.Health.Fhir.Api/Controllers/FhirController.cs
@@ -248,7 +248,7 @@ namespace Microsoft.Health.Fhir.Api.Controllers
             string type,
             [FromQuery(Name = KnownQueryParameterNames.At)] PartialDateTime at,
             [FromQuery(Name = KnownQueryParameterNames.Since)] PartialDateTime since,
-            [FromQuery(Name = KnownQueryParameterNames.Since)] PartialDateTime before,
+            [FromQuery(Name = KnownQueryParameterNames.Before)] PartialDateTime before,
             [FromQuery(Name = KnownQueryParameterNames.Count)] int? count,
             string ct)
         {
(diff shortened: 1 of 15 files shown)


src/Microsoft.Health.Fhir.Api/Features/Routing/UrlResolver.cs
```diff
@@ -113,13 +113,13 @@ public Uri ResolveResourceUrl(Resource resource, bool includeVersion = false)
             return new Uri(uriString);
         }
 
-        public Uri ResolveRouteUrl(IEnumerable<Tuple<string, string>> unsupportedSearchParams = null, string continuationToken = null)
+        public Uri ResolveRouteUrl(IEnumerable<Tuple<string, string>> unsupportedSearchParams = null, string continuationToken = null, IDictionary<string, object> routeValues = null)
```
brendankowitz: Is this required since we already have access to `Request.Query`. I think you just need to ensure that `_before` doesn't end up in `unsupportedSearchParams`. //cc @jackliums

src/Microsoft.Health.Fhir.Core/Features/Search/SearchService.cs
```diff
@@ -72,21 +75,59 @@ public async Task<Bundle> SearchCompartmentAsync(string compartmentType, string
...
+                            KnownQueryParameterNames.Since));
+                }
+
+                if (before != null)
+                {
+                    // _at and _since cannot be both specified.
+                    throw new InvalidSearchOperationException(
+                        string.Format(
+                            CultureInfo.InvariantCulture,
+                            Core.Resources.AtAndSinceCannotBeBothSpecified,
```
jackliums: nit: should rename this string since it's not really specifically for at and since.

src/Microsoft.Health.Fhir.Core/Features/Search/SearchService.cs
```diff
@@ -27,6 +27,9 @@ public abstract class SearchService : ISearchService, IProvideCapability
         private readonly IBundleFactory _bundleFactory;
         private readonly IFhirDataStore _fhirDataStore;
 
+        // Value which is subtracted from Now when querying _history without _before specified
+        private readonly int historyCurrentTimeBuffer = -3;
```
jackliums: nit: give the variable name unit (e.g., InSeconds) or make it into TimeSpan for readability.

src/Microsoft.Health.Fhir.Core/Features/Search/SearchService.cs
```diff
@@ -105,9 +146,27 @@ public async Task<Bundle> SearchCompartmentAsync(string compartmentType, string
...
+                    queryParameters.Add(Tuple.Create(SearchParameterNames.LastUpdated, $"ge{since}"));
+
+                    // when since is specified without _before, we will assume a value for before that is
+                    // Now - <buffer>  the buffer value allows us to have confidence that ongoing
+                    // edits are captured should the _before value be used in a future _history query
+
+                    if (before == null)
+                    {
+                        addedBefore = Clock.UtcNow.AddSeconds(historyCurrentTimeBuffer);
+                        before = new PartialDateTime(addedBefore.Value);
```
jackliums: nit: more efficient if this just simply add the query parameter directly. Creating a new PartialDateTime requires parsing of an already fully populated DateTimeOffset into individual component then stich it together again using StringBuilder.

src/Microsoft.Health.Fhir.Core/Features/Search/BundleFactory.cs
```diff
@@ -118,8 +118,16 @@ public Bundle CreateHistoryBundle(IEnumerable<Tuple<string, string>> unsupported
                 }
             }
 
+            IDictionary<string, object> routeValues = new Dictionary<string, object>();
+
+            // Add the _before value to the self link if we added it to the search parameters
+            if (addedBefore != null)
+            {
+                routeValues.Add(KnownQueryParameterNames.Before, addedBefore.Value.ToString("o"));
```
jackliums: The supplied _before parameter is a PartialDateTime but this is DateTimeOffset so the if the input was /Patient/_history?_before=2019 then the self-link returned in the bundle will be /Patient/_history?_before=2019-01-01T00:00:00.00000Z (or equivalent of that). I just want to make sure that's intentional.

src/Microsoft.Health.Fhir.Api/Features/Routing/UrlResolver.cs
feordin: When a caller does not specify a value for _before in the query, we will pick a specific time to bound the query, and then add it to the selflink so that caller can know what exact timestamp was used.  This timestamp can be used for a new query to get a new batch of history records.  I am open to thoughts/suggestions on a better approach.

---
In reply to: [282159094](https://github.com/microsoft/fhir-server/pull/455#discussion_r282159094) [](ancestors = 282159094)

src/Microsoft.Health.Fhir.Core/Features/Search/BundleFactory.cs
feordin: We should only add a new _before parameter if it was not specified by the user.  If the user specified a value for _before, it should remain on the selflink, unchanged

---
In reply to: [282223550](https://github.com/microsoft/fhir-server/pull/455#discussion_r282223550) [](ancestors = 282223550)

src/Microsoft.Health.Fhir.Core/Features/Search/SearchService.cs
```diff
@@ -27,6 +27,9 @@ public abstract class SearchService : ISearchService, IProvideCapability
         private readonly IBundleFactory _bundleFactory;
         private readonly IFhirDataStore _fhirDataStore;
 
+        // Value which is subtracted from Now when querying _history without _before specified
+        private readonly int historyCurrentTimeBufferInSeconds = -3;
```
hansenms: This is a bit of a red flag to me. After reflecting on this a bit, I think we should not try to add `_before` when user does not specify. This number will depend on a number of things that may not be in out control.

src/Microsoft.Health.Fhir.Api/Features/Routing/UrlResolver.cs
feordin: I have reverted the code that added the _before param when none was specified by the caller.

src/Microsoft.Health.Fhir.Core/Features/Search/BundleFactory.cs
feordin: I have reverted the code that added the _before param when none was specified.

src/Microsoft.Health.Fhir.Core/Features/Search/SearchService.cs
feordin: I have reverted the code that added the _before param when none was specified by the caller.

src/Microsoft.Health.Fhir.Api/Controllers/FhirController.cs
```diff
@@ -221,10 +222,11 @@ public async Task<IActionResult> Read(string type, string id)
         public async Task<IActionResult> SystemHistory(
             [FromQuery(Name = KnownQueryParameterNames.At)] PartialDateTime at,
```
jackliums: Not your code, but I just realized that the `at` parasmeter is an instant type. Filed new [bug](https://github.com/microsoft/fhir-server/issues/463) for this.

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Search/SearchServiceTests.cs
```diff
@@ -157,7 +157,7 @@ public async Task GivenAMissingResourceId_WhenSearchingHistory_ThenAResourceNotF
 
             _searchService.SearchImplementation = options => new SearchResult(new ResourceWrapper[0], null);
 
-            await Assert.ThrowsAsync<ResourceNotFoundException>(() => _searchService.SearchHistoryAsync(resourceType, resourceId, null, null, null, null, CancellationToken.None));
+            await Assert.ThrowsAsync<ResourceNotFoundException>(() => _searchService.SearchHistoryAsync(resourceType, resourceId, null, null, null, null, null, CancellationToken.None));
```
jackliums: nit: can we add named parameter for readability?

test/Microsoft.Health.Fhir.Tests.E2E/Rest/HistoryTests.cs
```diff
@@ -55,6 +59,265 @@ public async Task WhenGettingSystemHistory_GivenNoType_TheServerShouldReturnTheA
...
 
+        [Fact]
+        public void WhenGettingSystemHistory_GivenAValueForSince_TheServerShouldReturnOnlyRecordsModifiedAfterSinceValue()
+        {
+            var since = GetStartTimeForHistoryTest();
+            var sinceUriString = HttpUtility.UrlEncode(since.ToString("o"));
+            Thread.Sleep(1000);
+            _createdResource.Resource.Comment = "Changed by E2E test";
+
+            var updatedResource = Client.UpdateAsync<Observation>(_createdResource).GetAwaiter().GetResult();
```
brendankowitz: Test method could be `async Task` which will remove the `GetAwaiters` and `.Results`

test/Microsoft.Health.Fhir.Tests.E2E/Rest/HistoryTests.cs
```diff
@@ -55,6 +59,265 @@ public async Task WhenGettingSystemHistory_GivenNoType_TheServerShouldReturnTheA
...
+
+            Thread.Sleep(5000); // ensure that the server has fully processed the PUT
```
brendankowitz: Not sure about needing this, should the CosmosDb sessions not handle this type of consistency? (`SessionMessageHandler`) // @johnstairs

johnstairs: We use session consistency in the E2E tests and take care of passing around the session token. You should not need this.

test/Microsoft.Health.Fhir.Tests.E2E/Rest/HistoryTests.cs
feordin: I started with the History tests as async, but that caused the multiple versions (JSON vs XML) to run in parallel and then the number of edits was no longer predictable and the tests would not run reliably.  So I removed all the async code.

test/Microsoft.Health.Fhir.Tests.E2E/Rest/HistoryTests.cs
feordin: I have removed or reduced the wait times.

test/Microsoft.Health.Fhir.Tests.E2E/Rest/HistoryTests.cs
brendankowitz: Makes sense. Wondering if this might work on the HistoryTests class? `[CollectionDefinition(DisableParallelization = true)]`
https://github.com/xunit/xunit/issues/1179

feordin: Thanks!  That was a good suggestion.  Tests updated with the DisableParallelization enabled.

Issue #400, opened 2019-03-27
Support SQL Server as a data provider


johnstairs:

justinbhopper:
@jackliums @johnstairs I'm curious if there are any guesses how well the SQL Server persistence layer might perform in comparison to the Cosmos solution?  

I'm guessing adding support for SQL Server is mostly driven from a desire to increase support/compatibility, and not because it is believed it will become the superior data storage medium for fhir-server, but I'm just speculating...

hansenms:
The SQL persistence provider is being implemented to support features such as chained searches, _include, _revinclude, transactions, etc. These things are hard to support using Cosmos. Hope this helps.

hansenms:
It is available for SQL now.

PR #456, opened 2019-05-08
Add sql configuration for web/arm template

brandonpollett:
## Description
Updates the web project to be able to run in either SQL server or CosmosDB. Additionally updates the arm template to support both deployments. Enables E2E tests for supported scenarios (CRUD) in sql server. 

To support this, the PR build and CI build will need to be updated.

## Related issues
Relates to #400 

## Testing
E2E tests were enabled for SQL Server CRUD.


samples/templates/default-azuredeploy.json
```diff
@@ -144,9 +144,13 @@
...
             "FhirServer:Security:Authentication:Audience": "[parameters('securityAuthenticationAudience')]",
-            "CosmosDb:ContinuationTokenSizeLimitInKb": "1"
+            "CosmosDb:ContinuationTokenSizeLimitInKb": "1",
+            "SqlServer:Initialize": "true",
+            "DataStore": "[if(equals(parameters('solutionType'),'FhirServerCosmosDB'), 'CosmosDb', 'SqlServer')]"
         },
-        "combinedFhirServerConfigProperties": "[union(variables('staticFhirServerConfigProperties'), parameters('additionalFhirServerConfigProperties'))]"
+        "combinedFhirServerConfigProperties": "[union(variables('staticFhirServerConfigProperties'), parameters('additionalFhirServerConfigProperties'))]",
+        "computedSqlServerReference": "[concat('Microsoft.Sql/servers/', variables('serviceName'))]",
+        "sqlAdminPass" : "[concat(uniqueString(resourceGroup().id, deployment().name, 'ac45f99e-035e-4712-941d-cafac13536ba'), 'x', '!')]"
```
johnstairs: This can be reverse-engineered. It should be passed in as a parameter I think (since we don't necessarily want to change it with each deployment).

samples/templates/default-azuredeploy.json
```diff
@@ -245,8 +254,9 @@
             }
         },
         {
+            "condition": "[equals(parameters('solutionType'), 'FhirServerCosmosDB')]",
```
jackliums: We might want to split the Cosmos DB and SQL server ARM template out to separate file and include that conditionally? Doesn't have to be in this PR but it might be easier to read if we separate them by components?

samples/templates/default-azuredeploy.json
```diff
@@ -264,6 +274,58 @@
...
+            "condition": "[equals(parameters('solutionType'), 'FhirServerSqlServer')]",
+            "name": "[variables('serviceName')]",
+            "type": "Microsoft.Sql/servers",
+            "apiVersion": "2015-05-01-preview",
+            "location": "[resourceGroup().location]",
+            "tags": {
+                "FhirServerSolution": "[parameters('solutionType')]"
+            },
+            "properties": {
+                "administratorLogin": "[uniqueString(resourceGroup().id)]",
```
jackliums: Is this setting the administrator user name? If it's generating a hash based on the resource group name, how would we know what that is since it's not being displayed or returned anywhere?

samples/templates/default-azuredeploy.json
```diff
@@ -264,6 +274,58 @@
...
+                "administratorLoginPassword": "[parameters('sqlAdminPassword')]",
+                "version": "12.0"
+            },
+            "resources": [
+                {
+                    "condition": "[equals(parameters('solutionType'), 'FhirServerSqlServer')]",
```
jackliums: Do we need these additional conditions since the parent resource already has the same condition?

samples/templates/default-azuredeploy.json
```diff
@@ -264,6 +274,58 @@
...
+                    "location": "[resourceGroup().location]",
+                    "tags": {
+                        "FhirServerSolution": "[parameters('solutionType')]"
+                    },
+                    "name": "FHIR",
+                    "properties": {
+                        "collation": "SQL_Latin1_General_CP1_CI_AS"
+                    },
+                    "sku": {
+                        "name": "Standard",
```
jackliums: nit: should these be configurable through parameters?

samples/templates/default-azuredeploy.json
```diff
@@ -297,30 +359,46 @@
...
         },
         {
+            "condition": "[equals(parameters('solutionType'), 'FhirServerCosmosDB')]",
             "type": "Microsoft.KeyVault/vaults/secrets",
             "name": "[concat(variables('serviceName'), '/CosmosDb--Host')]",
             "apiVersion": "2015-06-01",
             "properties": {
                 "contentType": "text/plain",
-                "value": "[reference(concat('Microsoft.DocumentDb/databaseAccounts/', variables('serviceName'))).documentEndpoint]"
+                "value": "[if(equals(parameters('solutionType'), 'FhirServerCosmosDB'), reference(concat('Microsoft.DocumentDb/databaseAccounts/', variables('serviceName')), '2015-04-08').documentEndpoint, '')]"
```
jackliums: Do we need this condition here since condition above already check to make sure the solution type is cosmos db?

samples/templates/default-azuredeploy.json
jackliums: I guess it's in the KeyVault so we could grab it from there.

brandonpollett: Depends on where you want to use it. It's stored in the connection string we put in keyvault and it's also displayed in the portal when you go to `Query Editor (Preview)`.

I'm not opposed to making it something known, but just thought this added one extra dimension that you had to know to get access.

src/Microsoft.Health.Fhir.Web/Startup.cs
```diff
@@ -24,9 +25,16 @@ public virtual void ConfigureServices(IServiceCollection services)
         {
             services.AddDevelopmentIdentityProvider(Configuration);
 
-            services.AddFhirServer(Configuration)
-                .AddExportWorker()
-                .AddCosmosDb(Configuration);
+            Core.Registration.IFhirServerBuilder fhirServerBuilder = services.AddFhirServer(Configuration).AddExportWorker();
+
+            if (Configuration["DataStore"].Equals(KnownDataStores.CosmosDb, StringComparison.InvariantCultureIgnoreCase))
```
jackliums: nit: save as a variable

src/Microsoft.Health.Fhir.Web/Startup.cs
```diff
@@ -24,9 +25,16 @@ public virtual void ConfigureServices(IServiceCollection services)
...
+            {
+                fhirServerBuilder.AddCosmosDb(Configuration);
+            }
+            else if (Configuration["DataStore"].Equals(KnownDataStores.SqlServer, StringComparison.InvariantCultureIgnoreCase))
+            {
+                services.AddExperimentalSqlServer();
```
jackliums: any reason this is not following the pattern above?

samples/templates/default-azuredeploy.json
brandonpollett: Yes. The way that arm appears to do lookups is that it evaluates these regardless of the condition on the resource -- so without these checks you end up with not found exceptions.

src/Microsoft.Health.Fhir.Web/appsettings.json
```diff
@@ -56,6 +56,7 @@
       "MaxWaitTimeInSeconds": 5
     }
   },
+  "DataStore": "CosmosDb",
```
jackliums: nit: should this be under FhirServer since it's setting the data store for the FHIR server?

brandonpollett: I don't think so since this is a setting only for our web project and not the actual fhir server code.

samples/templates/default-azuredeploy.json
brandonpollett: I'll have to double check.

src/Microsoft.Health.Fhir.Web/Startup.cs
```diff
@@ -24,9 +25,16 @@ public virtual void ConfigureServices(IServiceCollection services)
...
-            services.AddFhirServer(Configuration)
-                .AddExportWorker()
-                .AddCosmosDb(Configuration);
+            Core.Registration.IFhirServerBuilder fhirServerBuilder = services.AddFhirServer(Configuration).AddExportWorker();
+
+            if (Configuration["DataStore"].Equals(KnownDataStores.CosmosDb, StringComparison.InvariantCultureIgnoreCase))
+            {
+                fhirServerBuilder.AddCosmosDb(Configuration);
+            }
+            else if (Configuration["DataStore"].Equals(KnownDataStores.SqlServer, StringComparison.InvariantCultureIgnoreCase))
```
jackliums: I wonder if this process could be a little bit more automatic. E.g., some class with some attribute to indicate that data store type and the code automatically discover and register it based on that attribute?

samples/templates/default-azuredeploy.json
brandonpollett: I think long term that will probably make sense. There's going to likely be a number of more options in the SQL arm template so I'm not sure currently if it makes more sense to make two completely separate templates or use a base template with two sub templates depending on the data store selected.

samples/templates/default-azuredeploy.json
brandonpollett: I think at some point, yes. There are 5 different settings for SKU and without knowing which configuration we were going to recommend, I went for hard coding the cheapest to get tests passing.

https://docs.microsoft.com/en-us/azure/templates/microsoft.sql/2017-10-01-preview/servers/databases#Sku

src/Microsoft.Health.Fhir.Web/Startup.cs
brandonpollett: The pattern of using the `FhirServerBuilder`?

samples/templates/default-azuredeploy.json
brandonpollett: It fails when you do a cosmosdb deployment with the following exception when you don't include it:

```json
{
  "error": {
    "code": "ParentResourceNotFound",
    "message": "Can not perform requested operation on nested resource. Parent resource 'msh-fhir-pr-456' not found."
  }
}
```

samples/templates/default-azuredeploy.json
brandonpollett: Splitting out sql from cosmos arm templates.

PR #454, opened 2019-05-08
Adds codegen for table-valued parameters

johnstairs:
## Description

A intermediate PR to show updates to codegen to support table-valued parameters. First use is for adding ResourceWriteClaims (a.k.a. LastModifiedClaims). Programming model looks like this: (last parameter is new):
``` csharp
V1.UpsertResource.PopulateCommand(
    command,
    resourceTypeId: _model.GetResourceTypeId(resource.ResourceTypeName),
    resourceId: resource.ResourceId,
    eTag: weakETag == null ? null : (int?)etag,
    allowCreate: allowCreate,
    isDeleted: resource.IsDeleted,
    updatedDateTime: resource.LastModified,
    keepHistory: keepHistory,
    requestMethod: resource.Request.Method,
    rawResource: ms,
    resourceWriteClaims: GetResourceWriteClaims(resource));

// use command...

private IEnumerable<V1.ResourceWriteClaimTableTypeRow> GetResourceWriteClaims(ResourceWrapper resource)
{
    return resource.LastModifiedClaims?.Select(c => new V1.ResourceWriteClaimTableTypeRow(_model.GetClaimTypeId(c.Key), c.Value));
}
```

## Related issues
Part of  [AB#69044](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/69044)

PR #457, opened 2019-05-10
Compartment indexing and TVP generator codegen

johnstairs:
## Description
Builds on #454, adding Compartment indexing and adds some structure around how we generate rows for table-valued parameters. Each TVP has a generator type that is added to the DI container and some codegen that creates a generator that produces a struct with properties for each TVP. Programming model looks like this:

``` csharp
private readonly V1.UpsertResourceTvpGenerator<ResourceWrapper> _upsertResourceTvpGenerator;

// ...

V1.UpsertResource.PopulateCommand(
    command,
    resourceTypeId: _model.GetResourceTypeId(resource.ResourceTypeName),
    resourceId: resource.ResourceId,
    eTag: weakETag == null ? null : (int?)etag,
    allowCreate: allowCreate,
    isDeleted: resource.IsDeleted,
    updatedDateTime: resource.LastModified,
    keepHistory: keepHistory,
    requestMethod: resource.Request.Method,
    rawResource: stream,
    tableValuedParameters: _upsertResourceTvpGenerator.Generate(resource));
```

## Related issues
Part of [AB#69044](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/69044)


src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Model/ITableValuedParameterRowGenerator.cs
```diff
@@ -0,0 +1,20 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+
+namespace Microsoft.Health.Fhir.SqlServer.Features.Schema.Model
+{
+    /// <summary>
+    /// Generates a sequence of row structs for a table-valued parameter.
+    /// </summary>
+    /// <typeparam name="TInput">THe input type</typeparam>
```
cunninghamjc: Nit: "THe" -> "The"

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
```diff
@@ -227,6 +227,52 @@ CREATE CLUSTERED INDEX IXC_LastModifiedClaim on dbo.ResourceWriteClaim
...
+
+CREATE TABLE dbo.CompartmentAssignment
+(
+    ResourceSurrogateId bigint NOT NULL,
+    CompartmentTypeId tinyint NOT NULL,
+    ReferenceResourceId varchar(64) NOT NULL,
+    IsHistory bit NOT NULL,
+) WITH (DATA_COMPRESSION = PAGE)
+
+CREATE CLUSTERED INDEX IXC_CompartmentAssignment
```
cunninghamjc: Should this be unique?

johnstairs: In general, we are not doing anything at the moment to ensure that duplicates are not present in the resource and/or what the extractors yield. For example, for an observation resource, the same practitioner could appear twice in the list of performers. Or a patient could be named "John John", in which case the extracted name parameter would have "John" twice. 

Having duplicates does not bring any benefit and forces to do a DISTINCT clause when searching. 
So we probably should make sure everything is unique and filter out duplicates before adding them, but when it gets to strings, it has to be unique respecting the same collation rules as the DB column. I am deferring this to later, once we have sorted out the collation configuration. Sound ok?

cunninghamjc: Yes, sounds okay.

PR #458, opened 2019-05-10
Moves Stu3 specific functionality from Fhir.Core to Fhir.Stu3.Core

brendankowitz:
## Description
The first stage in splitting logic for supporting R4/Stu3

## Related issues
Addresses [AB#69057](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/69057)


src/Microsoft.Health.Fhir.Api.UnitTests/Features/Headers/FhirHeadersTests.cs
```diff
@@ -35,15 +37,16 @@ public FhirHeadersTests()
...
                 },
-            };
+            }.ToResourceElement();
         }
 
         [Fact]
         public void WhenSettingALocationHeader_ThenFhirResultHasALocationHeader()
         {
-            var locationUrl = new Uri($"http://localhost/{_mockResource.GetType().Name}/{_mockResource.Id}/_history/{_mockResource.Meta.VersionId}");
+            var locationUrl = new Uri($"http://localhost/{_mockResource.GetType().Name}/{_mockResource.Id}/_history/{_mockResource.VersionId}");
```
jackliums: nit: I don't think it matters since we are mocking it but for clarity purposes I think the URL is supposed to be something like `http://localhost/Observation/xxx` but it wil lbe `http://localhost/ResourceElement/xxx` after the refactor.

src/Microsoft.Health.Fhir.Api/Controllers/FhirController.cs
```diff
@@ -424,14 +424,14 @@ public async Task<IActionResult> SearchCompartmentByResourceType(string compartm
 
         private async Task<IActionResult> PerformCompartmentSearch(string compartmentType, string compartmentId, string resourceType, IReadOnlyList<Tuple<string, string>> queries)
         {
-            Bundle response = await _mediator.SearchResourceCompartmentAsync(compartmentType, compartmentId, resourceType, queries, HttpContext.RequestAborted);
+            var response = await _mediator.SearchResourceCompartmentAsync(compartmentType, compartmentId, resourceType, queries, HttpContext.RequestAborted);
```
jackliums: This should remain `Bundle` according to our coding guideline since the right-hand side does not clearly identify the object type.

src/Microsoft.Health.Fhir.Api/Features/Headers/FhirHeaders.cs
```diff
@@ -31,7 +33,7 @@ public static FhirResult SetLocationHeader(this FhirResult fhirResult, IUrlResol
 
         public static FhirResult SetETagHeader(this FhirResult fhirResult)
         {
-            var resource = fhirResult.Resource;
+            var resource = fhirResult.Resource.Instance.ToPoco<Resource>();
```
jackliums: Don't need to convert this back to poco. Can just access the versionId through `resource.VersionId`.

src/Microsoft.Health.Fhir.Api/Features/Headers/FhirHeaders.cs
```diff
@@ -52,7 +54,7 @@ public static FhirResult SetETagHeader(this FhirResult fhirResult, WeakETag weak
 
         public static FhirResult SetLastModifiedHeader(this FhirResult fhirResult)
         {
-            var resource = fhirResult.Resource;
+            var resource = fhirResult.Resource.Instance.ToPoco<Resource>();
```
jackliums: Same here. Can just access LastUpdated through `resource.lastUpdated`.

src/Microsoft.Health.Fhir.Api.UnitTests/Features/Headers/FhirHeadersTests.cs
brendankowitz: Thanks! I'll need help trying to find things like this to tidy up

test/Microsoft.Health.Fhir.Tests.E2E/Rest/VReadTests.cs
```diff
@@ -30,7 +31,7 @@ public VReadTests(HttpIntegrationTestFixture<Startup> fixture)
         [Trait(Traits.Priority, Priority.One)]
         public async Task WhenGettingAResource_GivenAnIdAndVersionId_TheServerShouldReturnTheAppropriateResourceSuccessfully()
         {
-            Observation createdResource = await Client.CreateAsync(Samples.GetDefaultObservation());
+            Observation createdResource = await Client.CreateAsync(Samples.GetDefaultObservation().ToPoco<Observation>());
```
jackliums: What's the plan for when we need to start testing STU3 and R4?

src/Microsoft.Health.Fhir.Api.UnitTests/Features/Audit/AuditLoggingFilterAttributeTests.cs
```diff
@@ -77,7 +78,7 @@ public void GivenAController_WhenExecutedAction_ThenAuditLogShouldBeLogged()
         {
             const HttpStatusCode expectedStatusCode = HttpStatusCode.Created;
 
-            var fhirResult = new FhirResult(new Patient() { Name = { new HumanName() { Text = "TestPatient" } } });
+            var fhirResult = new FhirResult(new Patient() { Name = { new HumanName() { Text = "TestPatient" } } }.ToResourceElement());
```
jackliums: Is there a way to directly deserialize into ITypedElement without having to deserialize into a resource first?

src/Microsoft.Health.Fhir.Core/Features/Context/FhirRequestContext.cs
```diff
@@ -55,7 +56,7 @@ public class FhirRequestContext : IFhirRequestContext
 
         public string CorrelationId { get; }
 
-        public Coding RequestType { get; }
+        public ResourceElement RequestType { get; }
```
jackliums: nit: I think we can have our own type instead of using ResourceElement here and just convert it to Coding when needed. I think it was coding because it was easy to use before but now we have to evaluate through FHIR path, it might be easier just to use a ValueTuple or custom struct/class.

src/Microsoft.Health.Fhir.Api/Features/ActionResults/FhirResult.cs
```diff
@@ -78,5 +83,36 @@ public static FhirResult NoContent()
                 StatusCode = HttpStatusCode.NoContent,
             };
         }
+
+        /// <inheritdoc />
```
jackliums: One of the reason for the base class is to share this logic. Looks like `Result.ToPoco()` is the only difference between the two. Can we still share most of the code and just have the derived class provide that (could extend BaseActionResult to take additional type parameter?)

src/Microsoft.Health.Fhir.Api/Features/Filters/OperationOutcomeExceptionFilterAttribute.cs
```diff
@@ -53,7 +54,7 @@ public override void OnActionExecuted(ActionExecutedContext context)
                     new OperationOutcome
```
jackliums: We probably should avoid using OperationOutcome and have our own class? Or is that part of the refactor for the moving the APIs to STU3 assembly?

src/Microsoft.Health.Fhir.Api/Features/Headers/FhirResultExtensions.cs
```diff
@@ -52,13 +54,12 @@ public static FhirResult SetETagHeader(this FhirResult fhirResult, WeakETag weak
 
         public static FhirResult SetLastModifiedHeader(this FhirResult fhirResult)
         {
-            var resource = fhirResult.Result;
-            if (resource != null)
+            ResourceElement resource = fhirResult.Result;
+
+            var lastUpdated = resource?.LastUpdated;
+            if (lastUpdated.HasValue)
```
jackliums: nit: I prefer keep it consistent with `if (lastUpdated != null)`.

src/Microsoft.Health.Fhir.Api/Modules/FhirModule.cs
```diff
@@ -57,12 +59,28 @@ public void Load(IServiceCollection services)
             services.AddSingleton(xmlParser);
             services.AddSingleton(xmlSerializer);
 
-            services.AddSingleton<IReadOnlyDictionary<ResourceFormat, Func<string, Resource>>>(x =>
+            services.AddSingleton<IReadOnlyDictionary<FhirResourceFormat, Func<string, string, DateTimeOffset, ResourceElement>>>(x =>
```
jackliums: This could be registered with `FhirResourceFormat` and `Func<string, Resource>` right? Since the only difference is that JsonSerializer vs XmlSerializer but the Parse method is the same.

Something like:
```
services.AddSingleton<IReadOnlyDictionary<FhirResourceFormat, Func<string, Resource>>>(x =>
{
    return new Dictionary<FhirResourceFormat, Func<string, Resource>>
    {
        { FhirResourceFormat.Json, jsonParser.Parse<Resource> },
        { FhirResourceFormat.Xml, xmlParser.Parse<Resource> },
    };
});
```

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Compartment/CompartmentIndexerTests.cs
```diff
@@ -30,14 +30,14 @@ public void GivenSearchIndicesWithResourceTypeAndCompartmentType_WhenExtracted_T
             var compartmentIndexer = new CompartmentIndexer(compartmentManager);
 
             HashSet<string> compParams = null;
-            compartmentManager.TryGetSearchParams(resourceType, compartmentType, out compParams)
+            compartmentManager.TryGetSearchParams(resourceType.ToString(), compartmentType.ToString(), out compParams)
```
jackliums: Should these be updated to use string instead of enum?

src/Microsoft.Health.Fhir.Api/Modules/FhirModule.cs
```diff
@@ -144,6 +162,10 @@ public void Load(IServiceCollection services)
 
             services.AddSingleton<INarrativeHtmlSanitizer, NarrativeHtmlSanitizer>();
 
+            var stu3ModelFactory = new Stu3ModelFactory();
+            services.Add(c => stu3ModelFactory).Singleton().AsSelf().AsImplementedInterfaces();
```
jackliums: nit: can we keep the parameter name consistent across (it's showing up as `c`, `x`, and `provider` in the same file).

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Resources/Upsert/UpsertResourceValidatorTests.cs
```diff
@@ -23,9 +26,10 @@ public class UpsertResourceValidatorTests
         public void GivenAResourceWithoutInvalidId_WhenValidatingUpsert_ThenInvalidShouldBeReturned(string id)
         {
             var validator = new UpsertResourceValidator(new NarrativeHtmlSanitizer(NullLogger<NarrativeHtmlSanitizer>.Instance));
-            Observation resource = Samples.GetDefaultObservation();
+            var resource = Samples.GetDefaultObservation().Instance.ToPoco<Observation>();
```
jackliums: Probably can use  `.UpdateId(id)` to avoid converting back and forth?

src/Microsoft.Health.Fhir.Core/Configs/AuthorizationConfiguration.cs
```diff
@@ -21,18 +21,16 @@ public class AuthorizationConfiguration
...
                 foreach (var validationError in role.Validate(new ValidationContext(role)))
                 {
-                    issues.Add(new OperationOutcome.IssueComponent
-                    {
-                        Severity = OperationOutcome.IssueSeverity.Fatal,
-                        Code = OperationOutcome.IssueType.Invalid,
-                        Diagnostics = validationError.ErrorMessage,
-                    });
+                    issues.Add(new OperationOutcomeIssue(
+                            "Fatal",
```
jackliums: `OperationOutcomeConstants.IssueSeverity.Fatal`

src/Microsoft.Health.Fhir.Core/Features/Compartment/CompartmentResourceHandler.cs
```diff
@@ -36,7 +35,7 @@ public async Task<CompartmentResourceResponse> Handle(CompartmentResourceRequest
         {
             EnsureArg.IsNotNull(message, nameof(message));
 
-            Bundle bundle = await _searchService.SearchCompartmentAsync(message.CompartmentType, message.CompartmentId, message.ResourceType, message.Queries, cancellationToken);
+            var bundle = await _searchService.SearchCompartmentAsync(message.CompartmentType, message.CompartmentId, message.ResourceType, message.Queries, cancellationToken);
```
jackliums: nit: should not be var

src/Microsoft.Health.Fhir.Core/Features/Persistence/ResourceWrapper.cs
```diff
@@ -25,10 +25,10 @@ public class ResourceWrapper
            : this(
                  IsNotNull(resource).Id,
```
jackliums: nit: `IsNotNull` can be replaced by `EnsureArg.IsNotNull`

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/ReferenceSearchValueParser.cs
```diff
@@ -58,7 +57,7 @@ public ReferenceSearchValue Parse(string s)
                     return new ReferenceSearchValue(
                         ReferenceKind.InternalOrExternal,
                         null,
-                        resourceType,
+                        resourceTypeInString,
```
jackliums: We should validate to make sure this is a valid resource type (i think Enum.Parse will throw exception if it's not valid).

src/Microsoft.Health.Fhir.Core/Messages/Get/GetSystemCapabilitiesResponse.cs
```diff
@@ -4,19 +4,20 @@
...
-using Hl7.Fhir.Model;
+using Hl7.Fhir.ElementModel;
+using Microsoft.Health.Fhir.Core.Models;
 
 namespace Microsoft.Health.Fhir.Core.Messages.Get
 {
     public class GetSystemCapabilitiesResponse
     {
-        public GetSystemCapabilitiesResponse(CapabilityStatement capabilityStatement)
+        public GetSystemCapabilitiesResponse(ResourceElement capabilityStatement)
```
jackliums: Should we add validation to make sure the InstanceType is correct? E.g., if someone passes in ResourceElement with underlying object being Bundle should be rejected?

src/Microsoft.Health.Fhir.Core/Models/FhirResourceFormat.cs
```diff
@@ -0,0 +1,14 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Fhir.Core.Models
+{
+    public enum FhirResourceFormat
+    {
+        Xml = 1,
+        Json = 2,
+        Unknown = 3,
```
jackliums: Unknown should be the first item with 0

src/Microsoft.Health.Fhir.Core/Models/OperationOutcomeIssue.cs
```diff
@@ -0,0 +1,23 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Fhir.Core.Models
+{
+    public class OperationOutcomeIssue
+    {
+        public OperationOutcomeIssue(string severity, string code, string diagnostics)
```
jackliums: nit: null check

src/Microsoft.Health.Fhir.Core/Models/ResourceElement.cs
```diff
@@ -0,0 +1,53 @@
...
+    public class ResourceElement
+    {
+        private readonly Lazy<EvaluationContext> _context;
+
+        public ResourceElement(ITypedElement instance)
+        {
+            EnsureArg.IsNotNull(instance, nameof(instance));
+
+            Instance = instance;
+            _context = new Lazy<EvaluationContext>(() => new EvaluationContext(instance));
```
jackliums: Could just be lazy loaded property?

src/Microsoft.Health.Fhir.Api/Features/Filters/OperationOutcomeExceptionFilterAttribute.cs
brendankowitz: Yep, splitting API will be different PR

src/Microsoft.Health.Fhir.Core/Models/FhirResourceFormat.cs
brendankowitz: Agree, these are the same as in Fhir.Stu3.ResourceFormat. Should we deviate?

test/Microsoft.Health.Fhir.Tests.E2E/Rest/VReadTests.cs
brendankowitz: This is a separate task. We'll need to leverage the new testing attributes. We'll also need to spike perhaps wrapping the FhirClient with an agnostic proxy

src/Microsoft.Health.Fhir.Api.UnitTests/Features/Audit/AuditLoggingFilterAttributeTests.cs
brendankowitz: I'll keep looking into this

src/Microsoft.Health.Fhir.Core/Models/IModelFactory.cs
```diff
@@ -0,0 +1,26 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using Hl7.Fhir.ElementModel;
+using Hl7.FhirPath;
+
+namespace Microsoft.Health.Fhir.Core.Models
+{
+    public interface IModelFactory
```
jackliums: This doesn't sound like a factory. IModelInfoProvider? Resolver?

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/ReferenceSearchValue.cs
```diff
@@ -21,11 +20,11 @@ public class ReferenceSearchValue : ISearchValue
         /// <param name="baseUri">The base URI of the resource.</param>
         /// <param name="resourceType">The resource type.</param>
         /// <param name="resourceId">The resource id.</param>
-        public ReferenceSearchValue(ReferenceKind referenceKind, Uri baseUri, ResourceType? resourceType, string resourceId)
+        public ReferenceSearchValue(ReferenceKind referenceKind, Uri baseUri, string resourceType, string resourceId)
         {
             if (baseUri != null)
             {
-                EnsureArg.IsNotNull(resourceType, nameof(resourceType));
+                EnsureArg.IsNotNullOrWhiteSpace(resourceType, nameof(resourceType));
```
jackliums: Should we also validate to make sure it's supported ResourceType?

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/ReferenceSearchValueParser.cs
```diff
@@ -46,7 +47,11 @@ public ReferenceSearchValue Parse(string s)
             if (match.Success)
             {
                 string resourceTypeInString = match.Groups[ResourceTypeCapture].Value;
-                ResourceType resourceType = Enum.Parse<ResourceType>(resourceTypeInString);
+
+                if (!ModelFactory.IsKnownResource(resourceTypeInString))
+                {
+                    throw new RequestNotValidException(string.Format(Resources.ResourceNotSupported, resourceTypeInString));
```
jackliums: I think maybe `ArgumentException` is more appropriate since Parse could be used by other calls too.

src/Microsoft.Health.Fhir.Core/Models/CodeableConceptInfo.cs
```diff
@@ -0,0 +1,25 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using EnsureThat;
+
+namespace Microsoft.Health.Fhir.Core.Models
+{
+    public class CodeableConceptInfo
```
jackliums: Maybe named this something like `SystemValuePair`? `SystemAndValue`?

jackliums: Actually, shouldn't this be system and code so more like SystemCodePair? Then you can collapse CodingInfo into here as well.

src/Microsoft.Health.Fhir.Core/Models/FhirResourceFormat.cs
jackliums: I think our code should follow our coding practices. I don't imagine us doing direct cast anyways if we need to convert so i don't think it would be a problem if the number is different?

src/Microsoft.Health.Fhir.Core/Models/FhirSpecification.cs
```diff
@@ -0,0 +1,14 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Fhir.Core.Models
+{
+    public enum FhirSpecification
+    {
+        Any,
```
jackliums: When woudl we use Any?

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Search/SearchValues/ReferenceSearchValueParserTests.cs
```diff
@@ -22,6 +23,8 @@ public class ReferenceSearchValueParserTests
 
         public ReferenceSearchValueParserTests()
         {
+            ModelExtensions.SetModelFactory();
```
jackliums: Let's chat about this a little bit later. The ambient context might make thigns a little bit challenging when we have STU3 and R4 tests running in parallel (potentially).

src/Microsoft.Health.Fhir.Api/Modules/FhirModule.cs
brendankowitz: Added a local method that sets the common properties

src/Microsoft.Health.Fhir.Core/Messages/Get/GetSystemCapabilitiesResponse.cs
brendankowitz: I don't have a good solution for this atm

src/Microsoft.Health.Fhir.Core/Models/FhirResourceFormat.cs
brendankowitz: Sure

src/Microsoft.Health.Fhir.Core/Models/FhirSpecification.cs
brendankowitz: No reason yet, will remove

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Search/SearchValues/ReferenceSearchValueParserTests.cs
brendankowitz: Agree

src/Microsoft.Health.Fhir.Core/Models/IModelFactory.cs
brendankowitz: Going with `IModelInfoProvider`

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/ReferenceSearchValue.cs
brendankowitz: Added `ModelInfoProvider.EnsureValidResourceType`

src/Microsoft.Health.Fhir.Core/Models/CodeableConceptInfo.cs
brendankowitz: Merged both into `CodingInfo`

src/Microsoft.Health.Fhir.Stu3.Core/Extensions/ModelExtensions.cs
```diff
@@ -0,0 +1,93 @@
...
+using Hl7.Fhir.Model;
+using Microsoft.Health.Fhir.Core.Models;
+
+namespace Microsoft.Health.Fhir.Core.Extensions
+{
+    public static class ModelExtensions
+    {
+        static ModelExtensions()
+        {
+            SetModelInfoProvider();
```
jackliums: The ambient context worries me a little bit. I think it's fine for now since it's already quite large change but we should think about this a little bit more.

src/Microsoft.Health.Fhir.Api/Modules/FhirModule.cs
```diff
@@ -57,12 +59,31 @@ public void Load(IServiceCollection services)
...
+            {
+                return new Dictionary<FhirResourceFormat, Func<string, string, DateTimeOffset, ResourceElement>>
                 {
-                    { ResourceFormat.Json, str => jsonParser.Parse<Resource>(str) },
-                    { ResourceFormat.Xml, str => xmlParser.Parse<Resource>(str) },
+                    {
+                        FhirResourceFormat.Json, (str, version, lastModified) =>
+                        {
+                            var resource = jsonParser.Parse<Resource>(str);
+                            return SetMetadata(resource, version, lastModified);
```
jackliums: Just curious why this logic needs to happen in IoC container? It could be implemented in one place in `ResourceDeserializer.DeserializeRaw` right? Then the registration will much simpler just FhirResourceFormat against a method. I feel like having business logic injected in IoC container is a little odd?

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Search/Expressions/Parsers/SearchValueExpressionBuilderTests.cs
```diff
@@ -584,7 +584,7 @@ public void GivenAReferenceUsingResourceTypeAndResourceIdTargetingInternalResour
             const string resourceId = "123";
             string reference = $"{resourceType}/{resourceId}";
 
-            _referenceSearchValueParser.Parse(reference).Returns(new ReferenceSearchValue(ReferenceKind.Internal, null, resourceType, resourceId));
+            _referenceSearchValueParser.Parse(reference).Returns(new ReferenceSearchValue(ReferenceKind.Internal, null, resourceType.ToString(), resourceId));
```
jackliums: nit: these could just be string from beginning. I will fix them as I refactor the tests.

src/Microsoft.Health.Fhir.Stu3.Core/Features/Search/Expressions/Expression.cs
```diff
@@ -4,14 +4,15 @@
...
+using EnsureThat;
 using Hl7.Fhir.Model;
 
 namespace Microsoft.Health.Fhir.Core.Features.Search.Expressions
 {
     /// <summary>
     /// Represents an expression.
     /// </summary>
-    public abstract class Expression
+    public abstract class Expression : IExpression
```
jackliums: We shouldn't need to move the expression stuff since it should be generic enough to be version agnostic.

src/Microsoft.Health.Fhir.Stu3.Core/Features/Resources/Create/CreateResourceHandler.cs
```diff
@@ -48,7 +50,7 @@ public async Task<UpsertResourceResponse> Handle(CreateResourceRequest message,
 
             resource.VersionId = result.Wrapper.Version;
 
-            return new UpsertResourceResponse(new SaveOutcome(resource, SaveOutcomeType.Created));
+            return new UpsertResourceResponse(new SaveOutcome(resource.ToResourceElement(), SaveOutcomeType.Created));
```
jackliums: I think the FhirDataStore could just return ResourceElement instead of Resource right?

src/Microsoft.Health.Fhir.Core/Models/SearchParameterInfo.cs
```diff
@@ -0,0 +1,21 @@
...
+    public class SearchParameterInfo
+    {
+        public SearchParameterInfo(string name)
+        {
+            EnsureArg.IsNotNullOrWhiteSpace(name, nameof(name));
+
+            Name = name;
+        }
+
+        public string Name { get; set; }
```
jackliums: remove setter

src/Microsoft.Health.Fhir.Stu3.Core/Features/Resources/Create/CreateResourceHandler.cs
brendankowitz: Yes it should, I think this can be done in ab#69127

src/Microsoft.Health.Fhir.Stu3.Core/Features/Search/Expressions/Expression.cs
brendankowitz: Linking to the task ab#69162

src/Microsoft.Health.Fhir.Api/Modules/FhirModule.cs
brendankowitz: I'm not currently sure how to update properties on ITypedElement, ResourceDeserializer is still in Core so it doesn't have access to Resource. I don't have a good solution for this yet

PR #462, opened 2019-05-13
Support checking the job hash and return existing job if there is a match

jackliums:
## Description
Add support for de-dupling the export job. This could happen if user queues an export job but failed to get response. User might try again in this case, which could cause duplicated export job.

The job record now contains a hash which is generated by:

- The URI of the incoming job (which includes all of the parameters used by the job.
- The requestor claims (which identifies the user).

When the hash matches and the there is already a queued or running job, that job is returned. Otherwise a new job is created. This is best effort (without transaction) since the likelihood of this happens is small.

To support this, the existing `IClaimsIndexer` was refactored as `IClaimsExtractor` and is now moved under Security instead of under Persistent so the code can be reused. Also, added "iss" to the list of claims to be extracted in addition to "oid" since "oid" is in the context of "iss" and we now support changing the issuer in the managed service.

## Related issues
Addresses #346 . Work item: [AB#68901](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/68901).

## Testing
Unit test and integration test.


src/Microsoft.Health.Fhir.Core/Features/Security/ClaimsPrincipalExtractor.cs
```diff
@@ -10,14 +10,14 @@
 using Microsoft.Health.Fhir.Core.Configs;
 using Microsoft.Health.Fhir.Core.Features.Context;
 
-namespace Microsoft.Health.Fhir.Core.Features.Persistence
+namespace Microsoft.Health.Fhir.Core.Features.Security
 {
-    public class ClaimsIndexer : IClaimsIndexer
+    public class ClaimsPrincipalExtractor : IClaimsExtractor
```
brandonpollett: Maybe `PrincipalClaimsExtractor`?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/CreateExportRequestHandler.cs
```diff
@@ -31,18 +38,33 @@ public async Task<CreateExportResponse> Handle(CreateExportRequest request, Canc
...
+            var hashObject = new
+            {
+                request.RequestUri,
+                RequestorClaims = requestorClaims,
+                request.DestinationInfo,
+            };
 
-            // Store the destination secret
-            SecretWrapper result = await _secretStore.SetSecretAsync(jobRecord.SecretName, request.DestinationInfo.ToJson());
+            string hash = JsonConvert.SerializeObject(hashObject).ComputeHash();
```
brandonpollett: Should we enforce ordering on the claims collection so that it serializes the same every time?

src/Microsoft.Health.Fhir.Core/Extensions/StringExtensions.cs
```diff
@@ -6,13 +6,21 @@
...
     {
-        public static string ComputeHash(string data)
+        /// <summary>
+        /// Computes SHA256 hash based of <paramref name="data"/>.
+        /// </summary>
+        /// <param name="data">The data to compute hash on.</param>
+        /// <returns>The computed hash string.</returns>
+        public static string ComputeHash(this string data)
         {
+            EnsureArg.IsNotNull(data, nameof(data));
```
brandonpollett: Is hashing whitespace valid?

test/Microsoft.Health.Fhir.Tests.Integration/Features/Operations/Export/CreateExportRequestHandlerTests.cs
```diff
@@ -0,0 +1,151 @@
...
+        {
+            var request = new CreateExportRequest(RequestUrl, DestinationType, ConnectionString);
+
+            CreateExportResponse response = await _createExportRequestHandler.Handle(request, _cancellationToken);
+
+            var newRequest = new CreateExportRequest(RequestUrl, DestinationType, ConnectionString);
+
+            CreateExportResponse newResponse = await _createExportRequestHandler.Handle(request, _cancellationToken);
+
+            Assert.NotNull(newResponse);
```
brandonpollett: nit: should we also assert the `response` isn't null?

test/Microsoft.Health.Fhir.Tests.Integration/Features/Operations/Export/CreateExportRequestHandlerTests.cs
```diff
@@ -0,0 +1,151 @@
...
+        [Fact]
+        public async Task GivenDifferentRequestor_WhenCreatingAnExportJob_ThenNewJobShouldBeCreated()
+        {
+            _claimsExtractor.ExtractImpl = () => new[] { KeyValuePair.Create("oid", "user1") };
+
+            var request = new CreateExportRequest(RequestUrl, DestinationType, ConnectionString);
+
+            CreateExportResponse response = await _createExportRequestHandler.Handle(request, _cancellationToken);
+
+            _claimsExtractor.ExtractImpl = () => new[] { KeyValuePair.Create("oid", "user1") };
```
brandonpollett: How are these different requestors? It looks like the claims extractor extracts the same thing. It looks like the connection string is what is changing.

src/Microsoft.Health.Fhir.Core/Features/Security/ClaimsPrincipalExtractor.cs
jackliums: That's what I had originally but I renamed it because the .NET type is called ClaimsPrincipal but I guess we are extracting the Claims so PrincipalClaimsExtractor makes more sense.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/CreateExportRequestHandler.cs
jackliums: Good catch. Originally this was in JobRecord which had `requestorClaims?.OrderBy(claim => claim.Key, StringComparer.Ordinal).ToList();`. I moved the code out but forgot to move the ordering.

src/Microsoft.Health.Fhir.Core/Extensions/StringExtensions.cs
jackliums: if one wishes to compute hash of whitespace 😊

test/Microsoft.Health.Fhir.Tests.Integration/Features/Operations/Export/CreateExportRequestHandlerTests.cs
jackliums: It's already verified by the test above.

test/Microsoft.Health.Fhir.Tests.Integration/Features/Operations/Export/CreateExportRequestHandlerTests.cs
jackliums: The pending change was accidentally not staged before I committed change.

test/Microsoft.Health.Fhir.Tests.Integration/Features/Operations/Export/CreateExportRequestHandlerTests.cs
```diff
@@ -0,0 +1,172 @@
...
+            Assert.NotNull(response);
+            Assert.NotEmpty(response.JobId);
+
+            SecretWrapper secret = await _secretStore.GetSecretAsync($"Export-Destination-{response.JobId}");
+
+            Assert.NotNull(secret);
+        }
+
+        [Fact]
+        public async Task GivenThereIsAMatchingJob_WhenCreatingAnExportJob_ThenNewJobShouldBeCreated()
```
namadabu: _ThenExistingJobShouldBeReturned()

test/Microsoft.Health.Fhir.Tests.Integration/Features/Operations/FhirOperationConstants.cs
```diff
@@ -0,0 +1,17 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Fhir.Tests.Integration.Features.Operations
+{
+    public static class FhirOperationConstants
```
namadabu: nit: FhirOperationTestConstants?

jackliums: Fixed.

test/Microsoft.Health.Fhir.Tests.Integration/Features/Operations/Export/CreateExportRequestHandlerTests.cs
jackliums: Good catch.

Issue #104, opened 2018-10-13
The search indexer needs to support converting from one FHIR element to multiple ISearchValue


jackliums:
Just noticed that the expression for Location.near and Location.near-distance are both pointing to Location.position but Location.near is a token search parameter and Location.near-distance is a quantity search parameter.

Currently, the code assumes that each FHIR element can only be converted into one ISearchValue type, which does not seem the case anymore.

Issue #103, opened 2018-10-13
Quantity values need to be indexed with min and max


jackliums:
Condition.abatement-age search parameter is a quantity search parameter but the value can be a Range, which has low and high. To support this scenario, the quantity value needs to be indexed with both numbers in order to evaluate.

CaitlinV39:
Opened investigation story for this AB#75836

namadabu:
@CaitlinV39 I haven't yet looked into this as part of my current feature work. Not sure about others.

PR #464, opened 2019-05-14
Add Low and High to quantity and number search params

johnstairs:
## Description
Quantity and Number search parameters currently assume a single decimal value, but these can be ranges.

We will be setting up the SQL tables with ranges so this change updates the search extraction logic to specify low and high values. We are not adding support for range-based quantity and number searches in Cosmos DB in the PR. Rather, we store things as-is if `low == high`, but add new properties if `low != high`. We will need to perform a data migration before we can support these searches in Cosmos DB.

## Related issues
Addresses #104. Related to (but does not fix) #103.


src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/NumberSearchValue.cs
```diff
@@ -17,12 +17,23 @@ public class NumberSearchValue : ISearchValue
         /// Initializes a new instance of the <see cref="NumberSearchValue"/> class.
         /// </summary>
         /// <param name="number">The number value.</param>
-        public NumberSearchValue(decimal number) => Number = number;
+        /// <param name="low">The lower bound of the quantity range.</param>
```
jackliums: we could still have `NumberSearchValue(decimal number)` for convenience when there is only one value? It will just set both low and high the same number.

jackliums: Also, the `number` comment above doesn't apply anymore

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/NumberSearchValue.cs
```diff
@@ -17,12 +17,23 @@ public class NumberSearchValue : ISearchValue
...
+        /// <param name="high">The upper bound of the quantity range.</param>
+        public NumberSearchValue(decimal? low, decimal? high)
+        {
+            Low = low;
+            High = high;
+        }
+
+        /// <summary>
```
jackliums: We shouldn't allow both low and high be null.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Search/SearchIndexEntryJObjectGenerator.cs
```diff
@@ -83,14 +83,31 @@ void ISearchValueVisitor.Visit(DateTimeSearchValue dateTime)
 
         void ISearchValueVisitor.Visit(NumberSearchValue number)
         {
-            AddProperty(SearchValueConstants.NumberName, number.Number);
+            if (number.Low == number.High)
```
jackliums: For maintaining backwards compatibility until we can do migration, we will want to dual write to `SearchValueConstants.NumberName` when `Low == High`.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Search/SearchIndexEntryJObjectGenerator.cs
```diff
@@ -83,14 +83,31 @@ void ISearchValueVisitor.Visit(DateTimeSearchValue dateTime)
...
+            {
+                AddProperty(SearchValueConstants.NumberName, number.Low);
```
jackliums: Should we consider writing to low and high here as well? I guess if we are doing migration later, we could just do all of it at once for existing data.

johnstairs: Yup, I'll update

src/Microsoft.Health.Fhir.Core/Features/Search/Converters/FhirElementToSearchValueTypeConverter.cs
```diff
@@ -38,14 +39,15 @@ public IEnumerable<ISearchValue> ConvertTo(object value)
...
         }
 
         /// <summary>
         /// Converts the FHIR element to a list of <see cref="ISearchValue"/>.
         /// </summary>
         /// <param name="value">The value to convert.</param>
+        /// <param name="searchParameterType">The target type</param>
         /// <returns>A list of <see cref="ISearchValue"/>.</returns>
-        protected abstract IEnumerable<ISearchValue> ConvertTo(TFhirElement value);
+        protected abstract IEnumerable<ISearchValue> ConvertTo(TFhirElement value, SearchParamType searchParameterType);
```
jackliums: I was thinking about this. Instead of the ConvertTo method takes the serach parameter type, we could make two separate converters (e.g., `RangeToNumberSearchValueTypeConverter` and `RangeToQuantitySearchValueTypeConverter`) and have the `FhirElementToSearchValueTypeConverterManager` register the converters based on the ElementType + SearchParamType. That way each converter only does one thing and there won't be oddness with SearchParamType being passed in but ignored by most of the converters?

src/Microsoft.Health.Fhir.Core/Features/Search/Converters/RangeToNumberSearchValueTypeConverter.cs
```diff
@@ -0,0 +1,22 @@
...
+namespace Microsoft.Health.Fhir.Core.Features.Search.Converters
+{
+    /// <summary>
+    /// A converter used to convert from <see cref="Range"/> to a list of <see cref="NumberSearchValue"/>.
+    /// </summary>
+    public class RangeToNumberSearchValueTypeConverter : FhirElementToSearchValueTypeConverter<Range, NumberSearchValue>
+    {
+        protected override IEnumerable<NumberSearchValue> ConvertTo(Range value)
+        {
+            yield return new NumberSearchValue(value.Low?.Value, value.High?.Value);
```
jackliums: I think we need to put condition here to skip the creation of the NumberSearchValue if Low and High are both null. If there is malformed incoming data, we don't want it to fail the request.

src/Microsoft.Health.Fhir.Core/Features/Search/Converters/RangeToQuantitySearchValueTypeConverter.cs
```diff
@@ -0,0 +1,24 @@
...
+namespace Microsoft.Health.Fhir.Core.Features.Search.Converters
+{
+    /// <summary>
+    /// A converter used to convert from <see cref="Range"/> to a list of <see cref="QuantitySearchValue"/>.
+    /// </summary>
+    public class RangeToQuantitySearchValueTypeConverter : FhirElementToSearchValueTypeConverter<Range, QuantitySearchValue>
+    {
+        protected override IEnumerable<QuantitySearchValue> ConvertTo(Range value)
+        {
+            Quantity quantityRepresentative = value.Low ?? value.High;
```
jackliums: Same here. Skip if both low and high are null.

Also, maybe put comment and link to the spec indicating that the low and high must have the same unit? (http://hl7.org/fhir/STU3/datatypes.html#range)

src/Microsoft.Health.Fhir.Core/Features/Search/SearchIndexer.cs
```diff
@@ -227,5 +227,30 @@ private IEnumerable<SearchIndexEntry> ProcessNonCompositeSearchParameter(SearchP
 
             return results;
         }
+
+        private static Type GetSearchValueTypeForSearchParamType(SearchParamType searchParamType)
```
jackliums: We could also use attribute and regsiter the type against derived class instead of manually map it here (but i think this will need to change a bit with the R4 work anyway so we can see how it evolve later).

johnstairs: To be revisited once R4 lands.

PR #468, opened 2019-05-15
Add indexing for search parameters

johnstairs:
## Description
Stores extracted search parameter values into tables for each type (token, quantity, etc).

Composite search parameters are not implemented yet.

## Related issues
Addresses [AB#69044](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/69044)


src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
```diff
@@ -273,6 +273,392 @@ ON dbo.CompartmentAssignment
...
+    ReferenceResourceVersion int NULL
+)
+
+CREATE TABLE dbo.ReferenceSearchParam
+(
+    ResourceSurrogateId bigint NOT NULL,
+    SearchParamId smallint NOT NULL,
+    BaseUri varchar(128) NULL,
+    ReferenceResourceTypeId smallint NOT NULL,
+    ReferenceResourceId varchar(64) NOT NULL,
```
jackliums: I am not sure if 64 is long enough. In the case of external resource reference, this will be an URL.

johnstairs:
I'm going to be adding unit tests once the the schema and R4 refactoring settle

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
```diff
@@ -273,6 +273,392 @@ ON dbo.CompartmentAssignment
...
+    ReferenceResourceVersion int NULL,
+    IsHistory bit NOT NULL,
+) WITH (DATA_COMPRESSION = PAGE)
+
+CREATE CLUSTERED INDEX IXC_ReferenceSearchParam
+ON dbo.ReferenceSearchParam
+(
+    ResourceSurrogateId,
+    SearchParamId,
+    ReferenceResourceTypeId,
```
johnstairs: Going to move the TypeId later in the indexes for this table, because you can perform a search without specifying the type. @jackliums, I don't think we handle this part of the spec:
> Servers SHOULD reject a search where the logical id refers to more than one matching resource across different types

jackliums: That's not yet implemented. We do have the metadata about the target resource type so we should be able to implement this pretty easily.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/SearchParameterRowGenerator.cs
```diff
@@ -0,0 +1,37 @@
...
+        }
+
+        protected SqlServerFhirModel Model { get; }
+
+        public virtual IEnumerable<TRow> GenerateRows(ResourceMetadata input)
+        {
+            return input.GetSearchIndexEntriesByType(typeof(TSearchValue))
+                .Where(v => ShouldGenerateRow(v.SearchParameter, (TSearchValue)v.Value))
+                .Select(v => GenerateRow(Model.GetSearchParamId(v.SearchParameter.Url), v.SearchParameter, (TSearchValue)v.Value))
+                .Distinct();
```
johnstairs: Remove Distinct() until hashcode and equality implementeation is generated. Currently being done in the stored procedure.

cunninghamjc:
> --

Need to add new params to comments

---
Refers to: src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql:707 in 5a2c7c3. [](commit_id = 5a2c7c39e04fe1cb89b3173abae8029b51e72684, deletion_comment = False)

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
```diff
@@ -273,6 +273,392 @@ ON dbo.CompartmentAssignment
...
+(
+    SearchParamId smallint NOT NULL,
+    Text nvarchar(400) COLLATE Latin1_General_CI_AI NOT NULL
+)
+
+CREATE TABLE dbo.StringSearchParam
+(
+    ResourceSurrogateId bigint NOT NULL,
+    SearchParamId smallint NOT NULL,
+    Text nvarchar(400) COLLATE Latin1_General_CI_AI NOT NULL,
```
cunninghamjc: How did we pick 400?  What happens if we attempt to store a longer value?

johnstairs: Great question. This is not set in stone yet. Clustered and nonclustered cannot exceed 900 and 1700 bytes respectively. Including the resource surrogate ID in the clustered index probably means that there is not much to be gained by including the Text column. So we can make it longer if we only put it in nonclustered indexes.
But there still will be a limit. To lift the limit, we could have an overflow column that is not indexed that we only reach into when predicates are exceed the key column length.
Right now, the resource insert will fail if there is a string search param value > 400. 
FWIW, Hapi has a max length of 200.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/SearchParameterRowGenerator.cs
johnstairs: In next PR

PR #471, opened 2019-05-16
Updates to History batch test to make it more reliable

feordin:
## Description
Updates to the History batch test to make it more reliable

## Related issues
Addresses [issue [AB#69164](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/69164)].

### DIFF
--- a/test/Microsoft.Health.Fhir.Tests.E2E/Rest/HistoryTests.cs
+++ b/test/Microsoft.Health.Fhir.Tests.E2E/Rest/HistoryTests.cs
@@ -240,8 +240,8 @@ namespace Microsoft.Health.Fhir.Tests.E2E.Rest
 
             Thread.Sleep(500);
 
-            newResources.Add(await Client.CreateAsync(Samples.GetJsonSample("ObservationWithBloodPressure") as Observation));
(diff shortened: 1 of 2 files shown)


test/Microsoft.Health.Fhir.Tests.E2E/Rest/HistoryTests.cs
```diff
@@ -213,11 +213,35 @@ public async Task WhenGettingSystemHistory_GivenAQueryThatReturnsMoreThan10Resul
...
+                FhirResponse<Bundle> readResponse = Client.SearchAsync("_history?_since=" + sinceUriString).Result;
+
+                if (readResponse.Resource.Entry.Count == 10)
+                {
+                    historyHas10Entries = true;
+                    break;
+                }
+
+                Thread.Sleep(2000);
+                since = DateTime.UtcNow;
```
brandonpollett: Why is this being reset here? Wouldn't this never work if there was a lag and say one entry was created before we reset it and the other 9 after?

feordin: That is correct!  Thanks for catching this.

PR #477, opened 2019-05-20
Fix deadlock issue in upsert

johnstairs:
Under concurrent calls to the UpsertResource stored procedure, we would run into deadlocks. This was because the `UPDATE`  statement at the beginning of the procedure was taking more locks than intended. 

I've replaced this with a simpler (and more efficient) approach of doing a locking read of the current resource, and based on the results, either updating or deleting only if necessary.

Related to [AB#69044](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/69044)


cunninghamjc:
>             THROW 50404, 'Resource does not exist and create is not allowed', 1;

Just noticed -- do we need to explicitly rollback (or commit) the transaction before throwing?

---
Refers to: src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql:1245 in d59ee1f. [](commit_id = d59ee1f9a3c23a978352bb1c87eb0080e4cd10f8, deletion_comment = False)

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
```diff
@@ -1203,33 +1220,23 @@ AS
...
-        WHERE ResourceTypeId = @resourceTypeId AND ResourceId = @resourceId AND IsHistory = 0 AND (@isDeleted = 0 OR IsDeleted = 0)
-    END
+    -- variables for the existing version of the resource that will be replaced
+    DECLARE @previousResourceSurrogateId bigint
+    DECLARE @previousVersion bigint
+    DECLARE @previousIsDeleted bit
+
+    -- This should place a range lock on a row in the IX_Resource_ResourceTypeId_ResourceId nonclustered filtered index
+    SELECT @previousResourceSurrogateId = ResourceSurrogateId, @previousVersion = Version, @previousIsDeleted = IsDeleted 
+    FROM Resource WITH (UPDLOCK, HOLDLOCK)
```
cunninghamjc: Missing dbo.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
```diff
@@ -1366,7 +1388,7 @@ AS
     END
 
 
-    IF (@etag IS NOT NULL AND @etag <> (@version - 1)) BEGIN
+    IF (@etag IS NOT NULL AND @etag <> @previousVersion) BEGIN
```
cunninghamjc: Should we do this check earlier, before the updates/deletes?

johnstairs: Yeah, we should.

PR #482, opened 2019-05-21
Fix SQL integration tests

johnstairs:
1. Fix failure in SQL integration tests
1. Introduce `DataStore.All` and `Format.All` to make it easier to find tests that are not datastore-agnostic
1. Remove skip from `SqlDataReaderExtensionsTests`


brendankowitz:
:shipit:

Issue #389, opened 2019-03-22
InvalidOperationException in ExtractSearchIndexByCompartmentType() method


garamamo:
**Describe the bug**
The line 
                if (indexEntry.Value is ReferenceSearchValue refValue && CompartmentDefinitionManager.ResourceTypeToCompartmentType.TryGetValue(refValue.ResourceType.Value, out CompartmentType compartmentType))
 throws an invalidoperationexception exception (see below) when ResourceType.Value does not exist. Check for HasValue on refValue.ResourceType if there is no value and skip. This can happen when customer data is invalid.

Found from logs.

**Expected behavior**
No error thrown when resourcetype is not available for a reference search.

**Actual behavior**
InvalidOperationException ()
Severity Count Type HealthMethod MaxVersion Period Clusters AnyProblemId AnyItemId AnyOperationId
Error 118 InvalidOperationException () Fhir.Core.Features.Compartment.CompartmentIndexer.ExtractSearchIndexByCompartmentType Resolute_PaaS_CI.master.20190317.2 00:42:31.0343105 ["ncus"] System.InvalidOperationException at MediatR.Pipeline.RequestPostProcessorBehavior`2+<Handle>d__2.MoveNext 0af14e00-4bde-11e9-80f2-b9a3627a5611 379e99aea12bf84eaee780f2f068b242

PR #484, opened 2019-05-21
Issue #389: InvalidOperationException in ExtractSearchIndexByCompartmentType()

garamamo:
## Description
While extracting Compartment Indices from reference search values, the compartment type is derived from the resource type in the reference search value. However in some rare cases, the resource type is null and only the resource id is set in the search value. The indexer was throwing an InvalidOperationException in such cases as it was trying to derive the compartment type from a non existent resource type. The fix addresses this issue.
Note: Related, it might be possible for the reference search value parser to follow the resource id for such cases above and set the resource type, will discuss with Jack to see if we should open an issue for the same. Opened issue #485.

## Related issues
Addresses #389.

## Testing
Added unit tests for this condition. Also, updated/added more unit tests in the area.


src/Microsoft.Health.Fhir.Stu3.Core/Features/Definition/CompartmentIndexer.cs
```diff
@@ -70,7 +70,7 @@ public CompartmentIndices Extract(string resourceType, IReadOnlyCollection<Searc
 
             foreach (var indexEntry in searchIndices)
             {
-                if (indexEntry.Value is ReferenceSearchValue refValue && CompartmentDefinitionManager.ResourceTypeToCompartmentType.TryGetValue(refValue.ResourceType, out CompartmentType compartmentType))
+                if (indexEntry.Value is ReferenceSearchValue refValue && !string.IsNullOrWhiteSpace(refValue.ResourceType) && CompartmentDefinitionManager.ResourceTypeToCompartmentType.TryGetValue(refValue.ResourceType, out CompartmentType compartmentType))
```
johnstairs: As discussed, please open a bug for when the reference is constrained to a single type. Once we implement referential integrity checking, we will be able to resolve this in all cases.

garamamo: Opened issue #485.

johnstairs:
Please give the PR a descriptive title

PR #491, opened 2019-05-23
Search expression visitor updates

johnstairs:
1. Changed expression visitor methods to take a TContext parameter and return a TOutput. Also renamed them to `VisitBinary`, `VisitChained`, etc. instead of overloads of `Visit` to make code navigation easier.
1. Implemented a simple string representation of expressions to make debugging easier, inspired by s-expressions. These look like `(FieldEqual Quantity 1)` and `(And (FieldGreaterThan Quantity 1) (FieldLessThan Quantity 10))`.
1. Created an abstract rewriter class that rewrites an expression tree. We'll use it for tree transformation in the SQL provider.
1. Removed `IExpression`, since we only ever used `Expression`.

Relates to [AB#68735](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/68735)


src/Microsoft.Health.Fhir.CosmosDb/Features/Search/Queries/ExpressionQueryBuilder.cs
```diff
@@ -84,32 +81,28 @@ internal class ExpressionQueryBuilder : IExpressionVisitor
...
-                    expression.Expression.AcceptVisitor(this);
-                }
-                finally
-                {
-                    _fieldNameOverride = null;
-                }
+                // We do not currently support specifying the system for the _type parameter value.
+                // We would need to add it to the document, but for now it seems pretty unlikely that it will
+                // be specified when searching.
+                expression.Expression.AcceptVisitor(this, context.WithFieldNameOverride(fieldNameOverride: SearchValueConstants.RootResourceTypeName));
```
johnstairs: @feordin FYI, our changes will conflict here.

PR #494, opened 2019-05-23
Changes default ContinuationTokenSizeLimitInKb to 1

brendankowitz:
## Description
Changes default ContinuationTokenSizeLimitInKb to 1

### DIFF
--- a/src/Microsoft.Health.Fhir.Web/appsettings.json
+++ b/src/Microsoft.Health.Fhir.Web/appsettings.json
@@ -50,7 +50,7 @@
     "InitialDatabaseThroughput": null,
     "ConnectionMode": "Direct",
     "ConnectionProtocol": "Tcp",
-    "ContinuationTokenSizeLimitInKb": 2,
+    "ContinuationTokenSizeLimitInKb": 1,
     "DefaultConsistencyLevel": "Session",
     "PreferredLocations": [],
     "RetryOptions": {


src/Microsoft.Health.Fhir.Web/appsettings.json
```diff
@@ -50,7 +50,7 @@
     "InitialDatabaseThroughput": null,
     "ConnectionMode": "Direct",
     "ConnectionProtocol": "Tcp",
-    "ContinuationTokenSizeLimitInKb": 2,
+    "ContinuationTokenSizeLimitInKb": 1,
```
jackliums: what's the reason for changing to 1k and also do we need to worry about backwards compatibility? E.g., someone executed a search with 2k as limit and received a continuation token with 2k then the server gets upgraded and the search request comes in with 2k continuation token but now the limit is set to be 1k?

brendankowitz: We were testing if this was going to resolve problems executing history/paging tests against appservice. The appdeploy.json already sets 1k as the default also.
I'm not sure this makes the tests any more reliable though, thinking of abandoning

johnstairs: We actually have this applied on the arm template already, so this change should not be necessary

johnstairs: @brendankowitz , @feordin I think I figured out the issue. In CI, this setting was not applied so the continuation token would be longer than app service would accept. App Service then responds with a 404 without forwarding the request to us, which I believe is why we did not see the request in AI.

hansenms:
I added this to the template:
https://github.com/microsoft/fhir-server/pull/443

PR #499, opened 2019-05-28
Adds IndentedStringBuilder

johnstairs:
## Description
Introduces an `IntendedStringBuilder` class that combines the programming model of `StringBuilder` and 
`IndentedTextWriter`. It exposes the API surface of StringBuilder but allows you to set the indentation level, optionally in a disposable scope. The programming model looks something like this:

``` csharp
var sb = new IndentedStringBuilder(new StringBuilder())
    .AppendLine("class Foo").AppendLine("{");

using (sb.Indent())
{
    sb.AppendLine("Foo()").AppendLine("{");
    using (sb.Indent())
    {
        sb.Append("// ").AppendLine("hello");
    }

    sb.AppendLine("}");
}

sb.Append("}");
```
The class consists primarily of code generated at build time.

## Related issues
Part of [AB#68735](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/68735)


src/Microsoft.Health.Fhir.SqlServer/IndentedStringBuilder.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// <returns>A scope to be disposed when the indentation level is to be restored</returns>
+        internal IndentedScope Indent() => new IndentedScope(this);
+
+        protected virtual void AppendIndent()
+        {
+            if (_indentPending)
+            {
+                for (int i = 0; i < _indentLevel; i++)
+                {
+                    _inner.Append(IndentString);
```
brandonpollett: Could do something like `_inner.Append(new String(' ', _indentLevel * 4);`

johnstairs: That allocates a new string every time, so will be less efficient

brandonpollett: Or could pad the number of spaces.

brandonpollett: Same issue I suppose.

src/Microsoft.Health.Fhir.SqlServer.UnitTests/IndentedStringBuilderTests.cs
```diff
@@ -0,0 +1,36 @@
...
+                {
+                    sb.Append("// ").AppendLine("hello");
+                }
+
+                sb.AppendLine("}");
+            }
+
+            sb.Append("}");
+
+            Assert.Equal($"class Foo{Environment.NewLine}{{{Environment.NewLine}    Foo(){Environment.NewLine}    {{{Environment.NewLine}        // hello{Environment.NewLine}    }}{Environment.NewLine}}}", sb.ToString());
```
brendankowitz: Would literal work? `@""`? or not on linux?
Might be easier to see the indents in the test

johnstairs: I agree, but then the string will have different line endings depending on the platform, so you would need to normalize

PR #501, opened 2019-05-28
Updates HL7 packages in CI-Latest build

brendankowitz:
## Description
Updates Latest Package CI to include new R4 packages


build/.vsts-package-ci.yml
```diff
@@ -20,14 +20,12 @@ jobs:
...
-      pwsh: true
-      targetType: filePath
-      filePath: build/Update-Packages.ps1
-      arguments: -PackageName $(package.name) -PackageVersion $(packageVersion)
+        $FhirPackage = nuget list Hl7.Fhir.R4 -Prerelease -Source https://www.myget.org/F/fhir-net-api/api/v3/index.json | Out-String
+        $packageR4Version = $FhirPackage.Split()[1]
+        $FhirPackage = nuget list Hl7.Fhir.Stu3 -Prerelease -Source https://www.myget.org/F/fhir-net-api/api/v3/index.json | Out-String
+        $packageStu3Version = $FhirPackage.Split()[1]
+
+        & build/Update-Packages.ps1 -PackageName "Hl7.Fhir.Stu3" -PackageVersion $packageStu3Version | Out-Null
```
jackliums: Won't these two update the same file and therefore, we will only end up with csproj file that's R4?

brendankowitz: This just sets the Version property in the csproj. It looks like that Stu3/R4 can have different build revisions so they both need to be set independently

brendankowitz: This is for this build pipeline: https://microsofthealthoss.visualstudio.com/FhirServer/_build/results?buildId=1407

Issue #503, opened 2019-05-29
Removed the destination info from the request URI.


jackliums:
**Describe the bug**
We store the complete requestUri when saving the ExportJobRecord in the data store. This uri will contain destination connection information in the query parameters. We need to remove this information before storing the requestUri.

**To Reproduce**
1. Create a new export job with _destinationType and _destinationConnectionSettings.

**Expected behavior**
The request URI stored in the datastore should not contain the secret.

**Actual behavior**
The request URI stored in the data store contains the secret.

brandonpollett:
@jackliums can this be closed?

PR #504, opened 2019-05-29
Removed the destination info from the request URI.

jackliums:
## Description
Removed the destination info from the request URI. The destination info contains secret and needs to be stored in the secret store and should not be saved along with the job record.

## Related issues
Addresses #503. Work item: [AB#69209](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/69209).

## Testing
Integration test.

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Operations/Export/CreateExportRequestHandler.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Operations/Export/CreateExportRequestHandler.cs
@@ -5,9 +5,11 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.Specialized;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using System.Web;
 using EnsureThat;
 using MediatR;
 using Microsoft.Health.Fhir.Core.Extensions;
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Fhir.Core/Features/Operations/Export/CreateExportRequestHandler.cs
```diff
@@ -59,7 +61,16 @@ public async Task<CreateExportResponse> Handle(CreateExportRequest request, Canc
 
             if (outcome == null)
             {
-                var jobRecord = new ExportJobRecord(request.RequestUri, hash, requestorClaims);
+                // Remove the connection settings from the request URI store it in the secret store.
+                NameValueCollection queryParameters = HttpUtility.ParseQueryString(request.RequestUri.Query);
+
+                queryParameters.Remove(KnownQueryParameterNames.DestinationType);
```
jackliums: Because the connection settings contains secret, we should remove it from the RequestURI so that it's not stored in the datastore in plain text.

The [spec](https://github.com/HL7/bulk-data/blob/master/spec/export/index.md#response---complete-status) requires the original URI to be returned when the job is completed. However, we delete the secret after the job is completed so we won't able to add these parameters back. We could choose to store the secret indefinitely but that might be a security concern. I think it's a fine compromise to return the original URL without the secret. What do you think @hansenms?

test/Microsoft.Health.Fhir.Tests.Integration/Features/Operations/Export/CreateExportRequestHandlerTests.cs
```diff
@@ -154,6 +155,22 @@ public async Task GivenDifferentDestination_WhenCreatingAnExportJob_ThenNewJobSh
             Assert.NotEqual(response.JobId, newResponse.JobId);
         }
 
+        [Fact]
+        public async Task GivenDestinationTypeOrDestinationConnectionSettings_WhenCreatingAnExportJob_ThenItShouldBeRemovedFromRequestUri()
+        {
+            const string baseUrl = "http://localhost/$export?_count=100";
+
+            var requestUri = new Uri($"{baseUrl}&_destinationType=type&_destinationConnectionSettings=settings");
```
namadabu: nit: might be nice to have one more param after the _destinationConnectionSettings to make sure that is still part of the url after the removal.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/CreateExportRequestHandler.cs
jackliums: For now, I am going to implement the way it is. If we have feedback that the original URL must be reconstructed, we can fix it at that point.

PR #513, opened 2019-05-29
Package R4 Web project in build

brendankowitz:
## Description
R4.Web project is now packaged for WebDeploy

## Related issues
Addresses [AB#69059](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/69059).

## Testing
Infrastructure that will be hit by E2E tests


samples/templates/default-azuredeploy-sql.json
```diff
@@ -143,7 +151,8 @@
             "FhirServer:Security:Authentication:Audience": "[parameters('securityAuthenticationAudience')]",
             "CosmosDb:ContinuationTokenSizeLimitInKb": "1",
             "SqlServer:Initialize": "true",
-            "DataStore": "SqlServer"
+            "DataStore": "SqlServer",
+            "PROJECT": "[concat('src/Microsoft.Health.Fhir.', parameters('fhirVersion'),'.Web/Microsoft.Health.Fhir.', parameters('fhirVersion'), '.Web.csproj')]"
```
johnstairs: why is PROJECT uppercase?

brendankowitz: Setting is here: https://github.com/projectkudu/kudu/wiki/Customizing-deployments#using-app-settings-instead-of-a-deployment-file

build/package.yml
```diff
@@ -3,10 +3,18 @@ steps:
   # Package web
   
   - task: DotNetCoreCLI@2
-    displayName: 'dotnet publish web'
+    displayName: 'dotnet publish Stu3 web'
     inputs:
       command: publish
-      projects: '**/Microsoft.Health.Fhir.Web.csproj'
+      projects: '**/Microsoft.Health.Fhir.Stu3.Web.csproj'
+      arguments: '--runtime win10-x64 --output $(build.artifactStagingDirectory)/web --configuration $(buildConfiguration) --version-suffix $(build.buildnumber)'
```
jackliums: nit: since the only difference is the displayName and projects, maybe consider moving this task into a template so it can be reused or at least make the arguments into a variable and reuse here so that if we need to update, it will be in one place?

samples/templates/default-azuredeploy-sql.json
brendankowitz: Now merged if needed in `combinedFhirServerConfigProperties`

PR #514, opened 2019-05-30
Downgrade DocumentDB.Core nuget to 2.2.2

feordin:
## Description
Downgrade DocumentDB.Core nuget to 2.2.2 to avoid null reference exception with continuation tokens

## Related issues
Addresses [issue #].

## Testing
History tests which result in batched responses were enabled.


jackliums:
Let's abandon this PR since downgrade will not fix the issue and introduce new problem.

Issue #517, opened 2019-06-02
PartialDateTime.ToString() results in incorrect second component


johnstairs:
**Describe the bug**
The values of history parameters `_since` and `_before` are truncated before being used in a database query when the second component is less than 10. 

**To Reproduce**
1. Parse "`2018-01-25T10:12:03.2290000+00:00` as a PartialDateTime
2. Call `ToString()`

**Expected behavior**
`2018-01-25T10:12:03.2290000+00:00`

**Actual behavior**
`2018-01-25T10:12:3.2290000+00:00`
Note that the seconds are specified as `3` and not `03` as it should.
When parsed again, the it becomes `2018-01-25T10:12:00.0000000+00:00` because it thinks the seconds are not specified at all.

johnstairs:
@feordin @jack, this was causing instability in the history tests.

LTA-Thinking:
Fixed in PR #520

PR #520, opened 2019-06-03
SQL search implementation (parity with Cosmos DB)

johnstairs:
## Description
The SQL provider now supports the same search capabilities as the Cosmos DB provider. (Actually, it also supports ranges for quantity and number search parameters).
This is implemented as a number of transformations applied to the expression tree before finally emitting SQL.

### Example
This search query:
`/Patient/x/Observation?value-quantity=5.4|http://unitsofmeasure.org|mg`
is turned into this SQL:

``` SQL
DECLARE @p0 SmallInt = 77;
DECLARE @p1 TinyInt = 3;
DECLARE @p2 VarChar(64) = 'x';
DECLARE @p3 SmallInt = 825;
DECLARE @p4 Int = 12;
DECLARE @p5 NVarChar(256) = 'mg';
DECLARE @p6 Decimal(18,6) = 5.35;
DECLARE @p7 Decimal(18,6) = 5.45;

WITH cte0 AS
(
    FROM dbo.CompartmentAssignment
    WHERE IsHistory = 0
        AND ResourceTypeId = @p0
        AND CompartmentTypeId = @p1
        AND ReferenceResourceId = @p2
),
cte1 AS
(
    FROM dbo.QuantitySearchParam
    WHERE IsHistory = 0
        AND ResourceSurrogateId IN (SELECT Sid1 FROM cte0)
        AND ResourceTypeId = @p0
        AND SearchParamId = @p3
        AND SystemId = @p4
        AND QuantityCodeId IN (SELECT QuantityCodeId FROM dbo.QuantityCode WHERE Value = @p5)
        AND SingleValue IS NOT NULL AND SingleValue >= @p6
        AND SingleValue IS NOT NULL AND SingleValue <= @p7

),
cte2 AS
(
    SELECT Sid1 FROM cte1
    UNION ALL
    FROM dbo.QuantitySearchParam
    WHERE IsHistory = 0
        AND ResourceSurrogateId IN (SELECT Sid1 FROM cte0)
        AND ResourceTypeId = @p0
        AND SearchParamId = @p3
        AND SystemId = @p4
        AND QuantityCodeId IN (SELECT QuantityCodeId FROM dbo.QuantityCode WHERE Value = @p5)
        AND LowValue IS NOT NULL AND HighValue >= @p6
        AND LowValue IS NOT NULL AND LowValue <= @p7

),
cte3 AS
(
    SELECT DISTINCT TOP 11 Sid1
    FROM cte2
    ORDER BY Sid1 ASC
)
SELECT r.ResourceTypeId, r.ResourceId, r.Version, r.IsDeleted, r.ResourceSurrogateId, r.RequestMethod, r.RawResource
FROM dbo.Resource r
WHERE r.ResourceSurrogateId IN (SELECT Sid1 FROM cte3)
```

## Related issues
Addresses [AB#68735](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/68735) and #517


src/Microsoft.Health.Fhir.Core/Features/Persistence/ResourceRequest.cs
```diff
@@ -19,15 +19,14 @@ protected ResourceRequest()
 
         public ResourceRequest(Uri url, string method)
```
brendankowitz: Should we add a default value, i.e. url = null when we don't enforce?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/ConcatenationRewriter.cs
```diff
@@ -0,0 +1,76 @@
...
+    {
+        private readonly IExpressionVisitor<object, bool> _scout;
+
+        protected ConcatenationRewriter(IExpressionVisitor<object, bool> scout)
+        {
+            EnsureArg.IsNotNull(scout, nameof(scout));
+            _scout = scout;
+        }
+
+        public override Expression VisitSqlRoot(SqlRootExpression expression, object context)
```
brendankowitz: EnsureArgs?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/NormalizedSearchParameterQueryGeneratorFactory.cs
```diff
@@ -0,0 +1,176 @@
...
+            {
+                generator = GetGenerator(searchParameterInfo);
+                _cache.TryAdd(searchParameterInfo.Url, generator);
+            }
+
+            return generator;
+
+            NormalizedSearchParameterQueryGenerator GetGenerator(SearchParameterInfo param)
+            {
+                switch (param.Type)
```
brendankowitz: FYI We're adding a new enum `Special` for R4

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/ConcatenationRewriter.cs
johnstairs: We are only calling this from Accept methods, which pass `this`. I'd rather use C# 8 nullable parameters for this when it's available

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/NormalizedSearchParameterQueryGeneratorFactory.cs
```diff
@@ -0,0 +1,176 @@
...
+
+                    default:
+                        throw new InvalidOperationException($"Unexpected search parameter type {param.Type}");
+                }
+            }
+        }
+
+        public NormalizedSearchParameterQueryGenerator VisitBinary(BinaryExpression expression, object context)
+        {
+            throw new InvalidOperationException("Should not get here");
```
brandonpollett: Nit: exception message should be more descriptive.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Expressions/Parsers/SearchParameterExpressionParser.cs
```diff
@@ -93,7 +93,7 @@ public class SearchParameterExpressionParser : ISearchParameterExpressionParser
                         string.Format(CultureInfo.InvariantCulture, Core.Resources.ModifierNotSupported, modifier, searchParameter.Name));
                 }
 
-                outputExpression = Expression.Contains(FieldName.TokenText, null, value, true);
+                outputExpression = Expression.StartsWith(FieldName.TokenText, null, value, true);
```
brandonpollett: I believe for tokens, the ':text" modifier is actually a contains. 

From the docs (https://www.hl7.org/fhir/search.html): 
GET [base]/Condition?code:text=headache -> Search for any Condition with a code that has a text "headache" associated with it (either in the text, or a display)

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/NumericRangeRewriter.cs
```diff
@@ -0,0 +1,73 @@
...
+                default:
+                    return expression;
+            }
+
+            switch (expression.BinaryOperator)
+            {
+                case BinaryOperator.GreaterThan:
+                    return Expression.GreaterThan(highField, expression.ComponentIndex, expression.Value);
+                case BinaryOperator.GreaterThanOrEqual:
+                    return Expression.GreaterThanOrEqual(highField, expression.ComponentIndex, expression.Value);
```
brandonpollett: Would this be `GreaterThanOrEqual(lowField...`?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/NumericRangeRewriter.cs
```diff
@@ -0,0 +1,73 @@
...
+                case BinaryOperator.LessThan:
+                    return Expression.LessThan(lowField, expression.ComponentIndex, expression.Value);
+                case BinaryOperator.LessThanOrEqual:
+                    return Expression.LessThanOrEqual(lowField, expression.ComponentIndex, expression.Value);
```
brandonpollett: Same here but, `LessThanOrEqual(highField...`?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/NumericRangeRewriter.cs
```diff
@@ -0,0 +1,73 @@
...
+                case BinaryOperator.Equal:
+                case BinaryOperator.NotEqual:
+                default:
+                    throw new ArgumentOutOfRangeException(expression.BinaryOperator.ToString());
```
brandonpollett: Wouldn't these fall into the following parts of the spec (https://www.hl7.org/fhir/search.html#prefix): 

|  |  |  |
| ---  | ---   | --- |
| eq | the value for the parameter in the resource is equal to the provided value | the range of the search value fully contains the range of the target value |
| ne | the value for the parameter in the resource is not equal to the provided value | the range of the search value does not fully contain the range of the target value |

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/CompartmentSearchParameterQueryGenerator.cs
```diff
@@ -0,0 +1,34 @@
...
+
+            context.StringBuilder
+                .Append(V1.CompartmentAssignment.CompartmentTypeId)
+                .Append(" = ")
+                .Append(context.Parameters.AddParameter(V1.CompartmentAssignment.CompartmentTypeId, compartmentId))
+                .AppendLine()
+                .Append("AND ")
+                .Append(V1.CompartmentAssignment.ReferenceResourceId)
+                .Append(" = ")
+                .Append(context.Parameters.AddParameter(V1.CompartmentAssignment.ReferenceResourceId, expression.CompartmentId));
```
brandonpollett: The use of `compartmentId` and `expression.CompartmentId` is a bit confusing since it appears that one is a lookup table id and one is a resource id.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlSearchParameters.cs
```diff
@@ -0,0 +1,20 @@
...
+using Microsoft.Health.Fhir.Core.Models;
+using Microsoft.Health.Fhir.ValueSets;
+
+namespace Microsoft.Health.Fhir.SqlServer.Features.Search
+{
+    public static class SqlSearchParameters
+    {
+        public const string ResourceSurrogateIdParameterName = "_resourceSurrogateId";
+
+        public static readonly Uri ResourceSurrogateIdUri = new Uri("http://hl7.org/fhir/SearchParameter/Resource-surrogateId");
```
brandonpollett: nit: Since this is of our own creation, should we make it in a Uri that's unique to us?

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Expressions/Parsers/SearchParameterExpressionParser.cs
johnstairs: From the spec:
>`:text` [...] In this case, the search functions as a normal string search.

[...]

>This search is insensitive to casing and included combining characters, like accents or other diacritical marks. Punctuation and non-significant whitespace (e.g. repeated space characters, tab vs space) should also be ignored. By default, a field matches a string query if the value of the field equals or starts with the supplied parameter value, after both have been normalized by case and combining characters. 

Also hapi does not implement it as a contains.
http://hapi.fhir.org/baseDstu3/Observation?code:text=Respiratory matches "Respiratory Rate"
http://hapi.fhir.org/baseDstu3/Observation?code:text=Respiratory&code:text=Rate does not

Neither does Vonk:
https://vonk.fire.ly/Observation?_format=json&code:text=Respiratory&code:text=rate

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/NumericRangeRewriter.cs
johnstairs: No, because you are looking for overlapping ranges. `lowfield` could start before the search value and still overlap with the search range

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/NumericRangeRewriter.cs
johnstairs: The Core layer, when creating these expressions, turns all equality expressions into range expressions, so we do not need to handle equality here.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlSearchParameters.cs
johnstairs: Makes sense

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/NumericRangeRewriter.cs
johnstairs: Same

src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/ExpressionRewriterWithInitialContext.cs
```diff
@@ -0,0 +1,12 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Fhir.Core.Features.Search.Expressions
+{
+    public abstract class ExpressionRewriterWithInitialContext<TContext> : ExpressionRewriter<TContext>, IExpressionVisitorWithInitialContext<TContext, Expression>
+    {
+        public abstract TContext InitialContext { get; }
```
jackliums: Any reason why this couldn't just be `public virtual TContext InitialContext => default` and get rid of `ExpressionRewriterWrithDefaultInitialContext`?

src/Microsoft.Health.Fhir.Core/Models/SearchParameterInfo.cs
```diff
@@ -19,10 +19,27 @@ public class SearchParameterInfo
...
+                name,
+                url,
+                Enum.TryParse<SearchParamType>(searchParamType, out var type) ? (SearchParamType?)type : null,
+                components,
+                expression,
+                targetResourceTypes)
+        {
+        }
+
+        public SearchParameterInfo(
```
jackliums: When would we have SearchParameterInfo without url, type, and expresion and etc defined?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -0,0 +1,301 @@
...
+                            delimited.BeginDelimitedElement();
+                            tableExpression.NormalizedPredicate.AcceptVisitor(tableExpression.SearchParameterQueryGenerator, this);
+                        }
+                    }
+
+                    StringBuilder.AppendLine(")");
+                    break;
+
+                case TableExpressionKind.Top:
+                    StringBuilder.Append("SELECT DISTINCT TOP ").Append(context.MaxItemCount + 1).AppendLine(" Sid1 ")
```
cunninghamjc: We should be able to use a param instead (e.g. top (@maxItemCount)) to increase reuse of cached plans.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/BundleFactory.cs
```diff
@@ -101,7 +101,7 @@ public ResourceElement CreateHistoryBundle(IEnumerable<Tuple<string, string>> un
                         Request = new Bundle.RequestComponent
                         {
                             Method = hasVerb ? (Bundle.HTTPVerb?)httpVerb : null,
-                            Url = r.Request?.Url?.ToString(),
+                            Url = hasVerb ? $"{resource.InstanceType}/{(httpVerb == Bundle.HTTPVerb.POST ? null : resource.Id)}" : null,
```
jackliums: what's this change for?

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Expressions/Parsers/SearchParameterExpressionParser.cs
jackliums: Why was this changed to StartsWith? The spec says it should be partial search so shouldn't that be contains?

> For token: :text (the match does a partial searches on the text portion of a CodeableConcept or the display portion of a Coding), instead of the default search which uses codes. Other defined modifiers are :in, :below, :above and :not-in which are described below.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SearchParameterQueryGenerator.cs
```diff
@@ -0,0 +1,164 @@
...
+                .AppendLine(context.Parameters.AddParameter(searchParamIdColumn, searchParamId).ParameterName);
+
+            return context;
+        }
+
+        public override SqlQueryGenerator VisitMultiary(MultiaryExpression expression, SqlQueryGenerator context)
+        {
+            if (expression.MultiaryOperation == MultiaryOperator.Or)
+            {
+                context.StringBuilder.Append('(');
```
cunninghamjc: Nit: single quotes here but double quotes below (for the closing paren).  Make them consistent?  Same in SqlQueryGenerator.cs

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/SqlFieldName.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    public static class SqlFieldName
+    {
+        public const FieldName ResourceSurrogateId = (FieldName)100;
+        public const FieldName LastUpdated = (FieldName)101;
+        public const FieldName TextOverflow = (FieldName)102;
+        public const FieldName NumberLow = (FieldName)103;
+        public const FieldName NumberHigh = (FieldName)104;
+        public const FieldName QuantityLow = (FieldName)105;
+        public const FieldName QuantityHigh = (FieldName)108;
```
jackliums: nit: just curious was this intentionally 108 skipping 106 & 107?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlQueryParameterManager.cs
```diff
@@ -0,0 +1,64 @@
...
+
+                _lookup.Add(key, parameter);
+            }
+
+            return parameter;
+        }
+
+        private string NextParameterName() => $"@p{_parameters.Count}";
+
+        public SqlParameter AddParameter(object value)
```
cunninghamjc: Is this used?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/TableExpressionKind.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// </summary>
+        NotExists,
+
+        /// <summary>
+        /// Represents a table expression that yields all possible resource IDs
+        /// </summary>
+        All,
+
+        /// <summary>
+        /// Represents a table expression that applies a TOP operator it its predecessor.
```
jackliums: nit: in?

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/BundleFactory.cs
johnstairs: We should be returning return relative paths here (and we do not store it in SQL)

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Expressions/Parsers/SearchParameterExpressionParser.cs
johnstairs: See earlier thread:  https://github.com/microsoft/fhir-server/pull/520#discussion_r290797558

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlQueryParameterManager.cs
johnstairs: It will to implement parameterized TOP :)

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/ConcatenationRewriter.cs
```diff
@@ -0,0 +1,76 @@
...
+            List<TableExpression> newTableExpressions = null;
+            for (var i = 0; i < expression.TableExpressions.Count; i++)
+            {
+                TableExpression tableExpression = expression.TableExpressions[i];
+
+                if (tableExpression.NormalizedPredicate.AcceptVisitor(_scout, null))
+                {
+                    if (newTableExpressions == null)
+                    {
+                        newTableExpressions = new List<TableExpression>();
```
jackliums: nit: `newTableExpressions = expression.TableExpression.Take(i).ToList()`

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/ConcatenationRewriter.cs
```diff
@@ -0,0 +1,76 @@
...
+            {
+                return expression;
+            }
+
+            return new SqlRootExpression(newTableExpressions, expression.DenormalizedExpressions);
+        }
+
+        public override Expression VisitTable(TableExpression tableExpression, object context)
+        {
+            var normalizedPredicate = tableExpression.NormalizedPredicate.AcceptVisitor(this, context);
```
jackliums: this doesn't need to check if the instance is the same or not?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/DateTimeRangeRewriter.cs
```diff
@@ -0,0 +1,86 @@
...
+    /// (And (FieldGreaterThanOrEqual DateTimeStart x) (FieldLessThanOrEqual DateTimeEnd y) (FieldLessThanOrEqual DateTimeStart y)).
+    /// This rewriting constrains the range scan over the index (DateTimeStart, DateTimeEnd).
+    /// </summary>
+    internal class DateTimeRangeRewriter : ExpressionRewriterWithDefaultInitialContext<object>
+    {
+        internal static readonly DateTimeRangeRewriter Instance = new DateTimeRangeRewriter();
+
+        public override Expression VisitSearchParameter(SearchParameterExpression expression, object context)
+        {
+            if ((expression.Parameter.Type == SearchParamType.Date
```
jackliums: I think some comment would help here why this rule exists. Is it because the start/end of LastUpdated will always be the same and therefore don't need the rewrite? Also, is composite here because one of the component of composite could be date?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/ConcatenationRewriter.cs
johnstairs: This is extracted out to a method in #527.

src/Microsoft.Health.Fhir.Core/Models/SearchParameterInfo.cs
brendankowitz: There are a few tests that do this. The original model didn't validate against being empty

src/Microsoft.Health.Fhir.SqlServer.UnitTests/Features/Search/Expressions/FlatteningRewriterTests.cs
```diff
@@ -0,0 +1,58 @@
...
+namespace Microsoft.Health.Fhir.SqlServer.UnitTests.Features.Search.Expressions
+{
+    public class FlatteningRewriterTests
+    {
+        [Fact]
+        public void GivenAMultiaryExpressionWithASingleElement_WhenFlattened_RemovesTheMultiary()
+        {
+            MultiaryExpression inputExpression = Expression.And(Expression.Equals(FieldName.Number, null, 1));
+            Expression visitedExpression = inputExpression.AcceptVisitor(FlatteningRewriter.Instance);
+            Assert.Equal("(FieldEqual Number 1)", visitedExpression.ToString());
```
jackliums: When we have time later, should we consider validating the actual output expressions instead of string since that removes the dependencies on the ToString implementation. (There are bunch of helpers in `SearchExpressionTestHelper` to help validating expression programmatically).

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/ConcatenationRewriter.cs
johnstairs: We only visit this if the scout says it needs to be rewritten. I'll add a debug assert.

src/Microsoft.Health.Fhir.SqlServer.UnitTests/Features/Search/Expressions/FlatteningRewriterTests.cs
johnstairs: We can discuss, in my opinion the test code is clearer when you compare strings rather than procedurally comparing object trees.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/DateTimeRangeRewriter.cs
johnstairs: Correct. Adding commments.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/SqlFieldName.cs
johnstairs: Was not intentional

PR #521, opened 2019-06-04
Enable R4 E2E tests

brandonpollett:
## Description
Enables R4 E2E tests.

## Related issues
Addresses [AB#69062](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/69062).

## Testing
E2E tests are enabled for R3 and R4.


brandonpollett:
There are a few items to still be addressed:

- [ ] Add support for running E2E in cloud via Environment Variables
- [ ] Look at composite search tests that need to be updated for R4
- [ ] Cleanup of code/de-dupe
- [ ] Look at enabling R4 for Smart tests.

Microsoft.Health.Fhir.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.28729.10
+# Visual Studio 15
+VisualStudioVersion = 15.0.28307.645
```
brandonpollett: Revert downgrade.

src/Microsoft.Health.Fhir.Tests.Common/ICustomFhirClient.cs
```diff
@@ -0,0 +1,154 @@
...
+
+        Task<string> GetBearerToken(TestApplication clientApplication, TestUser user);
+
+        List<KeyValuePair<string, string>> GetAppSecuritySettings(TestApplication clientApplication);
+
+        List<KeyValuePair<string, string>> GetUserSecuritySettings(TestApplication clientApplication, TestUser user);
+
+        Task GetSecuritySettings(string fhirServerMetadataUrl);
+
+        ResourceElement GetDefaultObservation();
```
brandonpollett: Extract these from FhirClient to some sort of resource element extensions/provider.

src/Microsoft.Health.Fhir.Core/AssemblyInfo.cs
```diff
@@ -9,3 +9,5 @@
 [assembly: NeutralResourcesLanguage("en-us")]
 [assembly: InternalsVisibleTo("Microsoft.Health.Fhir.R4.Api.UnitTests")]
 [assembly: InternalsVisibleTo("Microsoft.Health.Fhir.Stu3.Api.UnitTests")]
+[assembly: InternalsVisibleTo("Microsoft.Health.Fhir.Tests.E2E.R4")]
```
jackliums: I don't think the E2E assemblies should have dependency on Core assembly (I haven't seen what the exact dependencies are). I think the E2E tests should be performed purely through HTTP layer and shouldn't need to know the internals of the core implementation.

src/Microsoft.Health.Fhir.Tests.Common/EmbeddedResourceManager.cs
```diff
@@ -13,7 +13,12 @@ public static class EmbeddedResourceManager
     {
         public static string GetStringContent(string embeddedResourceSubNamespace, string fileName, string extension)
         {
-            string resourceName = $"{typeof(EmbeddedResourceManager).Namespace}.{embeddedResourceSubNamespace}.{ModelInfoProvider.Version}.{fileName}.{extension}";
+            return GetStringContent(embeddedResourceSubNamespace, fileName, extension, ModelInfoProvider.Version.ToString());
```
jackliums: Since now this class has a dependency of ModelInfoProvider, I think we should use IoC container to inject this instance and remove the depdencies on the static instance (which I think is the direction we want to go eventually so might as well start cleaning it.)

src/Microsoft.Health.Fhir.Tests.Common/FhirException.cs
```diff
@@ -4,19 +4,18 @@
...
 using System;
-using System.Linq;
 using System.Net;
 using System.Net.Http;
 using System.Net.Http.Headers;
-using Hl7.Fhir.Model;
+using Microsoft.Health.Fhir.Core.Models;
 
-namespace Microsoft.Health.Fhir.Tests.E2E.Common
+namespace Microsoft.Health.Fhir.Tests.Common
```
jackliums: I think for E2E test, I think we should have STU3 and R4 projects with version specific tests implemented that and have them also include a shared project which hosts all of the common infrastructure and tests.

Most of the client will use something like the Firely .NET library and deserialize/deserialize resources directly and use HttpClient so I think it's beneficial to keep the test mimic that as opposed to use something like ResourceElement and introduce another layer that does conversions. 

What do you think?

src/Microsoft.Health.Fhir.Tests.Common/Microsoft.Health.Fhir.Tests.Common.csproj
```diff
@@ -86,12 +98,24 @@
     <EmbeddedResource Include="TestFiles\R4\AuthConfigWithValidRoles.json" />
     <EmbeddedResource Include="TestFiles\R4\BasicExampleNarrative.json" />
     <EmbeddedResource Include="TestFiles\R4\BasicExampleNarrative.xml" />
+    <EmbeddedResource Include="TestFiles\R4\Condition-For-Patient-f001.json" />
```
jackliums: Should these be moved out of common and put into R4 and Stu3 projects respectively?

src/Microsoft.Health.Fhir.Tests.Common/FhirException.cs
brandonpollett: The point of the current test structure is to make use of the custom fixture `[HttpIntegrationFixtureArgumentSets(DataStore.CosmosDb, Format.Json, FhirVersion.All)]`. Since the FhirVersion is essentially another vector of the type of test, I felt the approach in this PR make more sense.

src/Microsoft.Health.Fhir.Core/AssemblyInfo.cs
brandonpollett: The dependency this enables is for some constants used in the smart tests. 

Additionally, there is a dependency for `ResourceElement`, but that's publicly exposed in core.

src/Microsoft.Health.Fhir.Core/Extensions/IntersectExtensions.cs
```diff
@@ -71,7 +76,7 @@ internal static class IntersectExtensions
                 issues.Add(string.Format(CultureInfo.InvariantCulture, Core.Resources.InvalidListConfigSetting, fieldName));
             }
 
-            if (config.Select(selector).GroupBy(x => x).Any(x => x.Count() > 1))
+            if (config.Select(selectorDistinct).GroupBy(x => x).Any(x => x.Count() > 1))
```
brendankowitz: I ended up reverting this change from my pr. 
It really depends if having the same search parameter listed with different 'profiles' is valid or not? I know it used to break api management

src/Microsoft.Health.Fhir.R4.Core/Features/Conformance/BaseCapabilities.json
```diff
@@ -15,7 +15,7 @@
     "kind": [ "capability" ],
     "fhirVersion": "4.0.0",
     "format": [
-        "json"
+        "application/fhir+json"
```
brendankowitz: Are these json files now up to date with master?

src/Microsoft.Health.Fhir.R4.Core/Features/Definition/search-parameters.json
```diff
@@ -24263,11 +24263,11 @@
       "modifier" : ["missing"],
       "component" : [{
         "definition" : "http://hl7.org/fhir/SearchParameter/DocumentReference-relatesto",
-        "expression" : "code"
+        "expression" : "target"
```
brendankowitz: @jackliums we should log this as a bug with FHIR

src/Microsoft.Health.Fhir.Tests.Common/EmbeddedResourceManager.cs
brendankowitz: Agree, I'd accept this in a followup PR also though. I'll make a task for cleaning up the static ModelInfoProvider on a broader level

src/Microsoft.Health.Fhir.Tests.Common/FhirException.cs
brendankowitz: We can further review/get feedback, for now I like the `HttpIntegrationFixtureArgumentSets` approach also

src/Microsoft.Health.Fhir.Tests.Common/IVersionSpecificFhirClient.cs
```diff
@@ -0,0 +1,141 @@
...
+
+        Task<string> GetBearerToken(TestApplication clientApplication, TestUser user);
+
+        List<KeyValuePair<string, string>> GetAppSecuritySettings(TestApplication clientApplication);
+
+        List<KeyValuePair<string, string>> GetUserSecuritySettings(TestApplication clientApplication, TestUser user);
+
+        Task GetSecuritySettings(string fhirServerMetadataUrl);
+
+        ResourceElement GetDefaultObservation();
```
brendankowitz: It would be good to put these on a different interface, i.e. `IFhirTestResources`or something.

src/Microsoft.Health.Fhir.Tests.Common/Microsoft.Health.Fhir.Tests.Common.csproj
brendankowitz: We could possibly do this in another pass. At least in `Tests.Common` they are only compiled/embedded once. This assembly is shared with all test projects

src/Microsoft.Health.Fhir.Tests.Common/Microsoft.Health.Fhir.Tests.Common.csproj
```diff
@@ -139,6 +175,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.Health.Abstractions\Microsoft.Health.Abstractions.csproj" />
+    <ProjectReference Include="..\Microsoft.Health.Fhir.Api\Microsoft.Health.Fhir.Api.csproj" />
```
brendankowitz: Is this ref used by this project?

test/Microsoft.Health.Fhir.Tests.E2E/Common/KnownResourceTypes.cs
```diff
@@ -0,0 +1,32 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Fhir.Tests.E2E.Common
+{
+    public static class KnownResourceTypes
```
brendankowitz: Would this be useful in ValueSets or Core? I know @feordin was using a known type

src/Microsoft.Health.Fhir.Tests.Common/EmbeddedResourceManager.cs
jackliums: Oh wait. Nevermind. This is for test.

src/Microsoft.Health.Fhir.Tests.Common/FhirException.cs
jackliums: What I worry about is the readability of the code and maintenanbility of the test and the productivity of the dev who has to write new tests.

1. It moves away from the resource class so updating resource require new extension method to be written (if it doesn't already exist) and then update to FhirClient to be able to set these properties.
2. The logic is now located spread across different places (ModelExtensions, FhirClient, and the tests themselves).
3. Everything is ResourceElement now which is a bit hard to read. We would have to rely on the variable name to be able to tell what resource it's manipulating.
4. Accessing the properties of the resource becomes a bit hard since it has to rely on fhir path.

For E2E tests, we almost for certain will have to manipulate the resource directly to create objects with certain values. While it makes sense to consolidate into ResourceElement for the server code, I don't know if it makes that much sense for the E2E tests.

src/Microsoft.Health.Fhir.Tests.Common/FhirResponse{T}.cs
```diff
@@ -4,12 +4,10 @@
 // -------------------------------------------------------------------------------------------------
 
 using System.Net.Http;
-using Hl7.Fhir.Model;
 
-namespace Microsoft.Health.Fhir.Tests.E2E.Common
+namespace Microsoft.Health.Fhir.Tests.Common
```
jackliums: should T be ResourceElement at least?

test/Microsoft.Health.Fhir.Tests.E2E.Shared/FhirClient.cs
```diff
@@ -0,0 +1,565 @@
...
+            var parser = new Hl7.Fhir.Serialization.FhirJsonParser();
+            return parser.Parse<Resource>(json).ToResourceElement();
+        }
+
+        public void Validate(ResourceElement resourceElement)
+        {
+            DotNetAttributeValidation.Validate(resourceElement.ToPoco(), true);
+        }
+
+        public ResourceElement UpdateId(ResourceElement resourceElement, string id)
```
jackliums: nit: maybe make this file partial and separate the implementation of the "operations" and "test extensions"?

test/Microsoft.Health.Fhir.Tests.E2E.Stu3/AuthenticationSettings.cs
```diff
@@ -0,0 +1,20 @@
...
+using static Microsoft.Health.Fhir.Tests.Common.EnvironmentVariables;
+
+namespace Microsoft.Health.Fhir.Tests.E2E.Common
+{
+    /// <summary>
+    /// Authentication Settings
+    /// </summary>
+    public static class AuthenticationSettings
+    {
+        public static string Scope => GetEnvironmentVariableWithDefault("Scope", DevelopmentIdentityProviderConfiguration.Audience);
```
jackliums: This file and the `AuthenticationSettings.cs` in Stu3 seems to be identical?

test/Microsoft.Health.Fhir.Tests.E2E/Crucible/CrucibleDataSource.cs
```diff
@@ -7,12 +7,13 @@
...
 using System.Net.Http;
 using System.Threading.Tasks;
-using Hl7.Fhir.Rest;
+using Microsoft.Health.Fhir.Tests.Common.FixtureParameters;
 using Microsoft.Health.Fhir.Tests.E2E.Crucible.Client;
-using FhirClient = Microsoft.Health.Fhir.Tests.E2E.Common.FhirClient;
 
 namespace Microsoft.Health.Fhir.Tests.E2E.Crucible
 {
+    extern alias E2EStu3;
```
jackliums: Does crucible tests need to be decorated with the FhirVersion.Stu3 so it only runs for STU3?

test/Microsoft.Health.Fhir.Tests.E2E/Rest/Audit/AuditTests.cs
```diff
@@ -411,11 +407,11 @@ private async Task ExecuteAndValidate(Func<FhirClient, Task> clientSetup, HttpSt
             const string url = "Patient/123";
 
             // Create a new client with no token supplied.
-            var client = new FhirClient(_fixture.CreateHttpClient(), ResourceFormat.Json);
+            var client = (IVersionSpecificFhirClient)Activator.CreateInstance(_client.GetType(), _fixture.CreateHttpClient(), Format.Json, _fixture.FhirVersion);
```
jackliums: Maybe add a method to FhirClient (something like `SetupAnonymous` or something) to just reuse the same instance of the client but allow the token to be temporarily removed for a given call?

test/Microsoft.Health.Fhir.Tests.E2E.Shared/FhirClient.cs
```diff
@@ -0,0 +1,565 @@
...
+
+                    SecuritySettings = (true, authorizeUrl, tokenUrl);
+                    return;
+                }
+            }
+
+            SecuritySettings = (false, null, null);
+        }
+
+        public ResourceElement GetDefaultObservation()
```
jackliums: This feels a little odd (where the function is provided).

brandonpollett:
Close in favor of #531

PR #527, opened 2019-06-05
Combine search parameters where possible and optimize date range queries

johnstairs:
## Description
Adds a couple of visitors that perform some optimizations. The first is to combine expressions across the same search parameter under one `SearchParameterExpression`. So for the search `RiskAssessment?probability=gt0.8&probability=lt0.9` we can do a single range query over `NumberSearchParam` and do not have to do a join.

Building on that, the next is to rewrite date range queries. The search query `Procedure?date=ge2010-05-12&date=le2010-05-20` can be quite expensive because it will end up as `(And (FieldGreaterThan DateTimeEnd '2010-05-12') (FieldLessThan DateTimeStart '2010-05-20'))`. Note that the LHS and RHS predicates are over different fields.

One way we can optimize this is to notice that the vast majority of datetime ranges in the database will be less than a day. This means that for most of the data, we can apply a fixed range on DateTimeStart, knowing, in this example, that it will always be greater than 2010-05-11 in addition to always being less than 2010-05-20. We can create a filtered nonclustered index where the range is greater than one day. Over this index, we'll do the original query, but it will be on a much smaller set of row.

For this example, the SQL looks like:
``` SQL
DECLARE @p0 SmallInt = 89;
DECLARE @p1 SmallInt = 182;
DECLARE @p2 DateTime2 = '5/12/2010 12:00:00 AM';
DECLARE @p3 DateTime2 = '5/20/2010 11:59:59 PM';
DECLARE @p4 DateTime2 = '5/11/2010 12:00:00 AM';

WITH cte0 AS
(
    FROM dbo.DateTimeSearchParam
    WHERE IsHistory = 0
        AND ResourceTypeId = @p0
        AND SearchParamId = @p1
        AND IsLongerThanADay = 1
        AND EndDateTime >= @p2
        AND StartDateTime <= @p3

),
cte1 AS
(
    SELECT Sid1 FROM cte0
    UNION ALL
    FROM dbo.DateTimeSearchParam
    WHERE IsHistory = 0
        AND ResourceTypeId = @p0
        AND SearchParamId = @p1
        AND IsLongerThanADay = 0
        AND EndDateTime >= @p2
        AND StartDateTime >= @p4
        AND StartDateTime <= @p3

),
cte2 AS
(
    SELECT DISTINCT TOP 11 Sid1
    FROM cte1
    ORDER BY Sid1 ASC
)
SELECT r.ResourceTypeId, r.ResourceId, r.Version, r.IsDeleted, r.ResourceSurrogateId, r.RequestMethod, r.RawResource
FROM dbo.Resource r
WHERE r.ResourceSurrogateId IN (SELECT Sid1 FROM cte2)
```

## Related issues
Addresses [AB#69346](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/69346)


src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/BinaryOperator.cs
```diff
@@ -10,11 +10,11 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Expressions
...
     public enum BinaryOperator
     {
-        Equal,
-        GreaterThan,
-        GreaterThanOrEqual,
-        LessThan,
-        LessThanOrEqual,
-        NotEqual,
+        Equal = 0,
+        GreaterThan = 1,
```
cunninghamjc: We may want to add a comment on the dependency between these enum values and DateTimeBoundedRangeRewriter.

PR #528, opened 2019-06-05
Azure Export Destination Client

namadabu:
## Description
MVP of an Azure Export Destination Client that can export data to an Azure storage account.

## Related issues
Addresses [[AB#69124](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/69124)].

## Testing
Tested manually by using Postman and targeting a personal azure storage account for exporting data.


src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportDestinationClient/AzureExportDestinationClient.cs
```diff
@@ -0,0 +1,104 @@
...
+        private Dictionary<Uri, CloudBlockBlobWrapper> _uriToBlobMapping = new Dictionary<Uri, CloudBlockBlobWrapper>();
+        private Dictionary<(Uri FileUri, uint PartId), Stream> _streamMappings = new Dictionary<(Uri FileUri, uint PartId), Stream>();
+
+        public string DestinationType => "azure-block-blob";
+
+        public async Task ConnectAsync(string connectionSettings, CancellationToken cancellationToken)
+        {
+            EnsureArg.IsNotNullOrWhiteSpace(connectionSettings, nameof(connectionSettings));
+
+            CloudStorageAccount cloudAccount = null;
```
jackliums: nit: these two can be combined into one: `if (!CloudStorageAccount.TryParse(connectionSettings, out CloudStorageAccount cloudAccount))`

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportDestinationClient/AzureExportDestinationClient.cs
```diff
@@ -0,0 +1,104 @@
...
+            if (!CloudStorageAccount.TryParse(connectionSettings, out cloudAccount))
+            {
+                throw new CantConnectToDestinationException();
+            }
+
+            _blobClient = cloudAccount.CreateCloudBlobClient();
+
+            // We will need to accept a container name/reference instead of using root container.
```
jackliums: is this a todo?

src/Microsoft.Health.Fhir.Shared.Api/Modules/OperationsModule.cs
```diff
@@ -28,7 +28,11 @@ public void Load(IServiceCollection services)
                 .Transient()
                 .AsSelf();
 
-            services.Add<IExportDestinationClient>(sp => sp.GetRequiredService<InMemoryExportDestinationClient>())
+            services.Add<AzureExportDestinationClient>()
+                .Transient()
+                .AsSelf();
+
+            services.Add<IExportDestinationClient>(sp => sp.GetRequiredService<AzureExportDestinationClient>())
```
namadabu: Add the InMemoryDestinationClient also.

src/Microsoft.Health.Fhir.Stu3.Web/appsettings.json
```diff
@@ -4,7 +4,7 @@
       "UseStrictConformance": true
     },
     "Security": {
-      "Enabled": true,
+      "Enabled": false,
```
namadabu: Revert all the changes in this file before check-in

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportDestinationClient/AzureExportDestinationClient.cs
```diff
@@ -0,0 +1,104 @@
...
+            return Task.FromResult(blockBlob.Uri);
+        }
+
+        public async Task WriteFilePartAsync(Uri fileUri, uint partId, byte[] bytes, CancellationToken cancellationToken)
+        {
+            EnsureArg.IsNotNull(fileUri, nameof(fileUri));
+            EnsureArg.IsNotNull(bytes, nameof(bytes));
+            CheckIfClientIsConnected();
+
+            var key = (fileUri, partId);
```
jackliums: Now that I think about it, the key just need to be fileUri since commit will commit all of the parts and remove the mapping.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportDestinationClient/AzureExportDestinationClient.cs
```diff
@@ -0,0 +1,104 @@
...
+            {
+                Stream stream = mapping.Value;
+
+                // Reset the position.
+                stream.Position = 0;
+
+                CloudBlockBlobWrapper blobWrapper = _uriToBlobMapping[mapping.Key.Item1];
+
+                var blockId = Convert.ToBase64String(Encoding.ASCII.GetBytes(mapping.Key.Item2.ToString("d6")));
+                await blobWrapper.UploadBlockAsync(blockId, mapping.Value, md5Hash: null, cancellationToken);
```
jackliums: You could do this step in parallel to increase throughput.

src/Microsoft.Health.Fhir.Stu3.Web/appsettings.json
jackliums: Shouldn't check-in this change.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportDestinationClient/AzureExportDestinationClient.cs
```diff
@@ -0,0 +1,104 @@
...
+        private CloudBlobContainer _blobContainer = null;
+
+        private Dictionary<Uri, CloudBlockBlobWrapper> _uriToBlobMapping = new Dictionary<Uri, CloudBlockBlobWrapper>();
+        private Dictionary<(Uri FileUri, uint PartId), Stream> _streamMappings = new Dictionary<(Uri FileUri, uint PartId), Stream>();
+
+        public string DestinationType => "azure-block-blob";
+
+        public async Task ConnectAsync(string connectionSettings, CancellationToken cancellationToken)
+        {
+            EnsureArg.IsNotNullOrWhiteSpace(connectionSettings, nameof(connectionSettings));
```
jackliums: I think we said connectionSettings will be base64 encoded because it might contain characters valid for URL.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportDestinationClient/AzureExportDestinationClient.cs
```diff
@@ -0,0 +1,104 @@
...
+
+            CloudStorageAccount cloudAccount = null;
+            if (!CloudStorageAccount.TryParse(connectionSettings, out cloudAccount))
+            {
+                throw new CantConnectToDestinationException();
```
jackliums: Maybe just a generic `DestinationConnectionException` which can be used when connection settings is not valid or when the write operation fails?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportDestinationClient/AzureExportDestinationClient.cs
```diff
@@ -0,0 +1,104 @@
...
+                Stream stream = mapping.Value;
+
+                // Reset the position.
+                stream.Position = 0;
+
+                CloudBlockBlobWrapper blobWrapper = _uriToBlobMapping[mapping.Key.Item1];
+
+                var blockId = Convert.ToBase64String(Encoding.ASCII.GetBytes(mapping.Key.Item2.ToString("d6")));
+                await blobWrapper.UploadBlockAsync(blockId, mapping.Value, md5Hash: null, cancellationToken);
+
```
jackliums: If these operation throw exceptions, it should be wrapped in common exception so the core logic knows how to deal with it. (e.g., connection issue that can be retried vs authentication issue which might not be retriable.)

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportDestinationClient/CloudBlockBlobWrapper.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        public CloudBlockBlobWrapper(CloudBlockBlob blockBlob)
+        {
+            EnsureArg.IsNotNull(blockBlob, nameof(blockBlob));
+
+            _cloudBlob = blockBlob;
+
+            // For now we assume that the blockblob we are using is always new. If it is an existing one,
+            // we will have to get the existing list of block ids.
+            _existingBlockIds = new HashSet<string>();
```
jackliums: I don't think you can use HashSet here since the order of the blocks that's been committed matters.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportDestinationClient/AzureExportDestinationClient.cs
```diff
@@ -0,0 +1,104 @@
...
+            foreach (KeyValuePair<(Uri, uint), Stream> mapping in _streamMappings)
+            {
+                Stream stream = mapping.Value;
+
+                // Reset the position.
+                stream.Position = 0;
+
+                CloudBlockBlobWrapper blobWrapper = _uriToBlobMapping[mapping.Key.Item1];
+
+                var blockId = Convert.ToBase64String(Encoding.ASCII.GetBytes(mapping.Key.Item2.ToString("d6")));
```
jackliums: Now i am wondering if there is any point base64 encoding this. I forgot why I added that step but `mapping.Key.item2.ToString("d6")` already outputs string like '0000001' so it doesn't seem like the base64 encoding is necessary.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportDestinationClient/AzureExportDestinationClient.cs
```diff
@@ -0,0 +1,104 @@
...
+            CloudStorageAccount cloudAccount = null;
+            if (!CloudStorageAccount.TryParse(connectionSettings, out cloudAccount))
+            {
+                throw new CantConnectToDestinationException();
+            }
+
+            _blobClient = cloudAccount.CreateCloudBlobClient();
+
+            // We will need to accept a container name/reference instead of using root container.
+            _blobContainer = _blobClient.GetRootContainerReference();
```
jackliums: Instead of root container, should we consider create a container with the job id?

namadabu: Good idea. Updated the call to take in a `containerId`

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportDestinationClient/AzureExportDestinationClient.cs
namadabu: I am going to leave it like this for now since we need the partId to generate the blockId. I will refactor it later.

src/Microsoft.Health.Fhir.Core/Microsoft.Health.Fhir.Core.csproj
```diff
@@ -9,6 +9,7 @@
     <PackageReference Include="Ensure.That" Version="8.1.1" />
     <PackageReference Include="FluentValidation" Version="8.1.3" />
     <PackageReference Include="MediatR" Version="6.0.0" />
+    <PackageReference Include="Microsoft.Azure.Storage.Blob" Version="10.0.3" />
```
jackliums: This should not be in Core. You should create a separate assembly and have the Azure dependencies in there.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportDestinationClient/AzureExportDestinationClient.cs
```diff
@@ -0,0 +1,126 @@
...
+        private Dictionary<Uri, CloudBlockBlobWrapper> _uriToBlobMapping = new Dictionary<Uri, CloudBlockBlobWrapper>();
+        private Dictionary<(Uri FileUri, uint PartId), Stream> _streamMappings = new Dictionary<(Uri FileUri, uint PartId), Stream>();
+
+        public string DestinationType => "azure-block-blob";
+
+        public async Task ConnectAsync(string connectionSettings, CancellationToken cancellationToken, string containerId = null)
+        {
+            EnsureArg.IsNotNullOrWhiteSpace(connectionSettings, nameof(connectionSettings));
+
+            string decodedConnectionString;
```
jackliums: nit: This could happen at the layer above since every client must implement the base64 decoding. It can be decoded before it's even stored in keyvault.

jackliums: (doesn't have to be in this PR.)

src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/AzureExportDestinationClient.cs
```diff
@@ -0,0 +1,127 @@
...
+                {
+                    await blobWrapper.UploadBlockAsync(blockId, mapping.Value, md5Hash: null, cancellationToken);
+                    await blobWrapper.CommitBlockListAsync(cancellationToken);
+                }));
+            }
+
+            await Task.WhenAll(uploadAndCommitTasks);
+
+            // We can clear the stream mappings once we commit everything in memory.
+            _streamMappings.Clear();
```
jackliums: Streams should be disposed.

src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/AzureExportDestinationClient.cs
```diff
@@ -0,0 +1,127 @@
...
+
+            await _blobContainer.CreateIfNotExistsAsync();
+        }
+
+        public Task<Uri> CreateFileAsync(string fileName, CancellationToken cancellationToken)
+        {
+            EnsureArg.IsNotNullOrWhiteSpace(fileName, nameof(fileName));
+            CheckIfClientIsConnected();
+
+            CloudBlockBlob blockBlob = _blobContainer.GetBlockBlobReference(fileName);
```
jackliums: Also set the content type of the blob `application/fhir+ndjson`

src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/AzureExportDestinationClient.cs
```diff
@@ -0,0 +1,133 @@
...
+                    await blobWrapper.CommitBlockListAsync(cancellationToken);
+                }));
+            }
+
+            await Task.WhenAll(uploadAndCommitTasks);
+
+            // We can clear the stream mappings once we commit everything in memory.
+            foreach (Stream memStream in _streamMappings.Values)
+            {
+                memStream.Dispose();
```
jackliums: why not dispose it inside the task above after you finish commiting?

src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/AzureExportDestinationClient.cs
```diff
@@ -0,0 +1,133 @@
...
+
+                // Reset the position.
+                stream.Position = 0;
+
+                CloudBlockBlobWrapper blobWrapper = _uriToBlobMapping[mapping.Key.Item1];
+
+                var blockId = Convert.ToBase64String(Encoding.ASCII.GetBytes(mapping.Key.Item2.ToString("d6")));
+                uploadAndCommitTasks.Add(Task.Run(async () =>
+                {
+                    await blobWrapper.UploadBlockAsync(blockId, mapping.Value, md5Hash: null, cancellationToken);
```
jackliums: nit: `mapping.Value` could just be `stream`.

src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/Registration/FhirServerBuilderExportDestinationClientRegistrationExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+using EnsureThat;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Health.Extensions.DependencyInjection;
+using Microsoft.Health.Fhir.Azure.ExportDestinationClient;
+using Microsoft.Health.Fhir.Core.Features.Operations.Export.ExportDestinationClient;
+using Microsoft.Health.Fhir.Core.Registration;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    public static class FhirServerBuilderExportDestinationClientRegistrationExtensions
```
jackliums: This file and `FhirServerBuilderSecretStoreRgistrationExtensions` could collapse into one class (you can still provide 2 different methods but you don't need 2 classes).

src/Microsoft.Health.Fhir.Azure/FhirServerBuilderAzureRegistrationExtensions.cs
```diff
@@ -41,12 +59,12 @@ public static IFhirServerBuilder AddKeyVaultSecretStore(this IFhirServerBuilder
...
-                        var tokenProvider = new AzureServiceTokenProvider();
-                        var kvClient = new KeyVaultClient(new KeyVaultClient.AuthenticationCallback(tokenProvider.KeyVaultTokenCallback));
+                {
+                    var tokenProvider = new AzureServiceTokenProvider();
+                    var kvClient = new KeyVaultClient(new KeyVaultClient.AuthenticationCallback(tokenProvider.KeyVaultTokenCallback));
 
-                        return new KeyVaultSecretStore(kvClient, new Uri(keyVaultConfig.Endpoint));
-                    })
+                    return new KeyVaultSecretStore(kvClient, new Uri(keyVaultConfig.Endpoint));
+                })
```
namadabu: nit: spacing seems a little weird here with .Singleton() starting with an extra tab compared to the previous line.

PR #530, opened 2019-06-05
Exclude the definition files from GeoPol tool.

jackliums:
## Description
Exclude the definition file from the GeoPol tool.


jackliums:
Exclude the definition file from the GeoPol tool.

PR #535, opened 2019-06-07
Local devbox DB naming and bootstrap improvements

johnstairs:
1. STU3 and R4 launchSettings.json profiles should use different databases
1. Create the database if it does not exist when SqlServer:Initialize config is true
1. Read launchSettings.json when configuring test server.


src/Microsoft.Health.Fhir.Stu3.Web/appsettings.json
```diff
@@ -4,7 +4,7 @@
       "UseStrictConformance": true
     },
     "Security": {
-      "Enabled": true,
+            "Enabled": false,
```
brandonpollett: Revert?

johnstairs: Yikes! Yes!

PR #537, opened 2019-06-09
Adds support for search parameter chaining with SQL

johnstairs:
## Description
Implements search parameter chaining. Does NOT include reverse changing (`_has`).

### Examples:

`/Claim?organization.name=falmouth%20walk%20in%20medical%20center`

`/Encounter?participant:Practitioner.name=Jane,Norm`

`/DiagnosticReport?result.code-value-quantity=http://loinc.org|787-2$83`

Multi-level chaining:

`/Claim?encounter.service-provider.name=falmouth%20walk%20in%20medical%20center`

Note that in the case where a reference can be to multiple types you must specify the target type. (we could lift that restriction later if we want). So you cannot specify `/Observation?subject.name=lisa`, instead you must say `/Observation?subject:Patient.name=lisa`

## Related issues
Addresses [AB#68734](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/68734)

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/IndentedStringBuilder.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/IndentedStringBuilder.cs
@@ -37,12 +37,23 @@ namespace Microsoft.Health.Fhir.SqlServer
         /// <returns>A scope to be disposed when the indentation level is to be restored</returns>
         internal IndentedScope Indent() => new IndentedScope(this);
 
-        [Obsolete("Use overload with table alias instead")]
+        /// <summary>
+        /// Appends a column name to this instance.
+        /// </summary>
+        /// <param name="column">The column</param>
+        /// <returns>This instance</returns>
+        [Obsolete("Use overload with table alias instead")] // Catch calls an raise compiler warnings.
         public IndentedStringBuilder Append(Column column)
         {
             return Append(column, null);
         }
 
+        /// <summary>
+        /// Appends a column name to this instance.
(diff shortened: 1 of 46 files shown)


src/Microsoft.Health.Fhir.SqlServer/IndentedStringBuilder.cs
```diff
@@ -36,6 +37,60 @@ internal int IndentLevel
         /// <returns>A scope to be disposed when the indentation level is to be restored</returns>
         internal IndentedScope Indent() => new IndentedScope(this);
 
+        /// <summary>
+        /// Appends a column name to this instance.
+        /// </summary>
+        /// <param name="column">The column</param>
+        /// <returns>This instance</returns>
+        [Obsolete("Use overload with table alias instead")] // Catch calls an raise compiler warnings.
```
cunninghamjc: Why not just remove this method?

johnstairs: The idea was to catch cases in the future where we forget to call the overload with the alias. If we remove this, calls to `Append(myColumn)` will resolve the the object `Append(Object)` overload. We don't need to keep this, I was thinking it could be helpful.

cunninghamjc: Got it, that makes sense.

PR #540, opened 2019-06-10
Added bulk export doc.

jackliums:
## Description
Added doc for brief description of the bulk export feature.

## Related issues
Addresses #346. Work item: [AB#69125](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/69125).

### DIFF
--- a/docs/BulkExport.md
+++ b/docs/BulkExport.md
@@ -1,0 +1,19 @@
+# Bulk Export
+
+This feature allows data from the FHIR server to be exported. More detail can be found in the [spec](https://github.com/HL7/bulk-data/blob/master/spec/export/index.md).
+
+The feature is currently under preview and is turned off by default. To enable the feature, update the `FhirServer:Operations:Export:Enabled` setting to be true.
+
+## Specifying destination
+
+One customization that we have done in addition to the spec is requiring the destination information to be supplied. The destination type can be specified by using `_destinationType` query parameter and any destination specific settings can be specified as BASE64 encoded string by using `_destinationConnectionSettings` query parameter.
+
+Currently, we only support Azure Blob storage is supported. This can be specified with `_destinationType=azure-block-blob` and `_destinationConnectionSettings` is the BASE64 encoding of the SAS token to the blob storage.
+
+Example of the input:
+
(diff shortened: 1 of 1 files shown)


docs/BulkExport.md
```diff
@@ -0,0 +1,19 @@
+# Bulk Export
+
+This feature allows data from the FHIR server to be exported. More detail can be found in the [spec](https://github.com/HL7/bulk-data/blob/master/spec/export/index.md).
+
+The feature is currently under preview and is turned off by default. To enable the feature, update the `FhirServer:Operations:Export:Enabled` setting to be true.
+
+## Specifying destination
+
+One customization that we have done in addition to the spec is requiring the destination information to be supplied. The destination type can be specified by using `_destinationType` query parameter and any destination specific settings can be specified as BASE64 encoded string by using `_destinationConnectionSettings` query parameter.
```
jmandel: Note this is a breaking change from the specification if it adds a new, required parameter. We should call it out as such (and ideally document that we plan to support a default destination to maintain compatibility).

jackliums: Sure I can make it more clear.

hansenms:
I think we need an example here. Starting with I have a storage account and I create a SAS token, etc. Then I structure my request by taking this `{...}` and base64 encoding, etc.

docs/BulkExport.md
```diff
@@ -0,0 +1,37 @@
...
+
+The feature is currently under preview and is turned off by default. To enable the feature, update the `FhirServer:Operations:Export:Enabled` setting to be true.
+
+## Specifying destination
+
+One customization that we have done in addition to the spec is _requiring_ the destination information to be supplied. This is a breaking change from the spec. We are planning to support default destination to maintain compatibility with spec.
+
+The destination type can be specified by using `_destinationType` query parameter and any destination specific settings can be specified as base64 encoded string by using `_destinationConnectionSettings` query parameter.
+
+Currently, we only support Azure Blob storage as the destination. This can be specified with `_destinationType=azure-block-blob` and `_destinationConnectionSettings=[base64 encoding of the connection string]`.
```
namadabu: I am not sure about the terminology here. But do we want to specifically call out the connection string should be for the storage account itself (and not, say, for a container/blob). Or is it already implied?

jackliums: I think it's implied. There is also an example of what the connection string looks like in the spec.

PR #541, opened 2019-06-12
Fixes incorrect query on string overflow

johnstairs:
## Description
The logic that accounts for string search parameter values being greater than the maximum we index, and falling into an overflow column, was incorrect.

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/StringOverflowRewriter.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/StringOverflowRewriter.cs
@@ -18,14 +18,33 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Search.Expressions.Visitors
     /// </summary>
     internal class StringOverflowRewriter : ConcatenationRewriter
     {
+        private readonly ISearchParameterDefinitionManager _searchParameterDefinitionManager;
+
         public StringOverflowRewriter(ISearchParameterDefinitionManager searchParameterDefinitionManager)
             : base(new Scout(searchParameterDefinitionManager))
         {
+            _searchParameterDefinitionManager = searchParameterDefinitionManager;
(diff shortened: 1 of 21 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/StringOverflowRewriter.cs
```diff
@@ -28,18 +31,45 @@ public override Expression VisitString(StringExpression expression, object conte
...
+                    {
+                        return false;
+                    }
+                }
+
                 switch (expression.StringOperator)
                 {
                     case StringOperator.Equals:
                     case StringOperator.NotStartsWith:
-                    case StringOperator.StartsWith:
```
cunninghamjc: Why remove StartsWith?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/StringOverflowRewriter.cs
```diff
@@ -14,42 +18,72 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Search.Expressions.Visitors
...
-        internal static readonly StringOverflowRewriter Instance = new StringOverflowRewriter();
+        private readonly ISearchParameterDefinitionManager _searchParameterDefinitionManager;
 
-        private StringOverflowRewriter()
-            : base(new Scout())
+        public StringOverflowRewriter(ISearchParameterDefinitionManager searchParameterDefinitionManager)
+            : base(new Scout(searchParameterDefinitionManager))
         {
+            _searchParameterDefinitionManager = searchParameterDefinitionManager;
+            EnsureArg.IsNotNull(searchParameterDefinitionManager, nameof(searchParameterDefinitionManager));
```
cunninghamjc: Nit: any reason for checking for null after assigning instead of before?

PR #543, opened 2019-06-17
Parsing changes to support reverse chaining

johnstairs:
We needed to modify search expression "key" parsing to handle reverse chaining. Parsing is working but creating the reverse chained expression is not implemented yet (nor is executing it, obviously).

Uses `ReadOnlySpan<char>` for tokenizing for efficiency. We can eventually spanify/memory-ify more if we choose to.

@brandonpollett, for you to run with from here :)


brandonpollett:
Closing and re-opening under a PR opened by me.

PR #544, opened 2019-06-18
Support ability to resume export

namadabu:
## Description
Currently we always start an export job from the beginning irrespective of what state it is in. This means we could end up duplicating work in cases where we had some data exported before we got interrupted.  This change adds code to allow resuming an export job so that we avoid duplicating work. 

## Related issues
Addresses [[AB#68703](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/68703)].

## Testing
Added UTs.
Manual E2E testing using azure storage account.

### DIFF
--- a/src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/AzureExportDestinationClient.cs
+++ b/src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/AzureExportDestinationClient.cs
@@ -6,6 +6,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
@@ -93,7 +94,9 @@ namespace Microsoft.Health.Fhir.Azure.ExportDestinationClient
         {
             CheckIfClientIsConnected();
 
-            var uploadAndCommitTasks = new List<Task>();
+            // Upload all blocks for each blob that was modified.
+            var uploadTasks = new List<Task>();
+            var wrappersToCommit = new List<CloudBlockBlobWrapper>();
             foreach (KeyValuePair<(Uri, uint), Stream> mapping in _streamMappings)
             {
                 Stream stream = mapping.Value;
@@ -104,24 +107,57 @@ namespace Microsoft.Health.Fhir.Azure.ExportDestinationClient
                 CloudBlockBlobWrapper blobWrapper = _uriToBlobMapping[mapping.Key.Item1];
 
                 var blockId = Convert.ToBase64String(Encoding.ASCII.GetBytes(mapping.Key.Item2.ToString("d6")));
-                uploadAndCommitTasks.Add(Task.Run(async () =>
-                {
-                    await blobWrapper.UploadBlockAsync(blockId, stream, md5Hash: null, cancellationToken);
-                    await blobWrapper.CommitBlockListAsync(cancellationToken);
 
-                    stream.Dispose();
-                }));
+                uploadTasks.Add(blobWrapper.UploadBlockAsync(blockId, stream, md5Hash: null, cancellationToken));
+                wrappersToCommit.Add(blobWrapper);
             }
 
-            await Task.WhenAll(uploadAndCommitTasks);
+            await Task.WhenAll(uploadTasks);
+
+            // Commit all the blobs that were uploaded.
+            var commitTasks = new List<Task>();
+            foreach (CloudBlockBlobWrapper wrapper in wrappersToCommit)
+            {
+                commitTasks.Add(wrapper.CommitBlockListAsync(cancellationToken));
+            }
+
+            await Task.WhenAll(commitTasks);
 
             // We can clear the stream mappings once we commit everything in memory.
+            foreach (Stream stream in _streamMappings.Values)
+            {
+                stream.Dispose();
+            }
+
             _streamMappings.Clear();
+        }
+
+        public async Task InitializeForResumingExportAsync(CancellationToken cancellationToken)
+        {
+            CheckIfClientIsConnected();
+
+            IEnumerable<IListBlobItem> blobs = _blobContainer.ListBlobs();
+
+            // For each blob, dowloand the committed block list and create a corresponding CloudBlockBlobWrapper.
+            foreach (IListBlobItem blobItem in blobs)
+            {
+                CloudBlockBlob blob = blobItem as CloudBlockBlob;
+                if (blob == null)
+                {
+                    throw new DestinationConnectionException(Resources.UnableToResumeExport);
+                }
+
+                // We are going to consider only committed blocks.
+                IEnumerable<ListBlockItem> blockIdlist = await blob.DownloadBlockListAsync();
+                var wrapper = new CloudBlockBlobWrapper(blob, blockIdlist.Where(x => x.Committed == true).Select(x => x.Name).ToList());
+
+                _uriToBlobMapping.Add(blob.Uri, wrapper);
+            }
         }
 
         private void CheckIfClientIsConnected()
         {
-            if (_blobClient == null)
+            if (_blobClient == null || _blobContainer == null)
             {
                 throw new DestinationConnectionException(Resources.DestinationClientNotConnected);
             }
--- a/src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/AzureExportDestinationClient.cs
+++ b/src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/AzureExportDestinationClient.cs
@@ -67,8 +67,12 @@ namespace Microsoft.Health.Fhir.Azure.ExportDestinationClient
             CheckIfClientIsConnected();
 
             CloudBlockBlob blockBlob = _blobContainer.GetBlockBlobReference(fileName);
-            blockBlob.Properties.ContentType = "application/fhir+ndjson";
-            _uriToBlobMapping.Add(blockBlob.Uri, new CloudBlockBlobWrapper(blockBlob));
+
+            if (!_uriToBlobMapping.ContainsKey(blockBlob.Uri))
+            {
+                blockBlob.Properties.ContentType = "application/fhir+ndjson";
+                _uriToBlobMapping.Add(blockBlob.Uri, new CloudBlockBlobWrapper(blockBlob));
+            }
 
             return Task.FromResult(blockBlob.Uri);
         }
--- a/src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/AzureExportDestinationClient.cs
+++ b/src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/AzureExportDestinationClient.cs
@@ -132,25 +132,24 @@ namespace Microsoft.Health.Fhir.Azure.ExportDestinationClient
             _streamMappings.Clear();
         }
 
-        public async Task InitializeForResumingExportAsync(CancellationToken cancellationToken)
+        public async Task OpenFilesAsync(IList<Uri> fileUris, CancellationToken cancellationToken)
         {
+            EnsureArg.IsNotNull(fileUris, nameof(fileUris));
             CheckIfClientIsConnected();
 
-            IEnumerable<IListBlobItem> blobs = _blobContainer.ListBlobs();
-
-            // For each blob, dowload the committed block list and create a corresponding CloudBlockBlobWrapper.
-            foreach (IListBlobItem blobItem in blobs)
+            foreach (Uri uri in fileUris)
             {
-                CloudBlockBlob blob = blobItem as CloudBlockBlob;
-                if (blob == null)
-                {
-                    // Not a block blob. Skip it.
-                    continue;
-                }
+                var blob = new CloudBlockBlob(uri, _blobClient.Credentials);
 
                 // We are going to consider only committed blocks.
-                IEnumerable<ListBlockItem> blockIdlist = await blob.DownloadBlockListAsync();
-                var wrapper = new CloudBlockBlobWrapper(blob, blockIdlist.Where(x => x.Committed == true).Select(x => x.Name).ToList());
+                IEnumerable<ListBlockItem> blockIds = await blob.DownloadBlockListAsync(
+                    BlockListingFilter.Committed,
+                    accessCondition: null,
+                    options: null,
+                    operationContext: null,
+                    cancellationToken);
+
+                var wrapper = new CloudBlockBlobWrapper(blob, blockIds.Select(x => x.Name).ToList());
 
                 _uriToBlobMapping.Add(blob.Uri, wrapper);
             }
--- a/src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/AzureExportDestinationClient.cs
+++ b/src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/AzureExportDestinationClient.cs
@@ -13,6 +13,7 @@ using System.Threading.Tasks;
 using EnsureThat;
 using Microsoft.Azure.Storage;
 using Microsoft.Azure.Storage.Blob;
+using Microsoft.Extensions.Logging;
 using Microsoft.Health.Fhir.Core.Features.Operations.Export.ExportDestinationClient;
 
 namespace Microsoft.Health.Fhir.Azure.ExportDestinationClient
@@ -24,6 +25,15 @@ namespace Microsoft.Health.Fhir.Azure.ExportDestinationClient
 
         private Dictionary<Uri, CloudBlockBlobWrapper> _uriToBlobMapping = new Dictionary<Uri, CloudBlockBlobWrapper>();
         private Dictionary<(Uri FileUri, uint PartId), Stream> _streamMappings = new Dictionary<(Uri FileUri, uint PartId), Stream>();
+
+        private readonly ILogger _logger;
+
+        public AzureExportDestinationClient(ILogger<AzureExportDestinationClient> logger)
+        {
+            EnsureArg.IsNotNull(logger, nameof(logger));
+
+            _logger = logger;
+        }
 
         public string DestinationType => "azure-block-blob";
 
@@ -100,8 +110,9 @@ namespace Microsoft.Health.Fhir.Azure.ExportDestinationClient
 
             // Upload all blocks for each blob that was modified.
             Task[] uploadTasks = new Task[_streamMappings.Count];
+            CloudBlockBlobWrapper[] wrappersToCommit = new CloudBlockBlobWrapper[_streamMappings.Count];
+
             int index = 0;
-            var wrappersToCommit = new List<CloudBlockBlobWrapper>();
             foreach (KeyValuePair<(Uri, uint), Stream> mapping in _streamMappings)
             {
                 // Reset stream position.
@@ -112,9 +123,8 @@ namespace Microsoft.Health.Fhir.Azure.ExportDestinationClient
                 var blockId = Convert.ToBase64String(Encoding.ASCII.GetBytes(mapping.Key.Item2.ToString("d6")));
 
                 uploadTasks[index] = blobWrapper.UploadBlockAsync(blockId, stream, md5Hash: null, cancellationToken);
+                wrappersToCommit[index] = blobWrapper;
                 index++;
-
-                wrappersToCommit.Add(blobWrapper);
             }
 
             await Task.WhenAll(uploadTasks);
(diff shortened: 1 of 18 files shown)


src/Microsoft.Health.Fhir.Core/Features/Operations/Export/Models/ExportJobRecord.cs
```diff
@@ -66,7 +66,7 @@ protected ExportJobRecord()
         public int SchemaVersion { get; private set; }
 
         [JsonProperty(JobRecordProperties.Output)]
-        public List<ExportFileInfo> Output { get; private set; } = new List<ExportFileInfo>();
+        public Dictionary<string, ExportFileInfo> Output { get; private set; } = new Dictionary<string, ExportFileInfo>();
```
jackliums: Prefer expose interface rather than concrete types (both here and below).

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -77,26 +76,34 @@ public async Task ExecuteAsync(ExportJobRecord exportJobRecord, WeakETag weakETa
...
 
-                // Connect to the destination using appropriate client.
-                _exportDestinationClient = _exportDestinationClientFactory.Create(destinationInfo.DestinationType);
-
-                await _exportDestinationClient.ConnectAsync(destinationInfo.DestinationConnectionString, cancellationToken, _exportJobRecord.Id);
+                // If we are resuming a job, we can detect that by checking the progress info from the job record.
+                // If it is null, then we know we are processing a new job.
+                if (_exportJobRecord.Progress != null)
+                {
+                    // Update our internal mapping with list of files that we have already committed.
```
jackliums: isn't the _exportJobRecord.Output and _resourceTypeToFileInforMapping the same thing?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -77,26 +76,34 @@ public async Task ExecuteAsync(ExportJobRecord exportJobRecord, WeakETag weakETa
...
+                }
+                else
+                {
+                    _exportJobRecord.Progress = new ExportJobProgress(continuationToken: null, page: 0);
+                }
 
                 ExportJobProgress progress = _exportJobRecord.Progress;
 
-                // Current page will be used to organize a set of search results into a group so that they can be committed together.
+                // Current page will be used to organize a set of search results into a batch so that they can be committed together.
```
jackliums: page should be changed to batch.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -77,26 +76,34 @@ public async Task ExecuteAsync(ExportJobRecord exportJobRecord, WeakETag weakETa
...
                 uint currentBatchId = progress.Page;
 
-                // The first item is placeholder for continuation token so that it can be updated efficiently later.
                 var queryParameters = new Tuple<string, string>[]
                 {
-                    null,
+                    Tuple.Create(KnownQueryParameterNames.ContinuationToken, progress.ContinuationToken),
```
jackliums: The comment should still be relevant. Maybe just say that the first item will be for continuation token and will be updated as process continues?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -106,46 +113,40 @@ public async Task ExecuteAsync(ExportJobRecord exportJobRecord, WeakETag weakETa
...
-                    foreach (ResourceWrapper resourceWrapper in searchResult.Results)
-                    {
-                        await ProcessResourceWrapperAsync(resourceWrapper, currentBatchId, cancellationToken);
-                    }
+                    await ProcessSearchResults(searchResult.Results, currentBatchId, cancellationToken);
 
                     if (searchResult.ContinuationToken == null)
                     {
                         // No more continuation token, we are done.
                         break;
```
jackliums: nit: the `progress.Page != 0` below in the if statement can be removed now since you changed the order of execution.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/GetExportRequestHandler.cs
```diff
@@ -38,7 +39,7 @@ public async Task<GetExportResponse> Handle(GetExportRequest request, Cancellati
                     outcome.JobRecord.QueuedTime,
                     outcome.JobRecord.RequestUri,
                     requiresAccessToken: false,
-                    outcome.JobRecord.Output,
+                    outcome.JobRecord.Output.Values.ToList(),
```
jackliums: it would be nice if this is ordered in some way (e.g., by resource type or eventually by resource type + part number).

src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/AzureExportDestinationClient.cs
```diff
@@ -104,24 +107,57 @@ public async Task CommitAsync(CancellationToken cancellationToken)
...
             // We can clear the stream mappings once we commit everything in memory.
+            foreach (Stream stream in _streamMappings.Values)
+            {
+                stream.Dispose();
+            }
+
             _streamMappings.Clear();
         }
 
+        public async Task InitializeForResumingExportAsync(CancellationToken cancellationToken)
```
jackliums: Instead of downloading all blobs, can we create a method like OpenFileAsync and open individual one as needed since we know the file name? Or make the method CreateIfNotExistAsync or something?

src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/AzureExportDestinationClient.cs
```diff
@@ -93,7 +94,9 @@ public async Task CommitAsync(CancellationToken cancellationToken)
         {
             CheckIfClientIsConnected();
 
-            var uploadAndCommitTasks = new List<Task>();
+            // Upload all blocks for each blob that was modified.
+            var uploadTasks = new List<Task>();
```
jackliums: nit: specify the initial capacity so the list doesn't have to grow. Also, use the array instead of list here since you know the size and it's cheaper to allocate.

src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/AzureExportDestinationClient.cs
```diff
@@ -104,24 +111,57 @@ public async Task CommitAsync(CancellationToken cancellationToken)
...
-                uploadAndCommitTasks.Add(Task.Run(async () =>
-                {
-                    await blobWrapper.UploadBlockAsync(blockId, stream, md5Hash: null, cancellationToken);
-                    await blobWrapper.CommitBlockListAsync(cancellationToken);
 
-                    stream.Dispose();
-                }));
+                uploadTasks.Add(blobWrapper.UploadBlockAsync(blockId, stream, md5Hash: null, cancellationToken));
+                wrappersToCommit.Add(blobWrapper);
             }
```
jackliums: could be simplified:

```
Task[] commitTasks = wrappersToCommit.Select(wrapper => wrapper.CommitBlockListAsync(cancellationToken)).ToArray();
```

src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/AzureExportDestinationClient.cs
```diff
@@ -104,24 +111,57 @@ public async Task CommitAsync(CancellationToken cancellationToken)
...
+        {
+            CheckIfClientIsConnected();
+
+            IEnumerable<IListBlobItem> blobs = _blobContainer.ListBlobs();
+
+            // For each blob, dowloand the committed block list and create a corresponding CloudBlockBlobWrapper.
+            foreach (IListBlobItem blobItem in blobs)
+            {
+                CloudBlockBlob blob = blobItem as CloudBlockBlob;
+                if (blob == null)
```
jackliums: I don't know if you can make this decision. If someone creates a page blob in the location, it should not affect the export and we should just ignore and move on.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
namadabu: Ya. Initially the Output field was a List and I had written the code based on that. But then it made more sense to change it into a Dictionary but I missed out on updating the code to take advantage of the fact that both Output and _resourceTypeToFileInforMapping are using the same datastructure. I have updated the code to remove the copying over part.

src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/AzureExportDestinationClient.cs
namadabu: Ya, makes sense.

src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/AzureExportDestinationClient.cs
namadabu: Renamed the method to OpenFilesAsync and passed in the list of specific files that it needs to open for resuming export.

src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/AzureExportDestinationClient.cs
```diff
@@ -93,35 +98,76 @@ public async Task CommitAsync(CancellationToken cancellationToken)
         {
             CheckIfClientIsConnected();
 
-            var uploadAndCommitTasks = new List<Task>();
+            // Upload all blocks for each blob that was modified.
+            Task[] uploadTasks = new Task[_streamMappings.Count];
+            int index = 0;
+            var wrappersToCommit = new List<CloudBlockBlobWrapper>();
```
namadabu: Use an array instead of List here also

src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/OrderedSetOfBlockIds.cs
```diff
@@ -0,0 +1,62 @@
...
+                _itemList.Add(item);
+            }
+        }
+
+        /// <summary>
+        /// Returns a list of all items that were added to the <see cref="OrderedSetOfBlockIds"/>
+        /// </summary>
+        public List<string> ToList()
+        {
+            return _itemList;
```
namadabu: Return a copy of the list

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -77,26 +77,37 @@ public async Task ExecuteAsync(ExportJobRecord exportJobRecord, WeakETag weakETa
...
+                // If we are resuming a job, we can detect that by checking the progress info from the job record.
+                // If it is null, then we know we are processing a new job.
+                if (_exportJobRecord.Progress != null)
+                {
+                    List<Uri> fileUris = _resourceTypeToFileInfoMapping.Values.Select(fileInfo => fileInfo.FileUri).ToList();
+                    await _exportDestinationClient.OpenFilesAsync(fileUris, cancellationToken);
 
-                // TODO: For now, always restart from the beginning. We will support resume in another work item.
-                _exportJobRecord.Progress = new ExportJobProgress(continuationToken: null, page: 0);
+                    _logger.LogTrace("Finished opening existing files for resuming export.");
```
jackliums: I think there was some miscommunication here. I don't think we need to open all of the files that's in progress. You could just load the file when you need it. 

e.g.,

in ProcessSearchResults

1. Check to see if there is a mapping in _resourceTypeToFileInfoMapping.
2. If there is mapping, use it
3. If there is no mapping, then check to see if there is a file or not from progress and open file if there is a file or create file is there isn't one.

This way it doesn't require you to open all files and download all block list up front. Only when it's necessary.

namadabu: Updated the code accordingly.

src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/AzureExportDestinationClient.cs
```diff
@@ -93,35 +108,69 @@ public async Task CommitAsync(CancellationToken cancellationToken)
...
             }
 
-            await Task.WhenAll(uploadAndCommitTasks);
+            await Task.WhenAll(uploadTasks);
+
+            // Commit all the blobs that were uploaded.
+            Task[] commitTasks = wrappersToCommit.Select(wrapper => wrapper.CommitBlockListAsync(cancellationToken)).ToArray();
+            await Task.WhenAll(commitTasks);
 
             // We can clear the stream mappings once we commit everything in memory.
```
jackliums: nit: not in memory anymore.

src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/CloudBlockBlobWrapper.cs
```diff
@@ -20,18 +20,21 @@ namespace Microsoft.Health.Fhir.Azure.ExportDestinationClient
...
-        private readonly List<string> _existingBlockIds;
+        private readonly OrderedSetOfBlockIds _existingBlockIds;
         private readonly CloudBlockBlob _cloudBlob;
 
         public CloudBlockBlobWrapper(CloudBlockBlob blockBlob)
+            : this(blockBlob, new List<string>())
+        {
+        }
+
+        public CloudBlockBlobWrapper(CloudBlockBlob blockBlob, List<string> blockList)
```
jackliums: prefer interface: IEnumerable

src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/OrderedSetOfBlockIds.cs
```diff
@@ -0,0 +1,62 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+using EnsureThat;
+
+namespace Microsoft.Health.Fhir.Azure.ExportDestinationClient
+{
+    /// <summary>
+    /// A modified implementation of an ordered hash set using a linked list and hashset. This class is specifically used
```
jackliums: This is using a list not linked list.

src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/OrderedSetOfBlockIds.cs
```diff
@@ -0,0 +1,62 @@
...
+    /// for maintaining the existing block ids that are part of a blob in <see cref="CloudBlockBlobWrapper"/>. We need to
+    /// maintain ordering of the block ids but also make sure that we don't have multiple copies of a block id. Hence
+    /// the need for an ordered hash set.
+    /// </summary>
+    internal class OrderedSetOfBlockIds
```
jackliums: I think you could use built-in SortedSet<T> since we are using the page number as the id.

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Export/ExportJobTaskTests.cs
```diff
@@ -301,6 +301,73 @@ public async Task GivenDeleteSecretFailed_WhenExecuted_ThenJobStatusShouldBeUpda
...
+                Options.Create(_exportJobConfiguration),
+                _searchService,
+                _resourceToByteArraySerializer,
+                _exportDestinationClientFactory,
+                NullLogger<ExportJobTask>.Instance);
+
+            numberOfSuccessfulPages = 6;
+            await secondExportJobTask.ExecuteAsync(_exportJobRecord, _weakETag, _cancellationToken);
+
+            exportedIds = _inMemoryDestinationClient.GetExportedData(new Uri("Patient.ndjson", UriKind.Relative));
```
jackliums: In the real world scenario, the destination client will be re-created so perhaps instead of reuse the in memory destination client, reinstantiate it and only capture the new entires?

also, shouldn't it also capture 5?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -183,48 +184,66 @@ private async Task UpdateJobStatus(OperationStatus operationStatus, bool updateE
...
-            // Check whether we already have an existing file for the current resource type.
-            if (!_resourceTypeToFileInfoMapping.TryGetValue(resourceType, out ExportFileInfo exportFileInfo))
+            foreach (ResourceWrapper resourceWrapper in searchResults)
             {
-                string fileName = resourceType + ".ndjson";
+                string resourceType = resourceWrapper.ResourceTypeName;
 
-                Uri fileUri = await _exportDestinationClient.CreateFileAsync(fileName, cancellationToken);
+                // Check whether we already have an existing file for the current resource type.
+                ExportFileInfo exportFileInfo;
```
jackliums: these two lines can be combined

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -183,48 +184,66 @@ private async Task UpdateJobStatus(OperationStatus operationStatus, bool updateE
...
+                if (!_resourceTypeToFileInfoMapping.TryGetValue(resourceType, out exportFileInfo))
+                {
+                    // Check whether we have seen this file previously (in situations where we are resuming an export)
+                    if (_exportJobRecord.Output.TryGetValue(resourceType, out exportFileInfo))
+                    {
+                        // A file already exists for this resource type. Let us open the file on the client.
+                        await _exportDestinationClient.OpenFileAsync(exportFileInfo.FileUri, cancellationToken);
 
-                exportFileInfo = new ExportFileInfo(resourceType, fileUri, sequence: 0);
+                        _resourceTypeToFileInfoMapping.Add(resourceType, exportFileInfo);
```
jackliums: this line is here and in the else clause so you can move it outside.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -170,11 +163,19 @@ public async Task ExecuteAsync(ExportJobRecord exportJobRecord, WeakETag weakETa
...
-                await UpdateJobStatus(OperationStatus.Failed, updateEndTimestamp: true, cancellationToken);
+                await UpdateAndCommitJobStatus(OperationStatus.Failed, updateEndTimestamp: true, cancellationToken);
+            }
+        }
+
+        private void UpdateJobOutputData(IDictionary<string, ExportFileInfo> resourceMapping)
+        {
+            foreach (KeyValuePair<string, ExportFileInfo> kvp in resourceMapping)
+            {
+                _exportJobRecord.Output.TryAdd(kvp.Key, kvp.Value);
```
jackliums: Couldn't this be added after line 234? Since the only time you will add a new entry to Output is when you create a new file?

src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/AzureExportDestinationClient.cs
```diff
@@ -93,35 +108,69 @@ public async Task CommitAsync(CancellationToken cancellationToken)
         {
             CheckIfClientIsConnected();
 
-            var uploadAndCommitTasks = new List<Task>();
+            // Upload all blocks for each blob that was modified.
+            Task[] uploadTasks = new Task[_streamMappings.Count];
+            CloudBlockBlobWrapper[] wrappersToCommit = new CloudBlockBlobWrapper[_streamMappings.Count];
+
+            int index = 0;
```
jackliums: nit: the code might look cleaner if you just use for instead of foreach.

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Export/ExportJobTaskTests.cs
namadabu: The reason for the discrepancy between the `numberOfSuccessfulPages` and `exportedIds` is because the `numberOfCalls` is incremented before the exception is thrown in the search service. That's why 2 ends up being skipped and we only end up with 0134 (and no 2 or 5). I'll update the code to increment `numberOfCalls` after the exception is thrown so that the resulting numbers are in order.

src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/OrderedSetOfBlockIds.cs
namadabu: We won't be able to use SortedSet here because we are base64 encoding the page number into a string. And in that case SortedSet will alter the order. If we change the way we generate the blockId, we can modify this part accordingly.

PR #548, opened 2019-06-19
Fixes bug where Resources with no search parameters could not be saved

brendankowitz:
## Description
Fixes bug where Resources with no search parameters could not be saved

## Related issues
Addresses [AB#69396](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/69396).

## Testing
Added regression test case to E2E tests with new resource type that was previously causing the error

### DIFF
--- a/test/Microsoft.Health.Fhir.R4.Tests.E2E/Rest/VersionSpecificTests.cs
+++ b/test/Microsoft.Health.Fhir.R4.Tests.E2E/Rest/VersionSpecificTests.cs
@@ -26,6 +26,16 @@ namespace Microsoft.Health.Fhir.Tests.E2E.Rest
         }
 
         [Fact]
+        public async Task GivenAnObservationDefinition_WhenCreating_ThenTheCorrectResponseShouldBeReturned()
+        {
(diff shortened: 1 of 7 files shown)


test/Microsoft.Health.Fhir.R4.Tests.E2E/Rest/VersionSpecificTests.cs
```diff
@@ -25,6 +25,16 @@ public async Task GivenR4Server_WhenCapabilityStatementIsRetrieved_ThenCorrectVe
             await TestCapabilityStatementFhirVersion("4.0.0");
         }
 
+        [Fact]
+        public async Task GivenAnObservationDefinition_WhenCreating_ThenTheCorrectResponseShouldBeReturned()
+        {
+            var resource = Samples.GetJsonSample<ObservationDefinition>("ObservationDefinition-example");
```
jackliums: nit: maybe add the test in SearchParameterDefinitionBuilderTests since this logic is not version specific?

brendankowitz: Good point, I'll add a test there too

PR #551, opened 2019-06-20
Add SQL support for Reverse Chaining

brandonpollett:
## Description
Updates parsing to support reverse chaining and add SQL server support for reverse chain queries

## Related issues
Addresses [AB#69375](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/69375).

## Testing
Added additional E2E tests for reverse chains and combinations of reverse and forward chaining. Additionally, the existing chained search tests still work.

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/ChainedExpression.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/ChainedExpression.cs
@@ -19,11 +19,13 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Expressions
         /// <param name="resourceType">The resource type that supports this search expression.</param>
         /// <param name="referenceSearchParameter">The search parameter that establishes the reference</param>
         /// <param name="targetResourceType">The target resource type.</param>
+        /// <param name="reversed">If this is a reversed chain parameter.</param>
         /// <param name="expression">The search expression.</param>
         public ChainedExpression(
             string resourceType,
             SearchParameterInfo referenceSearchParameter,
             string targetResourceType,
+            bool reversed,
             Expression expression)
         {
             EnsureArg.IsTrue(ModelInfoProvider.IsKnownResource(resourceType), nameof(resourceType));
@@ -34,6 +36,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Expressions
             ResourceType = resourceType;
             ReferenceSearchParameter = referenceSearchParameter;
             TargetResourceType = targetResourceType;
+            Reversed = reversed;
             Expression = expression;
         }
 
@@ -51,6 +54,8 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Expressions
         /// Gets the target resource type.
         /// </summary>
         public string TargetResourceType { get; }
+
+        public bool Reversed { get; set; }
 
         /// <summary>
         /// Gets the search expression.
--- a/src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/ChainedExpression.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/ChainedExpression.cs
@@ -19,7 +19,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Expressions
         /// <param name="resourceType">The resource type that supports this search expression.</param>
         /// <param name="referenceSearchParameter">The search parameter that establishes the reference</param>
         /// <param name="targetResourceType">The target resource type.</param>
-        /// <param name="reversed">If this is a reversed chain parameter.</param>
+        /// <param name="reversed">If this is a reversed chained expression.</param>
         /// <param name="expression">The search expression.</param>
         public ChainedExpression(
             string resourceType,
@@ -55,7 +55,10 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Expressions
         /// </summary>
         public string TargetResourceType { get; }
 
-        public bool Reversed { get; set; }
+        /// <summary>
+        /// Get if the expression is reversed.
+        /// </summary>
+        public bool Reversed { get; }
 
         /// <summary>
         /// Gets the search expression.
@@ -71,7 +74,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Expressions
 
         public override string ToString()
         {
-            return $"(Chain {ReferenceSearchParameter.Name}:{TargetResourceType} {Expression})";
+            return $"({(Reversed ? "Reverse " : string.Empty)}Chain {ReferenceSearchParameter.Name}:{TargetResourceType} {Expression})";
         }
     }
 }
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Expressions/Parsers/ExpressionParser.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Expressions/Parsers/ExpressionParser.cs
@@ -8,7 +8,6 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 using EnsureThat;
-using Hl7.Fhir.Model;
 using Hl7.Fhir.Utility;
 using Microsoft.Health.Fhir.Core.Features.Definition;
 using Microsoft.Health.Fhir.Core.Models;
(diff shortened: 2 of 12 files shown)


src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/ChainedExpression.cs
```diff
@@ -52,6 +55,8 @@ public class ChainedExpression : Expression
         /// </summary>
         public string TargetResourceType { get; }
 
+        public bool Reversed { get; set; }
```
jackliums: Should this flag be exposed in the ToString below?

brandonpollett: It should... I had that change sitting on my local version and had apparently forgotten to push. Updated now.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Expressions/Parsers/ExpressionParser.cs
```diff
@@ -61,50 +59,72 @@ public Expression Parse(string resourceType, string key, string value)
...
 
-            if (paths?.Length == 0)
+        private Expression ParseImpl(string resourceType, ReadOnlySpan<char> key, string value)
+        {
+            if (TryConsume("_has:".AsSpan(), ref key))
             {
-                throw new SearchParameterNotSupportedException(resourceType, key);
+                if (!TrySplit(':', ref key, out var type))
+                {
+                    throw new Exception("missing type");
```
jackliums: We should not be throwing generic Exception type.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Expressions/Parsers/ExpressionParser.cs
```diff
@@ -61,50 +59,72 @@ public Expression Parse(string resourceType, string key, string value)
...
-            PathSegment[] paths = key.Split(new[] { SearchParams.SEARCH_CHAINSEPARATOR }, StringSplitOptions.RemoveEmptyEntries)
-                 .Select(path => PathSegment.Parse(path))
-                 .ToArray();
+            return ParseImpl(resourceType, key.AsSpan(), value);
+        }
 
-            if (paths?.Length == 0)
+        private Expression ParseImpl(string resourceType, ReadOnlySpan<char> key, string value)
+        {
+            if (TryConsume("_has:".AsSpan(), ref key))
```
jackliums: could `"_has:".AsSpan()` be stored as constant?

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Expressions/Parsers/ExpressionParser.cs
```diff
@@ -61,50 +59,72 @@ public Expression Parse(string resourceType, string key, string value)
...
             {
-                throw new SearchParameterNotSupportedException(resourceType, key);
+                if (!TrySplit(':', ref key, out var type))
```
jackliums: nit: instead of var, should be the actual type since we can't infer the type from code.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Expressions/Parsers/ExpressionParser.cs
```diff
@@ -61,50 +59,72 @@ public Expression Parse(string resourceType, string key, string value)
...
+                {
+                    throw new Exception("missing type");
+                }
+
+                if (!TrySplit(':', ref key, out var refParam))
+                {
+                    throw new Exception("missing search reference search param");
+                }
+
+                SearchParameterInfo refSearchParameter = _searchParameterDefinitionManager.GetSearchParameter(type.ToString(), refParam.ToString());
```
jackliums: `type.ToString()` can be saved into a variable and used here and below.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -256,13 +256,28 @@ public object VisitTable(TableExpression tableExpression, SearchOptions context)
                             .Append(" = ").Append(V1.Resource.ResourceId, resourceTableAlias);
                     }
 
+                    // Denormalized predicates on reverse chains need to be applied via a join to the resource table
```
jackliums: Do we need test for paging or is that a separate user story?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/ChainingSearchTests.cs
```diff
@@ -61,7 +58,7 @@ public async Task GivenAChainedSearchExpressionOverASimpleParameter_WhenSearched
 
             Bundle bundle = await Client.SearchAsync(ResourceType.DiagnosticReport, query);
 
-            ValidateBundle(bundle, Fixture.SmithDiagnosticReport, Fixture.TrumanDiagnosticReport);
+            ValidateBundle(bundle, Fixture.SmithSnomedDiagnosticReport, Fixture.SmithLoincDiagnosticReport, Fixture.TrumanSnomedDiagnosticReport, Fixture.TrumanLoincDiagnosticReport);
         }
 
         [Fact]
```
jackliums: Can we also add a test to do multiple chaining? e.g., `result.patient.organization.address-city=Seattle`

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Expressions/Parsers/ExpressionParser.cs
```diff
@@ -113,65 +133,72 @@ private Expression ParseChainedExpression(string resourceType, SearchParameterIn
...
 
-            ResourceType? scopedTargetResourceType = null;
-
             // Check to see if the client has specifically specified the target resource type to scope to.
-            if (targetResourceType != null)
+            if (!string.IsNullOrEmpty(targetResourceType))
             {
                 // A target resource type is specified.
-                if (!Enum.TryParse(targetResourceType, out ResourceType parsedResourceType))
+                if (!Enum.TryParse<ResourceType>(targetResourceType, out _))
```
brendankowitz: Probably could just use `IModelInfoProvider.IsKnownResource()`

src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/ChainedExpression.cs
```diff
@@ -52,6 +55,11 @@ public class ChainedExpression : Expression
         /// </summary>
         public string TargetResourceType { get; }
 
+        /// <summary>
+        /// Get if the expression is reversed.
+        /// </summary>
+        public bool Reversed { get; }
```
brendankowitz: I've been wondering if it makes to sense to specify `ReverseChainedExpression` as its own class (or inherited class). Then if polymorphism can smooth out some of the conditionals below. Would mean that `TableExpressionKind.ReverseChain` might be a concept too. Thoughts?

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Expressions/Parsers/ExpressionParser.cs
brandonpollett: It doesn't look like you can have a ReadOnlySpan<char> as a const, but I did make the string `_has:` a const.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
brandonpollett: I added paging tests for the various combinations of chaining/reverse chaining.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/ChainingSearchTests.cs
brandonpollett: This was covered in one level, but I added another level like you mentioned.

src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/ChainedExpression.cs
brandonpollett: I think it's possible to split it out, but I'm not sure how much cleaner it would be. Most of the logic is in the SQL generation step and splitting it doesn't appear to make that any cleaner/easier. Most of the logic between chain and reverse chain is the same, just swapping what we consider `Sid1` and `Sid2` and the occasional table change so we're filtering correctly.

PR #553, opened 2019-06-20
Merge ModelInfoProviders

brendankowitz:
## Description
Merges ModelInfoProvider classes


src/Microsoft.Health.Fhir.R4.Core/VersionSpecificModelInfoProvider.cs
```diff
@@ -3,15 +3,15 @@
...
-using System.Reflection;
-using Microsoft.Health.Fhir.Core;
+using Microsoft.Health.Fhir.Core.Models;
 
-namespace Microsoft.Health.Fhir.Api.Modules
+namespace Microsoft.Health.Fhir.Core
 {
-    public static class KnownAssemblies
+    /// <summary>
+    /// Provides agnostic access to R4 FHIR Models and Resources
```
brandonpollett: nit: I don't think this is agnostic since it's specific to R4

src/Microsoft.Health.Fhir.Shared.Api/Modules/KnownAssemblies.cs
```diff
@@ -12,6 +12,6 @@ public static class KnownAssemblies
     {
         public static Assembly Core => typeof(Clock).Assembly;
 
-        public static Assembly CoreVersionSpecific => typeof(R4ModelInfoProvider).Assembly;
+        public static Assembly CoreVersionSpecific => typeof(VersionSpecificModelInfoProvider).Assembly;
```
jackliums: nit: I wonder if we should have some marker interface or something instead of using clock above. Someone might accidentally refactor the code and move Clock to say a utility library or something?

src/Microsoft.Health.Fhir.Shared.Core/Extensions/ModelExtensions.cs
```diff
@@ -20,11 +20,7 @@ public static class ModelExtensions
         /// </summary>
         public static void SetModelInfoProvider()
         {
-#if Stu3
-            ModelInfoProvider.SetProvider(new Stu3ModelInfoProvider());
-#elif R4
-            ModelInfoProvider.SetProvider(new R4ModelInfoProvider());
-#endif
+            ModelInfoProvider.SetProvider(new VersionSpecificModelInfoProvider());
```
jackliums: Can we remove the macro definition now?

brendankowitz: There is still 1 more use in `SearchParameterDefinitionBuilder`

PR #565, opened 2019-07-02
GeoPolitical Update: Fix for GeoPolExclude into master branch

JoeDRowan:
## Description
Suggestion for excluding file from Geo Political which concerns some flags at present.
Note @jackliums that we can use wildcards if we want to exclude multiple files e.g. *.json

### DIFF
--- a/GeoPol.xml
+++ b/GeoPol.xml
@@ -18,8 +18,9 @@
     <ExcludeFolder>GeoPol.xml</ExcludeFolder>
     <ExcludeFolder>THIRDPARTYNOTICES.md</ExcludeFolder>
     <ExcludeFolder>src\Microsoft.Health.Fhir.Stu3.Core\Features\Definition\search-parameters.json</ExcludeFolder>
-    <ExcludeFolder>src\Microsoft.Health.Fhir.R4.Core\Features\Definition\search-parameters.json</ExcludeFolder>
     <ExcludeFolder>src\Microsoft.Health.Fhir.Stu3.Core\Features\Conformance\AllCapabilities.json</ExcludeFolder>
+    <ExcludeFolder>src\Microsoft.Health.Fhir.R4.Core\Features\Definition\search-parameters.json</ExcludeFolder>
     <ExcludeFolder>src\Microsoft.Health.Fhir.R4.Core\Features\Conformance\AllCapabilities.json</ExcludeFolder>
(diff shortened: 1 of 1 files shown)


GeoPol.xml
```diff
@@ -18,8 +18,9 @@
     <ExcludeFolder>GeoPol.xml</ExcludeFolder>
     <ExcludeFolder>THIRDPARTYNOTICES.md</ExcludeFolder>
     <ExcludeFolder>src\Microsoft.Health.Fhir.Stu3.Core\Features\Definition\search-parameters.json</ExcludeFolder>
-    <ExcludeFolder>src\Microsoft.Health.Fhir.R4.Core\Features\Definition\search-parameters.json</ExcludeFolder>
     <ExcludeFolder>src\Microsoft.Health.Fhir.Stu3.Core\Features\Conformance\AllCapabilities.json</ExcludeFolder>
+    <ExcludeFolder>src\Microsoft.Health.Fhir.R4.Core\Features\Definition\search-parameters.json</ExcludeFolder>
     <ExcludeFolder>src\Microsoft.Health.Fhir.R4.Core\Features\Conformance\AllCapabilities.json</ExcludeFolder>
+    <ExcludeFolder>src\Microsoft.Health.Fhir.R4.Core\Features\Conformance\DefaultCapabilities.json</ExcludeFolder>
```
brandonpollett: If we're excluding the R4 `DefaultCapabilities.json`, should we also exclude the STU3 `DefaultCapabilities.json`?

JoeDRowan: I really don't know @brandonpollett where these files are used and whether they should be scanned for GeoPolitical issues or not. If in doubt about whether a file should be included, best to include. If it is an external file that we cannot change, then we should exclude from a scan as it would create un-fixable noise issues otherwise. Remember wildcards are supported e.g. *.json

brandonpollett: These files are used to generate our metadata endpoint that is exposed to the customers. Some of it can probably be changed and some of it can't. Ultimately it just needs to conform to the spec (details here: https://www.hl7.org/fhir/capabilitystatement.html). I guess another question to ask, is why was the STU3 default capabilities added to the exclude list?

JoeDRowan: That file threw up some GeoPolitical issues @brandonpollett which @jackliums investigated and figured that we should exclude from future scans.

PR #570, opened 2019-07-05
Changes to DotNetCore 2.2.107

brendankowitz:
## Description
Updates .Net Core SDK version to 2.2.107 (from 2.2.300 which isn't compatible with Kudu build/deployments and VS2017). 2.2.107 uses the same runtime version as 2.2.300.

## Related issues
Addresses [AB#69468](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/69468)

## Testing
Existing tests were run.


brandonpollett:
Looks like you should be good after updating the `PR Internal Checks` to use the 2.2.107 version of the SDK.

PR #576, opened 2019-07-11
Added logic to allow export job to be canceled.

jackliums:
## Description
Added logic to allow export job to be canceled.

## Related issues
Addresses #346. Work item: [AB#69519](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/69519).

## Testing
Unit test and manual testing. E2E tests will be added as a separate work item.


src/Microsoft.Health.Fhir.Shared.Api/Controllers/ExportController.cs
```diff
@@ -148,6 +151,16 @@ public async Task<IActionResult> GetExportStatusById(string idParameter)
             return exportActionResult;
         }
 
+        [HttpDelete]
+        [Route(KnownRoutes.ExportJobLocation, Name = RouteNames.CancelExport)]
+        [AuditEventType(AuditEventSubType.Export)]
+        public async Task<IActionResult> CancelExport(string id)
```
namadabu: `id` needs to be renamed to `idParameter` to match the `KnownRoutes.ExportJobLocation` template. Else the parameter value will be null

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/CancelExportRequestHandler.cs
```diff
@@ -0,0 +1,67 @@
...
+                case OperationStatus.Completed:
+                case OperationStatus.Failed:
+                    return new CancelExportResponse(HttpStatusCode.Conflict);
+            }
+
+            // Try to cancel the job.
+            outcome.JobRecord.Status = OperationStatus.Cancelled;
+            outcome.JobRecord.CancelledTime = Clock.UtcNow;
+
+            await _retryPolicy.ExecuteAsync(() => _fhirOperationDataStore.UpdateExportJobAsync(outcome.JobRecord, outcome.ETag, cancellationToken));
```
namadabu: If we get a `JobConflictException` here, wouldn't we hit the same error on the retries since we are not passing in the updated ETag?

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Export/CancelExportRequestHandlerTests.cs
```diff
@@ -0,0 +1,133 @@
...
+            ExportJobOutcome outcome = await ExecuteCancelExportAsync(operationStatus, HttpStatusCode.Conflict);
+
+            Assert.Equal(operationStatus, outcome.JobRecord.Status);
+            Assert.Null(outcome.JobRecord.CancelledTime);
+        }
+
+        [Theory]
+        [InlineData(OperationStatus.Queued)]
+        [InlineData(OperationStatus.Running)]
+        public async Task GivenAFhirMediator_WhenCancelingExistingExportJobThatHasNotCompleted_ThenConflictStatusCodeShouldbeReturned(OperationStatus operationStatus)
```
namadabu: ThenAcceptedStatusCodeShouldbeReturned

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Export/CancelExportRequestHandlerTests.cs
```diff
@@ -0,0 +1,133 @@
...
+
+            ServiceProvider provider = collection.BuildServiceProvider();
+            _mediator = new Mediator(type => provider.GetService(type));
+        }
+
+        [Theory]
+        [InlineData(OperationStatus.Cancelled)]
+        [InlineData(OperationStatus.Completed)]
+        [InlineData(OperationStatus.Failed)]
+        public async Task GivenAFhirMediator_WhenCancelingExistingExportJobHasAlreadyCompleted_ThenConflictStatusCodeShouldbeReturned(OperationStatus operationStatus)
```
namadabu: nit: WhenCancelingExistingExportJobThatHasAlreadyCompleted

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -156,47 +156,72 @@ public async Task ExecuteAsync(ExportJobRecord exportJobRecord, WeakETag weakETa
...
+            using (IScoped<IFhirOperationDataStore> fhirOperationDataStore = _fhirOperationDataStoreFactory())
+            {
+                updatedOutcome = await fhirOperationDataStore.Value.GetExportJobByIdAsync(_exportJobRecord.Id, cancellationToken);
+            }
+
+            if (updatedOutcome.JobRecord.Status == OperationStatus.Cancelled)
+            {
+                _logger.LogInformation("The job was canceled by another process.");
+
+                // The job was canceled. Merge the output without updating the status.
```
namadabu: Is there a reason we want to store the updated Output? Since the job has anyway been cancelled, will the Output data be used?

jackliums: I think it's still useful to update the Output because consider the following case:

1. Job persist the output information at T1. 
2. Job starts processing the next batches (and let's say `MaximumNumberOfResourcesPerQuery = 100` and `NumberOfPagesPerCommit = 10`)
3. The user cancels the job while Step 2 is happening.

In this case, the job did process additional 1000 resources but updating the job record would fail with JobConflictException because the job was canceled. I think it would be nice to actually update the Output one last time to reflect the latest number otherwise, depending on the configuration values, the discrepancy could be very large.

But this brings up an interesting issue that I didn't think about which is that with the current implementation, it will only stop the job once the job finish processing the batch but the batch size could be large depending on what configuration setting is so it might take a while before the batch finishes. We might need to add check in the loop or something to detect cancellation earlier.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ExportController.cs
jackliums: Yeah I missed this one with the merge. Will fix it.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/CancelExportRequestHandler.cs
jackliums: Good catch. I meant to have the whole logic within the retry block so it can fetch the latest change and update the status. Updated the code and updated the unit test.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
jackliums: I thought about this a bit more and I think you are right. If the job was cancelled before the next commit, there could be additional resources exported but they are not committed yet so it won't show up in the destination anyways so we don't need to update the Output.

namadabu: I agree that it if we are returning the Output to the user, it makes sense that we return the updated one (so that the user can determine what they want to do with it). I initially thought that we will not be returning an Output response for such cancelled export requests. I just read through the spec again and they don't specifically mention how the server should respond (for the get export status call) once the export job gets cancelled. I think we have the freedom to decide whether we want to call it a success or failure.

But marking this as success and returning the Output containing currently exported data might create another kind of issue. Depending on when the client polls for the export status, they can receive different kinds of Output data but the operation status as Cancelled. It might be difficult for the client to determine when they will get the "final" Output data after polling the status. Like you mentioned above, this can be mitigated if we poll for the status within the loop. But is the perf hit that comes with the polling worth it? I feel it might be better to not promise the client anything in the case of a cancelled export job.

jackliums: I added the "lightweight" check using ETag in the loop. It returns 304 Not Modified if the document is not updated and only costs 1 RU as opposed to fetch the whole document. This way it can detect cancellation quicker without putting too much load on the backend. We can try it out and see how it is and decide if we want to keep it or not.

src/Microsoft.Health.CosmosDb/Features/Storage/RetryExceptionPolicy.cs
```diff
@@ -22,7 +22,7 @@ public static bool IsTransient(DocumentClientException ex)
             // Detects "449 Retry With" - The operation encountered a transient error. This only occurs on write operations. It is safe to retry the operation.
             // Detects "429 Too Many Request" - The collection has exceeded the provisioned throughput limit. Retry the request after the server specified retry after duration.
             // For more information see: https://docs.microsoft.com/en-us/rest/api/documentdb/http-status-codes-for-documentdb
-            if (ex.StatusCode == (HttpStatusCode)449 || ex.StatusCode == (HttpStatusCode)429)
+            if (ex.StatusCode == (HttpStatusCode)449 || ex.StatusCode == HttpStatusCode.Conflict)
```
namadabu: This should still be 429 right?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -156,47 +156,72 @@ public async Task ExecuteAsync(ExportJobRecord exportJobRecord, WeakETag weakETa
...
+            await UpdateJobRecordAsync(cancellationToken);
         }
 
-        private async Task UpdateAndCommitJobRecord(ExportJobRecord jobRecord, CancellationToken cancellationToken)
+        private async Task UpdateJobRecordAsync(CancellationToken cancellationToken)
         {
             using (IScoped<IFhirOperationDataStore> fhirOperationDataStore = _fhirOperationDataStoreFactory())
             {
-                ExportJobOutcome updatedExportJobOutcome = await fhirOperationDataStore.Value.UpdateExportJobAsync(jobRecord, _weakETag, cancellationToken);
+                ExportJobOutcome updatedExportJobOutcome = await fhirOperationDataStore.Value.UpdateExportJobAsync(_exportJobRecord, _weakETag, cancellationToken);
```
namadabu: This could still throw a `JobConflictException` since we will be updating the JobRecord during this stage.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -125,22 +131,57 @@ public async Task ExecuteAsync(ExportJobRecord exportJobRecord, WeakETag weakETa
...
+                            return;
+                        }
+                    }
                 }
 
                 // Commit one last time for any pending changes.
                 await _exportDestinationClient.CommitAsync(cancellationToken);
 
-                await UpdateAndCommitJobStatus(OperationStatus.Completed, updateEndTimestamp: true, cancellationToken);
+                await CompleteJobAsync(OperationStatus.Completed, cancellationToken);
```
namadabu: This could still throw a JobConflictException since we will be updating the JobRecord during this stage.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
namadabu: This reduces the probability that a user will receive different Output data when polling for the status of a cancelled export job (depending on the timing of the poll and the final job record update) but does not completely eliminate it. I still feel that we are trying to optimize for something that the end-user might not even need (they might just be interested in seeing whether the export job was cancelled rather than what state it was in when it got cancelled).

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
jackliums: Chat about this a little bit. Removed the logic that tries to keep the output up-to-date.

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Export/ExportJobTaskTests.cs
```diff
@@ -29,6 +29,7 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Operations.Export
 {
     public class ExportJobTaskTests
     {
+        private const string PatientFileName = "Patient.ndjson";
```
namadabu: nit: Can be removed.

jackliums: Updated the code to actually use the constant.

Issue #453, opened 2019-05-07
More explicit exception should be thrown when continuation token is invalid


jackliums:
**Describe the bug**
If an invalid continuation token is passed in, the server throws generic 500 Internal Server Error. We should throw more specific error here.

**To Reproduce**
1. Make a search call with some invalid continuation token.

**Expected behavior**
The server throw an error indicating that the continuation token is invalid.

**Actual behavior**
500 Internal Server Error is thrown with generic error message.

Issue #452, opened 2019-05-07
We should base64 encoding the continuation token.


jackliums:
**Describe the bug**

Currently, if the persistent store is Cosmos DB, we return the continuation token from Cosmos DB back to the caller as our continuation token in the next link. 

The string we returned in JSON is URL encoded. The next link in the .NET STU3 library is an instance of URI.

If the caller makes the next call using the instance of URI or .OriginalString, everything works. But if the caller uses .ToString to convert the instance of URI back to string and use that, then it unescape string. Because Cosmos DB continuation token contains + and :, when the next calls is made, it throws error because the continuation token is not valid anymore.

**To Reproduce**
1. Import a lot of Patient resource.
2. Execute a search and get the next link that contains + and :.
3. Get the next link URL as a string by executing .ToString on the URI instance deserialized from the response.
4. Make the new search code.

**Expected behavior**
The search results should be returned.

**Actual behavior**
500 Internal Server Error is returned.

PR #584, opened 2019-07-16
Encode the continuation token

LTA-Thinking:
## Description
Base 64 encode the continuation token and send a proper error message when a bad continuation token is passed.

## Related issues
Addresses #452 and #453.

## Testing
Updated existing unit test to check that the token is encoded.

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Search/SearchService.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Search/SearchService.cs
@@ -51,8 +51,16 @@ namespace Microsoft.Health.Fhir.Core.Features.Search
         {
             SearchOptions searchOptions = _searchOptionsFactory.Create(resourceType, queryParameters);
 
-            // Execute the actual search.
-            return await SearchInternalAsync(searchOptions, cancellationToken);
+            try
+            {
+                // Execute the actual search.
+                return await SearchInternalAsync(searchOptions, cancellationToken);
+            }
+            catch (Exception)
+            {
+                // Should a logging statement be added?
+                throw new RequestNotValidException(Resources.InvalidContinuationToken);
+            }
         }
 
         /// <inheritdoc />
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/BundleFactory.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/BundleFactory.cs
@@ -36,55 +36,50 @@ namespace Microsoft.Health.Fhir.Core.Features.Search
 
         public ResourceElement CreateSearchBundle(SearchResult result)
         {
-            EnsureArg.IsNotNull(result, nameof(result));
-
-            // Create the bundle from the result.
-            var bundle = new Bundle();
-
-            if (result != null)
+            return CreateBundle(result, Bundle.BundleType.Searchset, r =>
             {
-                IEnumerable<Bundle.EntryComponent> entries = result.Results.Select(r =>
-                {
-                    ResourceElement resource = _deserializer.Deserialize(r);
+                ResourceElement resource = _deserializer.Deserialize(r);
 
-                    return new Bundle.EntryComponent
+                return new Bundle.EntryComponent
+                {
+                    FullUrlElement = new FhirUri(_urlResolver.ResolveResourceUrl(resource)),
+                    Resource = resource.Instance.ToPoco<Resource>(),
+                    Search = new Bundle.SearchComponent
                     {
-                        FullUrlElement = new FhirUri(_urlResolver.ResolveResourceUrl(resource)),
-                        Resource = resource.Instance.ToPoco<Resource>(),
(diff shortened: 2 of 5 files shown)


src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
```diff
@@ -58,13 +59,22 @@ public SearchOptions Create(string compartmentType, string compartmentId, string
...
+                    // This is an unreachable case. The mapping of the query parameters makes it so only one continuation token can exist.
                     if (continuationToken != null)
                     {
                         throw new InvalidSearchOperationException(
                             string.Format(Core.Resources.MultipleQueryParametersNotAllowed, KnownQueryParameterNames.ContinuationToken));
                     }
 
-                    continuationToken = query.Item2;
+                    // Checks if the continuation token is base 64 bit encoded. Needed for systems that have cached continuation tokens from before they were encoded.
+                    if (Regex.IsMatch(query.Item2, "^([A-Za-z0-9+/]{4})*([A-Za-z0-9+/]{3}=|[A-Za-z0-9+/]{2}==)?$"))
```
feordin: >if (Regex.IsMatch(query.Item2, "^([A-Za-z0-9+/]{4})*([A-Za-z0-9+/]{3}=|[A-Za-z0-9+/]{2}==)?$")) [](start = 20, length = 95)

Is there any chance that older continuation tokens which are not base 64 encoded will pass this regex test?  I am not sure what they look like.

LTA-Thinking: The unencoded token seems to be JSON formatted, so it will have characters like { that wouldn't show up in a Base 64 encoded string. Here is an example of an old token taken from the debugger: 
"{\"token\":\"+RID:CvwaAKHaQvz9FgAAAAAAAA==#RT:1#TRC:10#FPC:AgEAAAAGAP2WAcD8/w==\",\"range\":{\"min\":\"\",\"max\":\"FF\"}}"

feordin: Sounds good.

brendankowitz: We should also use compiled regex (maybe in a static variable) here since it could be a common call

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/BundleFactory.cs
```diff
@@ -93,43 +88,23 @@ public ResourceElement CreateHistoryBundle(SearchResult result)
...
-                            Url = hasVerb ? $"{resource.InstanceType}/{(httpVerb == Bundle.HTTPVerb.POST ? null : resource.Id)}" : null,
-                        },
-                        Response = new Bundle.ResponseComponent
-                        {
-                            LastModified = r.LastModified,
-                            Etag = WeakETag.FromVersionId(r.Version).ToString(),
-                        },
-                    };
-                });
+                IEnumerable<Bundle.EntryComponent> entries = result.Results.Select(selectionFunction);
```
brendankowitz: Any reason to not just pass in IEnumerable<Bundle.EntryComponent> instead of the delegate?

LTA-Thinking: I don't see how you could pass in the entries object to the Select method. Select returns a filled list, it doesn't have an overload that takes in a list and fills it.

src/Microsoft.Health.Fhir.Core/Features/Search/SearchService.cs
```diff
@@ -51,8 +51,16 @@ protected SearchService(ISearchOptionsFactory searchOptionsFactory, IFhirDataSto
...
             SearchOptions searchOptions = _searchOptionsFactory.Create(resourceType, queryParameters);
 
-            // Execute the actual search.
-            return await SearchInternalAsync(searchOptions, cancellationToken);
+            try
+            {
+                // Execute the actual search.
+                return await SearchInternalAsync(searchOptions, cancellationToken);
+            }
+            catch (Exception)
```
jackliums: This might be a problem since it treats any error as invalid continuation token. Instead, I think we should create our own InvalidContinuationTokenException or something like that and catch specific exception type in `FhirCosmosSearchService` and convert the exception type.

brendankowitz: This can be updated in https://github.com/microsoft/fhir-server/pull/589

LTA-Thinking: The specific exception type is a DocumentClientException. The issue I had when I tried to catch it specifically is that the class definition can't be found. It's in the Azure namespace so I don't know why a reference can't be made to it.

Issue #578, opened 2019-07-11
lost Sorting data while parsing incoming search parameters


devcs21:
**Describe the bug**
if one sends sorting parameters to fhir server it  lost them in SearchOptionsFactory so it's impossible to sort

**To Reproduce**
1. open Visual Studio, go to SearchOptionsFactory->Create method
2. Set a breakpoint on " return searchOptions;" 
3. using postman/fiddler/(or custom FHIR client of your choice) send a request with "sort" parameter in URL  something like that https://localhost:44348/CommunicationRequest?_sort=-occurrence&occurrence=lt2019-07-12 
4. once reached breakpoint go to "watch" window:
searchParams variable **contains** sorting parameter (searchParams.Sort property  contains correct sorting data) but searchOptions does **not**

**Expected behavior**
Add sorting parameter data to returning searchOptions

**Actual behavior**
sorting parameters are not processed in any way

possible fix (our current workaround):
1. modify Microsoft.Health.Fhir.Core.Features.Search.SearchOptions by adding new property:
public IEnumerable<Tuple<string, string>> Sort { get;  set; }

2. modify Microsoft.Health.Fhir.Core.Features.Search.SearchOptionsFactory by adding this code **before** "return searchOptions;" statement :
             if (null != searchParams.Sort && 0 < searchParams.Sort.Count)
            {
                searchOptions.Sort = searchParams.Sort.Select(sorting => Tuple.Create(sorting.Item1, 
                                                                               sorting.Item2.ToString()));
            }

            return searchOptions;

jackliums:
We haven't implemented the sorting yet and that's why it's not being handled currently. But we could expose it in the SearchOptions at least so if someone needs to use it, it won't be lost.

devcs21:
@jackliums it would be great!

PR #586, opened 2019-07-16
Add sort parameters to the search options

LTA-Thinking:
## Description
Add sort parameters that were entered as part of a search to the search options. This does not add support for sorting.

## Related issues
Addresses #578.

## Testing
New test is added to check that sort parameters are added properly to the search options.

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Search/SearchOptions.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Search/SearchOptions.cs
@@ -62,6 +62,6 @@ namespace Microsoft.Health.Fhir.Core.Features.Search
         /// <summary>
         /// Gets the list of sorting parameters. The second item in a tuple is whether the sort is accending or decending.
         /// </summary>
-        public IEnumerable<Tuple<string, string>> Sort { get; internal set; }
+        public IEnumerable<Tuple<string, SortOrder>> Sort { get; internal set; }
     }
 }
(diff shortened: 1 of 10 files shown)


src/Microsoft.Health.Fhir.Core/Features/Search/SearchOptions.cs
```diff
@@ -58,5 +58,10 @@ internal set
         /// Gets the list of search parameters that were not used in the search.
         /// </summary>
         public IReadOnlyList<Tuple<string, string>> UnsupportedSearchParams { get; internal set; }
+
+        /// <summary>
+        /// Gets the list of sorting parameters. The second item in a tuple is whether the sort is accending or decending.
+        /// </summary>
+        public IEnumerable<Tuple<string, string>> Sort { get; internal set; }
```
jackliums: We should define the sort order as an enum (I know it's redundant but string is not friendly here). Eventually, i think the FHIR .NET library will have a base library that contains all "common" objects that would work across versions so we can switch to that later.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
```diff
@@ -170,6 +170,12 @@ public SearchOptions Create(string compartmentType, string compartmentId, string
 
             searchOptions.UnsupportedSearchParams = unsupportedSearchParameters;
 
+            // Sort is currently not implimented. The sort parameters are added to the search options to allow for future development.
+            if (searchParams.Sort != null && searchParams.Sort.Count > 0)
+            {
+                searchOptions.Sort = searchParams.Sort.Select(sorting => Tuple.Create(sorting.Item1, (SortOrder)sorting.Item2));
```
jackliums: let's do Enum.Parse instead of direct casting. In case the value changes (which is unlikely... but you never know), we would fail as oppose to cast it to something unexpectedly.

PR #596, opened 2019-07-31
Improved error handling for export requests

namadabu:
## Description
Added code to return operation outcome in cases where export fails.
Added a level of abstraction for secret store errors.

## Related issues
Addresses [issue [AB#69469](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/69469)].

## Testing
Some UTs and integration tests. Manual E2E testing using Postman


src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -166,12 +166,27 @@ public async Task ExecuteAsync(ExportJobRecord exportJobRecord, WeakETag weakETa
...
+                _exportJobRecord.FailureReason = dce.Message;
+                await CompleteJobAsync(OperationStatus.Failed, cancellationToken);
+            }
             catch (Exception ex)
             {
                 // The job has encountered an error it cannot recover from.
                 // Try to update the job to failed state.
                 _logger.LogError(ex, "Encountered an unhandled exception. The job will be marked as failed.");
 
+                _exportJobRecord.FailureReason = "Unknown Error";
```
jackliums: Can this be ex.Message? If so, then these 3 conditions could just be collapsed into this one and you can output the error message in the log?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/GetExportRequestHandler.cs
```diff
@@ -46,6 +46,10 @@ public async Task<GetExportResponse> Handle(GetExportRequest request, Cancellati
 
                 exportResponse = new GetExportResponse(HttpStatusCode.OK, jobResult);
             }
+            else if (outcome.JobRecord.Status == OperationStatus.Failed || outcome.JobRecord.Status == OperationStatus.Canceled)
+            {
+                exportResponse = new GetExportResponse(HttpStatusCode.InternalServerError, outcome.JobRecord.FailureReason);
```
jackliums: I don't think this is right. InternalServerError signals that this particular GetExport request has failed not that the job itself has failed. If the job is failed, it should still return 200 OK and the user can look at the OperationOutcome within the response to figure out what happened (or we can return some flag/status to indicate the result of the job).

src/Microsoft.Health.Fhir.Core/Messages/Export/CreateExportResponse.cs
```diff
@@ -9,13 +9,32 @@ namespace Microsoft.Health.Fhir.Core.Messages.Export
 {
     public class CreateExportResponse
     {
-        public CreateExportResponse(string id)
+        public string JobId { get; private set; }
+
+        public bool Successful { get; private set; }
```
jackliums: nit: this flag can just check if JobId is null or not.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/GetExportRequestHandler.cs
namadabu: I think that's a valid point. But the spec (https://github.com/HL7/bulk-data/blob/master/spec/export/index.md#bulk-data-status-request) recommends returning an error code for the `get status` request if the export fails. The way I interpreted it was that `get status` serves as a "proxy" to the export status and hence it is okay to return an error code even though the `get status` request itself succeeded. And since we are returning 500 to the end user, I decided to use that at the controller level also.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
namadabu: Since we will be returning data in `FailureReason` to the end-user, I want to avoid setting it to `ex.Message` because we can't know for certain what it's value will be in cases of unhandled exceptions.

src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/AzureExportDestinationClient.cs
```diff
@@ -48,12 +49,12 @@ public async Task ConnectAsync(string connectionSettings, CancellationToken canc
             }
             catch (Exception)
             {
-                throw new DestinationConnectionException(Resources.InvalidConnectionSettings);
+                throw new DestinationConnectionException(Resources.InvalidConnectionSettings, HttpStatusCode.BadRequest);
```
jackliums: We should consider validating this during queuing since there is no point queuing the job if the input is invalid.

src/Microsoft.Health.Fhir.Azure/KeyVault/KeyVaultSecretStore.cs
```diff
@@ -21,20 +25,67 @@ public class KeyVaultSecretStore : ISecretStore
...
 
         public async Task<SecretWrapper> GetSecretAsync(string secretName, CancellationToken cancellationToken)
         {
             EnsureArg.IsNotNullOrWhiteSpace(secretName);
 
-            SecretBundle result = await _keyVaultClient.GetSecretAsync(_keyVaultUri.AbsoluteUri, secretName, cancellationToken);
+            SecretBundle result;
+            try
+            {
+                result = await _retryPolicy.ExecuteAsync(async () =>
```
jackliums: nit: ExecuteAsync takes Func<Task> so this could just be simplified as `await _retryPolicy.ExecuteAsync(() => _keyVaultClient.GetSecretAsync(...))`.

src/Microsoft.Health.Fhir.Azure/KeyVault/KeyVaultSecretStore.cs
```diff
@@ -21,20 +25,67 @@ public class KeyVaultSecretStore : ISecretStore
...
+                {
+                    return await _keyVaultClient.GetSecretAsync(_keyVaultUri.AbsoluteUri, secretName, cancellationToken);
+                });
+            }
+            catch (KeyVaultErrorException kve)
+            {
+                HttpStatusCode statusCode = kve.Response != null ? kve.Response.StatusCode : HttpStatusCode.InternalServerError;
```
jackliums: I am not sure about passing through the KeyVault status code. I think we should return code like Unauthorized but any other code probably should just get categorized as InternalServerError?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -155,9 +156,9 @@ public async Task ExecuteAsync(ExportJobRecord exportJobRecord, WeakETag weakETa
                     // Best effort to delete the secret. If it fails to delete, then move on.
                     await _secretStore.DeleteSecretAsync(_exportJobRecord.SecretName, cancellationToken);
                 }
-                catch (Exception ex)
+                catch (SecretStoreException sse)
```
jackliums: nit: I think this should remain Exception since what we want is fire and forget.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ExportController.cs
```diff
@@ -89,6 +89,11 @@ public class ExportController : Controller
 
             CreateExportResponse response = await _mediator.ExportAsync(_fhirRequestContextAccessor.FhirRequestContext.Uri, destinationType, destinationConnectionString, HttpContext.RequestAborted);
 
+            if (!response.Successful)
```
jackliums: This could be thrown from the handler since it's not hosting layer specific.

src/Microsoft.Health.Fhir.Azure/KeyVault/KeyVaultSecretStore.cs
namadabu: Updated KeyVaultSecretStore to return only Unauthorized or Forbidden response codes and return InternalServerError for all other cases.

src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/AzureExportDestinationClient.cs
namadabu: Added a check to see whether string is base64 encoded at the filter attribute layer.

PR #604, opened 2019-08-02
Move exception handling for invalid continuation token.

LTA-Thinking:
## Description
Moved the exception handling for an invalid continuation token lower in the call stack and made it look at a specific exception type.

## Related issues
#453

### DIFF
--- a/src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/FhirRequestContextDocumentClientExtensions.cs
+++ b/src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/FhirRequestContextDocumentClientExtensions.cs
@@ -40,6 +40,10 @@ namespace Microsoft.Health.Fhir.CosmosDb.Features.Storage
                 {
                     throw new RequestRateExceededException(dce.RetryAfter);
                 }
+                else if (dce.Message.Contains("Invalid Continuation Token", StringComparison.OrdinalIgnoreCase))
+                {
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/FhirRequestContextDocumentClientExtensions.cs
```diff
@@ -40,6 +40,10 @@ public static void ProcessException(this IFhirRequestContext requestContext, Exc
                 {
                     throw new RequestRateExceededException(dce.RetryAfter);
                 }
+                else if (dce.Message.Contains("Invalid Continuation Token", StringComparison.OrdinalIgnoreCase))
```
brendankowitz: Only feedback I had was to ask if there was anything in `GetSubStatusCode` that could help with detecting this?

brendankowitz: https://github.com/microsoft/fhir-server/blob/master/src/Microsoft.Health.CosmosDb/Features/Storage/DocumentClientExceptionExtensions.cs

LTA-Thinking: Looking at the values in the ResponseHeaders through the debugger I don't see any value labeled as a sub-status, or any value that would help with detecting this.

Issue #605, opened 2019-08-03
Audit log should include caller's IP address


jackliums:
**User story**
As an administrator, I want to be able to review audit logs with caller ip address so that I can detect abnormal access patterns.

**Acceptance criteria**
1. When I see audit log, the IP address of the caller should be included.

PR #606, opened 2019-08-03
Add caller IP address to audit log.

jackliums:
## Description
- Add caller IP address to the audit log.
- Moved a few files around.

## Related issues
Addresses #605. Work item: [AB#69613](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/69613).

## Testing
Unit test, manual test.


jackliums:
I moved some of the files from Microsoft.Health.Fhir.Api to Microsoft.Health.Fhir.Shared.Api since this project has associated unit test projects and the files I have moved do have unit test associated with them.

src/Microsoft.Health.Fhir.Shared.Api/Utilities/CallerIpAddressRetriever.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        public string CallerIpAddress
+        {
+            get
+            {
+                // Determine the caller IP address.
+                // If the X-Forward-For header is supplied, use that; otherwise, get the IP from the HTTP context.
+                return _fhirRequestContextAccessor.FhirRequestContext.RequestHeaders.TryGetValue("X-Forwarded-For", out StringValues value) && value.Any() ?
+                    value.First() :
+                    _httpContextAccessor.HttpContext.Connection.RemoteIpAddress?.ToString();
```
staladay: Shouldn't you check to make sure Connection is not null before you check RemoteIpAddress.

jackliums: If the request reaches here, I don't think Connection will be null but I can add the extra check.

src/Microsoft.Health.Fhir.Shared.Api/Microsoft.Health.Fhir.Shared.Api.projitems
```diff
@@ -9,10 +9,17 @@
     <Import_RootNamespace>Microsoft.Health.Fhir.Shared.Api</Import_RootNamespace>
   </PropertyGroup>
   <ItemGroup>
+    <Compile Include="$(MSBuildThisFileDirectory)Controllers\AadSmartOnFhirProxyController.cs" />
```
brendankowitz: Why do these need to be moved to shared?

src/Microsoft.Health.Fhir.Shared.Api/Microsoft.Health.Fhir.Shared.Api.projitems
```diff
@@ -9,10 +9,17 @@
...
     <Compile Include="$(MSBuildThisFileDirectory)Controllers\ExportController.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Controllers\FhirController.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Features\ActionResults\FhirResult.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Features\ActionResults\OperationOutcomeResult.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)Features\Audit\AuditHelper.cs" />
```
brendankowitz: Why do these need to be moved to shared?

src/Microsoft.Health.Fhir.Core/Utilities/ICallerIpAddressRetriever.cs
```diff
@@ -0,0 +1,18 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Fhir.Core.Utilities
+{
+    /// <summary>
+    /// Provides functionality to retrieve caller IP address.
+    /// </summary>
+    public interface ICallerIpAddressRetriever
```
brendankowitz: `Accessor` looks to be the AspNet suffix where the interface has only a property

src/Microsoft.Health.Fhir.Shared.Api/Utilities/CallerIpAddressRetriever.cs
```diff
@@ -0,0 +1,41 @@
...
+        public CallerIpAddressRetriever(IFhirRequestContextAccessor fhirRequestContextAccessor, IHttpContextAccessor httpContextAccessor)
+        {
+            EnsureArg.IsNotNull(fhirRequestContextAccessor, nameof(fhirRequestContextAccessor));
+            EnsureArg.IsNotNull(httpContextAccessor, nameof(httpContextAccessor));
+
+            _fhirRequestContextAccessor = fhirRequestContextAccessor;
+            _httpContextAccessor = httpContextAccessor;
+        }
+
+        public string CallerIpAddress
```
brendankowitz: Is this already a thing with `UseForwardedHeaders`?
- https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.builder.forwardedheadersextensions.useforwardedheaders?view=aspnetcore-2.2#Microsoft_AspNetCore_Builder_ForwardedHeadersExtensions_UseForwardedHeaders_Microsoft_AspNetCore_Builder_IApplicationBuilder_
- https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/proxy-load-balancer?view=aspnetcore-2.2#troubleshoot

src/Microsoft.Health.Fhir.Shared.Api/Microsoft.Health.Fhir.Shared.Api.projitems
jackliums: I thought I left comment about the change but I guess I didn't.

The shared project (Microsoft.Health.Fhir.Shared.Api) has associated unit test projects (Microsoft.Health.Shared.Api.UnitTests) but the regular project doesn't. 

Some of the class in the Microsoft.Health.Fhir.Api had its unit tests in Microsoft.Health.Shared.Api.UnitTests which didn't seem right. So I moved those classes to the shared project. Alternatively, I would make another unit test project for Microsoft.health.Fhir.Api.

What's the guidance on what should go into shared and what shouldn't?

src/Microsoft.Health.Fhir.Shared.Api/Utilities/CallerIpAddressRetriever.cs
jackliums: :open_mouth: I didn't know it existed. That makes the code simpler.

src/Microsoft.Health.Fhir.Shared.Api/Microsoft.Health.Fhir.Shared.Api.projitems
jackliums: With the new changes, this is no longer needed so I reverted back.

src/Microsoft.Health.Fhir.Shared.Api/Utilities/CallerIpAddressRetriever.cs
jackliums: This was much easier to implement.

src/Microsoft.Health.Fhir.Shared.Api/Registration/FhirServerServiceCollectionExtensions.cs
```diff
@@ -45,18 +46,23 @@ public static class FhirServerServiceCollectionExtensions
                 options.RespectBrowserAcceptHeader = true;
             });
 
+            services.Configure<ForwardedHeadersOptions>(options =>
+            {
+                options.ForwardedHeaders = ForwardedHeaders.All;
```
brendankowitz: Are there any security implications to specifying all?

jackliums: That's a good question.

The 3 options are protocol, host, and ip address. We need protocol and ip address since that's what are we logging. But if we are forwarding those anyways, seems like we can forward host too.

We can bring this up during the security meeting on Friday.

jackliums:
Refactored the code a little bit to move the common logic to the AuditHelper so that individual logger doesn't have to repeat the same implementation (e.g., getting the caller IP address now happens in AuditHelper and so IAuditLogger will receive it as string so it can simply log it).

src/Microsoft.Health.Fhir.Api/Features/Audit/AuditHelper.cs
```diff
@@ -66,15 +63,21 @@ public string GetAuditEventType(string controllerName, string actionName)
...
-            Log(AuditAction.Executing, controllerName, actionName, statusCode: null, resourceType: null);
+            EnsureArg.IsNotNull(claimsExtractor, nameof(claimsExtractor));
+            EnsureArg.IsNotNull(httpContext, nameof(httpContext));
+
+            Log(AuditAction.Executing, controllerName, actionName, statusCode: null, resourceType: null, httpContext, claimsExtractor);
         }
 
         /// <inheritdoc />
-        public void LogExecuted(string controllerName, string actionName, HttpStatusCode statusCode, string responseResultType)
+        public void LogExecuted(string controllerName, string actionName, string responseResultType, HttpContext httpContext, IClaimsExtractor claimsExtractor)
```
brendankowitz: Why pass `HttpContext httpContext` and `IClaimsExtractor claimsExtractor` in over having them injected?

jackliums: We pass in `IClaimsExtractor` because we use different extractor depending on where the call is from. The regular FHIR APIs will extract the claims from principal whereas the Smart-On-Proxy controller extracts the claims from Form/Query.

For `HttpContext`, we could inject `IHttpContextAccessor` but all of the callers already have access to HttpContext already anyways, so I didn't feel like we needed to inject the accessor and get the context again.

Issue #12, opened 2018-09-17
Run search E2E tests in parallel


jackliums:
Right now, the parallel execution is disabled because search test needs to setup the data (by deleting known resources first). This is invasive since it's deleting all Patients or all Observation resources and requires the tests to be executed serially. One thing we could do is to attach custom identifiers to the search tests and update the search tests to include this identifier in the search itself. The downside is that string search becomes string search + token search (for identifier) but it should speed up the tests and will not accidentally wipe off data if executed against wrong database.

johnstairs:
IMO yes

CaitlinV39:
AB#76158

CaitlinV39:
Not prioritzed for now so were going to close.

PR #625, opened 2019-08-22
Split out HttpIntegrationTestFixture and introduce caching

johnstairs:
## Description
This changes the way we use and reuse `FhirClient`s across our end to end tests. The existing `HttpIntegrationTestFixture` has been broken apart into reusable pieces. The breakdown now looks like this:

- A `TestFhirServerFactory` is registered as an assembly fixture, meaning it can be reused across the tests in the assembly. It's job is to create and cache `TestFhirServer`.
- `TestFhirServer` is an abstract with two concrete implementations: `InProcTestFhirServer` and `RemoteTestFhirServer`. `TestFhirServer` creates and caches `FhirClient` instances that are setup to connect to the server in question (either one hosted in-proc or remote). 
- `HttpIntegrationTestFixture` now just calls `TestFhirServerFactory.GetTestFhirServer()` and `TestFhirServer.GetFhirClient()` to get a FHIR client for a test class.

I also changed the `FhirClient` API slightly for switching principals. Rather than mutating the instance, these methods will now return a new `FhirClient` instance using the desired identity.

Overall, this reduces E2E test execution time on a local box (where we use the in-proc test server) from ~4.5 minutes to ~1.5 minutes, a **3x** improvement.

We still have disabled parallelization within the E2E projects, so there is more we can do (see #12), but this is a good start.

### DIFF
--- a/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/TestFhirServer.cs
+++ b/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/TestFhirServer.cs
@@ -39,9 +39,9 @@ namespace Microsoft.Health.Fhir.Tests.E2E.Rest
             return GetFhirClient(format, TestApplications.ServiceClient, null, reusable);
         }
 
-        public FhirClient GetFhirClient(ResourceFormat format, TestApplication clientApplication, TestUser user, bool cacheable = true)
+        public FhirClient GetFhirClient(ResourceFormat format, TestApplication clientApplication, TestUser user, bool reusable = true)
         {
-            if (!cacheable)
+            if (!reusable)
             {
                 return CreateFhirClient(format, clientApplication, user);
             }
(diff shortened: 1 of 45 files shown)


test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/TestFhirServer.cs
```diff
@@ -0,0 +1,136 @@
...
+        }
+
+        public Uri BaseAddress { get; }
+
+        public FhirClient GetFhirClient(ResourceFormat format, bool reusable = true)
+        {
+            return GetFhirClient(format, TestApplications.ServiceClient, null, reusable);
+        }
+
+        public FhirClient GetFhirClient(ResourceFormat format, TestApplication clientApplication, TestUser user, bool cacheable = true)
```
brendankowitz: `cacheable` is called `reusable` above

PR #626, opened 2019-08-22
Organizing solution folders

brendankowitz:
## Description
Re-organizes solution folders to group project layers, feedback welcome:

image


johnstairs:
Maybe while you're at it, add folders under test for E2E and Integration

Issue #580, opened 2019-07-12
ArgumentException() thrown when search request contains "-_count"


jackliums:
**Describe the bug**
We have an unhandled exception during the GET Fhir/SearchByResourceType request flow. It ends up in a 500 response code. Seems like the incoming request contains a "-_count" parameter that is not a valid SearchParams.

**To Reproduce**
1. Search by using `Patient?_summary=_count`

**Expected behavior**
400 Bad Request should be thrown.

**Actual behavior**
500 Internal Server Error is thrown.

brendankowitz:
Should it be valid to use the int value of the enum? For instance `/Observation?_summary=3` is the same as `/Observation?_summary=count`. `/Observation?_summary=100` can be used, but is ignored.

brendankowitz:
Lets also add that using numeric values should also result in a 400

brendankowitz:
A 500 is also thrown when executed against the metadata endpoint:
https://msh-fhir-ci-r4.azurewebsites.net/metadata?_format=json&_summary=xyz

Ivanidzo4ka:
Do you mind if I take care of this one?

One thing which comes to my mind when I'm looking into current code and why it's happening it looks like main problem is in absence of validation for query parameters.
From what I see query parameters just get passed down to search without much of validation.
https://github.com/microsoft/fhir-server/blob/00d18530cdcdd2e2a56afe6a24455a4d3149e76a/src/Microsoft.Health.Fhir.Shared.Api/Controllers/FhirController.cs#L417

Is there are reason to push that to search itself instead of doing basic validation through model binding?
Would it be beneficial to have strongly typed object instead of `IReadOnlyList<Tuple<string, string>> queries` or strings give project some flexibility?

CaitlinV39:
AB#76055

PR #630, opened 2019-08-29
Throw Bad request exception if search param parsing fails.

deepak-ms:
## Description
Existing code flow in case of "?_summary=_count",
1. SearchOptionsFactory adds incorrect params to unsupportedParamList and proceeds ahead to retrieve results.
2. In Output formatter for response, exception is thrown while parsing SummaryType enum for query param.
3. Above exception results in 500 code.

With this change, SearchOptionsFactory throws BadRequestException (and SummaryType parsing is avoided in outputFormatter) which results in 400 code.

## Related issues
Addresses [issue #580].

## Testing
Verified manually and added E2E test.

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Api/Features/Formatters/FhirJsonOutputFormatter.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Api/Features/Formatters/FhirJsonOutputFormatter.cs
@@ -6,9 +6,11 @@
 using System;
 using System.Buffers;
 using System.IO;
+using System.Net;
 using System.Text;
 using EnsureThat;
 using Hl7.Fhir.Model;
+using Hl7.Fhir.Rest;
 using Hl7.Fhir.Serialization;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc.Formatters;
@@ -70,7 +72,10 @@ namespace Microsoft.Health.Fhir.Api.Features.Formatters
                     jsonWriter.Formatting = Formatting.Indented;
                 }
 
-                _fhirJsonSerializer.Serialize((Resource)context.Object, jsonWriter, context.HttpContext.GetSummaryType(_logger));
+                _fhirJsonSerializer.Serialize(
+                    (Resource)context.Object,
+                    jsonWriter,
+                    (context.HttpContext.Response.StatusCode == (int)HttpStatusCode.OK) ? context.HttpContext.GetSummaryType(_logger) : SummaryType.False);
                 await jsonWriter.FlushAsync();
             }
         }
--- a/src/Microsoft.Health.Fhir.Shared.Api/Features/Formatters/FhirJsonOutputFormatter.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Api/Features/Formatters/FhirJsonOutputFormatter.cs
@@ -6,11 +6,9 @@
 using System;
 using System.Buffers;
 using System.IO;
-using System.Net;
 using System.Text;
 using EnsureThat;
 using Hl7.Fhir.Model;
-using Hl7.Fhir.Rest;
 using Hl7.Fhir.Serialization;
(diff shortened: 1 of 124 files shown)


deepak-ms:
Note that status 200 is returned right now for requests like "/Observation?_count=", "/Observation?_summary=" or "/Observation?=" (basically if any param is empty), and status 400 is returned if the params are non-empty and invalid (e.g. "/Observation?_count=abc".

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
```diff
@@ -98,10 +99,7 @@ public SearchOptions Create(string compartmentType, string compartmentId, string
                     }
                     catch (Exception ex)
                     {
-                        _logger.LogInformation(ex, "Failed to parse the query parameter. Skipping.");
-
-                        // There was a problem parsing the parameter. Add it to list of unsupported parameters.
-                        unsupportedSearchParameters.Add(query);
+                        throw new BadRequestException(ex.Message);
```
brandonpollett: Does this change the behavior in the case where we just don't support a search parameter? 

For example, if somebody were to send `Location?type1=test` would this return an error?

brandonpollett: Also, there's some discussion around this in the spec: https://www.hl7.org/fhir/search.html#errors

deepak-ms: I think it changes the behavior for only those requests which are syntactically incorrect (validation failed by HL7 package). "Location?type1=test" still returns result (with 200 OK) ignoring the unsupported params.

src/Microsoft.Health.Fhir.Shared.Api/Features/Formatters/FhirJsonOutputFormatter.cs
```diff
@@ -70,7 +72,10 @@ public override async Task WriteResponseBodyAsync(OutputFormatterWriteContext co
                     jsonWriter.Formatting = Formatting.Indented;
                 }
 
-                _fhirJsonSerializer.Serialize((Resource)context.Object, jsonWriter, context.HttpContext.GetSummaryType(_logger));
+                _fhirJsonSerializer.Serialize(
+                    (Resource)context.Object,
+                    jsonWriter,
+                    (context.HttpContext.Response.StatusCode == (int)HttpStatusCode.OK) ? context.HttpContext.GetSummaryType(_logger) : SummaryType.False);
```
brendankowitz: I feel like this logic should be resolved in `GetSummaryType`. Perhaps in this case if we're already validating the logic we change `GetSummaryType` to `TryParse` of the enum and don't throw

deepak-ms: I agree to the logic to get resolved inside GetSummaryType if that's meant only for this specific use case (which is actually the case right now). I think handling the logic outside allows GetSummaryType to be used in other scenarios as well (which expect exception in case of invalid value). In last discussion, Jack was inclined towards latter (current change). Let's discuss it together when he's back.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
jackliums: Can we put comment here just to describe this briefly? I think it will help whoever reading the code.

src/Microsoft.Health.Fhir.Shared.Api/Features/Formatters/FhirJsonOutputFormatter.cs
deepak-ms: Moved the check inside GetSummaryType.

PR #631, opened 2019-08-29
Fixed potential nullref when throwing OperationFailedException.

namadabu:
## Description
We are trying to access `FailureDetails` in `ExportJobRecord` without doing a null check first.

## Related issues
Addresses [issue [AB#70143](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/70143)].

## Testing
Added relevant UT

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Operations/Export/GetExportRequestHandler.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Operations/Export/GetExportRequestHandler.cs
@@ -48,9 +48,11 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Export
             }
             else if (outcome.JobRecord.Status == OperationStatus.Failed || outcome.JobRecord.Status == OperationStatus.Canceled)
             {
+                string failureReason = outcome.JobRecord.FailureDetails != null ? outcome.JobRecord.FailureDetails.FailureReason : Resources.UnknownError;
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Fhir.Core/Features/Operations/Export/GetExportRequestHandler.cs
```diff
@@ -48,9 +48,11 @@ public async Task<GetExportResponse> Handle(GetExportRequest request, Cancellati
             }
             else if (outcome.JobRecord.Status == OperationStatus.Failed || outcome.JobRecord.Status == OperationStatus.Canceled)
             {
+                string failureReason = outcome.JobRecord.FailureDetails != null ? outcome.JobRecord.FailureDetails.FailureReason : Resources.UnknownError;
+                HttpStatusCode failureStatusCode = outcome.JobRecord.FailureDetails != null ? outcome.JobRecord.FailureDetails.FailureStatusCode : HttpStatusCode.InternalServerError;
```
feordin: Is it the desired behavior to return a 500 error?

namadabu: When a job fails, the expectation is that the `FailureDetails` data is populated with the appropriate status code and we return that to the client. In the unexpected case that `FailureDetails` is not populated, we are defaulting to 500

PR #637, opened 2019-09-05
Add customer request headers to audit log

feordin:
## Description
Add the capability to store custom information in the audit logs by including headers in the request with a unique prefix.

## Related issues
Addresses [issue [AB#69944](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/69944)].

## Testing
New automated tests added, as well as manual testing.

### DIFF
--- a/src/Microsoft.Health.Fhir.Api/Features/Audit/AuditConstants.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/Audit/AuditConstants.cs
@@ -1,0 +1,18 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Fhir.Api.Features.Audit
+{
+    public static class AuditConstants
+    {
+        public const string CustomAuditHeaderPrefix = "X-MS-HEALTHCAREAPIS-AUDIT-";
+
+        public const string CustomAuditHeaderKeyValue = "CustomAuditHeaderCollection";
+
+        public const int MaximumNumberOfCustomHeaders = 10;
+
+        public const int MaximumLengthOfCustomHeader = 2048;
+    }
+}
--- a/src/Microsoft.Health.Fhir.Api/Features/Audit/AuditHeaderReader.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/Audit/AuditHeaderReader.cs
@@ -1,0 +1,48 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.Health.Fhir.Api.Features.Audit
+{
+    public class AuditHeaderReader : IAuditHeaderReader
+    {
+        public IReadOnlyDictionary<string, string> Read(HttpContext httpContext)
+        {
+            if (!httpContext.Items.ContainsKey(AuditConstants.CustomAuditHeaderKeyValue))
+            {
+                var customHeaders = new Dictionary<string, string>();
+
+                foreach (var headerName in httpContext.Request.Headers.Keys)
+                {
+                    if (headerName.StartsWith(AuditConstants.CustomAuditHeaderPrefix, StringComparison.OrdinalIgnoreCase))
+                    {
+                        var headerValue = httpContext.Request.Headers[headerName].ToString();
+                        if (headerValue.Length > AuditConstants.MaximumLengthOfCustomHeader)
+                        {
+                            throw new AuditHeaderException(headerName, headerValue.Length);
+                        }
+
+                        customHeaders[headerName] = headerValue;
+                    }
+                }
+
+                if (customHeaders.Count > 10)
+                {
+                    throw new AuditHeaderException(customHeaders.Count);
+                }
+
+                httpContext.Items[AuditConstants.CustomAuditHeaderKeyValue] = customHeaders;
+                return customHeaders;
+            }
+            else
+            {
+                return httpContext.Items[AuditConstants.CustomAuditHeaderKeyValue] as IReadOnlyDictionary<string, string>;
+            }
+        }
+    }
+}
--- a/src/Microsoft.Health.Fhir.Api/Features/Audit/AuditHeaderReader.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/Audit/AuditHeaderReader.cs
@@ -6,11 +6,20 @@
 using System;
 using System.Collections.Generic;
 using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Options;
+using Microsoft.Health.Fhir.Core.Configs;
 
 namespace Microsoft.Health.Fhir.Api.Features.Audit
 {
     public class AuditHeaderReader : IAuditHeaderReader
     {
+        private readonly AuditConfiguration _auditConfiguration;
+
+        public AuditHeaderReader(IOptions<AuditConfiguration> auditConfiguration)
+        {
+            _auditConfiguration = auditConfiguration.Value;
+        }
+
         public IReadOnlyDictionary<string, string> Read(HttpContext httpContext)
         {
             if (!httpContext.Items.ContainsKey(AuditConstants.CustomAuditHeaderKeyValue))
@@ -19,7 +28,7 @@ namespace Microsoft.Health.Fhir.Api.Features.Audit
 
                 foreach (var headerName in httpContext.Request.Headers.Keys)
                 {
-                    if (headerName.StartsWith(AuditConstants.CustomAuditHeaderPrefix, StringComparison.OrdinalIgnoreCase))
+                    if (headerName.StartsWith(_auditConfiguration.CustomAuditHeaderPrefix, StringComparison.OrdinalIgnoreCase))
                     {
                         var headerValue = httpContext.Request.Headers[headerName].ToString();
                         if (headerValue.Length > AuditConstants.MaximumLengthOfCustomHeader)
@@ -31,7 +40,7 @@ namespace Microsoft.Health.Fhir.Api.Features.Audit
                     }
                 }
 
-                if (customHeaders.Count > 10)
+                if (customHeaders.Count > AuditConstants.MaximumNumberOfCustomHeaders)
                 {
                     throw new AuditHeaderException(customHeaders.Count);
                 }
--- a/src/Microsoft.Health.Fhir.Api/Features/Audit/AuditHeaderReader.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/Audit/AuditHeaderReader.cs
@@ -5,8 +5,10 @@
 
 using System;
 using System.Collections.Generic;
+using EnsureThat;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Options;
+using Microsoft.Extensions.Primitives;
 using Microsoft.Health.Fhir.Core.Configs;
 
 namespace Microsoft.Health.Fhir.Api.Features.Audit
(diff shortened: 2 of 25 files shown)


src/Microsoft.Health.Fhir.Api/Features/Audit/AuditHeaderReader.cs
```diff
@@ -0,0 +1,48 @@
...
+                        if (headerValue.Length > AuditConstants.MaximumLengthOfCustomHeader)
+                        {
+                            throw new AuditHeaderException(headerName, headerValue.Length);
+                        }
+
+                        customHeaders[headerName] = headerValue;
+                    }
+                }
+
+                if (customHeaders.Count > 10)
```
deepak-ms: use MaximumNumberOfCustomHeaders

src/Microsoft.Health.Fhir.Api/Resources.resx
```diff
@@ -123,6 +123,9 @@
   <data name="ContentTypeHeaderRequired" xml:space="preserve">
     <value>The "content-type" header is required.</value>
   </data>
+  <data name="CustomAuditHeaderTooLarge" xml:space="preserve">
+    <value>The value of the header {0} is  {1} and is too long so will not be added to the audit logs.  The maximum allowed size is {2}.  Please retry the request with a smaller value.</value>
```
LTA-Thinking: Nit: Double space after '{0} is' and before '{1}'

src/Microsoft.Health.Fhir.Api/Features/Audit/AuditConstants.cs
```diff
@@ -0,0 +1,18 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Fhir.Api.Features.Audit
+{
+    public static class AuditConstants
+    {
+        public const string CustomAuditHeaderPrefix = "X-MS-HEALTHCAREAPIS-AUDIT-";
```
brendankowitz: I don't think this should be in OSS.
@jackliums @hansenms thoughts?

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Audit/AuditHelperTests.cs
```diff
@@ -46,6 +46,7 @@ public class AuditHelperTests
 
         private readonly HttpContext _httpContext = new DefaultHttpContext();
         private readonly IClaimsExtractor _claimsExtractor = Substitute.For<IClaimsExtractor>();
+        private readonly IAuditHeaderReader _auditHeaderReader = new AuditHeaderReader();
```
brandonpollett: Any reason we are not substituting this?

src/Microsoft.Health.Fhir.Api/Resources.resx
```diff
@@ -179,6 +182,9 @@
   <data name="ToggleNavigation" xml:space="preserve">
     <value>Toggle navigation</value>
   </data>
+  <data name="TooManyCustomAuditHeaders" xml:space="preserve">
+    <value>A maximum of {0} custom audit headers allowed in the request.  {1} headers were detected.  Please retry the request with fewer cutom headers.</value>
```
cunninghamjc: Typo: "cutom" -> "custom"

src/Microsoft.Health.Fhir.Api/Resources.resx
```diff
@@ -123,6 +123,9 @@
   <data name="ContentTypeHeaderRequired" xml:space="preserve">
     <value>The "content-type" header is required.</value>
   </data>
+  <data name="CustomAuditHeaderTooLarge" xml:space="preserve">
+    <value>The value of the header {0} is {1} and is too long so will not be added to the audit logs.  The maximum allowed size is {2}.  Please retry the request with a smaller value.</value>
```
cunninghamjc: It looks like {1} is the size of the header value, so this text would read something like, "The value of the header X-FOO is 13789 and is too long."  That seems a little awkward/confusing.  Should it be, "The size of the header {0} is {1}"?

src/Microsoft.Health.Fhir.Api/Features/Audit/AuditHeaderReader.cs
```diff
@@ -0,0 +1,57 @@
...
+using Microsoft.Extensions.Options;
+using Microsoft.Health.Fhir.Core.Configs;
+
+namespace Microsoft.Health.Fhir.Api.Features.Audit
+{
+    public class AuditHeaderReader : IAuditHeaderReader
+    {
+        private readonly AuditConfiguration _auditConfiguration;
+
+        public AuditHeaderReader(IOptions<AuditConfiguration> auditConfiguration)
```
jackliums: nit: null check?

src/Microsoft.Health.Fhir.Api/Features/Audit/AuditHeaderReader.cs
```diff
@@ -0,0 +1,57 @@
...
+            _auditConfiguration = auditConfiguration.Value;
+        }
+
+        public IReadOnlyDictionary<string, string> Read(HttpContext httpContext)
+        {
+            if (!httpContext.Items.ContainsKey(AuditConstants.CustomAuditHeaderKeyValue))
+            {
+                var customHeaders = new Dictionary<string, string>();
+
+                foreach (var headerName in httpContext.Request.Headers.Keys)
```
jackliums: nit: if you do `foreach(KeyValuePair<string, StringValues> header in httpContext.Request.Headers)`, then you can access the key and the value without additional lookup.

src/Microsoft.Health.Fhir.Api/Features/Audit/AuditHeaderReader.cs
```diff
@@ -0,0 +1,57 @@
...
+        private readonly AuditConfiguration _auditConfiguration;
+
+        public AuditHeaderReader(IOptions<AuditConfiguration> auditConfiguration)
+        {
+            _auditConfiguration = auditConfiguration.Value;
+        }
+
+        public IReadOnlyDictionary<string, string> Read(HttpContext httpContext)
+        {
+            if (!httpContext.Items.ContainsKey(AuditConstants.CustomAuditHeaderKeyValue))
```
jackliums: nit: prefer TryGetValue

src/Microsoft.Health.Fhir.Api/Features/Audit/AuditLogger.cs
```diff
@@ -31,7 +31,8 @@ public class AuditLogger : IAuditLogger
             "Action: {Action}" + Environment.NewLine +
             "StatusCode: {StatusCode}" + Environment.NewLine +
             "CorrelationId: {CorrelationId}" + Environment.NewLine +
-            "Claims: {Claims}";
+            "Claims: {Claims}" + Environment.NewLine +
+            "CustomHeaderValues: {CustomHeaderValues}";
```
jackliums: I think to be on the safe side, we probably don't want to output this here in the default logger just like how we don't output IP address in this implementation.

src/Microsoft.Health.Fhir.Api/Features/Audit/AuditHeaderReader.cs
```diff
@@ -0,0 +1,57 @@
...
+    public class AuditHeaderReader : IAuditHeaderReader
+    {
+        private readonly AuditConfiguration _auditConfiguration;
+
+        public AuditHeaderReader(IOptions<AuditConfiguration> auditConfiguration)
+        {
+            _auditConfiguration = auditConfiguration.Value;
+        }
+
+        public IReadOnlyDictionary<string, string> Read(HttpContext httpContext)
```
jackliums: nit: null check

src/Microsoft.Health.Fhir.Api/Resources.resx
```diff
@@ -179,6 +182,9 @@
   <data name="ToggleNavigation" xml:space="preserve">
     <value>Toggle navigation</value>
   </data>
+  <data name="TooManyCustomAuditHeaders" xml:space="preserve">
+    <value>A maximum of {0} custom audit headers allowed in the request.  {1} headers were detected.  Please retry the request with fewer custom headers.</value>
+  </data>
```
jackliums: Leave some comment about what {0} and {1} would be for the translators. An example would be nice.

src/Microsoft.Health.Fhir.Api/Resources.resx
```diff
@@ -123,6 +123,9 @@
   <data name="ContentTypeHeaderRequired" xml:space="preserve">
     <value>The "content-type" header is required.</value>
   </data>
+  <data name="CustomAuditHeaderTooLarge" xml:space="preserve">
+    <value>The size of the header {0} is {1}.  The maximum allowed size is {2}.  Please retry the request with a smaller value.</value>
```
jackliums: The message here and below should be a bit more consistent. Maybe..

The maximum length of a custom audit header value is {0}. The supplied custom audit header '{1}' has length of {2}.
The maximum number of custom audit headers allowed is {0}. The number of custom audit headers supplied is {1}.

src/Microsoft.Health.Fhir.Core/Configs/AuditConfiguration.cs
```diff
@@ -0,0 +1,12 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Fhir.Core.Configs
+{
+    public class AuditConfiguration
+    {
+        public string CustomAuditHeaderPrefix { get; set; } = "X-MS-AZUREFHIR-AUDIT-";
```
jackliums: We should have validation to make sure the prefix is not empty (E.g., I cannot supply empty string in the config file).

src/Microsoft.Health.Fhir.Api/Resources.resx
JoeDRowan: Agreed - Explaining what the placeholders are allows translators to have context when translating. Remember to not actually use curly braces as part of the explanation in the Comments field as they would then be interpreted as rules to be parsed - Instead refer to placeholders as (say) PH0, PH1. Also add the following rule to ensure that the placeholders remain in the translation: {NumberedPlaceHolder=”{0}","{1}”} Ref: aka.ms/NExTLoc

JoeDRowan: This convention for explaining and controlling placeholders follows throughout this file @feordin

src/Microsoft.Health.Fhir.Api/Resources.resx
```diff
@@ -123,6 +123,10 @@
   <data name="ContentTypeHeaderRequired" xml:space="preserve">
     <value>The "content-type" header is required.</value>
   </data>
+  <data name="CustomAuditHeaderTooLarge" xml:space="preserve">
+    <value>The maximum length of a custom audit header value is {0}. The supplied custom audit header '{1}' has length of {2}.</value>
+    <comment>PH0 is a constant defining the max length of a header.  PH1 is the name  of the header sent in the request.  PH2 is the length of the incoming header value.  Example: The maximum length of a custom audit header value is 2048. The supplied custom audit header 'X-MS-AZUREFHIR-AUDIT-SITE' has length of 3072.</comment>
```
brendankowitz: In other comments we've used `{0}` instead of `PH0`

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Audit/AuditHeaderReaderTests.cs
```diff
@@ -0,0 +1,211 @@
...
+using Microsoft.Extensions.Primitives;
+using Microsoft.Health.Fhir.Api.Features.Audit;
+using Microsoft.Health.Fhir.Core.Configs;
+using Microsoft.Health.Fhir.Core.Exceptions;
+using NSubstitute;
+using Xunit;
+
+namespace Microsoft.Health.Fhir.Api.UnitTests.Features.Audit
+{
+    public class AuditHeaderReaderTests
```
brendankowitz: Most other tests are following the BDD style test naming convention GivenX_WhenY_ThenZ. Might be worth quickly revising these

src/Microsoft.Health.Fhir.Api/Resources.resx
JoeDRowan: This is actually correct @brendankowitz - Using curly braces { } in a comment will get interpreted by the Localization engine as a rule to be interpreted e.g. {Locked} {NumberedPlaceholder="{0}"} etc. So to refer to curly braces but not as rules, we cannot use { } and the convention set is to refer to them as PH0, PH1 etc.  If other places have used { } in comments that are not rules, they should change to PH0...

brendankowitz: Thanks for clarification @JoeDRowan, just calling this out because the file is not consistent

PR #643, opened 2019-09-09
Returns 412 error for incorrect if-match header in R4

rotodd:
## Description

- In R4, the server returns a 412 Precondition Failed status code if the version id given in the If-Match header does not match (see [R4 spec](http://hl7.org/fhir/http.html#concurrency))
r4_response
- In STU3, the server returns a 409 Conflict status code if the version id given in the If-Match header does not match (see [STU3 spec](http://hl7.org/fhir/STU3/http.html#concurrency))
stu3_response

## Related issues

- [#AB69420](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/69420)

## Testing

- Separated out existing E2E tests to account for version-dependent behavior
- Added new E2E and integration tests

### DIFF
--- a/src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
+++ b/src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
@@ -126,7 +126,7 @@ namespace Microsoft.Health.Fhir.CosmosDb.Features.Storage
                 switch (dce.GetSubStatusCode())
                 {
                     case HttpStatusCode.PreconditionFailed:
-                        throw new ResourceConflictException(weakETag);
+                        throw new PreconditionFailedException(string.Format(Core.Resources.ResourceVersionConflict, weakETag?.VersionId));
                     case HttpStatusCode.NotFound:
                         if (cosmosWrapper.IsDeleted)
                         {
--- a/src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
+++ b/src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
@@ -135,7 +135,7 @@ namespace Microsoft.Health.Fhir.CosmosDb.Features.Storage
 
                         if (weakETag != null)
                         {
-                            throw new ResourceConflictException(weakETag);
+                            throw new ResourceNotFoundException(string.Format(Core.Resources.ResourceNotFoundByIdAndVersion, resource.ResourceTypeName, resource.ResourceId, weakETag.VersionId));
                         }
                         else if (!allowCreate)
                         {
--- a/src/Microsoft.Health.Fhir.R4.Core/Features/Resources/Upsert/UpsertResourceHandler.cs
+++ b/src/Microsoft.Health.Fhir.R4.Core/Features/Resources/Upsert/UpsertResourceHandler.cs
@@ -24,6 +24,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Resources.Upsert
             }
             catch (ResourceConflictException)
             {
+                // In R4, the server returns a 412 Precondition Failed status code if the version id given in the If-Match header does not match
                 throw new PreconditionFailedException(string.Format(Core.Resources.ResourceVersionConflict, message.WeakETag?.VersionId));
             }
 
--- a/src/Microsoft.Health.Fhir.R4.Core/Features/Resources/Upsert/UpsertResourceHandler.cs
+++ b/src/Microsoft.Health.Fhir.R4.Core/Features/Resources/Upsert/UpsertResourceHandler.cs
@@ -1,0 +1,33 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Threading;
+using System.Threading.Tasks;
+using MediatR;
+using Microsoft.Health.Fhir.Core.Exceptions;
+using Microsoft.Health.Fhir.Core.Features.Persistence;
+using Microsoft.Health.Fhir.Core.Messages.Upsert;
+
+namespace Microsoft.Health.Fhir.Core.Features.Resources.Upsert
+{
+    public partial class UpsertResourceHandler : BaseResourceHandler, IRequestHandler<UpsertResourceRequest, UpsertResourceResponse>
+    {
+        private async Task<UpsertOutcome> UpsertAsync(UpsertResourceRequest message, ResourceWrapper resourceWrapper, bool allowCreate, bool keepHistory, CancellationToken cancellationToken)
+        {
+            UpsertOutcome result;
+
+            try
+            {
+                result = await FhirDataStore.UpsertAsync(resourceWrapper, message.WeakETag, allowCreate, keepHistory, cancellationToken);
+            }
+            catch (ResourceConflictException)
+            {
+                throw new PreconditionFailedException(string.Format(Core.Resources.ResourceVersionConflict, message.WeakETag?.VersionId));
+            }
+
+            return result;
+        }
+    }
+}
--- a/src/Microsoft.Health.Fhir.R4.Core/Features/Resources/Upsert/UpsertResourceHandler.cs
+++ b/src/Microsoft.Health.Fhir.R4.Core/Features/Resources/Upsert/UpsertResourceHandler.cs
@@ -6,7 +6,6 @@
 using System.Threading;
 using System.Threading.Tasks;
 using MediatR;
-using Microsoft.Health.Fhir.Core.Exceptions;
 using Microsoft.Health.Fhir.Core.Features.Persistence;
 using Microsoft.Health.Fhir.Core.Messages.Upsert;
 
@@ -16,17 +15,8 @@ namespace Microsoft.Health.Fhir.Core.Features.Resources.Upsert
     {
         private async Task<UpsertOutcome> UpsertAsync(UpsertResourceRequest message, ResourceWrapper resourceWrapper, bool allowCreate, bool keepHistory, CancellationToken cancellationToken)
         {
-            UpsertOutcome result;
-
-            try
-            {
-                result = await FhirDataStore.UpsertAsync(resourceWrapper, message.WeakETag, allowCreate, keepHistory, cancellationToken);
-            }
-            catch (ResourceConflictException)
-            {
-                // In R4, the server returns a 412 Precondition Failed status code if the version id given in the If-Match header does not match
-                throw new PreconditionFailedException(string.Format(Core.Resources.ResourceVersionConflict, message.WeakETag?.VersionId));
-            }
+            // In R4, the server returns a 412 Precondition Failed status code if the version id given in the If-Match header does not match
+            UpsertOutcome result = await FhirDataStore.UpsertAsync(resourceWrapper, message.WeakETag, allowCreate, keepHistory, cancellationToken);
 
             return result;
         }
--- a/src/Microsoft.Health.Fhir.R4.Core/Features/Resources/Upsert/UpsertResourceHandler.cs
+++ b/src/Microsoft.Health.Fhir.R4.Core/Features/Resources/Upsert/UpsertResourceHandler.cs
@@ -15,10 +15,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Resources.Upsert
     {
         private async Task<UpsertOutcome> UpsertAsync(UpsertResourceRequest message, ResourceWrapper resourceWrapper, bool allowCreate, bool keepHistory, CancellationToken cancellationToken)
         {
-            // In R4, the server returns a 412 Precondition Failed status code if the version id given in the If-Match header does not match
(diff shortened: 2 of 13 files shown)


johnstairs:
This looks good. We just need to keep in mind that we will need to rework the implementation of version-specific behaviors when we move away from shared projects.

src/Microsoft.Health.Fhir.R4.Core/Features/Resources/Upsert/UpsertResourceHandler.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        private async Task<UpsertOutcome> UpsertAsync(UpsertResourceRequest message, ResourceWrapper resourceWrapper, bool allowCreate, bool keepHistory, CancellationToken cancellationToken)
+        {
+            UpsertOutcome result;
+
+            try
+            {
+                result = await FhirDataStore.UpsertAsync(resourceWrapper, message.WeakETag, allowCreate, keepHistory, cancellationToken);
+            }
+            catch (ResourceConflictException)
```
brendankowitz: It might be better to rewire this to have the correct behavior for R4 and alter it only for Stu3. So the UpsertAsync method in the `*FhirDataStore` throws the `PreconditionFailedException` and the Stu3 handler provides the backwards compatibility. Thoughts?

rotodd: Makes sense, fixed in db09f23.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
```diff
@@ -126,7 +126,7 @@ public sealed class CosmosFhirDataStore : IFhirDataStore, IProvideCapability
                 switch (dce.GetSubStatusCode())
                 {
                     case HttpStatusCode.PreconditionFailed:
-                        throw new ResourceConflictException(weakETag);
+                        throw new PreconditionFailedException(string.Format(Core.Resources.ResourceVersionConflict, weakETag?.VersionId));
                     case HttpStatusCode.NotFound:
                         if (cosmosWrapper.IsDeleted)
                         {
```
rotodd: @brendankowitz the spec doesn't seem to clearly outline how to handle the scenario where an incorrect if-match header is provided for a resource that does not exist. My interpretation would be that we should return a 404 error code on line 138 below, but I could also see how 409/412 might make sense.

jackliums: I think that interpretation is correct. If the resource doesn't exist, we shoudl fail with 404. Only if the resource exists, then we will take if-match header into consideration.

src/Microsoft.Health.Fhir.R4.Core/Features/Resources/Upsert/UpsertResourceHandler.cs
```diff
@@ -0,0 +1,24 @@
...
+using Microsoft.Health.Fhir.Core.Features.Persistence;
+using Microsoft.Health.Fhir.Core.Messages.Upsert;
+
+namespace Microsoft.Health.Fhir.Core.Features.Resources.Upsert
+{
+    public partial class UpsertResourceHandler : BaseResourceHandler, IRequestHandler<UpsertResourceRequest, UpsertResourceResponse>
+    {
+        private async Task<UpsertOutcome> UpsertAsync(UpsertResourceRequest message, ResourceWrapper resourceWrapper, bool allowCreate, bool keepHistory, CancellationToken cancellationToken)
+        {
+            // In R4, the server returns a 412 Precondition Failed status code if the version id given in the If-Match header does not match
```
jackliums: Is this for R4 only?

http://hl7.org/fhir/STU3/http.html#concurrency seems to indicate that it should also return 412.

rotodd: I'm fairly confident it's just for R4 - in the section that you linked to in the Stu3 spec, there's a line that says:

> If the version id given in the If-Match header does not match, the server returns a 409 Conflict status code instead of updating the resource

jackliums: oh you are right. I missed that part.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
rotodd: Mini update, I came across this comment in the spec:

> The update interaction creates a new current version for an existing resource or creates an initial version if no resource already exists for the given id.

So just to clarify, if a correct if-match header is provided in an update call for a resource that does not exist, a new resource should be created. The if-match header would have to specify a version of 1 in this case.

If an incorrect if-match header is provided in an update call for resource that does not exist, I'm not sure if I should return 412 (for R4) and 409 (for Stu3) or if I should ignore the if-match header and create a resource with version 1.

Let me know what you think @jackliums @brendankowitz

src/Microsoft.Health.Fhir.R4.Core/Features/Resources/Upsert/UpsertResourceHandler.cs
johnstairs: This comment makes sense in the context of this code review, but seems out of place right here in the code. We're not doing anything special for this scenario in the code that follows.

src/Microsoft.Health.Fhir.Stu3.Core/Features/Resources/Upsert/UpsertResourceHandler.cs
```diff
@@ -0,0 +1,35 @@
...
+    public partial class UpsertResourceHandler : BaseResourceHandler, IRequestHandler<UpsertResourceRequest, UpsertResourceResponse>
+    {
+        private async Task<UpsertOutcome> UpsertAsync(UpsertResourceRequest message, ResourceWrapper resourceWrapper, bool allowCreate, bool keepHistory, CancellationToken cancellationToken)
+        {
+            // In STU3, the server returns a 409 Conflict status code if the version id given in the If-Match header does not match
+            UpsertOutcome result;
+
+            try
+            {
+                result = await FhirDataStore.UpsertAsync(resourceWrapper, message.WeakETag, allowCreate, keepHistory, cancellationToken);
```
johnstairs: I would just return here and not have the variable.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -120,14 +120,28 @@ public async Task<UpsertOutcome> UpsertAsync(ResourceWrapper resource, WeakETag
...
+                                    throw new MethodNotAllowedException(Core.Resources.ResourceCreationNotAllowed);
+                                }
+
+                                break;
                             default:
                                 _logger.LogError(e, "Error from SQL database on upsert");
                                 throw;
                         }
+
+                        _logger.LogError(e, "Unhandled Document Client Exception");
```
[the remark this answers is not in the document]
rotodd: Fixed in 68e0f8f.

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/FhirStorageTests.cs
```diff
@@ -143,6 +143,34 @@ public async Task GivenASavedResource_WhenUpserting_ThenTheExistingResourceIsUpd
...
+
+        [Theory]
+        [InlineData("1")]
+        [InlineData("-1")]
+        [InlineData("0")]
+        public async Task WhenUpsertingWithCreateDisabledAndIntegerETagHeader_GivenANonexistentResource_TheServerShouldReturnResourceNotFoundResponse(string versionId)
+        {
+            var observation = _conformance.Rest[0].Resource.Find(r => r.Type == ResourceType.Observation);
+            observation.UpdateCreate = false;
+            observation.Versioning = CapabilityStatement.ResourceVersionPolicy.Versioned;
```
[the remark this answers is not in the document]
rotodd: Good idea, fixed in a4f9a07.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -120,14 +120,28 @@ public async Task<UpsertOutcome> UpsertAsync(ResourceWrapper resource, WeakETag
                     {
                         switch (e.Number)
                         {
-                            case SqlErrorCodes.NotFound:
```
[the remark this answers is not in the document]
rotodd: Let me know what you think of 45d574d.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -120,14 +120,25 @@ public async Task<UpsertOutcome> UpsertAsync(ResourceWrapper resource, WeakETag
...
                         {
+                            case SqlErrorCodes.PreconditionFailed:
+                                throw new PreconditionFailedException(string.Format(Core.Resources.ResourceVersionConflict, weakETag?.VersionId));
                             case SqlErrorCodes.NotFound:
+                                if (weakETag != null)
+                                {
+                                    throw new ResourceNotFoundException(string.Format(Core.Resources.ResourceNotFoundByIdAndVersion, resource.ResourceTypeName, resource.ResourceId, weakETag.VersionId));
+                                }
+
+                                break;
```
johnstairs: you can do `goto default` instead.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -68,7 +68,7 @@ public async Task<UpsertOutcome> UpsertAsync(ResourceWrapper resource, WeakETag
             int etag = 0;
             if (weakETag != null && !int.TryParse(weakETag.VersionId, out etag))
             {
-                throw new ResourceConflictException(weakETag);
+                throw new PreconditionFailedException(string.Format(Core.Resources.ResourceVersionConflict, weakETag?.VersionId));
```
johnstairs: I guess one issue is if this is provided but not an integer, we throw, but we don't know what the error should be. We could set the etag to a sentinel value (like -1) and send it though.

rotodd: Oh, good idea. Looking into this now - I think I could potentially get the Cosmos/Sql behavior to match this way.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
rotodd: Nice, fixed in 5636452.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
rotodd: Implemented in 7649951 and updated the [expected outcomes summary sheet](https://microsoft-my.sharepoint.com/:x:/g/personal/rotodd_microsoft_com1/EVWfudFyn4JFnT5o5LJrlNMBig2YlTSSbeRb-0WDmj4DVQ?e=0y4PVh).

test/Microsoft.Health.Fhir.Stu3.Tests.Integration/Persistence/FhirStorageTests.cs
```diff
@@ -0,0 +1,52 @@
...
+
+            var newResourceValues = Samples.GetJsonSample("WeightInGrams").ToPoco();
+            newResourceValues.Id = saveResult.Resource.Id;
+
+            await Assert.ThrowsAsync<ResourceConflictException>(async () =>
+                await Mediator.UpsertResourceAsync(newResourceValues.ToResourceElement(), WeakETag.FromVersionId("invalidVersion")));
+        }
+
+        [Fact]
+        [FhirStorageTestsFixtureArgumentSets(DataStore.SqlServer)]
```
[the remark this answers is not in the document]
rotodd: As mentioned below, sql/cosmos behaviours match as of 7649951.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -68,7 +68,7 @@ public async Task<UpsertOutcome> UpsertAsync(ResourceWrapper resource, WeakETag
             int etag = 0;
             if (weakETag != null && !int.TryParse(weakETag.VersionId, out etag))
             {
-                throw new ResourceConflictException(weakETag);
+                etag = -1;
```
johnstairs: This would be a good place to add a comment explaining why :)

johnstairs:
:shipit: !

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
rotodd: Thought you might say that... added in f9a779a!

PR #644, opened 2019-09-09
Refactored the audit event type mapping.

jackliums:
## Description
Refactored the audit event type mapping logic so that it can be used by different components. 

## Testing
Unit tests and manual test.

### DIFF
--- a/src/Microsoft.Health.Fhir.Api/Features/Audit/AuditEventTypeMapping.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/Audit/AuditEventTypeMapping.cs
@@ -18,6 +18,19 @@ namespace Microsoft.Health.Fhir.Api.Features.Audit
     /// <summary>
     /// Provides the ability to lookup audit event type.
     /// </summary>
+    /// <remarks>
+    /// <para>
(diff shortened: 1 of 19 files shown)


src/Microsoft.Health.Fhir.Api/Features/Audit/AuditEventTypeMapping.cs
```diff
@@ -0,0 +1,67 @@
...
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc.Controllers;
+using Microsoft.AspNetCore.Mvc.Infrastructure;
+using Microsoft.Health.Extensions.DependencyInjection;
+using Microsoft.Health.Fhir.Api.Extensions;
+
+namespace Microsoft.Health.Fhir.Api.Features.Audit
+{
+    /// <summary>
+    /// Provides the ability to lookup audit event type.
```
johnstairs: Might want to provide a little more context here and why this solution is necessary

jackliums: That's a good idea. I will add the comments.

PR #646, opened 2019-09-10
Add batch/transaction route

brandonpollett:
## Description
Adds the batch/transaction route to the server. The route is current non-functional and should return either a `405` in the case of a bundle posted or a `400` if anything else is posted.

## Related issues
Addresses [AB#70188](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/70188)

## Testing
Add E2E tests for batch and transaction to check that expected status codes are returned.

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Api/Controllers/FhirController.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Api/Controllers/FhirController.cs
@@ -378,6 +378,7 @@ namespace Microsoft.Health.Fhir.Api.Controllers
         /// <summary>
         /// Searches across all resource types.
         /// </summary>
+        [HttpGet]
         [Route("", Name = RouteNames.SearchAllResources)]
         [AuditEventType(AuditEventSubType.SearchSystem)]
         [Authorize(PolicyNames.ReadPolicy)]
@@ -511,6 +512,21 @@ namespace Microsoft.Health.Fhir.Api.Controllers
             VersionsResult response = await _mediator.GetOperationVersionsAsync(HttpContext.RequestAborted);
 
             return new OperationVersionsResult(response, HttpStatusCode.OK);
+        }
+
+        /// <summary>
+        /// Handles batch and transaction requests
+        /// </summary>
+        /// <param name="bundle">The bundle being posted</param>
+        [HttpPost]
+        [Route("", Name = RouteNames.PostBundle)]
+        [AuditEventType(AuditEventSubType.Batch)] // TODO: Our current auditing implementation only allows one audit event type attribute even though this action handles two.
+        [Authorize(PolicyNames.WritePolicy)]
+        public async Task<IActionResult> BatchAndTransactions([FromBody] Resource bundle)
+        {
+            ResourceElement bundleResponse = await _mediator.PostBundle(bundle.ToResourceElement());
+
+            return FhirResult.Create(bundleResponse);
         }
     }
 }
--- a/src/Microsoft.Health.Fhir.Shared.Api/Controllers/FhirController.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Api/Controllers/FhirController.cs
@@ -18,6 +18,7 @@ using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Microsoft.Extensions.Primitives;
 using Microsoft.Health.Fhir.Api.Configs;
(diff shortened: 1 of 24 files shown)


src/Microsoft.Health.Fhir.Shared.Api/Controllers/FhirController.cs
```diff
@@ -512,5 +513,20 @@ public async Task<IActionResult> Versions()
...
             return new OperationVersionsResult(response, HttpStatusCode.OK);
         }
+
+        /// <summary>
+        /// Handles batch and transaction requests
+        /// </summary>
+        /// <param name="bundle">The bundle being posted</param>
+        [HttpPost]
+        [Route("", Name = RouteNames.PostBundle)]
+        [AuditEventType(AuditEventSubType.Batch)] // TODO: Our current auditing implementation only allows one audit event type attribute even though this action handles two.
```
jackliums: For transaction/batch, we will have to figure out the type based on the "type" property?

src/Microsoft.Health.Fhir.Shared.Api/Controllers/FhirController.cs
```diff
@@ -512,5 +513,20 @@ public async Task<IActionResult> Versions()
...
+        [Authorize(PolicyNames.WritePolicy)]
```
jackliums: Batch can also do read right?

Should we have batch as a separate policy?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/TransactionTests.cs
```diff
@@ -0,0 +1,45 @@
...
+        [Fact]
+        [Trait(Traits.Priority, Priority.One)]
+        public async Task WhenSubmittingATransaction_GivenAProperBundle_ThenNotSupportedIsReturned()
+        {
+            FhirException ex = await Assert.ThrowsAsync<FhirException>(() => Client.PostBundleAsync(Samples.GetDefaultTransaction().ToPoco<Bundle>()));
+
+            Assert.Equal(HttpStatusCode.MethodNotAllowed, ex.StatusCode);
+        }
+
+        [Fact]
```
jackliums: This is redundant since the same test already exists in the batch one.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/FhirController.cs
jackliums: I wonder if we can use routing constraints or something to separate these two out?

brandonpollett: Correct, the audit event type will be determined by the `type` of the bundle resource posted. 

I don't believe route constraints will work (at least out of the box) because [`IRouteConstraint` ](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.routing.irouteconstraint?view=aspnetcore-2.2) only contains the route data - not the payload.

There are at least a couple of options, but I think the decision merits more thought as a separate story:
- Move the auditing down to the mediator level
- Make the auditing attribute aware of payload
- Make the auditing a custom middleware

src/Microsoft.Health.Fhir.Shared.Api/Controllers/FhirController.cs
brandonpollett: My original thought was that `write` implicitly allowed `read`, but this doesn't seem to be the way it's currently implemented. We have this exact same issue today with `ConditionalCreate` and `ConditionalUpdate`. Let me give this some more thought.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/FhirController.cs
brandonpollett: We may be able to use an [`IActionConstraint`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.actionconstraints.iactionconstraint?view=aspnetcore-2.2) to achieve the route selection.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/FhirController.cs
brandonpollett: I ended up just adding `write` and `read` policies.

Issue #22, opened 2018-09-18
'ap' prefix on a search for an integer quantity does not behave as expected


johnstairs:
Given a service where the following returns a single item:
http://localhost:53727/Observation?component-value-quantity=eq107
Then this should also find the item:
http://localhost:53727/Observation?component-value-quantity=ap105
But it does not.

From http://hl7.org/fhir/stu3/search.html#prefix:

> the value for the parameter in the resource is approximately the same to the provided value.
Note that the recommended value for the approximation is 10% of the stated value (or for a date, 10% of the gap between now and the date), but systems may choose other values where appropriate

PR #654, opened 2019-09-16
Add approximate value to number/quantity.

brandonpollett:
## Description
Adds approximate value for number and quantity. If you pass in the value `15.0` it will search for items with values between `13.45` and `16.55`

## Related issues
Addresses #22 .

## Testing
Update to the unit test that covered the `ap` modifiers and manual test of E2E scenario talked about in the issue.


src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Expressions/Parsers/SearchValueExpressionBuilderHelper.cs
```diff
@@ -338,8 +338,14 @@ private Expression GenerateNumberExpression(FieldName fieldName, decimal number)
 
             switch (_comparator)
             {
-                case SearchComparator.Eq:
                 case SearchComparator.Ap:
+                    var approximateModifier = Math.Abs(number * Convert.ToDecimal(ApproximateMultiplier));
+                    lowerBound -= approximateModifier;
+                    upperBound += approximateModifier;
+                    return Expression.And(
```
brandonpollett: Could probably use a `goto case SearchComparator.Eq;` here after modifying the lower and upper bound.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Expressions/Parsers/SearchValueExpressionBuilderHelper.cs
```diff
@@ -338,8 +338,12 @@ private Expression GenerateNumberExpression(FieldName fieldName, decimal number)
 
             switch (_comparator)
             {
-                case SearchComparator.Eq:
                 case SearchComparator.Ap:
+                    var approximateModifier = Math.Abs(number * Convert.ToDecimal(ApproximateMultiplier));
```
johnstairs: maybe make the constant a decimal instead?

Issue #536, opened 2019-06-09
Validation messages are missing details


johnstairs:
**Describe the bug**
The server provides insufficient information when rejecting a POST because of validation errors

**To Reproduce**
```
curl -X POST \
  https://localhost:44348/DiagnosticReport \
  -H 'Accept: */*' \
  -H 'Cache-Control: no-cache' \
  -H 'Connection: keep-alive' \
  -H 'Content-Type: application/json' \
  -H 'Host: localhost:44348' \
  -H 'User-Agent: PostmanRuntime/7.13.0' \
  -H 'accept-encoding: gzip, deflate' \
  -H 'cache-control: no-cache' \
  -H 'content-length: 39' \
  -d '{
	"resourceType": "DiagnosticReport"
}'
```
**Expected behavior**
Information about which elements are missing

**Actual behavior**
``` json
{
    "resourceType": "OperationOutcome",
    "id": "bd982592-509b-42c2-be8f-90fc304869bb",
    "issue": [
        {
            "severity": "error",
            "code": "invalid",
            "diagnostics": "Element with min. cardinality 1 cannot be null"
        },
        {
            "severity": "error",
            "code": "invalid",
            "diagnostics": "Element with min. cardinality 1 cannot be null"
        }
    ]
}
```

johnstairs:
cc @brendankowitz

ElYusubov:
I am seeing same generic error on FHIR API as well.
@johnstairs & @brendankowitz is there an ETA when error details will be added?

![image](https://user-images.githubusercontent.com/29330538/64457694-4f58dc80-d0c1-11e9-98f2-5eeffb6feccd.png)

brendankowitz:
I'm pretty sure the information is there, it's just not being passed through correctly. I'll look at getting this scheduled.

PR #655, opened 2019-09-18
Adds location to validator error messages

rotodd:
## Description

- The location is now included in the attribute validator error messages

validation_messages_with_location

## Related issues

- [#AB70211](https://microsofthealth.visualstudio.com/Health/_workitems/edit/70211) #536

## Testing

- Updated existing unit tests to ensure a FHIR path is included in validation messages
- Added a new unit test for attribute validation

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Validation/AttributeValidator.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Validation/AttributeValidator.cs
@@ -43,7 +43,9 @@ namespace Microsoft.Health.Fhir.Core.Features.Validation
                     foreach (var error in results)
                     {
                         ITypedElement typedElement = resourceElement.Instance;
-                        var fullFhirPath = typedElement.InstanceType + "." + error.MemberNames?.FirstOrDefault();
+
+                        var fullFhirPath = typedElement.InstanceType;
+                        fullFhirPath += error.MemberNames?.FirstOrDefault() == null ? $"" : "." + error.MemberNames?.FirstOrDefault();
 
                         yield return new ValidationFailure(fullFhirPath, error.ErrorMessage);
                     }
--- a/src/Microsoft.Health.Fhir.Core/Features/Validation/AttributeValidator.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Validation/AttributeValidator.cs
@@ -10,6 +10,7 @@ using EnsureThat;
 using FluentValidation;
 using FluentValidation.Results;
 using FluentValidation.Validators;
+using Hl7.Fhir.ElementModel;
 using Microsoft.Health.Fhir.Core.Models;
 using Task = System.Threading.Tasks.Task;
 using ValidationResult = System.ComponentModel.DataAnnotations.ValidationResult;
@@ -41,7 +42,10 @@ namespace Microsoft.Health.Fhir.Core.Features.Validation
                 {
                     foreach (var error in results)
                     {
-                        yield return new ValidationFailure(error.MemberNames?.FirstOrDefault(), error.ErrorMessage);
+                        ITypedElement typedElement = resourceElement.Instance;
+                        var fullFhirPath = typedElement.InstanceType + "." + error.MemberNames?.FirstOrDefault();
+
+                        yield return new ValidationFailure(fullFhirPath, error.ErrorMessage);
                     }
                 }
             }
--- a/src/Microsoft.Health.Fhir.Core/Features/Validation/AttributeValidator.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Validation/AttributeValidator.cs
@@ -45,7 +45,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Validation
                         ITypedElement typedElement = resourceElement.Instance;
 
                         var fullFhirPath = typedElement.InstanceType;
-                        fullFhirPath += error.MemberNames?.FirstOrDefault() == null ? $"" : "." + error.MemberNames?.FirstOrDefault();
+                        fullFhirPath += string.IsNullOrEmpty(error.MemberNames?.FirstOrDefault()) ? string.Empty : "." + error.MemberNames?.FirstOrDefault();
 
                         yield return new ValidationFailure(fullFhirPath, error.ErrorMessage);
                     }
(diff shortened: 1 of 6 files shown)


src/Microsoft.Health.Fhir.Core/Features/Validation/AttributeValidator.cs
```diff
@@ -41,7 +42,10 @@ public IEnumerable<ValidationFailure> Validate(PropertyValidatorContext context)
                 {
                     foreach (var error in results)
                     {
-                        yield return new ValidationFailure(error.MemberNames?.FirstOrDefault(), error.ErrorMessage);
+                        ITypedElement typedElement = resourceElement.Instance;
+                        var fullFhirPath = typedElement.InstanceType + "." + error.MemberNames?.FirstOrDefault();
```
johnstairs: It's not entirely obvious that this works for paths that are deeper than top-level fields, but from your screenshot it looks like it does. :) Could you add a unit test to verify the path?

src/Microsoft.Health.Fhir.Core/Features/Validation/ResourceNotValidException.cs
```diff
@@ -41,7 +41,8 @@ public ResourceNotValidException(IEnumerable<ValidationFailure> validationFailur
                     Issues.Add(new OperationOutcomeIssue(
                             OperationOutcomeConstants.IssueSeverity.Error,
                             OperationOutcomeConstants.IssueType.Invalid,
-                            failure.ErrorMessage));
+                            failure.ErrorMessage,
+                            new[] { failure.PropertyName }));
```
brendankowitz: Lets only create the array if PropertyName is NotNullOrEmpty

src/Microsoft.Health.Fhir.Core/Features/Validation/AttributeValidator.cs
brendankowitz: The code `error.MemberNames?.FirstOrDefault()` reads that it could be a null value. In this case perhaps we just use the InstanceType?

rotodd: Yep definitely!

brendankowitz: This might not be a problem right now since we currently have recursion set to false: 
https://github.com/microsoft/fhir-server/blob/master/src/Microsoft.Health.Fhir.Core/Features/Validation/AttributeValidator.cs#L40

src/Microsoft.Health.Fhir.Core/Features/Validation/ResourceNotValidException.cs
rotodd: Good idea, fixed in 2b34f6d.

src/Microsoft.Health.Fhir.Core/Features/Validation/AttributeValidator.cs
rotodd: Makes sense, fixed in 2b34f6d.

src/Microsoft.Health.Fhir.Core/Features/Validation/AttributeValidator.cs
```diff
@@ -41,7 +42,12 @@ public IEnumerable<ValidationFailure> Validate(PropertyValidatorContext context)
                 {
                     foreach (var error in results)
                     {
-                        yield return new ValidationFailure(error.MemberNames?.FirstOrDefault(), error.ErrorMessage);
+                        ITypedElement typedElement = resourceElement.Instance;
+
+                        var fullFhirPath = typedElement.InstanceType;
+                        fullFhirPath += error.MemberNames?.FirstOrDefault() == null ? $"" : "." + error.MemberNames?.FirstOrDefault();
```
brendankowitz: Small tweak suggestion:
```suggestion
                        fullFhirPath += string.IsNullOrEmpty(error.MemberNames?.FirstOrDefault()) ? string.Empty : "." + error.MemberNames?.FirstOrDefault();
```

src/Microsoft.Health.Fhir.Core/Features/Validation/AttributeValidator.cs
rotodd: Let me know what you think of cf799da and if you think it makes sense to add new unit tests for AttributeValidator.

src/Microsoft.Health.Fhir.Core/Features/Validation/AttributeValidator.cs
rotodd: Ah, right. Thanks. Fixed in dcf6b94.

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Validation/Narratives/NarrativeValidatorTests.cs
```diff
@@ -27,29 +31,39 @@ public NarrativeValidatorTests()
...
         }
 
         [Theory]
         [MemberData(nameof(XssStrings))]
-        public void GivenAnInvalidNarrative_WhenProcessingABundle_ThenAValidationMessageIsCreated(string maliciousNarrative)
+        public void GivenAnInvalidNarrative_WhenProcessingABundle_ThenAValidationMessageWithAFhirPathIsCreated(string maliciousNarrative)
         {
             var defaultObservation = Samples.GetDefaultObservation().ToPoco<Observation>();
             defaultObservation.Text.Div = maliciousNarrative;
+            var observationInstance = defaultObservation.ToResourceElement().Instance;
```
brendankowitz: Perhaps just use `defaultObservation.TypeName` below instead of converting.
There is also a `[ResourceElement].InstanceType` that exists without needing to navigate to the Instance

rotodd: Oh cool, thanks. Fixed in 9c8d694.

src/Microsoft.Health.Fhir.Core/Features/Validation/AttributeValidator.cs
rotodd: Added an AttributeValidator unit test in cc717d1.

Issue #657, opened 2019-09-19
Some routes return correctly but throw Xml InvalidOperationException in console


brendankowitz:
**Describe the bug**
Navigating to some routes and specifying xml formatting will return correctly, but the following error is logged to the console

```
Microsoft.AspNetCore.Mvc.Formatters.XmlSerializerOutputFormatter:Warning: An error occurred while trying to create an XmlSerializer for the type 'Hl7.Fhir.Model.Resource'.

System.InvalidOperationException: There was an error reflecting type 'Hl7.Fhir.Model.Resource'. ---> System.InvalidOperationException: There was an error reflecting property 'Meta'. ---> System.InvalidOperationException: There was an error reflecting type 'Hl7.Fhir.Model.Meta'. ---> System.InvalidOperationException: Cannot serialize member 'Hl7.Fhir.Model.Meta.Profile' of type 'System.Collections.Generic.IEnumerable`1[[System.String, System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]]', see inner exception for more details. ---> System.NotSupportedException: Cannot serialize member Hl7.Fhir.Model.Meta.Profile of type System.Collections.Generic.IEnumerable`1[[System.String, System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]] because it is an interface.
   --- End of inner exception stack trace ---
   at System.Xml.Serialization.StructModel.CheckSupportedMember(TypeDesc typeDesc, MemberInfo member, Type type)
   at System.Xml.Serialization.StructModel.GetPropertyModel(PropertyInfo propertyInfo)
   at System.Xml.Serialization.StructModel.GetFieldModel(MemberInfo memberInfo)
   at System.Xml.Serialization.XmlReflectionImporter.InitializeStructMembers(StructMapping mapping, StructModel model, Boolean openModel, String typeName, RecursionLimiter limiter)
   at System.Xml.Serialization.XmlReflectionImporter.ImportStructLikeMapping(StructModel model, String ns, Boolean openModel, XmlAttributes a, RecursionLimiter limiter)
   at System.Xml.Serialization.XmlReflectionImporter.ImportTypeMapping(TypeModel model, String ns, ImportContext context, String dataType, XmlAttributes a, Boolean repeats, Boolean openModel, RecursionLimiter limiter)
   --- End of inner exception stack trace ---
[...]

PR #658, opened 2019-09-19
Adds a GenericXmlOutputSerializer that only activates for non-fhir types

brendankowitz:
## Description
Adds a GenericXmlSerializer that only serializes non-FHIR types

## Related issues
Addresses #657

## Testing
Manually inspected console output


src/Microsoft.Health.Fhir.Shared.Api/Features/Formatters/GenericXmlOutputFormatter.cs
```diff
@@ -0,0 +1,27 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using EnsureThat;
+using Hl7.Fhir.Model;
+using Microsoft.AspNetCore.Mvc.Formatters;
+
+namespace Microsoft.Health.Fhir.Api.Features.Formatters
+{
+    public class GenericXmlOutputFormatter : XmlSerializerOutputFormatter
```
deepak-ms: Should we rename it to 'NonFhirResourceXmlOutputFormatter'?

brendankowitz: Thats the original name I had :P

Issue #649, opened 2019-09-11
Incorrect response status code for invalid Accept header on GET requests


brendankowitz:
**Describe the bug**

406 is the appropriate response when the Accept header requests a format that the server does not support, and 415 Unsupported Media Type when the client posts a format that is not supported to the server.

Ref: http://hl7.org/fhir/formats.html#wire

**To Reproduce**
1. GET /Patient?_format=application/foobar

**Expected behavior**
When not a post it should be 406

**Actual behavior**
A response status code 415 returned

PR #661, opened 2019-09-20
Returns 406 error for invalid Accept header on GET request

rotodd:
## Description

- The server now returns a 406 Not Acceptable status code if the Accept header requests a format that the server does not support (see [spec](http://hl7.org/fhir/formats.html#wire))
- Previously, a 415 Unsupported Media Type status code was returned

406_error_on_get_with_invalid_header

## Related issues

- [#AB70307](https://microsofthealth.visualstudio.com/Health/_workitems/edit/70307)
- #649 

## Testing

- Updated existing unit test


brendankowitz:
Can you also check that a get request with `accept: application/invalid` return 406.
When posting, content-type: application/invalid should be 415.

rotodd:
Need to look into the GET/POST differentiation a bit more as well as the differences between the _format parameter and Accept header.

Issue #669, opened 2019-09-26
ApiNotificationMiddleware could fail if FhirRequestContext is not set


YazanMSFT:
**Describe the bug**
There are cases where it's possible that the FhirRequestContext is null. The ApiNotificationMiddleware should handle those gracefully.

**To Reproduce**
This generally happens for request types that do not have an authz header.

**Expected behavior**
API response is preserved, no ApiNotification is published.

**Actual behavior**
An internal server error is thrown.

PR #670, opened 2019-09-26
Failures in emitting API notifications should not cause the response to fail

YazanMSFT:
## Description
This PR addresses two issues:

1) There are cases when the FhirRequestContext is not set. We should check for null in those cases and not attempt to publish an API notification.
2) If any errors occur while publishing the notification, they should not cause the API to fail.

## Related issues
Addresses [issue #669].

## Testing
Unit tests were added that replicate the failure case. Verified they failed before the fix and succeed after.

### DIFF
--- a/src/Microsoft.Health.Fhir.Api/Features/ApiNotifications/ApiNotificationMiddleware.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/ApiNotifications/ApiNotificationMiddleware.cs
@@ -87,7 +87,7 @@ namespace Microsoft.Health.Fhir.Api.Features.ApiNotifications
                     catch (Exception e)
                     {
                         // Failures in publishing API notifications should not cause the API to return an error.
-                        _logger.LogError(e, "Failure while publishing API notification.");
+                        _logger.LogCritical(e, "Failure while publishing API notification.");
                     }
                 }
             }
--- a/src/Microsoft.Health.Fhir.Api/Features/ApiNotifications/ApiNotificationMiddleware.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/ApiNotifications/ApiNotificationMiddleware.cs
@@ -3,6 +3,7 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
+using System;
 using System.Net;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Fhir.Api/Features/ApiNotifications/ApiNotificationMiddleware.cs
```diff
@@ -62,24 +63,32 @@ public async Task Invoke(HttpContext context)
...
+
+                            await _mediator.Publish(apiNotification, CancellationToken.None);
+
+                            if (fhirRequestContext.StorageRequestMetrics != null)
+                            {
+                                await _mediator.Publish(fhirRequestContext.StorageRequestMetrics, CancellationToken.None);
+                            }
                         }
                     }
+                    catch (Exception e)
```
brandonpollett: Why not just check for a null FhirRequestContext and return instead of catching all the exceptions? This seems overly broad and may hide other issues in the future.

brendankowitz: Need to confirm, but if a NotificationHandler throws, it might bubble up here. The this middleware shouldn't fail here because of this

brendankowitz:
> Any idea why the CORs E2E test didn't catch this?

This middleware isn't registered in the OSS server. I proposed either registering it (but there are no consumers in OSS) or moving it out

src/Microsoft.Health.Fhir.Api/Features/ApiNotifications/ApiNotificationMiddleware.cs
YazanMSFT: Yeah, looks like an exception in a handler will bubble up to this middleware.

src/Microsoft.Health.Fhir.Api/Features/ApiNotifications/ApiNotificationMiddleware.cs
```diff
@@ -62,24 +63,32 @@ public async Task Invoke(HttpContext context)
...
+                    {
+                        // Failures in publishing API notifications should not cause the API to return an error.
+                        _logger.LogError(e, "Failure while publishing API notification.");
```
brandonpollett: I think we should change this to a LogCritical since we're catching everything to make this case stand out more.

PR #676, opened 2019-10-03
Added FHIR service name length validation check

anrudraw:
## Description
This PR has modified default-azuredeploy-sql.json and default-azuredeploy.json  ARM Templates for validating the length of fhir service name upfront.

## Related issues
Addresses [issue #].

## Testing


hansenms:
This only changes it in the SQL version of the template. I believe it should be consistently in both.

anrudraw:
@hansenms updated Cosmos DB template as well.

PR #690, opened 2019-10-15
Add PATCH route and E2E test

brandonpollett:
## Description
Adds a `PATCH` route that returns a 405 status code and an operation outcome that says patch is not supported. 

## Related issues
Addresses [AB#70318](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/70318).

## Testing
An E2E test was added to make sure a patch request returns the appropriate status code.

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Api/Controllers/FhirController.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Api/Controllers/FhirController.cs
@@ -25,6 +25,7 @@ using Microsoft.Health.Fhir.Api.Features.Filters;
 using Microsoft.Health.Fhir.Api.Features.Headers;
 using Microsoft.Health.Fhir.Api.Features.Routing;
 using Microsoft.Health.Fhir.Api.Features.Security;
+using Microsoft.Health.Fhir.Core.Exceptions;
 using Microsoft.Health.Fhir.Core.Extensions;
 using Microsoft.Health.Fhir.Core.Features;
 using Microsoft.Health.Fhir.Core.Features.Context;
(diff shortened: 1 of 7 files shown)


src/Microsoft.Health.Fhir.Shared.Api/Controllers/FhirController.cs
```diff
@@ -389,6 +390,22 @@ public async Task<IActionResult> Delete(string typeParameter, string idParameter
...
+        /// <param name="idParameter">The identifier.</param>
+        [HttpPatch]
+        [Route(KnownRoutes.ResourceTypeById)]
+        [AuditEventType(AuditEventSubType.Patch)]
+        [Authorize(PolicyNames.WritePolicy)]
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Performance", "CA1822:Mark members as static", Justification = "Controller methods won't be called if static.")]
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Usage", "CA1801:Review unused parameters", Justification = "Need the parameters for routing to work.")]
+        public Task<IActionResult> Patch(string typeParameter, string idParameter)
+        {
+            throw new MethodNotAllowedException(Resources.PatchNotSupported);
```
namadabu: nit: maybe we can re-use `Resources.UnsupportedOperation` here instead of creating a new one?

brandonpollett: Patch is an http method, not an [FHIR operation](https://www.hl7.org/fhir/operations.html).

PR #693, opened 2019-10-18
Implements _total parameter (accurate/none)

rotodd:
## Description

- Enables clients to request that the total number of matching resources be included in the returned bundle as described in the [spec](https://www.hl7.org/fhir/search.html#total)
- The implementation varies depending on the configured database:
    - SQL: two read queries are performed in one transaction, allowing for read-committed isolation
    - Cosmos DB: two read queries are performed (this operation is not atomic)
- Note that returning an estimate of the total number of matching resources is still to-be-implemented (see [AB#70590](https://microsofthealth.visualstudio.com/Health/_workitems/edit/70590))

total-param

## Related issues

Addresses [AB#70590](https://microsofthealth.visualstudio.com/Health/_workitems/edit/70110).

## Testing

- Adds two new E2E tests


src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -67,17 +67,51 @@ internal class SqlServerSearchService : SearchService
...
+            {
+                connection.Open();
+
+                SearchResult searchResult;
+
+                // If we should include the total count of matching search results
+                if (searchOptions.IncludeTotal == TotalType.Accurate && !searchOptions.CountOnly)
+                {
+                    // Begin a transaction so we can perform two atomic reads.
+                    using (var transaction = connection.BeginTransaction(IsolationLevel.ReadCommitted))
```
deepak-ms: Shouldn't a higher level of isolation be used here for count to be same as number of results?

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
```diff
@@ -90,6 +90,17 @@ public SearchOptions Create(string compartmentType, string compartmentId, string
                     // Query parameter with empty value is not supported.
                     unsupportedSearchParameters.Add(query);
                 }
+                else if (query.Item1 == KnownQueryParameterNames.Total)
+                {
+                    if (string.Compare(query.Item2, TotalType.Accurate.ToString(), StringComparison.OrdinalIgnoreCase) == 0)
```
brandonpollett: Could we convert the `TotalType` to a normal enum and do something like `query.Item2.GetValueByEnumLiteral<TotalType>()` to get the enum value?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
johnstairs: That is a good point. The database is configured with read committed snapshot isolation, so each read by default will be based on a consistent point in time. It was my impression that this was based on the start of the transaction, but it looks like it is from the start of each statement, even when running in a transaction (please confirm). So we will want, and will will need to enable, snapshot isolation. This will give the desired behavior without blocking writes. 
Apologies, @rotodd, it looks like I didn't get this completely right.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
```diff
@@ -90,6 +90,17 @@ public SearchOptions Create(string compartmentType, string compartmentId, string
                     // Query parameter with empty value is not supported.
                     unsupportedSearchParameters.Add(query);
                 }
+                else if (query.Item1 == KnownQueryParameterNames.Total)
```
brendankowitz: This should probably also be StringComparison.OrdinalIgnoreCase

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
```diff
@@ -40,10 +40,23 @@ public class FhirCosmosSearchService : SearchService
...
         {
-            return await ExecuteSearchAsync(
+            SearchResult searchResult = await ExecuteSearchAsync(
                 _queryBuilder.BuildSqlQuerySpec(searchOptions),
                 searchOptions,
                 cancellationToken);
+
+            if (searchOptions.IncludeTotal == TotalType.Accurate && !searchOptions.CountOnly)
+            {
+                searchOptions.CountOnly = true;
```
brendankowitz: I think I'd prefer this manipulation to be done in a method on SearchOptions for now. i.e. `searchOptions.AsCountOnly()`. The property can then remain as private.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
rotodd: @johnstairs I did some reading to confirm, and you're right - specifying `IsolationLevel.ReadCommitted` with the setting `READ_COMMITTED_SNAPSHOT ON` would definitely work. I noticed that specifying `IsolationLevel.Snapshot` with the setting `ALLOW_SNAPSHOT_ISOLATION ON` would also work.

I decided to go with the second option because the first one will affect all transactions where the isolation level is either unset or set to ReadCommitted. The second option seemed to be more explicit with fewer potential side effects (more info [here](https://docs.microsoft.com/en-us/dotnet/framework/data/adonet/sql/snapshot-isolation-in-sql-server?redirectedfrom=MSDN)).

Since John is on vacation, @brandonpollett or @deepak-ms may you take a look at 3c160b1?

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
rotodd: Good call on getting rid of the weird TotalType class.

`GetValueByEnumLiteral` didn't agree with case sensitivity, but `Enum.TryParse` seems to do the job. Let me know what you think of e6364ac.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
rotodd: Thanks, fixed in 9df2fc7.

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
rotodd: Feels a bit strange adding a set method to a class of properties with internal setters, but I understand that manipulating CountOnly directly in these files isn't super nice. Let me know how 809259a looks.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
deepak-ms: @rotodd, what i understand from sql docs is that first option won't work for this scenario since (as John also pointed) row snapshot is from start of the statement (not transaction). So only second option will work here.

deepak-ms: FWIW, 
i got confused by "If you set the READ_COMMITTED_SNAPSHOT database option to ON, the database engine uses row versioning and **snapshot** isolation as the default, instead of using locks to protect the data." in the docs.
So I just tested it out. 
Result : If both ALLOW_SNAPSHOT_ISOLATION and READ_COMMITTED_SNAPSHOT are ON,
then ReadCommitted still behaves as ReadCommitted (reads based on start of **statement**).

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
```diff
@@ -90,6 +90,20 @@ public SearchOptions Create(string compartmentType, string compartmentId, string
...
+                    if (Enum.TryParse<TotalType>(query.Item2, true, out var totalType))
+                    {
+                        if (totalType == TotalType.Accurate)
+                        {
+                            searchOptions.IncludeTotal = TotalType.Accurate;
+                        }
+                        else if (totalType == TotalType.None)
+                        {
+                            searchOptions.IncludeTotal = TotalType.None;
+                        }
```
brandonpollett: Is there any reason we can't just do `searchOptions.IncludTotal = totalType;` here?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
rotodd: Ah interesting. Thank you! I was mislead by that as well. I'm interested to hear about how you tested it?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
```diff
@@ -43,6 +43,9 @@ IF ((SELECT is_ansi_nulls_on FROM sys.databases WHERE database_id = DB_ID()) = 0
     ALTER DATABASE CURRENT SET ANSI_NULLS ON
 END
 
+-- Allow access to versioned rows if Snapshot isolation level is set for a specific transaction
+ALTER DATABASE CURRENT SET ALLOW_SNAPSHOT_ISOLATION ON
+
```
rotodd: As discussed with @brandonpollett, this change may not be reflected (1.sql won't be rerun for existing processes), causing any SQL transactions that specify `IsolationLevel.Snapshot` to throw an exception.

Since it would be better to risk having mismatched count and search results than risk an exception being thrown, I'll revert to using `IsolationLevel.ReadCommitted`. I also created a user story to update the isolation level once schema migration is implemented (see [#AB70783](https://microsofthealth.visualstudio.com/Health/_workitems/edit/70783)).

FYI @deepak-ms @johnstairs

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
rotodd: Nope, good idea. Fixed in 85c7130.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
rotodd: Implemented in 8e9b00e.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
deepak-ms: @rotodd, used 2 interleaving transactions for testing. [This](https://www.interfacett.com/blogs/understanding-isolation-levels-sql-server-2008-r2-2012-examples/) doc might be useful.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
```diff
@@ -90,6 +90,13 @@ public SearchOptions Create(string compartmentType, string compartmentId, string
                     // Query parameter with empty value is not supported.
                     unsupportedSearchParameters.Add(query);
                 }
+                else if (string.Compare(query.Item1, KnownQueryParameterNames.Total, StringComparison.OrdinalIgnoreCase) == 0)
+                {
+                    if (Enum.TryParse<TotalType>(query.Item2, true, out var totalType))
```
jackliums: If it fails, should it throw BadRequest? Also, if estimate is not supported, should we throw error indicating it's not supported?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -115,104 +149,107 @@ private async Task<SearchResult> SearchImpl(SearchOptions searchOptions, bool hi
...
             {
-                connection.Open();
-
-                using (SqlCommand sqlCommand = connection.CreateCommand())
+                // If we are sending multiple search queries in one transaction
+                if (transaction != null)
                 {
-                    var stringBuilder = new IndentedStringBuilder(new StringBuilder());
+                    // We must assign both the transaction and connection objects to the Command object for a pending local transaction.
+                    sqlCommand.Connection = connection;
```
brandonpollett: Is this not handled already by the `connection.CreateCommand()` above?

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
```diff
@@ -40,10 +40,23 @@ public class FhirCosmosSearchService : SearchService
...
         {
-            return await ExecuteSearchAsync(
+            SearchResult searchResult = await ExecuteSearchAsync(
                 _queryBuilder.BuildSqlQuerySpec(searchOptions),
                 searchOptions,
                 cancellationToken);
+
+            if (searchOptions.IncludeTotal == TotalType.Accurate && !searchOptions.CountOnly)
+            {
+                searchOptions.AsCountOnly(true);
```
jackliums: I think for Cosmos DB, this won't scale because it means for every request, it has to recompute the total count (even though that's might not be changing). Counting for Cosmos DB is also surprisingly slow, so that's another problem too.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
rotodd: Good question. Since it wasn't specified in the spec, I followed the implementation of `_pretty`  (which falls back to the default setting of `false` when an invalid/incorrect value is passed).

Looking at the implementations of `_summary` and `_format` now, it seems like your suggestion makes more sense. Do you think the lack of error messages for invalid `_pretty` values should be logged as a bug?

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
jackliums: Here is an example you can try:

https://ms.portal.azure.com/#@microsoft.onmicrosoft.com/resource/subscriptions/17af5f40-c564-4afe-ada0-fe7193bd474a/resourceGroups/jackliu-ip2/providers/Microsoft.DocumentDb/databaseAccounts/jackliu-ip2/queryExplorer

I have about 3 million resources in this Cosmos DB. If you run the query `SELECT VALUE COUNT(1) FROM c WHERE c.isSystem = false AND c.isHistory = false AND c.isDeleted = false`, which is pretty close to what we would generate, then you will see that it dispatches multiple requests (each against one partition) and each of that costs like 3600RU, which quickly uses up all available RUs of the database and that can cause DDoS very easily.

I think for this to scale

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
```diff
@@ -40,10 +40,23 @@ public class FhirCosmosSearchService : SearchService
...
+                var totalSearchResult = await ExecuteSearchAsync(
```
jackliums: It seems like what we want is the ability to tell the search implementation that we want numbers not the actual list of results. I think we can refactor the code a little bit so that it's not "forced" to fit the exsiting call patterns.

E.g., maybe keep SearchOptions strickly result of parsing the user input so it reflects what user has input and allow the search to perform in various different ways through other parameters or options. 

Does that make sense?

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
rotodd: As discussed, here are my action items moving forward:

- Reach out to the Cosmos DB team to see if there is a better approach (done)
- For now, throw a Not Supported exception in FhirCosmosSearchService.cs if the user attempts to specify `_total = accurate` (still to be done)

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
rotodd: As discussed, I plan to refactor to avoid modifying the SearchOptions object in this file.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
rotodd: Logged a [bug](https://microsofthealth.visualstudio.com/Health/_workitems/edit/70807) for the _pretty parameter and added invalid input handling for the _total parameter in 9830b25.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
rotodd: Right thanks, makes sense that setting it a second time isn't necessary (I was originally mislead by [this](https://docs.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqltransaction?view=netcore-2.0)). 

Fixed in 56ad72d

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
rotodd: Implemented in 9830b25.

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
rotodd: Implemented in 3357067 and d43999e.

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
rotodd: .. adding this back in after all in d43999e.

Reasoning: the Cosmos team is rolling out changes that should greatly reduce the computation cost of the queries that we are executing. Thanks @jackliums for looking into this.

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
```diff
@@ -59,7 +72,8 @@ public class FhirCosmosSearchService : SearchService
         private async Task<SearchResult> ExecuteSearchAsync(
```
jackliums: Instead of adding a boolean, I think we should just split the method; it's not sharing much logic anyways. Perhaps having `ExecutSearchAsync` and `ExecuteCountSearchAsync` or something like that?

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
```diff
@@ -68,7 +82,7 @@ public class FhirCosmosSearchService : SearchService
                 RequestContinuation = searchOptions.ContinuationToken,
             };
 
-            if (searchOptions.CountOnly)
+            if (searchOptions.CountOnly || calculateTotalCount)
             {
                 return new SearchResult(
                     (await _fhirDataStore.ExecuteDocumentQueryAsync<int>(sqlQuerySpec, feedOptions, cancellationToken)).Single(),
```
jackliums: The ContinuationToken is specified above. What does that mean when total is specified? For example, let's say the initial count search returned 1000 items and the search returned 100 items. When the user follows the continuation link to get the second page, what would the total be? Is it still 1000 or would it actually be 900?

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
```diff
@@ -40,10 +40,23 @@ public class FhirCosmosSearchService : SearchService
             SearchOptions searchOptions,
             CancellationToken cancellationToken)
         {
-            return await ExecuteSearchAsync(
+            SearchResult searchResult = await ExecuteSearchAsync(
                 _queryBuilder.BuildSqlQuerySpec(searchOptions),
                 searchOptions,
                 cancellationToken);
+
+            if (searchOptions.IncludeTotal == TotalType.Accurate && !searchOptions.CountOnly)
```
jackliums: Are we intentionally computing the total count every time search is executed? Alternatively, we could just compute the total during the first search and somehow carry that information through ct or snapshot or something so we don't need to compute it again and again right? (although we will need to figure out how to deal with the fact that resources could be manipulated while the user is paging through the search result).

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/Queries/QueryBuilder.cs
```diff
@@ -38,11 +38,11 @@ public QueryBuilderHelper()
                 _queryHelper = new QueryHelper(_queryBuilder, _queryParameterManager, SearchValueConstants.RootAliasName);
             }
 
-            public SqlQuerySpec BuildSqlQuerySpec(SearchOptions searchOptions)
+            public SqlQuerySpec BuildSqlQuerySpec(SearchOptions searchOptions, bool calculateTotalCount = false)
```
jackliums: I think we should change the signature to: `public SqlQuerySpec BuildSqlQuerySpec(Expression expression, bool countOnly)` or something like that. We should also update `GenerateHistorySql` below to match as well so `public SqlQuerySpec GenerateHistorySql(Expression expression)`. Actually, if you are touching it, might as well rename them so the names are consistent (e.g., GenerateSearchSql and GenerateHistorySql or something).

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
rotodd: Right - I've chatted with @johnstairs about this. I think this is an optimization that is definitely worth considering.

I created [#AB70827](https://microsofthealth.visualstudio.com/Health/_workitems/edit/70827).

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/Queries/QueryBuilder.cs
rotodd: As discussed, I created [#AB70828](https://microsofthealth.visualstudio.com/Health/_workitems/edit/70828).

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
rotodd: Same as below, I created [#AB70828](https://microsofthealth.visualstudio.com/Health/_workitems/edit/70828).

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
rotodd: I created #AB70827 to look into this.

PR #701, opened 2019-10-24
Make bundle sub requests have correct audit info.

brandonpollett:
## Description
Updates the bundle handler to set a new `FhirRequestContext` per sub request. This allows the audit logger to emit the correct information in the audit log including what the URL request would be if the call was made directly.

## Related issues
Addresses [AB#70317](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/70317)

## Testing
Manually tested.

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
@@ -16,12 +16,14 @@ using Hl7.Fhir.Model;
 using Hl7.Fhir.Serialization;
 using MediatR;
 using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Extensions;
 using Microsoft.AspNetCore.Http.Features.Authentication;
 using Microsoft.AspNetCore.Http.Headers;
 using Microsoft.AspNetCore.Routing;
 using Microsoft.Extensions.Primitives;
 using Microsoft.Health.Fhir.Api.Features.ContentTypes;
 using Microsoft.Health.Fhir.Api.Features.Headers;
+using Microsoft.Health.Fhir.Api.Features.Routing;
 using Microsoft.Health.Fhir.Core.Exceptions;
 using Microsoft.Health.Fhir.Core.Extensions;
 using Microsoft.Health.Fhir.Core.Features.Context;
@@ -97,6 +99,8 @@ namespace Microsoft.Health.Fhir.Api.Features.Resources.Bundle
                 httpContext.Response.Body = new MemoryStream();
 
                 var requestUri = new Uri(_fhirRequestContextAccessor.FhirRequestContext.BaseUri, entry.Request.Url);
+                httpContext.Request.Scheme = requestUri.Scheme;
+                httpContext.Request.Host = new HostString(requestUri.Host, requestUri.Port);
                 httpContext.Request.Path = requestUri.LocalPath;
                 httpContext.Request.QueryString = new QueryString(requestUri.Query);
                 httpContext.Request.Method = entry.Request.Method.ToString();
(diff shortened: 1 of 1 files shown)


src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -97,6 +99,8 @@ private async Task FillRequestLists(List<Hl7.Fhir.Model.Bundle.EntryComponent> b
                 httpContext.Response.Body = new MemoryStream();
 
                 var requestUri = new Uri(_fhirRequestContextAccessor.FhirRequestContext.BaseUri, entry.Request.Url);
+                httpContext.Request.Scheme = requestUri.Scheme;
```
jackliums: would the entry.Request.Url always be relative URL (and if not do we have validation?)

brandonpollett: According to the spec the `entry.Request.Url` is "The URL for this entry, relative to the root (the address to which the request is posted)." See [bundle spec](https://www.hl7.org/fhir/bundle-definitions.html#Bundle.entry.request.url) for details.

jackliums: Also, how does it handle correlationId? We are not passing by the HTTP header. Is it because this is within the same transient session and IoC contains returns the same instance of CorrelationIdProvider?

jackliums: It's using JsonContenType below is that something we can assume? If the incoming request was Xml, would this still work?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -146,6 +150,23 @@ private async Task ExecuteRequests(Hl7.Fhir.Model.Bundle responseBundle, Hl7.Fhi
...
+                        httpContext.Request.GetDisplayUrl(),
+                        originalFhirRequestContext.BaseUri.OriginalString,
+                        originalFhirRequestContext.CorrelationId,
+                        httpContext.Request.Headers,
+                        httpContext.Response.Headers,
+                        resourceType?.ToString())
+                    {
+                        Principal = originalFhirRequestContext.Principal,
+                    };
+                    _fhirRequestContextAccessor.FhirRequestContext = newFhirRequestContext;
```
jackliums: Do we not need to restore this context after it's done?

jackliums: I feel like there are a lot of conversions happening (hence more allocations and etc). And now reading through the code, mocking with the context and "simulating" the calls concerns me a bit also and i am not sure if we have enough test coverage in various areas we have to confidently say this works without regressions. Let's chat more on Monday.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
brandonpollett: > Also, how does it handle correlationId? 

We set the correlationId on the FhirRequestContext that we pass to the sub calls.

brandonpollett: > It's using JsonContenType below is that something we can assume? If the incoming request was Xml, would this still work?

Once it's within the context of the main bundle call we're only working with the pocos. We can choose from there to send all the subsequent requests as json. There's a decent amount of parsing/serializing going on here that we could optimize by not relying on the http request.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
brandonpollett: Because the accessor is backed by an asynclocal, the flow should only carry from outer async calls to inner async calls. Once you're out of the particular async call, it should revert to the value that it was.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -146,6 +150,23 @@ private async Task ExecuteRequests(Hl7.Fhir.Model.Bundle responseBundle, Hl7.Fhi
                 if (request.Handler != null)
                 {
                     HttpContext httpContext = request.HttpContext;
+
+                    IFhirRequestContext originalFhirRequestContext = _fhirRequestContextAccessor.FhirRequestContext;
+
+                    request.RouteData.Values.TryGetValue(KnownActionParameterNames.ResourceType, out object resourceType);
+                    var newFhirRequestContext = new FhirRequestContext(
```
jackliums: shoudl we add a test to make sure that correct audit log is produced for the batch call? It might only run locally (since the existing audit tests only works with in-memory server at the moment) but would still be nice to have coverage?

brandonpollett: Yes, we should. The plan is to do that as part of the stories AB#70787 and AB#70788 once we change the feature flag to be enabled by default.

PR #702, opened 2019-10-24
Move metrics notification for Cosmos to datastore provider.

brandonpollett:
## Description
This PR updates the way that metrics are emitted from the service to support discrete database metrics for individual requests within a batch/transaction. The Cosmos DB data provider has been updated to emit metrics for each of the calls that makes individually. 

## Related issues
Addresses [AB#70785](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/70785).

## Testing
- Updated existing tests
- Added additional E2E tests
- Added additional unit tests


src/Microsoft.Health.Fhir.CosmosDb/Features/Metrics/CosmosMetricProcessor.cs
```diff
@@ -0,0 +1,119 @@
...
+using MediatR;
+using Microsoft.Azure.Documents.Client;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Primitives;
+using Microsoft.Health.CosmosDb.Features.Storage;
+using Microsoft.Health.Fhir.Core.Features.Context;
+
+namespace Microsoft.Health.Fhir.CosmosDb.Features.Metrics
+{
+    public class CosmosMetricProcessor : ICosmosMetricProcessor
```
jackliums: Perhaps chagne this name to `CosmosResponseProcessor` or `CosmosDbResponseProcessor` (we should really standardize the naming but doesn't need to be in this one)? It seems like it's doing more than processsing metrics (e.g., it's setting the session token)

src/Microsoft.Health.Fhir.CosmosDb/Features/Metrics/CosmosMetricProcessor.cs
```diff
@@ -0,0 +1,119 @@
...
+
+            AddRequestChargeToFhirRequestContext(responseRequestCharge, collectionSizeUsageKilobytes, statusCode);
+        }
+
+        private void AddRequestChargeToFhirRequestContext(double responseRequestCharge, long? collectionSizeUsage, HttpStatusCode? statusCode)
+        {
+            IFhirRequestContext requestContext = _fhirRequestContextAccessor.FhirRequestContext;
+
+            // If there has already been a request to the database for this request, then we want to append a second charge header.
+            if (requestContext.ResponseHeaders.ContainsKey(CosmosDbHeaders.RequestCharge))
```
jackliums: nit: use TryGetValue here and can use Concat from the existing StringValues below.

src/Microsoft.Health.Fhir.CosmosDb/Features/Metrics/CosmosMetricProcessor.cs
```diff
@@ -0,0 +1,119 @@
...
+
+            _fhirRequestContextAccessor = fhirRequestContextAccessor;
+            _mediator = mediator;
+            _logger = logger;
+        }
+
+        public void ProcessResponse<T>(T resourceResponseBase)
+            where T : IResourceResponseBase
+        {
+            ProcessResponse(resourceResponseBase.SessionToken, resourceResponseBase.RequestCharge, resourceResponseBase.CollectionSizeUsage, resourceResponseBase.StatusCode);
```
jackliums: We should check with Yazan but I think we don't want to publish the CollectionSizeUsage here. I've found that the number was unreliable and does not match the actual collection size (i think it was more like partition size or something like that).

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosExceptionProcessor.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Net;
+using EnsureThat;
+using Microsoft.Azure.Documents;
+using Microsoft.Health.Abstractions.Exceptions;
+using Microsoft.Health.Fhir.Core.Features.Context;
+using Microsoft.Health.Fhir.CosmosDb.Features.Metrics;
+
+namespace Microsoft.Health.Fhir.CosmosDb.Features.Storage
+{
+    public class CosmosExceptionProcessor : ICosmosExceptionProcessor
```
jackliums: Maybe just change `CosmosMetricProcessor` to `CosmosResponseProcessor` and combine this logic in there? They all seem to be related to processing the response from Cosmos DB anyways.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Metric/MetricTests.cs
```diff
@@ -0,0 +1,93 @@
...
+        private readonly MetricTestFixture _fixture;
+        private readonly FhirClient _client;
+
+        private readonly MetricHandler _metricHandler;
+
+        public MetricTests(MetricTestFixture fixture)
+        {
+            _fixture = fixture;
+            _client = fixture.FhirClient;
+            _metricHandler = _fixture.MetricHandler;
```
jackliums: This could be null right?

src/Microsoft.Health.Fhir.CosmosDb/Features/Metrics/CosmosMetricProcessor.cs
```diff
@@ -0,0 +1,119 @@
...
+            if (statusCode.HasValue && statusCode == HttpStatusCode.TooManyRequests)
+            {
+                cosmosMetrics.ThrottledCount = 1;
+            }
+
+            cosmosMetrics.RequestCount = 1;
+
+            try
+            {
+                _mediator.Publish(cosmosMetrics, CancellationToken.None);
```
brendankowitz: For this try/catch to work you'll need to declare the method `async void` and `await` here. The method will return as soon as this is hit though, without waiting for publish to finish.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosResponseProcessor.cs
```diff
@@ -0,0 +1,149 @@
...
+
+            if (collectionSizeUsage.HasValue)
+            {
+                cosmosMetrics.CollectionSizeUsageKilobytes = collectionSizeUsage;
+            }
+
+            if (statusCode.HasValue && statusCode == HttpStatusCode.TooManyRequests)
+            {
+                cosmosMetrics.ThrottledCount = 1;
+            }
```
YazanMSFT: As discussed, we can switch this to a bool and remove RequestCount

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosResponseProcessor.cs
```diff
@@ -0,0 +1,149 @@
...
+        /// Adds request charge to the response headers and throws a <see cref="RequestRateExceededException"/>
+        /// if the status code is 429.
+        /// </summary>
+        /// <param name="ex">The exception</param>
+        public async Task ProcessException(Exception ex)
+        {
+            if (_fhirRequestContextAccessor.FhirRequestContext == null)
+            {
+                return;
+            }
```
YazanMSFT: Is this an expected case or should we log a warning?

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosResponseProcessor.cs
```diff
@@ -0,0 +1,149 @@
...
+
+            EnsureArg.IsNotNull(ex, nameof(ex));
+
+            if (ex is DocumentClientException dce)
+            {
+                await ProcessResponse(null, dce.RequestCharge, null, dce.StatusCode);
+
```
YazanMSFT: Not sure if this is a change you made or if it was already like that, but could you put the parameter name for the two nulls for clarity?

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosResponseProcessor.cs
brandonpollett: This behavior was previously in the extension method, so I would assume this is an expected case.

src/Microsoft.Health.Fhir.CosmosDb/Features/Metrics/CosmosMetricProcessor.cs
brandonpollett: Refactored to be `async`.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Metric/MetricTests.cs
brandonpollett: Fixed.

Issue #703, opened 2019-10-24
SMART on FHIR proxy token endpoint returns 415 when no `scope` field is missing in token response from Azure AD.


hansenms:
**Describe the bug**
SMART on FHIR proxy token endpoint returns 415 when no `scope` field is missing in token response from Azure AD.

**To Reproduce**
1. Set up a client application with no scopes asked for resource.  
2. Configure FHIR server with SMART on FHIR Proxy on.
3. Use SMART on FHIR App launcher to initiate token flow.

**Expected behavior**
Token response should be returned to client with `scope` field absent or empty.

**Actual behavior**
Returns 415 Invalid media type.

PR #704, opened 2019-10-24
Added null check for missing scopes

hansenms:
## Description
Adds check for null when no scopes are returned in SMART on FHIR proxy token endpoint.

## Related issues
Addresses #703.

### DIFF
--- a/src/Microsoft.Health.Fhir.Api/Controllers/AadSmartOnFhirProxyController.cs
+++ b/src/Microsoft.Health.Fhir.Api/Controllers/AadSmartOnFhirProxyController.cs
@@ -389,23 +389,27 @@ namespace Microsoft.Health.Fhir.Api.Controllers
             tokenResponse["client_id"] = clientId;
 
             // Replace fully qualifies scopes with short scopes and replace $
-            string[] scopes = tokenResponse["scope"].ToString().Split(' ');
-            var scopesBuilder = new StringBuilder();
+            string[] scopes = tokenResponse["scope"]?.ToString().Split(' ');
 
-            foreach (var s in scopes)
+            if (scopes != null)
             {
-                if (IsAbsoluteUrl(s))
-                {
-                    var scopeUri = new Uri(s);
-                    scopesBuilder.Append($"{scopeUri.Segments.Last().Replace('$', '/')} ");
-                }
-                else
-                {
-                    scopesBuilder.Append($"{s.Replace('$', '/')} ");
-                }
-            }
+                var scopesBuilder = new StringBuilder();
 
-            tokenResponse["scope"] = scopesBuilder.ToString().TrimEnd(' ');
+                foreach (var s in scopes)
+                {
+                    if (IsAbsoluteUrl(s))
+                    {
+                        var scopeUri = new Uri(s);
+                        scopesBuilder.Append($"{scopeUri.Segments.Last().Replace('$', '/')} ");
+                    }
+                    else
+                    {
+                        scopesBuilder.Append($"{s.Replace('$', '/')} ");
+                    }
+                }
+
+                tokenResponse["scope"] = scopesBuilder.ToString().TrimEnd(' ');
+            }
 
             return new ContentResult()
             {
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Fhir.Api/Controllers/AadSmartOnFhirProxyController.cs
```diff
@@ -389,23 +389,27 @@ public AadSmartOnFhirProxyController(IOptions<SecurityConfiguration> securityCon
             tokenResponse["client_id"] = clientId;
 
             // Replace fully qualifies scopes with short scopes and replace $
-            string[] scopes = tokenResponse["scope"].ToString().Split(' ');
-            var scopesBuilder = new StringBuilder();
+            string[] scopes = tokenResponse["scope"]?.ToString().Split(' ');
 
-            foreach (var s in scopes)
+            if (scopes != null)
```
brandonpollett: Should this throw a bad request exception?

hansenms: Apparently, it is OK to have a token response with no scopes (if none were requested), in which case, we should just not do anything with them. It is not an error on the AAD side and the token is valid, so I think this is OK? Any concens?

brandonpollett: Ok, makes sense. I just wanted to make sure that it wasn't an issue because we throw bad requests elsewhere in this code.

hansenms:
@brandonpollett @namadabu @feordin I added a unit test for this (and the replacements we do). Please have a look.

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Controllers/AadSmartOnFhirProxyControllerTests.cs
```diff
@@ -187,5 +189,41 @@ public void GivenMissingAudParamToV1AAd_WhenAuthorizeRequestAction_ThenRedirectR
 
             Assert.ThrowsAsync<AadSmartOnFhirProxyBadRequestException>(() => _controller.Token(grantType, compoundCode, redirectUri, clientId, clientSecret)).Wait();
         }
+
+        [Theory]
+        [InlineData(null, null)]
+        [InlineData("user_impersonation test.scope", "user_impersonation test.scope")]
+        [InlineData("patient$Observation.read", "patient/Observation.read")]
+        [InlineData("patient$Observation.read patient$Encounter.read", "patient/Observation.read patient/Encounter.read")]
+        public void GivenScopesInTokenResponse_WhenTokenRequestAction_ThenCorrectScopesReturned(string inScopes, string outScopes)
```
brandonpollett: nit: parameter names - probably something like `tokenScopes` and `expectedScopes`

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Controllers/AadSmartOnFhirProxyControllerTests.cs
```diff
@@ -187,5 +189,41 @@ public void GivenMissingAudParamToV1AAd_WhenAuthorizeRequestAction_ThenRedirectR
...
+        {
+            const string grantType = "authorization_code";
+            const string clientId = "1234";
+            const string clientStringPlaceHolder = "XYZ";
+            const string compoundCode = "eyAiY29kZSIgOiAiZm9vIiB9";
+            Uri redirectUri = new Uri("https://localhost");
+
+            _urlResolver.ResolveRouteNameUrl(Arg.Any<string>(), Arg.Any<IDictionary<string, object>>()).Returns(redirectUri);
+
+            JObject content = JObject.Parse("{}");
```
brandonpollett: Can you just use `new JObject()` or something equivalent here?

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Controllers/AadSmartOnFhirProxyControllerTests.cs
```diff
@@ -187,5 +189,41 @@ public void GivenMissingAudParamToV1AAd_WhenAuthorizeRequestAction_ThenRedirectR
...
         }
+
+        [Theory]
+        [InlineData(null, null)]
+        [InlineData("user_impersonation test.scope", "user_impersonation test.scope")]
+        [InlineData("patient$Observation.read", "patient/Observation.read")]
+        [InlineData("patient$Observation.read patient$Encounter.read", "patient/Observation.read patient/Encounter.read")]
+        public void GivenScopesInTokenResponse_WhenTokenRequestAction_ThenCorrectScopesReturned(string inScopes, string outScopes)
+        {
+            const string grantType = "authorization_code";
```
brandonpollett: `const` isn't really needed here.

PR #706, opened 2019-10-25
Capability statement enhancements/refactor

brendankowitz:
## Description

- Large refactor of Capability generation
  - Removes DefaultCapabilities.json and AllCapabilities.json
  - Moves capability code into Core
- Adds ICapabilityStatementBuilder

## Related issues
Addresses [AB#70801](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/70801).

## Testing
- Unit tests
- E2E Tests


src/Microsoft.Health.Fhir.Shared.Api/Modules/FeatureFlags/XmlFormatter/XmlFormatterConfiguration.cs
```diff
@@ -39,15 +40,16 @@ public void PostConfigure(string name, MvcOptions options)
         {
             if (_featureConfiguration.SupportsXml)
             {
-                _configuredConformanceProvider.ConfigureOptionalCapabilities(statement => statement.Format = statement.Format.Concat(new[] { KnownContentTypes.XmlContentType }));
+                _configuredConformanceProvider
+                    .ConfigureOptionalCapabilities(statement => statement.Format.Add(KnownContentTypes.XmlContentType));
```
brendankowitz: This might not be needed

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -48,20 +49,23 @@ internal class SqlServerFhirDataStore : IFhirDataStore, IProvideCapability
...
             EnsureArg.IsNotNull(searchParameterTypeMap, nameof(searchParameterTypeMap));
+            EnsureArg.IsNotNull(modelInfoProvider, nameof(modelInfoProvider));
             EnsureArg.IsNotNull(upsertResourceTvpGenerator, nameof(upsertResourceTvpGenerator));
             EnsureArg.IsNotNull(logger, nameof(logger));
             EnsureArg.IsNotNull(coreFeatures, nameof(coreFeatures));
 
             _configuration = configuration;
             _model = model;
             _searchParameterTypeMap = searchParameterTypeMap;
+            _modelInfoProvider = modelInfoProvider;
```
brendankowitz: Is this still being used?

src/Microsoft.Health.Fhir.Core/Features/Conformance/CapabilityStatementBuilder.cs
```diff
@@ -0,0 +1,230 @@
...
+                };
+                listedRestComponent.Resource.Add(resourceComponent);
+            }
+
+            action(resourceComponent);
+
+            return this;
+        }
+
+        public ICapabilityStatementBuilder TryAddRestInteraction(string resourceType, string interaction)
```
jackliums: nit: there is no "Try" here right? It will always add the interaction?

src/Microsoft.Health.Fhir.Core/Features/Conformance/CapabilityStatementBuilder.cs
```diff
@@ -0,0 +1,230 @@
...
+                    c.SearchParam.Add(searchParam);
+                }
+            });
+
+            return this;
+        }
+
+        public ICapabilityStatementBuilder AddDefaultResourceInteractions()
+        {
+            foreach (var resource in _modelInfoProvider.GetResourceTypeNames())
```
jackliums: nit: string instead of var?

src/Microsoft.Health.Fhir.Core/Features/Conformance/CapabilityStatementBuilder.cs
brendankowitz: All the collections are HashSets so it's possible the call will be ignored. I'd also be ok with changing it to `AddRestInteraction`

src/Microsoft.Health.Fhir.Core/Features/Conformance/CapabilityStatementBuilder.cs
```diff
@@ -0,0 +1,230 @@
...
+            {
+                if (string.Equals(resource, KnownResourceTypes.Parameters, StringComparison.Ordinal))
```
jackliums: nit: maybe a bit of comment here will help people later? Why we are skipping certain types and what not?

src/Microsoft.Health.Fhir.Core/Features/Conformance/CapabilityStatementBuilder.cs
```diff
@@ -0,0 +1,230 @@
...
+                        Name = x.Name,
+                        Type = x.Type,
+                        Definition = x.Url,
+                        Documentation = x.Description,
+                    }));
+
+                    TryAddRestInteraction(resource, TypeRestfulInteraction.SearchType);
+                }
+
+                TryAddRestInteraction(resource, TypeRestfulInteraction.HistoryType);
```
jackliums: Should these two be added in the DefaultResourceInteractions above? Doesn't seem to be related to search paramters?

src/Microsoft.Health.Fhir.Core/Features/Conformance/CapabilityStatementBuilder.cs
jackliums: Yeah after reading towards the end where JsonSerializer was created, I just realized that it might be hashset.

src/Microsoft.Health.Fhir.Core/Features/Conformance/Models/PropertyEqualityComparer.cs
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        public bool Equals(T x, T y)
+        {
+            if (x == null || y == null)
+            {
+                return false;
+            }
+
+            return _propertiesToCompare.All(property => ReferenceEquals(property(x), property(y)) || property(x) == property(y));
```
jackliums: nit: save the variable so it doesn't have to evaluate again?

src/Microsoft.Health.Fhir.Core/Features/Conformance/CapabilityStatementBuilder.cs
```diff
@@ -0,0 +1,230 @@
...
+            return this;
+        }
+
+        public ICapabilityStatementBuilder TryAddRestInteraction(string resourceType, string interaction)
+        {
+            EnsureArg.IsNotNullOrEmpty(resourceType, nameof(resourceType));
+            EnsureArg.IsNotNullOrEmpty(interaction, nameof(interaction));
+            EnsureArg.IsTrue(_modelInfoProvider.IsKnownResource(resourceType), nameof(resourceType), x => GenerateTypeErrorMessage(x, resourceType));
+
+            UpdateRestResourceComponent(resourceType, c =>
```
jackliums: Is the capability statement built before the server starts? (e.g., does this need to be thread-safe?)

src/Microsoft.Health.Fhir.Core/Features/Conformance/IProvideCapability.cs
```diff
@@ -13,7 +13,7 @@ public interface IProvideCapability
         /// <summary>
         /// Allows this component to add capabilities into the specified ListedCapabilityStatement.
         /// </summary>
-        /// <param name="statement">The ListedCapabilityStatement to be added to.</param>
-        void Build(IListedCapabilityStatement statement);
+        /// <param name="builder">The ListedCapabilityStatement to be added to.</param>
```
jackliums: nit: comment needs to be updated.

src/Microsoft.Health.Fhir.Core/Features/Conformance/ListedConformanceStatementExtensions.cs
```diff
@@ -4,23 +4,19 @@
...
-        bool? ConditionalCreate { get; set; }
-
-        bool? ReadHistory { get; set; }
-
-        IList<string> SearchRevInclude { get; }
-
-        IList<string> SearchInclude { get; }
-
-        void AddResourceVersionPolicy(string policy);
+            return restComponents.Single(x => x.Mode == ListedCapabilityStatement.ServerMode);
```
jackliums: The `ListedCapabilityStatement.Rest` is already a hashset using `PropertyEqualityComparer` with Mode. Would that be utilized here when finding element. If not, can we utilize it?

src/Microsoft.Health.Fhir.Core/Features/Conformance/Models/PropertyEqualityComparer.cs
```diff
@@ -0,0 +1,50 @@
...
+        }
+
+        public int GetHashCode(T obj)
```
jackliums: nit: obj could be null

src/Microsoft.Health.Fhir.Core/Features/Conformance/Serialization/DefaultOptionHashSetJsonConverter.cs
```diff
@@ -0,0 +1,65 @@
...
+
+            serializer.Serialize(writer, defaultOption);
+        }
+
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
+            throw new NotImplementedException();
+        }
+
+        public override bool CanConvert(Type objectType)
```
jackliums: nit: set CanRead = false?

src/Microsoft.Health.Fhir.Core/Features/Conformance/Serialization/DefaultOptionHashSetJsonConverter.cs
```diff
@@ -0,0 +1,65 @@
...
+{
+    internal class DefaultOptionHashSetJsonConverter : JsonConverter
+    {
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            EnsureArg.IsNotNull(writer, nameof(writer));
+            EnsureArg.IsNotNull(value, nameof(value));
+            EnsureArg.IsNotNull(serializer, nameof(serializer));
+
+            dynamic defaultOption = ((dynamic)value).DefaultOption;
```
jackliums: Instead of using dynamic, could define an interface like IDefaultOption or something which expose a property `object DefaultOption` then cast the object to that interface and access the property? It guards against potential renaming in the future and also no performance hit?

src/Microsoft.Health.Fhir.Core/Features/Conformance/Serialization/DefaultOptionHashSetJsonConverter.cs
```diff
@@ -0,0 +1,65 @@
...
+                    }
+
+                    if (defaultOption == item)
+                    {
+                        exists = true;
+                        break;
+                    }
+                }
+
+                if (!exists)
```
jackliums: Maybe a bit comment about why this logic exist would be nice. When would the default option won't exist and why the first item would be okay in that case?

jackliums: Sounds like even though the item is hashset, but it will only serialize one object? is that correct understanding?

src/Microsoft.Health.Fhir.Core/Features/Conformance/Serialization/EnumLiteralJsonConverter.cs
```diff
@@ -0,0 +1,43 @@
...
+namespace Microsoft.Health.Fhir.Core.Features.Conformance.Serialization
+{
+    internal class EnumLiteralJsonConverter : JsonConverter
+    {
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            EnsureArg.IsNotNull(writer, nameof(writer));
+            EnsureArg.IsNotNull(serializer, nameof(serializer));
+
+            if (value is Enum obj)
```
jackliums: FhirUtility already has a method called `GetLiteral`, which I think does this. It's in Hl7.Fhir.Support NuGet package which i think is version agnostic?

src/Microsoft.Health.Fhir.Core/Features/Conformance/SystemConformanceProvider.cs
```diff
@@ -0,0 +1,109 @@
...
+                {
+                    _configurationUpdates.Clear();
+                    _sem.Release();
+                }
+            }
+
+            return _listedCapabilityStatement;
+        }
+
+        public void ConfigureOptionalCapabilities(Action<ListedCapabilityStatement> builderAction)
```
jackliums: Is this guaranteed to be called before GetCapabilityStatementAsync is called?

jackliums: I guess you do have guard against it. I wonder if we should consider making sure the capability statement is built as part of the warm up before the server starts taking requests?

src/Microsoft.Health.Fhir.Core/Models/CodingInfo.cs
```diff
@@ -21,5 +22,10 @@ public CodingInfo(string system, string code)
         public string System { get; set; }
 
         public string Code { get; set; }
+
```
jackliums: Do we still need this? `Coding` seems to be in Hl7.Fhir.Support NuGet package now so that's version agnostic right?

src/Microsoft.Health.Fhir.Core/Features/Conformance/CapabilityStatementBuilder.cs
brendankowitz: That should be fine. History capabilities are tied to search and not the DataStore which is why they originated here.

src/Microsoft.Health.Fhir.Core/Features/Conformance/CapabilityStatementBuilder.cs
brendankowitz: We lock inside `SystemConformanceProvider`

src/Microsoft.Health.Fhir.Core/Features/Conformance/ListedConformanceStatementExtensions.cs
brendankowitz: I don't think so? PropertyEqualityComparer is ensuring only 1 RestComponent of each Mode is added (Server/Client). This is explicitly fetching the Server rest component.

src/Microsoft.Health.Fhir.Core/Features/Conformance/Models/PropertyEqualityComparer.cs
brendankowitz: Not sure what you mean? This evaluates all specified properties on ObjA against the same properties on ObjB. The instances are passed in for comparison. `_propertiesToCompare` is an array of compiled functions (getters), so should also be fast.

src/Microsoft.Health.Fhir.Core/Features/Conformance/Serialization/DefaultOptionHashSetJsonConverter.cs
brendankowitz: Yes, even though our server can support multiple options, the capability statement only has one (in these cases). In a later task I want to use these sets to create a JSON Schema that defines all the valid options you can use to override

brendankowitz: Also, perhaps the first item is not ok, perhaps it should throw an exception if the option is misconfigured.

src/Microsoft.Health.Fhir.Core/Features/Conformance/SystemConformanceProvider.cs
brendankowitz: Makes sense

src/Microsoft.Health.Fhir.Core/Models/CodingInfo.cs
brendankowitz: We don't need this class anymore, I think I'll defer this to a different PR

src/Microsoft.Health.Fhir.Core/Models/ModelInfoProvider.cs
```diff
@@ -89,6 +90,12 @@ public static EvaluationContext GetEvaluationContext(ITypedElement element)
             return _modelInfoProvider.GetEvaluationContext(element);
         }
 
+        public static IStructureDefinitionSummaryProvider GetStructureDefinitionSummaryProvider()
+        {
+            EnsureProvider();
+            return _modelInfoProvider.GetStructureDefinitionSummaryProvider();
```
jackliums: nit: could define a property and wrap around _modelInfoProvider

```
private static IModelInfoProvider
{
    get
    {
        EnsureProvider();

        return _modelInfoProvider;
    }
}
```

src/Microsoft.Health.Fhir.Shared.Core/VersionSpecificModelInfoProvider.cs
```diff
@@ -59,5 +60,10 @@ public EvaluationContext GetEvaluationContext(ITypedElement element)
         {
             return new FhirEvaluationContext(element);
         }
+
+        public IStructureDefinitionSummaryProvider GetStructureDefinitionSummaryProvider()
+        {
+            return new PocoStructureDefinitionSummaryProvider();
```
jackliums: Does this need new instance every time?

src/Microsoft.Health.Fhir.Shared.Core/VersionSpecificModelInfoProvider.cs
```diff
@@ -42,7 +43,7 @@ public bool IsKnownCompartmentType(string compartmentType)
 
         public string[] GetResourceTypeNames()
         {
-            return Enum.GetNames(typeof(ResourceType));
+            return ModelInfo.SupportedResources.ToArray();
```
jackliums: Seems like this could be cached.

src/Microsoft.Health.Fhir.Core/Features/Conformance/Models/PropertyEqualityComparer.cs
jackliums: `property(x)` and `property(y)` only need to be evaluated once instead of twice in the delegate.

src/Microsoft.Health.Fhir.Core/Features/Conformance/Models/ListedCapabilityStatement.cs
```diff
@@ -0,0 +1,56 @@
...
+using System.Collections.Generic;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Health.Fhir.Core.Features.Conformance.Models
+{
+    public class ListedCapabilityStatement
+    {
+        internal const string ServerMode = "server";
+        internal const string CapabilityStatement = "CapabilityStatement";
```
jackliums: This doesn't seem to be used?

src/Microsoft.Health.Fhir.Core/Features/Conformance/Models/ListedCapabilityStatement.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        public ListedCapabilityStatement()
+        {
+            Status = new DefaultOptionHashSet<string>("draft");
+            Kind = new DefaultOptionHashSet<string>("capability");
+            Rest = new HashSet<ListedRestComponent>(new PropertyEqualityComparer<ListedRestComponent>(x => x.Mode));
+            Format = new HashSet<string>();
```
jackliums: nit: StringComparer

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Conformance/ConformanceBuilderTests.cs
```diff
@@ -0,0 +1,89 @@
...
+            bool hasCreate = (bool)statement.Scalar($"{ResourceQuery("AuditEvent")}.interaction.where(code = '{TypeRestfulInteraction.Create}').exists()");
+            bool noUpdate = (bool)statement.Scalar($"{ResourceQuery("AuditEvent")}.interaction.where(code = '{TypeRestfulInteraction.Update}').exists()");
+            bool noDelete = (bool)statement.Scalar($"{ResourceQuery("AuditEvent")}.interaction.where(code = '{TypeRestfulInteraction.Delete}').exists()");
+
+            Assert.True(hasCreate);
+            Assert.False(noUpdate);
+            Assert.False(noDelete);
+        }
+
+        private static string ResourceQuery(string resource)
```
jackliums: Should we also add some tests for Parameter and Binary since those also have some special logic. And one test for positive case like Pantient or something to make sure it has the correct capability built?

brendankowitz: Yes, added a test for Parameter, removed logic for Binary (I didn't find anything specific in the spec)

src/Microsoft.Health.Fhir.Core/Features/Conformance/Models/ListedCapabilityStatement.cs
```diff
@@ -0,0 +1,55 @@
...
+
+namespace Microsoft.Health.Fhir.Core.Features.Conformance.Models
+{
+    public class ListedCapabilityStatement
+    {
+        internal const string ServerMode = "server";
+
+        public ListedCapabilityStatement()
+        {
+            Status = new DefaultOptionHashSet<string>("draft", StringComparer.Ordinal);
```
fmeira, line 19: Hi @brendankowitz, how should the CapabilityStatement.Status be set to "active" if the model does not allow for changing it?

PR #710, opened 2019-10-28
Transaction Support for Bundle Processing

poadhika:
## Description
This PR introduces basic support for transaction processing. It uses the same structure as batch processing except for starting and ending of transaction. This implementation uses transactionscope to support transaction for multiple sqlconnections.

## Related issues
Addresses [https://microsofthealth.visualstudio.com/Health/_workitems/edit/68731](url)

## Testing
This change was tested manually as the feature is still not enabled by default so E2E tests cannot be written/updated. Some unit tests has been added.


src/Microsoft.Health.Fhir.Core/Features/Persistence/IFhirDataStore.cs
```diff
@@ -20,5 +20,11 @@ public interface IFhirDataStore
         Task<ResourceWrapper> GetAsync(ResourceKey key, CancellationToken cancellationToken);
 
         Task HardDeleteAsync(ResourceKey key, CancellationToken cancellationToken);
+
+        void BeginTransactionScope();
+
+        void CompleteTransactionScope();
+
+        void Dispose();
```
brendankowitz: I think if you have a `Dispose` method you should be taking a dependency on the IDisposable interface

src/Microsoft.Health.Fhir.Core/Features/Persistence/IFhirDataStore.cs
```diff
@@ -20,5 +20,11 @@ public interface IFhirDataStore
         Task<ResourceWrapper> GetAsync(ResourceKey key, CancellationToken cancellationToken);
 
         Task HardDeleteAsync(ResourceKey key, CancellationToken cancellationToken);
+
+        void BeginTransactionScope();
+
+        void CompleteTransactionScope();
```
brendankowitz: Why are these defined in TransactionHandler and also in this interface?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -266,5 +268,20 @@ public void Build(IListedCapabilityStatement statement)
                 statement.TryAddRestInteraction(SystemRestfulInteraction.Transaction);
             }
         }
+
+        public void Dispose()
+        {
+            _transactionScope?.Dispose();
+        }
+
+        public void BeginTransactionScope()
```
brendankowitz: I think we should follow more closely with the .net patterns of, so in the datastore it only needs 1 method for setup:
```csharp
ITransaction BeginTransaction()
```
ITransaction could have a `SqlTransaction` implementation.

ITransaction has the other methods:
```csharp
public interface ITransaction : IDisposable {
void Complete();
void Dispose();
}
```

In calling code, it would then be easy to wrap in a using:

```csharp
using(var transaction = _transactionHandler.BeginTransaction()) {
// do things
transaction.Commit();
}
```
Any errors would call Dispose without Commit and cause a rollback

Thoughts?

src/Microsoft.Health.Fhir.Core/Features/Persistence/TransactionHandler.cs
```diff
@@ -0,0 +1,36 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using EnsureThat;
+
+namespace Microsoft.Health.Fhir.Core.Features.Persistence
+{
+    public class TransactionHandler : ITransactionHandler
```
brandonpollett: Would the code still work if we required `IFhirDataStore` to implement `ITransactionHandler` and didn't have this class at all?

src/Microsoft.Health.Fhir.Core/Features/Persistence/BundleProcessingStatus.cs
```diff
@@ -0,0 +1,13 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Fhir.Core.Features.Persistence
+{
+    public enum BundleProcessingStatus
+    {
+        SUCCEEDED,
```
brandonpollett: nit: Enums in this code base aren't all capitals.

src/Microsoft.Health.Fhir.Shared.Api/Features/ActionResults/FhirResult.cs
```diff
@@ -68,6 +92,17 @@ public static FhirResult NotFound()
             };
         }
 
+        /// <summary>
+        /// Returns a BadRequest response
+        /// </summary>
+        public static FhirResult BadRequest()
```
brandonpollett: Is this used anywhere?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -66,21 +75,58 @@ public async Task<BundleResponse> Handle(BundleRequest bundleRequest, Cancellati
...
+                    _transactionHandler.CompleteTransactionScope();
+                }
+                finally
+                {
+                    _transactionHandler.Dispose();
+                }
+            }
+            catch (TransactionAbortedException ex)
+            {
+                bundleProcessingStatus = BundleProcessingStatus.FAILED;
```
brandonpollett: Could we just throw a new exception type here and handle the exception like the rest of the exceptions in the code base and return an operation outcome with a bad request status? Per the [spec](https://www.hl7.org/fhir/http.html#transaction-response) `For a failed transaction, the server returns a single OperationOutcome instead of a Bundle.`

src/Microsoft.Health.Fhir.Shared.Api/Features/ActionResults/FhirResult.cs
```diff
@@ -46,6 +48,28 @@ public static FhirResult Create(ResourceElement resource, HttpStatusCode statusC
...
+        /// <param name="bundleResponse">The resonse bundle.</param>
+        public static FhirResult Create(BundleResponse bundleResponse)
+        {
+            EnsureArg.IsNotNull(bundleResponse, nameof(bundleResponse));
+
+            if (bundleResponse.BundleProcessingStatus == BundleProcessingStatus.FAILED)
+            {
+                return new FhirResult(bundleResponse.Bundle)
+                {
+                    StatusCode = HttpStatusCode.BadRequest,
```
jackliums: Bundle could be failed to process for various reasons. Is it always going to be BadRequest (meaning client has send bad request)?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -66,21 +75,58 @@ public async Task<BundleResponse> Handle(BundleRequest bundleRequest, Cancellati
...
+                {
+                    Type = Hl7.Fhir.Model.Bundle.BundleType.BatchResponse,
+                };
+
+                await ExecuteAllRequests(responseBundle);
+                return new BundleResponse(responseBundle.ToResourceElement());
             }
 
-            await FillRequestLists(bundleResource.Entry);
+            if (bundleResource.Type == Hl7.Fhir.Model.Bundle.BundleType.Transaction)
```
jackliums: nit: else if?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -66,21 +75,58 @@ public async Task<BundleResponse> Handle(BundleRequest bundleRequest, Cancellati
...
+            {
+                var responseBundle = new Hl7.Fhir.Model.Bundle
+                {
+                    Type = Hl7.Fhir.Model.Bundle.BundleType.TransactionResponse,
+                };
+
+                return await ExecuteTransactionForAllRequests(responseBundle);
```
jackliums: CosmosDB doesn't support trasaction. Instead of failing with 500 (since the CosmosFhirDataStore would throw NotImplementedException), should this method check against some sort of "capability" and returned not supported exception if the operation is not supported?

src/Microsoft.Health.Fhir.Shared.Api/Features/ActionResults/FhirResult.cs
```diff
@@ -46,6 +48,28 @@ public static FhirResult Create(ResourceElement resource, HttpStatusCode statusC
...
 
+        /// <summary>
+        /// Creates a FHIR result with the specified parameters
+        /// </summary>
+        /// <param name="bundleResponse">The resonse bundle.</param>
+        public static FhirResult Create(BundleResponse bundleResponse)
+        {
+            EnsureArg.IsNotNull(bundleResponse, nameof(bundleResponse));
+
+            if (bundleResponse.BundleProcessingStatus == BundleProcessingStatus.FAILED)
```
jackliums: Is this needed? If the operation failed, most likely you will be returning OperationOutcome instead of Bundle for transaction.

poadhika: This will not be needed. I have a updated PR that use OperationOutcome.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
poadhika: Yes,  the same behavior happens here. Execution will reach to bundlehandler after validating the capabilitystatement. For cosmosDB, transaction will be turned off and Not Supported exception will be thrown.

src/Microsoft.Health.Fhir.Shared.Api/Features/ActionResults/FhirResult.cs
poadhika: No Will remove it.

src/Microsoft.Health.Fhir.Core/Features/Persistence/BundleProcessingStatus.cs
poadhika: This enum will not be reuired as i will be using operationoutcome. I will  delete this file in my next update.

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/ActionResults/FhirResultTests.cs
```diff
@@ -58,5 +62,39 @@ public void GivenANotFoundStatus_WhenReturningAResult_ThenTheStatusCodeIsSetCorr
             Assert.Null(result.Result);
             Assert.Equal(HttpStatusCode.NotFound, result.StatusCode.GetValueOrDefault());
         }
+
+        [Fact]
+        public void GivenASuccessfulBundleProcessingStatus_WhenReturningAResult_ThenBadRequestIsReturned()
```
deepak-ms: typo : "BadRequest"

poadhika: Updated the test with based on new change

src/Microsoft.Health.Fhir.Shared.Api/Features/ActionResults/FhirResult.cs
poadhika: I have now parsed the status code from operation outcome and used that to map the proper exception but still defaults to badrequest.

src/Microsoft.Health.Fhir.Core/Features/Persistence/TransactionHandler.cs
poadhika: Implemented as suggested

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -66,21 +75,58 @@ public async Task<BundleResponse> Handle(BundleRequest bundleRequest, Cancellati
...
+            }
+
+            throw new MethodNotAllowedException(Microsoft.Health.Fhir.Api.Resources.UnsupportedOperation);
```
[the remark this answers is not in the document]
poadhika: updated the correct message type

src/Microsoft.Health.Fhir.Core/Features/Persistence/ITransactionHandler.cs
```diff
@@ -0,0 +1,16 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Fhir.Core.Features.Persistence
+{
+    public interface ITransactionHandler
```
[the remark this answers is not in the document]
poadhika: Done

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
poadhika: Done as per suuggestion.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -42,47 +46,113 @@ public partial class BundleHandler : IRequestHandler<BundleRequest, BundleRespon
...
+                _logger.LogError("Error while parsing transaction exception. Throwing BadRequest as a default exception.");
+                throw new BadRequestException("Transaction aborted while processing request");
+            }
 
             return new BundleResponse(responseBundle.ToResourceElement());
         }
 
+        private static void ThrowTransactionException(Hl7.Fhir.Model.Bundle responseBundle)
+        {
+            EntryComponent responseBundleEntryComponent = responseBundle.Entry?.First();
```
poadhika: Here I am parsing the status code from operationcome.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -167,6 +242,18 @@ private async Task ExecuteRequests(Hl7.Fhir.Model.Bundle responseBundle, Hl7.Fhi
                         if (entryComponentResource.ResourceType == ResourceType.OperationOutcome)
                         {
                             entryComponent.Response.Outcome = entryComponentResource;
+
+                            if (responseBundle.Type == Hl7.Fhir.Model.Bundle.BundleType.TransactionResponse)
+                            {
+                                _bundleProcessingStatus = BundleProcessingStatus.Failed;
```
brandonpollett: Could we just throw the transaction exception here and let it bubble up?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -167,6 +242,18 @@ private async Task ExecuteRequests(Hl7.Fhir.Model.Bundle responseBundle, Hl7.Fhi
...
+                                entryComponent.FullUrl = request.HttpContext.Request.Path;
+
+                               // transaction responsebundle only contains operationoutcome in case of failure.
```
brandonpollett: nit: inconsistent spacing

src/Microsoft.Health.Fhir.Core/Features/Persistence/ITransactionHandler.cs
```diff
@@ -0,0 +1,15 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Transactions;
+
+namespace Microsoft.Health.Fhir.Core.Features.Persistence
+{
+    public interface ITransactionHandler : IDisposable
+    {
+        TransactionScope BeginTransaction();
```
brandonpollett: We might want to think about abstracting the `TransactionScope` here to something else. It seems like an implementation detail that could differ between datastores.

src/Microsoft.Health.Fhir.Shared.Api/Features/ActionResults/FhirResult.cs
brandonpollett: Is it only used in the tests? If so, then this should be in a test project.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
poadhika: Actually we can. Not sure why that never occurred to me. And that emoji is my reaction after I found out it can be done this way :)

src/Microsoft.Health.Fhir.Core/Features/Persistence/ITransactionHandler.cs
poadhika: Updated.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
```diff
@@ -269,5 +269,14 @@ public void Build(IListedCapabilityStatement statement)
                 statement.TryAddRestInteraction(SystemRestfulInteraction.Batch);
             }
         }
+
+        ITransactionScope ITransactionHandler.BeginTransaction()
+        {
+            throw new NotImplementedException();
```
jackliums: nit: i think NotSupportedException is more appropriate.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -66,23 +76,67 @@ public async Task<BundleResponse> Handle(BundleRequest bundleRequest, Cancellati
...
+                {
+                    Type = Hl7.Fhir.Model.Bundle.BundleType.TransactionResponse,
+                };
 
-            await FillRequestLists(bundleResource.Entry);
+                return await ExecuteTransactionForAllRequests(responseBundle);
+            }
 
-            var responseBundle = new Hl7.Fhir.Model.Bundle
+            throw new MethodNotAllowedException(Microsoft.Health.Fhir.Api.Resources.OnlyCertainBundleTypesSupported);
```
brandonpollett: We should update this message to say something like `Bundles of type '{type}' are not supported.`

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -66,23 +76,67 @@ public async Task<BundleResponse> Handle(BundleRequest bundleRequest, Cancellati
...
+                    await ExecuteAllRequests(responseBundle);
 
-            await ExecuteAllRequests(responseBundle);
+                    transaction.Complete();
+                }
+            }
+            catch (TransactionAbortedException)
+            {
+                _logger.LogError("Failed to commit a transaction. Throwing BadRequest as a default exception.");
+                throw new BadRequestException("Transaction aborted while processing request");
```
brandonpollett: This text should be in `resources.resx`

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -66,23 +76,67 @@ public async Task<BundleResponse> Handle(BundleRequest bundleRequest, Cancellati
...
+            }
 
             return new BundleResponse(responseBundle.ToResourceElement());
         }
 
+        private static void ThrowTransactionException(EntryComponent responseBundle)
+        {
+            int responseStatus = int.Parse(responseBundle?.Response?.Status);
+            string errorMessage = $"Transaction aborted while processing request url: '{responseBundle?.FullUrl}'. {responseBundle?.Response?.Outcome}";
```
brandonpollett: What does this error message end up looking like to the client? Is it an operation outcome with a message of some string followed by an encoded version of the internal operation outcome? We may be able to instead just return multiple issues in the operation outcome.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -167,6 +221,11 @@ private async Task ExecuteRequests(Hl7.Fhir.Model.Bundle responseBundle, Hl7.Fhi
                         if (entryComponentResource.ResourceType == ResourceType.OperationOutcome)
                         {
                             entryComponent.Response.Outcome = entryComponentResource;
+
+                            if (responseBundle.Type == Hl7.Fhir.Model.Bundle.BundleType.TransactionResponse)
```
brandonpollett: We should verify that we don't return an `OperationOutcome` on any successful operations. I think a delete may return one per the spec. See: https://www.hl7.org/fhir/http.html#summary

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -66,23 +76,81 @@ public async Task<BundleResponse> Handle(BundleRequest bundleRequest, Cancellati
...
+                _logger.LogError("Failed to commit a transaction. Throwing BadRequest as a default exception.");
+                throw new BadRequestException(Api.Resources.GeneralTransactionAbortedError);
+            }
 
             return new BundleResponse(responseBundle.ToResourceElement());
         }
 
+        private static void ThrowTransactionException(HttpContext httpContext, Resource responseOperationOutcome)
+        {
+            var operationOutcomeIssues = GetOperationOutcomeIssues(((OperationOutcome)responseOperationOutcome).Issue);
```
brandonpollett: Can we change the incoming parameter to the type `OperationOutcome` since this is already checked/expected in the code that calls this method?

src/Microsoft.Health.Fhir.Core/Exceptions/PreconditionFailedException.cs
```diff
@@ -20,5 +21,19 @@ public PreconditionFailedException(string message)
                     OperationOutcomeConstants.IssueType.Required,
                     message));
         }
+
+        public PreconditionFailedException(string message, List<OperationOutcomeIssue> operationOutcomeIssues)
```
brandonpollett: Would it make more sense to just have something like a `TransactionFailedException` that takes in a status code and list of operation outcomes and does this logic? That way we're not repeating this in three different places since our main goal is to flow through the status code? Status code could even just default to `400`.

Inside of `OperationOutcomeExceptionFilterAttribute.cs` we could set the status code to the one that's set on the type.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -66,23 +76,81 @@ public async Task<BundleResponse> Handle(BundleRequest bundleRequest, Cancellati
...
+                {
+                    Type = Hl7.Fhir.Model.Bundle.BundleType.BatchResponse,
+                };
+
+                await ExecuteAllRequests(responseBundle);
+                return new BundleResponse(responseBundle.ToResourceElement());
+            }
+            else if (bundleResource.Type == Hl7.Fhir.Model.Bundle.BundleType.Transaction)
             {
-                throw new MethodNotAllowedException(Microsoft.Health.Fhir.Api.Resources.OnlyCertainBundleTypesSupported);
```
brandonpollett: If we're no longer using `OnlyCertainBundleTypesSupported` can you remove it?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -66,23 +76,81 @@ public async Task<BundleResponse> Handle(BundleRequest bundleRequest, Cancellati
...
+            }
 
             return new BundleResponse(responseBundle.ToResourceElement());
         }
 
+        private static void ThrowTransactionException(HttpContext httpContext, Resource responseOperationOutcome)
+        {
+            var operationOutcomeIssues = GetOperationOutcomeIssues(((OperationOutcome)responseOperationOutcome).Issue);
+
+            var errorMessage = string.Format(Api.Resources.TransactionAbortedError, httpContext.Request.Method, httpContext.Request.Path);
```
brandonpollett: Does `httpContext.Request.Path` contain the querystring values also? If not, we could have two different `GET` calls for `Patient` and `Patient?_lastUpdated=gt2019` and they may show as the same in the error message.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -66,23 +76,81 @@ public async Task<BundleResponse> Handle(BundleRequest bundleRequest, Cancellati
...
+                case (int)HttpStatusCode.PreconditionFailed:
+                    throw new PreconditionFailedException(errorMessage, operationOutcomeIssues);
+                case (int)HttpStatusCode.NotFound:
+                    throw new ResourceNotFoundException(errorMessage, operationOutcomeIssues);
+                default:
+                    throw new BadRequestException(errorMessage, operationOutcomeIssues);
+            }
+        }
+
+        private static List<OperationOutcomeIssue> GetOperationOutcomeIssues(List<OperationOutcome.IssueComponent> operationoutcomeIssueList)
```
brandonpollett: This could possibly make sense as an extension.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
poadhika: It has already been removed.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
poadhika: Yes. It has querystring values as well.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -266,5 +267,16 @@ public void Build(IListedCapabilityStatement statement)
                 statement.TryAddRestInteraction(SystemRestfulInteraction.Transaction);
             }
         }
+
+        public ITransactionScope BeginTransaction()
+        {
+            _transactionScope = new DefaultTransactionScope();
```
brendankowitz: Does SqlServerFhirDataStore need to keep a reference to the transaction in this case? I feel that it is creating and assigning ownership to the calling component. Thoughts?

src/Microsoft.Health.Fhir.Core.UnitTests/Persistence/TransactionFailedExceptionTests.cs
```diff
@@ -0,0 +1,70 @@
...
+        [Fact]
+        public void GivenAnEmptyListOfOperationOutComeIssues_WhenInitialized_ThenOneOperationOutcomeIssueShouldBeAdded()
+        {
+            string message = "message";
+            HttpStatusCode statusCode = HttpStatusCode.Processing;
+
+            var exception = new TransactionFailedException(message, statusCode, new List<OperationOutcomeIssue>());
+
+            Assert.NotNull(exception.Issues);
+            Assert.Equal(1, exception.Issues.Count);
```
brandonpollett: nit: `Assert.Single(exception.Issues.Count`

src/Microsoft.Health.Fhir.Core.UnitTests/Persistence/TransactionFailedExceptionTests.cs
```diff
@@ -0,0 +1,70 @@
...
+
+            var issue = exception.Issues.First();
+            Assert.Equal(OperationOutcomeConstants.IssueSeverity.Error, issue.Severity);
+            Assert.Equal(OperationOutcomeConstants.IssueType.Processing, issue.Code);
+            Assert.Equal(message, issue.Diagnostics);
+        }
+
+        private static List<OperationOutcomeIssue> GetOperationOutcomeIssues(string message)
+        {
+            var issues = new List<OperationOutcomeIssue>();
```
brandonpollett: nit: could use object initialization here

src/Microsoft.Health.Fhir.Core.UnitTests/Persistence/TransactionFailedExceptionTests.cs
```diff
@@ -0,0 +1,70 @@
...
+
+            var exception = new TransactionFailedException(message, statusCode, operationOutComeIssues);
+
+            Assert.NotNull(exception.Issues);
+            Assert.Equal(3, exception.Issues.Count);
+
+            var firtIssue = exception.Issues.First();
+            Assert.Equal(OperationOutcomeConstants.IssueSeverity.Error, firtIssue.Severity);
+            Assert.Equal(OperationOutcomeConstants.IssueType.Processing, firtIssue.Code);
+            Assert.Equal(message, firtIssue.Diagnostics);
```
brandonpollett: nit: could probably have a private method that does the asserts since we're doing the same asserts 4 times in the code.

src/Microsoft.Health.Fhir.Core/Features/Persistence/TransactionFailedException.cs
```diff
@@ -0,0 +1,34 @@
...
+            Debug.Assert(operationOutcomeIssues != null, "OperationOutcomeIssues should not be null");
+
+            ResponseStatusCode = httpStatusCode;
+
+            Issues.Add(new OperationOutcomeIssue(
+                    OperationOutcomeConstants.IssueSeverity.Error,
+                    OperationOutcomeConstants.IssueType.Processing,
+                    message));
+
+            operationOutcomeIssues.ForEach(x => Issues.Add(x));
```
brandonpollett: nit: could probably do something like `Issues.AddRange(operationOutcomeIssues);`

PR #711, opened 2019-10-29
add configuration option for database creation

nkiok:
## Description
Adds a configuration option whether to create or not the SQL database (if not already present).
  
This allows support of the following scenario:
When using a connection string that specifies a contained database user, the initial catalog must be specified

[To connect as a contained database user, the connection string must always contain a parameter for the user database so that the Database Engine knows which database is responsible for managing the authentication process.](https://docs.microsoft.com/en-us/sql/relational-databases/security/contained-database-users-making-your-database-portable?view=sql-server-ver15)

## Testing
Manually tested

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Configs/SqlServerDataStoreConfiguration.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Configs/SqlServerDataStoreConfiguration.cs
@@ -23,6 +23,6 @@ namespace Microsoft.Health.Fhir.SqlServer.Configs
         /// <summary>
         /// Allows the experimental schema initializer to attempt to create the database if not present.
         /// </summary>
-        public bool EnsureDatabase { get; set; }
+        public bool AllowDatabaseCreation { get; set; }
     }
 }
--- a/src/Microsoft.Health.Fhir.SqlServer/Configs/SqlServerDataStoreConfiguration.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Configs/SqlServerDataStoreConfiguration.cs
@@ -19,5 +19,10 @@ namespace Microsoft.Health.Fhir.SqlServer.Configs
         /// If set, this applies schema 1 which resets all the data in the database. This is temporary until the schema migration tool is complete.
         /// </summary>
         public bool DeleteAllDataOnStartup { get; set; }
+
+        /// <summary>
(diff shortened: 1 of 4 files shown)


src/Microsoft.Health.Fhir.SqlServer/Configs/SqlServerDataStoreConfiguration.cs
```diff
@@ -19,5 +19,10 @@ public class SqlServerDataStoreConfiguration
         /// If set, this applies schema 1 which resets all the data in the database. This is temporary until the schema migration tool is complete.
         /// </summary>
         public bool DeleteAllDataOnStartup { get; set; }
+
+        /// <summary>
+        /// Allows the experimental schema initializer to attempt to create the database if not present.
+        /// </summary>
+        public bool AllowDatabaseCreation { get; set; }
```
brendankowitz: To match current behavior I'm assuming we might want this turned on by default
```suggestion
        public bool AllowDatabaseCreation { get; set; } = true;
```

brendankowitz:
@johnstairs or @brandonpollett can you give this a quick look?

src/Microsoft.Health.Fhir.SqlServer/Configs/SqlServerDataStoreConfiguration.cs
nkiok: Might not actually, the way i see it deployments are mostly likely to occur in Azure, the default behaviour would be that the database is already present. From a security standpoint it is not a good idea for an app to have access to a database via a privileged account that has permissions to create databases.

brendankowitz: Makes sense. What about developing locally, are there any additional steps there?

nkiok: We can definitely make it "visible" or explicit for local development, add this config property to both .Web (R4, Stu3) projects `launchSettings.json` file. Out of the box the functionality remains the same as before

```
"SqlServer": {
      "commandName": "Project",
      "environmentVariables": {
        ..........
        "SqlServer:Initialize": "true",
        "SqlServer:AllowDatabaseCreation": "true",
```

PR #722, opened 2019-11-06
Returns error on invalid _pretty parameter input

rotodd:
## Description
- Returns 400 Bad Request when an invalid [_pretty parameter](https://www.hl7.org/fhir/http.html#parameters) is passed in (previously, the default value of false was assumed)

invalid-pretty

## Related issues
Addresses [#AB70807](https://microsofthealth.visualstudio.com/Health/_workitems/edit/70807).

## Testing
Adds a new e2e test.

### DIFF
--- a/src/Microsoft.Health.Fhir.Api/Resources.resx
+++ b/src/Microsoft.Health.Fhir.Api/Resources.resx
@@ -139,6 +139,10 @@
   <data name="InvalidLaunchContext" xml:space="preserve">
     <value>Invalid launch context parameters.</value>
   </data>
+  <data name="InvalidPrettyParameter" xml:space="preserve">
+    <value>The requested "_pretty" parameter is invalid.</value>
+    <comment>{NumberedPlaceHolder="_pretty"}</comment>
+  </data>
   <data name="InvalidRedirectUri" xml:space="preserve">
     <value>Provided value for redirect Uri: {0} is not a valid Uri.</value>
     <comment>{0} is a parameter passed in for redirection.</comment>
@@ -218,9 +222,6 @@
   <data name="UnsupportedParameterValue" xml:space="preserve">
     <value>The supplied value for "{0}" paramter is invalid.</value>
     <comment>{0} is parameter name</comment>
-  </data>
-  <data name="UnsupportedPrettyParameter" xml:space="preserve">
-    <value>The requested "_pretty" parameter is not supported.</value>
   </data>
   <data name="UnsupportedResourceType" xml:space="preserve">
     <value>Requested operation does not support resource type: {0}.</value>
--- a/src/Microsoft.Health.Fhir.Api/Resources.resx
+++ b/src/Microsoft.Health.Fhir.Api/Resources.resx
@@ -219,6 +219,9 @@
     <value>The supplied value for "{0}" paramter is invalid.</value>
     <comment>{0} is parameter name</comment>
   </data>
+  <data name="UnsupportedPrettyParameter" xml:space="preserve">
+    <value>The requested "_pretty" parameter is not supported.</value>
+  </data>
   <data name="UnsupportedResourceType" xml:space="preserve">
     <value>Requested operation does not support resource type: {0}.</value>
     <comment>{0} is the resource type.</comment>
--- a/src/Microsoft.Health.Fhir.Shared.Api/Features/ContentTypes/ContentTypeService.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Api/Features/ContentTypes/ContentTypeService.cs
@@ -86,7 +86,7 @@ namespace Microsoft.Health.Fhir.Api.Features.ContentTypes
 
             string prettyParameterValue = GetParameterValueFromQueryString(httpContext, KnownQueryParameterNames.Pretty);
 
-            if (prettyParameterValue != null && !bool.TryParse(prettyParameterValue, out bool isPretty))
+            if (prettyParameterValue != null && !bool.TryParse(prettyParameterValue, out _))
             {
                 throw new BadRequestException(Api.Resources.UnsupportedPrettyParameter);
             }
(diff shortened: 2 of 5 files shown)


src/Microsoft.Health.Fhir.Shared.Api/Features/ContentTypes/ContentTypeService.cs
```diff
@@ -139,5 +147,27 @@ private async Task<bool> IsFormatSupportedAsync(ResourceFormat resourceFormat)
                 }
             });
         }
+
+        private static string GetPrettyFromQueryString(HttpContext context)
```
brandonpollett: This looks pretty much like `GetFormatFromQueryString`. Is it possible to combine them to something like `GetValueFromQueryString(HttpContext context, string key)`?

src/Microsoft.Health.Fhir.Shared.Api/Features/ContentTypes/ContentTypeService.cs
```diff
@@ -82,6 +83,13 @@ public async Task CheckRequestedContentTypeAsync(HttpContext httpContext)
                     }
                 }
             }
+
+            string prettyParameterValue = GetPrettyFromQueryString(httpContext);
+
+            if (prettyParameterValue != null && !bool.TryParse(prettyParameterValue, out bool isPretty))
```
brandonpollett: You might be able to just call `bool.TryParse` without the null check. If memory serves, it does a null check internally. Also, we can probably change the out parameter to be `out _` since we don't actually care about the return value.

src/Microsoft.Health.Fhir.Shared.Api/Features/Formatters/HttpContextExtensions.cs
```diff
@@ -48,7 +48,7 @@ public static bool GetIsPretty(this HttpContext context)
             {
                 if (!bool.TryParse(query, out bool isPretty))
                 {
-                    // Assume no pretty formatting if parameter can't be parsed.
+                    // ContentTypeService validates the _pretty parameter. This is reached if other errors are encountered when parsing the query string.
```
brandonpollett: What other errors actually cause this?

rotodd: If both `format=blah` and `pretty=blah`, this will be reached when the operation outcome for the error is being formatted.

src/Microsoft.Health.Fhir.Shared.Api/Features/ContentTypes/ContentTypeService.cs
rotodd: Right, I originally didn't have this null check for that reason.  What I realized was here `prettyParameterValue=null` is used to indicate that `_pretty` wasn't included in the query string (and we wouldn't want this to fail in that case).

And cool thanks, didn't know you could discard out variables. Fixed in 19651a.

src/Microsoft.Health.Fhir.Shared.Api/Features/ContentTypes/ContentTypeService.cs
rotodd: Good idea. Implemented in 86f3822.

src/Microsoft.Health.Fhir.Api/Resources.resx
```diff
@@ -219,6 +219,9 @@
     <value>The supplied value for "{0}" paramter is invalid.</value>
     <comment>{0} is parameter name</comment>
   </data>
+  <data name="UnsupportedPrettyParameter" xml:space="preserve">
+    <value>The requested "_pretty" parameter is not supported.</value>
+  </data>
```
brandonpollett: Probably need to add a comment of `{NumberedPlaceHolder="_pretty"}` so that it wouldn't be translated.

src/Microsoft.Health.Fhir.Shared.Api/Features/ContentTypes/ContentTypeService.cs
```diff
@@ -82,23 +83,30 @@ public async Task CheckRequestedContentTypeAsync(HttpContext httpContext)
...
+            }
         }
 
-        private static string GetFormatFromQueryString(HttpContext context)
+        private static string GetParameterValueFromQueryString(HttpContext context, string parameterName)
         {
             EnsureArg.IsNotNull(context, nameof(context));
 
-            // If executing in a rethrown error context, ensure we carry the specified format
+            // If executing in a rethrown error context, ensure we carry the specified formatting.
```
brandonpollett: nit: should `formatting` be something like `query string value`?

src/Microsoft.Health.Fhir.Shared.Api/Features/ContentTypes/ContentTypeService.cs
```diff
@@ -82,23 +83,30 @@ public async Task CheckRequestedContentTypeAsync(HttpContext httpContext)
                     }
                 }
             }
+
+            string prettyParameterValue = GetParameterValueFromQueryString(httpContext, KnownQueryParameterNames.Pretty);
+
+            if (prettyParameterValue != null && !bool.TryParse(prettyParameterValue, out _))
+            {
+                throw new BadRequestException(Api.Resources.UnsupportedPrettyParameter);
```
namadabu: nit: can we use `UnsupportedParameterValue` instead of adding a new one?

src/Microsoft.Health.Fhir.Api/Resources.resx
rotodd: Thanks! Added in fccf656.

src/Microsoft.Health.Fhir.Shared.Api/Features/ContentTypes/ContentTypeService.cs
rotodd: Yes that definitely makes more sense. Fixed in dbcf721.

src/Microsoft.Health.Fhir.Shared.Api/Features/ContentTypes/ContentTypeService.cs
rotodd: I ended up changing this resource to better fit the error code that I'm returning in fccf656. Otherwise, I think you're right that it would have been nice to have a single message for both _pretty and _format.

PR #725, opened 2019-11-06
Adds an implementation to fetch script

rbans96:
## Description
GET call on /_schema/versions/{id}/script route returns script

## Related issues
Addresses [AB#68808](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/68808)

## Testing
Added unit tests and e2e tests

Success Screenshot
image

POST verb to fail
image

Content-Disposition Header
image


src/Microsoft.Health.Fhir.SqlServer.Api/Controllers/SchemaController.cs
```diff
@@ -72,7 +73,7 @@ public ActionResult SqlScript(int id)
         {
             _logger.LogInformation($"Attempting to get script for schema version: {id}");
 
-            throw new NotImplementedException(Resources.ScriptNotImplemented);
+            return Content(SchemaUpgradeRunner.GetMigrationScript(id));
```
brandonpollett: We need to specify the content type here of `application/sql`. Additionally, we probably want to change the return type `File` and also specify a filename which will then set the `Content-Disposition` header.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/SchemaUpgradeRunner.cs
```diff
@@ -50,7 +50,7 @@ public void ApplySchema(int version)
             _logger.LogInformation("Completed applying schema {version}", version);
         }
 
-        private static string GetMigrationScript(int version)
+        public static string GetMigrationScript(int version)
```
brandonpollett: Since this is used by more than just the `SchemaUpgradeRunner` now, this might make sense in a different file.

src/Microsoft.Health.Fhir.SqlServer.Api/Features/Filters/NullArgumentExceptionFilterAttribute.cs
```diff
@@ -0,0 +1,26 @@
...
+
+using System;
+using System.Net;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Mvc.Filters;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Health.Fhir.SqlServer.Api.Features.Filters
+{
+    public class NullArgumentExceptionFilterAttribute : ActionFilterAttribute
```
brandonpollett: This seems like an overly broad approach to capture attempting getting a script that doesn't exist. This will likely capture anywhere in the stack that an `ArgumentNullException` occurs. We should probably update the code where it's failing to do a null check and throw a specific error type to handle. 

We can then implement a more general filter similar to `OperationOutcomeExceptionFilterAttribute` for just the schema controller. That would allow us to combine this logic with the `NotImplementedExceptionFilter` and handle the output in a standard manner.

src/Microsoft.Health.Fhir.SqlServer.Api/Controllers/SchemaController.cs
rbans96: Addressed it 👍

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/SchemaUpgradeRunner.cs
rbans96: addressed it 👍

src/Microsoft.Health.Fhir.SqlServer.Api/Features/Filters/NullArgumentExceptionFilterAttribute.cs
rbans96: addressed it 👍

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Sql/SchemaTests.cs
```diff
@@ -90,6 +88,41 @@ public async Task WhenRequestingScript_GivenNonIntegerVersion_TheServerShouldRet
...
+        }
+
+        [Fact]
+        public async Task WhenRequestingScript_GivenDeleteMethod_TheServerShouldReturnNotFound()
+        {
+            await SendAndVerifyStatusCode(HttpMethod.Delete, "_schema/versions/1/script", HttpStatusCode.NotFound);
+        }
+
+        [Fact]
+        public async Task WhenRequestingScript_GivenSchemaIdFound_TheServerShouldReturnScript()
```
brandonpollett: We should test failure cases too.

src/Microsoft.Health.Fhir.SqlServer.Api/Controllers/SchemaController.cs
```diff
@@ -68,11 +69,11 @@ public ActionResult CurrentVersion()
...
         [AllowAnonymous]
         [Route(KnownRoutes.Script, Name = RouteNames.Script)]
-        public ActionResult SqlScript(int id)
+        public FileContentResult SqlScript(int id)
         {
             _logger.LogInformation($"Attempting to get script for schema version: {id}");
-
-            throw new NotImplementedException(Resources.ScriptNotImplemented);
+            string fileName = $"{id}.sql";
+            return File(Encoding.ASCII.GetBytes(ScriptProvider.GetMigrationScript(id)), "application/json", fileName);
```
brandonpollett: We might want to look into having a version of `GetMigrationScript` that returns something that doesn't require the conversion to text and then to bytes.

src/Microsoft.Health.Fhir.SqlServer/Resources.resx
```diff
@@ -120,4 +120,7 @@
   <data name="InvalidContinuationToken" xml:space="preserve">
     <value>The provided continuation token is not valid.</value>
   </data>
+  <data name="ScriptNotFound" xml:space="preserve">
+    <value>The provided version is unknown</value>
```
brandonpollett: nit: missing period.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Sql/SchemaTests.cs
rbans96: Failure cases are like - schemaId not found and Verbs except GET.
Yes I  missed to add testcase for schemaId not found- work in progress. While testcases are already added for the failure verbs(POST, PUT, DELETE). 
Please let me know if I am missing anything here. Thanks

src/Microsoft.Health.Fhir.SqlServer.Api/Controllers/SchemaController.cs
rbans96: Addressed it.

src/Microsoft.Health.Fhir.SqlServer.Api/Features/Filters/HttpExceptionFilterAttribute.cs
```diff
@@ -0,0 +1,50 @@
...
+
+namespace Microsoft.Health.Fhir.SqlServer.Api.Features.Filters
+{
+    internal class HttpExceptionFilterAttribute : ActionFilterAttribute
+    {
+        public override void OnActionExecuted(ActionExecutedContext context)
+        {
+        EnsureArg.IsNotNull(context, nameof(context));
+
+        if (context?.Exception == null)
```
brandonpollett: nit: since we're null checking above, the `?` operator should never be reached.

src/Microsoft.Health.Fhir.SqlServer.Api/Features/Filters/HttpExceptionFilterAttribute.cs
```diff
@@ -0,0 +1,50 @@
...
+using Microsoft.AspNetCore.Mvc.Filters;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Health.Fhir.SqlServer.Api.Features.Filters
+{
+    internal class HttpExceptionFilterAttribute : ActionFilterAttribute
+    {
+        public override void OnActionExecuted(ActionExecutedContext context)
+        {
+        EnsureArg.IsNotNull(context, nameof(context));
```
brandonpollett: nit: indentation is off.

src/Microsoft.Health.Fhir.SqlServer.Api/Features/Filters/HttpExceptionFilterAttribute.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        if (context?.Exception == null)
+        {
+            return;
+        }
+
+        if (!context.ExceptionHandled)
+            {
+                var resultJson = new JObject();
+                if (context.Exception != null)
```
brandonpollett: This should always be true since the code is already checking that `context.Exception` equals null above and returning.

src/Microsoft.Health.Fhir.SqlServer.Api/Features/Filters/HttpExceptionFilterAttribute.cs
rbans96: 👍 My bad

Issue #411, opened 2019-04-05
PartialDateTime should not allow Hour and Minute without Seconds


jackliums:
**Describe the bug**

According to the spec (http://hl7.org/fhir/STU3/datatypes.html#datetime):

> A date, date-time or partial date (e.g. just year or year + month) as used in human communication. If hours and minutes are specified, a time zone SHALL be populated. Seconds must be provided due to schema type constraints but may be zero-filled and may be ignored. Dates SHALL be valid dates. The time "24:00" is not allowed.

Current implementation allows hour and minute without seconds so that should be changed.

johnstairs:
@LTA-Thinking this part of the spec does not seem to be enforced:
> If hours and minutes are specified, a time zone SHALL be populated

LTA-Thinking:
Strange. We seem to have code to handle that senario: https://github.com/microsoft/fhir-server/blob/13a410bea18435288e7725b5be8fd324c0dac310/src/Microsoft.Health.Fhir.Core/Models/PartialDateTime.cs#L110. I'll look into it.

rotodd:
As discussed with @jackliums, the current implementation where seconds are populated if hours and minutes are specified is fine.

The [spec](http://hl7.org/fhir/search.html#date) does not indicate that seconds need to be provided by the user when specifying a `DateTime` search parameter.

Issue #25, opened 2018-09-18
Problems with invalid date search


johnstairs:
Add this resource to a FHIR server:
https://www.hl7.org/fhir/diagnosticreport-example.json
note the field `"issued": "2011-03-04T11:45:33+11:00",`

Perform the search:
`/DiagnosticReport?issued=2011-03-04T10:45:33`
It matches. Cool.

Now try this one:
`/DiagnosticReport?issued=0-03-04T10:45:33` 
You get a 500.

Now this one: 
`/DiagnosticReport?issued=201111111-03-04T10:45:33`
It actually matches the document!
So does this one:
`/DiagnosticReport?issued=2011jasdlka-03-04T10:45:33`

PR #726, opened 2019-11-06
Require seconds if hours and minutes entered

LTA-Thinking:
## Description
Add a requirement that seconds be specified if the hours and minutes in a timestamp are given. This is a replacement of #585.
## Related issues
Addresses #411, #25.

## Testing
Unit tests have been updated to pass seconds and a new one was added to test that seconds are required.


src/Microsoft.Health.Fhir.Core/Models/PartialDateTime.cs
```diff
@@ -292,7 +301,7 @@ public static PartialDateTime Parse(string s)
             catch (Exception ex) when (ex is ArgumentException)
             {
                 // The input value was parsed correctly but one of the value provided were out of range.
-                throw new FormatException("Input string was not in a correct format. At least one portion of a date was invalid or out of range.", ex);
+                throw new RequestNotValidException("Input string was not in a correct format. At least one portion of a date was invalid or out of range: " + ex.Message);
```
johnstairs: Include ex as the inner exception

src/Microsoft.Health.Fhir.Core/Models/PartialDateTime.cs
```diff
@@ -292,7 +301,7 @@ public static PartialDateTime Parse(string s)
             catch (Exception ex) when (ex is ArgumentException)
```
johnstairs: Just `catch (ArgumentException ex)`

src/Microsoft.Health.Fhir.Core/Models/PartialDateTime.cs
LTA-Thinking: This type of exception doesn't have a constructor that takes an inner exception.

johnstairs:
While you're at it,
``` http
https://localhost:44348/Observation?_lastUpdated=1
```

generates a 500

src/Microsoft.Health.Fhir.Core/Models/PartialDateTime.cs
johnstairs: If it do 
``` HTTP
/Observation?_lastUpdated=2013-04-03T15:30
```

This error is returned from the API call:
> Input string was not in a correct format. At least one portion of a date was invalid or out of range: The 'second' portion of a date must be specified if 'hour' and 'minute' are specified.\r\nParameter name: second

The "Parameter name: second" portion does not make sense from the user's context. Ideally, the error would provide the search parameter name, but we don't do that in other cases (but feel free to fix that)

LTA-Thinking: Refactored to remove outer exception.

src/Microsoft.Health.Fhir.Core/Models/PartialDateTime.cs
LTA-Thinking: Refactored out the catch

src/Microsoft.Health.Fhir.Core/Models/PartialDateTime.cs
LTA-Thinking: Fixed exception message

LTA-Thinking:
When the code is parsing times the regex it is using won't match to invalid times. This means that inputs like 2018-02-01T36 or 2018-02-01T12:97 will succeed with the hours and minutes being ignored, while an input like 2018-02-01T12:42:75 will give an exception that seconds is unspecified. Working on remedying this problem.

LTA-Thinking:
Bringing in Issue #25

brendankowitz:
Closing as this PR is covered in #933

Issue #729, opened 2019-11-07
Refactor _total parameter implementation so that SearchOptions is returned to its original state


rotodd:
**User story**

We are mutating SearchOptions in both the SQL and Cosmos search service classes. Return it to its original state in a try-finally block. This will allow the properties of the SearchOptions to remain in the state set by user input.

**Acceptance criteria**

The SearchOptions object is not modified after it is used in the search service classes.

CaitlinV39:
@rotodd Does this item have additional work pending?

rotodd:
> @rotodd Does this item have additional work pending?

@CaitlinV39 nope, closing it now. Thanks 😊

PR #731, opened 2019-11-07
Refactors _total parameter implementation to reset search options to its original state

rotodd:
## Description
- Returns the SearchOptions object to its original state if it is modified for a query where `_total=accurate`

## Related issues
Addresses #729 and [#AB71063](https://microsofthealth.visualstudio.com/Health/_workitems/edit/71063).

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
@@ -88,16 +88,16 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Search
 
                             // Perform a second read to get the count.
                             var countOnlySearchResult = await SearchImpl(searchOptions, false, connection, cancellationToken, transaction);
-                            
+
                             searchResult.TotalCount = countOnlySearchResult.TotalCount;
+
+                            transaction.Commit();
                         }
                         finally
                         {
                             // Reset search options to its original state.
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -80,17 +80,24 @@ protected override async Task<SearchResult> SearchInternalAsync(SearchOptions se
                     // Begin a transaction so we can perform two atomic reads.
                     using (var transaction = connection.BeginTransaction(IsolationLevel.ReadCommitted))
                     {
-                        searchResult = await SearchImpl(searchOptions, false, connection, cancellationToken, transaction);
+                        try
+                        {
+                            searchResult = await SearchImpl(searchOptions, false, connection, cancellationToken, transaction);
```
johnstairs: This should be outside of the try block. If it throws, you don't want to revert.

rotodd: If this throws, searchOptions.CountOnly will already be set to false (due to the if statement on line 78). 

I agree that this doesn't necessarily need to be in the try block and that it's a bit redundant to reset a false value to false a second time. Given this is already merged, how would you recommend moving forward?

johnstairs: Not a problem, I would just address it in the next change.

PR #732, opened 2019-11-08
Adds caching for health check endpoint

brendankowitz:
## Description
Adds a 1 second cache and semaphore for the health check endpoint. This ensures only 1 thread will do the work of calling into the health checks at most once per second.

## Related issues
Addresses [AB#70779](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/70779)

## Testing
* Unit tests added
* Existing E2E tests


src/Microsoft.Health.Fhir.Api/Features/HealthCheck/HealthCheckCachingMiddleware.cs
```diff
@@ -0,0 +1,108 @@
...
+using Microsoft.Extensions.Logging;
+using Microsoft.Health.Fhir.Core;
+
+namespace Microsoft.Health.Fhir.Api.Features.HealthCheck
+{
+    /// <summary>
+    /// This is not the same as HealthCheckOptions.AllowCachingResponses.
+    /// This Middleware provides short term caching to protect against DoS attacks on the health check endpoint.
+    /// </summary>
+    internal sealed class HealthCheckCachingMiddleware : IDisposable
```
jackliums: Is there any reason we are not caching the result in `CosmosHealthCheck` and `SqlServerHealthCheck`?

brendankowitz: I also had an implementation for this approach. Middleware seemed like a better catch-all for all health checks as the goal is to add end point protection.

For consideration this was my implementation for caching the check itself:
```
using System;
using System.Threading;
using System.Threading.Tasks;
using EnsureThat;
using Microsoft.Extensions.Diagnostics.HealthChecks;

namespace Microsoft.Health.Fhir.Core
{
    public class CachedHealthCheck<T> : IHealthCheck
        where T : class, IHealthCheck
    {
        private readonly T _healthCheck;
        private DateTimeOffset _lastChecked;
        private HealthCheckResult _lastResult;
        private readonly SemaphoreSlim _semaphore = new SemaphoreSlim(1, 1);

        public CachedHealthCheck(T healthCheck)
        {
            EnsureArg.IsNotNull(healthCheck, nameof(healthCheck));

            _healthCheck = healthCheck;
        }

        public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken)
        {
            if (_lastChecked >= Clock.UtcNow.AddSeconds(-1))
            {
                return _lastResult;
            }

            await _semaphore.WaitAsync(cancellationToken);
            try
            {
                if (_lastChecked >= Clock.UtcNow.AddSeconds(-1))
                {
                    return _lastResult;
                }

                _lastResult = await _healthCheck.CheckHealthAsync(context, cancellationToken);
                _lastChecked = Clock.UtcNow;

                return _lastResult;
            }
            finally
            {
                _semaphore.Release();
            }
        }
    }
}
```

johnstairs: If we are converned about CPU cycles serializing the response, the /metadata endpoint will be a lot heavier. Perhaps output caching should be extended to it (with a longer ttl)

brendankowitz: I was thinking of addressing metadata in as part of this https://github.com/microsoft/fhir-server/compare/personal/bkowitz/adds-aspnetcore-responsecompression?expand=1

src/Microsoft.Health.Fhir.Api/Features/HealthCheck/HealthCheckCachingMiddleware.cs
```diff
@@ -0,0 +1,108 @@
...
+        private async Task<bool> WriteCachedResult(Stream originalBodyStream, bool forceWrite = false)
+        {
+            if (forceWrite || (_lastResultBuffer != null && _lastCheckTime >= Clock.UtcNow.Add(-_cacheDuration)))
+            {
+                if (!forceWrite)
+                {
+                    _logger.LogDebug("Writing cached healthcheck from {0:o}", _lastCheckTime);
+                }
+
+                await originalBodyStream.WriteAsync(_lastResultBuffer, 0, _lastResultBuffer.Length);
```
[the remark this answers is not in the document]
brendankowitz: Thanks, updated the `if` check

src/Microsoft.Health.Fhir.Api/Features/HealthCheck/HealthCheckCachingMiddleware.cs
brendankowitz: @jackliums @johnstairs in your opinion(s) do you feed its cleaner to go with the above `CachedHealthCheck<T>` approach?

brendankowitz: Updated

src/Microsoft.Health.Fhir.Shared.Api/Modules/HealthChecks/CachedHealthCheck.cs
```diff
@@ -0,0 +1,70 @@
...
+namespace Microsoft.Health.Fhir.Api.Modules.HealthChecks
+{
+    internal sealed class CachedHealthCheck : IHealthCheck, IDisposable
+    {
+        private readonly IServiceProvider _provider;
+        private readonly Func<IServiceProvider, IHealthCheck> _healthCheck;
+        private DateTimeOffset _lastChecked;
+        private HealthCheckResult _lastResult;
+        private readonly SemaphoreSlim _semaphore = new SemaphoreSlim(1, 1);
+        private readonly TimeSpan _cacheTime = TimeSpan.FromSeconds(1);
```
johnstairs: can be static

src/Microsoft.Health.Fhir.Shared.Api/Modules/HealthChecks/CachedHealthCheck.cs
```diff
@@ -0,0 +1,70 @@
...
+                {
+                    return _lastResult;
+                }
+
+                using (var scope = _provider.CreateScope())
+                {
+                    var check = _healthCheck.Invoke(scope.ServiceProvider);
+
+                    _lastResult = await check.CheckHealthAsync(context, cancellationToken);
+                    _lastChecked = Clock.UtcNow;
```
johnstairs: If this throws, no caching? Did you consider storing the ExceptionDispatchInfo?

src/Microsoft.Health.Fhir.Shared.Api/Modules/HealthChecks/HealthCheckConfiguration.cs
```diff
@@ -0,0 +1,49 @@
...
+        private readonly IServiceProvider _serviceProvider;
+
+        public HealthCheckConfiguration(IServiceProvider serviceProvider)
+        {
+            EnsureArg.IsNotNull(serviceProvider, nameof(serviceProvider));
+
+            _serviceProvider = serviceProvider;
+        }
+
+        public void PostConfigure(string name, HealthCheckServiceOptions options)
```
jackliums: Is this method called before the ServiceProvider is materialized/build? I guess I don't know enough about the order of execution for this pipeline. Would there be a chance where the IHealthCheck instance is already created by the time it gets here?

Also, we could do this in the registration too right? Like at the very end of the registration, we can have a loop that loops through all ServiceDescriptor with service type of IHealthCheck and just overwrites that? Is there any advantage/disadvantage?

src/Microsoft.Health.Fhir.Shared.Api/Modules/HealthChecks/HealthCheckConfiguration.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            var list = options.Registrations.ToArray();
```
jackliums: nit: coding style says only use var if right hand-size is clear (here and below in foreach).

src/Microsoft.Health.Fhir.Shared.Api/Modules/HealthChecks/HealthCheckConfiguration.cs
brendankowitz: After reviewing the framework code, this _seems_ like the best hook for what we're trying to do. This is called just after the container is built, but before the app starts, so it is safe in terms of ordering of registrations. I think this might be better than wrapping the registration in the container as it doesn't interfere with any of the HealthCheck extensions or HealthCheck lifetimes/scopes, these can all still be setup as desired.

PR #740, opened 2019-11-12
Transaction-Bundle validation

poadhika:
## Description
This PR adds the functionality that validates the following:
 Within a transaction the same resource/id cannot be present more than once within DELETEs, POSTs, PUTs, or PATCHes. If this occurs, the transaction should fail.

## Related issues
Addresses  https://microsofthealth.visualstudio.com/Health/_workitems/edit/70639

## Testing
Unit Tests has been added.


src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionValidator.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            string fullUrl = component.FullUrl;
+
+            if (fullUrl != null)
+            {
+                if (fullUrl.StartsWith("urn", StringComparison.OrdinalIgnoreCase) || component.Request.Method == HTTPVerb.POST)
+                {
+                    return fullUrl;
+                }
+            }
```
brandonpollett: This may only come into play for `POST`s. For those, it may be sufficient to just use the `fullUrl` regardless of if it starts with urn or not.

The rest of the time, the `component.Request.Url` will most likely be the accurate portion to use.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionValidator.cs
```diff
@@ -0,0 +1,45 @@
...
+                if (resourceIdList.Contains(resourceId))
+                {
+                    throw new RequestNotValidException(string.Format(Api.Resources.ResourcesMustBeUnique, resourceId));
+                }
+
+                resourceIdList.Add(resourceId);
+            }
+        }
+
+        private static string GetResourceUrl(EntryComponent component)
```
brandonpollett: There's also some actions that aren't GETs that realistically we don't have to worry about. Operations via `POST` is an example. https://www.hl7.org/fhir/http.html#summary

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionValidator.cs
```diff
@@ -0,0 +1,45 @@
...
+
+            if (fullUrl != null)
+            {
+                if (fullUrl.StartsWith("urn", StringComparison.OrdinalIgnoreCase) || component.Request.Method == HTTPVerb.POST)
+                {
+                    return fullUrl;
+                }
+            }
+
+            return component.Request.Url;
```
brandonpollett: I think according to the spec, we're also supposed to check resolved IDs for conditional creates and deletes.

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Resources/Bundle/TransactionExceptionHandlerTests.cs
```diff
@@ -0,0 +1,65 @@
...
+
+            for (int i = 0; i < operationOutcomeIssues.Count; i++)
+            {
+                Assert.Equal(operationOutcomeIssues[i].Severity.ToString(), parsedOperationOutcomeIssueList[i].Severity);
+                Assert.Equal(operationOutcomeIssues[i].Code.ToString(), parsedOperationOutcomeIssueList[i].Code);
+                Assert.Equal(operationOutcomeIssues[i].Diagnostics, parsedOperationOutcomeIssueList[i].Diagnostics);
+            }
+        }
+
+        private static OperationOutcome GetOperationOutcomeList()
```
brandonpollett: nit: Maybe update the name since this isn't returning a list

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Resources/Bundle/TransactionExceptionHandlerTests.cs
```diff
@@ -0,0 +1,65 @@
...
+using Microsoft.Health.Fhir.Api.Features.Resources.Bundle;
+using Microsoft.Health.Fhir.Core.Features.Persistence;
+using Xunit;
+
+namespace Microsoft.Health.Fhir.Api.UnitTests.Features.Resources.Bundle
+{
+    public class TransactionExceptionHandlerTests
+    {
+        [Fact]
+        public void GivenAnOperationOutCome_WhenExecuted_ThenACorrectExceptionIsThrown()
```
brandonpollett: nit: `Outcome`, not `OutCome`

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Resources/Bundle/TransactionValidatorTests.cs
```diff
@@ -0,0 +1,56 @@
...
+            {
+                var resourceUrlList = new HashSet<string>();
+                foreach (Hl7.Fhir.Model.Bundle.EntryComponent entry in requestBundle.ToPoco<Hl7.Fhir.Model.Bundle>().Entry)
+                {
+                    TransactionValidator.ValidateTransaction(resourceUrlList, entry);
+                }
+            }
+            catch (Exception ex)
+            {
+                Assert.True(ex is RequestNotValidException);
```
brandonpollett: Might want to assert the correct url is in the message too.

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Resources/Bundle/TransactionValidatorTests.cs
```diff
@@ -0,0 +1,56 @@
...
+            var requestBundle = Samples.GetJsonSample("Bundle-TransactionWithMultipleEntriesModifyingSameResource");
+
+            ValidateIfBundleEntryIsUnique(requestBundle);
+        }
+
+        private static void ValidateIfBundleEntryIsUnique(Core.Models.ResourceElement requestBundle)
+        {
+            try
+            {
+                var resourceUrlList = new HashSet<string>();
```
brandonpollett: Could move this to the constructor and then just do `Assert.Throws` in the failing cases.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionValidator.cs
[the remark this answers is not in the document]
poadhika: My understanding is fullURL will be present in POST. UUID will be used as a fullURL since the resource is not assigned a created yet and will not have the absoulute url/identity to represent it. According to spec "For transactions and batches, entries MAY not have fullURLs when the entry.request.method = POST, and the resource has no identity. Note that even in this case, there may still be a fullURL in a transaction on a POST so that relationships between resources can be represented"

poadhika: Per our discussion, I will address this scenario after the dependent stories are complete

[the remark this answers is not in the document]
poadhika: It will now return fullURl only for the valid post requests. Valid here means if the post is not for operations and also it will ignore if resourcetype is bundle.

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Resources/Bundle/TransactionValidatorTests.cs
poadhika: I keep it as is but combined two tests  into one. Let me know if I need to change it.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionValidator.cs
```diff
@@ -0,0 +1,54 @@
...
+            }
+        }
+
+        private static bool ValidateBundleEntry(EntryComponent entry)
+        {
+            string requestUrl = entry.Request.Url;
+
+            return !((entry.Request.Method == HTTPVerb.GET)
+                || requestUrl.Contains("$", StringComparison.InvariantCulture)
+                || requestUrl.Contains("_search", StringComparison.InvariantCulture)
```
brandonpollett: `_search` should only appear in `POST` so we don't need to check it all the time

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionValidator.cs
```diff
@@ -0,0 +1,54 @@
...
+                    resourceIdList.Add(resourceId);
+                }
+            }
+        }
+
+        private static bool ValidateBundleEntry(EntryComponent entry)
+        {
+            string requestUrl = entry.Request.Url;
+
+            return !((entry.Request.Method == HTTPVerb.GET)
```
brandonpollett: We should comment for why these rules are being enforced.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionValidator.cs
```diff
@@ -0,0 +1,54 @@
...
+{
+    public static class TransactionValidator
+    {
+        public static void ValidateTransaction(HashSet<string> resourceIdList, EntryComponent entry)
+        {
+            if (ValidateBundleEntry(entry))
+            {
+                string resourceId = GetResourceUrl(entry);
+
+                if (resourceId != null)
```
brandonpollett: What would the behavior be of a `FullUrl` of `""` on a `POST` be?

poadhika: It should skip that case as well. I have now updated the check.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -127,37 +126,22 @@ private async Task<BundleResponse> ExecuteTransactionForAllRequests(Hl7.Fhir.Mod
             return new BundleResponse(responseBundle.ToResourceElement());
         }
 
-        private static void ThrowTransactionException(HttpContext httpContext, OperationOutcome operationOutcome)
+        private async Task FillRequestLists(Hl7.Fhir.Model.Bundle.BundleType? bundleType, List<Hl7.Fhir.Model.Bundle.EntryComponent> bundleEntries)
         {
-            var operationOutcomeIssues = GetOperationOutcomeIssues(operationOutcome.Issue);
+            var resourceUrlList = new HashSet<string>();
```
jackliums: Should always specify StringComparer when using string as key.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionValidator.cs
```diff
@@ -0,0 +1,54 @@
...
+                    resourceIdList.Add(resourceId);
+                }
+            }
+        }
+
+        private static bool ValidateBundleEntry(EntryComponent entry)
+        {
+            string requestUrl = entry.Request.Url;
+
+            // Check for duplicate resources within a bundle entry is skipped if the entry is bundle or if the request within a entry is not modifying the resource.
```
jackliums: Can you explain this a little bit. Are you saying that if an entry inside the bundle is another bundle, we won't validate across these two bundles?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionValidator.cs
```diff
@@ -0,0 +1,54 @@
...
+            return !(entry.Resource.ResourceType == Hl7.Fhir.Model.ResourceType.Bundle
+                || entry.Request.Method == HTTPVerb.GET
+                || (entry.Request.Method == HTTPVerb.POST && requestUrl.Contains("_search", StringComparison.InvariantCulture))
```
jackliums: Should we disallow using POST for search in batch/transaction context? I feel like it doesn't make sense because you cannot specify the body of the request in the bundle?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionValidator.cs
jackliums: I think the name might be a little misleading. It's not really validating the bundle entry, but rather it's check to see if validation is needed?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionValidator.cs
```diff
@@ -0,0 +1,54 @@
...
+                || requestUrl.Contains("$", StringComparison.InvariantCulture));
+        }
+
+        private static string GetResourceUrl(EntryComponent component)
+        {
+            if (component.Request.Method == HTTPVerb.POST)
+            {
+                return component.FullUrl;
```
jackliums: What happens if the request is conditional create/update? In that case, the request URL contains search parameters; would that be addressed as part of work mentioned below?

jackliums: Also, what do we do in the case the URL is specified absolute/relative that points to the same resource?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionValidator.cs
jackliums: I am not sure how we will support this through bundle. To search using POST and _search endpoint, one must specify the Content-Type header to be application/x-www-form-urlencded and the search criteria needs to be specified through body. I don't think neither of this is possible through bundle entry (correct me if I am wrong).

Should we add validation just to not allow this?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -127,37 +126,22 @@ private async Task<BundleResponse> ExecuteTransactionForAllRequests(Hl7.Fhir.Mod
...
-        private async Task FillRequestLists(List<Hl7.Fhir.Model.Bundle.EntryComponent> bundleEntries)
-        {
             foreach (Hl7.Fhir.Model.Bundle.EntryComponent entry in bundleEntries)
             {
                 if (entry.Request.Method == null)
                 {
                     continue;
                 }
 
+                if (bundleType == Hl7.Fhir.Model.Bundle.BundleType.Transaction)
```
jackliums: [Spec](http://hl7.org/fhir/http.html#brules) says:

> For a batch, there SHALL be no interdependencies between the different entries in the Bundle that cause change on the server. The success or failure of one change SHOULD not alter the success or failure or resulting content of another change.

But seems like we are implemeting this for transaciton?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionExceptionHandler.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.Net;
+using Hl7.Fhir.Model;
+using Microsoft.Health.Fhir.Core.Features.Persistence;
+using Microsoft.Health.Fhir.Core.Models;
+
+namespace Microsoft.Health.Fhir.Api.Features.Resources.Bundle
+{
+    public static class TransactionExceptionHandler
+    {
+        public static void ThrowTransactionException(string errorMessage, HttpStatusCode statusCode, OperationOutcome operationOutcome)
```
jackliums: Seems like only being used inside BundleHandler. Any specific reason that this was move to a public static class?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionValidator.cs
brandonpollett: That's a good point. I think throwing in that case makes the most sense.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
brandonpollett: In the [transaction section](http://hl7.org/fhir/http.html#trules) it states:

> If any resource identities (including resolved identities from conditional update/delete) overlap in steps 1-3, then the transaction SHALL fail.

Steps 1-3 refers to 

> 1. Process any DELETE interactions
> 2. Process any POST interactions
> 3. Process any PUT or PATCH interactions

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionValidator.cs
poadhika: > Can you explain this a little bit. Are you saying that if an entry inside the bundle is another bundle, we won't validate across these two bundles?

Yes. For now this is the case.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionValidator.cs
poadhika: > What happens if the request is conditional create/update? In that case, the request URL contains search parameters; would that be addressed as part of work mentioned below?

I am addressing the conditional create/update scenario as a part of the next story I am working on as they are similar.

poadhika: > Also, what do we do in the case the URL is specified absolute/relative that points to the same resource?

We are returning FullURL only in POST. We are using the request.url in other cases( i.e. for PUT and DELETE). In case for POST, fullURl is mostly uuid as the resource willnot have logicalId yet. POST and PUT can be creating the same resource in case of conditional create/update when the search parameter is the same identifier. I am addressing that as a part of my story that I am currently working on. Does this answer your question?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionExceptionHandler.cs
poadhika: This was moved in order to make unit testing easier.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionValidator.cs
jackliums: > We are returning FullURL only in POST.

This method is processing the incoming request, crafted by the caller, right?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
jackliums: Okay so for this PR, it's only addressing the transaction part and there is another work item to track for batch?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -127,37 +126,22 @@ private async Task<BundleResponse> ExecuteTransactionForAllRequests(Hl7.Fhir.Mod
...
+                {
+                    TransactionValidator.ValidateTransaction(resourceUrlList, entry);
```
jackliums: What is the advantage of having the caller keeping track of list of resource URLs instead of just pass the bundle in and let the function validates it upfront? For example, can you just pass the bundle to this ValidateTransaction (by the name, i think it needs to be renamed. It's not really validating the transaction.), and just have all of the logic encapsulated in that method?

At this point, the server hasn't replaced all of the ID yet so it seems like the detection is only applicable to resource IDs supplied by the user, in which case, it could just do the validation upfront and fail the request right away instead of allocating all these objects below that's just goint to be deallocated anyways?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
poadhika: I checked the behavior on other fhir test servers but didnot get the answer i am looking for. For now, i will create a separate story for batch.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
poadhika: As discussed, will move validation as a separate module. For conditionacreate/update, will resolve the id's before executing the transaction.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionValidator.cs
```diff
@@ -0,0 +1,66 @@
...
+using System;
+using System.Collections.Generic;
+using Microsoft.Health.Fhir.Core.Exceptions;
+using static Hl7.Fhir.Model.Bundle;
+
+namespace Microsoft.Health.Fhir.Api.Features.Resources.Bundle
+{
+    public static class TransactionValidator
+    {
+        // Validates if transaction bundle contains multiple entries that are modifying the same resource.
```
brandonpollett: nit: convert this to the standard method documentation format

src/Microsoft.Health.Fhir.Api/Resources.resx
```diff
@@ -253,4 +257,7 @@
     <value>STU3 API &lt;small&gt;preview&lt;/small&gt;</value>
     <comment>{NumberedPlaceHolder=”STU3”}</comment>
   </data>
+  <data name="InvalidBundleEntry" xml:space="preserve">
+    <value>Requested operation does not support _search interation using POST.</value>
```
brandonpollett: nit: I think `_search` and `POST` might need to be numbered placeholders.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionValidator.cs
```diff
@@ -0,0 +1,66 @@
...
+            HTTPVerb? requestMethod = entry.Request.Method;
+
+            // Search operations using _search and POST endpoint is not supported for bundle.
+            if (requestMethod == HTTPVerb.POST && requestUrl.Contains("_search", StringComparison.InvariantCulture))
+            {
+                throw new RequestNotValidException(string.Format(Api.Resources.InvalidBundleEntry, entry.Request.Url));
+            }
+
+            // Check for duplicate resources within a bundle entry is skipped if the entry is bundle or if the request within a entry is not modifying the resource.
+            return !(entry.Resource?.ResourceType == Hl7.Fhir.Model.ResourceType.Bundle
```
brandonpollett: We should file an issue to look at bundles w/ bundles inside. It might make sense to throw an exception in the case for now since we don't have a good way of validating it. Thoughts?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionValidator.cs
```diff
@@ -0,0 +1,66 @@
...
+                 || requestMethod == HTTPVerb.GET
+                 || requestUrl.Contains("$", StringComparison.InvariantCulture));
+        }
+
+        private static string GetResourceUrl(EntryComponent component)
+        {
+            if (component.Request.Method == HTTPVerb.POST)
```
brandonpollett: We probably want to also check that the value is not null or empty since I'm not sure there's validation on it coming in to the server for `POST`s if we're not already checking this case.

poadhika: are you referring to Method value or fullURl value? If this is for fullurl, then its okay to have this value null or empty since we are addressing that in the caller method.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionValidator.cs
poadhika: good point. I will file a issue for bundles supporting bundles. Right now i can throw a validation exception.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionValidator.cs
brandonpollett: Right, I missed that check on this call through.

PR #742, opened 2019-11-13
Optimizes _total parameter counting

rotodd:
## Description
- Implements an optimization for the [_total parameter](http://hl7.org/fhir/search.html#total) so that counting the number of matching resources only happens once

- Considers two scenarios:
  1. The results fit on one page
      - The Total value is calculated by counting the number of results returned from the first query
      - Only one query is issued
      - The possibility of a race condition is eliminated

  2. The results span multiple pages
      - Two queries are issued to include the Total value on the first page
      - The _total parameter is removed from the link to the next page, so the Total value is excluded from subsequent pages unless the user explicitly adds it to the Next link

total-parameter-counts-once

## Related issues
Addresses [#AB70827](https://microsofthealth.visualstudio.com/Health/_workitems/edit/70827).

## Testing
- Updates existing E2E test

### DIFF
--- a/src/Microsoft.Health.Fhir.Api/Features/Routing/UrlResolver.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/Routing/UrlResolver.cs
@@ -134,7 +134,6 @@ namespace Microsoft.Health.Fhir.Api.Features.Routing
                     if (string.Equals(searchParam.Key, KnownQueryParameterNames.Total, StringComparison.OrdinalIgnoreCase) && removeTotalParameter)
                     {
                         // Don't add the _total parameter to the route values.
-                        // TODO: Check how this impacts multiple parameters with the same name?
                         continue;
                     }
 
--- a/src/Microsoft.Health.Fhir.Api/Features/Routing/UrlResolver.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/Routing/UrlResolver.cs
@@ -111,7 +111,7 @@ namespace Microsoft.Health.Fhir.Api.Features.Routing
             return new Uri(uriString);
         }
 
-        public Uri ResolveRouteUrl(IEnumerable<Tuple<string, string>> unsupportedSearchParams = null, IReadOnlyList<(string parameterName, string reason)> unsupportedSortingParameters = null, string continuationToken = null)
+        public Uri ResolveRouteUrl(IEnumerable<Tuple<string, string>> unsupportedSearchParams = null, IReadOnlyList<(string parameterName, string reason)> unsupportedSortingParameters = null, string continuationToken = null, bool removeTotalParameter = false)
         {
             string routeName = _fhirRequestContextAccessor.FhirRequestContext.RouteName;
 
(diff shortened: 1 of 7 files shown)


src/Microsoft.Health.Fhir.Api/Features/Routing/UrlResolver.cs
```diff
@@ -130,8 +130,14 @@ public Uri ResolveRouteUrl(IEnumerable<Tuple<string, string>> unsupportedSearchP
                 foreach (KeyValuePair<string, StringValues> searchParam in Request.Query)
                 {
                     // Remove the parameter if:
-                    // 1. It is the _sort parameter and the parameter is not supported for sorting
-                    // 2. The parameter is not supported.
+                    // 1. It is the _total parameter, since we only want to count for the first page of results.
+                    if (string.Equals(searchParam.Key, KnownQueryParameterNames.Total, StringComparison.OrdinalIgnoreCase) && removeTotalParameter)
```
rotodd: This assumes that we should remove `_total` from the Next link, regardless of what it is set to.

In other words, `_total` could be set to either `accurate` or `none`, and in both scenarios it would be removed from the link to the next page.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -68,58 +68,52 @@ internal class SqlServerSearchService : SearchService
...
+                    // Count the results on the page.
+                    searchResult.TotalCount = searchResult.Results.Count();
+                }
+                else
+                {
+                    try
                     {
-                        try
-                        {
-                            searchResult = await SearchImpl(searchOptions, false, connection, cancellationToken, transaction);
```
rotodd: I moved this out of the try-block, addressing @johnstairs's [comment](https://github.com/microsoft/fhir-server/pull/731#discussion_r344317975) from my previous PR.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -68,58 +68,52 @@ internal class SqlServerSearchService : SearchService
...
+            {
+                searchResult = await SearchImpl(searchOptions, false, cancellationToken);
 
-                // If we should include the total count of matching search results
-                if (searchOptions.IncludeTotal == TotalType.Accurate && !searchOptions.CountOnly)
+                // If this is the first page and there aren't any more pages
+                if (searchOptions.ContinuationToken == null && searchResult.ContinuationToken == null)
                 {
-                    // Begin a transaction so we can perform two atomic reads.
-                    using (var transaction = connection.BeginTransaction(IsolationLevel.ReadCommitted))
```
rotodd: I removed this (for now) and updated the User Story [#AB70783](https://microsofthealth.visualstudio.com/Health/_workitems/edit/70783).

src/Microsoft.Health.Fhir.Api/Features/Routing/UrlResolver.cs
johnstairs: perf suggestion: put the removeTotalParameter first, so that we will not bother doing string.equals calls were removeTotalParameter is false.

rotodd: Thanks, implemented in fb17e80.

PR #743, opened 2019-11-13
Updated to preserve the response order in batch

rbans96:
## Description
BundleHandler is updated to keep the order of the requests in a batch so that bundling of the responses are in the same order as of the requests (earlier it was bundled in the order of the processing of each request)

## Related issues
[AB#71104](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/71104)

## Testing
Existing E2E tests and more E2E tests will be covered as part of #70787

**Local Testing - Screenshot(Request vs Response)**

image
image

**Handles empty request in Bundle**
image


src/Microsoft.Health.Fhir.R4.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -14,16 +14,16 @@ namespace Microsoft.Health.Fhir.Api.Features.Resources.Bundle
     /// </summary>
     public partial class BundleHandler
     {
-        private Dictionary<Hl7.Fhir.Model.Bundle.HTTPVerb, List<RouteContext>> GenerateRequestDictionary()
+        private Dictionary<Hl7.Fhir.Model.Bundle.HTTPVerb, Dictionary<RouteContext, int>> GenerateRequestDictionary()
```
brandonpollett: We should just use something like a list of tuples here instead of dictionaries. For example: `List<(RouteContext routeContext, int order)>`

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -179,7 +184,16 @@ private async Task ExecuteRequests(Hl7.Fhir.Model.Bundle responseBundle, Hl7.Fhi
...
-                responseBundle.Entry.Add(entryComponent);
+                if (responseBundle.Entry?.Any() != true)
+                {
+                    List<EntryComponent> entryComponents = new EntryComponent[entrySize].ToList();
+                    entryComponents[entry.Value] = entryComponent;
+                    responseBundle.Entry = entryComponents;
+                }
+                else
+                {
+                    responseBundle.Entry[entry.Value] = entryComponent;
```
brandonpollett: We can remove the else and the `entryComponents[entry.Value] = entryComponent;` the and always do `responseBundle.Entry[entry.Value] = entryComponent;`.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -179,7 +184,16 @@ private async Task ExecuteRequests(Hl7.Fhir.Model.Bundle responseBundle, Hl7.Fhi
                     entryComponent.Response = new Hl7.Fhir.Model.Bundle.ResponseComponent { Status = ((int)HttpStatusCode.NotFound).ToString() };
                 }
 
-                responseBundle.Entry.Add(entryComponent);
+                if (responseBundle.Entry?.Any() != true)
+                {
+                    List<EntryComponent> entryComponents = new EntryComponent[entrySize].ToList();
+                    entryComponents[entry.Value] = entryComponent;
+                    responseBundle.Entry = entryComponents;
```
brandonpollett: This could be simplified to `responseBundle.Entry = new List<EntryComponent>(entrySize)`

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -179,7 +184,16 @@ private async Task ExecuteRequests(Hl7.Fhir.Model.Bundle responseBundle, Hl7.Fhi
                     entryComponent.Response = new Hl7.Fhir.Model.Bundle.ResponseComponent { Status = ((int)HttpStatusCode.NotFound).ToString() };
                 }
 
-                responseBundle.Entry.Add(entryComponent);
+                if (responseBundle.Entry?.Any() != true)
```
brandonpollett: Would it be sufficient to only do `if(responseBundle.Entry != null)`?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -38,10 +39,11 @@ public partial class BundleHandler : IRequestHandler<BundleRequest, BundleRespon
         private readonly IFhirRequestContextAccessor _fhirRequestContextAccessor;
         private readonly FhirJsonSerializer _fhirJsonSerializer;
         private readonly FhirJsonParser _fhirJsonParser;
-        private readonly Dictionary<Hl7.Fhir.Model.Bundle.HTTPVerb, List<RouteContext>> _requests;
+        private readonly Dictionary<Hl7.Fhir.Model.Bundle.HTTPVerb, Dictionary<RouteContext, int>> _requests;
         private readonly IHttpAuthenticationFeature _httpAuthenticationFeature;
         private readonly IRouter _router;
         private readonly IServiceProvider _requestServices;
+        private int entrySize;
```
brandonpollett: Should be named `_entrySize`. Might want to consider renaming it something like `_requestCount`.

src/Microsoft.Health.Fhir.R4.Api/Features/Resources/Bundle/BundleHandler.cs
rbans96: 👍

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
rbans96: I wrote this first but then found out.. responseBundle.Entry is not null but count is zero.. so the statement validates for both (null and count). And if Entry list is null or count is zero.. I created a list with the capacity of requests size

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
rbans96: 👍

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
rbans96: I wrote this first too and found out that it creates a list and assigns the capacity to entrySize but the list itself is empty ie. entry can't be added to list[index]. So with the above logic, if Entry is null or count is zero, I am creating a list with the rows to the length of entrySize and setting it to responseBundle. Then for the subsequent iteration only else part will be executed

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
rbans96: It will not work because  if Entry is null or count is zero, I am creating a list with the rows to the length of entrySize and setting it to responseBundle. Then for the subsequent iteration only else part will be executed

brandonpollett: What I mean is it should be able to be simplified to the following:
```csharp
if (responseBundle.Entry?.Any() != true)
{
    responseBundle.Entry = new EntryComponent[_requestCount].ToList();
}

responseBundle.Entry[tuple.Item2] = entryComponent;
```

rbans96: got it, thanks for the clarification

brandonpollett:
Can you look at adding a unit test to cover this since we now have the `BundleHandlerTests.cs` in the code?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -288,7 +293,14 @@ private async Task ExecuteRequests(Hl7.Fhir.Model.Bundle responseBundle, Hl7.Fhi
                     };
                 }
 
-                responseBundle.Entry.Add(entryComponent);
+                if (responseBundle.Entry?.Any() != true)
+                {
+                    List<EntryComponent> entryComponents = new EntryComponent[_requestCount].ToList();
+                    entryComponents[tuple.Item2] = entryComponent;
```
brandonpollett: This should be able to be removed since we're doing the same thing below.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -207,8 +211,9 @@ private static void AddHeaderIfNeeded(string headerKey, string headerValue, Http
 
         private async Task ExecuteRequests(Hl7.Fhir.Model.Bundle responseBundle, Hl7.Fhir.Model.Bundle.HTTPVerb httpVerb)
         {
-            foreach (RouteContext request in _requests[httpVerb])
+            foreach (var tuple in _requests[httpVerb])
```
brandonpollett: nit: for readability we could use deconstruction here -- so something like `foreach ((RouteContext routeContext, int entryIndex) in _requests[httpVerb])`

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Resources/Bundle/BundleHandlerTests.cs
```diff
@@ -131,17 +132,83 @@ public async Task GivenABundleWithAGet_WhenNotAuthorized_ReturnsABundleResponseW
...
+
+            var bundleResource = bundleResponse.Bundle.ToPoco<Hl7.Fhir.Model.Bundle>();
+            Assert.Equal(Hl7.Fhir.Model.Bundle.BundleType.BatchResponse, bundleResource.Type);
+            Assert.Equal(3, bundleResource.Entry.Count);
+            Assert.Equal("403", bundleResource.Entry[0].Response.Status);
+            Assert.Equal("404", bundleResource.Entry[1].Response.Status);
+            Assert.Equal("200", bundleResource.Entry[2].Response.Status);
+        }
+
+        private void RouteAsyncFunction(CallInfo callInfo)
```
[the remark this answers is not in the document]
rbans96: Agree, but at this point there are couple of test cases which are referencing this method. If we localize this method, we would end up with redundant code. We might need to refactor this method at later point in time for future test cases. Thoughts?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
jackliums: Adding a comment in the code could help whoever read the code later to understand why it's done this way.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
jackliums: This does not need to happen in the foreach loop right? It could just happen once before the loop?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -193,7 +197,7 @@ private async Task FillRequestLists(List<Hl7.Fhir.Model.Bundle.EntryComponent> b
                     RouteData = routeContext.RouteData,
                 };
 
-                _requests[entry.Request.Method.Value].Add(routeContext);
+                _requests[entry.Request.Method.Value].Add((routeContext, order++));
```
jackliums: what happen if the entry.Request.Method is not one that's defined in the dictionary? For example, Patch is not defined in Stu3 BundleHandler but it doesn't stop me from specifying Patch in the request as one of the request. Is there validation that happens before this?

jackliums: Oh nevermind, I see that this is HTTPVerb enum which is defined differently for each version.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -151,6 +153,8 @@ private static List<OperationOutcomeIssue> GetOperationOutcomeIssues(List<Operat
 
         private async Task FillRequestLists(List<Hl7.Fhir.Model.Bundle.EntryComponent> bundleEntries)
         {
+            int order = 0;
+            _requestCount = bundleEntries.Count;
             foreach (Hl7.Fhir.Model.Bundle.EntryComponent entry in bundleEntries)
             {
                 if (entry.Request.Method == null)
```
jackliums: Would this be a problem with the order counting? I don't know what the response looks like but assume you have an incoming bundle of GET, POST, empty, DELETE, when it encounter the empty request, it skips is but does not increment the order, what would the response look like?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
jackliums: Not sure why my comments are not showing up:

This does not need to happen in the foreach loop right? It could just happen once before the loop? I think it might be a little easier to read if:

1. Create an array of size _requestCount before foreach loop
2. Sets the entryComponent to the corresponding index in foreach loop
3. After the forech loop, either do responseBundle.Entry.AddRange(array) or do array.ToList()

src/Microsoft.Health.Fhir.R4.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -14,16 +14,16 @@ namespace Microsoft.Health.Fhir.Api.Features.Resources.Bundle
...
-            return new Dictionary<Hl7.Fhir.Model.Bundle.HTTPVerb, List<RouteContext>>
+            return new Dictionary<Hl7.Fhir.Model.Bundle.HTTPVerb, List<(RouteContext, int)>>
             {
-                { Hl7.Fhir.Model.Bundle.HTTPVerb.DELETE, new List<RouteContext>() },
-                { Hl7.Fhir.Model.Bundle.HTTPVerb.GET, new List<RouteContext>() },
-                { Hl7.Fhir.Model.Bundle.HTTPVerb.HEAD, new List<RouteContext>() },
-                { Hl7.Fhir.Model.Bundle.HTTPVerb.PATCH, new List<RouteContext>() },
-                { Hl7.Fhir.Model.Bundle.HTTPVerb.POST, new List<RouteContext>() },
-                { Hl7.Fhir.Model.Bundle.HTTPVerb.PUT, new List<RouteContext>() },
+                { Hl7.Fhir.Model.Bundle.HTTPVerb.DELETE, new List<(RouteContext, int)>() },
```
jackliums: Not your code, but I would like to suggest a small:

What we care about is the order in which the operations gets executed (and GenerateRequestDictionary really is just operating on top of that) so, I think we can simplify the code to:

In BundleHandler.cs, define a member variable like

`private readonly Hl7.Fhir.Model.Bundle.HTTPVerb[] _verbExecutionOrder;`

In shared BundleHandler class, add `:this()` after the constructor.

Then in the Stu3 and R4 specific BundleHandler class, add the private constructor that initializes version specific orders:

Stud3:
```
        private BundleHandler()
        {
            _verbExecutionOrder = new []
            {
                Hl7.Fhir.Model.Bundle.HTTPVerb.DELETE,
                Hl7.Fhir.Model.Bundle.HTTPVerb.POST,
                Hl7.Fhir.Model.Bundle.HTTPVerb.PUT,
                Hl7.Fhir.Model.Bundle.HTTPVerb.GET,
            };
        }
```

R4:
```
        private BundleHandler()
        {
            _verbExecutionOrder = new []
            {
                Hl7.Fhir.Model.Bundle.HTTPVerb.DELETE,
                Hl7.Fhir.Model.Bundle.HTTPVerb.POST,
                Hl7.Fhir.Model.Bundle.HTTPVerb.PUT,
                Hl7.Fhir.Model.Bundle.HTTPVerb.PATCH,
                Hl7.Fhir.Model.Bundle.HTTPVerb.GET,
                Hl7.Fhir.Model.Bundle.HTTPVerb.HEAD,
            };
        }
```
Remove both GenerateRequestDictionary and ExecuteAllRequests from version specific BundleHandler.cs files and move the logic into the shared BundleHandler.cs

rbans96: 👍

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
rbans96: good point.. so as of now let's say request has 10 entries(with one empty entry), then response has 9 entries as its not appending anything for the empty one.

jackliums: I don't think that would be a desired behavior. We should either reject the call altogether (because it contains invalid entry) or return the empty entry where it is so that the caller can correlate their request to the response.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -71,13 +75,24 @@ public BundleHandler(IHttpContextAccessor httpContextAccessor, IFhirRequestConte
...
+            _requests = _verbExecutionOrder.ToDictionary(verb => verb, _ => new List<(RouteContext, int)>());
 
             _httpAuthenticationFeature = httpContextAccessor.HttpContext.Features.Get<IHttpAuthenticationFeature>();
             _router = httpContextAccessor.HttpContext.GetRouteData().Routers.First();
             _requestServices = httpContextAccessor.HttpContext.RequestServices;
         }
 
+        private async Task ExecuteAllRequests(Hl7.Fhir.Model.Bundle responseBundle)
+        {
+            EntryComponent[] entryComponents = new EntryComponent[_requestCount];
```
jackliums: a bit of comment here would help (and why new List(_requestCount) won't work).

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
jackliums: Since empty entry is probably invalid, one potential fix is to return an OperationOutcome indicating that it was a bad request for this entry.

PR #744, opened 2019-11-13
Fix for ResourceType not being set in metric notifications

YazanMSFT:
## Description
The ResourceType field in FhirRequestContext was not set correctly since the middleware uses RouteData to set it, but RouteData is not set when the middleware is run. Fixing this by adding a filter.

## Related issues
Addresses [issue #745 ].

## Testing
Unit tests.


src/Microsoft.Health.Fhir.Api/Features/Filters/FhirRequestContextResourceTypeFilterAttribute.cs
```diff
@@ -0,0 +1,43 @@
...
+using EnsureThat;
+using Microsoft.AspNetCore.Mvc.Filters;
+using Microsoft.AspNetCore.Routing;
+using Microsoft.Health.Fhir.Api.Features.Routing;
+using Microsoft.Health.Fhir.Core.Features.Context;
+
+namespace Microsoft.Health.Fhir.Api.Features.Filters
+{
+    [AttributeUsage(AttributeTargets.Class)]
+    public class FhirRequestContextResourceTypeFilterAttribute : ActionFilterAttribute
```
brandonpollett: We might want to think about combining this with `FhirRequestContextRouteNameFilterAttribute.cs` and updating the name to be something about hydrating the `FhirRequestContext`.

src/Microsoft.Health.Fhir.Api/Features/Filters/FhirRequestContextResourceTypeFilterAttribute.cs
```diff
@@ -0,0 +1,43 @@
...
+            RouteData routeData = context.RouteData;
+
+            object resourceType = null;
+
+            if (routeData != null && routeData.Values != null)
+            {
+                routeData.Values.TryGetValue(KnownActionParameterNames.ResourceType, out resourceType);
+            }
+
+            _fhirRequestContextAccessor.FhirRequestContext.ResourceType = resourceType?.ToString();
```
brandonpollett: could just use the `fhirRequestContext` we're declaring above or remove that declaration. Also, there's probably no reason to call this if the `TryGetValue` fails above.

src/Microsoft.Health.Fhir.Api/Features/Filters/FhirRequestContextResourceTypeFilterAttribute.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        public override void OnActionExecuting(ActionExecutingContext context)
+        {
+            IFhirRequestContext fhirRequestContext = _fhirRequestContextAccessor.FhirRequestContext;
+
+            RouteData routeData = context.RouteData;
+
+            object resourceType = null;
+
+            if (routeData != null && routeData.Values != null)
```
brandonpollett: Could we do `if (routeData?.Values != null)`

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/FhirRequestContextRouteDataPopulatingFilterAttribute.cs
```diff
@@ -65,6 +66,19 @@ public override void OnActionExecuting(ActionExecutingContext context)
                     }
                 }
             }
+
+            // Set the resource type based on the route data
+            RouteData routeData = context.RouteData;
+
+            object resourceType = null;
```
brandonpollett: nit: we could inline this declaration to the out variable below.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Metric/MetricTests.cs
```diff
@@ -70,22 +74,28 @@ private async Task ExecuteAndValidate<T>(Func<Task<T>> action, params (Type type
 
             var result = await action() as HttpResponseMessage;
 
-            foreach ((Type type, int count) expectedNotification in expectedNotifications)
+            foreach ((Type type, int count, string resourceType) expectedNotification in expectedNotifications)
             {
                 if (expectedNotification.count == 0)
                 {
-                    Assert.False(_metricHandler.HandleCountDictionary.TryGetValue(expectedNotification.type, out var count));
+                    Assert.False(_metricHandler.NotificationMapping.TryGetValue(expectedNotification.type, out var items));
```
brandonpollett: nit: we can do `out _` since we're not using this value.

src/Microsoft.Health.Fhir.Core/Features/Context/FhirRequestContext.cs
```diff
@@ -62,6 +62,6 @@ public class FhirRequestContext : IFhirRequestContext
 
         public IDictionary<string, StringValues> ResponseHeaders { get; }
 
-        public string ResourceType { get; }
+        public string ResourceType { get; set; }
```
jackliums: I think for RouteName and ResourceType, we should consider throw an example if the value has not be set to avoid weird bugs down the road where depending when it's access, it might return data or null.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Metric/MetricHandler.cs
```diff
@@ -8,33 +8,32 @@
...
-            HandleCountDictionary.Clear();
+            NotificationMapping.Clear();
         }
 
-        public Task Handle(IMetricsNotification notification, CancellationToken cancellationToken)
+        public Task Handle(INotification notification, CancellationToken cancellationToken)
         {
             Type notificationType = notification.GetType();
-            if (HandleCountDictionary.ContainsKey(notificationType))
+            if (NotificationMapping.ContainsKey(notificationType))
```
jackliums: TryGetValue to avoid double lookup

src/Microsoft.Health.Fhir.Core/Features/Context/FhirRequestContext.cs
YazanMSFT: I tried to do that, but endpoint like health/check would fail in that case, where it doesn't go through the MVC pipeline. I discussed it with John and we decided that it should be OK to leave it without throwing an exception, and having tests that ensure it's set for the endpoints we expect it to be set.

PR #748, opened 2019-11-15
Updates package 'IdentityServer4' to version '3.0.2'

MicrosoftHealthService:


feordin:
Build error shows that IdentityServer4 is only comptabile with .Net Core 3.0.  Until we move to 3.0, we won't be able to take this update.

PR #753, opened 2019-11-18
add docker and kubernetes

marrobi:
## Description
Added dockerfile, docker compose and kubernetes config and readme.

Things to sort:
[ ] passwords/connection string for sql
[ ] Split restore and build in docker image?
[ ] Which projects actually need building?
[ ] can config be done using env vars?


marrobi:
Sorry - wrong fork!

Issue #760, opened 2019-11-21
Patient Search with Multiple Given Names return zero results.


granadacoder:
**Describe the bug**

When searching for a Patient using multiple given names, (more than 1), the results come back empty on legitimate inputs.

**To Reproduce**
1.  Use Sql Server (not CosmoDB)
2. Add some Patient(s) Resources to the database.  With Multiple Given Names.

One like this:

```
                "name": [
                    {
                        "family": "Smith",
                        "given": [
                            "John",
                            "Apple",
                            "Jingerheimer"
                        ]
                    }
                ]
```
second one like this:

```
                "name": [
                    {
                        "family": "Smith",
                        "given": [
                            "John",
                            "Banana",
                            "Jingerheimer"
                        ]
                    }
                ]

```
Same two patients seen below with more json.

```

{
    "resourceType": "Bundle",
    "id": "269caf66-0ccc-43e7-b9a5-f16f84db0149",
    "meta": {
        "lastUpdated": "2019-11-20T19:30:26.858917+00:00"
    },
    "type": "searchset",
    "link": [
        {
            "relation": "self",
            "url": "https://localhost:44348/Patient?given=Jingerheimer"
        }
    ],
    "entry": [
        {
            "fullUrl": "https://localhost:44348/Patient/504f6bd3-e9b4-4846-8948-97bf09c70722",
            "resource": {
                "resourceType": "Patient",
                "id": "504f6bd3-e9b4-4846-8948-97bf09c70722",
                "meta": {
                    "versionId": "1",
                    "lastUpdated": "2019-11-20T19:26:11.005+00:00"
                },
                "identifier": [
                    {
                        "system": "ssn",
                        "value": "111-11-1111"
                    },
                    {
                        "system": "uuid",
                        "value": "da55d068e0784b359fa97498a11543c5"
                    }
                ],
                "name": [
                    {
                        "family": "Smith",
                        "given": [
                            "John",
                            "Apple",
                            "Jingerheimer"
                        ]
                    }
                ]
            },
            "search": {
                "mode": "match"
            }
        },
        {
            "fullUrl": "https://localhost:44348/Patient/10054ce9-6141-4eca-bc5b-0978f8c8afcb",
            "resource": {
                "resourceType": "Patient",
                "id": "10054ce9-6141-4eca-bc5b-0978f8c8afcb",
                "meta": {
                    "versionId": "1",
                    "lastUpdated": "2019-11-20T19:26:48.962+00:00"
                },
[...]

PR #763, opened 2019-11-22
Remove TableExpressionCombiner

johnstairs:
## Description
Removes an SQL optimization rewriter that can produce incorrect results.

Given a seach like `probability=gt0.8&probability=lt0.9`, it was intended to collapse the SQL down to a single CTE with a predicate like `WHERE value > 0.8 AND value < 0.9` instead of intersecting the results of two queries.

This optimization is only safe to do when used against search parameters that will yield at most one value per resource. We currently do not have the necessary information to make that determination (though we should add it), so we are removing this pass altogether.

## Related issues
Addresses #760 

## Testing
Added integration test


test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/StringSearchTestFixture.cs
```diff
@@ -21,13 +21,13 @@ public StringSearchTestFixture(DataStore dataStore, Format format, TestFhirServe
             FhirClient.DeleteAllResources(ResourceType.Patient).Wait();
 
             Patients = FhirClient.CreateResourcesAsync<Patient>(
-                p => SetPatientInfo(p, "Seattle", "Smith"),
-                p => SetPatientInfo(p, "Portland", "Williams"),
-                p => SetPatientInfo(p, "Vancouver", "Anderson"),
-                p => SetPatientInfo(p, LongString, "Murphy"))
+                    p => SetPatientInfo(p, "Seattle", "Smith", given: "Bea"),
```
brandonpollett: nit: Should we add an additional case where the patient either shares the family or given name? That way we can also verify if you search `name=Bea&name=Smith`, it only returns one that matches both?

johnstairs: Good suggestion.

PR #764, opened 2019-11-22
Adds batch e2e tests

rbans96:
## Description
Adds e2e Audit, Batch and Metric tests

## Related issues
Addresses [AB#70787](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/70787)

## Testing
Run tests


rbans96:
Accidentally created this branch while I was checked out to other branch and not master - so it shows few commits history(the commits previous to "initial commit") of other branch.

test/Microsoft.Health.Fhir.R4.Tests.E2E/Rest/Audit/AuditTests.cs
```diff
@@ -0,0 +1,41 @@
...
+                return;
+            }
+
+            var expectedActions = new List<string> { "batch", "delete", "delete", "create", "create", "create", "create", "update", "update", "update", "update", "update", "update", "search-type", "search-type", "read", "read", "batch" };
+            var expectedPathSegments = new List<string> { string.Empty, "Patient/234", "Patient/234", "Patient", "Patient", "Patient", "Patient", "Patient/123", "Patient/123", "Patient?identifier=http:/example.org/fhir/ids|456456", "Patient?identifier=http:/example.org/fhir/ids|456456", "Patient/123", "Patient/123", "Patient?name=peter", "Patient?name=peter", "Patient/12334", "Patient/12334", string.Empty };
+            var expectedStatusCodes = new List<HttpStatusCode> { HttpStatusCode.OK, HttpStatusCode.OK, HttpStatusCode.NoContent, HttpStatusCode.OK, HttpStatusCode.Created, HttpStatusCode.NoContent, HttpStatusCode.OK, HttpStatusCode.OK, HttpStatusCode.OK, HttpStatusCode.OK, HttpStatusCode.OK, HttpStatusCode.OK, HttpStatusCode.PreconditionFailed, HttpStatusCode.OK, HttpStatusCode.OK, HttpStatusCode.NotFound, HttpStatusCode.NotFound, HttpStatusCode.OK };
+            await ExecuteAndValidateBatch(
```
brandonpollett: Might be cleaner to make this a list of tuples such that we have something like `List<(string auditType, string pathSegments, HttpStatusCode statusCode)>`

Thoughts?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/BatchTests.cs
```diff
@@ -27,11 +40,63 @@ public BatchTests(HttpIntegrationTestFixture fixture)
...
+        public async Task WhenSubmittingABatch_GivenAProperBundleWithFewFailedRequests_ThenSuccessIsReturnedAndOperationOutcomesForFailedRequests()
+        {
+            Resource requestBundle = Samples.GetDefaultBatch().ToPoco<Bundle>();
+
+            FhirResponse<Bundle> fhirResponse = await Client.PostBundleAsync(requestBundle);
+            Assert.NotNull(fhirResponse);
+            Assert.Equal(HttpStatusCode.OK, fhirResponse.StatusCode);
+            foreach (var entry in fhirResponse.Resource.Entry)
+            {
+                if (entry.Response.Status.StartsWith("4"))
```
brandonpollett: We should verify that we got the expected number of exceptions and that they were in the expected index.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/BatchTests.cs
```diff
@@ -27,11 +40,63 @@ public BatchTests(HttpIntegrationTestFixture fixture)
...
+        {
+            FhirClient tempClient = Client.CreateClientForUser(TestUsers.ReadOnlyUser, TestApplications.NativeClient);
+            Resource requestBundle = Samples.GetDefaultBatch().ToPoco<Bundle>();
+
+            FhirResponse<Bundle> fhirResponse = await tempClient.PostBundleAsync(requestBundle);
+            Assert.NotNull(fhirResponse);
+            Assert.Equal(HttpStatusCode.OK, fhirResponse.StatusCode);
+
+            // Since first five requests are POST/PUT
+            for (int i = 0; i < 5; i++)
```
brandonpollett: We also need to verify that the reads succeeded.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Metric/MetricTests.cs
```diff
@@ -23,6 +23,7 @@
 
 namespace Microsoft.Health.Fhir.Tests.E2E.Rest.Metric
 {
+    [Trait(Traits.Category, Categories.Batch)]
```
brandonpollett: We only want to mark the actual batch test with this category.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Metric/MetricTests.cs
```diff
@@ -64,6 +65,18 @@ public async Task GivenHealthCheckPath_WhenInvoked_MetricNotificationsNotEmitted
...
         }
 
+        [Trait(Traits.Priority, Priority.One)]
+        [Fact]
+        public async Task GivenABatch_WhenInvoked_MetricNotificationsShouldBeEmitted()
+        {
+            _metricHandler?.ResetCount();
+
+            await ExecuteAndValidate(
+                () => _client.CreateAsync(Samples.GetDefaultBatch().ToPoco()),
```
brandonpollett: I think this is going to just create a bundle on the server and not actually do a batch. You most likely need to do `_client.PostBundleAsync`

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/BatchTests.cs
rbans96: This method just verifies that operationOutcome is never empty for the exceptions. All HttpStatusCodes(including exceptions) and its index are verified as part of WhenSubmittingABatch_GivenAProperBundle_ThenSuccessIsReturnedForBatchAndExpectedStatusCodesPerRequests method.

test/Microsoft.Health.Fhir.R4.Tests.E2E/KnownCrucibleTests.cs
```diff
@@ -434,8 +434,6 @@ public class KnownCrucibleTests
             "testscript-example-update/SETUP",
             "transactionandbatchtest/XFER0",
             "transactionandbatchtest/XFER10",
-            "transactionandbatchtest/XFER11",
-            "transactionandbatchtest/XFER12",
```
rbans96: Got a hint from the crucible test failure as below-
image

test/Microsoft.Health.Fhir.Stu3.Tests.E2E/KnownCrucibleTests.cs
```diff
@@ -429,8 +429,6 @@ public class KnownCrucibleTests
             "testscript-example-update/SETUP",
             "transactionandbatchtest/XFER0",
             "transactionandbatchtest/XFER10",
-            "transactionandbatchtest/XFER11",
-            "transactionandbatchtest/XFER12",
```
rbans96: Got a hint from the crucible test failure as below-
image

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Audit/AuditTests.cs
```diff
@@ -21,8 +21,11 @@
 
 namespace Microsoft.Health.Fhir.Tests.E2E.Rest.Audit
 {
+    /// <summary>
+    /// Provides Audit specific tests.
+    /// </summary
     [HttpIntegrationFixtureArgumentSets(DataStore.CosmosDb, Format.Json)]
-    public class AuditTests : IClassFixture<AuditTestFixture>
+    public partial class AuditTests : IClassFixture<AuditTestFixture>
     {
```
poadhika: Why is this class now partial?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Audit/AuditTests.cs
```diff
@@ -21,8 +21,11 @@
 
 namespace Microsoft.Health.Fhir.Tests.E2E.Rest.Audit
 {
+    /// <summary>
+    /// Provides Audit specific tests.
+    /// </summary
     [HttpIntegrationFixtureArgumentSets(DataStore.CosmosDb, Format.Json)]
-    public class AuditTests : IClassFixture<AuditTestFixture>
+    public partial class AuditTests : IClassFixture<AuditTestFixture>
```
poadhika: Why is this class now partial?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/BatchTests.cs
poadhika: Instead of only checking for failed operation, can we assert for both success and failures here and validating the operation outcome message for failed cases?

rbans96: Since success and failure status codes for each entry in bundle are verified in another test method. I think it would be redundant if I assert for success status codes here as well for the same bundle.. Thoughts?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Audit/AuditTests.cs
rbans96: Because of data issue in stu3, the returned status codes for few entries in the bundle are different from R4. So leveraging partial class functionality,  AuditTest for the Batch is written specificly for R4.

test/Microsoft.Health.Fhir.R4.Tests.E2E/Rest/Audit/AuditTests.cs
```diff
@@ -0,0 +1,58 @@
...
+        [Trait(Traits.Priority, Priority.One)]
+        public async Task GivenABatch_WhenPost_ThenAuditLogEntriesShouldBeCreated()
+        {
+            if (!_fixture.IsUsingInProcTestServer)
+            {
+                // This test only works with the in-proc server with customized middleware pipeline
+                return;
+            }
+
+            List<(string expectedActions, string expectedPathSegments, HttpStatusCode expectedStatusCodes)> expectedList = new List<(string, string, HttpStatusCode)>
```
poadhika: nit: change expectedActions to expectedAuditType

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/BatchTests.cs
```diff
@@ -42,5 +88,35 @@ public async Task WhenSubmittingABatch_GivenANonBundleResource_ThenBadRequestIsR
 
             Assert.Equal(HttpStatusCode.BadRequest, ex.StatusCode);
         }
+
+        private void ValidateResourceStatusCode(Bundle resource)
```
poadhika: might want to rename the method name to ValidateResourceOutput since we are validating both status and operationoutcome in case of failures as well?

test/Microsoft.Health.Fhir.R4.Tests.E2E/Rest/Audit/AuditTests.cs
```diff
@@ -0,0 +1,58 @@
...
+using Microsoft.Health.Fhir.Core.Extensions;
+using Microsoft.Health.Fhir.Tests.Common;
+using Xunit;
+
+namespace Microsoft.Health.Fhir.Tests.E2E.Rest.Audit
+{
+    /// <summary>
+    /// Provides R4 specific tests.
+    /// </summary>
+    /// [Trait(Traits.Category, Categories.Batch)]
```
brendankowitz: We might need this after we release

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/BatchTests.cs
```diff
@@ -27,11 +40,44 @@ public BatchTests(HttpIntegrationTestFixture fixture)
 
         [Fact]
         [Trait(Traits.Priority, Priority.One)]
-        public async Task WhenSubmittingABatch_GivenAProperBundle_ThenNotSupportedIsReturned()
+        public async Task WhenSubmittingABatch_GivenAProperBundle_ThenSuccessIsReturnedForBatchAndExpectedStatusCodesPerRequests()
```
brendankowitz: Perhaps `GivenAValidBundle`

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/BatchTests.cs
```diff
@@ -42,5 +88,34 @@ public async Task WhenSubmittingABatch_GivenANonBundleResource_ThenBadRequestIsR
...
+
+        private void ValidateResourceOutput(Bundle resource)
+        {
+            Assert.Equal("201", resource.Entry[0].Response.Status);
+            Assert.True("201".Equals(resource.Entry[1].Response.Status) || "200".Equals(resource.Entry[1].Response.Status), "Conditional Create");
+            Assert.True("200".Equals(resource.Entry[2].Response.Status) || "201".Equals(resource.Entry[2].Response.Status), "Update");
+            Assert.True("201".Equals(resource.Entry[3].Response.Status) || "200".Equals(resource.Entry[3].Response.Status), "Update or Create");
+
+            // Passes locally each time but fails in pipeline. Investigation is required
+            // Assert.True("412".Equals(resource.Entry[4].Response.Status) || "404".Equals(resource.Entry[4].Response.Status), "Conditional update");
```
brendankowitz: Needed?

rbans96: It should not be a blocker as this is only failing for sqlServer, json for stu3 and R4 but not sure why.
Thoughts?

poadhika: are you saying its failing for sql+json combination?

rbans96: Its just failing for sql+xml and R4. Sorry for the confusion

poadhika: Based on our conversation  it looks like its failing for sql-xml combination for R4. It might makes sense to spend sometime on checking why its faiing in server?

src/Microsoft.Health.Fhir.Stu3.Web/appsettings.json
```diff
@@ -22,7 +22,7 @@
             "SupportsXml": true
         },
         "CoreFeatures": {
-            "SupportsBatch": false,
+            "SupportsBatch": true,
```
jackliums: Just curious, is this something that's ready for prime time and we want to enable by default now? If it's for testing only, there is a test configuration override that you can do (you can reference corstestconfiguration.json for example).

test/Microsoft.Health.Fhir.R4.Tests.E2E/Rest/Audit/AuditTests.cs
```diff
@@ -0,0 +1,58 @@
...
+            if (!_fixture.IsUsingInProcTestServer)
+            {
+                // This test only works with the in-proc server with customized middleware pipeline
+                return;
+            }
+
+            List<(string expectedActions, string expectedPathSegments, HttpStatusCode expectedStatusCodes)> expectedList = new List<(string, string, HttpStatusCode)>
+            {
+                ("batch", string.Empty, HttpStatusCode.OK),
+                ("delete", "Patient/234", HttpStatusCode.OK),
```
jackliums: Is this allowed? I thought the same resource within a batch/transaction cannot have multiple actions (even though it's no-op)?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Audit/AuditTests.cs
jackliums: Does that mean we are not supporting Batch for Stu3? If we do support for Stu3, can we have specific tests written for Stu3 as well?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Audit/AuditTests.cs
```diff
@@ -389,6 +392,59 @@ private async Task ExecuteAndValidate<T>(Func<Task<FhirResponse<T>>> action, str
...
+
+            string correlationId = response.Headers.GetValues(RequestIdHeaderName).FirstOrDefault();
+
+            Assert.NotNull(correlationId);
+
+            string expectedAppId = TestApplications.ServiceClient.ClientId;
+
+            IReadOnlyList<AuditEntry> auditList = _auditLogger.GetAuditEntriesByCorrelationId(correlationId);
+
+            Assert.Equal(18, auditList.Count);
```
jackliums: I think the test should be as simple as possible and doesn't have too many conditions. 1. because more condition in test = harder to read/maintain, 2. more logic means it could have error in the logic itself. Also, the logic in this method is very much tied to the test (E.g., where did 18 come from? Why are there even and odd validations?)

test/Microsoft.Health.Fhir.R4.Tests.E2E/Rest/Audit/AuditTests.cs
rbans96: True. Two entries for the same resource is in the list are to verify the audit entries(executing and executed) per resource. To avoid confusion, refactored the expectedList to contain HttpStatusCodes as null for executing entries

src/Microsoft.Health.Fhir.Stu3.Web/appsettings.json
rbans96: @brandonpollett Thoughts?

brandonpollett: The original thought was to enable it by default now, but I could see the use of holding off until we get a bit more testing complete and the transaction tests in, too. 

@rbans96 - You can look into how this functions with the `corstestconfiguration.json` or `exporttestconfiguration.json` which are located in the `test\Configuration` folder. It will also require a bit of a change in the build system to make work which I can help with.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/BatchTests.cs
```diff
@@ -42,5 +88,32 @@ public async Task WhenSubmittingABatch_GivenANonBundleResource_ThenBadRequestIsR
 
             Assert.Equal(HttpStatusCode.BadRequest, ex.StatusCode);
         }
+
+        private void ValidateResourceOutput(Bundle resource)
```
jackliums: nit: is there advantage of splitting this into a separate method? It's only used by `WhenSubmittingABatch_GivenAValidBundle_ThenSuccessIsReturnedForBatchAndExpectedStatusCodesPerRequests` and specifically validating the result of that test so seems like the logic just belongs to that method?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Audit/AuditTests.cs
```diff
@@ -389,6 +392,48 @@ private async Task ExecuteAndValidate<T>(Func<Task<FhirResponse<T>>> action, str
                 ae => ValidateExecutedAuditEntry(ae, expectedAction, expectedResourceType, expectedUri, expectedStatusCode, correlationId, expectedAppId, ExpectedClaimKey));
         }
 
+        private async Task ExecuteAndValidateBatch<T>(Func<Task<FhirResponse<T>>> action, List<(string, string, HttpStatusCode?, ResourceType?)> expectedList)
```
brandonpollett: nit: use named tuples to increase readability (e.g. something like `(string auditAction, string route, HttpStatusCode? statusCode, ResourceType resourceType?)`)

PR #765, opened 2019-11-25
Intra-bundle resource reference resolution

brandonpollett:
## Description
This PR adds code to go through the resources that are part of a transaction to resolve any resource references that are part of the transaction. It does this by matching any `fullUrl`s in the posts with any resource references within resources in the transaction.

## Related issues
Addresses [AB#70309](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/70309).

## Testing
Unit test of the resource extension, manual testing of E2E scenario, all existing tests continue to pass.


src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -326,6 +353,42 @@ private async Task ExecuteRequests(Hl7.Fhir.Model.Bundle responseBundle, Hl7.Fhi
...
+            foreach (EntryComponent entry in bundleEntries)
+            {
+                if (entry.Request.Method != HTTPVerb.POST)
+                {
+                    continue;
+                }
+
+                // We've already come across this ID
+                if (!idDictionary.TryGetValue(entry.FullUrl, out _))
+                {
```
poadhika: Since caller can put any value in fullURL, what will happen if  the same value is given for a FullURl for two different post entries?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -326,6 +353,42 @@ private async Task ExecuteRequests(Hl7.Fhir.Model.Bundle responseBundle, Hl7.Fhi
             }
         }
 
+        private static void PopulateReferenceIdDictionary(IEnumerable<EntryComponent> bundleEntries, IDictionary<string, string> idDictionary)
+        {
+            foreach (EntryComponent entry in bundleEntries)
+            {
+                if (entry.Request.Method != HTTPVerb.POST)
+                {
```
poadhika: Is there a reason why we are handling this only for POST and not for PUT?  Does the reference resolution applies to only create scenarios and not for update?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
brandonpollett: That would be stopped by the code in the PR that does the verification.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
brandonpollett: `PUT` requires there to be an explicit ID provided in the request. 

My assumption (could be incorrect) is that if you are referencing something that is the subject of a `PUT` you would have already updated the references to it to use that explicit ID provided in the request.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/ResourceExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+namespace Microsoft.Health.Fhir.Api.Features.Resources
+{
+    public static class ResourceExtensions
+    {
+        private const string ChildrenPropertyName = "Children";
+
+        public static IEnumerable<T> GetAllChildren<T>(
+            this Resource resource)
+        {
+            var stack = new Stack<(PropertyInfo propertyInfo, object baseObject)>(resource.GetType().GetProperties().Where(x => x.Name == ChildrenPropertyName).Select(x => (x, resource as object)));
```
brendankowitz: Seems like only 1 property should have the name `Children`?

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Create/CreateResourceHandler.cs
```diff
@@ -34,8 +34,11 @@ public async Task<UpsertResourceResponse> Handle(CreateResourceRequest message,
 
             var resource = message.Resource.Instance.ToPoco<Resource>();
 
-            // If an Id is supplied on create it should be removed/ignored
-            resource.Id = null;
+            // Check to see if we want to persist the supplied Id
+            if (!message.PersistId)
```
brendankowitz: Can we just use `UpsertResourceRequest` instead in this case?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/ResourceExtensions.cs
brendankowitz: I read further down. I think this could also work as a recursive method with `GetAllChildren<T>(this Base resource)`. Which could then `foreach(r in b.Children)` (since this seems to be the only property of interest), I think we could keep static typing for this

brandonpollett: Yep, I was able to refactor this into a recursive method.

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Create/CreateResourceHandler.cs
brandonpollett: We possibly could, but I think we may get into some cases where we'd get unexpected behavior. 

A concrete example that looks like would already exists is that in `R4` it looks like the `UpsertResourceHandler` does a check for a setting called `RequireETag`. If that setting was enabled, you wouldn't be able to create in a transaction. We obviously could get around this by doing an additional check, but I'm not sure that end state is better

brandonpollett: This was updated to using a settable func so that we're only modifying how create/upsert creates a new id.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -253,13 +287,15 @@ private async Task ExecuteRequests(Hl7.Fhir.Model.Bundle responseBundle, Hl7.Fhi
 
                     _bundleHttpContextAccessor.HttpContext = httpContext;
 
+                    _resourceIdProvider.Create = () => persistedId;
```
brendankowitz: Not a blocker, but since this provider is still ambient in a context, I still feel this should be reverted to the previous state after the invoke

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -319,6 +355,41 @@ private async Task ExecuteRequests(Hl7.Fhir.Model.Bundle responseBundle, Hl7.Fhi
...
+        {
+            foreach (EntryComponent entry in bundleEntries)
+            {
+                if (entry.Request.Method != HTTPVerb.POST)
+                {
+                    continue;
+                }
+
+                // We've already come across this ID
+                if (!idDictionary.TryGetValue(entry.FullUrl, out _))
```
brendankowitz: Is this just ContainsKey?:
```suggestion
                if (!idDictionary.ContainsKey(entry.FullUrl))
```

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -319,6 +355,41 @@ private async Task ExecuteRequests(Hl7.Fhir.Model.Bundle responseBundle, Hl7.Fhi
...
+                {
+                    // This id is new to us
+                    var insertId = Guid.NewGuid().ToString();
```
brendankowitz:
```suggestion
                    var insertId = _resourceIdProvider.Create();
```

PR #741, opened 2019-11-12
Conditional Create/Update respects SearchEntryMode

brendankowitz:
## Description
Conditional Create/Update must respect SearchEntryMode.Match when searching for existing items

## Related issues
Addresses #730

## Testing
Existing unit/integration tests

PR #769, opened 2019-12-02
Conditional reference resolution in bundle transaction

poadhika:
## Description
A transaction bundle can include resources that have conditional references (searches). This PR addresses the resolution of those references. This PR also resolves and validates if the resources are uniue for the entires that have conditionalcreates and updates.

## Related issues
Addresses UserStory: [AB#70305](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/70305)

## Testing
Unit Tests has been added


src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -217,6 +235,43 @@ private async Task FillRequestLists(List<Hl7.Fhir.Model.Bundle.EntryComponent> b
...
+                if (referenceIdDictionary.TryGetValue(reference.Reference, out var referenceId))
+                {
+                    reference.Reference = referenceId;
+                }
+                else
+                {
+                    if (reference.Reference.Contains("?", StringComparison.InvariantCulture))
+                    {
+                        string[] queries = reference.Reference.Split("?");
+                        string resourceType = queries[0];
```
brandonpollett: Not sure if we need to guard against a system wide search if it's possible here.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionValidator.cs
```diff
@@ -42,6 +62,70 @@ public static void ValidateTransactionBundle(Hl7.Fhir.Model.Bundle bundle)
...
+                string resourceType = null;
+                StringValues conditionalQueries;
+                HTTPVerb requestMethod = (HTTPVerb)entry.Request.Method;
+                bool conditionalCreate = requestMethod == HTTPVerb.POST;
+                bool condtionalUpdate = requestMethod == HTTPVerb.PUT || requestMethod == HTTPVerb.DELETE;
+
+                if (condtionalUpdate)
+                {
+                    string[] conditinalUpdateParameters = entry.Request.Url.Split("?");
+                    resourceType = conditinalUpdateParameters[0];
```
brandonpollett: Same here for potentially a system wide search?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
poadhika: Can you please elaborate this one?

poadhika:
> You might want to look at https://github.com/microsoft/fhir-server/pull/741/files#diff-fd7a91f4021cb77ba58cdf737e58aece for some of the caveats while doing conditional stuff. I'm wondering if there's a way to consolidate between the two so we don't have to do the same filtering in multiple places.

I have now updated TransactionValidator to use baseconditionalhandler .

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -127,6 +132,9 @@ public async Task<BundleResponse> Handle(BundleRequest bundleRequest, Cancellati
             var bundleResource = bundleRequest.Bundle.ToPoco<Hl7.Fhir.Model.Bundle>();
             _bundleType = bundleResource.Type;
 
+            // For resources within a transaction, we need to validate if they are referring to each other and throw an exception in such case.
+            await _transactionValidator.ValidateBundle(bundleResource);
```
brendankowitz: The comment says within a transaction, but it will run for both batch and transaction, which is valid?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -211,6 +217,8 @@ private async Task FillRequestLists(List<Hl7.Fhir.Model.Bundle.EntryComponent> b
                             persistedId = value.resourceId;
                         }
                     }
+
+                    // await ResolveBundleReferencesAsync(entry, referenceIdDictionary);
```
brendankowitz: Needed?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -251,6 +259,44 @@ private async Task FillRequestLists(List<Hl7.Fhir.Model.Bundle.EntryComponent> b
             }
         }
 
+        public async Task ResolveIntraBundleReferences(EntryComponent entry, Dictionary<string, (string resourceId, string resourceType)> referenceIdDictionary)
+        {
+            IEnumerable<ResourceReference> references = entry.Resource.GetAllChildren<ResourceReference>();
+
+            foreach (ResourceReference reference in references)
+            {
+                // We've already come across this ID
```
brendankowitz: Perhaps?
```suggestion
                // Checks to see if this reference has already been assigned an Id
```

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -251,6 +259,44 @@ private async Task FillRequestLists(List<Hl7.Fhir.Model.Bundle.EntryComponent> b
...
+                if (referenceIdDictionary.TryGetValue(reference.Reference, out var referenceInformation))
+                {
+                    reference.Reference = $"{referenceInformation.resourceType}/{referenceInformation.resourceId}";
+                }
+                else
+                {
+                    if (reference.Reference.Contains("?", StringComparison.InvariantCulture))
```
brendankowitz: I'd probably just use StringComparison.Ordinal (byte comparison) unless we have a reason not to

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -251,6 +259,44 @@ private async Task FillRequestLists(List<Hl7.Fhir.Model.Bundle.EntryComponent> b
...
+                    {
+                        string[] queries = reference.Reference.Split("?");
+                        string resourceType = queries[0];
```
brendankowitz: Should we check `IModelInfoProvider.IsKnownResource` here to ensure we have a valid value?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionValidator.cs
```diff
@@ -5,35 +5,61 @@
...
+            : base(fhirDataStore, searchService, conformanceProvider, resourceWrapperFactory, resourceIdProvider)
+        {
+        }
+
         /// <summary>
         /// Validates if transaction bundle contains multiple entries that are modifying the same resource.
         /// </summary>
         /// <param name="bundle"> The input bundle</param>
-        public static void ValidateTransactionBundle(Hl7.Fhir.Model.Bundle bundle)
+        public async Task ValidateBundle(Hl7.Fhir.Model.Bundle bundle)
```
brendankowitz: EnsureArg

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionValidator.cs
```diff
@@ -5,35 +5,61 @@
...
+
             var resourceIdList = new HashSet<string>(StringComparer.Ordinal);
 
             foreach (var entry in bundle.Entry)
             {
                 if (ShouldValidateBundleEntry(entry))
                 {
-                    string resourceId = GetResourceUrl(entry);
+                    string resourceId = await GetResourceId(entry);
+                    string conditionalCreateQuery = entry.Request.IfNoneExist ?? entry.Request.IfNoneExist;
```
brendankowitz: Null coalesce looks redundant here

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionValidator.cs
```diff
@@ -5,35 +5,61 @@
...
         /// <param name="bundle"> The input bundle</param>
-        public static void ValidateTransactionBundle(Hl7.Fhir.Model.Bundle bundle)
+        public async Task ValidateBundle(Hl7.Fhir.Model.Bundle bundle)
         {
+            if (bundle.Type != BundleType.Transaction)
+            {
+                return;
+            }
+
             var resourceIdList = new HashSet<string>(StringComparer.Ordinal);
```
brendankowitz: Possibly use StringComparer.OrdinalIgnoreCase here

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionValidator.cs
```diff
@@ -42,6 +68,75 @@ public static void ValidateTransactionBundle(Hl7.Fhir.Model.Bundle bundle)
...
         }
 
+        private static string BuildRequestUrlForConditionalQueries(EntryComponent entry, string conditionalCreateQuery)
+        {
+            return conditionalCreateQuery == null ? entry.Request.Url : entry.Request.Url + "?" + conditionalCreateQuery;
+        }
+
+        private async Task<string> GetResourceId(EntryComponent entry)
+        {
+            if (entry.Request.IfNoneExist == null && !entry.Request.Url.Contains("?", StringComparison.InvariantCulture))
```
brendankowitz: string.IsNullOrWhitespace and StringComparison.Ordinal

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionValidator.cs
```diff
@@ -42,6 +68,75 @@ public static void ValidateTransactionBundle(Hl7.Fhir.Model.Bundle bundle)
...
+
+                return entry.Request.Url;
+            }
+            else
+            {
+                string resourceType = null;
+                StringValues conditionalQueries;
+                HTTPVerb requestMethod = (HTTPVerb)entry.Request.Method;
+                bool conditionalCreate = requestMethod == HTTPVerb.POST;
+                bool condtionalUpdate = requestMethod == HTTPVerb.PUT || requestMethod == HTTPVerb.DELETE;
```
brendankowitz: We don't support conditional delete yet

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionValidator.cs
```diff
@@ -42,6 +68,75 @@ public static void ValidateTransactionBundle(Hl7.Fhir.Model.Bundle bundle)
...
+            {
+                throw new RequestNotValidException(string.Format(Api.Resources.InvalidConditionalReferenceParameters, requestUrl));
+            }
+
+            Tuple<string, string>[] conditionalParameters = QueryHelpers.ParseQuery(conditionalQueries)
+                              .SelectMany(query => query.Value, (query, value) => Tuple.Create(query.Key, value)).ToArray();
+
+            var searchResourceRequest = new SearchResourceRequest(resourceType, conditionalParameters);
+
+            return await Search(searchResourceRequest.ResourceType, searchResourceRequest.Queries, CancellationToken.None);
```
brendankowitz: Can CancellationToken be passed through the stack? This might be particularly useful for a bundle as they could take some time

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionValidator.cs
```diff
@@ -5,35 +5,61 @@
...
+using Microsoft.Health.Fhir.Core.Features.Resources;
+using Microsoft.Health.Fhir.Core.Features.Search;
+using Microsoft.Health.Fhir.Core.Messages.Search;
 using Microsoft.Health.Fhir.Core.Models;
 using static Hl7.Fhir.Model.Bundle;
 
 namespace Microsoft.Health.Fhir.Api.Features.Resources.Bundle
 {
-    public static class TransactionValidator
+    public class TransactionValidator : BaseConditionalHandler
```
brendankowitz: This _seems_ like its just validating resource references as in `BundleReferenceValidator`, thoughts?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
poadhika: it runs for transaction only. The method ValidateBundle checks resourceType before proceeding.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
poadhika: Not needed. Forhot to remove it.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionValidator.cs
poadhika: Yes. But was accounting the possiblity that we might support it in future. But perhaps not a good idea for now. I will remove it.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionValidator.cs
poadhika: The validation is only reuired for the transaction bundle and therefore it was name TransactionValidator. I can rename it to TransactionBundleReferenceValidator maybe?

brendankowitz: This seems more descriptive to me, initially I thought it was validating that the transaction actions happened successfully

poadhika: I renamed it to TransactionBundleValidator  because in addition to validating bundlereference, it also validates uniques ogf entries even if the entry doesnot contain reference and also checks if there are entries with operations that are not allow. So that name seemed appropriate,

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Resources/Bundle/BundleHandlerTests.cs
```diff
@@ -192,6 +206,90 @@ public async Task GivenABundle_WhenMultipleRequests_ReturnsABundleResponseWithCo
...
+                   "123",
+                   "1",
+                   "Patient",
+                   new RawResource("data", Core.Models.FhirResourceFormat.Json),
+                   null,
+                   DateTimeOffset.MinValue,
+                   false,
+                   null,
+                   null,
+                   null));
```
brendankowitz: Perhaps extract this to a method so you can use it below

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -127,7 +132,10 @@ public async Task<BundleResponse> Handle(BundleRequest bundleRequest, Cancellati
             var bundleResource = bundleRequest.Bundle.ToPoco<Hl7.Fhir.Model.Bundle>();
             _bundleType = bundleResource.Type;
 
-            await FillRequestLists(bundleResource.Entry);
+            // For resources within a transaction, we need to validate if they are referring to each other and throw an exception in such case.
+            await _transactionBundleValidator.ValidateBundle(bundleResource, cancellationToken);
+
+            await FillRequestLists(bundleResource.Entry, cancellationToken);
```
brendankowitz: I'm now thinking it might make more sense to move both these methods into the if/else, e.g:
```csharp
if (_bundleType == BundleType.Batch)
{
    await FillRequestLists(bundleResource.Entry, cancellationToken);
    // ...
}
else if (_bundleType == BundleType.Transaction)
{
    await _transactionBundleValidator.ValidateBundle(bundleResource, cancellationToken);
    await FillRequestLists(bundleResource.Entry, cancellationToken);
    // ...
}
```

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionBundleValidator.cs
```diff
@@ -0,0 +1,161 @@
...
+        public async Task ValidateBundle(Hl7.Fhir.Model.Bundle bundle, CancellationToken cancellationToken)
+        {
+            var resourceIdList = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+
+            foreach (var entry in bundle.Entry)
+            {
+                if (ShouldValidateBundleEntry(entry))
+                {
+                    string resourceId = await GetResourceId(entry, cancellationToken);
+                    string conditionalCreateQuery = entry.Request.IfNoneExist ?? entry.Request.IfNoneExist;
```
brendankowitz: This looks redundant?
```suggestion
                    string conditionalCreateQuery = entry.Request.IfNoneExist;
```

poadhika: I was assigning the value only if the conditionalParameter existed. But since I am checking null when I am using the variable. It is redundant check and can be removed. I will remove it.

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Resources/Bundle/TransactionBundleValidatorTests.cs
```diff
@@ -0,0 +1,90 @@
...
+        {
+            IFhirDataStore fhirDataStore = Substitute.For<IFhirDataStore>();
+            Lazy<IConformanceProvider> conformanceProvider = Substitute.For<Lazy<IConformanceProvider>>();
+            IResourceWrapperFactory resourceWrapperFactory = Substitute.For<IResourceWrapperFactory>();
+            ResourceIdProvider resourceIdProvider = Substitute.For<ResourceIdProvider>();
+            _transactionBundleValidator = new TransactionBundleValidator(fhirDataStore, conformanceProvider, resourceWrapperFactory, _searchService, resourceIdProvider);
+        }
+
+        [Fact]
+        public async Task GivenABundleWithUniqueResources_TransactionBundleValidatorShouldNotThrowExceptionAsync()
```
brandonpollett: nit: naming convetion

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -251,6 +259,44 @@ private async Task FillRequestLists(List<Hl7.Fhir.Model.Bundle.EntryComponent> b
...
+                    {
+                        string[] queries = reference.Reference.Split("?");
+                        string resourceType = queries[0];
+                        string conditionalQueries = queries[1];
+
+                        SearchResultEntry[] results = await _transactionBundleValidator.GetExistingResourceId(entry.Request.Url, resourceType, conditionalQueries, cancellationToken);
+
+                        int? resultCount = results?.Length;
+
+                        if (!resultCount.HasValue || resultCount != 1)
```
brandonpollett: Could this be simplified to `if (results == null || results.Length != 1)`?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -251,6 +259,44 @@ private async Task FillRequestLists(List<Hl7.Fhir.Model.Bundle.EntryComponent> b
...
+                        {
+                            throw new RequestNotValidException(string.Format(Api.Resources.InvalidConditionalReference, reference.Reference));
+                        }
+
+                        string resourceId = $"{resourceType}/{results[0].Resource.ResourceId}";
+
+                        referenceIdDictionary.Add(reference.Reference, (resourceId, resourceType));
```
brandonpollett: For a `Patient` with id `123` this would insert a tuple of `("Patient/123", "Patient")` right? 

If it ran into the same reference again, wouldn't the first `if` statement set the references to `Patient/Patient/123`? Do we have a test case that would cover this case?

poadhika: Yes, that was the bug. I have now updated the unittest as well to cover that scenario.

src/Microsoft.Health.Fhir.Api/Resources.resx
```diff
@@ -258,7 +258,18 @@
...
\ No newline at end of file
+  <data name="InvalidConditionalReference" xml:space="preserve">
+    <value>Given conditional reference '{0}' does not resolve to a resource.</value>
+    <comment>{0} is the conditional reference parameter.</comment>
+  </data>
+  <data name="InvalidConditionalReferenceParameters" xml:space="preserve">
+    <value>Resource type and query parameter must be present in a given request '{0}'.</value>
+    <comment>{0} is the request url.</comment>
+  </data>
+  <data name="ConditionalOperationNotSelectiveEnough" xml:space="preserve">
```
brandonpollett: Does something similar exist in `Microsoft.Health.Fhir.Core.Resources`?

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Resources/Bundle/BundleHandlerTests.cs
```diff
@@ -192,6 +207,72 @@ public async Task GivenABundle_WhenMultipleRequests_ReturnsABundleResponseWithCo
             Assert.Equal("200", bundleResource.Entry[2].Response.Status);
         }
 
+        [Fact]
+        public async Task GivenATransactionBundleWithConditionalReferences_WhenResolved_ThenReferencesValuesAreUpdatedCorrectly()
```
brandonpollett: Are we testing the `BundleHandler` or the `TransactionBundleValidator` here? Would it make sense to make the `TransactionBundleValidator` mockable instead? Thoughts?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -251,6 +259,42 @@ private async Task FillRequestLists(List<Hl7.Fhir.Model.Bundle.EntryComponent> b
             }
         }
 
+        public async Task ResolveIntraBundleReferences(EntryComponent entry, Dictionary<string, (string resourceId, string resourceType)> referenceIdDictionary, CancellationToken cancellationToken)
```
brandonpollett: nit: we should probably change this name since it's now all references.

src/Microsoft.Health.Fhir.Api/Resources.resx
poadhika: Yes but I was unable to refer that. Therefore added the same in API.

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Resources/Bundle/BundleHandlerTests.cs
poadhika: As per our discussion l am leaving it as is.

PR #770, opened 2019-12-02
Refactored the notification middleware.

jackliums:
## Description
Refactored the code base so that ApiNotificationMiddleware is now being registered in the IoC container so that the instance can be replaced when needed.

Also, cleaned up the code around how the FhirRequestContext is being initialized. In FhirRequestContextMiddleware, the RouteData and Principal will have the correct value anyways because Mvc middle has not executed yet so removed these. Added FhirRequestContextBeforeAuthenticationMiddleware and together with FhirRequestContextAfterAuthenticationMiddleware and FhirRequestContextRouteDataPopulatingFilterAttribute, it tries to initialize the FhirRequestContext as soon as possible.

FhirRequestContextBeforeAuthenticationMiddleware runs before the Authentication middleware and is needed in case the authentication/authorization fails. In this case, the Mvc middleware is not executed, which means the code will not populate anything related to route (audit event, resource type and etc). The logic in this middleware catches that condition and tries to populate these data.

FhirRequestContextAfterAuthenticationMiddleware  runs after the Authenticaiton middleware and will populate the principal.

FhirRequestContextRouteDataPopulatingFilterAttribute runs after Mvc middleware and will populate route when the request passed authentication/authorization check.

We should look at this and refactor some of the code and maybe remove dependencies on the authorization middleware at some point.

## TODO

One thing I started to clean up but end up not included in this PR is around audit. Currently, the executed audit event is logged in both AuditMiddleware and AuditLoggingFilterAttribute. This is again because Authentication middleware runs before Mvc middleware so AuditMiddleware is responsible for logging the case where the request failed with authentication/authorization error whereas AuditLoggingFilterAttribute logs the requests that passed authentication check. Ideally, all of the executed audit log should happen in AuditMiddlware (e.g., if we want to also log latency along with the audit log, which is something the many Azure service does). However, currently, the batch/transaction has dependencies on AuditLoggingFilterAttribute to log child requests. Once that change is checked in, I can revisit this later.

## Related issues
Addresses [AB#70213](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/70213).

## Testing
Unit test, integration test, and manu tests.

### DIFF
--- a/src/Microsoft.Health.Fhir.Api/Features/ApiNotifications/ApiNotificationMiddleware.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/ApiNotifications/ApiNotificationMiddleware.cs
@@ -9,55 +9,58 @@ using System.Threading;
 using System.Threading.Tasks;
 using EnsureThat;
 using MediatR;
-using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Logging;
+using Microsoft.Health.Fhir.Api.Extensions;
 using Microsoft.Health.Fhir.Core;
 using Microsoft.Health.Fhir.Core.Features.Context;
 
 namespace Microsoft.Health.Fhir.Api.Features.ApiNotifications
 {
-    public class ApiNotificationMiddleware
+    public class ApiNotificationMiddleware : IMiddleware
     {
-        private readonly RequestDelegate _next;
         private readonly IFhirRequestContextAccessor _fhirRequestContextAccessor;
         private readonly IMediator _mediator;
         private readonly ILogger<ApiNotificationMiddleware> _logger;
 
         public ApiNotificationMiddleware(
-            RequestDelegate next,
             IFhirRequestContextAccessor fhirRequestContextAccessor,
             IMediator mediator,
             ILogger<ApiNotificationMiddleware> logger)
         {
-            EnsureArg.IsNotNull(next, nameof(next));
             EnsureArg.IsNotNull(fhirRequestContextAccessor, nameof(fhirRequestContextAccessor));
             EnsureArg.IsNotNull(mediator, nameof(mediator));
             EnsureArg.IsNotNull(logger, nameof(logger));
 
-            _next = next;
             _fhirRequestContextAccessor = fhirRequestContextAccessor;
             _mediator = mediator;
             _logger = logger;
         }
 
-        public async Task Invoke(HttpContext context)
+        public async Task InvokeAsync(HttpContext context, RequestDelegate next)
         {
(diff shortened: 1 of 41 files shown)


src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/FhirRequestContextRouteDataPopulatingFilterAttribute.cs
```diff
@@ -67,16 +79,17 @@ public override void OnActionExecuting(ActionExecutingContext context)
...
-                    fhirRequestContext.ResourceType = resourceType?.ToString();
-                }
-            }
+        public override void OnActionExecuted(ActionExecutedContext context)
+        {
+            // The result can either be a FhirResult or an OperationOutcomeResult which both extend BaseActionResult.
+            var result = context.Result as IResourceActionResult;
+
+            _fhirRequestContextAccessor.FhirRequestContext.ResourceType = result?.GetResultTypeName();
+
```
YazanMSFT: It seems like we have two different methods being used in this filter for populating the resource type, is it possible for them to be different within the same service call?

jackliums: It is possible. For example, create/read/update, they operate on a single resource type and is identical for both request and response. However, in the case of search, the incoming request could indicate that it's a search for Patient, but the response is Bundle.

src/Microsoft.Health.Fhir.Api/Features/ApiNotifications/ApiNotificationMiddleware.cs
```diff
@@ -9,55 +9,59 @@
...
             }
 
+            await InvokeInternalAsync(context, next);
+        }
+
+        protected virtual async Task InvokeInternalAsync(HttpContext context, RequestDelegate next)
+        {
             var apiNotification = new ApiResponseNotification();
 
             using (var timer = _logger.BeginTimedScope("ApiNotificationMiddleware") as ActionTimer)
```
brendankowitz: Perhaps nameof(ApiNotificationMiddleware)

src/Microsoft.Health.Fhir.Api/Features/ApiNotifications/ApiNotificationMiddleware.cs
```diff
@@ -9,55 +9,59 @@
...
 
-                await _next(context);
+            if (!context.Request.IsFhirRequest())
+            {
+                // Don't emit events for internal calls.
+                await next(context);
                 return;
             }
 
+            await InvokeInternalAsync(context, next);
```
brendankowitz: Perhaps `InvokeWithNotification`?

jackliums: Renamed.

PR #772, opened 2019-12-04
Runs crucible for sql

rbans96:
## Description
Adds crucible support for SQL

## Related issues
Addresses : User Story 70320

## Testing
Run crucible tests locally
image

### DIFF
--- a/test/Microsoft.Health.Fhir.Shared.Tests.Crucible/CrucibleDataSource.cs
+++ b/test/Microsoft.Health.Fhir.Shared.Tests.Crucible/CrucibleDataSource.cs
@@ -16,9 +16,11 @@ namespace Microsoft.Health.Fhir.Tests.E2E.Crucible
     public class CrucibleDataSource
     {
         private const int PastResultsValidInMinutes = 10;
+        private static DataStore _dataStore;
 
-        public CrucibleDataSource()
+        public CrucibleDataSource(DataStore dataStore)
         {
+            _dataStore = dataStore;
             TestRun = new Lazy<ServerTestRun>(() => GetTestRunAsync().GetAwaiter().GetResult());
         }
 
@@ -50,7 +52,7 @@ namespace Microsoft.Health.Fhir.Tests.E2E.Crucible
             using (var testFhirServerFactory = new TestFhirServerFactory())
             {
                 var fhirClient = testFhirServerFactory
-                    .GetTestFhirServer(DataStore.CosmosDb, null)
+                    .GetTestFhirServer(_dataStore, null)
                     .GetFhirClient(ResourceFormat.Json);
 
                 if (fhirClient.SecuritySettings.SecurityEnabled)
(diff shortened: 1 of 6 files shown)


brendankowitz:
I'm pretty sure you'll need to set something inside `SetTestServerAsync` so that the test server names are something like "PR123_R4_SQL"

brendankowitz:
The crucible server is: https://crucible.mshapis.com/ you should be able to see both variations show up after the E2E tests run

rbans96:
> The crucible server is: https://crucible.mshapis.com/ you should be able to see both variations show up after the E2E tests run

yes I can see that :)
image

test/Microsoft.Health.Fhir.Shared.Tests.Crucible/CrucibleDataSource.cs
```diff
@@ -88,5 +90,10 @@ private async Task<ServerTestRun> GetTestRunAsync()
 
             return new ServerTestRun(crucible.ServerBase, result);
         }
+
+        public static DataStore GetDataStore()
+        {
+            return _dataStore;
+        }
```
brendankowitz: This looks like it should be a property

test/Microsoft.Health.Fhir.Shared.Tests.Crucible/CrucibleTestFixture.cs
```diff
@@ -37,12 +40,14 @@ public void Run(string id)
 
             if (findTest != null)
             {
+                bool isCosmosDb = CrucibleDataSource.GetDataStore().Equals(DataStore.CosmosDb);
```
brendankowitz: In general we prefer to use string.Equals and also pass in the type of string comparison, so for example here: `string.Equals(a, b, StringComparison.OrdinalIgnoreCase)`

test/Microsoft.Health.Fhir.Shared.Tests.Crucible/CrucibleTestFixture.cs
```diff
@@ -37,12 +40,14 @@ public void Run(string id)
...
                 var failures = findTest.Result
                     .Where(x =>
                     {
                         var testName = $"{x.TestId ?? findTest.TestId}/{x.Id}";
-                        return x.Status == "fail" && !KnownCrucibleTests.KnownFailures.Contains(testName) && !KnownCrucibleTests.KnownBroken.Contains(testName)
+                        bool commonFailures = x.Status == "fail" && !KnownCrucibleTests.KnownCommonFailures.Contains(testName) && !KnownCrucibleTests.KnownBroken.Contains(testName)
                                && !x.Message.ToString().Contains(KnownCrucibleTests.BundleCountFilter);
+                        return isCosmosDb ? commonFailures && !KnownCrucibleTests.KnownCosmosDbFailures.Contains(testName) : commonFailures && !KnownCrucibleTests.KnownSqlServerFailures.Contains(testName);
```
brendankowitz: As a suggestion we might be able to simplify by creating a `knownFailures` variable with this logic built-in to use here and below.
e.g. `var knownFailures = commonFailures.Concat(isCosmosDb ? KnownCrucibleTests.KnownCosmosDbFailures : KnownCrucibleTests.KnownSqlServerFailures).ToArray();`

test/Microsoft.Health.Fhir.Shared.Tests.Crucible/CrucibleDataSource.cs
rbans96: Having DataStore as a property, leads to create all other methods as non static and [memberData] should be static so it results in a compilation error

Issue #664, opened 2019-09-23
ValidateResourceIdFilterAttribute returns 'TypeName' in error message location instead of 'Id'


brendankowitz:
**Describe the bug**
ValidateResourceIdFilterAttribute returns 'TypeName' in error message location instead of 'Id' when the Id of the PUT body doesn't match the URL.

**To Reproduce**
1. PUT to /Observation/f002 using a different Id in the payload e.g. f001
2. The validation message that is returned has the text "TypeName" in the location field

**Expected behavior**
The location should either be omitted or contain the relevant field "Observation.id"

**Actual behavior**
The text "TypeName" in the location field, which doesn't make sense

```
{
    "resourceType": "OperationOutcome",
    "id": "c5459154-064b-4bbd-b6bd-081ffb6e8196",
    "issue": [
        {
            "severity": "error",
            "code": "invalid",
            "diagnostics": "Id in the URL must match id in the resource.",
            "location": [
                "TypeName"
            ]
        }
    ]
}
```

PR #781, opened 2019-12-10
Updates location in operationOutcome

rbans96:
## Description
The location updates to contain the relevant field "Observation.id" when Observation object resource Id not matched OR without Id
Fixes the order of not reachable code

## Related issues
Addresses #664

## Testing
Unit tests and manual tests.
image

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Filters/ValidateResourceIdFilterTests.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Filters/ValidateResourceIdFilterTests.cs
@@ -5,6 +5,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using Hl7.Fhir.Model;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
@@ -14,6 +15,7 @@ using Microsoft.AspNetCore.Routing;
 using Microsoft.Health.Fhir.Api.Features.Filters;
 using Microsoft.Health.Fhir.Api.Features.Routing;
 using Microsoft.Health.Fhir.Core.Features.Validation;
+using Microsoft.Health.Fhir.Core.Models;
 using Xunit;
 
 namespace Microsoft.Health.Fhir.Api.UnitTests.Features.Filters
@@ -47,7 +49,8 @@ namespace Microsoft.Health.Fhir.Api.UnitTests.Features.Filters
 
             var context = CreateContext(observation, observation.Id.ToUpper());
 
-            Assert.Throws<ResourceNotValidException>(() => filter.OnActionExecuting(context));
+            var exception = Assert.Throws<ResourceNotValidException>(() => filter.OnActionExecuting(context));
+            Assert.Equal("Observation.id", exception.Issues.First<OperationOutcomeIssue>().Location.First());
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/ValidateResourceIdFilterAttribute.cs
```diff
@@ -24,19 +25,20 @@ public override void OnActionExecuting(ActionExecutingContext context)
             if (context.RouteData.Values.TryGetValue(KnownActionParameterNames.Id, out var actionId) &&
                 context.ActionArguments.TryGetValue(KnownActionParameterNames.Resource, out var parsedModel))
             {
-                if (!string.Equals((string)actionId, ((Resource)parsedModel).Id, StringComparison.Ordinal))
```
rbans96: Moves up Id null check if-block since it is not reachable as this usecase getting handled by id not matched if-block

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Filters/ValidateResourceIdFilterTests.cs
```diff
@@ -75,6 +78,9 @@ public void GivenAnObservationAction_WhenPuttingAnObservationObjectWithoutAnId_T
             var context = CreateContext(observation, Guid.NewGuid().ToString());
 
             Assert.Throws<ResourceNotValidException>(() => filter.OnActionExecuting(context));
+
+            var exception = Assert.Throws<ResourceNotValidException>(() => filter.OnActionExecuting(context));
```
brandonpollett: You should be able to remove the `Assert.Throws<ResourceNotValidException>(() => filter.OnActionExecuting(context));` above since we're doing the same thing in this line and keeping the exception.

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/ValidateResourceIdFilterAttribute.cs
```diff
@@ -24,19 +25,20 @@ public override void OnActionExecuting(ActionExecutingContext context)
...
+                string location = string.Concat(parsedModel.GetType().ToString().Split('.').Last(), ".id");
```
brandonpollett: You might be able to do `context.RouteData.Values.TryGetValue(KnownActionParameterNames.ResourceType, out var resourceType)` to get the respource type.

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/ValidateResourceIdFilterAttribute.cs
```diff
@@ -24,19 +24,21 @@ public override void OnActionExecuting(ActionExecutingContext context)
             if (context.RouteData.Values.TryGetValue(KnownActionParameterNames.Id, out var actionId) &&
                 context.ActionArguments.TryGetValue(KnownActionParameterNames.Resource, out var parsedModel))
             {
-                if (!string.Equals((string)actionId, ((Resource)parsedModel).Id, StringComparison.Ordinal))
+                string location = context.RouteData.Values.TryGetValue(KnownActionParameterNames.ResourceType, out var resourceType) ?
```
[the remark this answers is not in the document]
rbans96: Good point, thanks

PR #782, opened 2019-12-11
Implements SQL export job creation

rotodd:
## Description
 
- Implements the method to create an export job, the first step in adding SQL support for [bulk export](https://microsofthealth.visualstudio.com/Health/_wiki/wikis/Resolute.wiki/25/BulkExportDesignSpec)
- Creates a new `ExportJob` table with the following columns:
  - `Id`: the unique ID of the job
  - `Status`: the status of the job (set to `Queued` during export job creation)
  - `HeartbeatDateTime`: used to check if jobs are stale when acquiring export jobs (set to `null` during export job creation)
  - `QueuedDateTime`: the time the job is created/queued
  - `RawJobRecord`: the raw job record, zipped (similar to the `RawResource` column in `UpsertResource`)
  - `JobVersion`: used to manage concurrency when deleting jobs
- Adds a stored procedure for adding a new job record to the `ExportJob` table
- Updates SQL code generation files to allow the [rowversion](https://docs.microsoft.com/en-us/sql/t-sql/data-types/rowversion-transact-sql?view=sql-server-ver15) SQL data type to be used for the `JobVersion` column in `ExportJob`
 
- Note: merging into a feature branch

sql-export-create

## Related issues
Addresses [#AB71329](https://microsofthealth.visualstudio.com/Health/_workitems/edit/71329).

## Testing

- Enables running the integration test for export job creation with SQL and adds support for operation datastore tests on the SQL side

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
@@ -1273,7 +1273,7 @@ CREATE UNIQUE NONCLUSTERED INDEX IX_ExportJob_Status_HeartbeatDateTime_QueuedDat
     Status,
     HeartbeatDateTime,
     QueuedDateTime
-) -- TODO: Modify indexes as needed when implementing method to acquire export jobs.
+) -- TODO: Modify indexes as needed when implementing remaining sql export methods.
 
 GO
 
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
@@ -1255,9 +1255,9 @@ GO
 **************************************************************/
 CREATE TABLE dbo.ExportJob
 (
-    JobId varchar(64) COLLATE Latin1_General_100_CS_AS NOT NULL,
-    JobStatus varchar(10) NOT NULL,
-    HeartbeatTimeStamp datetimeoffset(7) NULL,
+    Id varchar(64) COLLATE Latin1_General_100_CS_AS NOT NULL,
+    Status varchar(10) NOT NULL,
+    HeartbeatDateTime datetimeoffset(7) NULL,
     QueuedDateTime datetimeoffset(7) NOT NULL,
     RawJobRecord varbinary(max) NOT NULL,
 	JobVersion rowversion NOT NULL
@@ -1265,15 +1265,15 @@ CREATE TABLE dbo.ExportJob
 
 CREATE UNIQUE CLUSTERED INDEX IXC_ExportJob ON dbo.ExportJob
 (
-    JobId
+    Id
 )
 
-CREATE UNIQUE NONCLUSTERED INDEX IX_ExportJob_JobStatus_HeartbeatTimeStamp_QueuedDateTime ON dbo.ExportJob
+CREATE UNIQUE NONCLUSTERED INDEX IX_ExportJob_Status_HeartbeatDateTime_QueuedDateTime ON dbo.ExportJob
 (
-    JobStatus,
-    HeartbeatTimeStamp,
+    Status,
+    HeartbeatDateTime,
     QueuedDateTime
-) -- TODO: Modify this as needed when implementing method to acquire export jobs.
+) -- TODO: Modify indexes as needed when implementing method to acquire export jobs.
 
 GO
 
@@ -1292,7 +1292,6 @@ CREATE SEQUENCE dbo.ResourceSurrogateIdUniquifierSequence
         CACHE 1000000
 GO
 
-
 /*************************************************************
     Stored procedure for exporting
 **************************************************************/
@@ -1304,20 +1303,21 @@ GO
 --     Creates a new row to the ExportJob table, adding a new job to the queue of jobs to be processed.
 --
 -- PARAMETERS
---     @jobId
+--     @id
 --         * The ID of the export job record
---     @jobStatus
+--     @status
 --         * The status of the export job
---     @heartbeatTimeStamp
---         * The time the export job is pinged
---     @queuedDataTime
+--     @queuedDateTime
 --         * The time the export job is queued
 --     @rawJobRecord
 --         * A compressed UTF16-encoded JSON document
+--
+-- RETURN VALUE
+--     The row version of the created export job.
+--
 CREATE PROCEDURE dbo.CreateExportJob
-    @jobId varchar(64),
-    @jobStatus varchar(10),
-    @heartbeatTimeStamp datetimeoffset(7) = NULL,
+    @id varchar(64),
+    @status varchar(10),
     @queuedDateTime datetimeoffset(7),
     @rawJobRecord varbinary(max)
 AS
@@ -1327,9 +1327,9 @@ AS
     BEGIN TRANSACTION
 
     INSERT INTO dbo.ExportJob
-        (JobId, JobStatus, HeartbeatTimeStamp, QueuedDateTime, RawJobRecord)
+        (Id, Status, QueuedDateTime, RawJobRecord)
     VALUES
-        (@jobId, @jobStatus, @heartbeatTimeStamp, @queuedDateTime, @rawJobRecord)
+        (@id, @status, @queuedDateTime, @rawJobRecord)
   
     SELECT CAST(MIN_ACTIVE_ROWVERSION() AS INT)
 
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
@@ -1260,7 +1260,7 @@ CREATE TABLE dbo.ExportJob
     HeartbeatDateTime datetimeoffset(7) NULL,
     QueuedDateTime datetimeoffset(7) NOT NULL,
     RawJobRecord varbinary(max) NOT NULL,
-	JobVersion rowversion NOT NULL
+    JobVersion rowversion NOT NULL
 )
 
 CREATE UNIQUE CLUSTERED INDEX IXC_ExportJob ON dbo.ExportJob
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
@@ -1259,7 +1259,8 @@ CREATE TABLE dbo.ExportJob
     JobStatus varchar(10) NOT NULL,
     HeartbeatTimeStamp datetimeoffset(7) NULL,
     QueuedDateTime datetimeoffset(7) NOT NULL,
-    RawJobRecord varbinary(max) NOT NULL
+    RawJobRecord varbinary(max) NOT NULL,
+	JobVersion rowversion NOT NULL
 )
 
 CREATE UNIQUE CLUSTERED INDEX IXC_ExportJob ON dbo.ExportJob
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
@@ -1331,6 +1331,8 @@ AS
     VALUES
         (@jobId, @jobStatus, @heartbeatTimeStamp, @queuedDateTime, @rawJobRecord)
   
+    SELECT CAST(MIN_ACTIVE_ROWVERSION() AS INT)
+
     COMMIT TRANSACTION
 GO
 
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Model/Column.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Model/Column.cs
@@ -434,9 +434,9 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Schema.Model
         }
     }
 
-    public class RowVersionColumn : Column<byte[]>
+    public class TimestampColumn : Column<byte[]>
     {
-        public RowVersionColumn(string name)
+        public TimestampColumn(string name)
             : base(name, SqlDbType.Timestamp, true)
         {
         }
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Model/Column.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Model/Column.cs
@@ -439,6 +439,8 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Schema.Model
     /// </summary>
     /// <remarks>
     /// The timestamp data type is a synonym for the rowversion data type, and its syntax is now deprecated in SQL.
+    /// The rowversion data type is used in the SQL code, and it is translated to timestamp on the C# side because
+    /// the SqlDbType enum only supports timestamp.
     /// </remarks>
     public class TimestampColumn : Column<byte[]>
     {
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Model/Column.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Model/Column.cs
@@ -455,7 +455,10 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Schema.Model
             const int length = 8;
 
             byte[] bytes = new byte[length];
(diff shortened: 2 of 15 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Model/Column.cs
```diff
@@ -434,6 +434,30 @@ public override void Set(SqlDataRecord record, int ordinal, byte? value)
...
+            reader.GetBytes(Metadata.Name, ordinal, 0, bytes, 0, length);
+
+            return bytes;
+        }
+
+        public override void Set(SqlDataRecord record, int ordinal, byte[] value)
+        {
+            record.SetBytes(ordinal, 0, value, 0, value.Length);
+        }
+    }
```
rotodd: [This](https://docs.microsoft.com/en-us/dotnet/framework/data/adonet/sql-server-data-type-mappings) seems to indicate that both timestamp and rowversion data types are mapped to a byte array.

At the same time, [documentation for rowversion](https://docs.microsoft.com/en-us/sql/t-sql/data-types/rowversion-transact-sql?view=sql-server-ver15) specifies:

`A nonnullable rowversion column is semantically equivalent to a binary(8) column. A nullable rowversion column is semantically equivalent to a varbinary(8) column.`

Should behavior here differ based on if the column is specified as nullable or not?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
```diff
@@ -1250,6 +1250,33 @@ WITH (DATA_COMPRESSION = PAGE)
 
 GO
 
+/*************************************************************
+    Export Job
+**************************************************************/
+CREATE TABLE dbo.ExportJob
```
brandonpollett: Is the plan to move these additions to a `2.sql` before merging the feature branch to master?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Model/Column.cs
```diff
@@ -434,6 +434,30 @@ public override void Set(SqlDataRecord record, int ordinal, byte? value)
...
+    public class TimestampColumn : Column<byte[]>
+    {
+        public TimestampColumn(string name)
+            : base(name, SqlDbType.Timestamp, true)
+        {
+        }
+
+        public override byte[] Read(SqlDataReader reader, int ordinal)
+        {
+            // The row version storage size is 8 bytes.
```
brandonpollett: nit: might want to indicate that this is synonymous with row version in the class documentation.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
```diff
@@ -5,25 +5,80 @@
...
         {
             IReadOnlyCollection<ExportJobOutcome> returnValue = new List<ExportJobOutcome>();
             return Task.FromResult(returnValue);
         }
 
-        public Task<ExportJobOutcome> CreateExportJobAsync(ExportJobRecord jobRecord, CancellationToken cancellationToken)
+        public async Task<ExportJobOutcome> CreateExportJobAsync(ExportJobRecord jobRecord, CancellationToken cancellationToken)
         {
-            throw new NotImplementedException();
+            using (var connection = new SqlConnection(_configuration.ConnectionString))
```
brandonpollett: Not sure if the export can ever be part of a FHIR transaction, but if so the connection will need to be handled differently.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
```diff
@@ -5,25 +5,80 @@
...
+                        command,
+                        jobRecord.Id,
+                        jobRecord.Status.ToString(),
+                        jobRecord.QueuedTime,
+                        stream);
+
+                    int? rowVersion = (int?)await command.ExecuteScalarAsync(cancellationToken);
+
+                    // The row version should nerver be null.
+                    Ensure.That(rowVersion).IsNotNull();
```
brandonpollett: What would this look like when it bubbles up to the user that requested it?

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Features/Operations/FhirOperationDataStoreTests.cs
```diff
@@ -17,16 +17,15 @@
 
 namespace Microsoft.Health.Fhir.Tests.Integration.Features.Operations
 {
-    [Collection(FhirOperationTestConstants.FhirOperationTests)]
```
brandonpollett: Just for my knowledge, why was the collection attribute removed?

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlServerFhirStorageTestHelper.cs
```diff
@@ -19,9 +19,14 @@ public SqlServerFhirStorageTestHelper(string connectionString)
             _connectionString = connectionString;
         }
 
-        public Task DeleteAllExportJobRecordsAsync()
+        public async Task DeleteAllExportJobRecordsAsync()
```
brandonpollett: Can this take a cancellation token?

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlServerFhirStorageTestsFixture.cs
```diff
@@ -33,6 +34,7 @@ public class SqlServerFhirStorageTestsFixture : IServiceProvider, IAsyncLifetime
         private readonly string _masterConnectionString;
         private readonly string _databaseName;
         private readonly IFhirDataStore _fhirDataStore;
+        private IFhirOperationDataStore _fhirOperationDataStore;
```
brandonpollett: Can this be `readonly`?

tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Sql/SqlVisitor.cs
```diff
@@ -218,7 +218,21 @@ private static string GetTableTypeNameWithoutVersionSuffix(SchemaObjectName obje
...
+            if (Enum.TryParse(sqlDbTypeToNormalize, true, out SqlDbType sqlDbType))
+            {
+                normalizedSqlDbType = sqlDbType.ToString();
+            }
+            else
+            {
+                if (string.Compare(sqlDbTypeToNormalize, "RowVersion", StringComparison.InvariantCultureIgnoreCase) == 0)
+                {
+                    // The timestamp data type is a synonym for the rowversion data type. Its syntax is now deprecated in SQL.
+                    normalizedSqlDbType = nameof(SqlDbType.Timestamp);
```
brandonpollett: Would it make more sense to do `SqlDbType.Timestamp.ToString()` since that's essentially what we're doing above?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
rotodd: Yep, I plan to complete [#AB71190](https://microsofthealth.visualstudio.com/Health/_workitems/edit/71190) before merging.

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlServerFhirStorageTestsFixture.cs
rotodd: Right, thanks. Fixed in e789186.

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlServerFhirStorageTestHelper.cs
rotodd: Let me know if e789186 looks ok.

tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Sql/SqlVisitor.cs
rotodd: That makes sense - fixed in 96e92f1.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Model/Column.cs
rotodd: I agree! Added in d6cbf1f.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
rotodd: Ah, good point. I throw `OperationFailedException` in 1014e9d.

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Features/Operations/FhirOperationDataStoreTests.cs
rotodd: I actually didn't intend to take this out, thanks for catching it. Added it back in and included a few more tests in 0002444.

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Features/Operations/FhirOperationDataStoreTests.cs
```diff
@@ -18,23 +18,25 @@
...
+        private readonly CancellationToken _cancellationToken;
 
         private readonly CreateExportRequest _exportRequest = new CreateExportRequest(new Uri("http://localhost/ExportJob"), "destinationType", "destinationConnection");
 
-        public CosmosFhirOperationDataStoreTests(FhirStorageTestsFixture fixture)
+        public FhirOperationDataStoreTests(FhirStorageTestsFixture fixture)
         {
             _operationDataStore = fixture.OperationDataStore;
             _testHelper = fixture.TestHelper;
+            _cancellationToken = new CancellationTokenSource().Token;
```
brandonpollett: We would probably be better off storing the `CancellationTokenSource` in a field so that we could call `CancellationTokenSource.Cancel()`. 

In looking at the code, we may just want to update the `DeleteAllExportJobRecordsAsync` method to have an optional cancellation token parameter (e.g. `CancellationToken cancellationToken = default` and not actually pass it in from here since we're never actually cancelling the task.

rotodd: Sounds good, made the cancellation token an optional parameter in 0cabf1a.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
rotodd: @namadabu @jackliums what are your thoughts on having export available as part of a FHIR transaction?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Model/Column.cs
```diff
@@ -434,6 +434,36 @@ public override void Set(SqlDataRecord record, int ordinal, byte? value)
...
+        {
+        }
+
+        public override byte[] Read(SqlDataReader reader, int ordinal)
+        {
+            // The rowversion storage size is 8 bytes.
+            const int length = 8;
+
+            byte[] bytes = new byte[length];
+            reader.GetBytes(Metadata.Name, ordinal, 0, bytes, 0, length);
```
jackliums: can we add named parameter to these 0s? Makes reading the code without IDE a bit easier.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
jackliums: The spec specific calls out that operations are supported within batch/transaction so I suppose that it should be allowed. But it does make reference detection harder though. Should ask the FHIR community.

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlServerFhirStorageTestHelper.cs
```diff
@@ -19,9 +20,14 @@ public SqlServerFhirStorageTestHelper(string connectionString)
...
         }
 
-        public Task DeleteAllExportJobRecordsAsync()
+        public async Task DeleteAllExportJobRecordsAsync(CancellationToken cancellationToken = default)
         {
-            throw new System.NotImplementedException();
+            using (var connection = new SqlConnection(_connectionString))
+            {
+                var command = new SqlCommand("DELETE FROM dbo.ExportJob", connection);
+                command.Connection.Open();
```
jackliums: OpenAsync?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Model/Column.cs
```diff
@@ -434,6 +434,36 @@ public override void Set(SqlDataRecord record, int ordinal, byte? value)
         }
     }
 
+    /// <summary>
+    /// Represents a rowversion type column.
+    /// </summary>
+    /// <remarks>
+    /// The timestamp data type is a synonym for the rowversion data type, and its syntax is now deprecated in SQL.
```
jackliums: Can you explain a little bit about what you meant that it's now deprecated in SQL? We should not be using anything that's deprecated.

We don't have to use rowversion. We could alternatively just generate automatically incremented count or something.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Model/Column.cs
rotodd: Good idea, thanks. Added in be0c10f and af5964a.

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlServerFhirStorageTestHelper.cs
rotodd: Right thanks - fixed in 18fa292.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Model/Column.cs
rotodd: Sure. The first "note" in the [rowversion documentation](https://docs.microsoft.com/en-us/sql/t-sql/data-types/rowversion-transact-sql?view=sql-server-ver15) indicates that the timestamp syntax is deprecated.

I used the rowversion type in 1.sql, but I ran into issues when this was translated into C# objects because the [SqlDbType enum](https://docs.microsoft.com/en-us/dotnet/api/system.data.sqldbtype?view=netframework-4.8) does not support RowVersion (only Timestamp).

I added some code in SqlVisitor.cs that translates `RowVersion` to `Timestamp`, which causes the name of the generated column object to be `TimestampColumn` (instead of `RowVersionColumn`). 

I updated the documentation to make this a bit more clear in b7efe7d - but if you have suggestions on how to improve this, I'm all ears.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
rotodd: Created a post [here](http://community.fhir.org/t/bulk-export-and-fhir-transactions/1688) and [#AB71716](https://microsofthealth.visualstudio.com/Health/_workitems/edit/71716) to look into this further.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Model/Column.cs
rotodd: @johnstairs I believe you would the best person to answer this. Going to merge this PR to my feature branch for now, but this can be revisited after the holidays if needed.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Model/Column.cs
rotodd: Similar to above, I'm going to merge this PR to my feature branch for now, but this can be revisited after the holidays if needed.

PR #785, opened 2019-12-11
Add check for reference type

brandonpollett:
## Description
Currently if you `POST` a transaction with a reference that is an `Identifier` as opposed to a `Reference` a `NullRefException` is thrown.

## Related issues
Addresses [AB#71488](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/71488).

## Testing
Added unit tests. Manual E2E test verifies behavior.

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Resources/Bundle/BundleHandlerTests.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Resources/Bundle/BundleHandlerTests.cs
@@ -209,6 +209,44 @@ namespace Microsoft.Health.Fhir.Api.UnitTests.Features.Resources.Bundle
         }
 
         [Fact]
+        public async Task GivenATransactionBundleWithIdentifierReferences_WhenResolved_ThenReferencesValuesAreNotUpdated()
+        {
+            var observation = new Observation
+            {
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Resources/Bundle/BundleHandlerTests.cs
```diff
@@ -223,7 +261,7 @@ public async Task GivenATransactionBundleWithConditionalReferences_WhenResolved_
 
             foreach (var entry in bundle.Entry)
             {
-                IEnumerable<ResourceReference> references = entry.Resource.GetAllChildren<ResourceReference>();
+                List<ResourceReference> references = entry.Resource.GetAllChildren<ResourceReference>().ToList();
```
rbans96: Any specific reason to convert to List since we just intent to iterate over it? Just curious for my knowledge.

brandonpollett: To avoid possible multiple enumerations - https://www.jetbrains.com/help/resharper/PossibleMultipleEnumeration.html

PR #792, opened 2019-12-17
Adds profile under resource

rbans96:
## Description
Adds Profile for each individual resource

## Related issues
Addresses #724

## Testing
Unit tests and manual test

image

PR #793, opened 2019-12-18
Add $validate endpoint

LTA-Thinking:
## Description
Added the $validate endpoint to allow validation of FHIR resources without persisting them to the database. The endpoint is controlled through a flag that is set to be enabled by default.
FHIR spec reference: http://hl7.org/fhir/resource-operation-validate.html

## Testing
Unit and E2E tests were added for the new endpoint. Manual testing with Postman was also performed.


src/Microsoft.Health.Fhir.Shared.Api/Controllers/ValidateController.cs
```diff
@@ -0,0 +1,70 @@
...
+            _coreFeatures = coreFeatures.Value;
+        }
+
+        [HttpPost]
+        [Route(KnownRoutes.ValidateResourceType)]
+        [AuditEventType(AuditEventSubType.Read)]
+        [Authorize(PolicyNames.ReadPolicy)]
+        public async Task<IActionResult> Validate([FromBody] Resource resource)
+        {
+            if (!_coreFeatures.SupportsValidate || resource.ResourceType == ResourceType.Parameters)
```
deepak-ms: what's special about Parameters resourceType?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ValidateTests.cs
```diff
@@ -0,0 +1,111 @@
...
+
+        [Theory]
+        [InlineData(
+            "Patient/$validate",
+            "{\"resourceType\":\"Patient\",\"name\":\"test, one\"}",
+            "Type checking the data: Since type HumanName is not a primitive, it cannot have a value (at Resource.name[0])")]
+        [InlineData(
+            "Observation/$validate",
+            "{\"resourceType\":\"Observation\",\"code\":{\"coding\":[{\"system\":\"system\",\"code\":\"code\"}]}}",
+            "Element with min. cardinality 1 cannot be null")]
```
deepak-ms: which element?

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ValidateController.cs
LTA-Thinking: Parameters is a wrapper resource that can be put around other resources, but isn't persisted. There is a future story to support it for this endpoint, but it requires extra work. Here is the reference for it: https://www.hl7.org/fhir/parameters.html

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ValidateTests.cs
LTA-Thinking: The name/path is in another parameter of the returned operation outcome. I added another test to cover this case.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ValidateController.cs
```diff
@@ -0,0 +1,70 @@
...
+            {
+                var outcome = new OperationOutcome();
```
brandonpollett: Should just be able to throw an exception here and we've got code that will catch it and correctly setup the `OperationOutcome`. See `OperationOutcomeExceptionFilterAttribute`

src/Microsoft.Health.Fhir.Core/Features/Operations/Validate/ValidateOperationHandler.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    public class ValidateOperationHandler : IRequestHandler<ValidateOperationRequest, ValidateOperationResponse>
+    {
+        public Task<ValidateOperationResponse> Handle(ValidateOperationRequest request, CancellationToken cancellationToken)
+        {
+            return Task.FromResult(new ValidateOperationResponse(
+                new OperationOutcomeIssue(
+                    OperationOutcomeConstants.IssueSeverity.Information,
+                    OperationOutcomeConstants.IssueType.Informational,
+                    "All OK")));
```
brandonpollett: Text should be in a resource file.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ValidateController.cs
brandonpollett: There also seem to be some parameters for `$validate` that we don't support, so we may want to also let the user know of those - https://www.hl7.org/fhir/resource-operation-validate.html

src/Microsoft.Health.Fhir.Core/Configs/CoreFeatureConfiguration.cs
```diff
@@ -19,5 +19,10 @@ public class CoreFeatureConfiguration
         /// Gets or sets a value indicating whether Transaction is enabled or not.
         /// </summary>
         public bool SupportsTransaction { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether Validate is enabled or not.
+        /// </summary>
+        public bool SupportsValidate { get; set; }
```
brandonpollett: I'm not sure if there's plan to expose this via the `/metadata` endpoint, but if not, then this setting could probably just move to the `features` settings property versus the `coreFeatures`

LTA-Thinking: I was unsure about what the difference between the two was. I put it in here since it was in the Core project's feature folder. I'll move it if it makes more sense elsewhere.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ValidateController.cs
LTA-Thinking: Changed to throw an OperationNotImplementedException

src/Microsoft.Health.Fhir.Core/Features/Operations/Validate/ValidateOperationHandler.cs
LTA-Thinking: Moved

src/Microsoft.Health.Fhir.Core/Configs/CoreFeatureConfiguration.cs
LTA-Thinking: Flag moved

src/Microsoft.Health.Fhir.Api/Features/Routing/KnownRoutes.cs
```diff
@@ -33,6 +33,10 @@ internal class KnownRoutes
         public const string ExportResourceTypeById = ResourceTypeById + "/" + Export;
         public const string ExportJobLocation = OperationsConstants.Operations + "/" + OperationsConstants.Export + "/" + IdRouteSegment;
 
+        public const string Validate = "$validate";
+        public const string ValidateResourceType = ResourceType + "/" + Validate;
+        public const string ValidateResourceTypeById = ResourceTypeById + "/" + Validate;
```
namadabu: The `ValidateResourceTypeById` route does not seem to be supported currently. Should we support that also?

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ValidateController.cs
```diff
@@ -0,0 +1,62 @@
...
+    [ValidateModelState]
+    [Authorize(PolicyNames.FhirPolicy)]
+    public class ValidateController : Controller
+    {
+        private readonly IMediator _mediator;
+        private readonly FeatureConfiguration _features;
+
+        public ValidateController(IMediator mediator, IOptions<FeatureConfiguration> features)
+        {
+            _mediator = mediator;
```
namadabu: Null checks?

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Controllers/ValidateControllerTests.cs
```diff
@@ -0,0 +1,75 @@
...
+    public class ValidateControllerTests
+    {
+        private ValidateController _validateController;
+        private IMediator _mediator = Substitute.For<IMediator>();
+
+        public ValidateControllerTests()
+        {
+            _validateController = GetController(true);
+        }
+
```
namadabu: nit: maybe you can just use `Assert.ThrowsAsync()` here? Something like `OperationNotImplementedException ex = await Assert.ThrowsAsync<OperationNotImplementedException>(() => disabledValidateController.Validate(payload));`

test/Microsoft.Health.Fhir.Stu3.Tests.E2E/KnownCrucibleTests.cs
```diff
@@ -39,18 +39,13 @@ public class KnownCrucibleTests
             "connectathon-15-organization-locate-endpoint-xml/02-OrganizationSearchName",
             "connectathon-15-patient-base-client-id-json/Step4-PatientSearch",
             "connectathon-15-patient-base-client-id-xml/Step4-PatientSearch",
```
namadabu: Should this file be part of the current change?

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Controllers/ValidateControllerTests.cs
```diff
@@ -0,0 +1,75 @@
...
+        [Fact]
+        public async void GivenAValidateRequest_WhenTheServerDoesNotSupportValidate_ThenANotSupportedErrorIsReturned()
+        {
+            var disabledValidateController = GetController(false);
+            var payload = new Observation();
+
+            try
```
namadabu: nit: maybe you can just use `Assert.ThrowsAsync()` here? Something like `OperationNotImplementedException ex = await Assert.ThrowsAsync<OperationNotImplementedException>(() => disabledValidateController.Validate(payload));`

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Controllers/ValidateControllerTests.cs
```diff
@@ -0,0 +1,75 @@
...
+using Microsoft.Health.Fhir.Api.Configs;
+using Microsoft.Health.Fhir.Api.Controllers;
+using Microsoft.Health.Fhir.Core.Extensions;
+using Microsoft.Health.Fhir.Core.Features.Operations;
+using NSubstitute;
+using Xunit;
+
+namespace Microsoft.Health.Fhir.Api.UnitTests.Controllers
+{
+    public class ValidateControllerTests
```
namadabu: Should we add another UT to test the `ResourceType.Parameters` code path?

test/Microsoft.Health.Fhir.Stu3.Tests.E2E/KnownCrucibleTests.cs
LTA-Thinking: It was causing e2e test failures, so I updated it to fix them. Some of these tests seem to be a bit flaky, or I don't fully understand what it going on with them.

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Controllers/ValidateControllerTests.cs
LTA-Thinking: I didn't add a test for that because adding support for it is the next story. I can add one if there is value, but it would be removed in a week or two when support is added.

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Controllers/ValidateControllerTests.cs
LTA-Thinking: Good idea, replaced.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ValidateController.cs
LTA-Thinking: Added

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Controllers/ValidateControllerTests.cs
LTA-Thinking: Added test

src/Microsoft.Health.Fhir.Api/Features/Routing/KnownRoutes.cs
LTA-Thinking: Added

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ValidateController.cs
LTA-Thinking: Added exception responses if the user passes a mode or profile.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ValidateController.cs
namadabu: I think it will be better to do an `EnsureArg.IsNotNull(features?.Value, nameof(features))` to make sure both are not null.

LTA-Thinking: I don't know why that isn't showing here. In my local view there are EnsureArg.IsNotNulls for both of them and VS is telling me I have no uncommitted changes.

LTA-Thinking: For some reason GitHub wasn't showing all the commits by default. At the top if you select 'Changes from' and pick all commits you should see it.

namadabu: I wasn't referring to the `mediator` null check actually. I was referring to `features.` It would be good to check whether `features.Value` is also not null (apart from checking if `features` is not null) since that is the actual `FeatureConfiguration` object we use. Sorry for the confusion.

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/OperationOutcomeExceptionFilterAttribute.cs
```diff
@@ -79,6 +80,19 @@ public override void OnActionExecuted(ActionExecutedContext context)
                         operationOutcomeResult.StatusCode = HttpStatusCode.ServiceUnavailable;
                         break;
                     case ResourceNotValidException _:
+                        if (context.ActionDescriptor is ControllerActionDescriptor)
+                        {
+                            var controllerDescriptor = (ControllerActionDescriptor)context.ActionDescriptor;
+                            if (controllerDescriptor.ControllerTypeInfo.Name == "ValidateController")
```
namadabu: I am not familiar with the `ControllerActionDescriptor` class, but can we use `ControllerName` here instead of `ControllerTypeInfo.Name`

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ValidateController.cs
LTA-Thinking: No, my bad. I didn't see the checks and then quick read your comment. That is a good idea to check, I'll add that.

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Controllers/ValidateControllerTests.cs
```diff
@@ -0,0 +1,117 @@
...
+    {
+        [Fact]
+        public async void GivenAValidateRequest_WhenTheServerDoesNotSupportValidate_ThenANotSupportedErrorIsReturned()
+        {
+            ValidateController disabledValidateController = GetController(false);
+            Resource payload = new Observation();
+
+            OperationNotImplementedException ex = await Assert.ThrowsAsync<OperationNotImplementedException>(() => disabledValidateController.Validate(payload, null, null));
```
namadabu: nit: For `.Validate(payload, null, null)`, it would be nicer to include the parameter names when passing in non-descriptive values (such as null, 0, etc). Makes it easier to read the code.

LTA-Thinking: Changed to use named variables since when the method is put into a Func it losses the names of its parameters.

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/OperationOutcomeExceptionFilterAttribute.cs
LTA-Thinking: Good catch, changed.

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/OperationOutcomeExceptionFilterAttribute.cs
```diff
@@ -79,6 +80,19 @@ public override void OnActionExecuted(ActionExecutedContext context)
                         operationOutcomeResult.StatusCode = HttpStatusCode.ServiceUnavailable;
                         break;
                     case ResourceNotValidException _:
+                        if (context.ActionDescriptor is ControllerActionDescriptor)
+                        {
+                            var controllerDescriptor = (ControllerActionDescriptor)context.ActionDescriptor;
+                            if (controllerDescriptor.ControllerName == "Validate")
```
brandonpollett: Should probably use a `string.Equals` with a `StringComparison`

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/OperationOutcomeExceptionFilterAttribute.cs
```diff
@@ -79,6 +80,19 @@ public override void OnActionExecuted(ActionExecutedContext context)
                         operationOutcomeResult.StatusCode = HttpStatusCode.ServiceUnavailable;
                         break;
                     case ResourceNotValidException _:
+                        if (context.ActionDescriptor is ControllerActionDescriptor)
```
brandonpollett: I believe this can be simplified to `if (context.ActionDescriptor is ControllerActionDescriptor controllerDescriptor)`

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ValidateTests.cs
```diff
@@ -0,0 +1,185 @@
...
+    [Trait(Traits.Category, Categories.Validate)]
+    [HttpIntegrationFixtureArgumentSets(DataStore.All, Format.All)]
+    public class ValidateTests : IClassFixture<HttpIntegrationTestFixture>
+    {
+        private readonly HttpClient _client;
+        private readonly JsonSerializer _serializer;
+
+        public ValidateTests(HttpIntegrationTestFixture fixture)
+        {
+            _client = fixture.HttpClient;
```
brandonpollett: Any particular reason we aren't using the `FhirClient` that's exposed test fixture?

src/Microsoft.Health.Fhir.Core/Resources.resx
```diff
@@ -389,6 +389,9 @@
   <data name="UserRequestedCancellation" xml:space="preserve">
     <value>User requested cancellation of operation.</value>
   </data>
+  <data name="ValidationPassed" xml:space="preserve">
+    <value>All OK</value>
```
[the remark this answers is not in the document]
LTA-Thinking: This one is copied from the spec.

src/Microsoft.Health.Fhir.Api/Resources.resx
```diff
@@ -279,4 +279,16 @@
     <value>Resource type '{0}' in the reference '{1}' is not supported.</value>
     <comment>{0} is the resource type. {1} is the reference.</comment>
   </data>
+  <data name="ValidateWithParametersNotSupported" xml:space="preserve">
+    <value>Resources of type Parameters are not supported by this endpoint.</value>
+  </data>
+  <data name="ValidationNotSupported" xml:space="preserve">
+    <value>$validate is not a supported endpoint.</value>
```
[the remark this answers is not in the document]
LTA-Thinking: Added

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/OperationOutcomeExceptionFilterAttribute.cs
LTA-Thinking: Agreed, changed

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/OperationOutcomeExceptionFilterAttribute.cs
LTA-Thinking: Changed

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ValidateTests.cs
LTA-Thinking: Mostly because I didn't know about it. Tests updated to use the FhirClient and a new method added to the client for validate requests.

test/Microsoft.Health.Fhir.Shared.Tests.E2E.Common/FhirClient.cs
```diff
@@ -314,6 +314,20 @@ public async Task<FhirResponse<Bundle>> PostBundleAsync(Resource bundle)
             return await CreateResponseAsync<Bundle>(response);
         }
 
+        public async Task<OperationOutcome> ValidateAsync(string uri, string resource)
+        {
+            // resource is a string instead of a Resource object because the validate endpoint is frequently sent invalid resources that couldn't be parsed.
```
brandonpollett: Probably makes more sense as method documentation.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ValidateTests.cs
```diff
@@ -0,0 +1,155 @@
...
+
+using Hl7.Fhir.Model;
+using Microsoft.Health.Fhir.Tests.Common;
+using Microsoft.Health.Fhir.Tests.Common.FixtureParameters;
+using Xunit;
+
+namespace Microsoft.Health.Fhir.Tests.E2E.Rest
+{
+    [Trait(Traits.Category, Categories.Validate)]
+    [HttpIntegrationFixtureArgumentSets(DataStore.All, Format.Json)]
```
brandonpollett: We might want to have at least a single test that will test the XML code path if validate is also enabled for it.

LTA-Thinking: Would that involve making another file for XML tests, or can I apply that trait attribute to individual tests?

brandonpollett: You can apply the trait to individual tests. See `WhenVersionsEndpointIsCalledWithXml_GivenAValidXmlAcceptHeaderIsProvided_ThenServerShouldReturnOK` for an example.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ValidateController.cs
```diff
@@ -0,0 +1,93 @@
...
+            {
+                throw new OperationNotImplementedException(Resources.ValidationNotSupported);
+            }
+
+            if (resource.ResourceType == ResourceType.Parameters)
+            {
+                throw new OperationNotImplementedException(Resources.ValidateWithParametersNotSupported);
+            }
+
+            if (profile != null)
```
brendankowitz: string.IsNullOrEmpty

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ValidateTests.cs
LTA-Thinking: Test added

src/Microsoft.Health.Fhir.Core/Features/Conformance/Models/ReferenceComponent.cs
```diff
@@ -0,0 +1,17 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Fhir.Core.Models
+{
+    public class ReferenceComponent
```
brendankowitz: Yep, we should wait and use the model in https://github.com/microsoft/fhir-server/pull/792

LTA-Thinking: Done

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ValidateController.cs
LTA-Thinking: Fixed

test/Microsoft.Health.Fhir.Shared.Tests.E2E.Common/FhirClient.cs
LTA-Thinking: Changed

test/Microsoft.Health.Fhir.Shared.Tests.E2E.Common/FhirClient.cs
```diff
@@ -315,6 +315,26 @@ public async Task<FhirResponse<Bundle>> PostBundleAsync(Resource bundle)
             return await CreateResponseAsync<Bundle>(response);
         }
 
+        /**
```
brandonpollett: Weird format for comments, should just be `///` on each line

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ValidateTests.cs
```diff
@@ -0,0 +1,170 @@
...
+        public async void GivenAValidateRequest_WhenTheResourceIsValid_ThenAnOkMessageIsReturned(string path, string payload)
+        {
+            OperationOutcome outcome = await _client.ValidateAsync(path, payload);
+
+            Assert.Single(outcome.Issue);
+            CheckOperationOutcomeIssue(
+                outcome.Issue[0],
+                OperationOutcome.IssueSeverity.Information,
+                OperationOutcome.IssueType.Informational,
+                "All OK");
```
brandonpollett: Maybe a private string const since it's used in multiple places in this file?

src/Microsoft.Health.Fhir.Core/Features/KnownQueryParameterNames.cs
```diff
@@ -10,44 +10,48 @@ namespace Microsoft.Health.Fhir.Core.Features
...
 
-        public const string DestinationType = "_destinationType";
+        /// <summary>
+        /// The pretty query parameter.
+        /// </summary>
+        /// <remarks>True if the client wishes to request for pretty-printed resources (either in JSON or XML), false otherwise.</remarks>
+        public const string Pretty = "_pretty";
 
-        public const string DestinationConnectionSettings = "_destinationConnectionSettings";
+        public const string Profile = "profile";
```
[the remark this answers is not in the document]
LTA-Thinking: Unfortunately this is also a specified query parameter: 
https://www.hl7.org/fhir/resource-operation-validate.html#examples

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Controllers/ValidateControllerTests.cs
```diff
@@ -0,0 +1,123 @@
...
+using Microsoft.Health.Fhir.Api.Controllers;
+using Microsoft.Health.Fhir.Core.Extensions;
+using Microsoft.Health.Fhir.Core.Features.Operations;
+using Microsoft.Health.Fhir.Core.Models;
+using NSubstitute;
+using Xunit;
+
+namespace Microsoft.Health.Fhir.Api.UnitTests.Controllers
+{
+    public class ValidateControllerTests
```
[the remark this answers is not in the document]
LTA-Thinking: We can't simulate a happy path request in the unit tests since the mediator is not mocked in a useful way. The happy path is tested through the E2E tests.

src/Microsoft.Health.Fhir.Api/Resources.resx
```diff
@@ -279,6 +279,19 @@
     <value>Resource type '{0}' in the reference '{1}' is not supported.</value>
     <comment>{0} is the resource type. {1} is the reference.</comment>
   </data>
+  <data name="ValidateWithParametersNotSupported" xml:space="preserve">
+    <value>Resources of type Parameters are not supported by this endpoint.</value>
```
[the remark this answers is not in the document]
LTA-Thinking: Changed

src/Microsoft.Health.Fhir.Core/Features/Operations/Validate/ValidateOperationHandler.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Threading.Tasks;
+using MediatR;
+using Microsoft.Health.Fhir.Core.Messages.Operation;
+using Microsoft.Health.Fhir.Core.Models;
+
+namespace Microsoft.Health.Fhir.Core.Features.Validation
+{
+    public class ValidateOperationHandler : IRequestHandler<ValidateOperationRequest, ValidateOperationResponse>
+    {
+        public Task<ValidateOperationResponse> Handle(ValidateOperationRequest request, CancellationToken cancellationToken)
```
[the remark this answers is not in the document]
LTA-Thinking: Added

src/Microsoft.Health.Fhir.Core/Resources.resx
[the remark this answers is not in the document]
LTA-Thinking: Added

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/OperationOutcomeExceptionFilterAttribute.cs
```diff
@@ -80,6 +81,17 @@ public override void OnActionExecuted(ActionExecutedContext context)
                         operationOutcomeResult.StatusCode = HttpStatusCode.ServiceUnavailable;
                         break;
                     case ResourceNotValidException _:
+                        if (context.ActionDescriptor is ControllerActionDescriptor controllerDescriptor)
+                        {
+                            if (controllerDescriptor.ControllerName.Equals("Validate", StringComparison.OrdinalIgnoreCase))
```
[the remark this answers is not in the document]
LTA-Thinking: Moved

test/Microsoft.Health.Fhir.Shared.Tests.E2E.Common/FhirClient.cs
LTA-Thinking: This is interesting to me. /** is Java syntax for documentation comments. In the docs it says C# supports this syntax (as an equivalent to ///), but no one uses it. Switched to ///.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ValidateTests.cs
LTA-Thinking: Changed

src/Microsoft.Health.Fhir.Core/Features/KnownQueryParameterNames.cs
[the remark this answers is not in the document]
brandonpollett: Yea, that's why I suggested naming (the variable name) it something specific for validation since it would probably make more sense to have the more broad variable be `Profile = "_profile"`.

PR #796, opened 2019-12-19
Add logging for 403 in batch/transaction operations.

brandonpollett:
## Description
Updates to allow the batch/transaction operations to correctly audit a 403 response. Previously the value was only audited in the middleware, but with this change it is also logged when setting the 403 status code for the sub-requests.

Additionally, some refactoring of the batch E2E tests along with changes to make them less flakey.

## Related issues
Addresses [AB#71338](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/71338).

## Testing
Added/updated unit tests and E2E tests.


src/Microsoft.Health.Fhir.Api/Features/Audit/AuditLoggingFilterAttribute.cs
```diff
@@ -40,7 +41,11 @@ public override void OnResultExecuted(ResultExecutedContext context)
         {
             EnsureArg.IsNotNull(context, nameof(context));
 
-            _auditHelper.LogExecuted(context.HttpContext, _claimsExtractor);
+            // The status code of 403 is only ever encountered here in the case of a failed transaction and will be correctly audited in the middleware.
```
brandonpollett: Comment should also mention batch or just switch to bundle operations?

src/Microsoft.Health.Fhir.Shared.Web/DevelopmentIdentityProviderRegistrationExtensions.cs
```diff
@@ -188,6 +188,7 @@ public override void Load()
                     Data["DevelopmentIdentityProvider:Enabled"] = bool.TrueString;
                     Data["FhirServer:Security:Authentication:Audience"] = DevelopmentIdentityProviderConfiguration.Audience;
                     Data["FhirServer:Security:PrincipalClaims:0"] = DevelopmentIdentityProviderConfiguration.LastModifiedClaim;
+                    Data["FhirServer:Security:PrincipalClaims:0"] = DevelopmentIdentityProviderConfiguration.ClientIdClaim;
```
brandonpollett: index should be 1

src/Microsoft.Health.Fhir.Api/Features/Bundle/BundleAwareJwtBearerHandler.cs
```diff
@@ -22,13 +24,28 @@ namespace Microsoft.Health.Fhir.Api.Features.Bundle
...
+        private readonly IClaimsExtractor _claimsExtractor;
 
-        public BundleAwareJwtBearerHandler(IOptionsMonitor<JwtBearerOptions> options, ILoggerFactory logger, UrlEncoder encoder, IDataProtectionProvider dataProtection, ISystemClock clock, IBundleHttpContextAccessor bundleHttpContextAccessor)
+        public BundleAwareJwtBearerHandler(
+            IOptionsMonitor<JwtBearerOptions> options,
+            ILoggerFactory logger,
+            UrlEncoder encoder,
+            IDataProtectionProvider dataProtection,
+            ISystemClock clock,
+            IBundleHttpContextAccessor
```
jackliums: wrong new line?

src/Microsoft.Health.Fhir.Api/Features/Bundle/BundleAwareJwtBearerHandler.cs
```diff
@@ -41,6 +58,7 @@ protected override async Task HandleForbiddenAsync(AuthenticationProperties prop
             if (_bundleHttpContextAccessor.HttpContext != null)
             {
                 _bundleHttpContextAccessor.HttpContext.Response.StatusCode = 403;
+                _auditHelper.LogExecuted(_bundleHttpContextAccessor.HttpContext, _claimsExtractor);
```
jackliums: When would this be logged?

jackliums:
I think we should really look at refactoring how we handle the request with batch/transaction in mind now.

Now the audit is being logged in 3 different places depending on the reuqest code path, which could be fragile and hard to maintain long term.

Can we file user story to investigate what ideally the system should look like and what refactoring is needed and prioritize that?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Audit/AuditTests.cs
```diff
@@ -383,6 +383,95 @@ public async Task GivenAResource_WhenNotAuthorized_ThenAuditLogEntriesShouldBeCr
                 expectedAppId: TestApplications.NativeClient.ClientId);
         }
 
+        [HttpIntegrationFixtureArgumentSets(DataStore.All, Format.Json)]
+        [Fact]
+        [Trait(Traits.Priority, Priority.One)]
+        [Trait(Traits.Category, Categories.Batch)]
+        public async Task GivenABatch_WhenPost_ThenAuditLogEntriesShouldBeCreated()
```
jackliums: I thought there was a reason why the test was split into R4 and Stu3 because there was some difference between the two?

jackliums: Oh I see, it's the http status code that's different?

src/Microsoft.Health.Fhir.Api/Features/Bundle/BundleAwareJwtBearerHandler.cs
brandonpollett: Yes, this is triggered each time a sub-request encounters the `403`. The code added to the `AuditLoggingFilterAttribute` is used filter out the audit logging in the case of a `403`.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Audit/AuditTests.cs
brandonpollett: Yea, once I fixed the differences between the between the sample batches the only differences were the status codes in the case of an `IfMatch` failure.

PR #799, opened 2019-12-26
Adds operationOutcome for type not present

rbans96:
## Description
Adds specific operationOutcome diagnostic for type not present

## Related issues
Addresses Bug [AB#71574](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/71574)

## Testing
Manual testing
image

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Resources.resx
+++ b/src/Microsoft.Health.Fhir.Core/Resources.resx
@@ -358,6 +358,9 @@
   <data name="SortNotSupported" xml:space="preserve">
     <value>The _sort parameter is not supported.</value>
   </data>
+  <data name="TypeNotPresent" xml:space="preserve">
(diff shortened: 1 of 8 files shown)


src/Microsoft.Health.Fhir.Core/Resources.resx
```diff
@@ -358,6 +358,9 @@
   <data name="SortNotSupported" xml:space="preserve">
     <value>The _sort parameter is not supported.</value>
   </data>
+  <data name="TypeNotPresent" xml:space="preserve">
+    <value>The type is not present.</value>
```
poadhika: Can we be more specific on error message. For example "Bundle type is not present. Possible values are: transaction, batch".

rbans96:
> Can we add a test for this?

Added e2e test

Issue #802, opened 2020-01-02
Docker for development doesn't work


dixpac:
I'm trying to setup development following docker instructions here https://github.com/microsoft/fhir-server/blob/master/samples/docker/README.md

`env SAPASSWORD='mypass' docker-compose -f samples/docker/docker-compose.yaml up -d`

But it doesn't work I get this error (tried on OSX mojave and Ubuntu 18.94)
<img width="1280" alt="Screenshot 2020-01-02 at 12 54 37" src="https://user-images.githubusercontent.com/7427365/71666000-54702480-2d5f-11ea-87ba-9ede338187e6.png">

Help would be much appreciated :) ?

dixpac:
I'm .net beginner. It seems that docker image `mcr.microsoft.com/dotnet/core/sdk:2.2-bionic`
doesn't have a SKD with version `2.2.408` but `2.2.207`. When I change to `2.2.207` here [here](https://github.com/microsoft/fhir-server/blob/master/samples/docker/docker_global.json#L3)
Then this step executes successfully but I got error on publish [here](https://github.com/microsoft/fhir-server/blob/master/samples/docker/Dockerfile#L14)

hansenms:
@marrobi

marrobi:
@dixpac odd. Just pulled from master and ran the compose command and the images built ok.

We are doing something slightly odd to override the default SDK version, this line: https://github.com/microsoft/fhir-server/blob/e2a8a845563770f812badfc693fccc6317d82fed/samples/docker/Dockerfile#L8 to set it to `2.2.408` as thats waht was in the image `2.2-bionic` image

Hmm, just re pulled the `2.2-bionic` image and seems the tag is associated with a new image. I can reproduce the error. Leave it with me while I figure out how the image versions work..

marrobi:
@dixpac weird that the SDK version has gone from `2.2.408` to `2.2.207` in a newer image. I'll update the `global.json` and the image tag to `2.2.207-bionic`

I made the change exactly as you have above and the build succeeded. Could you post the error you got on publish?

@hansenms do you know we can ensure we always use the SDK version available in the image? Otherwise each time it is changed on Docker hub the build will break. Will try to find out.

hansenms:
Well, I think what we ultimately need to do is to have the Docker build done as part of our PR build process, so we know it is working. We should then publish the images so that we at least know where to find a working version. We have a work item for this #779.

hansenms:
[...]

PR #803, opened 2020-01-02
Remove sdk in global.json for docker builds.

marrobi:
## Description
Removes the SDK version from `global.json` when building the Docker image

## Related issues
Addresses #802

## Testing
Build works on WSL - however should wait for confirmation that issue is resolved on #802 before merging.


marrobi:
I've actually worked out if we just remove the pinned SDK version - i.e. remove the SDK from `global.json` the image builds successfully. This also means that future updates to 2.2/new 2.2.x docker images should not break the build.

@hansenms thoughts?

hansenms:
@marrobi I think the problem may be that we don't know what might break in another version. There is now an opportunity to build it in a way that we have not tested.

hansenms:
@brendankowitz could you OK this one?

brendankowitz:
Just wanting to confirm the expected behavior, is having the file with an empty entry here different to just removing this file? 
Are we expecting the docker container to build with the latest most appropriate version of .net?

For now I'm fine with either, just wanting to understand the intent

hansenms:
@marrobi is it better to just remove the file?

marrobi:
That works too! Done.

Issue #788, opened 2019-12-13
Using sql, posting a resource with a search url reference returns 500


brandonpollett:
**Describe the bug**
When posting a resource to a SQL backed server that has a reference like the one found below, the server returns a `500` response code

```json
"subject": {
    "reference": "Patient?identifier=2fa15bc7-8866-461a-9000-f739e425860b"
  }
```

**To Reproduce**
1. Post an observation like this one to the `/Observation` route

```json
{
  "resourceType": "Observation",
  "id": "example",
  "text": {
    "status": "generated",
    "div": "<div xmlns=\"http://www.w3.org/1999/xhtml\"><p><b>Generated Narrative with Details</b></p><p><b>id</b>: example</p><p><b>status</b>: final</p><p><b>category</b>: Vital Signs <span>(Details : {http://hl7.org/fhir/observation-category code 'vital-signs' = 'Vital Signs', given as 'Vital Signs'})</span></p><p><b>code</b>: Body Weight <span>(Details : {LOINC code '29463-7' = 'Body weight', given as 'Body Weight'}; {LOINC code '3141-9' = 'Body weight Measured', given as 'Body weight Measured'}; {SNOMED CT code '27113001' = 'Body weight', given as 'Body weight'}; {http://acme.org/devices/clinical-codes code 'body-weight' = 'body-weight', given as 'Body Weight'})</span></p><p><b>subject</b>: <a>Patient/example</a></p><p><b>context</b>: <a>Encounter/example</a></p><p><b>effective</b>: 28/03/2016</p><p><b>value</b>: 185 lbs<span> (Details: UCUM code [lb_av] = 'lb_av')</span></p></div>"
  },
  "status": "final",
  "category": [
    {
      "coding": [
        {
          "system": "http://hl7.org/fhir/observation-category",
          "code": "vital-signs",
          "display": "Vital Signs"
        }
      ]
    }
  ],
  "code": {
    "coding": [
      {
        "system": "http://loinc.org",
        "code": "29463-7",
        "display": "Body Weight"
      },
      {
        "system": "http://loinc.org",
        "code": "3141-9",
        "display": "Body weight Measured"
      },
      {
        "system": "http://snomed.info/sct",
        "code": "27113001",
        "display": "Body weight"
      },
      {
        "system": "http://acme.org/devices/clinical-codes",
        "code": "body-weight",
        "display": "Body Weight"
      }
    ]
  },
  "subject": {
[...]

PR #806, opened 2020-01-08
Fix post Observation with search reference

rbans96:
## Description
Fixes the 500 error when search url reference is present in the Observation.
If search url reference is present then POST/Observation is success
If search url reference is not present or returns more than one entry then POST/Observation fails with 400.

Refactors the code - moved couple of methods and Resource Extensions class from api to core layer because of its reusability in the core layer

## Related issues
Addresses #788

## Testing
Unit tests and e2e tests


src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionBundleValidator.cs
```diff
@@ -68,6 +65,12 @@ public async Task ValidateBundle(Hl7.Fhir.Model.Bundle bundle, CancellationToken
             }
         }
 
+        public async Task ResolveBundleReferences(EntryComponent entry, Dictionary<string, (string resourceId, string resourceType)> referenceIdDictionary, CancellationToken cancellationToken)
```
poadhika: Not sure if this would be the right location for this method. As per class name, this class should only handle validation of bundle?

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/BaseConditionalHandler.cs
```diff
@@ -54,5 +60,66 @@ protected async Task<SearchResultEntry[]> Search(string instanceType, IReadOnlyL
 
             return matchedResults;
         }
+
+        public async Task ResolveBundleReferencesAsync(Resource resource, Dictionary<string, (string resourceId, string resourceType)> referenceIdDictionary, string requestUrl, CancellationToken cancellationToken)
```
poadhika: Similarly., not sure if this would be the correct location for this method. Should this be in a separate class?

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Create/CreateResourceHandler.cs
```diff
@@ -38,6 +44,8 @@ public async Task<UpsertResourceResponse> Handle(CreateResourceRequest message,
             // If an Id is supplied on create it should be removed/ignored
             resource.Id = null;
 
+            await ResolveBundleReferencesAsync(resource, _referenceIdDictionary, resource.ResourceType.ToString(), cancellationToken);
```
poadhika: Are we only resolving conditional reference here? What would be the behavior in non-conditionalreference such as  resource/resourceid or any urn?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ObservationResolveReferenceTests.cs
```diff
@@ -0,0 +1,74 @@
...
+            var uniqueIdentifier = Guid.NewGuid().ToString();
+
+            patient.Identifier.First().Value = uniqueIdentifier;
+
+            // Create patient resource
+            await Client.CreateAsync(patient);
+
+            observation.Subject.Reference = $"Patient?identifier={uniqueIdentifier}";
+
+            FhirResponse<Observation> updateResponse = await Client.CreateAsync(
```
poadhika: nit: createResponse?

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Create/CreateResourceHandler.cs
poadhika: Also we might want to change the name of methodname as we are no longer resolving only bundle reference?

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Resources/BaseConditionalHandlerTests.cs
```diff
@@ -0,0 +1,163 @@
...
+            IFhirDataStore fhirDataStore = Substitute.For<IFhirDataStore>();
+            Lazy<IConformanceProvider> conformanceProvider = Substitute.For<Lazy<IConformanceProvider>>();
+            IResourceWrapperFactory resourceWrapperFactory = Substitute.For<IResourceWrapperFactory>();
+            ResourceIdProvider resourceIdProvider = Substitute.For<ResourceIdProvider>();
+            _testBaseConditionalHandler = new TestBaseConditionalHandler(fhirDataStore, _searchService, conformanceProvider, resourceWrapperFactory, resourceIdProvider);
+        }
+
+        [Fact]
+        public async Task GivenATransactionBundleWithIdentifierReferences_WhenResolved_ThenReferencesValuesAreNotUpdated()
+        {
```
poadhika: I donot think, this is a right class for these bundle specific tests.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionBundleValidator.cs
rbans96: +1, removed from here

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/BaseConditionalHandler.cs
rbans96: Since we need to resolve references for conditional resource whether transaction, bundle or observation So I think it makes sense to have it on this class. Thoughts?

poadhika: yeah, the method name confused me. We can change the name of method to something that's not only pointing to bundle, since its resolving reference within othertype of resources as well?

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Resources/BaseConditionalHandlerTests.cs
rbans96: These tests are to validate the behavior of resolveRefereneAsSync method and now this method is moved to BaseConditionalHandler, So I think its good to have unit tests in the corresponding test class ie. BaseConditionalHandlerTests. Thoughts?

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Resources/Bundle/BundleHandlerTests.cs
```diff
@@ -229,138 +224,6 @@ public async Task GivenABundle_WhenMultipleRequests_ReturnsABundleResponseWithCo
...
-            var referenceIdDictionary = new Dictionary<string, (string resourceId, string resourceType)>();
-            foreach (var entry in bundle.Entry)
-            {
-                var exception = await Assert.ThrowsAsync<RequestNotValidException>(() => _bundleHandler.ResolveBundleReferences(entry, referenceIdDictionary, CancellationToken.None));
-                Assert.Equal(exception.Message, expectedMessage);
-            }
-        }
-
-        [Fact]
-        public async Task GivenAConfigurationEntryLimit_WhenExceeded_ThenBundleEntryLimitExceededExceptionShouldBeThrown()
```
poadhika: Did we delete this test by mistake? It is not related to resolvingreference?

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Resources/BaseConditionalHandlerTests.cs
poadhika: I guess it makes sense. Since all these tests are using bundle as a resource, I was thinking if it will make sense to put these separately. But there is no harm to leave them here too.

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Resources/Bundle/BundleHandlerTests.cs
rbans96: Good catch. There were some merge conflicts from the master, resolved them and now this test is here.

PR #807, opened 2020-01-08
Add bundle entry limit to configuration and enforce it.

brandonpollett:
## Description
Adds a limit to the number of entries that can be in a bundle that is posted for either a transaction or batch. 

The default value is set to 500 for both R3 and STU4. If the value is not specified (or the value is specified as 0) it does not limit the number of resources.

## Related issues
Addresses [AB#70863](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/70863).

## Testing
Manual testing and added a unit test.


src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/OperationOutcomeExceptionFilterAttribute.cs
```diff
@@ -81,6 +82,7 @@ public override void OnActionExecuted(ActionExecutedContext context)
                     case ResourceNotValidException _:
                     case BadRequestException _:
                     case RequestNotValidException _:
+                    case BundleEntryLimitExceededException _:
```
namadabu: nit: might be helpful to add a UT to make sure this exception is mapped to BadRequest

PR #812, opened 2020-01-10
Implements SQL export job acquisition

rotodd:
## Description
- Implements the method to acquire export jobs, another step in adding SQL support for [bulk export](https://microsofthealth.visualstudio.com/Health/_wiki/wikis/Resolute.wiki/25/BulkExportDesignSpec)
- Most of the logic lives in a stored procedure, which:
1. Determines how many jobs can be picked up
2. Picks up jobs that are either stale or marked as queued
3. Marks the acquired jobs as running and timestamps them
- Updates SQL code generation files to allow the binary SQL data type to be used for the JobVersion column in a temporary table variable

Export jobs being created and then acquired (`MaximumNumberOfConcurrentJobsAllowed = 5`, `JobPollingFrequency = 10s`):
queued-to-running

## Related issues
Addresses [#AB71331](https://microsofthealth.visualstudio.com/Health/_workitems/edit/71331).

## Testing
- Enables running the existing Cosmos integration tests for export job acquisition with SQL


src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
```diff
@@ -1259,7 +1259,7 @@ CREATE TABLE dbo.ExportJob
     Status varchar(10) NOT NULL,
     HeartbeatDateTime datetimeoffset(7) NULL,
     QueuedDateTime datetimeoffset(7) NOT NULL,
-    RawJobRecord varbinary(max) NOT NULL,
+    RawJobRecord varchar(max) NOT NULL,
```
rotodd: I am currently storing the raw job record as a JSON string to facilitate updating the status in the `AcquireExportJobs` procedure. This *seems* to be alright since export jobs records are fairly small.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
```diff
@@ -1336,6 +1321,82 @@ AS
...
+    SELECT TOP (@limit) Id, JobVersion
+    FROM dbo.ExportJob
+    WHERE (Status = 'Queued' OR (Status = 'Running' AND HeartbeatDateTime <= @expirationDateTime))
+    ORDER BY HeartbeatDateTime, QueuedDateTime
+
+    DECLARE @updatedJobs TABLE (RawJobRecord varchar(max) NOT NULL, JobVersion binary(8) NOT NULL)
+
+    -- Update each available job's status to running both in the export table's status column and in the raw export job record JSON.
+    UPDATE dbo.ExportJob
+    SET Status = 'Running', HeartbeatDateTime = @heartbeatDateTime, RawJobRecord = REPLACE(RawJobRecord, '"status":1', '"status":2')
```
rotodd: Is there a nicer way to update the JSON document?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Model/Column.cs
```diff
@@ -472,6 +471,49 @@ public override void Set(SqlDataRecord record, int ordinal, byte[] value)
         }
     }
 
+    public class BinaryColumn : Column<byte[]>
+    {
+        public BinaryColumn(string name, int length)
+            : base(name, SqlDbType.Binary, true, length)
```
rotodd: @johnstairs I've hard coded `Nullable = true` here. Is there a way to collect this info from the SQL file and pass it in to the Column constructor?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
brandonpollett: Maybe this: https://docs.microsoft.com/en-us/sql/t-sql/functions/json-modify-transact-sql?view=sql-server-ver15

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
```diff
@@ -1336,6 +1321,82 @@ AS
...
+    OUTPUT inserted.RawJobRecord, inserted.JobVersion
+    INTO @updatedJobs
+    WHERE Id IN (SELECT Id FROM @availableJobs) AND JobVersion IN (SELECT JobVersion FROM @availableJobs)
```
brendankowitz: Might want to check this statement. This seems like it would be true as long as an Id existed in the `@availableJobs` with a version from any of the `@availableJobs`. It doesn't look to be linking the Id, Version record specifically

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
rotodd: Thanks, fixed in 0a8cccd.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
rotodd: Makes sense, thanks. Looking into this now.

rotodd: Let me know how fa6ecae looks

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Model/Column.cs
johnstairs: DId you look at `SqlVisitor.IsColumnNullable`?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
```diff
@@ -1278,22 +1278,7 @@ CREATE UNIQUE NONCLUSTERED INDEX IX_ExportJob_Status_HeartbeatDateTime_QueuedDat
...
-        INCREMENT BY 1
-        MINVALUE 0
-        MAXVALUE 79999
-        CYCLE
-        CACHE 1000000
-GO
-
-/*************************************************************
-    Stored procedure for exporting
+    Stored procedures for exporting
```
johnstairs: Because this is all in one big file, I thing it would be good to put the definitions related to export after those related to CRUD, so I suggest moving these to the bottom of the file.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
johnstairs: I think that's fine.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
```diff
@@ -1336,6 +1321,82 @@ AS
...
+    @heartbeatDateTime datetimeoffset(7)
+AS
+    SET NOCOUNT ON
+    SET XACT_ABORT ON
+
+    SET TRANSACTION ISOLATION LEVEL SERIALIZABLE
+    BEGIN TRANSACTION
+
+    -- Get the number of jobs that are running and not stale.
+    -- Acquire a table lock, which will be held for the entire transaction.
```
johnstairs: exclusive table lock. I'd also add why.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
```diff
@@ -1336,6 +1321,82 @@ AS
...
+--
+-- DESCRIPTION
+--     Timestamps the available export jobs and sets their statuses to running.
+--
+-- PARAMETERS
+--     @expirationDateTime
+--         * The time before which an export job is considered stale
+--     @maximumNumberOfConcurrentJobsAllowed
+--         * The maximum number of running jobs we can have at once
+--     @heartbeatDateTime
```
johnstairs: It's not a big deal here, but you could use the SQL functions to determine the current time instead of passing in times. (you can still parameterize the expiration timespans). The reason would be that different FHIR servers can each have differing clock skews.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
```diff
@@ -1336,6 +1321,82 @@ AS
...
+    DECLARE @availableJobs TABLE (Id varchar(64) COLLATE Latin1_General_100_CS_AS NOT NULL, JobVersion binary(8) NOT NULL)
+
+    -- Get the available jobs, which are export jobs that are queued or stale.
+    INSERT INTO @availableJobs
+    SELECT TOP (@limit) Id, JobVersion
+    FROM dbo.ExportJob
+    WHERE (Status = 'Queued' OR (Status = 'Running' AND HeartbeatDateTime <= @expirationDateTime))
+    ORDER BY HeartbeatDateTime, QueuedDateTime
+
+    DECLARE @updatedJobs TABLE (RawJobRecord varchar(max) NOT NULL, JobVersion binary(8) NOT NULL)
```
johnstairs: I think (but I'm not certain) that you can omit the table variable declaration, the `INTO` clause in the `UPDATE` below, and the subsequent `SELECT`. The `OUTPUT` clause should be included in the stored procedure's resultset because of the the `OUTPUT` clause.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
```diff
@@ -97,7 +74,51 @@ public Task<ExportJobOutcome> GetExportJobByHashAsync(string hash, CancellationT
...
+
+                using (SqlDataReader sqlDataReader = await sqlCommand.ExecuteReaderAsync(CommandBehavior.SequentialAccess, cancellationToken))
+                {
+                    while (await sqlDataReader.ReadAsync(cancellationToken))
+                    {
+                        (string rawJobRecord, byte[] rowVersionAsBytes) = sqlDataReader.ReadRow(V1.ExportJob.RawJobRecord, V1.ExportJob.JobVersion);
+
+                        var exportJobRecord = JsonConvert.DeserializeObject<ExportJobRecord>(rawJobRecord);
+
+                        if (BitConverter.IsLittleEndian)
```
johnstairs: You could consider doing conversion in SQL where you don't have to worry about endianess: `CONVERT(bigint, JobVersion)`.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
```diff
@@ -97,7 +74,51 @@ public Task<ExportJobOutcome> GetExportJobByHashAsync(string hash, CancellationT
...
+                        {
+                            Array.Reverse(rowVersionAsBytes);
+                        }
+
+                        const int startIndex = 0;
+                        var rowVersionAsDecimalString = BitConverter.ToInt32(rowVersionAsBytes, startIndex).ToString();
```
johnstairs: ToInt64

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
rotodd: I agree that returning `bigint` would be cleaner. I'm able to implement this - however I would no longer be able to use the `SqlDataReader.ReadRow(column0, column1...)` method to read the value (and would need to opt for `SqlDataReader.GetInt64(columnName, ordinal)` instead).

I decided to leave the conversion in the C# code for now but am happy to change it if you think the alternative is better practice.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
rotodd: Right thanks, makes sense. Moved in ff5f3e1.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
rotodd: Updated in ff5f3e1.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
rotodd: Fixed in 933a13d.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
rotodd: Yes you're right! Awesome. Removed in 7caed89.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
rotodd: Good idea. Switched to `SYSUTCDATETIME()` in 5633f4a.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Model/Column.cs
rotodd: As discussed, I hard coded `Nullable = false` in 613031e, since the rowversion value can never be `null`.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
johnstairs: I think you should try to stick with all datetimes on the SQL side. You are still taking UtcNow on the FHIR server and passing it in to the stored procedure.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
```diff
@@ -97,7 +74,47 @@ public Task<ExportJobOutcome> GetExportJobByHashAsync(string hash, CancellationT
...
         {
-            throw new NotImplementedException();
+            // TODO: Implement this method.
+            return Task.FromResult(new ExportJobOutcome(jobRecord, eTag));
+        }
+
+        public async Task<IReadOnlyCollection<ExportJobOutcome>> AcquireExportJobsAsync(ushort maximumNumberOfConcurrentJobsAllowed, TimeSpan jobHeartbeatTimeoutThreshold, CancellationToken cancellationToken)
+        {
+            // We will consider a job to be stale if its timestamp is smaller than or equal to this.
+            DateTimeOffset expirationTime = Clock.UtcNow - jobHeartbeatTimeoutThreshold;
```
jackliums: Is there particular reason why this timestamp is calculated in the FE as opposed to just pass in the threshold and have the BE calculate the cutoff time?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
```diff
@@ -1818,3 +1747,128 @@ AS
...
+    SELECT TOP (@limit) Id, JobVersion
+    FROM dbo.ExportJob
+    WHERE (Status = 'Queued' OR (Status = 'Running' AND HeartbeatDateTime <= @expirationDateTime))
+    ORDER BY HeartbeatDateTime, QueuedDateTime
+
+    DECLARE @heartbeatDateTime datetime2(7) = SYSUTCDATETIME()
+
+    -- Update each available job's status to running both in the export table's status column and in the raw export job record JSON.
+    UPDATE dbo.ExportJob
+    SET Status = 'Running', HeartbeatDateTime = @heartbeatDateTime, RawJobRecord = JSON_MODIFY(RawJobRecord,'$.status', 2)
```
jackliums: Is the status property being serialized as number instead of string? In Cosmos DB, we serialize the value as string so if we can do that here, I think it would be better. One less thing to maintain (e.g., BE already knows the status value Running, if we serialize it as string, then we don't need to also know that Running = 2).

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
```diff
@@ -97,7 +74,47 @@ public Task<ExportJobOutcome> GetExportJobByHashAsync(string hash, CancellationT
...
+
+                        var exportJobRecord = JsonConvert.DeserializeObject<ExportJobRecord>(rawJobRecord);
+
+                        if (BitConverter.IsLittleEndian)
+                        {
+                            Array.Reverse(rowVersionAsBytes);
+                        }
+
+                        const int startIndex = 0;
+                        var rowVersionAsDecimalString = BitConverter.ToInt64(rowVersionAsBytes, startIndex).ToString();
```
jackliums: nit: why not just use named parameter here? e.g., BitConverter.ToInt64(rowVersionAsBytes, startIndex: 0)?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
```diff
@@ -97,7 +74,47 @@ public Task<ExportJobOutcome> GetExportJobByHashAsync(string hash, CancellationT
...
+
+                        acquiredJobs.Add(new ExportJobOutcome(exportJobRecord, WeakETag.FromVersionId(rowVersionAsDecimalString)));
+                    }
+                }
+
+                return new ReadOnlyCollection<ExportJobOutcome>(acquiredJobs);
```
jackliums: You don't need to allocate another colleciton. Just return acquiredJobs.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
rotodd: Nope, you're right that it makes sense to implement this on the SQL side. John pointed out the same thing. Fixed in d9a5f4b.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
rotodd: Right, thanks. Fixed in d9a5f4b.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
rotodd: Oops thank you. Fixed in 9f3d322.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
rotodd: Thanks, fixed in 9f3d322.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
rotodd: Ah yes, it was being serialized as an int. I changed it to a string in 255ef52. Thanks!

PR #840, opened 2020-01-17
Adds docker to build

brendankowitz:


hansenms:
I think we should push these images to an Azure container registry and upon successful tests, build, tests, etc. we should publish to mcr,microsoft.com (https://azure.microsoft.com/en-us/blog/microsoft-syndicates-container-catalog/). I think build and push would look something like:

```
- task: Docker@2
  displayName: Login to ACR
  inputs:
    command: login
    containerRegistry: fhir-server-acr

- task: Docker@2
  displayName: Build and Push
  inputs:
    command: buildAndPush
    repository: fhirserver
    tags: |
      $(Build.BuildId)
      $(Build.buildNumber)

- task: Docker@2
  displayName: Logout of ACR
  inputs:
    command: logout
    containerRegistry: fhirserver-acr
```

brendankowitz:
I'll test this out. I think PRs should just build/validate, I'll look to add a step to CI to push to a registry

PR #841, opened 2020-01-17
Update www-authenticate header with additional information

brandonpollett:
## Description
This PR updates the `www-authenticate` header to include variables for `authorization_uri`, `resource_id`, and `realm`. The goal of this update is to aid the client in the event of a `401` with some information regarding obtaining a ticket.

## Related issues
Addresses [AB#70967](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/70967).

## Testing
The E2E tests were updated to check the values returned from an un-authorized request.


src/Microsoft.Health.Fhir.Api/Modules/SecurityModule.cs
```diff
@@ -52,6 +52,7 @@ public void Load(IServiceCollection services)
                         options.Authority = _securityConfiguration.Authentication.Authority;
                         options.Audience = _securityConfiguration.Authentication.Audience;
                         options.RequireHttpsMetadata = true;
+                        options.Challenge = $"Bearer authorization_uri={_securityConfiguration.Authentication.Authority}, resource_id={_securityConfiguration.Authentication.Audience}, realm={_securityConfiguration.Authentication.Audience}";
```
hansenms: I think this is actually supposed to be the authorization URI, so:

```
$"Bearer authorization_uri={_securityConfiguration.Authentication.Authority}/oauth2/authorize, resource_id={_securityConfiguration.Authentication.Audience}, realm={_securityConfiguration.Authentication.Audience}"
```

hansenms: Actually, upon looking into this. Your code is correct. I had noticed some products that do it differently, but this is correct according to AAD docs.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/BasicAuthTests.cs
```diff
@@ -26,6 +29,8 @@ namespace Microsoft.Health.Fhir.Tests.E2E.Rest
     [HttpIntegrationFixtureArgumentSets(DataStore.CosmosDb, Format.Json)]
     public class BasicAuthTests : IClassFixture<HttpIntegrationTestFixture>
     {
+        private static readonly Regex WwwAuthenticatePattern = new Regex(@".*authorization_uri\s*=\s*(?<authorization_uri>[^\s,]*)?, resource_id=\""*(?<resource_id>[^""]+)\""*, realm=\""*(?<realm>[^""]+)\""*", RegexOptions.IgnoreCase);
```
hansenms: Not sure it is hugely important, but the header you are generating does not have '"' around the `resource_id` or the `realm` string. But this regex optionally matches them. It should maybe also be `\""?` (zero or one).

brandonpollett: This regex was a bit more robust than needed for the test. I trimmed it down to be more in line with what we actually expect the header to be.

PR #846, opened 2020-01-22
Implements SQL export job update and get by hash, id

rotodd:
## Description
- Implements the final three SQL [bulk export](https://microsofthealth.visualstudio.com/Health/_wiki/wikis/Resolute.wiki/25/BulkExportDesignSpec) methods:
1. Get an export job given its ID
2. Get an export job given a hash of its ID
3. Update an export job
- Adds three new stored procedures, which follow the Cosmos side implementation

## Demo
1. Create a test patient (since we are starting with an empty SQL database)
2. Send an export request, passing in a Base64 encoded connection string to a test storage account
3. An export job will be created, and its status will change: queued --> running --> completed
4. The exported patient resource is visible in the Azure storage account

sql-export-demo

## Related issues
Addresses [#AB71335](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Resolute/Stories/?workitem=71335) and [#AB71336](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Resolute/Stories/?workitem=71336).

## Testing
- Enables running the existing Cosmos integration tests for export job getters with SQL
- Adds a new test for testing export job update

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
@@ -1790,6 +1790,8 @@ GO
 -- PARAMETERS
 --     @id
 --         * The ID of the export job record
+--     @hash
+--         * The SHA256 hash of the export job record ID
 --     @status
 --         * The status of the export job
 --     @queuedDateTime
@@ -1802,6 +1804,7 @@ GO
 --
 CREATE PROCEDURE dbo.CreateExportJob
     @id varchar(64),
+    @hash varchar(64),
     @status varchar(10),
     @queuedDateTime datetimeoffset(7),
     @rawJobRecord varchar(max)
@@ -1812,9 +1815,9 @@ AS
     BEGIN TRANSACTION
 
     INSERT INTO dbo.ExportJob
-        (Id, Status, QueuedDateTime, RawJobRecord)
+        (Id, Hash, Status, QueuedDateTime, RawJobRecord)
     VALUES
-        (@id, @status, @queuedDateTime, @rawJobRecord)
+        (@id, @hash, @status, @queuedDateTime, @rawJobRecord)
   
     SELECT CAST(MIN_ACTIVE_ROWVERSION() AS INT)
 
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
@@ -1754,6 +1754,7 @@ GO
 CREATE TABLE dbo.ExportJob
 (
     Id varchar(64) COLLATE Latin1_General_100_CS_AS NOT NULL,
+    Hash varchar(64) COLLATE Latin1_General_100_CS_AS NOT NULL,
     Status varchar(10) NOT NULL,
     HeartbeatDateTime datetime2(7) NULL,
     QueuedDateTime datetimeoffset(7) NOT NULL,
@@ -1766,8 +1767,9 @@ CREATE UNIQUE CLUSTERED INDEX IXC_ExportJob ON dbo.ExportJob
     Id
 )
 
-CREATE UNIQUE NONCLUSTERED INDEX IX_ExportJob_Status_HeartbeatDateTime_QueuedDateTime ON dbo.ExportJob
+CREATE UNIQUE NONCLUSTERED INDEX IX_ExportJob_Hash_Status_HeartbeatDateTime_QueuedDateTime ON dbo.ExportJob
 (
+    Hash,
     Status,
     HeartbeatDateTime,
     QueuedDateTime
@@ -1850,6 +1852,36 @@ GO
 
 --
 -- STORED PROCEDURE
+--     Gets an export job given the hash of its ID.
+--
+-- DESCRIPTION
+--     Retrieves the export job record from the ExportJob table that has the matching hash.
+--
+-- PARAMETERS
+--     @hash
+--         * The SHA256 hash of the export job record ID
+--
+-- RETURN VALUE
+--     The matching export job.
+--
+CREATE PROCEDURE dbo.GetExportJobByHash
+    @hash varchar(64)
+AS
+    SET NOCOUNT ON
+
+    SET XACT_ABORT ON
+    BEGIN TRANSACTION
+
+    SELECT TOP(1) RawJobRecord, JobVersion
+    FROM dbo.ExportJob
+    WHERE Hash = @hash AND (Status = 'Queued' OR Status = 'Running')
+    ORDER BY HeartbeatDateTime, QueuedDateTime ASC
+
+    COMMIT TRANSACTION
+GO
+
+--
+-- STORED PROCEDURE
 --     Acquires export jobs.
 --
 -- DESCRIPTION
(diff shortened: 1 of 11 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
```diff
@@ -1810,9 +1815,111 @@ AS
...
+AS
+    SET NOCOUNT ON
+
+    SET XACT_ABORT ON
+    BEGIN TRANSACTION
+
+    SELECT TOP(1) RawJobRecord, JobVersion
+    FROM dbo.ExportJob
+    WHERE Hash = @hash AND (Status = 'Queued' OR Status = 'Running')
+    ORDER BY HeartbeatDateTime, QueuedDateTime ASC
```
rotodd: This follows the Cosmos-side implementation - however, I'm not certain that we need to include a filter for the status to be running or queued. Is there a reason why we filter for this when using Cosmos? @jackliums @namadabu

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
```diff
@@ -1810,9 +1815,111 @@ AS
...
+-- RETURN VALUE
+--     The matching export job.
+--
+CREATE PROCEDURE dbo.GetExportJobById
+    @id varchar(64)
+AS
+    SET NOCOUNT ON
+
+    SET XACT_ABORT ON
+    BEGIN TRANSACTION
```
johnstairs: You don't need a transaction here because there is only one statement in the transaction

rotodd: Right - thanks. Fixed in b5cae90.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
namadabu: We are using the hash comparison as a "best-effort" to prevent re-queuing the same job while we are still processing the first one. The reason we are filtering on the status is to make sure we are not blocking valid retries of jobs. As an example, let's say the initial export request failed because the user had forgotten to configure their storage account properly. They fix the config and then retry the same export request. We should process the export request in this scenario (even though the hash will be the same) and hence queue a new job.

rotodd: Ah I see. Thanks for the explanation!

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
```diff
@@ -1810,9 +1815,101 @@ AS
...
+
+    -- We will timestamp the jobs when we update them to track stale jobs.
+    DECLARE @heartbeatDateTime datetime2(7) = SYSUTCDATETIME()
+
+    UPDATE dbo.ExportJob
+    SET Status = @status, HeartbeatDateTime = @heartbeatDateTime, QueuedDateTime = @queuedDateTime, RawJobRecord = @rawJobRecord
+    WHERE Id = @id
```
rotodd: I believe a lock should be acquired and the row version should be checked before the job record is updated here. Looking into this now.

rotodd: Added this in bafd98a and c41548c. @johnstairs may you take a look at this when you have a chance?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
```diff
@@ -71,21 +72,93 @@ public async Task<ExportJobOutcome> CreateExportJobAsync(ExportJobRecord jobReco
...
+                }
+                catch (SqlException e)
+                {
+                    if (e.Number == SqlErrorCodes.PreconditionFailed)
+                    {
+                        throw new JobConflictException();
+                    }
+                    else
+                    {
+                        _logger.LogError(e, Core.Resources.SqlErrorOnExportUpdate);
```
namadabu: I don't think you need to create a `Resource` string for logging. It's only required for messages which are returned to the customer.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
```diff
@@ -71,21 +72,93 @@ public async Task<ExportJobOutcome> CreateExportJobAsync(ExportJobRecord jobReco
...
+            byte[] rowVersionAsBytes = GetRowVersionAsBytes(eTag);
+
+            using (SqlConnectionWrapper sqlConnectionWrapper = _sqlConnectionWrapperFactory.ObtainSqlConnectionWrapper(true))
+            using (SqlCommand sqlCommand = sqlConnectionWrapper.CreateSqlCommand())
+            {
+                V1.UpdateExportJob.PopulateCommand(
+                    sqlCommand,
+                    jobRecord.Id,
+                    jobRecord.Status.ToString(),
+                    jobRecord.QueuedTime,
```
namadabu: `QueuedTime` will always remain the same for an export job. I don't think we need to pass it in for the update call.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
```diff
@@ -71,21 +72,93 @@ public async Task<ExportJobOutcome> CreateExportJobAsync(ExportJobRecord jobReco
...
+                    JsonConvert.SerializeObject(jobRecord, _jsonSerializerSettings),
+                    rowVersionAsBytes);
+
+                try
+                {
+                    var rowVersion = (int?)await sqlCommand.ExecuteScalarAsync(cancellationToken);
```
namadabu: Just curious why you are returning the rowVersion as an int here instead of the byte[] ?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
rotodd: Good point. I thought about this and originally ended up leaving it in because I used it to prioritize older queued jobs when acquiring export jobs in the `dbo.AcquireExportJobs` procedure.

Looking at the Cosmos implementation, I noticed that the queued jobs are timestamped (the same way I timestamp running jobs), and this is used when deciding which jobs to pick up. I took out `QueuedDatetime` from the `dbo.ExportJob` table and updated `HeartbeatDateTime` in `dbo.CreateExportJob` in bd0d061. This simplifies the SQL side implementation and aligns its behaviour with the Cosmos side. Thank you!

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
rotodd: Right, thanks. Fixed in 1632801.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
rotodd: It seemed easier to do the conversion from bytes to an integer value in SQL, since the information is captured and stored as a string representation of a decimal in the `ExportJobOutcome.WeakETag` property on the C# side.

I am not tied to doing it this way though! Let me know if you have a nicer/cleaner implementation in mind.

rotodd: As discussed, I moved the conversion to the C# side and made use of the existing conversion methods for code readability in b6bad92.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/1.sql
rotodd: To move forward, I'm going to merge this to my feature branch for now. If needed, it can be readdressed when I merge to master.

PR #852, opened 2020-01-25
Support export to storage account set in config

namadabu:
## Description
For bulk export, we require users to pass in a _destinationType and _destinationConnectionSettings parameter along with the request. This information will be used to connect to the appropriate destination and export data there. With the PR, we are removing this functionality and changing the way export works. The storage account information should now be added to the app config and this will be used for exporting data. The default template has also been updated to support such a scenario.

TODO:
-Export Destination client should handle token expiry
-Add more tests to ExportJobTask

## Related issues
Addresses [issue [AB#69774](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/69774) [AB#71879](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/71879) [AB#71880](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/71880)].

## Testing
Existing tests. Manually tested by deploying a fhir-server instance (using the updated template) and verified that I could export data to the associated storage account.


samples/templates/default-azuredeploy.json
```diff
@@ -158,7 +167,10 @@
             "FhirServer:Security:Authentication:Authority": "[parameters('securityAuthenticationAuthority')]",
             "FhirServer:Security:Authentication:Audience": "[parameters('securityAuthenticationAudience')]",
             "CosmosDb:ContinuationTokenSizeLimitInKb": "1",
-            "DataStore": "CosmosDb"
+            "DataStore": "CosmosDb",
```
brendankowitz: What happens with the SQL template deployment?

namadabu: I haven't added these changes this to the SQL template since we don't yet support export for SQL.

samples/templates/default-azuredeploy.json
```diff
@@ -126,14 +126,22 @@
...
+            "defaultValue": "R4",
             "allowedValues": [
                 "Stu3",
                 "R4"
             ],
             "metadata": {
                 "description": "Only applies when specifying Repository Url"
             }
+        },
+        "storageAccountName": {
```
jackliums: Should we derive the storage account name instead of asking user to enter it so it behaves like other dependencies (e.g., KeyVault, AppInsights and etc)?

I added the storage account integration for the dicom-server template: https://github.com/microsoft/dicom-server/pull/23/files

samples/templates/default-azuredeploy.json
```diff
@@ -340,6 +352,34 @@
...
+            "type": "Microsoft.Storage/storageAccounts/providers/roleAssignments",
+            "apiVersion": "2018-09-01-preview",
+            "name": "[concat(parameters('storageAccountName'), '/Microsoft.Authorization/', guid(uniqueString(parameters('storageAccountName'))))]",
+            "dependsOn": [
+                "[parameters('storageAccountName')]",
+                "[variables('serviceName')]"
+            ],
+            "properties": {
+                "roleDefinitionId": "[variables('roleDefinitionId')]",
+                "principalId": "[reference(concat('Microsoft.Web/sites/', variables('serviceName')), '2018-11-01', 'full').identity.principalId]"
```
jackliums: Should minimize to only permission needed (e.g., blob)

src/Microsoft.Health.Fhir.Azure/AzureAccessTokenProvider.cs
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        public string AccessTokenProviderType => "azure";
+
+        public async Task<string> GetAccessTokenForResourceAsync(Uri resourceUri, CancellationToken cancellationToken)
+        {
+            EnsureArg.IsNotNull(resourceUri, nameof(resourceUri));
+
+            string accessToken = await _azureServiceTokenProvider.GetAccessTokenAsync(resourceUri.ToString(), cancellationToken: cancellationToken);
+            if (accessToken == null)
```
jackliums: should it throw exception if it fails to get the access token instead of null?

src/Microsoft.Health.Fhir.Core/Configs/ExportJobConfiguration.cs
```diff
@@ -15,6 +15,23 @@ public class ExportJobConfiguration
         /// </summary>
         public bool Enabled { get; set; }
 
+        /// <summary>
```
jackliums: This will be a breaking change so we need to be explicit about how we want to handle it.

If this code is deployed and there is an existing export job record, the new code will fail, which is fine if that's what we decide to do. Just want to make sure we think about breaking change scenarios when changing existing code.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/AccessTokenProvider/InMemoryAccessTokenProvider.cs
```diff
@@ -0,0 +1,24 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using EnsureThat;
+
+namespace Microsoft.Health.Fhir.Core.Features.Operations.Export.AccessTokenProvider
+{
+    public class InMemoryAccessTokenProvider : IAccessTokenProvider
```
jackliums: Is this needed in the production code?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/Models/DestinationInfo.cs
```diff
@@ -12,9 +11,6 @@ public class DestinationInfo
     {
         public DestinationInfo(string destinationType, string destinationConnectionString)
         {
-            EnsureArg.IsNotNullOrWhiteSpace(destinationType, nameof(destinationType));
```
jackliums: This class is no longer needed.

src/Microsoft.Health.Fhir.Api/Features/Filters/ValidateExportRequestFilterAttribute.cs
```diff
@@ -28,24 +24,8 @@ internal class ValidateExportRequestFilterAttribute : ActionFilterAttribute
...
-        public ValidateExportRequestFilterAttribute(IExportDestinationClientFactory exportDestinationClientFactory)
+        public ValidateExportRequestFilterAttribute()
         {
-            EnsureArg.IsNotNull(exportDestinationClientFactory, nameof(exportDestinationClientFactory));
-
-            _exportDestinationClientFactory = exportDestinationClientFactory;
-
-            _supportedQueryParams = new HashSet<string>(StringComparer.Ordinal)
-            {
-                KnownQueryParameterNames.DestinationType,
```
jackliums: Remove these two from KnownQueryParameterNames

src/Microsoft.Health.Fhir.R4.Web/appsettings.json
```diff
@@ -38,7 +38,10 @@
         },
         "Operations": {
             "Export": {
-                "Enabled": false,
+                "Enabled": true,
+                "StorageAccountType": null,
```
jackliums: Can this be move to top level? E.g., a storage account that can be used for other purpose too not just export?

src/Microsoft.Health.Fhir.Azure/AzureAccessTokenProvider.cs
```diff
@@ -0,0 +1,48 @@
...
+            _logger = logger;
+        }
+
+        public string AccessTokenProviderType => "azure";
+
+        public async Task<string> GetAccessTokenForResourceAsync(Uri resourceUri, CancellationToken cancellationToken)
+        {
+            EnsureArg.IsNotNull(resourceUri, nameof(resourceUri));
+
+            string accessToken = await _azureServiceTokenProvider.GetAccessTokenAsync(resourceUri.ToString(), cancellationToken: cancellationToken);
```
feordin: Are there any kind of transient errors it would make sense to have a retry here?  Or do we retry at a higher level?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -211,16 +208,33 @@ private async Task UpdateJobRecordAsync(CancellationToken cancellationToken)
             }
         }
 
-        // Get destination info from secret store, create appropriate export client and connect to destination.
+        // Get destination information, create appropriate export client and connect to destination.
```
jackliums: I don't think we want to aquire token on every loop right? The token is valid for certain period of time, so we should be able to reuse it until we have to refresh it?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportDestinationClient/IExportDestinationClient.cs
```diff
@@ -29,6 +29,16 @@ public interface IExportDestinationClient
         /// <returns>A <see cref="Task"/> representing connection operation.</returns>
         Task ConnectAsync(string connectionSettings, CancellationToken cancellationToken, string containerId = null);
 
+        /// <summary>
```
jackliums: Have you considered maybe delegate the "authentication" part and glue them together in IoC container?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
jackliums: Also, i don't think it's the export job task's responsibility to acquire access token.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobConfigurationValidator.cs
```diff
@@ -0,0 +1,76 @@
...
+        public bool ValidateExportJobConfig()
+        {
+            if (string.IsNullOrWhiteSpace(_exportJobConfiguration.StorageAccountType) ||
+                !_exportDestinationClientFactory.IsSupportedDestinationType(_exportJobConfiguration.StorageAccountType))
+            {
+                throw new ExportJobConfigValidationException(string.Format(Resources.UnsupportedDestinationType, _exportJobConfiguration.StorageAccountType), HttpStatusCode.BadRequest);
+            }
+
+            // Check whether the config contains a uri to a storage account or a connection string.
+            if (Uri.TryCreate(_exportJobConfiguration.StorageAccountConnection, UriKind.Absolute, out Uri resultUri))
```
jackliums: You already have an interface for handling acess token (i would make it more generic so it can handle how authentication is handled) and then just separate the connection to another implementation of the interface rather than depending on whther the conection is a "url" or not.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ExportController.cs
```diff
@@ -78,16 +77,14 @@ public class ExportController : Controller
...
-            [FromQuery(Name = KnownQueryParameterNames.DestinationConnectionSettings)] string destinationConnectionString)
+        public async Task<IActionResult> Export()
         {
             if (!_exportConfig.Enabled)
             {
                 throw new RequestNotValidException(string.Format(Resources.UnsupportedOperation, OperationsConstants.Export));
             }
 
-            CreateExportResponse response = await _mediator.ExportAsync(_fhirRequestContextAccessor.FhirRequestContext.Uri, destinationType, destinationConnectionString, HttpContext.RequestAborted);
+            CreateExportResponse response = await _mediator.ExportAsync(_fhirRequestContextAccessor.FhirRequestContext.Uri, HttpContext.RequestAborted);
```
jackliums: A few more things we should address (doesn't necessarily need to be in this PR).

1. If the user cancels the export or if export fails, we should delete all exported data from the storage.
2. We need to figure out how to expire exported results for completed job.

src/Microsoft.Health.Fhir.Api/Features/Filters/ValidateExportRequestFilterAttribute.cs
namadabu: Removed.

samples/templates/default-azuredeploy.json
namadabu: Update the template to generate storage account name. Also added an option for users to choose whether they want to enable export.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/Models/DestinationInfo.cs
namadabu: Deleted

src/Microsoft.Health.Fhir.R4.Web/appsettings.json
namadabu: I feel that it would be okay to keep it within export for now and move it later when we decide to use it for other purposes.

src/Microsoft.Health.Fhir.Azure/AzureAccessTokenProvider.cs
namadabu: Changed to throw an exception when we don't get back an access token

src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/AzureExportDestinationClient.cs
```diff
@@ -42,26 +55,89 @@ public async Task ConnectAsync(string connectionSettings, CancellationToken canc
...
             if (!CloudStorageAccount.TryParse(decodedConnectionString, out CloudStorageAccount cloudAccount))
             {
                 throw new DestinationConnectionException(Resources.InvalidConnectionSettings, HttpStatusCode.BadRequest);
             }
 
             _blobClient = cloudAccount.CreateCloudBlobClient();
+            await CreateContainerAsync(_blobClient, containerId);
+        }
+
+        public async Task ConnectHelperAsync(string connectionString, CancellationToken cancellationToken, string containerId = null)
```
namadabu: Renaming and merging this with `ConnectWithAccessTokenAsync()`

samples/templates/default-azuredeploy.json
namadabu: Update the sql template also with export options since we know support export in SQL also!

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/AccessTokenProvider/InMemoryAccessTokenProvider.cs
namadabu: Removed `InMemoryAccessTokenProvider`.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
namadabu: The original change had it only getting the token at the beginning of the export process. Once we connect, we get into the while loop for processing the export. Like you mentioned, I have moved the  acquiring access token logic out of the `ExportJobTask` class. Now that logic is abstracted away as part of the `IExportDestinationClient`

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportDestinationClient/IExportDestinationClient.cs
namadabu: Removed the `ConnectWithAccessTokenAsync` method

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ExportController.cs
brendankowitz: TTL on the document could be a solution for this (but would need a different solution for sql)

src/Microsoft.Health.Fhir.Azure/AzureAccessTokenProvider.cs
namadabu: Will handle retries as part of a separate PR since this PR is already huge. Created WI #72260 to track it

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ExportController.cs
namadabu: Created work items to track these issues (and other nice to haves)

samples/templates/default-azuredeploy.json
namadabu: The `roleDefinitionId` variable is defined to be the guid that corresponds to the `Storage Blob Data Contributor` role (Reference - https://docs.microsoft.com/en-us/azure/role-based-access-control/built-in-roles). If you are referring to the 'full' in the `reference` call, that just corresponds to returning the full object (so that we can get the `identity.principalId` of the app service.

src/Microsoft.Health.Fhir.Core/Microsoft.Health.Fhir.Core.csproj
```diff
@@ -25,6 +25,7 @@
     <PackageReference Include="Ensure.That" Version="8.1.2" />
     <PackageReference Include="FluentValidation" Version="8.6.1" />
     <PackageReference Include="MediatR" Version="7.0.0" />
+    <PackageReference Include="Microsoft.Azure.Services.AppAuthentication" Version="1.3.1" />
```
namadabu: Removed this

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobConfigurationValidator.cs
namadabu: Refactored the code to handle authentication in a separate interface/class and removed the `ExportJobConfigurationValidator`

samples/templates/default-azuredeploy-sql.json
```diff
@@ -138,6 +145,8 @@
         "msdeployPackageUrlToUse": "[if(empty(parameters('msdeployPackageUrl')),variables('defaultMsdeployPackageUrl'),parameters('msdeployPackageUrl'))]",
         "deployAppInsights": "[and(parameters('deployApplicationInsights'),not(variables('isMAG')))]",
         "appInsightsName": "[concat('AppInsights-', variables('serviceName'))]",
+        "roleDefinitionId": "[concat('/subscriptions/', subscription().subscriptionId, '/providers/Microsoft.Authorization/roleDefinitions/', 'ba92f5b4-2d11-453d-a403-e96b0029c9fe')]",
```
jackliums: nit: perhaps a more specific name? like storageBlobDataContributerRoleId?

src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/AzureConnectionStringClientInitializer.cs
```diff
@@ -0,0 +1,49 @@
...
+            {
+                throw new ExportClientInitializerException(Resources.InvalidConnectionSettings, HttpStatusCode.BadRequest);
+            }
+
+            if (!CloudStorageAccount.TryParse(_exportJobConfiguration.StorageAccountConnection, out CloudStorageAccount cloudAccount))
+            {
+                throw new ExportClientInitializerException(Resources.InvalidConnectionSettings, HttpStatusCode.BadRequest);
+            }
+
+            CloudBlobClient blobClient = cloudAccount.CreateCloudBlobClient();
```
jackliums: if this method fails to parse the connection string, would it throw exception? if so, does it need to be translated to ExportClientInitializerException?

src/Microsoft.Health.Fhir.R4.Web/appsettings.json
```diff
@@ -39,7 +39,10 @@
...
```
jackliums: You could simplify even further by remove this and only support blob storage in the Azure implementation.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -211,16 +188,18 @@ private async Task UpdateJobRecordAsync(CancellationToken cancellationToken)
             }
         }
 
-        // Get destination info from secret store, create appropriate export client and connect to destination.
-        private async Task GetDestinationInfoAndConnectAsync(CancellationToken cancellationToken)
+        // Create appropriate export client and connect to destination.
+        private async Task GetDestinationClientAndConnectAsync(CancellationToken cancellationToken)
```
jackliums: This still doesn't address the issue with expiring access token (e.g., it's acquiring token before the while loop but the while loop could take longer than the validation period of the access token). Would that be fixed in this PR or another PR?

src/Microsoft.Health.Fhir.Core/Configs/ExportJobConfiguration.cs
```diff
@@ -15,6 +15,25 @@ public class ExportJobConfiguration
...
+        /// Should be a connection string to the required storage account.
+        /// </summary>
+        public string StorageAccountConnection { get; set; }
+
+        /// <summary>
+        /// Determines the storage account connection that will be used to export data to.
+        /// Should be a uri pointing to the required storage account.
+        /// </summary>
+#pragma warning disable CA1056 // Uri properties should not be strings
+        public string StorageAccountUri { get; set; }
```
jackliums: Have you tried to make this Uri instead of string?

samples/templates/default-azuredeploy-sql.json
namadabu: Updated the name

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
namadabu: Yes. I will add that ability along with retries as part of another PR. The corresponding WI is #72260

src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/AzureConnectionStringClientInitializer.cs
namadabu: The method description does not mention anything about throwing an exception. And I assumed it should be okay to call this since we successfully parse the connection string in the previous step. To be safe, I have added a try-catch block around this call also.

src/Microsoft.Health.Fhir.R4.Web/appsettings.json
namadabu: Deleted this config and related classes that depended on this flow. Also moved the `InMemoryExportDestinationClient` to the tests project

Issue #856, opened 2020-01-27
FHIR Server deployment from ARM template fails


arieschwartzman:
Running this command on the master branch of the repo:

```
New-AzResourceGroupDeployment -TemplateUri https://raw.githubusercontent.com/Microsoft/fhir-server/master/samples/templates/default-azuredeploy.json -ResourceGroupName $rg.ResourceGroupName -serviceName $fhirServiceName
```
Gives this error when deploying on the AppService on West US 2 region:

```
Command: "D:\home\site\deployments\tools\deploy.cmd"
Handling ASP.NET Core Web Application deployment.
A compatible installed .NET Core SDK for global.json version [3.1.100] from [D:\home\site\repository\global.json] was not found
Install the [3.1.100] .NET Core SDK or update [D:\home\site\repository\global.json] with an installed .NET Core SDK:
  1.1.14 [D:\Program Files (x86)\dotnet\sdk]
  2.1.509 [D:\Program Files (x86)\dotnet\sdk]
  2.2.109 [D:\Program Files (x86)\dotnet\sdk]
Failed exitCode=-2147450735, command=dotnet restore "D:\home\site\repository\Microsoft.Health.Fhir.sln"
An error has occurred during web site deployment.
A compatible installed .NET Core SDK for global.json version [3.1.100] from [D:\home\site\repository\global.json] was not found\r\nInstall the [3.1.100] .NET Core SDK or update [D:\home\site\repository\global.json] with an installed .NET Core SDK:\r\nD:\Program Files (x86)\SiteExtensions\Kudu\85.11214.4277\bin\Scripts\starter.cmd "D:\home\site\deployments\tools\deploy.cmd"
```

It worked fine just last week

hansenms:
This is acknowledged. It appears that App Service does not have support for .NET Core 3.1 for Kudu builds. The SDK is not installed. We have to come up with a mitigation strategy and will also investigate when it might be available. 

As a workaround, you can still build locally and deploy the build to app service (which is what our CI/CD does) You just cannot build in the app service. If you set the `repositoryUrl` template parameter to `""` (blank string), the code will not be deployed, but all the database, app service, etc. will bet set up. You can then deploy your build to the app service and you should be good to go.

brandonpollett:
Here's some discussion around 3.1 in App Service - https://github.com/Azure/app-service-announcements-discussions/issues/129

brendankowitz:
For now, try specifying this branch from the last .net 2.2 release:
```
[...]

PR #858, opened 2020-01-28
Use webdeploy package instead of kudu builds

brendankowitz:
## Description
Kudu is not able to build .net core 3 projects, we need to move away from this approach for OSS template deployments.
This PR updates the templates to use a WebDeploy package that is created by the CI process and copied to blob storage.

## Related issues
Addresses #856.

## Testing
Manual deployments

Test from this branch:
```
New-AzResourceGroupDeployment -TemplateUri https://github.com/microsoft/fhir-server/raw/personal/bkowitz/issue-856-deploytemplate/samples/templates/default-azuredeploy.json -ResourceGroupName $rg.ResourceGroupName -serviceName $fhirServiceName
```

### DIFF
--- a/samples/templates/default-azuredeploy-sql.json
+++ b/samples/templates/default-azuredeploy-sql.json
@@ -63,18 +63,11 @@
                 "description": "Enable Azure AAD SMART on FHIR Proxy"
             }
         },
-        "repositoryUrl": {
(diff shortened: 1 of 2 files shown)


brendankowitz:
PR Build parameters will need to be updated to remove RepositoryUrl

PR #861, opened 2020-01-29
Accept Parameters resources in $validate requests

LTA-Thinking:
## Description
Add support for making validate requests using Parameter resources. Support for passing modes has also been added, although only the Delete mode will have functionality. All other modes will throw an Operation Not Implemented exception or a Bad Request depending on if the mode should be supported for the given request.

## Testing
New E2E tests are added to test the new functionality and manual testing with Postman was done.


src/Microsoft.Health.Fhir.Api/Features/Filters/ValidationModeFilterAttribute.cs
```diff
@@ -0,0 +1,75 @@
...
+            {
+                case "CREATE":
+                    throw new OperationNotImplementedException(string.Format(Resources.ValidationModeNotSupported, mode));
+                case "UPDATE":
+                    if (idMode)
+                    {
+                        throw new OperationNotImplementedException(string.Format(Resources.ValidationModeNotSupported, mode));
+                    }
+
+                    throw new BadRequestException(Resources.ValidationForUpdateAndDeleteNotSupported);
```
hsamhita: Should 'Supported' be changed to 'Valid', which goes well with BadRequestException here?

src/Microsoft.Health.Fhir.Api/Features/Filters/ValidationModeFilterAttribute.cs
```diff
@@ -0,0 +1,75 @@
...
+            StringValues modes;
+            if (context.HttpContext.Request.Query.TryGetValue(KnownQueryParameterNames.Mode, out modes))
+            {
+                if (modes.Count > 1)
+                {
+                    throw new BadRequestException(Resources.MultipleModesProvided);
+                }
+
+                var idMode = context.ActionDescriptor.DisplayName.Contains("ValidateById", StringComparison.OrdinalIgnoreCase);
+                var mode = modes[0];
```
namadabu, line 35: Shouldn't we validate that `modes.Count > 0` before trying to access it?

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ValidateController.cs
```diff
@@ -80,14 +103,29 @@ public async Task<IActionResult> Validate([FromBody] Resource resource, [FromQue
...
+            }
+
+            var paramProfile = resource.Parameter.Find(param => param.Name.Equals("profile", System.StringComparison.OrdinalIgnoreCase));
+            if (paramProfile != null && profile != null)
+            {
+                throw new BadRequestException(Resources.MultipleProfilesProvided);
+            }
+            else if (paramProfile != null && profile == null)
+            {
+                profile = paramProfile.Value.ToString();
```
namadabu, line 125: Shouldn't this throw a `OperationNotImplementedException(Resources.ValidateWithProfileNotSupported)` ?

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ValidateController.cs
```diff
@@ -50,26 +51,48 @@ public ValidateController(IMediator mediator, IOptions<FeatureConfiguration> fea
...
-            if (!string.IsNullOrEmpty(profile))
+            if (resource.ResourceType == ResourceType.Parameters)
             {
-                throw new OperationNotImplementedException(Resources.ValidateWithProfileNotSupported);
+                resource = ParseParameters((Parameters)resource, ref profile, ref mode);
             }
 
-            if (!string.IsNullOrEmpty(mode))
+            // This is the same as the filter that is applied in the ValidationModeFilter.
+            // It is needed here to cover the case of the mode being passed as part of a Parameters resource.
```
namadabu, line 86: Would it be possible to parse the resource in the `ValidationModeFilter`, check if it is `Parameters` type and do this validation in the filter itself?

src/Microsoft.Health.Fhir.Api/Features/Filters/ValidationModeFilterAttribute.cs
LTA-Thinking: I had assumed that the TryGetValue method wouldn't return a value if there were none, but it never hurts to have an extra check.

namadabu: It will return the default value for that type. And in the case of `StringValues` it would be an empty array of strings.

src/Microsoft.Health.Fhir.Api/Features/Filters/ValidationModeFilterAttribute.cs
```diff
@@ -0,0 +1,75 @@
...
+                    if (idMode)
+                    {
+                        throw new OperationNotImplementedException(string.Format(Resources.ValidationModeNotSupported, mode));
+                    }
+
+                    throw new BadRequestException(Resources.ValidationForUpdateAndDeleteNotSupported);
+                case "DELETE":
+                    if (idMode)
+                    {
+                        // This is done to bypass any exceptions that could be thrown in latter attribute filters.
```
namadabu, line 56: nit: can you also add a comment here explaining why the expected behavior for DELETE in idMode is to return "ValidationPassed"?

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ValidateController.cs
LTA-Thinking: The issue with doing it in the filter is that it is then done for all of our endpoints, as they all share the validate filter. Since the Parameters syntax is endpoint specific (different endpoints can have different properties) doing this would mean this controller would need a filter specifically for it.

src/Microsoft.Health.Fhir.Api/Features/Filters/ValidationModeFilterAttribute.cs
```diff
@@ -0,0 +1,75 @@
...
+                    throw new BadRequestException(Resources.MultipleModesProvided);
+                }
+
+                var idMode = context.ActionDescriptor.DisplayName.Contains("ValidateById", StringComparison.OrdinalIgnoreCase);
+                var mode = modes[0];
+                ParseMode(mode, idMode);
+            }
+        }
+
+        public static void ParseMode(string mode, bool idMode)
```
namadabu, line 40: nit: Instead of exposing this (and the other methods `ValidateId` and `ValidateType` ) as `public static` would it be cleaner to consolidate all these into a helper class (something like `ValidationHelpers`) since they are all related to validation?

src/Microsoft.Health.Fhir.Api/Features/Filters/ValidationModeFilterAttribute.cs
LTA-Thinking: Check added

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ValidateController.cs
LTA-Thinking: Good catch, I reordered the timing of the check above this so it would catch a profile in the parameters.

src/Microsoft.Health.Fhir.Api/Features/Filters/ValidationModeFilterAttribute.cs
LTA-Thinking: Added

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ValidateController.cs
LTA-Thinking: Ignore my above comment. I refactored the mode filter to do all the work.

src/Microsoft.Health.Fhir.Api/Features/Filters/ValidationModeFilterAttribute.cs
LTA-Thinking: I refactored the controller to no longer need access to these methods, so I made them private.

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/ValidationQueryFilterAndParameterParserAttribute.cs
```diff
@@ -0,0 +1,149 @@
...
+using Microsoft.Health.Fhir.Core.Features;
+using Microsoft.Health.Fhir.Core.Features.Operations;
+using Microsoft.Health.Fhir.Core.Features.Persistence;
+using Microsoft.Health.Fhir.Core.Features.Validation;
+using Microsoft.Health.Fhir.Core.Models;
+
+namespace Microsoft.Health.Fhir.Api.Features.Filters
+{
+    [AttributeUsage(AttributeTargets.Class)]
+    internal class ValidationQueryFilterAndParameterParserAttribute : ActionFilterAttribute
```
brendankowitz, line 24: Can some/most of this be moved to the Validator in the Mediatr pipeline? e.g. https://github.com/microsoft/fhir-server/blob/master/src/Microsoft.Health.Fhir.Core/Features/Operations/Validate/ValidateResourceOperationValidator.cs

The feature flag could stay in the api level

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/ValidateResourceTypeFilterAttribute.cs
```diff
@@ -24,13 +24,18 @@ public override void OnActionExecuting(ActionExecutingContext context)
...
                 context.ActionArguments.TryGetValue(KnownActionParameterNames.Resource, out var parsedModel))
             {
-                if (!string.Equals((string)actionModelType, ((Base)parsedModel).TypeName, StringComparison.OrdinalIgnoreCase))
-                {
-                    throw new ResourceNotValidException(new List<ValidationFailure>
+                ValidateType((Resource)parsedModel, (string)actionModelType);
+            }
+        }
+
+        private static void ValidateType(Resource resource, string expectedType)
```
namadabu, line 31: Since this is no longer being used outside of this filter, can we make it private to this class (or revert to the way it was before)?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ValidateTests.cs
```diff
@@ -139,6 +139,241 @@ public async void GivenAValidateByIdRequest_WhenTheResourceIdIsMissing_ThenADeta
...
+            Assert.Single(outcome.Issue);
+            CheckOperationOutcomeIssue(
+                outcome.Issue[0],
+                OperationOutcome.IssueSeverity.Information,
+                OperationOutcome.IssueType.Informational,
+                Success);
+        }
+
+        [Fact]
+        public async void GivenAValidateRequest_WhenAnInvlaidResourceIsPassedByParameter_ThenADetailedErrorIsReturned()
```
namadabu, line 158: nit: typo - Invalid

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ValidateTests.cs
```diff
@@ -139,6 +139,241 @@ public async void GivenAValidateByIdRequest_WhenTheResourceIdIsMissing_ThenADeta
...
+        [InlineData("UPDATE", "Resources can not be validated for update or delete at the default endpoint. An id must be provided in the URL.", OperationOutcome.IssueType.Invalid)]
+        [InlineData("DELETE", "Resources can not be validated for update or delete at the default endpoint. An id must be provided in the URL.", OperationOutcome.IssueType.Invalid)]
+        [InlineData("invalid", "invalid is not a valid validation mode.", OperationOutcome.IssueType.Invalid)]
+        public async void GivenAValidateRequest_WhenAModeIsPassed_ThenAnExceptionIsReturned(string mode, string message, OperationOutcome.IssueType issueType)
+        {
+            var payload = "{\"resourceType\": \"Patient\", \"id\": \"123\"}";
+
+            try
+            {
+                await _client.ValidateAsync("Patient/$validate?mode=" + mode, payload);
```
namadabu, line 184: We generally use `Assert.ThrowsAsync<FhirException>()` when trying to validate that an exception is thrown. This also avoids having the `Assert.True(false)` below (in order to fail the test if the exception is not thrown). You can use something like `FhirException ex = await Assert.ThrowsAsync<FhirException>()` and then validate on `ex`.

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/ValidationQueryFilterAndParameterParserAttribute.cs
```diff
@@ -0,0 +1,149 @@
...
+    {
+        private readonly FeatureConfiguration _features;
+
+        public ValidationQueryFilterAndParameterParserAttribute(IOptions<FeatureConfiguration> features)
+        {
+            EnsureArg.IsNotNull(features, nameof(features));
```
namadabu, line 30: nit: both validations can be done in one call by checking for `features?.Value`

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/ValidationQueryFilterAndParameterParserAttribute.cs
```diff
@@ -0,0 +1,149 @@
...
+            }
+
+            if (context.RouteData.Values.TryGetValue(KnownActionParameterNames.ResourceType, out var actionModelType) &&
+               context.ActionArguments.TryGetValue(KnownActionParameterNames.Resource, out var parsedModel))
+            {
+                if (((Resource)parsedModel).ResourceType == ResourceType.Parameters)
+                {
+                    var resource = ParseParameters((Parameters)parsedModel, ref profile, ref mode);
+
+                    context.ActionArguments[KnownActionParameterNames.Resource] = resource;
```
namadabu, line 78: Just to confirm what's happening here, we want to replace the original resource in the request (which is a Paramter resource) with the resource that was included as part of the Parameter resource.

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/ValidationQueryFilterAndParameterParserAttribute.cs
```diff
@@ -0,0 +1,149 @@
...
+using Microsoft.Health.Fhir.Api.Features.Routing;
+using Microsoft.Health.Fhir.Core.Features;
+using Microsoft.Health.Fhir.Core.Features.Operations;
+using Microsoft.Health.Fhir.Core.Features.Persistence;
+using Microsoft.Health.Fhir.Core.Features.Validation;
+using Microsoft.Health.Fhir.Core.Models;
+
+namespace Microsoft.Health.Fhir.Api.Features.Filters
+{
+    [AttributeUsage(AttributeTargets.Class)]
```
namadabu, line 23: UTs for this class?

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/ValidationQueryFilterAndParameterParserAttribute.cs
brendankowitz: I saw this as well, I think its so it validates with the `ValidateResourceTypeFilter`.

Perhaps a better option is to update the filter to allow this? `ValidateResourceTypeFilter(AllowParametersResource=true)`?

LTA-Thinking: That would allow the type filter to pass. I will also need to do this for the id filter, as it expects the id to be at the top level. My original idea with switching the value of the resource was so that as few areas as possible would need to know about whether it wrapped in a parameter resource.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ValidateTests.cs
LTA-Thinking: Changed

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ValidateTests.cs
LTA-Thinking: Fixed

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/ValidationQueryFilterAndParameterParserAttribute.cs
LTA-Thinking: Changed

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/ValidationQueryFilterAndParameterParserAttribute.cs
LTA-Thinking: I'll add some unit tests. Good to have double coverage.

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/ValidationQueryFilterAndParameterParserAttribute.cs
brendankowitz: If there are a lot of exceptions you're working around, I'd remove the Filter from the Controller level. It could be applied at the Action level if it makes sense.

Also, we could split the actions so a regular resource and Parameter resource go to different action methods, this could be done with IActionSelector.

Issue #863, opened 2020-01-30
Unstable Crucible E2E tests 


brendankowitz:
We should add these tests to the KnownBroken list as they are creating instability in CI builds:

```
connectathon-15-patient-fhirserver-03-read-server-id-json/01-ReadPatient
connectathon-15-patient-fhirserver-03-read-server-id-xml/01-ReadPatient
connectathon-15-patient-fhirserver-05-vread-server-id-json/02-VersionReadUpdatedPatient
```

PR #864, opened 2020-01-30
Adds come unstable crucible tests to knownbroken lists

brendankowitz:
Refs #863


LTA-Thinking:
I was having issues with some of these tests in my validate story.

I see some of the broken tests have links to github issues about why they are broken, but a lot more have nothing. Is it worthwhile to have a story to investigate why these tests aren't passing and log issues for them if needed?

brendankowitz:
@LTA-Thinking yes we should, right now I need a stable CI build/deployment. I'll make an issue for this

PR #871, opened 2020-02-04
Handle OperationCancelledException() thrown during health check

namadabu:
## Description
Our current health check implementation has a race condition which can end up causing an unhandled exception to be thrown. This happens when we are calling WaitAsync(cancellationToken) on the semaphoreslim and the cancellation token gets cancelled before the method returns. The WaitAsync method throws an OperationCanceledException in such scenarios and we don't catch it.

## Related issues
Addresses [issue [AB#72115](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/72115)].

## Testing
Added UT for corresponding change.

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Api/Modules/HealthChecks/CachedHealthCheck.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Api/Modules/HealthChecks/CachedHealthCheck.cs
@@ -51,7 +51,7 @@ namespace Microsoft.Health.Fhir.Api.Modules.HealthChecks
             catch (OperationCanceledException oce) when (cancellationToken.IsCancellationRequested)
             {
                 _logger.LogInformation(oce, $"Cancellation was requested for {nameof(CheckHealthAsync)}");
-                return HealthCheckResult.Healthy("Health check was canceled before completion.");
+                return _lastResult;
             }
 
             try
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Fhir.Shared.Api/Modules/HealthChecks/CachedHealthCheck.cs
```diff
@@ -44,7 +44,16 @@ public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context
...
 
-            await _semaphore.WaitAsync(cancellationToken);
+            try
+            {
+                await _semaphore.WaitAsync(cancellationToken);
+            }
+            catch (OperationCanceledException oce) when (cancellationToken.IsCancellationRequested)
+            {
+                _logger.LogInformation(oce, $"Cancellation was requested for {nameof(CheckHealthAsync)}");
+                return HealthCheckResult.Healthy("Health check was canceled before completion.");
```
namadabu: Open to suggestions on a better status response to return.

brendankowitz: Likely, if the request has been cancelled the client is not waiting for the response

namadabu: Yes, it shouldn't matter what we return since the caller is not going to get the response. But we need to return something. Maybe I can just return _lastResult (which will either be assigned or default values)

jackliums: I think returning last_Result makes sense.

namadabu: Updated to return _lastResult

PR #875, opened 2020-02-07
Fix MSDeploy dependency issue

brendankowitz:
## Description
Using the azuredeploy.json template sometimes fails with 

> AppGallery Deploy Failed: 'System.Threading.ThreadAbortException: Thread was being aborted. 

This is caused when the deployment starts and appsettings are applied. This then causes the app to reset aborting the deployment. This change adds the MSDeploy extension as a dependency before applying the appsettings.


hansenms:
Good one. I saw this problem today, actually.

brendankowitz:
@hansenms its causing the Integration test setup to fail regularly :(

PR #879, opened 2020-02-10
Generates a VLatest class for managing SQL schema versions

rotodd:
## Description
- Updates Code Gen to generate a `VLatest` class
- Refactors current code to make calls to `VLatest` (opposed to `V1` or `V2`)

This will prevent us from having to change numerous files when updating versions moving forward.

Next up, I'm planning on looking into updating Code Gen to detect VLatest automatically in [#AB72188](https://microsofthealth.visualstudio.com/Health/_workitems/edit/72188)

## Related issues
Addresses [#AB72041](https://microsofthealth.visualstudio.com/Health/_workitems/edit/72041).


src/Microsoft.Health.Fhir.SqlServer/Microsoft.Health.Fhir.SqlServer.csproj
```diff
@@ -25,6 +25,11 @@
       <Namespace>Microsoft.Health.Fhir.SqlServer.Features.Schema.Model</Namespace>
       <Args>$([System.IO.Path]::GetFullPath('$(MSBuildThisFileDirectory)\Features\Schema\Migrations\2.sql'))</Args>
     </Generated>
+    <Generated Include="Features\Schema\Model\VLatest.Generated.cs">
```
johnstairs: I can't really think of a scenario where we would need both V2 and VLatest classes. When V3 rolls around, we may have code that that conditionally references V2, in which case we would include codegen for it.

Similarly, I don't think we have any code referencing V1 anymore, so we can drop codegen for it.

rotodd: Right, fixed in 9f72df5. That's good to know for [#AB72041](https://microsofthealth.visualstudio.com/Health/_workitems/edit/72041).

Issue #334, opened 2019-01-14
Audit log might not log the correct resource type under certain circumstances


jackliums:
**Describe the bug**
Audit log might not log the correct resource type under certain circumstances.

_Non-existing resource_

**To Reproduce**
1. Make a get call to a non-existing resource

**Expected behavior**
An audit log entry is created with the requested resource type.

**Actual behavior**
Resource type is being logged as OperationOutcome.

_Deleting a resource_

**To Reproduce**
1. Create a resource
2. Delete the resource created in step 1.

**Expected behavior**
An audit log entry is created with the requested resource type.

**Actual behavior**
Resource type is being logged as null.

PR #884, opened 2020-02-13
RBAC restructuring

johnstairs:
Restructuring the RBAC model to align with Azure RBAC and for future data slice definitions, allowing a finer-grained access control. For now, there is no new functionality, but a number of changes:

1. Access control is moved out of the asp.net authorization system and instead enforced in the request handlers. 
1. Role definitions are specified in a roles.json file, outside of the asp.net core configuration system. The config binder swallows errors and the way the configuration system merges arrays from different sources can lead to expected results. So we load and validate from a dedicated file. The schema of the role definitions is also changing a bit.
1. Fixes audit logging issue #334.

Addresses #AB72271


src/Microsoft.Health.Fhir.Core/Features/Operations/Validate/ValidateOperationHandler.cs
```diff
@@ -5,21 +5,39 @@
...
+        }
+
         /// <summary>
         /// Handles validation requests that produced no errors. All validation is preformed before this is called.
         /// </summary>
         /// <param name="request">The request</param>
         /// <param name="cancellationToken">The CancellationToken</param>
         public Task<ValidateOperationResponse> Handle(ValidateOperationRequest request, CancellationToken cancellationToken)
         {
+            if (_authorizationService.CheckAccess(ResourceActions.ResourceValidate) != ResourceActions.ResourceValidate)
```
brendankowitz: I really like that we're moving these checks into the handlers. A possible enhancement might be to have the check access logic at a pre-processor level. Similar to the checks we do for validation and conformance (https://github.com/microsoft/fhir-server/blob/master/src/Microsoft.Health.Fhir.Core/Features/Validation/ValidateCapabilityPreProcessor.cs).

If the ResourceActions are static perhaps we could attribute the messages?

johnstairs: (PR is still a draft)
They are not entirely static. For instance, for a u PUT you must have `update` but `create` is optional. Also, for bundles and transactions, I'd like to perform one check for all actions that are actually needed (but I'm not sure we have that capability yet). 
But we could do the preprocessor if we pass the response to the handler. But then again, when we implement slices, we'll need more from the auth service (to determine if a create is within a slice for instance)

hansenms:
Overall, this looks great. Will take a close pass, but one high level naming thing, should we call it `dataActions` and `notDataActions`. It might seem a bit awkward here, but I am a bit concerned if there is confusion with the terms used in Azure RBAC where actions/notActions refer to control plane and dataActions/notDataActions refer to data plane.

hansenms:
The templates will need to be updated too. On that, it might be a bit confusing that the web apps are picking up most of their settings in the FHIR version specific web appsettings.json file, while the roles come from the roles.json file in the shared project.

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Security/RoleLoaderTests.cs
```diff
@@ -0,0 +1,224 @@
...
+                            name = "abc",
+                            dataActions = new[] { "*" },
+                            notDataActions = new[] { "hardDelete" },
+                            scopes = new[] { "/a" },
+                        },
+                    },
+                },
+            };
+
+            yield return new object[]
```
hansenms: Duplicate

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Security/RoleLoaderTests.cs
```diff
@@ -0,0 +1,224 @@
...
+using Microsoft.Health.Fhir.Core.Features.Settings;
+using Newtonsoft.Json.Linq;
+using NSubstitute;
+using Xunit;
+
+namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Security
+{
+    public class RoleLoaderTests
+    {
+        public static IEnumerable<object[]> GetInvalidRoles()
```
hansenms: Should be add one where there are two (or more) entries in scope?

src/Microsoft.Health.Fhir.Core/Features/Security/Role.cs
```diff
@@ -3,41 +3,22 @@
...
 
-using System.Collections.Generic;
-using System.ComponentModel.DataAnnotations;
-using System.Globalization;
+using EnsureThat;
 
 namespace Microsoft.Health.Fhir.Core.Features.Security
 {
-    public class Role : IValidatableObject
+    public class Role
```
hansenms: Wondering if we should add scope to `Role` even though it would always be `{ "/" }`. This to make it clearer later when we get to using it.

docs/Roles.md
```diff
@@ -1,96 +1,39 @@
...
+The set of data actions that can be part of a role definition are:
 
-In the app settings the nested array structure must be flattened and added to the `FhirServer:Security:Authorization:Roles` section of the configuration. Specifically, a roles array like:
+- `*` allows all data actions
+- `read` is required for reading and searching resources.
+- `write`is required for creating or updating resources.
+- `delete` is required for deleting resources. Hard-deleting requires this in addition to `hardDelete`
+- `hardDelete` is required, in addition to `delete`, for hard-deleting data.
+- `export` is required for exporting data, in addition to `read`
+- `resourceValidate` is required for invoking the [validate resource operation](https://www.hl7.org/fhir/operation-resource-validate.html).
```
deepak-ms: Will it be possible to read (or at least confirm the existence of) specific data on server as part of validate? If yes, would it be better for 'read' to include 'resourceValidate'?

johnstairs:
> The templates will need to be updated too. On that, it might be a bit confusing that the web apps are picking up most of their settings in the FHIR version specific web appsettings.json file, while the roles come from the roles.json file in the shared project.

Fixed. appsettings are now in the shared directory.

docs/Roles.md
johnstairs: If validation performs a requires a read (e.g. for referential integrity checking), it will require read permission.

PR #885, opened 2020-02-13
Docker readme tweak

gotdan:
Add note about SQL server password requirements, since if not met will lead to failed user creation and a broken server instance

### DIFF
--- a/samples/docker/README.md
+++ b/samples/docker/README.md
@@ -6,7 +6,7 @@ The following instructions detail how to build and run the FHIR Server in Docker
 
 ## Build and run with SQL Server using Docker Compose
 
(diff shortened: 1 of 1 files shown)


marrobi:
@gotdan good shout. Expect @hansenms will pick this up.

Interested to know if you found the Docker assets useful, and what was the use case?

Hoping we get some official images soon... as per https://github.com/microsoft/fhir-server/pull/840

gotdan:
@marrobi - I set up the docker instance as part of the dev environment for a FHIR cohort query tool I'm building and apart from running into the password issue it went smoothly and was a great fit for the use case. 

I also noticed that the default resources per bundle config of 500 is too low to handle sample data generated by Synthea (though the error message was clear and this was easy to change). Since loading Synthea data is probably a very common scenario for a test instance of the FHIR server, you may want to consider specifically documenting how to change the limit or raising the default to 1000 resources.

brendankowitz:
@brandonpollett I know you've done some perf testing for bundles, though this is also good feedback in https://github.com/microsoft/fhir-server/pull/885#issuecomment-586314063

brendankowitz:
Thanks for the contribution @gotdan (and @marrobi)!

Issue #894, opened 2020-02-20
Filter search parameters from search-parameters.json that are not supported


brendankowitz:
**User story**
Search parameters that are not supported by the system can be filtered out of the search-parameters.json file.
We should then use the latest/unmodified version of the file.

PR #900, opened 2020-02-26
Filter unsupported search parameters from CapabilityStatement

brendankowitz:
## Description
Filter unsupported search parameters.

- Adds "SupportedSearchParameterDefinitionManager" - all search parameter supported by the system
- Adds "SearchableSearchParameterDefinitionManager" - search parameters that have been indexed and ready to be exposed/searched

## Related issues
Addresses #894.

## Testing
Added unit tests that resolves all search parameter FHIR Path expressions and checks that the type has an index converter.


src/Microsoft.Health.Fhir.R4.Core/Features/Definition/unsupported-search-parameters.json
```diff
@@ -0,0 +1,131 @@
...
+        "http://hl7.org/fhir/SearchParameter/StructureDefinition-base",
+        "http://hl7.org/fhir/SearchParameter/StructureDefinition-base-path",
+        "http://hl7.org/fhir/SearchParameter/StructureDefinition-ext-context",
+        "http://hl7.org/fhir/SearchParameter/StructureDefinition-path",
+        "http://hl7.org/fhir/SearchParameter/StructureDefinition-valueset",
+        "http://hl7.org/fhir/SearchParameter/Subscription-url",
+        "http://hl7.org/fhir/SearchParameter/SubstanceSpecification-code",
+        "http://hl7.org/fhir/SearchParameter/TestScript-version"
+    ],
+    "PartialSupport": [
```
LTA-Thinking: Should this be in a different file, or maybe change the name of the file, since the name implies that only unsupported parameters are in this file?

src/Microsoft.Health.Fhir.Shared.Api/Modules/SearchModule.cs
```diff
@@ -37,8 +37,27 @@ public void Load(IServiceCollection services)
...
+                .Singleton()
+                .AsSelf();
+
+            services.Add<SupportedSearchParameterDefinitionManagerFactory>(c => c.GetRequiredService<SupportedSearchParameterDefinitionManager>)
+                .Singleton()
+                .AsSelf();
+
+            services.Add<SupportedSearchParameterDefinitionManager>()
+                .Singleton()
+                .AsSelf();
```
LTA-Thinking: Duplicate service added

LTA-Thinking:
What is the benefit of using a factory model for making the SearchableSearchParameterDefinitionManager and SupportedSearchParameterDefinitionManager? In the SearchModule the 'Factories' are just functions that return the singleton services, they do no work. Is there a thought that in the future the factory methods would need to modify the managers?

brendankowitz:
> What is the benefit of using a factory model for making the SearchableSearchParameterDefinitionManager and SupportedSearchParameterDefinitionManager? In the SearchModule the 'Factories' are just functions that return the singleton services, they do no work. Is there a thought that in the future the factory methods would need to modify the managers?

The factories are named delegates that return `ISearchParameterDefinitionManager`, there is no reference to the concrete implementation so usage/testing stays the same. 
I see 3 usage patterns in the future:
1. SearchParameterDefinitionManager that contains all search parameters known to the system
1. SearchParameterDefinitionManager that contains search parameters that are `supported` and can be indexed
1. SearchParameterDefinitionManager that contains search parameters that are `searchable` meaning they have been fully indexed

There are now 3 ISearchParameterDefinitionManagers in the system, named delegates is a way of asking for a specific one without knowing the concrete type.

I started updating individual usages but soon realized it's easy to miss, it seemed better to have this controlled by a specialized instance of the SearchParameterDefinitionManager.

Your right in perhaps factory is not the correct name in this case, in this case perhaps Resolver would be better. Usually I think Func/Factory.

src/Microsoft.Health.Fhir.R4.Core/Features/Definition/unsupported-search-parameters.json
brendankowitz: I see your point, but is it PartiallySupported or PartiallyUnsupported?

LTA-Thinking: Reading though the rest of the PR I agree this is the best place for them to keep all the references in one place. The file name still bugs me, but I can't think of anything better.

src/Microsoft.Health.Fhir.Core/Features/Conformance/SystemConformanceProvider.cs
```diff
@@ -29,15 +29,15 @@ public sealed class SystemConformanceProvider
 
         public SystemConformanceProvider(
             IModelInfoProvider modelInfoProvider,
-            ISearchParameterDefinitionManager searchParameterDefinitionManager,
+            SearchableSearchParameterDefinitionManagerResolver searchParameterDefinitionManagerResolver,
```
YazanMSFT: Is there a reason to switch this from taking in an Interface to taking in the concrete class?

YazanMSFT: Oh, I see this was discussed above :)

src/Microsoft.Health.Fhir.Shared.Core/Features/Definition/SearchableSearchParameterDefinitionManager.cs
```diff
@@ -0,0 +1,79 @@
...
+using Microsoft.Health.Fhir.Core.Features.Search;
+using Microsoft.Health.Fhir.Core.Models;
+using Microsoft.Health.Fhir.ValueSets;
+
+namespace Microsoft.Health.Fhir.Core.Features.Definition
+{
+    /// <summary>
+    /// A SearchParameterDefinitionManager that only returns actively searchable parameters.
+    /// If a non-searchable parameter is requested a <see cref="SearchParameterNotSupportedException"/> will be raised.
+    /// </summary>
```
YazanMSFT: This seems to be specific to GetSearchParameter(string resourceType, string name)

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/SearchParameters/SearchParameterToTypeResolver.cs
```diff
@@ -0,0 +1,378 @@
...
+using System.Text;
+using Hl7.Fhir.Introspection;
+using Hl7.Fhir.Model;
+using Hl7.FhirPath.Expressions;
+using Microsoft.Health.Fhir.Core.Models;
+using EnumerableReturnType=System.Collections.Generic.IEnumerable<Microsoft.Health.Fhir.Core.UnitTests.Features.Search.SearchParameterTypeResult>;
+using Expression = Hl7.FhirPath.Expressions.Expression;
+using Range = Hl7.Fhir.Model.Range;
+using ReturnType=Microsoft.Health.Fhir.Core.UnitTests.Features.Search.SearchParameterTypeResult;
+using SearchParamType = Microsoft.Health.Fhir.ValueSets.SearchParamType;
```
YazanMSFT: >e =  [](start = 20, length = 4)

Nit: any reason why some have spaces around the equals and others don't?

src/Microsoft.Health.Fhir.Shared.Core/Features/Definition/SearchableSearchParameterDefinitionManager.cs
brendankowitz: Removed the comment

PR #902, opened 2020-02-28
Sets parsing error message if exception message is null on model validation

rotodd:
## Description
Sometimes, when model validation for create requests fails, we return a 500. This is due to to a null or empty exception message being added to the model.

This change sets the error message to `"Error occurred when parsing model."` in this scenario.

## Related issues
Addresses [#AB72062](https://microsofthealth.visualstudio.com/Health/_workitems/edit/72062).

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Api/Features/Formatters/FhirJsonInputFormatter.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Api/Features/Formatters/FhirJsonInputFormatter.cs
@@ -95,8 +95,10 @@ namespace Microsoft.Health.Fhir.Api.Features.Formatters
 
                 if (delayedException != null)
                 {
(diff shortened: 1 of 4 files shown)


src/Microsoft.Health.Fhir.Shared.Api/Features/Formatters/FhirJsonInputFormatter.cs
```diff
@@ -95,8 +95,10 @@ public override Task<InputFormatterResult> ReadRequestBodyAsync(InputFormatterCo
 
                 if (delayedException != null)
                 {
+                    var errorMessage = string.IsNullOrEmpty(delayedException.Message) ? Api.Resources.ParsingError : delayedException.Message;
+
```
rotodd: @namadabu I haven't found a nice way to test this, since we aren't sure what the root cause of the null/empty exception messages is. Please let me know if you had something in mind.

PR #925, opened 2020-03-09
Create instance schema information

rbans96:
## Description
Create SQL schema and migration script for InstanceSchema information

## Related issues
Addresses #72783

## Testing
Manually tested.
Version can be seen upgraded to 3 and SchemaInformation table is created successfully for STU3 and R4
image
image

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/3.diff.sql
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/3.diff.sql
@@ -1,4 +1,21 @@
 /*************************************************************
+    Instance Schema
+**************************************************************/
+CREATE TABLE dbo.InstanceSchema
+(
+    Name varchar(64) COLLATE Latin1_General_100_CS_AS NOT NULL,
+    CurrentVersion int NOT NULL,
+    MaxVersion int NOT NULL,
+    MinVersion int NOT NULL,
+    Timeout datetime2(7) NOT NULL
+)
+
+CREATE UNIQUE CLUSTERED INDEX IXC_InstanceSchema ON dbo.InstanceSchema
+(
+    Name
+)
+
+/*************************************************************
     Stored procedures for InstanceSchema
 **************************************************************/
 --
@@ -32,7 +49,7 @@ AS
     SET XACT_ABORT ON
     BEGIN TRANSACTION
 
-    DECLARE @timeout datetime2(7) = SYSUTCDATETIME()
+    DECLARE @timeout datetime2(7) = DATEADD(minute, 2, SYSUTCDATETIME())
 
     INSERT INTO dbo.InstanceSchema
         (Name, CurrentVersion, MaxVersion, MinVersion, Timeout)
@@ -78,15 +95,15 @@ GO
 -- PARAMETERS
 --     @name
 --         * The unique name for a particular instance
---     @currentVersion
---         * The current version of the schema that the given instance is using
+--     @maxVersion
+--         * The maxium supported schema version for the given instance
 --
 -- RETURN VALUE
 --     The row version of the updated instance schema.
 --
 CREATE PROCEDURE dbo.UpdateInstanceSchema
     @name varchar(64),
-    @currentVersion int
+    @maxVersion int
     
 AS
     SET NOCOUNT ON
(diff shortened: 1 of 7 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/3.diff.sql
```diff
@@ -0,0 +1,153 @@
...
+--         * The unique name for a particular instance
+--     @maxVersion
+--         * The maxium supported schema version for the given instance
+--
+-- RETURN VALUE
+--     The row version of the updated instance schema.
+--
+CREATE PROCEDURE dbo.UpdateInstanceSchema
+    @name varchar(64),
+    @maxVersion int
```
brendankowitz: Should we also take in Current and Min versions? These might also change because of a code update

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/3.diff.sql
```diff
@@ -0,0 +1,153 @@
...
+    BEGIN
+        DECLARE @timeout datetime2(7) = DATEADD(minute, 2, SYSUTCDATETIME())
+
+        UPDATE dbo.InstanceSchema
+        SET MaxVersion = @maxVersion, Timeout = @timeout
+        WHERE Name = @name
+    END
+    ELSE
+    BEGIN
+        THROW 50404, 'Instance record not found', 1;
```
brandonpollett: What is the expected behavior of the instance if this happens?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/3.diff.sql
```diff
@@ -0,0 +1,153 @@
...
+    END
+  
+    SELECT MIN_ACTIVE_ROWVERSION()
+
+    COMMIT TRANSACTION
```
brandonpollett: Do we begin a transaction?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/3.diff.sql
```diff
@@ -0,0 +1,153 @@
...
+        UPDATE dbo.InstanceSchema
+        SET MaxVersion = @maxVersion, Timeout = @timeout
+        WHERE Name = @name
+    END
+    ELSE
+    BEGIN
+        THROW 50404, 'Instance record not found', 1;
+    END
+  
+    SELECT MIN_ACTIVE_ROWVERSION()
```
brandonpollett: Do we actually need to select anything?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/3.diff.sql
```diff
@@ -0,0 +1,153 @@
+/*************************************************************
+    Instance Schema
+**************************************************************/
+CREATE TABLE dbo.InstanceSchema
+(
+    Name varchar(64) COLLATE Latin1_General_100_CS_AS NOT NULL,
+    CurrentVersion int NOT NULL,
+    MaxVersion int NOT NULL,
+    MinVersion int NOT NULL,
+    Timeout datetime2(7) NOT NULL
```
brandonpollett: Do we need/want this much precision?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/3.diff.sql
rbans96: Good point. Agree for Current versions but I think Min version should not be updated. It should remain the same as specified by the code. Thoughts?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/3.diff.sql
rbans96: Good point! Since for now considering 2 mins for timeout, we even dont need milliseconds. Fixed it to datetime2(0).

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/3.diff.sql
rbans96: I think it should upsert instead update. Fixed it in the last commit. Let me know if you have other thoughts.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/3.diff.sql
rbans96: Removed in the last commit

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/3.diff.sql
rbans96: Good catch, we don't need transaction here. Updated in the last commit

rbans96:
Closing this PR as new PR is created to merge these changes into feature branch

rbans96:
> Should this code go into master or into a feature branch?

Thanks for bringing this up. Created new PR for these changes to go into feature branch.

PR #928, opened 2020-03-10
Adding retries for export when auth token expires

namadabu:
## Description
When exporting large amounts of data, we can run into an issue when the current auth token for the export client can expire. This change adds a retry mechanism where we attempt to get a new auth token and then continue export with the newly authorized export client

## Related issues
Addresses [issue [AB#72260](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/72260)].

## Testing
Existing tests, some UTs. Manually tested by expiring the access key of a test storage account (while export was running) and validated that a new token was retrieved to complete export.

### DIFF
--- a/src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/AzureExportDestinationClient.cs
+++ b/src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/AzureExportDestinationClient.cs
@@ -78,7 +78,7 @@ namespace Microsoft.Health.Fhir.Azure.ExportDestinationClient
             {
                 _logger.LogWarning(se, se.Message);
 
-                HttpStatusCode responseCode = ParseStorageException(se);
+                HttpStatusCode responseCode = StorageExceptionParser.ParseStorageException(se);
                 throw new DestinationConnectionException(se.Message, responseCode);
             }
         }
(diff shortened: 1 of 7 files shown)


src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/AzureExportDestinationClient.cs
```diff
@@ -119,9 +119,66 @@ public async Task CommitAsync(CancellationToken cancellationToken)
         {
             CheckIfClientIsConnected();
 
+            try
+            {
+                await Policy.Handle<StorageException>()
+                    .WaitAndRetryAsync(
+                        retryCount: 2,
+                        sleepDurationProvider: (retryCount) => TimeSpan.FromSeconds(5 * (retryCount - 1)),
```
brendankowitz: In other places in the code we have an 'exponential retry' policy defined as `.WaitAndRetryAsync(3, retryAttempt => TimeSpan.FromSeconds(Math.Pow(2, retryAttempt)))`, so the delay starts shorter and backs off exponentially, this one looks linear.
Perhaps we could create an extension method on `PolicyBuilder` that allows us to reference the pattern by name (ExponentialBackoffRetry, LinearBackoffRetry)?

src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/AzureExportDestinationClient.cs
```diff
@@ -205,11 +265,44 @@ private HttpStatusCode ParseStorageException(StorageException storageException)
                 }
                 catch (Exception)
```
brendankowitz: `Enum.TryParse` seems safer than catching a generic exception

namadabu: I thought of using `Enum.TryParse` initially, but then realized that the result would require further validation. As an example, `Enum.TryParse(intValue.ToString(), out HttpStatusCode responseCode)` with `intValue = 1` would return `True` and `responseCode = 1` even though it is an invalid HttpStatusCode. Hence I decided to use `IsDefined` since it would be easier and is the recommended solution in such situations (when converting int to string) - https://docs.microsoft.com/en-us/dotnet/api/system.enum.tryparse?view=netcore-3.1

brendankowitz: Ah got it, wasn't obvious that https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.storage.requestresult.httpstatuscode?view=azure-dotnet#Microsoft_Azure_Storage_RequestResult_HttpStatusCode is an int :)
If this is the case, and since you are already using IsDefined, any reason we shouldn't just cast? `(HttpStatusCode)intValue`?

namadabu: The StorageException class does seem a bit weird :). Yes, I should be able to just cast. Will update.

PR #931, opened 2020-03-10
Implements compatibility api

rbans96:
## Description
Implements the _schema/compatibility to return min and max compatible schema versions from the server
Adds stored procedure SelectMaxSupportedSchemaVersion

## Related issues
User Story [AB#72833](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/72833).

## Testing
Unit test and manual tests
image


src/Microsoft.Health.Fhir.SqlServer.Api/Controllers/SchemaController.cs
```diff
@@ -82,7 +87,15 @@ public ActionResult Compatibility()
         {
             _logger.LogInformation("Attempting to get compatibility");
 
-            throw new NotImplementedException(Resources.CompatibilityNotImplemented);
+            var minVersion = (int)_schemaInformation.MinimumSupportedVersion;
+            var compatibilityVersions = new Dictionary<string, object> { { "min", minVersion } };
```
brendankowitz: I think we should create POCO models going forward

src/Microsoft.Health.Fhir.Core/Features/Schema/IQueryProcessor.cs
```diff
@@ -0,0 +1,17 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Fhir.Core.Features.Schema
+{
+    public interface IQueryProcessor
```
brendankowitz: Using a Mediatr handler is the common pattern we currently have for calling functionality from the controller, (maybe `LatestSchemaVersionHandler`).

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/2.diff.sql
```diff
@@ -237,3 +237,28 @@ AS
...
+--  DESCRIPTION
+--      Selects the highest supported completed schema version
+--  PARAMETERS
+--      @version
+--          * The version number
+--
+--  RETURNS
+--      The highest version
+--
+CREATE PROCEDURE dbo.SelectMaxSupportedSchemaVersion
```
brendankowitz: Should be in V3?

rbans96: Yes, since 3.sql was not merged, so kept it in V2. I will move it to V3.

src/Microsoft.Health.Fhir.Core/Features/Schema/LatestSchemaVersionHandler.cs
```diff
@@ -0,0 +1,45 @@
...
+            EnsureArg.IsNotNull(fhirAuthorizationService, nameof(fhirAuthorizationService));
+            _fhirOperationDataStore = fhirOperationDataStore;
+            _fhirAuthorizationService = fhirAuthorizationService;
+        }
+
+        public async Task<GetCompatibilityVersionResponse> Handle(GetCompatibilityVersionRequest request, CancellationToken cancellationToken)
+        {
+            EnsureArg.IsNotNull(request, nameof(request));
+
+            if (await _fhirAuthorizationService.CheckAccess(DataActions.SchemaMigration) != DataActions.SchemaMigration)
```
rbans96: Due to some reason the api is throwing Authorization exception. I have added new enum SchemaMigration in DataAccess. Not sure what am I missing here :( Please suggest.

src/Microsoft.Health.Fhir.Core/Features/Operations/IFhirOperationDataStore.cs
```diff
@@ -56,5 +56,13 @@ public interface IFhirOperationDataStore
...
         /// <returns>A list of acquired export job.</returns>
         Task<IReadOnlyCollection<ExportJobOutcome>> AcquireExportJobsAsync(ushort maximumNumberOfConcurrentJobsAllowed, TimeSpan jobHeartbeatTimeoutThreshold, CancellationToken cancellationToken);
+
+        /// <summary>
+        /// Get compatible version.
+        /// </summary>
+        /// <param name="maxVersion">The maximum schema version specified by code</param>
+        /// /// <param name="cancellationToken">The cancellation token.</param>
+        /// <returns>The latest supported schema version from server.</returns>
+        Task<int> GetLatestCompatibleVersionAsync(int maxVersion, CancellationToken cancellationToken);
```
brendankowitz: In general things are looking good, some thoughts:
* Perhaps break this into its own interface `ISchemaMigrationDataStore`
* Think about how you can gracefully have the CosmosDb implementation return an unsupported status

src/Microsoft.Health.Fhir.Core/Features/Schema/LatestSchemaVersionHandler.cs
brendankowitz: @johnstairs any guidance here?
I think the controller action has `AllowAnonymous` so it will probably not be able to check roles/access for unauth'd users

rbans96: I discussed with John this morning, since schema migration will be run by admin by hitting few commands on tool, we might not need authorization check for schema migration specific apis. As of now i have removed this check. But I am all ears for any suggestions.

src/Microsoft.Health.Fhir.Core/Features/Operations/IFhirOperationDataStore.cs
rbans96: yes it was in my mind to separate this out. But I know Brandon is refactoring lot of files to break components to get sql out. So I am looking forward for his PR too. Anyways I will update it. Thank you!

src/Microsoft.Health.Fhir.Core/Features/Schema/ISchemaMigrationDataStore.cs
```diff
@@ -0,0 +1,21 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Health.Fhir.Core.Features.Schema
+{
+    public interface ISchemaMigrationDataStore
```
brandonpollett: Why does this live here instead of in the SQL packages?

src/Microsoft.Health.Fhir.Core/Features/Schema/LatestSchemaVersionHandler.cs
brandonpollett: I agree that the AuthN/AuthZ isn't needed for these APIs.

src/Microsoft.Health.Fhir.Core/Features/Operations/IFhirOperationDataStore.cs
brandonpollett: I think we should limit the schema stuff to the SQL packages. I'm not sure it has broad applicability to the Cosmos implementation and also the `_schema` routes only exist in the SQL packages.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Sql/SchemaTests.cs
```diff
@@ -54,11 +54,16 @@ public async Task WhenRequestingAvailable_GivenAServerThatHasSchemas_JsonShouldB
             Assert.Equal(scriptUrl, firstResult["script"]);
         }
 
-        [Theory]
-        [MemberData(nameof(Data))]
-        public async Task WhenRequestingSchema_GivenGetMethod_TheServerShouldReturnNotImplemented(string path)
+        [Fact]
+        public async Task WhenRequestingSchema_GivenGetMethodAndCompatibilityPath_TheServerShouldReturnSuccess()
+        {
+            await SendAndVerifyStatusCode(HttpMethod.Get, "_schema/compatibility", HttpStatusCode.OK);
```
brandonpollett: Probably want to verify that it returns the correct thing, too.

src/Microsoft.Health.Fhir.SqlServer.Api/Controllers/SchemaController.cs
```diff
@@ -78,11 +83,15 @@ public FileContentResult SqlScript(int id)
...
         [Route(KnownRoutes.Compatibility)]
-        public ActionResult Compatibility()
+        public async Task<IActionResult> Compatibility()
         {
             _logger.LogInformation("Attempting to get compatibility");
 
-            throw new NotImplementedException(Resources.CompatibilityNotImplemented);
+            int maxVersion = (int)_schemaInformation.MaximumSupportedVersion;
+            int minVersion = (int)_schemaInformation.MinimumSupportedVersion;
+            var compatibleResponse = await _mediator.GetCompatibleVersionAsync(minVersion, maxVersion, HttpContext.RequestAborted);
```
brandonpollett: This will vary based on the instance that you hit. I would think that it would be more helpful to use the current version in the DB along with the information we're recording in the instance table.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/3.sql
```diff
@@ -82,6 +82,31 @@ BEGIN
...
+--
+CREATE PROCEDURE dbo.SelectMaxSupportedSchemaVersion
+    @version int
+AS
+BEGIN
+    SET NOCOUNT ON
+
+    SELECT MAX(Version)
+    FROM SchemaVersion
+    WHERE Status = 'complete' AND Version <= @version
```
brandonpollett: This just seems like it's basically giving the current version in most cases unless you happen to hit an instance that hasn't updated to the latest version.

src/Microsoft.Health.Fhir.SqlServer.Api.UnitTests/Controllers/SchemaControllerTests.cs
```diff
@@ -58,11 +61,5 @@ public void GivenACurrentVersiontRequest_WhenNotImplemented_ThenNotImplementedSh
         {
             Assert.Throws<NotImplementedException>(() => _schemaController.CurrentVersion());
         }
-
-        [Fact]
-        public void GivenACompatibilityRequest_WhenNotImplemented_ThenNotImplementedShouldBeThrown()
```
brandonpollett: Since we're removing the error test, do we want to add a success test?

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Security/RequestHandlerCheckAccessTests.cs
```diff
@@ -50,7 +50,7 @@ void Dfs(Assembly a)
             Dfs(typeof(SecurityProvider).Assembly);
 
             return assemblies.SelectMany(a => a.GetTypes())
-                .Where(t => t.IsClass && !t.IsAbstract && t.GetInterfaces().Any(i => i.IsGenericType && i.GetGenericTypeDefinition() == typeof(IRequestHandler<,>)))
+                .Where(t => t.IsClass && string.Compare("LatestSchemaVersionHandler", t.Name, StringComparison.OrdinalIgnoreCase) != 0 && !t.IsAbstract && t.GetInterfaces().Any(i => i.IsGenericType && i.GetGenericTypeDefinition() == typeof(IRequestHandler<,>)))
```
brandonpollett: Shouldn't be needed right?

brendankowitz: There is also an `_exceptions` list above, would be a better place for this

src/Microsoft.Health.Fhir.Core/Features/Operations/IFhirOperationDataStore.cs
rbans96: Addressed it.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/3.sql
rbans96: Updated the query, please review and let me know your thoughts5

src/Microsoft.Health.Fhir.SqlServer.Api.UnitTests/Controllers/SchemaControllerTests.cs
rbans96: Test is added in the  LatestVersionSchemaHandler for success scenario  and there is no particular check in controller before it gets processed by the handler. Let me know if we still need one in controller

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Security/RequestHandlerCheckAccessTests.cs
rbans96: Its not needed as the handler is now moved to sql package

src/Microsoft.Health.Fhir.SqlServer.Api/Controllers/SchemaController.cs
```diff
@@ -24,8 +27,9 @@ public class SchemaController : Controller
         private readonly SchemaInformation _schemaInformation;
         private readonly IUrlResolver _urlResolver;
         private readonly ILogger<SchemaController> _logger;
+        private readonly IMediator _mediator;
 
-        public SchemaController(SchemaInformation schemaInformation, IUrlResolver urlResolver, ILogger<SchemaController> logger)
+        public SchemaController(SchemaInformation schemaInformation, IUrlResolver urlResolver, ILogger<SchemaController> logger, IMediator mediator)
```
brandonpollett: nit: convention is that logger should be last item

src/Microsoft.Health.Fhir.SqlServer.Api/Controllers/SchemaController.cs
```diff
@@ -78,11 +83,14 @@ public FileContentResult SqlScript(int id)
         [HttpGet]
         [AllowAnonymous]
         [Route(KnownRoutes.Compatibility)]
-        public ActionResult Compatibility()
+        public async Task<IActionResult> Compatibility()
         {
             _logger.LogInformation("Attempting to get compatibility");
 
-            throw new NotImplementedException(Resources.CompatibilityNotImplemented);
+            int minVersion = (int)_schemaInformation.MinimumSupportedVersion;
```
brandonpollett: This is going to vary based on the instance that you happen to hit if the code packages are different.

src/Microsoft.Health.Fhir.SqlServer/Features/Exceptions/RecordNotFoundException.cs
```diff
@@ -0,0 +1,26 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using EnsureThat;
+
+using Microsoft.Health.Fhir.Core.Exceptions;
+using Microsoft.Health.Fhir.Core.Models;
+
+namespace Microsoft.Health.Fhir.SqlServer.Features.Exceptions
+{
+    public class RecordNotFoundException : FhirException
```
brandonpollett: This probably shouldn't extend `FhirException` since it doesn't actually deal with FHIR.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Extensions/SchemaMediatorExtensions.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    public static class SchemaMediatorExtensions
+    {
+        public static async Task<GetCompatibilityVersionResponse> GetCompatibleVersionAsync(this IMediator mediator, int minVersion, CancellationToken cancellationToken)
+        {
+            EnsureArg.IsNotNull(mediator, nameof(mediator));
+
+            var request = new GetCompatibilityVersionRequest(minVersion);
+
+            GetCompatibilityVersionResponse response = await mediator.Send(request, cancellationToken);
```
brandonpollett: Could probably just `return await mediator.Send(request, cancellationToken);`

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/ISchemaMigrationDataStore.cs
```diff
@@ -0,0 +1,20 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Health.Fhir.SqlServer.Features.Schema
+{
+    public interface ISchemaMigrationDataStore
```
brandonpollett: Maybe just `ISchemaDataStore` or `ISchemaVersionDataStore`. Thoughts?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerSchemaMigrationDataStore.cs
```diff
@@ -0,0 +1,51 @@
...
+            EnsureArg.IsNotNull(sqlConnectionWrapperFactory, nameof(sqlConnectionWrapperFactory));
+            EnsureArg.IsNotNull(logger, nameof(logger));
+
+            _sqlConnectionWrapperFactory = sqlConnectionWrapperFactory;
+            _logger = logger;
+        }
+
+        public async Task<int> GetLatestCompatibleVersionAsync(CancellationToken cancellationToken)
+        {
+            using (SqlConnectionWrapper sqlConnectionWrapper = _sqlConnectionWrapperFactory.ObtainSqlConnectionWrapper(true))
```
brandonpollett: Does this actually need to enlist in a transaction? Is there an existing transaction?

src/Microsoft.Health.Fhir.SqlServer.Api/Controllers/SchemaController.cs
rbans96: Updated in the last commit to fetch minVersion too from InstanceSchema table

src/Microsoft.Health.Fhir.SqlServer.Api/Controllers/SchemaController.cs
```diff
@@ -34,6 +38,7 @@ public SchemaController(SchemaInformation schemaInformation, IUrlResolver urlRes
             _schemaInformation = schemaInformation;
             _urlResolver = urlResolver;
             _logger = logger;
+            _mediator = mediator;
```
brendankowitz: EnsureArg for mediator

rbans96:
This PR changes and addressed feedback are present in current version api PR because of the similar changes required.
Please refer https://github.com/microsoft/fhir-server/pull/938 for the changes

Issue #911, opened 2020-03-03
Refactor PartialDateTime to remove duplicate code


rotodd:
**User story**
The class `PartialDateTime.cs` has duplicate code for checking the formatting of the input date time string. Specifically, the RegEx in the `Parse()` method handles the majority of the checks in the `PartialDateTime` constructor.

Note 1: Tests that call this constructor will need to be updated.
Note 2: This relates to #25 and #411.

**Acceptance criteria**
1. `PartialDateTime.cs` is refactored to remove duplicate functionality.
2. Tests that call `PartialDateTime` constructor are updated.

PR #933, opened 2020-03-11
Returns error on invalid date time parameter

rotodd:
## Description

Previously, the `PartialDateTime` parser would read `datetime` input left to right, accepting the valid parts of the string until an incorrect value was reached (for examples, see #25).

Now, `400 Bad Request` is returned if any part of a `datetime` parameter is incorrectly formatted.

The new parser leverages the [DateTimeOffset.TryParseExact](https://docs.microsoft.com/en-us/dotnet/api/system.datetimeoffset.tryparseexact?view=netcore-3.1) method to parse input `datetime` strings.

With this implementation, we rely on the RegEx much less. However, it still benefits us in two ways:
1. It captures the parts of the `datetime` parameter that were not specified (the `DateTimeOffset` parser will auto-populate any information that is "missing")
2. It ensures that, if specified, the time zone hour is formatted with two digits (the `DateTimeOffset` parser allows `(+|-)h:mm`)

## Demo

A few test cases:

| Birthdate  | Is Valid? | Notes |
| ------------- | ------------- | ------------- |
| 1980-12-30T03:04:05.99Z  |  ✔ |  |
| 1980000000-12-30T03:04:05.99Z |  ❌  | Year must be formatted `yyyy` |
| 1980asdfghj-12-30T03:04:05.99Z |  ❌  | Year must be formatted `yyyy` |
| 1980-25-30T03:04:05.99Z |  ❌  | Month must be in range |
| 1980-12-32T03:04:05.99Z |  ❌  | Date must be in range |
| 1980-12-30T03:04:05.99-02:30 |  ✔  |  |
| 1980-12-30T03:04:05.99-2:30 |  ❌  | Time zone must be formatted `Z\|(+\|-)hh:mm`|
| le1980-12-30T03:04:05.99-02:30 |  ✔  |  |
| lg1980-12-30T03:04:05.99-02:30 |  ❌  | Invalid prefix |

date-time-formatting2

## Related issues

Addresses #25, #911 and [#AB72714](https://microsofthealth.visualstudio.com/Health/_workitems/edit/72714). 

## Testing

- Removes duplicate parsing logic in the `PartialDateTime` constructor that unit tests previously checked
  - These tests now call the `PartialDateTime.Parse` method
- Modifies unit tests in `SearchValueExpressionBuilderTests.cs` to call the `PartialDateTime` parser without binary prefixes, as this parsing occurs at a different layer
- Adds new unit tests
- Updates existing E2E tests where the time zone was not properly URL encoded

### DIFF
--- a/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Search/SearchValues/DateTimeSearchValueTests.cs
+++ b/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Search/SearchValues/DateTimeSearchValueTests.cs
@@ -83,7 +83,7 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Search.SearchValues
         public void GivenAPartialStartDateTime_WhenInitialized_ThenCorrectStartDateTimeShouldBeAssigned(string input, string start)
         {
             PartialDateTime inputDateTime = PartialDateTime.Parse(input);
-            PartialDateTime endDateTime = new PartialDateTime(2018);
+            PartialDateTime endDateTime = PartialDateTime.Parse("2018");
 
             DateTimeOffset expectedStartDateTime = DateTimeOffset.Parse(start);
 
@@ -97,7 +97,7 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Search.SearchValues
         public void GivenAPartialEndDateTime_WhenInitialized_ThenCorrectEndDateTimeShouldBeAssigned(string input, string start)
         {
             PartialDateTime inputDateTime = PartialDateTime.Parse(input);
-            PartialDateTime startDateTime = new PartialDateTime(2000);
+            PartialDateTime startDateTime = PartialDateTime.Parse("2000");
 
             DateTimeOffset expectedEndDateTime = DateTimeOffset.Parse(start);
 
--- a/src/Microsoft.Health.Fhir.Core.UnitTests/Models/PartialDateTimeTests.cs
+++ b/src/Microsoft.Health.Fhir.Core.UnitTests/Models/PartialDateTimeTests.cs
@@ -86,6 +86,24 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Models
         }
 
         [Theory]
+        [InlineData("2013-05-18T23:09+1:00")]
+        [InlineData("2013-05-18T23:09-4:30")]
+        [InlineData("2013-05-18T23:09+01:300")]
+        [InlineData("2013-05-18T23:09+01:77")]
+        [InlineData("2013-05-18T23:09+1")]
+        [InlineData("2013-05-18T23:09-7")]
+        [InlineData("2013-05-18T23:09-07")]
+        [InlineData("2013-05-18T23:09+Z")]
+        [InlineData("2013-05-18T23:09-Z")]
+        [InlineData("2013-05-18T23:09ZZ")]
+        [InlineData("2013-05-18T23:09-99:00")]
+        [InlineData("2013-05-18T23:09-789")]
+        public void GivenInvalidUtcOffset_WhenParsingPartialDateTime_ThenExceptionShouldBeThrown(string inputString)
+        {
+            Assert.Throws<FormatException>(() => PartialDateTime.Parse(inputString));
+        }
+
+        [Theory]
         [InlineData("0000-05-18T23:57:09.931094+01:00")] // Year cannot be less than 1.
         [InlineData("10000-05-18T23:57:09.931094+01:00")] // Year cannot be greater than 9999.
         [InlineData("2013-00-18T23:57:09.931094+01:00")] // Month cannot be less than 1.
@@ -111,6 +129,7 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Models
             yield return new object[] { "1999", 1999, null, null, null, null, null, null, null };
             yield return new object[] { "1999-10", 1999, 10, null, null, null, null, null, null };
             yield return new object[] { "1999-10-01", 1999, 10, 1, null, null, null, null, null };
+            yield return new object[] { "1999-10-18T12:35", 1999, 10, 18, 12, 35, null, null, 0 };
             yield return new object[] { "1999-10-18T12:35+01:00", 1999, 10, 18, 12, 35, null, null, 60 };
             yield return new object[] { "1999-10-18T12:35:55-02:30", 1999, 10, 18, 12, 35, 55, null, -150 };
             yield return new object[] { "1999-10-18T12:35:55Z", 1999, 10, 18, 12, 35, 55, null, 0 };
--- a/src/Microsoft.Health.Fhir.Core.UnitTests/Models/PartialDateTimeTests.cs
+++ b/src/Microsoft.Health.Fhir.Core.UnitTests/Models/PartialDateTimeTests.cs
@@ -27,33 +27,33 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Models
         [InlineData("05-18T23:57")]
         [InlineData("05-18T23:57:09")]
         [InlineData("05-18T23:57:09.931094")]
-        [InlineData("05-18T23:57%2B01:00")]
-        [InlineData("05-18T23:57:09%2B01:00")]
-        [InlineData("05-18T23:57:09.931094%2B01:00")]
+        [InlineData("05-18T23:57+01:00")]
+        [InlineData("05-18T23:57:09+01:00")]
+        [InlineData("05-18T23:57:09.931094+01:00")]
         [InlineData("2013-05T23:57")] // Month/date needs to be specified.
         [InlineData("2013-05T23:57:09")]
         [InlineData("2013-05T23:57:09.931094")]
-        [InlineData("2013-05T23:57%2B01:00")]
-        [InlineData("2013-05T23:57:09%2B01:00")]
-        [InlineData("2013-05T23:57:09.931094%2B01:00")]
+        [InlineData("2013-05T23:57+01:00")]
+        [InlineData("2013-05T23:57:09+01:00")]
+        [InlineData("2013-05T23:57:09.931094+01:00")]
         [InlineData("2013T23:57")] // Month and date need to be specified.
         [InlineData("2013T23:57:09")]
         [InlineData("2013T23:57:09.931094")]
-        [InlineData("2013T23:57%2B01:00")]
-        [InlineData("2013T23:57:09%2B01:00")]
-        [InlineData("2013T23:57:09.931094%2B01:00")]
+        [InlineData("2013T23:57+01:00")]
+        [InlineData("2013T23:57:09+01:00")]
+        [InlineData("2013T23:57:09.931094+01:00")]
         [InlineData("T23:57")] // Year, month and date need to be specified.
         [InlineData("T23:57:09")]
         [InlineData("T23:57:09.931094")]
-        [InlineData("T23:57%2B01:00")]
-        [InlineData("T23:57:09%2B01:00")]
-        [InlineData("T23:57:09.931094%2B01:00")]
+        [InlineData("T23:57+01:00")]
+        [InlineData("T23:57:09+01:00")]
+        [InlineData("T23:57:09.931094+01:00")]
         [InlineData("2013-05-18T23:09.931094")] // Hour/minute/second needs to be specified.
-        [InlineData("2013-05-18T23:09.931094%2B01:00")]
+        [InlineData("2013-05-18T23:09.931094+01:00")]
         [InlineData("2013-05-18T09.931094")] // Hour and minute need to be specified.
-        [InlineData("2013-05-18T09.931094%2B01:00")]
+        [InlineData("2013-05-18T09.931094+01:00")]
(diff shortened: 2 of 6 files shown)


src/Microsoft.Health.Fhir.Core.UnitTests/Features/Search/SearchValues/DateTimeSearchValueTests.cs
```diff
@@ -83,7 +83,7 @@ public void GivenAPartialDateTime_WhenInitialized_ThenCorrectStartDateTimeAndEnd
         public void GivenAPartialStartDateTime_WhenInitialized_ThenCorrectStartDateTimeShouldBeAssigned(string input, string start)
         {
             PartialDateTime inputDateTime = PartialDateTime.Parse(input);
-            PartialDateTime endDateTime = new PartialDateTime(2018);
```
rotodd: I made this constructor `private` - previously it was only referenced by unit tests.

src/Microsoft.Health.Fhir.Core.UnitTests/Models/PartialDateTimeTests.cs
```diff
@@ -14,172 +14,133 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Models
 {
     public class PartialDateTimeTests : IDisposable
```
rotodd: The big change made to this file was creating `PartialDateTime` objects by passing in a `datetime` string to the `PartialDateTime.Parse` method. This is the code path that is follow in a non-test environment.

For the most part, test cases remain the same.

src/Microsoft.Health.Fhir.Core/Models/PartialDateTime.cs
```diff
@@ -233,74 +139,124 @@ public PartialDateTime(DateTimeOffset dateTimeOffset)
...
+        /// <remarks>
+        /// The parser serves two different purposes: storing and searching.
+        /// There is a difference between how the date should be parsed for storing and how the date should be parsed for searching.
+        /// For date that should be stored, the time zone information must be present if time is specified.
+        /// From spec: http://hl7.org/fhir/datatypes.html#datetime, "If hours and minutes are specified, a time zone SHALL be populated."
+        /// However, if the date is being parsed for searching, then the time zone information is optional.
+        /// From spec: http://hl7.org/fhir/search.html#date, "you SHOULD provide a time zone if the time part is present."
+        /// As a result, the formats array allows the time zone to be optional.
+        /// </remarks>
+        private static string[] GenerateDateTimeOffsetFormats()
```
rotodd: Does it make sense to generate this array in a method?

src/Microsoft.Health.Fhir.Core.UnitTests/Models/PartialDateTimeTests.cs
```diff
@@ -14,172 +14,133 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Models
...
-            yield return new object[] { ParamNameHour, 2017, 1, 1, -1, 1, 1, 0m, 60 }; // Hour cannot be less than 0.
-            yield return new object[] { ParamNameHour, 2017, 1, 1, 24, 1, 1, 0m, 60 }; // Hour cannot be greater than 23.
-            yield return new object[] { ParamNameMinute, 2017, 1, 1, 1, -1, 1, 0m, 60 }; // Minute cannot be less than 0.
-            yield return new object[] { ParamNameMinute, 2017, 1, 1, 1, 61, 1, 0m, 60 }; // Minute cannot be greater than 59.
-            yield return new object[] { ParamNameSecond, 2017, 1, 1, 1, 1, -1, 0m, 60 }; // Second cannot be less than 0.
-            yield return new object[] { ParamNameSecond, 2017, 1, 1, 1, 1, 61, 0m, 60 }; // Second cannot be greater than 59.
-            yield return new object[] { ParamNameFraction, 2017, 1, 1, 1, 1, 1, -1m, 60 }; // Fraction cannot be less than 0.
-            yield return new object[] { ParamNameFraction, 2017, 1, 1, 1, 1, 1, 1m, 60 }; // Fraction cannot be greater than .
+            Assert.Throws<FormatException>(() => PartialDateTime.Parse(inputString));
```
LTA-Thinking: Why doesn't this one check the message like the ones above do?

src/Microsoft.Health.Fhir.Core.UnitTests/Models/PartialDateTimeTests.cs
```diff
@@ -224,19 +174,11 @@ public void GivenADateImtOffset_WhenInitialized_ThenCorrectPartialDateTimeShould
...
-            var dateTimeOffset = new DateTimeOffset(
-                year,
-                month,
-                day,
-                hour,
-                minute,
-                second,
-                millisecond,
-                utcOffset).AddTicks(fraction);
+            DateTimeOffset dateTimeOffset = new DateTimeOffset(year, month, day, hour, minute, second, millisecond, utcOffset).AddTicks(fraction);
```
LTA-Thinking: Why change the others to vars but change this one to be explicitly typed?

src/Microsoft.Health.Fhir.Core.UnitTests/Models/PartialDateTimeTests.cs
```diff
@@ -248,31 +190,32 @@ public void GivenADateImtOffset_WhenInitialized_ThenCorrectPartialDateTimeShould
...
         {
-            Assert.Throws<FormatException>(() => PartialDateTime.Parse("abc"));
+            Assert.Throws<FormatException>(() => PartialDateTime.Parse(inputString));
         }
 
         public static IEnumerable<object[]> GetParseData()
         {
-            yield return new object[] { "2017", 2017, null, null, null, null, null, null, null };
-            yield return new object[] { "2017-01", 2017, 1, null, null, null, null, null, null };
-            yield return new object[] { "2019-01-02", 2019, 1, 2, null, null, null, null, null };
```
LTA-Thinking: Are these three scenarios tested elsewhere?

src/Microsoft.Health.Fhir.Core/Models/PartialDateTime.cs
```diff
@@ -27,17 +27,18 @@ public class PartialDateTime
...
-        // This regular expression allows the time zone to be optional.
+        // This regular expression is used to capture which date time parts are specified by the user and which parts are not.
+        // This is required because date time parts left blank are used to indicate a time period.
+        // For example, 2000 is equivalent to an interval of [2000-01-01T00:00, 2000-12-31T23:59].
         private static readonly Regex DateTimeRegex = new Regex(
-            $@"-?(?<{YearCapture}>[0-9]{{4}})(-(?<{MonthCapture}>0[1-9]|1[0-2])(-(?<{DayCapture}>0[0-9]|[1-2][0-9]|3[0-1])(T(?<{HourCapture}>[01][0-9]|2[0-3]):(?<{MinuteCapture}>[0-5][0-9])(:((?<{SecondCapture}>[0-5][0-9])(?<{FractionCapture}>\.[0-9]+)?))?(?<{TimeZoneCapture}>Z|(\+|-)((0[0-9]|1[0-3]):[0-5][0-9]|14:00))?)?)?)?",
+            $@"-?(?<{YearCapture}>[0-9]{{4}})(-(?<{MonthCapture}>[0-9]{{2}}))?(-(?<{DayCapture}>[0-9]{{2}}))?(T(?<{HourCapture}>[0-9]{{2}}))?(:(?<{MinuteCapture}>[0-9]{{2}}))?(:(?<{SecondCapture}>[0-9]{{2}}))?((?<{FractionCapture}>\.[0-9]+))?((?<{TimeZoneCapture}>Z|(\+|-)(([0-9]{{2}}):[0-9]{{2}}))|(?<{InvalidTimeZoneCapture}>Z|(\+|-)(([0-9]{{1}}):[0-9]{{2}})))?",
```
LTA-Thinking: You can replace the [0-9]s with \d and the {{1}} towards the end is redundant.

src/Microsoft.Health.Fhir.Core/Models/PartialDateTime.cs
LTA-Thinking: I think this makes sense. This would be a large array to enumerate every option, and this makes typos less likely and easier to find.

src/Microsoft.Health.Fhir.Core/Models/PartialDateTime.cs
```diff
@@ -233,74 +139,124 @@ public PartialDateTime(DateTimeOffset dateTimeOffset)
...
+        /// From spec: http://hl7.org/fhir/search.html#date, "you SHOULD provide a time zone if the time part is present."
+        /// As a result, the formats array allows the time zone to be optional.
+        /// </remarks>
+        private static string[] GenerateDateTimeOffsetFormats()
+        {
+            var formats = new List<string>();
+
+            var dateFormats = new List<string> { "yyyy", "yyyy-MM", "yyyy-MM-dd" };
+
+            formats.AddRange(dateFormats);
```
LTA-Thinking: Nit: Sense dateFormats isn't used anywhere else you could put these into formats when the list is created.

src/Microsoft.Health.Fhir.Core.UnitTests/Models/PartialDateTimeTests.cs
rotodd: Yep exactly - they are included in `GetParameterNullData()`.

src/Microsoft.Health.Fhir.Core.UnitTests/Models/PartialDateTimeTests.cs
rotodd: Good point. I took it out because this exception can be thrown from two different places. They both send the same message, though, so I think it makes sense to add it in. Thanks!

src/Microsoft.Health.Fhir.Core.UnitTests/Models/PartialDateTimeTests.cs
rotodd: I believe I was trying to make ReSharper happy haha. I'll make my changes consistent across the file.

brendankowitz: The guidance is to use the explicit type when it isn't obvious, in this case you could use var since you directly see the type :) pretty sure Resharper has a setting to support this.

src/Microsoft.Health.Fhir.Core.UnitTests/Models/PartialDateTimeTests.cs
rotodd: Fixed in 5818fc1.

src/Microsoft.Health.Fhir.Core/Models/PartialDateTime.cs
rotodd: Good idea, thanks. Fixed in c9c8e78.

src/Microsoft.Health.Fhir.Core.UnitTests/Models/PartialDateTimeTests.cs
rotodd: Changed to var in 6c0373a.

src/Microsoft.Health.Fhir.Core/Models/PartialDateTime.cs
rotodd: Awesome, I was in need of feedback from a regex expert. Let me know how b05e9ef looks.

src/Microsoft.Health.Fhir.Core.UnitTests/Models/PartialDateTimeTests.cs
```diff
@@ -411,62 +348,18 @@ public void GivenAValidPartialDateTime_WhenToStringIsCalled_ThenCorrectStringSho
         [InlineData("de-DE", "2018-11-29T18:30:27.911+01:00", "2018-11-29T18:30:27,9110000+01:00")]
         [InlineData("en-GB", "2018-11-29T18:30:27.911+01:00", "2018-11-29T18:30:27.9110000+01:00")]
         [InlineData("en-US", "2018-11-29T18:30:27.911+01:00", "2018-11-29T18:30:27.9110000+01:00")]
-        public void GivenACulture_WhenToStringisCalled_ThenCorrectStringShouldBeReturned(string culture, string input, string expected)
+        public void GivenACulture_WhenToStringIsCalled_ThenCorrectStringShouldBeReturned(string culture, string inputString, string expectedString)
         {
-            System.Threading.Thread.CurrentThread.CurrentCulture = new System.Globalization.CultureInfo(culture);
-            PartialDateTime dateTime = PartialDateTime.Parse(input);
+            System.Threading.Thread.CurrentThread.CurrentCulture = new CultureInfo(culture);
```
brendankowitz: You could use the `Mock.Property` helper here to ensure the thread is set back to the original value:
```
using (Mock.Property(() => System.Threading.Thread.CurrentThread.CurrentCulture, new CultureInfo(culture)))
{
 //...
}
```

brendankowitz: Ah, already covered, nvm :)

PR #934, opened 2020-03-11
Breakout sharable sql components.

brandonpollett:
## Description

## Related issues
Addresses [issue #].

## Testing


src/Microsoft.Health.SqlServer.Api/Controllers/SchemaController.cs
```diff
@@ -45,10 +48,10 @@ public ActionResult AvailableVersions()
 
             var availableSchemas = new List<object>();
             var currentVersion = _schemaInformation.Current ?? 0;
-            foreach (var version in Enum.GetValues(typeof(SchemaVersion)).Cast<SchemaVersion>().Where(sv => sv >= currentVersion))
+            foreach (var version in Enum.GetValues(typeof(TSchemaVersionEnum)).Cast<int>().Where(sv => sv >= currentVersion))
```
brendankowitz: A little bit of thinking out loud here, but could this be driven from ISchemaInformation now? i.e. `for(int i = schemaInformation.Current; i <= schemaInformation.Max...)`?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlConnectionWrapperFactory.cs
```diff
@@ -4,7 +4,7 @@
 // -------------------------------------------------------------------------------------------------
 
 using EnsureThat;
-using Microsoft.Health.Fhir.SqlServer.Configs;
+using Microsoft.Health.SqlServer.Configs;
```
jackliums: Shouldn't these move into the common code too?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
```diff
@@ -18,9 +18,12 @@
 using Microsoft.Health.Fhir.Core.Exceptions;
 using Microsoft.Health.Fhir.Core.Features.Definition;
 using Microsoft.Health.Fhir.Core.Models;
-using Microsoft.Health.Fhir.SqlServer.Configs;
 using Microsoft.Health.Fhir.SqlServer.Features.Schema;
 using Microsoft.Health.Fhir.SqlServer.Features.Schema.Model;
+using Microsoft.Health.SqlServer.Configs;
```
jackliums: We need to do similar type of initialization in the DICOM side too (but just initializing the schema), so it would be good to have a common interface which each project could derive and it automatically just works

src/Microsoft.Health.Fhir.SqlServer/Registration/FhirServerBuilderSqlServerRegistrationExtensions.cs
```diff
@@ -6,17 +6,16 @@
...
-using Microsoft.Health.Fhir.SqlServer.Features.Health;
 using Microsoft.Health.Fhir.SqlServer.Features.Schema;
-using Microsoft.Health.Fhir.SqlServer.Features.Schema.Model;
 using Microsoft.Health.Fhir.SqlServer.Features.Search;
 using Microsoft.Health.Fhir.SqlServer.Features.Search.Expressions.Visitors;
 using Microsoft.Health.Fhir.SqlServer.Features.Storage;
+using Microsoft.Health.SqlServer.Api.Registration;
+using Microsoft.Health.SqlServer.Configs;
+using Microsoft.Health.SqlServer.Features.Schema.Model;
+using Microsoft.Health.SqlServer.Registration;
```
jackliums: nit: Maybe update the name too. it's no longer experimental.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/SchemaInformation.cs
```diff
@@ -3,20 +3,23 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
+using System;
+using Microsoft.Health.SqlServer.Features.Schema;
+
 namespace Microsoft.Health.Fhir.SqlServer.Features.Schema
 {
-    public class SchemaInformation
+    public class SchemaInformation : ISchemaInformation
```
jackliums: should this move to common too?

src/Microsoft.Health.Fhir.SqlServer/Microsoft.Health.Fhir.SqlServer.csproj
```diff
@@ -38,21 +31,8 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.Health.Fhir.Core\Microsoft.Health.Fhir.Core.csproj" />
-  </ItemGroup>
```
jackliums: is this removal intentional?

src/Microsoft.Health.Fhir.SqlServer.UnitTests/IndentedStringBuilderTests.cs
```diff
@@ -5,6 +5,7 @@
 
 using System;
 using System.Text;
+using Microsoft.Health.SqlServer;
```
jackliums: Should this test be moved now that IndentedStringBuilder is part of Microsoft.Health.SqlServer

src/Microsoft.Health.SqlServer.Api/Controllers/SchemaController.cs
```diff
@@ -44,11 +49,11 @@ public ActionResult AvailableVersions()
             _logger.LogInformation("Attempting to get available schemas");
 
             var availableSchemas = new List<object>();
-            var currentVersion = _schemaInformation.Current ?? 0;
-            foreach (var version in Enum.GetValues(typeof(SchemaVersion)).Cast<SchemaVersion>().Where(sv => sv >= currentVersion))
+            var currentVersion = _schemaInformation.Current ?? 1;
```
jackliums: is this intentional change?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlConnectionWrapperFactory.cs
brandonpollett: So far this is only a concept that's needed within the fhir-server. We can move it, but that would require a bit more refactoring as the common transaction interfaces live in the fhir-server core.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/SchemaInformation.cs
brandonpollett: Sounds good to me.

src/Microsoft.Health.Fhir.SqlServer/Microsoft.Health.Fhir.SqlServer.csproj
brandonpollett: Nope, good catch!

src/Microsoft.Health.SqlServer.Api/Controllers/SchemaController.cs
brandonpollett: Yes, because we no longer have access to the enum in this class, we need to initialize with 1 since we're just using the int iterator to build the data.

src/Microsoft.Health.Fhir.SqlServer.UnitTests/IndentedStringBuilderTests.cs
brandonpollett: Yes, along with `EnumerableExtensionsTests.cs`

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlConnectionWrapperFactory.cs
jackliums: The DICOM code actually has these registered but that might be because I just copied the code without looking at it closely. Let's not move the code and we can refactor the code in DICOM side when we incorporate the nuget package.

src/Microsoft.Health.SqlServer.Api/Controllers/SchemaController.cs
```diff
@@ -9,30 +9,35 @@
...
-            EnsureArg.IsNotNull(urlResolver, nameof(urlResolver));
+            EnsureArg.IsNotNull(scriptProvider, nameof(scriptProvider));
+            EnsureArg.IsNotNull(urlHelperFactory, nameof(urlHelperFactory));
+            EnsureArg.IsNotNull(actionContextAccessor, nameof(actionContextAccessor));
             EnsureArg.IsNotNull(logger, nameof(logger));
 
             _schemaInformation = schemaInformation;
-            _urlResolver = urlResolver;
+            _scriptProvider = scriptProvider;
+            _urlHelper = urlHelperFactory.GetUrlHelper(actionContextAccessor.ActionContext);
```
jackliums: Looks like my previous comment got lost. Would using the 'ControllerContext' property work here? It derives from `ActionContext`.

PR #938, opened 2020-03-16
Implements current version api

rbans96:
## Description
Implements GET //fhir-server/_schema/current

## Related issues
User Story 68809.

## Testing
Adds Unit tests, e2e tests and manual tests

Given -
InstanceSchema table has below entries
image
SchemaVersion table has below entries
image

Then output is -
image

And when InstanceSchema table has no entries and SchemaVersion table at above state, 
Then output is-
image


src/Microsoft.Health.Fhir.SqlServer.Api.UnitTests/Features/CurrentVersionSchemaHandlerTests.cs
```diff
@@ -0,0 +1,80 @@
...
+            ServiceProvider provider = collection.BuildServiceProvider();
+            _mediator = new Mediator(type => provider.GetService(type));
+            _cancellationToken = new CancellationTokenSource().Token;
+        }
+
+        [Fact]
+        public async Task GivenACurrentMediator_WhenCurrentRequest_ThenReturnsCurrentVersionInformation()
+        {
+            IList<CurrentVersionInformation> mockCurrentVersions = CurrentVersions();
+            SetupDataStore(_ => new GetCurrentVersionResponse(mockCurrentVersions));
```
namadabu: nit: `SetupDataStore` seems to be doing only one thing. Do we need to wrap that in a separate function? Might be cleaner to call it directly here?

src/Microsoft.Health.Fhir.SqlServer.Api/Features/Filters/HttpExceptionFilterAttribute.cs
```diff
@@ -39,6 +40,11 @@ public override void OnActionExecuted(ActionExecutedContext context)
                         context.Result = new JsonResult(resultJson) { StatusCode = (int)HttpStatusCode.NotFound };
                         context.ExceptionHandled = true;
                         break;
+
+                    case RecordNotFoundException _:
```
namadabu: nit: The cases for `RecordNotFoundException` and `FileNotFoundException` are doing the same actions on the result. We can simplify the code

src/Microsoft.Health.Fhir.SqlServer.Api/Features/LatestSchemaVersionHandler.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using EnsureThat;
+using MediatR;
+using Microsoft.Health.Fhir.SqlServer.Features.Schema;
+using Microsoft.Health.Fhir.SqlServer.Features.Schema.Messages.Get;
+
+namespace Microsoft.Health.Fhir.SqlServer.Api.Features
+{
+    public class LatestSchemaVersionHandler : IRequestHandler<GetCompatibilityVersionRequest, GetCompatibilityVersionResponse>
```
namadabu: Would be good to have the handler name similar to the name of the request that it handles to improve readability. Maybe something like `GetCompatibleSchemaVersionHandler` or something like that?

src/Microsoft.Health.Fhir.SqlServer.Api/Features/Filters/HttpExceptionFilterAttribute.cs
rbans96: Good point! Thanks

PR #946, opened 2020-03-23
Move shared SQL to nuget packages.

brandonpollett:
## Description
Update the project to use the shared sql nuget packages.

## Related issues
Addresses [AB#72572](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/72572).

## Testing
All existing tests continue to pass or moved except for one extension test that was marked as only debug.


src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
```diff
@@ -32,14 +31,11 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Storage
     /// many many times in the database. For more compact storage, we use IDs instead of the strings when referencing these.
     /// Also, because the number of distinct values is small, we can maintain all values in memory and avoid joins when querying.
     /// </summary>
-    public sealed class SqlServerFhirModel : IDisposable
+    public sealed class SqlServerFhirModel : SqlServerModelInitializer
```
jackliums: nit: maybe SqlServerSchemaInitializer since minimally, it needs to initialize the schema and model could be optional?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
```diff
@@ -139,21 +125,19 @@ public bool TryGetQuantityCodeId(string code, out int quantityCodeId)
             return _quantityCodeToId.TryGetValue(code, out quantityCodeId);
         }
 
-        public ValueTask EnsureInitialized() => _initializationOperation.EnsureInitialized();
-
-        private async Task Initialize()
+        protected override async Task Initialize()
         {
-            if (!_schemaInformation.Current.HasValue)
+            if (!SchemaInformation.Current.HasValue)
```
jackliums: Perhaps this should move to the base class since minimally, it should initialize the schema.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Sql/SchemaTests.cs
```diff
@@ -50,7 +50,7 @@ public async Task GivenAServerThatHasSchemas_WhenRequestingAvailable_JsonShouldB
             Assert.NotEmpty(jArrayResponse);
 
             JToken firstResult = jArrayResponse.First;
-            string scriptUrl = $"{_client.BaseAddress}_schema/versions/{firstResult["id"]}/script";
+            string scriptUrl = $"/_schema/versions/{firstResult["id"]}/script";
```
jackliums: should these test go to common code? and only have the FHIR specific ones here (if any?)

brandonpollett: Yes, good idea.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
brandonpollett: There's already a schema initializer. This class is for initializing the common data (i.e. resource types, systems, etc.). I'm open to calling it something else besides model, but I'm not sure I have a better name.

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlServerFhirStorageTestHelper.cs
```diff
@@ -185,8 +186,9 @@ SELECT t.name
         private SchemaInitializer CreateSchemaInitializer(string testConnectionString)
         {
             var config = new SqlServerDataStoreConfiguration { ConnectionString = testConnectionString, Initialize = true };
-            var schemaUpgradeRunner = new SchemaUpgradeRunner(config, NullLogger<SchemaUpgradeRunner>.Instance);
-            var schemaInformation = new SchemaInformation();
+            var schemaInformation = new SchemaInformation((int)SchemaVersion.V1, (int)SchemaVersion.V2);
```
brendankowitz: Is there a way we could determine max version automatically? Perhaps an extension method that does something like `Enum.GetValues(typeof(SchemaVersion)).Cast< SchemaVersion >().Max()`?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Sql/SchemaTests.cs
brandonpollett: Moved to common code in separate PR

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlServerFhirStorageTestHelper.cs
brandonpollett: We could technically do it that way. I was thinking it may make more sense to add a constants value like something below:

```dotnet
namespace Microsoft.Health.Fhir.SqlServer.Features.Schema
{
    public static class SchemaVersionConstants
    {
        public const int Max = (int)SchemaVersion.V2;
        public const int Min = (int)SchemaVersion.V1;
    }
}
```

brendankowitz: Yep also makes sense

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
brandonpollett: Moved to common code in separate PR

PR #949, opened 2020-03-24
Updates Instance Schema information

rbans96:
## Description
Insert and update the records in InstanceSchema table
Minor refactoring of SchemaInformation class

## Related issues
Addresses 72475

## Testing
Unit test and integration tests

Screenshot of polling and updating the InstanceSchema records based on compatibility versions fetched.
image
image


src/Microsoft.Health.Fhir.SqlServer/Features/Schema/SchemaInitializer.cs
```diff
@@ -59,18 +71,18 @@ internal void Initialize(bool forceIncrementalSchemaUpgrade = false)
...
             }
 
             // If the current schema version needs to be upgraded
-            if (_schemaInformation.Current < _schemaInformation.MaximumSupportedVersion)
+            if (_sqlServerDataStoreConfiguration.SchemaUpdatesEnabled && _schemaInformation.Current < _schemaInformation.MaximumSupportedVersion)
             {
                 // Apply each .diff.sql file one by one.
-                int current = (int?)_schemaInformation.Current ?? 0;
-                for (int i = current + 1; i <= (int)_schemaInformation.MaximumSupportedVersion; i++)
+                int current = _schemaInformation.Current ?? 0;
```
namadabu: This part is a bit confusing. Can `_schemaInformation.Current` be null here? If so, then the above comparison in the `if` condition would not have worked right? If it cannot be null, then we can directly use it?

src/Microsoft.Health.Fhir.SqlServer/Features/Exceptions/OperationFailedException.cs
```diff
@@ -0,0 +1,19 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using EnsureThat;
+
+namespace Microsoft.Health.Fhir.SqlServer.Features.Exceptions
+{
+    public class OperationFailedException : Exception
```
namadabu: We already have an `OperationFailedException` class in `Fhir.Core`. Can you rename this to something more specific to Sql to differentiate?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/SchemaInitializer.cs
```diff
@@ -59,18 +71,18 @@ internal void Initialize(bool forceIncrementalSchemaUpgrade = false)
...
+                for (int i = current + 1; i <= _schemaInformation.MaximumSupportedVersion; i++)
```
namadabu: Not your change, but just curious why we are calling `GetCurrentSchemaVersion()` after the end of the for loop. We don't seem to be doing anything with the result

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerSchemaDataStore.cs
```diff
@@ -108,15 +186,15 @@ public async Task<GetCurrentVersionResponse> GetCurrentVersionAsync(Cancellation
             return new GetCurrentVersionResponse(currentVersions);
         }
 
-        private int ConvertToInt(object o)
+        public static int ConvertToInt(object o)
```
namadabu: Any reason this needs to be `public static`?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerSchemaDataStore.cs
```diff
@@ -59,6 +59,84 @@ public async Task<GetCompatibilityVersionResponse> GetLatestCompatibleVersionAsy
...
+                    if (output.NullIfEmpty() == null)
+                    {
+                        throw new OperationFailedException(Resources.OperationFailed);
+                    }
+
+                    return output;
+                }
+                catch (SqlException e)
+                {
+                    _logger.LogError(e, "Error from SQL database on Upsert");
```
namadabu: Error from SQL database on Insert

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerSchemaDataStore.cs
```diff
@@ -59,6 +59,84 @@ public async Task<GetCompatibilityVersionResponse> GetLatestCompatibleVersionAsy
...
+                try
+                {
+                    var output = (string)await sqlCommand.ExecuteScalarAsync();
+
+                    if (output.NullIfEmpty() == null)
+                    {
+                        throw new OperationFailedException(Resources.OperationFailed);
+                    }
+
+                    return output;
```
namadabu: Do we need to return the `output`. Doesn't seem to be used by the caller.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/SchemaJobWorker.cs
```diff
@@ -0,0 +1,72 @@
...
+            using (IScoped<ISchemaDataStore> store = _schemaDataStoreFactory())
+            {
+                await store.Value.InsertInstanceSchemaInformation(instanceName, schemaInformation, cancellationToken);
+            }
+
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                try
+                {
+                    _logger.LogInformation($"Polling started at {TimeZoneInfo.ConvertTimeToUtc(DateTime.Now)}");
```
namadabu: nit: why not use `DateTimeOffset.UtcNow` ?

brendankowitz: For consistency use `Clock.UtcNow`

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/SchemaInitializer.cs
rbans96: Good point, Addressed!

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerSchemaDataStore.cs
rbans96: Good point, Addressed!

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/SchemaJobWorker.cs
```diff
@@ -0,0 +1,73 @@
...
+            using (IScoped<ISchemaDataStore> store = _schemaDataStoreFactory())
+            {
+                await store.Value.InsertInstanceSchemaInformation(instanceName, schemaInformation, cancellationToken);
+            }
+
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                try
+                {
+                    _logger.LogInformation($"Polling started at {Clock.UtcNow}");
```
rotodd, line 49: It looks like this message will be logged every minute. I am wondering if it might make sense to only log this before we enter the poll loop, opposed to every iteration?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/SchemaJobWorker.cs
```diff
@@ -0,0 +1,73 @@
...
+
+                    using (IScoped<ISchemaDataStore> store = _schemaDataStoreFactory())
+                    {
+                        GetCompatibilityVersionResponse getCompatibilityVersion = await store.Value.GetLatestCompatibleVersionAsync(cancellationToken);
+
+                        await store.Value.UpsertInstanceSchemaInformation(instanceName, getCompatibilityVersion.Versions, schemaInformation.Current.GetValueOrDefault(), cancellationToken);
```
rotodd, line 55: Just for my understanding: why is `InstanceSchema.MaxVersion` and `InstanceSchema.MinVersion` getting set to the compatible versions?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/SchemaJobWorker.cs
rbans96: That makes sense, fixed it.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/SchemaJobWorker.cs
rbans96: Good question. I was working on some refactoring and logic updates yesterday and this is also updated. Now it is set from SchemaInformation property to ensure that if server is restarted and any change in the property should also get updated in the InstanceSchema table.

rotodd: Nice, this makes sense.

Would it be possible to remove the C# call to `GetCurrentSchemaVersion()` and instead modify the SQL logic in the `UpsertInstanceSchema` stored procedure? You might be able to set `InstanceSchema.CurrentVersion` to `MAX(SchemaVersion.Version) WHERE Status = 'complete'` directly.

This would reduce SQL calls and would allow you to get-and-update concurrently.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerSchemaDataStore.cs
```diff
@@ -59,6 +59,80 @@ public async Task<GetCompatibilityVersionResponse> GetLatestCompatibleVersionAsy
...
 
+        public async Task InsertInstanceSchemaInformation(string name, SchemaInformation schemaInformation, CancellationToken cancellationToken)
+        {
+            using (SqlConnectionWrapper sqlConnectionWrapper = _sqlConnectionWrapperFactory.ObtainSqlConnectionWrapper())
+            using (SqlCommand sqlCommand = sqlConnectionWrapper.CreateSqlCommand())
+            {
+                VLatest.CreateInstanceSchema.PopulateCommand(
+                     sqlCommand,
+                     name,
+                     schemaInformation.Current.GetValueOrDefault(),
```
rotodd, line 70: Just curious, when would `schemaInformation.Current` be null?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/SchemaInitializer.cs
rotodd: I remember being confused by this naming as well and considered changing it to `GetAndUpdateCurrentSchemaVersion()`.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/SchemaJobWorker.cs
```diff
@@ -0,0 +1,72 @@
...
+                {
+                    using (IScoped<ISchemaDataStore> store = _schemaDataStoreFactory())
+                    {
+                        // Ensure schemaInformation has the latest current version
+                        schemaInformation.Current = await store.Value.UpsertInstanceSchemaInformation(instanceName, schemaInformation, cancellationToken);
+
+                        await store.Value.DeleteExpiredRecords();
+                    }
+
+                    await Task.Delay(_sqlServerDataStoreConfiguration.SchemaUpdatesJobPollingFrequency, cancellationToken);
```
brendankowitz, line 58: If an error happens before this, this loop will not wait and will continuously process the error. Thoughts on having the delay outside the using?

src/Microsoft.Health.Fhir.SqlServer.Api/Registration/FhirServerBuilderSqlServerRegistrationExtensions.cs
```diff
@@ -124,6 +129,21 @@ public static IFhirServerBuilder AddExperimentalSqlServer(this IFhirServerBuilde
...
+        /// Adds the Schema worker background service.
+        /// </summary>
+        /// <param name="fhirServerBuilder">The FHIR server builder.</param>
+        /// <returns>The builder.</returns>
+        public static IFhirServerBuilder AddSchemaWorker(
+            this IFhirServerBuilder fhirServerBuilder)
+        {
+            EnsureArg.IsNotNull(fhirServerBuilder, nameof(fhirServerBuilder));
+
+            fhirServerBuilder.Services.AddHostedService<SchemaJobWorkerBackgroundService>();
```
brendankowitz, line 142: Any downside in adding this directly into the `AddExperimentalSqlServer` method?

rbans96: Makes sense, Added directly into the AddExperimentalSqlServer method.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/SchemaJobWorker.cs
rbans96: Ah Good point! I fixed it by adding the delay in finally block to poll every minute even if its error.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/3.diff.sql
```diff
@@ -166,17 +163,9 @@ AS
...
-
-    Select @maxSchemaVersion = MIN(MaxVersion), @minSchemaVersion = MAX(MinVersion)
+    Select MAX(MinVersion), MIN(MaxVersion)
     FROM dbo.InstanceSchema
     WHERE Timeout > SYSUTCDATETIME()
-
-    SELECT @minSchemaVersion, MAX(Version)
-    FROM dbo.SchemaVersion
-    WHERE Status = 'complete' AND Version <= @maxSchemaVersion
-
```
rotodd, line 179: Why do we no longer need to check this?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/3.sql
```diff
@@ -2105,67 +2097,66 @@ GO
...
-    @currentVersion int,
     @maxVersion int,
     @minVersion int
     
 AS
     SET NOCOUNT ON
 
     DECLARE @timeout datetime2(0) = DATEADD(minute, 2, SYSUTCDATETIME())
-    IF EXISTS(SELECT *
+    DECLARE @currentVersion int = (SELECT COALESCE(MAX(Version), 0)
```
rotodd, line 2122: If upsert is called when `MAX(SchemaVersion.Version)` is null, should we be throwing an error instead of setting the value to 0?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/SchemaJobWorker.cs
```diff
@@ -0,0 +1,74 @@
...
+            _logger.LogInformation($"Polling started at {Clock.UtcNow}");
+
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                try
+                {
+                    using (IScoped<ISchemaDataStore> store = _schemaDataStoreFactory())
+                    {
+                        // Ensure schemaInformation has the latest current version
+                        schemaInformation.Current = await store.Value.UpsertInstanceSchemaInformation(instanceName, schemaInformation, cancellationToken);
```
rotodd, line 53: +1

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/ISchemaDataStore.cs
```diff
@@ -24,5 +24,29 @@ public interface ISchemaDataStore
...
+        /// </summary>
+        /// <returns>A task</returns>
+        Task DeleteExpiredRecords();
+
+        /// <summary>
+        /// Upsert current version information.
+        /// </summary>
+        /// /// <param name="name">The instance name.</param>
+        /// /// <param name="schemaInformation">The Schema information</param>
+        /// /// <param name="cancellationToken">The cancellation token.</param>
```
rotodd, line 39: Nit: looks like all the `param` comments are accidentally commented twice in this file.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/3.diff.sql
rbans96: Good question! Tool would need to call this compatibility api to verify if the given version is in compatibility range, if yes, it would proceed to run the sql script for this given version. While status = Current specifies the current version or previous versions for which migration script had already run. I realized while working on the tool side so here is the refactoring.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/3.sql
rbans96: I think this is just to take into account the first time when no schema version execution is complete. And we should not be throwing exception rather just upgrade to the available versions by the default upgrade or tool.

rotodd: Right, makes sense. Thanks 👍

PR #955, opened 2020-03-26
Handle null code value in search indexer

feordin:
## Description
Addresses a 500 error that was being thrown by the SearchIndexer for for null coded values

## Related issues
Addresses [AB#72608](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/72608).

## Testing
Added a unit test which repo'd the error.  Then saw it pass when the error was addressed.

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Converters/CodeOfTToTokenSearchValueTypeConverter.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Converters/CodeOfTToTokenSearchValueTypeConverter.cs
@@ -31,9 +31,16 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Converters
 
             EnsureArg.IsTrue(type.IsGenericType && type.GetGenericTypeDefinition() == FhirElementType, nameof(value));
 
-            ISystemAndCode systemAndCode = (ISystemAndCode)value;
+            var systemAndCode = value as ISystemAndCode;
 
-            yield return new TokenSearchValue(systemAndCode.System, systemAndCode.Code, null);
+            if (systemAndCode == null || systemAndCode.Code == null)
+            {
+                yield return null;
+            }
+            else
+            {
+                yield return new TokenSearchValue(systemAndCode.System, systemAndCode.Code, null);
+            }
         }
     }
 }
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Converters/CodeOfTToTokenSearchValueTypeConverter.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Converters/CodeOfTToTokenSearchValueTypeConverter.cs
@@ -33,7 +33,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Converters
 
             var systemAndCode = value as ISystemAndCode;
 
-            if (systemAndCode == null || systemAndCode.Code == null)
+            if (systemAndCode == null || string.IsNullOrEmpty(systemAndCode.Code)
             {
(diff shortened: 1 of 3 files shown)


src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Converters/CodeOfTToTokenSearchValueTypeConverter.cs
```diff
@@ -31,9 +31,16 @@ public IEnumerable<ISearchValue> ConvertTo(object value)
 
             EnsureArg.IsTrue(type.IsGenericType && type.GetGenericTypeDefinition() == FhirElementType, nameof(value));
 
-            ISystemAndCode systemAndCode = (ISystemAndCode)value;
+            var systemAndCode = value as ISystemAndCode;
 
-            yield return new TokenSearchValue(systemAndCode.System, systemAndCode.Code, null);
+            if (systemAndCode == null || systemAndCode.Code == null)
```
brendankowitz: Should `systemAndCode.Code` be exactly null or can IsNullOrWhitespace also count?

brendankowitz: Also is there a case where we System would be defined and not code?
Is the example from https://www.hl7.org/fhir/datatypes-examples.html#codeableconcept valid here:
```xml
 <unit>
    <coding>
      <system value="urn:oid:2.16.840.1.113883.19.5.2" />
      <code value="tab" />
      <display value="Tablet" />
    </coding>
    <coding>
      <system value="http://unitsofmeasure.org" />
    </coding>
  </unit>
```

feordin: I took a quick look at the firely code, where the exception is originating (https://github.com/FirelyTeam/fhir-net-api/blob/add8073b8d096f66cde13f54c940048ba4363e0a/src/Hl7.Fhir.Core/Model/Code.cs) and it seems we will only get the exception if the code is null.  If the System is null, we won't see an exception, we could create a search param with an empty system.  Currently our TokenSearchParam class allows that.  But perhaps we should not?

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Converters/CodeOfTToTokenSearchValueTypeConverter.cs
```diff
@@ -31,9 +31,16 @@ public IEnumerable<ISearchValue> ConvertTo(object value)
 
             EnsureArg.IsTrue(type.IsGenericType && type.GetGenericTypeDefinition() == FhirElementType, nameof(value));
 
-            ISystemAndCode systemAndCode = (ISystemAndCode)value;
+            var systemAndCode = value as ISystemAndCode;
 
-            yield return new TokenSearchValue(systemAndCode.System, systemAndCode.Code, null);
+            if (systemAndCode == null || string.IsNullOrEmpty(systemAndCode.Code)
```
brendankowitz:
```suggestion
            if (systemAndCode == null || string.IsNullOrEmpty(systemAndCode.Code))
```

rbans96:
While doing acceptance, I figured out that it is only happening for xml requests with code<T> being null/empty. Though it fixes the 500 error. But now when code is empty, it returns 400 - Bad Request. I believe it is due to parser.
While json with/without fix returns Success response.
We might need to file a new bug.

Screenshot for XML request(with fix)
image

Screenshot for JSON request
image

PR #956, opened 2020-03-26
Handle key inaccessible error in BYOK scenario

shmartel:
## Description
When a customer-managed key is not accessible, Cosmos DB will return a 403 to the FHIR server. In turn, the FHIR server should return a 403, as well as an appropriate OperationOutcome.  

## Related issues
Addresses [AB#73109](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/73109)

## Testing
- Ran all existing unit tests
- Added new unit tests
- Using a proof-of-concept console app to try to read a Cosmos DB document where the key is inaccessible, I have confirmed that the new logic in ProcessException to detect the error in question works.


Microsoft.Health.Fhir.sln
```diff
@@ -137,14 +137,53 @@ EndProject
 Global
 	GlobalSection(SharedMSBuildProjectFiles) = preSolution
 		test\Microsoft.Health.Fhir.Shared.Tests.E2E.Common\Microsoft.Health.Fhir.Shared.Tests.E2E.Common.projitems*{0478b687-7105-40f6-a2dc-81057890e944}*SharedItemsImports = 13
+		src\Microsoft.Health.Fhir.Shared.Core\Microsoft.Health.Fhir.Shared.Core.projitems*{070759a9-51d9-4967-8651-39cca8288c93}*SharedItemsImports = 5
```
shmartel: I'm not sure why all these new lines were added.  Will remove.

src/Microsoft.Health.CosmosDb/Features/Storage/DocumentClientExceptionExtensions.cs
```diff
@@ -21,10 +21,24 @@ public static class DocumentClientExceptionExtensions
...
+        {
+            return (HttpStatusCode?)exception.GetSubStatusValue();
+        }
+
+        /// <summary>
+        /// When a javascript stored procedure throws an exception in code, such as
+        /// throw new Error(404, "Document not found");
+        /// the status code of the exception will be 400 bad request, but there will be a substatus header
+        /// with the error code specified in the Error object (404 in this example).
+        /// This methods returns the value of substatus header if it is present.
```
shmartel: Typo.  Method.  Fix above as well

shmartel:
This change is now unblocked and can be revisited.  I will attempt to re-open this PR.

PR #961, opened 2020-03-27
Have PR release use docker images

LTA-Thinking:
## Description
Change the PR release to use a docker image instead of self compiled source.

## Testing
Manual testing


build/jobs/provision-deploy.yml
```diff
@@ -0,0 +1,87 @@
...
+     
+        $additionalProperties = $flattenedCors
+     
+        $additionalProperties["SqlServer__DeleteAllDataOnStartup"] = "true"
+        $additionalProperties["ASPNETCORE_FORWARDEDHEADERS_ENABLED"] = "true"
+
+        $templateParameters = @{
+            fhirVersion = "${{ parameters.version }}"
+            # appServicePlanResourceGroup = "${{ parameters.appServicePlanResourceGroup }}"
+            # appServicePlanName = "${{ parameters.appServicePlanName }}"
```
brendankowitz, line 56: Not needed now?

build/jobs/run-tests.yml
```diff
@@ -0,0 +1,133 @@
...
+      artifactName: 'IntegrationTests'
+
+  - task: UseDotNet@2
+    inputs:
+      useGlobalJson: true
+
+  - task: AzureKeyVault@1
+    displayName: 'Azure Key Vault: ${{ parameters.keyVaultName }}'
+    inputs:
+      azureSubscription: 'Microsoft Health Open Source Subscription'
```
brendankowitz, line 33: $(ConnectedServiceName)

LTA-Thinking: I'll fix this in the yaml PR. It'll show up here when I merge from master.

build/jobs/provision-deploy.yml
LTA-Thinking: Not needed currently. The shared App Service Plans don't work with the linux containers, but I think we will still want shared App Service Plans for cost reasons. I'll make new shared ones.

LTA-Thinking: Annoyingly I also had to make a new resource group as app service plans for windows and linux can't be in the same resource group and region. The shared plan is in msh-fhir-pr-linux.

PR #964, opened 2020-03-30
Adds current version in tool

rbans96:
## Description
Adds the support for current version information in Schema Manager tool
Renders the success and non success response in table view.

## Related issues
Addresses  #72784

## Testing
Manual testing

(Screenshot shows the success current command and invalid fhirServerUrl execution)
image


tools/FhirSchemaManager/InvokeAPI.cs
```diff
@@ -0,0 +1,45 @@
...
+using FhirSchemaManager.Model;
+using Newtonsoft.Json;
+
+namespace FhirSchemaManager
+{
+    internal class InvokeAPI : IInvokeAPI
+    {
+        public async Task<List<CurrentVersion>> GetCurrentVersionInformation(Uri serverUri)
+        {
+            var httpClient = new HttpClient
```
brandonpollett, line 21: Would it make sense to do this in the constructor of the `InvokeAPI` and make it available through a private field? When you get into a process that calls multiple methods, it may make it easier.

tools/FhirSchemaManager/InvokeAPI.cs
```diff
@@ -0,0 +1,45 @@
...
+            {
+                BaseAddress = serverUri,
+            };
+
+            var response = await httpClient.GetAsync(new Uri("/_schema/versions/current", UriKind.Relative));
```
brandonpollett, line 26: nit: put the URL in a const or static?

tools/FhirSchemaManager/Resources.resx
```diff
@@ -132,12 +132,24 @@
...
     <value>The root URL of the FHIR server to connect to.</value>
   </data>
   <data name="FhirServerRequiredValidation" xml:space="preserve">
     <value>You must include a FHIR server.</value>
   </data>
+  <data name="HttpRequestExceptionMessage" xml:space="preserve">
+    <value>No connection could be made to the target machine.</value>
+  </data>
+  <data name="RequestFailedMessage" xml:space="preserve">
+    <value>Request has failed due to fhirServerUrl is invalid or inactive.</value>
```
brandonpollett, line 151: nit: probably need a placeholder for `fhirServerUrl` since it's a parameter name. 

Do we know when the server is invalid vs when it's unreachable?

tools/FhirSchemaManager/Model/ErrorDescription.cs
```diff
@@ -0,0 +1,26 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System.Net;
+using EnsureThat;
+
+namespace FhirSchemaManager.Model
+{
+    public class ErrorDescription
+    {
+        public ErrorDescription(int statusCode, string message)
```
brandonpollett, line 13: You could make all of your exceptions have a base type of something like `SchemaManagerException` and pass that into `RenderErrror` instead of having this intermediate class.

Also, you may just be able to get away with passing all exceptions to `RenderError` and special case the schema manager specific ones.

tools/FhirSchemaManager/InvokeAPI.cs
rbans96: Makes sense, thanks!

tools/FhirSchemaManager/Resources.resx
rbans96: Right, when the url is invalid, exception is thrown by validators itself. So now updated errorMessage just for inactive.

tools/FhirSchemaManager/IInvokeAPI.cs
```diff
@@ -0,0 +1,17 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using FhirSchemaManager.Model;
+
+namespace FhirSchemaManager
+{
+    public interface IInvokeAPI
```
brendankowitz, line 13: What about something like IFhirSchemaClient?

tools/FhirSchemaManager/InvokeAPI.cs
```diff
@@ -0,0 +1,43 @@
...
+        private static HttpClient _httpClient;
+
+        public InvokeAPI()
+        {
+            _httpClient = new HttpClient();
+        }
+
+        public async Task<List<CurrentVersion>> GetCurrentVersionInformation(Uri serverUri)
+        {
+            _httpClient.BaseAddress = serverUri;
```
brendankowitz, line 29: I think this should be done in the constructor

tools/FhirSchemaManager/Exceptions/SchemaManagerException.cs
```diff
@@ -0,0 +1,19 @@
...
+using EnsureThat;
+
+namespace FhirSchemaManager.Exceptions
+{
+    public class SchemaManagerException : Exception
+    {
+        public SchemaManagerException(string message)
+            : base(message)
+        {
+            EnsureArg.IsNotNullOrWhiteSpace(message, nameof(message));
```
brendankowitz, line 16: Perhaps use `Debug.Assert(!string.IsNullOrEmpty(message), "Exception message should not be empty");` we want to avoid the possibility of raising an exception in the constructor of an exception

tools/FhirSchemaManager/Model/CurrentVersion.cs
```diff
@@ -0,0 +1,28 @@
...
+using System.Collections.Generic;
+using EnsureThat;
+
+namespace FhirSchemaManager.Model
+{
+    public class CurrentVersion
+    {
+        public CurrentVersion(int id, string status, List<string> servers)
+        {
+            EnsureArg.IsNotNull(status, nameof(status));
```
brendankowitz, line 15: EnsureArg on status and servers?

tools/FhirSchemaManager/IInvokeAPI.cs
brendankowitz: ...actually if this tool is shared, it should be ISchemaClient

PR #965, opened 2020-03-31
Adds FilebasedSearchParameterRegistry

brendankowitz:
## Description
Adds Read-only FilebasedSearchParameterRegistry

## Related issues
Addresses [AB#72423](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/72423).

## Testing
Added unit tests


build/jobs/provision-deploy.yml
```diff
@@ -85,6 +85,8 @@ jobs:
     inputs:
```
LTA-Thinking, line 85: This file shouldn't be part of this review.

src/Microsoft.Health.Fhir.Shared.Core/Features/Definition/SearchParameterDefinitionManager.cs
```diff
@@ -6,10 +6,14 @@
...
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
 using EnsureThat;
 using Hl7.Fhir.Serialization;
+using MediatR;
 using Microsoft.Health.Extensions.DependencyInjection;
 using Microsoft.Health.Fhir.Core.Features.Search;
+using Microsoft.Health.Fhir.Core.Messages.Search;
```
LTA-Thinking, line 16: Why does this need more imports when you are removing lines?

src/Microsoft.Health.Fhir.Shared.Core/Features/Definition/SearchParameterDefinitionBuilder.cs
```diff
@@ -29,31 +29,26 @@ internal class SearchParameterDefinitionBuilder
...
         {
             EnsureArg.IsNotNull(fhirJsonParser, nameof(fhirJsonParser));
             EnsureArg.IsNotNull(modelInfoProvider, nameof(modelInfoProvider));
             EnsureArg.IsNotNull(assembly, nameof(assembly));
-            EnsureArg.IsNotNullOrWhiteSpace(embeddedResourceName, nameof(embeddedResourceName));
 
             _fhirJsonParser = fhirJsonParser;
             _modelInfoProvider = modelInfoProvider;
             _assembly = assembly;
             _embeddedResourceName = embeddedResourceName;
```
rbans96, line 51: Do we need to validate embeddedResourceName?
EnsureArg.IsNotNull(embeddedResourceName, nameof(embeddedResourceName )

src/Microsoft.Health.Fhir.Shared.Core/Features/Definition/SearchParameterDefinitionBuilder.cs
```diff
@@ -292,21 +281,21 @@ void EnsureNoIssues()
...
-                info.IsSupported = false;
-                info.IsSearchable = false;
+                return spi;
             }
 
-            if (_unsupportedParams.PartialSupport.Contains(uri))
+            // Return SearchParameterInfo that has already been created for this Resource
+            if (_resourceTypeDictionary.TryGetValue(searchParameter.ResourceType.ToString(), out var spDictionary) &&
+                spDictionary.TryGetValue(searchParameter.Name, out var resourceSpi) &&
+                resourceSpi.Url == new Uri(searchParameter.Url))
```
rbans96, line 293: Curious, if we should compare Uri instances using equals() method?
https://docs.microsoft.com/en-us/dotnet/api/system.uri.equals?view=netframework-4.8

src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterStatusManager.cs
```diff
@@ -0,0 +1,80 @@
...
+                        Uri = p.Url,
+                        LastUpdated = Clock.UtcNow,
+                        Status = SearchParameterStatus.Supported,
+                    });
+
+                    p.IsSearchable = false;
+                    p.IsSupported = true;
+                }
+
+                updated.Add(p);
```
LTA-Thinking, line 69: Why make a new updated list? All the members of AllSearchParameters will be added to it, and they are being edited in place. Are updated and AllSearchParameters not identical?

src/Microsoft.Health.Fhir.Core/Features/Search/Registry/ResourceSearchParameterStatus.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Microsoft.Health.Fhir.Core.Features.Search.Registry
+{
+    public class ResourceSearchParameterStatus
+    {
+        public Uri Uri { get; set; }
+
+        public SearchParameterStatus Status { get; set; }
+
+        public bool? IsPartiallySupported { get; set; }
```
LTA-Thinking, line 16: What is the benefit of using the nullable boolean? It looks like you only set it to 'true' or don't set it.

build/jobs/provision-deploy.yml
brendankowitz: True, this is already merged, I pulled this out into a separate PR, will update from master

src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterStatusManager.cs
```diff
@@ -0,0 +1,80 @@
...
+            var updated = new List<SearchParameterInfo>();
+            var newParameters = new List<ResourceSearchParameterStatus>();
+
+            var parameters = (await _searchParameterRegistry.GetSearchParameterStatuses())
+                .ToDictionary(x => x.Uri);
+
+            // Set states of known parameters
+            foreach (var p in _searchParameterDefinitionManager.AllSearchParameters)
+            {
+                if (parameters.TryGetValue(p.Url, out var result))
```
namadabu, line 50: nit: can we replace `var` with explicit type here (and line 48). It's not evident what type they are and makes the code less readable.

src/Microsoft.Health.Fhir.Shared.Core/Features/Definition/SearchParameterDefinitionBuilder.cs
brendankowitz: Apparently in this case the same (https://docs.microsoft.com/en-us/dotnet/api/system.uri.op_equality?view=netcore-3.1)

```
This overload uses the Equals method to determine whether the two Uri instances are equivalent.
```

src/Microsoft.Health.Fhir.Core/Features/Search/Registry/ResourceSearchParameterStatus.cs
brendankowitz: Original thought is that I only know explicitly when Partial Support is true. The inverse is unset / unknown but not explicitly false, perhaps I'm over thinking it.

src/Microsoft.Health.Fhir.Stu3.Core.UnitTests/Microsoft.Health.Fhir.Stu3.Core.UnitTests.csproj
```diff
@@ -3,6 +3,9 @@
   <PropertyGroup>
     <TargetFramework>netcoreapp3.1</TargetFramework>
   </PropertyGroup>
+  <PropertyGroup>
+    <DefineConstants>Stu3</DefineConstants>
+  </PropertyGroup>
```
LTA-Thinking, line 8: Is this needed in the other two versions' files?

brendankowitz: 👍Added for consistency

PR #967, opened 2020-04-01
Updates error for invalid or unsupported _total parameter value

rotodd:
## Description

Previously, when an unsupported or invalid `_total` parameter value was provided, the error message that was returned was:

`"The requested \"_total\" parameter is not supported."`

This PR updates the error message to be:

`"The '_total' parameter value 'estimate' is unsupported. The supported values are: 'accurate', 'none'."` (in the case of an unsupported value)

`"The '_total' parameter value 'x' is invalid. The supported values are: 'accurate', 'none'."` (in the case of an invalid value)

total-error-messages

## Related issues
Addresses [73289](https://microsofthealth.visualstudio.com/Health/_workitems/edit/73289).

## Testing

Updates existing E2E tests to validate error message.

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
@@ -5,6 +5,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using System.Text.RegularExpressions;
 using EnsureThat;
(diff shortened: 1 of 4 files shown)


src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
```diff
@@ -92,19 +93,21 @@ public SearchOptions Create(string compartmentType, string compartmentId, string
                 }
                 else if (string.Compare(query.Item1, KnownQueryParameterNames.Total, StringComparison.OrdinalIgnoreCase) == 0)
                 {
+                    // Estimate is not yet supported.
+                    var supportedTotalTypes = new string($"'{TotalType.Accurate}', '{TotalType.None}'").ToLower(CultureInfo.CurrentCulture);
+
```
rbans96, line 98: Is there any specific reason to assign string through string constructor?
var supportedTotalTypes = $"'{TotalType.Accurate}', '{TotalType.None}'".ToLower(CultureInfo.CurrentCulture);

PR #968, opened 2020-04-03
Run the container image as a non-root user

chgl:
## Description
This adds a new, non-root user (called "nonroot") to the created container. The FHIR server is run by this user by default.

Using a user other than root is generally considered a  [good practice](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user). The (breaking?) consequence of this change is that ports < 1024 can not be bound, so the default listening port was set to 8080 instead of 80.

## Testing
This was tested locally by building the container using the instructions in the README, running it via the provided compose.yaml and writing and reading a few resources to the server.


chgl:
Sorry, I just noticed I missed updating some ports!

LTA-Thinking:
Thank you for the contribution. We are currently in the middle of reworking our automated builds and you need to merge from our master branch to pick up some of the new build scripts for the PR pass its checks.

LTA-Thinking:
In the docker Readme this also changes the port that the metadata endpoint will be at. The references to localhost/metadata need to be replaced with localhost:8080/metadata.

PR #971, opened 2020-04-06
Implements apply version command

rbans96:
## Description
Implements command Apply version (--version) in tool
The apply command executes the migration script for the given version if all the preconditions are met.

## Related issues
Addresses [AB#68810](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/68810)

## Testing
Manual tests(by updating the entries in schema table)
image


tools/FhirSchemaManager/Model/CompatibleVersion.cs
```diff
@@ -0,0 +1,25 @@
...
+
+using EnsureThat;
+
+namespace FhirSchemaManager.Model
+{
+    public class CompatibleVersion
+    {
+        public CompatibleVersion(int min, int max)
+        {
+            EnsureArg.IsNotNull<int>(min, nameof(min));
```
brendankowitz, line 14: What was the intent here? Int should never be null

tools/FhirSchemaManager/SchemaDataStore.cs
```diff
@@ -0,0 +1,53 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System.Data;
+using System.Data.SqlClient;
+
+namespace FhirSchemaManager
+{
+    public static class SchemaDataStore
+    {
+        public const string DeleteQuery = "DELETE FROM dbo.SchemaVersion WHERE Version = {0} AND Status = 'failed'";
```
brendankowitz, line 13: I think we should always use parameterized sql where we need to specify a value

tools/FhirSchemaManager/Commands/ApplyCommand.cs
```diff
@@ -4,16 +4,104 @@
...
     {
-        public static void Handler(string connectionString, Uri fhirServer, int version)
+        private static ISchemaClient schemaClient;
+
+        public static async Task HandlerAsync(string connectionString, Uri fhirServer, int version)
         {
-            Console.WriteLine($"--connection-string {connectionString}");
-            Console.WriteLine($"--fhir-server {fhirServer}");
-            Console.WriteLine($"--version {version}");
+            schemaClient = new SchemaClient(fhirServer);
```
brendankowitz, line 26: We shouldn't have a static variable that is set each time this method is called

tools/FhirSchemaManager/Commands/ApplyCommand.cs
```diff
@@ -4,16 +4,104 @@
...
+                return;
+            }
+        }
+
+        private static async Task ValidateVersion(int version)
+        {
+            CompatibleVersion compatibleVersion = await schemaClient.GetCompatibility();
+
+            // check if version lies in the compatibility range
+            if (!Enumerable.Range(compatibleVersion.Min, compatibleVersion.Max).Contains(version))
```
brendankowitz, line 93: Range will generate the items, its probably more efficient here to just use a logic check

tools/FhirSchemaManager/Commands/ApplyCommand.cs
```diff
@@ -4,16 +4,102 @@
...
     {
-        public static void Handler(string connectionString, Uri fhirServer, int version)
+        public static async Task HandlerAsync(string connectionString, Uri fhirServer, int version)
         {
-            Console.WriteLine($"--connection-string {connectionString}");
-            Console.WriteLine($"--fhir-server {fhirServer}");
-            Console.WriteLine($"--version {version}");
+            ISchemaClient schemaClient = new SchemaClient(fhirServer);
+
+            var region = new Region(
```
brendankowitz, line 26: region looks unused (from https://lgtm.com/projects/g/microsoft/fhir-server/rev/pr-2103ec5295e7746d520a6459ededf49176d5d02b)

tools/FhirSchemaManager/Commands/ApplyCommand.cs
```diff
@@ -4,16 +4,102 @@
...
+                    .ToList();
+
+                foreach (AvailableVersion availableVersion in availableVersions)
+                {
+                    string script = await schemaClient.GetScript(availableVersion.Script);
+
+                    await ValidateVersion(schemaClient, availableVersion.Id);
+
+                    // check if the record for given version exists in failed status
+                    SchemaDataStore.ExecuteQuery(connectionString, string.Join(SchemaDataStore.DeleteQuery, availableVersion.Id, SchemaDataStore.Failed), availableVersion.Id);
```
brendankowitz, line 58: Should this be Format instead of Join?
`string.Join(SchemaDataStore.DeleteQuery, availableVersion.Id, SchemaDataStore.Failed)`.

A better way would be with sql parameters, example here: https://docs.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqlcommand.parameters?view=netframework-4.8#examples

tools/FhirSchemaManager/SchemaDataStore.cs
```diff
@@ -0,0 +1,54 @@
...
+                    throw;
+                }
+            }
+        }
+
+        public static void ExecuteUpsertQuery(string connectionString, int version, string status)
+        {
+            using (var connection = new SqlConnection(connectionString))
+            {
+                var upsertCommand = new SqlCommand("dbo.UpsertSchemaVersion", connection)
```
brendankowitz, line 41: SqlCommand should also have a using

tools/FhirSchemaManager/Commands/ApplyCommand.cs
rbans96: Ah yes, Fixed

tools/FhirSchemaManager/Commands/ApplyCommand.cs
```diff
@@ -4,16 +4,97 @@
...
+            {
+                CommandUtils.PrintError(string.Format(Resources.QueryExecutionErrorMessage, ex.Message));
+                return;
+            }
+        }
+
+        private static async Task ValidateVersion(ISchemaClient schemaClient, int version)
+        {
+            // to ensure server side polling is completed
+            await Task.Delay(60000);
```
rbans96, line 81: Adding a minute delay because server side polling is scheduled at each minute interval. So to ensure it does not read any stale value between schema migration.

PR #978, opened 2020-04-14
Switch CI build to use yaml and docker

LTA-Thinking:
## Description
Moves the testing portion of the CI build to yaml and has the tests use docker.

As a part of this change the CI release pipeline will have the E2E tests removed as they are now done by the build pipeline. In addition the release pipeline will add a duplicate of the CI's docker image that has been tagged with a release number. The release pipeline will be automatically triggered but require manual approval to start, as is currently the case with the nuget and github release steps.


build/ci-variables.yml
```diff
@@ -0,0 +1,8 @@
+variables:
+    ResourceGroupRegion: 'southcentralus'
+    resourceGroupRoot: 'msh-fhir-ci'
+    appServicePlanResourceGroup: '$(resourceGroupRoot)-linux'
+    appServicePlanName: '$(resourceGroupRoot)-linux'
+    DeploymentEnvironmentName: '$(resourceGroupRoot)-d'
```
brendankowitz, line 6: What is `-d` for?

build/ci-pipeline.yml
```diff
@@ -0,0 +1,160 @@
...
+
+name: $(SourceBranchName)-$(Date:yyyyMMdd)$(Rev:-r)
+trigger: none
+
+variables:
+- template: ci-variables.yml
+- template: build-variables.yml
+
+stages:
+- stage: BuildUnitTests
```
namadabu, line 12: Just trying to understand where this build will be triggered from? Will it happen from master branch after a PR gets merged in?

build/ci-pipeline.yml
```diff
@@ -0,0 +1,160 @@
...
+      parameters:
+        packageArtifacts: false
+
+- stage: DockerBuild  
+  displayName: 'Build images'
+  dependsOn: []
+  jobs:
+  - template: ./jobs/docker-build-all.yml
+    parameters: 
+      tag: $(ImageTag)
```
namadabu, line 36: Kind of a tangential question. At what point will we assign the `<fhirVersion>:latest` tag to the latest build?

build/ci-variables.yml
LTA-Thinking: It's short for docker. I couldn't have it in the same resource group as the current CI resources, so I needed to add something to the name.

build/ci-pipeline.yml
LTA-Thinking: Yes, the trigger is set up in devops to run when something is merged to master.

build/ci-pipeline.yml
LTA-Thinking: The latest tag is a bit of a problem right now. The way that the docker repo works the latest tag is assigned to the image that was made most recently. This could be a CI build, release build, or PR build. The best option I think we have is to disable the latest build, but we haven't talked about this yet.

namadabu: Okay. Did not know that. I think we should update the deployment templates to use `master` as the default value for the `imagetag` parameter instead of using `latest`. So that we can prevent template users from deploying a pr build by mistage.

build/ci-variables.yml
brendankowitz: Let's rename it to replace the old ones

brendankowitz:
Can we also update the status badge at the top of the readme:
`[Build Status](https://microsofthealthoss.visualstudio.com/FhirServer/_build/latest?definitionId=27&branchName=master)`

build/ci-pipeline.yml
LTA-Thinking: Good idea. I changed it to ```release``` instead of ```master``` and I'll add that tag to the release pipeline, so it always corresponds to the latest release. That way there isn't the thought that the image is our current master branch.

brendankowitz: If it becomes a problem in future we could also push the PR images to a different registry. Come to think of it, it could also make it easier to clean things up

PR #979, opened 2020-04-15
Adds CosmosDb SearchParameterRegistry

brendankowitz:
## Description
Adds storage of the status of a search parameter into cosmosDb

## Related issues
Addresses [AB#73171](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/73171).

## Testing
Adds unit tests for new components

### DIFF
--- a/src/Microsoft.Health.Fhir.CosmosDb.UnitTests/Features/Storage/Registry/CosmosDbStatusRegistryInitializerTests.cs
+++ b/src/Microsoft.Health.Fhir.CosmosDb.UnitTests/Features/Storage/Registry/CosmosDbStatusRegistryInitializerTests.cs
@@ -1,0 +1,94 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.Documents;
+using Microsoft.Azure.Documents.Client;
+using Microsoft.Azure.Documents.Linq;
+using Microsoft.Health.Core;
+using Microsoft.Health.CosmosDb.Features.Storage;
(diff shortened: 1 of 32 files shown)


src/Microsoft.Health.Fhir.CosmosDb/Registration/FhirServerBuilderCosmosDbRegistrationExtensions.cs
```diff
@@ -101,11 +103,15 @@ private static IFhirServerBuilder AddCosmosDbPersistence(this IFhirServerBuilder
                 .AsService<ICollectionInitializer>();
 
             services.Add<FhirCollectionSettingsUpdater>()
-                .Singleton()
+                .Transient()
```
feordin, line 106: Changing these to transient will cause a new instance for each request, correct?  Is that necessary for these types?

src/Microsoft.Health.Fhir.CosmosDb.UnitTests/Features/Storage/Registry/CosmosDbStatusRegistryInitializerTests.cs
```diff
@@ -0,0 +1,93 @@
...
+
+            await _initializer.ExecuteAsync(documentClient, new DocumentCollection(), relativeCollectionUri);
+
+            await documentClient.Received().UpsertDocumentAsync(
+                relativeCollectionUri,
+                Arg.Any<SearchParameterStatusWrapper>());
+        }
+
+        [Fact]
+        public async Task GivenARegistryInitializer_WhenDatabaseIsNew_SearchParametersShouldBeUpserted()
```
feordin, line 73: Is the title correct for this test case?  It seems like we are verifying that we did not upsert anything because the database was not new.

src/Microsoft.Health.Fhir.CosmosDb/Registration/FhirServerBuilderCosmosDbRegistrationExtensions.cs
brendankowitz: I might review again, but these are only used once as far as I know. Having them as singleton means the container will hold onto the reference

PR #980, opened 2020-04-15
Implements --next option

rbans96:
## Description
Implements --next option for apply command

Command syntax-
>FhirSchemaManager apply --connection-string "server=(local);Initial Catalog=FHIR_R4;Integrated Security=true" --fhir-server https://localhost:44348 --next

## Related issues
Addresses #73489.

## Testing
Manual testing


tools/FhirSchemaManager/Commands/ApplyCommand.cs
```diff
@@ -38,7 +37,7 @@ public static async Task HandlerAsync(string connectionString, Uri fhirServer, i
                     availableVersions.RemoveAt(0);
                 }
 
-                availableVersions = availableVersions.Where(availableVersion => availableVersion.Id <= version)
+                availableVersions = availableVersions.Where(availableVersion => availableVersion.Id <= (next == true ? availableVersions.First().Id : version))
```
brendankowitz, line 40: nit: simplify into a variable `maxVersion` maybe?

rbans96: In the latest version PR, the version is moved to MutualExclusiveType class and then accessed as MutualExclusiveType.version.

PR #986, opened 2020-04-20
allows use of shared hosting plan in PR builds

brendankowitz:
## Description
Changes PR provisioning to put all sites into a single app service plan

### DIFF
--- a/build/jobs/provision-deploy.yml
+++ b/build/jobs/provision-deploy.yml
@@ -47,13 +47,15 @@ jobs:
      
         $additionalProperties = $flattenedCors
      
-        $additionalProperties["SqlServer__DeleteAllDataOnStartup"] = "true"
+        $additionalProperties["SqlServer__DeleteAllDataOnStartup"] = "false"
+        $additionalProperties["SqlServer__AllowDatabaseCreation"] = "true"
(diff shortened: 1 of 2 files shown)


build/jobs/provision-deploy.yml
```diff
@@ -47,13 +47,15 @@ jobs:
...
         $additionalProperties = $flattenedCors
      
-        $additionalProperties["SqlServer__DeleteAllDataOnStartup"] = "true"
+        $additionalProperties["SqlServer__DeleteAllDataOnStartup"] = "false"
+        $additionalProperties["SqlServer__AllowDatabaseCreation"] = "true"
         $additionalProperties["ASPNETCORE_FORWARDEDHEADERS_ENABLED"] = "true"
 
         $templateParameters = @{
             fhirVersion = "${{ parameters.version }}"
             appServicePlanResourceGroup = "${{ parameters.appServicePlanResourceGroup }}"
```
LTA-Thinking, line 56: To make a app service plan the resource group shouldn't be passed. There is a condition that is currently missing from the docker scripts that is present in the old scripts to prevent creating a new app service if an app service resource group is passed. This is being put back in in PR #988.

PR #988, opened 2020-04-21
Fix docker deploy template

LTA-Thinking:
## Description
Have the docker deployment templates respect fhir version and ask for image tag.


build/jobs/run-tests.yml
```diff
@@ -88,18 +88,18 @@ jobs:
...
       Inline: |
         $secrets = Get-AzKeyVaultSecret -VaultName $(DeploymentEnvironmentName)-ts
-
+        Write-Host "From TS"
         foreach($secret in $secrets)
         {
             $environmentVariableName = $secret.Name.Replace("--","_")
 
             $secretValue = Get-AzKeyVaultSecret -VaultName $(DeploymentEnvironmentName)-ts -Name $secret.Name
+            Write-Host "Overwriting $environmentVariableName with value $secretValue"
```
namadabu, line 97: Do we need to log this here? Or is it just for debugging purposes?

LTA-Thinking: Good catch, removed

Issue #999, opened 2020-04-30
Potential ExportJobWorker continuous loop


brendankowitz:
**Describe the bug**
Is seems that in the following code in [ExportJobWorker](https://github.com/microsoft/fhir-server/blob/master/src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobWorker.cs#L46) that when an exception is thrown the loop will immediately retry. If the exception continues to occur, the loop continiously tries without delay. This causes a large number of logged exception.

```
while (!cancellationToken.IsCancellationRequested)
{
    try
    {
        // DoWork();
        
        await Task.Delay(_exportJobConfiguration.JobPollingFrequency, cancellationToken);
    }
    catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
    {
        // Cancel requested.
    }
    catch (Exception ex)
    {
        // The job failed.
        _logger.LogError(ex, "Unhandled exception in the worker.");
    }
}
```

```
Microsoft.Azure.Documents.ForbiddenException: Message: {"Errors":["The script with id 'p-FdAJ7nEGgFAAAAAAAAgA==' is blocked for execution because it has violated its allowed resource limit several times."]} 

PartitionKeyRangeId: 0, IsValid: True, 
StatusCode: 403, SubStatusCode: 0, RequestCharge: 0.62, 
ResourceType: StoredProcedure, OperationType: ExecuteJavaScript , 
SDK: Windows/10.0.14393 documentdb-netcore-sdk/2.9.2
at Microsoft.Azure.Documents.TransportClient.ThrowServerException(String resourceAddress, StoreResponse storeResponse, Uri physicalAddress, Guid activityId, DocumentServiceRequest request) 
at Microsoft.Azure.Documents.Rntbd.TransportClient.InvokeStoreAsync(Uri physicalAddress, ResourceOperation resourceOperation, DocumentServiceRequest request)
at Microsoft.Azure.Documents.ConsistencyWriter.WritePrivateAsync(DocumentServiceRequest request, TimeoutHelper timeout, Boolean forceRefresh) 
at Microsoft.Azure.Documents.BackoffRetryUtility`1.ExecuteRetryAsync(Func`1 callbackMethod, Func`3 callShouldRetry, Func`1 inBackoffAlternateCallbackMethod, TimeSpan minBackoffForInBackoffCallback, CancellationToken cancellationToken, Action`1 preRetryCallback) 
[...]

PR #1003, opened 2020-05-04
Fixed continuous loop in ExportJobWorker when exception is thrown

namadabu:
## Description
In the ExportJobWorker we don't respect the time delay when an exception is thrown within the `ExecuteAsync` loop. This causes us to continuously retry.

This code fixes that. It also introduces a mechanism of increasing the delay between retries as long as we encounter exceptions within the execute loop.

## Related issues
Addresses [issue #999].

## Testing
Added UTs.

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobWorker.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobWorker.cs
@@ -96,6 +96,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Export
                 {
                     // Cancellation requested
                     _logger.LogInformation("ExportJobWorker: Cancellation requested.");
+                    break;
                 }
             }
         }
--- a/src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobWorker.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobWorker.cs
@@ -26,6 +26,8 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Export
         private readonly Func<IExportJobTask> _exportJobTaskFactory;
         private readonly ILogger _logger;
 
+        private TimeSpan _delayBeforeNextPoll;
+
         public ExportJobWorker(Func<IScoped<IFhirOperationDataStore>> fhirOperationDataStoreFactory, IOptions<ExportJobConfiguration> exportJobConfiguration, Func<IExportJobTask> exportJobTaskFactory, ILogger<ExportJobWorker> logger)
         {
             EnsureArg.IsNotNull(fhirOperationDataStoreFactory, nameof(fhirOperationDataStoreFactory));
@@ -37,6 +39,8 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Export
             _exportJobConfiguration = exportJobConfiguration.Value;
             _exportJobTaskFactory = exportJobTaskFactory;
             _logger = logger;
+
+            _delayBeforeNextPoll = _exportJobConfiguration.JobPollingFrequency;
         }
 
         public async Task ExecuteAsync(CancellationToken cancellationToken)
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobWorker.cs
```diff
@@ -37,6 +39,8 @@ public ExportJobWorker(Func<IScoped<IFhirOperationDataStore>> fhirOperationDataS
             _exportJobConfiguration = exportJobConfiguration.Value;
             _exportJobTaskFactory = exportJobTaskFactory;
             _logger = logger;
+
+            _delayBeforeNextPoll = _exportJobConfiguration.JobPollingFrequency;
```
brendankowitz, line 43: nit: I think move the declaration of `_delayBeforeNextPoll` into `ExecuteAsync` since its only mutating state within that function

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobWorker.cs
```diff
@@ -69,16 +73,30 @@ public async Task ExecuteAsync(CancellationToken cancellationToken)
...
+                    _delayBeforeNextPoll = _exportJobConfiguration.JobPollingFrequency;
                 }
                 catch (Exception ex)
                 {
                     // The job failed.
                     _logger.LogError(ex, "Unhandled exception in the worker.");
+
+                    // Since acquiring jobs failed let us introduce a delay before we retry. We don't want to increase the delay between polls to more than an hour.
+                    _delayBeforeNextPoll *= 2;
+                    if (_delayBeforeNextPoll > TimeSpan.FromSeconds(3600))
```
brendankowitz, line 86: nit: make `TimeSpan.FromSeconds(3600)` a const

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobWorker.cs
namadabu: Moved.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobWorker.cs
namadabu: Updated

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobWorker.cs
```diff
@@ -69,16 +72,31 @@ public async Task ExecuteAsync(CancellationToken cancellationToken)
...
+                    {
+                        delayBeforeNextPoll = _maximumDelay;
+                    }
+                }
+
+                try
+                {
+                    await Task.Delay(delayBeforeNextPoll, cancellationToken);
+                }
+                catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
```
LTA-Thinking, line 95: Is there a chance a cancellation could be requested during the upper try block? It looks like it would be handled in the upper catch block and then would continue looping.

namadabu: Even in that scenario, the `await Task.Delay()` will throw a `TaskCanceledException` which will then be caught by the `catch (OperationCanceledException)` block. The while loop condition also checks for whether cancellation was requested (`while (!cancellationToken.IsCancellationRequested)`). So we will exit the loop one way or the other.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobWorker.cs
```diff
@@ -69,16 +72,31 @@ public async Task ExecuteAsync(CancellationToken cancellationToken)
...
+                {
+                    // Cancellation requested
+                    _logger.LogInformation("ExportJobWorker: Cancellation requested.");
```
deepak-ms, line 98: nit: "while (!cancellationToken.IsCancellationRequested)" exit path doesn't have this log entry. Should we remove (or add info in message of) the log here to avoid confusion while analyzing logs?

namadabu: Since both scenarios are caused by the same underlying cancellation (`Task.Delay()` throws the `TaskCanceledException` when `cancellationToken.IsCancellationRequested == true`), I moved the log outside the while loop. Now the logging will be uniform regardless of where we exit the loop.

PR #1014, opened 2020-05-11
Handle all BYOK errors from Cosmos DB

shmartel:
## Description
Cosmos DB has provide sub status codes and descriptions for all the possible BYOK errors returned from their service.

## Related issues
Addresses [#73777](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/73777)

## Testing
Added unit test coverage


src/Microsoft.Health.Fhir.Core/Exceptions/CmkAadClientCredentialsGrantFailureException.cs
```diff
@@ -0,0 +1,20 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using Microsoft.Health.Fhir.Core.Models;
+
+namespace Microsoft.Health.Fhir.Core.Exceptions
+{
+    public class CmkAadClientCredentialsGrantFailureException : FhirException
```
namadabu, line 10: I don't think these exceptions should be in Fhir.Core since they are specific to CosmosDb

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosResponseProcessor.cs
```diff
@@ -68,9 +68,31 @@ public async Task ProcessException(Exception ex)
                 {
                     throw new Core.Exceptions.RequestEntityTooLargeException();
                 }
-                else if (dce.StatusCode == HttpStatusCode.Forbidden && dce.GetSubStatusValue() == CosmosDbSubStatusValues.CustomerManagedKeyInaccessible)
+                else if (dce.StatusCode == HttpStatusCode.Forbidden)
                 {
-                    throw new Core.Exceptions.CustomerManagedKeyInaccessibleException();
+                    switch (dce.GetSubStatusValue())
+                    {
+                        case CosmosDbSubStatusValues.CmkAadClientCredentialsGrantFailure:
```
namadabu, line 75: Since these exceptions all deal with the same underlying component, would it be easier to just have a single exception that takes in the sub-status code and creates an `OperationOutcomeIssue` with the appropriate `Resource` message? The rest of the code then has to handle just this one exception type instead of multiple.

src/Microsoft.Health.Fhir.CosmosDb.UnitTests/Features/Storage/CosmosResponseProcessorTests.cs
```diff
@@ -156,19 +156,109 @@ public async Task GivenADocumentClientExceptionWithSpecificMessage_WhenProcessin
...
+            DocumentClientException documentClientException = CreateDocumentClientException("12.4", "fail", HttpStatusCode.Forbidden, CosmosDbSubStatusValues.CmkAadClientCredentialsGrantFailure.ToString());
 
-            await Assert.ThrowsAsync<CustomerManagedKeyInaccessibleException>(async () => await _cosmosResponseProcessor.ProcessException(documentClientException));
+            await Assert.ThrowsAsync<CmkAadClientCredentialsGrantFailureException>(async () => await _cosmosResponseProcessor.ProcessException(documentClientException));
+
+            ValidateExecution(expectedSessionToken: null, 12.4, false);
+        }
+
+        [Fact]
+        public async Task GivenADocumentClientExceptionWithCmkAadServiceUnavailableSubStatus_WhenProcessing_ThenExceptionShouldThrow()
```
namadabu, line 169: nit: Since all these tests differ only in the sub-status code that is passed to `CreateDocumentClientException`, you can just make the test a `[Theory]` and pass the different codes as `InlineData`. Similar to the one below: `GivenADocumentClientExceptionWithForbiddenStatusCodeAndUnknownSubStatus_WhenProcessing_ThenNothingElseShouldOccur`

namadabu: Never mind, just realized that it checks for a different exception in every test.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosResponseProcessor.cs
shmartel: That sounds like a great idea, will do.

src/Microsoft.Health.CosmosDb/Features/Storage/CosmosDbSubStatusValues.cs
```diff
@@ -7,6 +7,16 @@ namespace Microsoft.Health.CosmosDb.Features.Storage
 {
     public static class CosmosDbSubStatusValues
```
brendankowitz, line 8: `KnownCosmosDbSubStatusValues` would follow the convention we've used in other parts of the codebase for these types of specified values

src/Microsoft.Health.Fhir.Core/Exceptions/CmkAadClientCredentialsGrantFailureException.cs
brendankowitz: SQL also supports Customer Managed Keys, @shmartel do you know if these would be reused if we also support this in future?

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosResponseProcessor.cs
shmartel: This is way better, thanks!

src/Microsoft.Health.Fhir.CosmosDb.UnitTests/Features/Storage/CosmosResponseProcessorTests.cs
shmartel: Per your other suggestion, I now have only one exception.  Thus, I was able to use the [Theory].  Thanks!

src/Microsoft.Health.Fhir.Core/Resources.resx
```diff
@@ -175,8 +175,35 @@
...
+  </data>
+  <data name="CmkAadServiceUnavailable" xml:space="preserve">
+    <value>AAD service unavailable to get access to the customer managed key.</value>
+  </data>
+  <data name="CmkKeyVaultAuthenticationFailure" xml:space="preserve">
+    <value>The customer managed key is disabled, or the Key Vault does not grant permission to AAD.</value>
+  </data>
+  <data name="CmkKeyVaultKeyNotFound" xml:space="preserve">
+    <value>The customer managed key is not found.</value>
+  </data>
```
shmartel, line 189: Grammar:  I think customer-managed key is correct.  Also, when I look at the BYOK documentation guidance, a '-' is used.  I will switch to that.

src/Microsoft.Health.Fhir.Core/Exceptions/CmkAadClientCredentialsGrantFailureException.cs
shmartel: This is a good observation.  As they were originally implemented in my first commit, I agree they were specific to Cosmos DB, as SQL would not have the same set of errors, or their errors messages would be different.  

In my 2nd commit, I now have a generic CusotmerManagedKeyException that takes an error code input, and populates the error OperationOutcomeIssue with the corresponding message.  So, at first glance, it seems that this exception could get reused for SQL, as their error codes seem to be different: https://docs.microsoft.com/en-us/sql/relational-databases/security/encryption/sql-server-connector-maintenance-troubleshooting?view=sql-server-2017#AppendixC.  So, in theory, this exception logic could check both CosmosDB and SQL error codes and return the correct error message.

But, what happens if some day the Cosmos DB and SQL error codes overlap?  Or, one of the two have some undocumented error codes that overlap.  Then the code doesn't know which error message to use.  Unlikely, but possible.

So, I will play it safe and move this to the Cosmos DB project.  I'm definitely open to suggestions on this!

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosResponseProcessor.cs
shmartel, line 75: In commit #2 I will have one generic CMK exception, per feedback.  In this case, I think this code becomes much simpler if CosmosDbSubStatusValues is an enum.  I've done so; open to feedback.

src/Microsoft.Health.Fhir.Core/Exceptions/CmkAadClientCredentialsGrantFailureException.cs
shmartel: Commit 3 shows what the code looks like when CustomerManagedKeyException is moved into Microsoft.Health.Fhir.CosmosDb.  Thoughts, @namadabu , @brendankowitz ?  Given that it no longer has any Cosmos-specific types in it, I can see any argument that it could live in Core, with the caveat of an error code value collision (unlikely, but possible).

src/Microsoft.Health.CosmosDb/Features/Storage/KnownCosmosDbCmkSubStatusValue.cs
```diff
@@ -0,0 +1,25 @@
+// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.CosmosDb.Features.Storage
+{
+    /// <summary>
+    /// Cosmos DB sustomer-managed key (CMK) sub status values
```
namadabu, line 9: nit: typo - customer

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Filters/OperationOutcomeExceptionFilterTests.cs
```diff
@@ -200,9 +201,9 @@ public void GivenARequestEntityTooLargeException_WhenExecutingAnAction_ThenTheRe
         }
 
         [Fact]
-        public void GivenACustomerManagedKeyInaccessibleException_WhenExecutingAnAction_ThenTheResponseShouldBeAnOperationOutcome()
+        public void GivenAadClientCredentialsGrantFailureException_WhenExecutingAnAction_ThenTheResponseShouldBeAnOperationOutcome()
```
namadabu, line 204: Should the name of the test start with GivenACustomerManagedKeyException ?

src/Microsoft.Health.Fhir.Core/Exceptions/CmkAadClientCredentialsGrantFailureException.cs
namadabu: @shmartel Now that you have made the exception generic in nature, I feel that it would be better to move this to Core.Exceptions. This way it will be consistent across both the data layers (when we support BYOK for SQL). If there is some kind of collision of status code values in the future, we can always re-factor to move the status code -> error message mapping logic to the respective data layer and still keep `CustomerManagedKeyException`. Thoughts?

Also, one issue with the current implementation is that the Api layer now needs to 'know' about the `CosmosDb` layer in order to understand `CustomerManagedKeyException`. We should avoid that kind of dependency.

shmartel: Makes sense.  I will revert my last commit which puts the exception back in Core.

src/Microsoft.Health.Fhir.Core/Microsoft.Health.Fhir.Core.csproj
```diff
@@ -44,6 +44,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\Microsoft.Health.CosmosDb\Microsoft.Health.CosmosDb.csproj" />
```
brendankowitz, line 47: Core shouldn't reference CosmosDb

shmartel: Yes.  This was something I did try, but I expected this change to get reverted when I reverted to  Commit #3 to put the exception back in it's original location.  I'm not sure why it didn't.  Investigating...

shmartel: I've solved this by moving KnownCosmosDbCmkSubStatusValue into Core\Exceptions, so it lives alongside the CustomerManagedKeyException that leverages it.  The downside here is that we now have Cosmos DB related "info" in Core.  But, I couldn't find a better location that avoided the circular reference.  I'm open to suggestions.

shmartel: Based on an offline suggestion from @namadabu , I've moved KnownCosmosDbCmkSubStatusValue and the calculation of the BYOK error message into Microsoft.Health.Fhir.CosmosDb.  This removes any Cosmos-related knowledge from Core

src/Microsoft.Health.Fhir.Core/Resources.resx
```diff
@@ -238,7 +205,7 @@
   <data name="IdRequirements" xml:space="preserve">
     <value>Id must be any combination of upper or lower case ASCII letters ('A'..'Z', and 'a'..'z'), numerals ('0'..'9'), '-' and '.', with a length limit of 64 characters. (This might be an integer, an un-prefixed OID, UUID, or any other identifier pattern that meets these constraints.)</value>
   </data>
-  <data name="IfMatchHeaderRequiredForResource" xml:space="preserve">
```
shmartel, line 211: It appears my editor has removed some spacing from this file, unfortunatley.  I will fix.

src/Microsoft.Health.Fhir.Core/Exceptions/CustomerManagedKeyException.cs
```diff
@@ -7,14 +7,14 @@
...
-    public class CustomerManagedKeyInaccessibleException : FhirException
+    public class CustomerManagedKeyException : FhirException
     {
-        public CustomerManagedKeyInaccessibleException()
+        public CustomerManagedKeyException(string message)
         {
             Issues.Add(new OperationOutcomeIssue(
                     OperationOutcomeConstants.IssueSeverity.Error,
                     OperationOutcomeConstants.IssueType.Forbidden,
-                    Resources.CustomerManagedKeyInaccessible));
```
namadabu, line 17: Should we call the base constructor here so that the `message` gets logged?

shmartel: That sounds good, thanks.  I've made the change.

PR #1035, opened 2020-05-21
CI pipeline - Clear old data from storage accounts after e2e tests

namadabu:
## Description
The Export E2E tests create some containers in the storage accounts. Since the storage accounts for CI are persistent, we need a mechanism to clear the old data from the storage accounts from time to time. This change adds a step to the CI pipeline that takes care of removing containers older than one day.

## Related issues
Addresses [issue [AB#74093](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/74093)].

## Testing

### DIFF
--- a/build/ci-pipeline.yml
+++ b/build/ci-pipeline.yml
@@ -145,6 +145,7 @@ stages:
 
 - stage: cleanStorageAccounts
   displayName: 'Clean Storage Accounts'
+  dependsOn: []
   jobs:
     - template: ./jobs/clean-storage-accounts.yml
       parameters:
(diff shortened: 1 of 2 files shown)


build/ci-pipeline.yml
```diff
@@ -143,6 +143,13 @@ stages:
       version: R5
       keyVaultName: $(DeploymentEnvironmentNameR5)
 
+- stage: cleanStorageAccounts
+  displayName: 'Clean Storage Accounts'
+  jobs:
```
LTA-Thinking, line 148: Nit: you could add 'dependsOn: []' so that this runs every time the pipeline runs. Since it is only deleting data over a day old it shouldn't affect currently running tests.

namadabu: Updated

PR #1036, opened 2020-05-21
Extract FhirClient to separate library

brandonpollett:
## Description
This PR take the `FhirClient` in the E2E project and moves it to a separate `Microsoft.Health.Fhir.Client` library. 

## Related issues
Addresses [AB#74009](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/74009).

## Testing
Existing tests that make use of the `FhirClient` continue to pass.


Microsoft.Health.Fhir.sln
```diff
@@ -130,6 +130,14 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Health.Fhir.R5.Te
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Health.Fhir.R5.Tests.Integration", "test\Microsoft.Health.Fhir.R5.Tests.Integration\Microsoft.Health.Fhir.R5.Tests.Integration.csproj", "{E18E488B-EE9C-4735-81B1-65C74DAE83CF}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.Health.Fhir.R4.Client", "src\Microsoft.Health.Fhir.R4.Client\Microsoft.Health.Fhir.R4.Client.csproj", "{C65993F8-E77B-4F7A-BE49-5CAE9DCA3162}"
```
brandonpollett, line 133: Move these into a separate folder within the solution structure. Maybe `client` or something similar.

src/Microsoft.Health.Fhir.Shared.Client/FhirClient.cs
```diff
@@ -14,42 +13,59 @@
...
-        private readonly TestFhirServer _testFhirServer;
-        private readonly TestApplication _clientApplication;
-        private readonly TestUser _user;
-        private readonly Dictionary<string, string> _bearerTokens = new Dictionary<string, string>();
+        public const string SmartOAuthUriExtension = "http://fhir-registry.smarthealthit.org/StructureDefinition/oauth-uris";
+        public const string SmartOAuthUriExtensionToken = "token";
+        public const string SmartOAuthUriExtensionAuthorize = "authorize";
+
+        private const string IfNoneExist = "If-None-Exist";
+        private const string IfMatch = "If-Match";
```
jackliums, line 31: nit: maybe IfMatchHeaderName

src/Microsoft.Health.Fhir.Shared.Client/FhirClient.cs
```diff
@@ -81,31 +97,30 @@ public class FhirClient
...
             _mediaType = MediaTypeWithQualityHeaderValue.Parse(_contentType);
-            SetupAuthenticationAsync(clientApplication, user).GetAwaiter().GetResult();
+
+            SetupSecurity();
         }
 
         public ResourceFormat Format { get; }
 
-        public (bool SecurityEnabled, string AuthorizeUrl, string TokenUrl) SecuritySettings { get; }
+        public bool SecurityEnabled { get; private set; }
```
jackliums, line 106: Doesn't have to be in this PR, but I feel like we should separate the client and authentication logic at some point.

src/Microsoft.Health.Fhir.Shared.Client/FhirClient.cs
```diff
@@ -374,67 +389,26 @@ private async Task<FhirResponse<T>> CreateResponseAsync<T>(HttpResponseMessage r
                 string.IsNullOrWhiteSpace(content) ? null : (T)_deserialize(content));
         }
 
-        private async Task SetupAuthenticationAsync(TestApplication clientApplication, TestUser user = null)
+        private void SetupSecurity()
```
jackliums, line 392: We could also make this method public and have the caller explicitly initiate authentication process.

src/Microsoft.Health.Fhir.Shared.Client/FhirClientAuthExtensions.cs
```diff
@@ -0,0 +1,112 @@
...
+                new KeyValuePair<string, string>(OpenIdConnectParameterNames.Password, password),
+            };
+
+            await ObtainTokenAndSetOnClient(fhirClient, formData);
+        }
+
+        private static async Task ObtainTokenAndSetOnClient(FhirClient fhirClient, List<KeyValuePair<string, string>> formData)
+        {
+            using var formContent = new FormUrlEncodedContent(formData);
+            HttpResponseMessage tokenResponse = await fhirClient.HttpClient.PostAsync(fhirClient.TokenUri, formContent);
```
jackliums, line 105: does tokenResponse need to be disposed?

src/Microsoft.Health.Fhir.Shared.Client/FhirResponse{T}.cs
```diff
@@ -23,5 +23,10 @@ public FhirResponse(HttpResponseMessage response, T resource)
         {
             return response.Resource;
         }
+
+        public T ToT()
```
jackliums, line 27: Doesn't the Resource property above already givce you this?

brandonpollett: Yes, it does. This in response to https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2225?view=vs-2019 which talks about having an alternative member for languages that don't support overloaded operators. I think since this will be a library, it's right to have the method also.

src/Microsoft.Health.Fhir.Shared.Client/FhirClient.cs
brandonpollett: I liked the idea of exposing the `SetupSecurity` call to the using code to do, so I exposed it as the method `ConfigureSecurityOptions`. I also made the `SecurityEnabled` option a `bool?` so it could be known if the method had been called or not.

PR #1044, opened 2020-05-27
Fixes referencing conditional create/update resources in bundle

brandonpollett:
## Description
This PR fixes two separate issues that exist when referencing condtional resources within a transaction bundle.

1. Previously a reference to a resource that is conditionally created would update the reference correctly on the first post of the bundle, but subsequent posts would not reference the correct reference.

2. Previously a reference to a resource that was conditionally updated would not update the reference if the resource was created, but subsequent posts would update to the correct reference.

Also, some minor refactoring in the code to break some of the larger methods into smaller methods within the bundle handler and additional comments were added.

## Related issues
Addresses [AB#74138](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/74138).

## Testing
Created two new end to end tests cover these scenarios.


src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -235,75 +235,82 @@ private async Task FillRequestLists(List<EntryComponent> bundleEntries, Cancella
...
-                    await _referenceResolver.ResolveReferencesAsync(entry.Resource, _referenceIdDictionary, requestUrl, cancellationToken);
+        private async Task GenerateRequest(EntryComponent entry, int order, CancellationToken cancellationToken)
+        {
+            string persistedId = default;
+            HttpContext httpContext = new DefaultHttpContext { RequestServices = _requestServices };
+
+            var requestUrl = entry.Request?.Url;
+
+            // For resources within a transaction, we need to resolve any intrabundle references and potentially persist any internally assigned ids
+            HTTPVerb requestMethod = entry.Request.Method.Value;
```
LTA-Thinking, line 264: The assignment for requestUrl has a null check on Request. Is one needed here too?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -423,17 +374,100 @@ private async Task ExecuteRequests(Hl7.Fhir.Model.Bundle responseBundle, HTTPVer
             }
         }
 
+        private void CreateEntryComponent(HttpContext httpContext, EntryComponent entryComponent)
```
LTA-Thinking, line 377: I think this method should return an EntryComponent instead of taking it as an input argument. The current use case for it is passing in an unmodified new EntryComponent. I think it makes more sense for the method to make the new EntryComponent itself and pass it back.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -423,17 +374,100 @@ private async Task ExecuteRequests(Hl7.Fhir.Model.Bundle responseBundle, HTTPVer
...
+                if (entry.Request.Method == HTTPVerb.PUT && entry.Request.Url.Contains('?', StringComparison.InvariantCultureIgnoreCase) && !string.IsNullOrEmpty(entry.FullUrl))
+                {
+                    if (!idDictionary.ContainsKey(entry.FullUrl))
+                    {
+                        // This id is new to us
+                        var insertId = _resourceIdProvider.Create();
+                        entry.Resource.Id = insertId;
+
+                        idDictionary.Add(entry.FullUrl, (insertId, entry.Resource.TypeName));
+                    }
```
LTA-Thinking, line 456: Nit: add a continue statement.

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Resources/Bundle/TransactionBundleValidatorTests.cs
```diff
@@ -25,18 +23,20 @@ namespace Microsoft.Health.Fhir.Api.UnitTests.Features.Resources.Bundle
...
             _transactionBundleValidator = new TransactionBundleValidator(new ResourceReferenceResolver(_searchService, new QueryStringParser()));
+            _idDictionary = new Dictionary<string, (string resourceId, string resourceType)>();
         }
 
         [Fact]
         public async Task GivenATransactionBundle_WhenContainsUniqueResources_NoExceptionShouldBeThrown()
         {
             var requestBundle = Samples.GetJsonSample("Bundle-TransactionWithValidBundleEntry");
-            await _transactionBundleValidator.ValidateBundle(requestBundle.ToPoco<Hl7.Fhir.Model.Bundle>(), CancellationToken.None);
+            await _transactionBundleValidator.ValidateBundle(requestBundle.ToPoco<Hl7.Fhir.Model.Bundle>(), _idDictionary, CancellationToken.None);
```
LTA-Thinking, line 39: Could you add an assert to show that idDictionary is updated as expected?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/TransactionTests.cs
```diff
@@ -274,5 +275,69 @@ public async Task GivenATransactionBundleReferencesInResourceBody_WhenSuccessful
...
+        {
+            IEnumerable<ResourceReference> imagingStudyReferences = resource.GetAllChildren<ResourceReference>();
+            bool foundReference = false;
+
+            foreach (var reference in imagingStudyReferences)
+            {
+                if (reference.Reference.StartsWith("Patient"))
+                {
+                    Assert.Equal($"Patient/{patientId}", reference.Reference);
+                    foundReference = true;
```
LTA-Thinking, line 336: Would a return statement here make sense?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
brandonpollett: No, where `GenerateRequest` is called above has a null check already for `if (entry.Request?.Method == null)`.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionBundleValidator.cs
```diff
@@ -68,50 +70,47 @@ private static string BuildRequestUrlForConditionalQueries(EntryComponent entry,
...
+                return entry.Request.Method == HTTPVerb.POST ? entry.FullUrl : entry.Request.Url;
+            }
+
+            string resourceType = null;
+            StringValues conditionalQueries;
 
-                return entry.Request.Url;
+            if (entry.Request.Method == HTTPVerb.PUT)
+            {
+                string[] conditionalUpdate = entry.Request.Url.Split('?');
```
jackliums, line 86: not your code but I feel like we shold use existing library to parse URLs instead of implementing this ourselves. For example, I think there is a bug here if one of the query parameter contains value ? (if this were truly URL, additional ? will be encoded as %3F but because this is just a JSON string, i am not sure if it will always be encoded.) If there is additional ? in the value without encoded, then the conditionalQueries below might be wrong?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionBundleValidator.cs
```diff
@@ -68,50 +70,47 @@ private static string BuildRequestUrlForConditionalQueries(EntryComponent entry,
...
-                IReadOnlyCollection<SearchResultEntry> matchedResults = await _referenceResolver.GetExistingResourceId(entry.Request.Url, resourceType, conditionalQueries, cancellationToken);
-                int? count = matchedResults?.Count;
+            if (matchedResults?.Count > 1)
+            {
+                // Multiple matches: The server returns a 412 Precondition Failed error indicating the client's criteria were not selective enough
+                throw new PreconditionFailedException(string.Format(Api.Resources.ConditionalOperationInBundleNotSelectiveEnough, conditionalQueries));
+            }
 
-                if (count > 1)
+            if (matchedResults?.Count == 1)
```
jackliums, line 105: nit: `else if` here too

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionBundleValidator.cs
```diff
@@ -68,50 +70,47 @@ private static string BuildRequestUrlForConditionalQueries(EntryComponent entry,
...
+            {
+                // If this entry has a fullUrl, then save it to the idDictionary for matching later
+                if (!string.IsNullOrWhiteSpace(entry.FullUrl))
```
jackliums, line 108: does something validate that the FullUrl starts with `urn:oid:` in this case somewhere?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -423,17 +377,108 @@ private async Task ExecuteRequests(Hl7.Fhir.Model.Bundle responseBundle, HTTPVer
...
 
-                // We've already come across this ID
-                if (!string.IsNullOrWhiteSpace(entry.FullUrl) && !idDictionary.ContainsKey(entry.FullUrl))
+                if (string.IsNullOrWhiteSpace(entry.FullUrl))
+                {
+                    continue;
+                }
+
+                // We've not come across this ID
+                if (!idDictionary.ContainsKey(entry.FullUrl))
```
jackliums, line 481: the logic below (484-487) is now identical to logic above (461-464). Can we modify the condition to simplify the logic and add some comments on the details?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionBundleValidator.cs
```diff
@@ -32,10 +32,12 @@ public TransactionBundleValidator(ResourceReferenceResolver referenceResolver)
         /// <summary>
         /// This method validates if transaction bundle contains multiple entries that are modifying the same resource.
         /// It also validates if the request operations within a entry is a valid operation.
+        /// If a conditional create or update is executed and a resource exists, the value is populated in the idDictionary.
         /// </summary>
         /// <param name="bundle"> The input bundle</param>
+        /// <param name="idDictionary">The id dictionary that stores fullUrl to actual ids.</param>
         /// <param name="cancellationToken"> The cancellation token</param>
-        public async Task ValidateBundle(Hl7.Fhir.Model.Bundle bundle, CancellationToken cancellationToken)
+        public async Task ValidateBundle(Hl7.Fhir.Model.Bundle bundle, IDictionary<string, (string resourceId, string resourceType)> idDictionary, CancellationToken cancellationToken)
```
jackliums, line 40: maybe not in this PR, but i feel like this class is doing more and more stuff. It's not just validation anymore. it's also loading the resources, building lookup dictionary.

Perhaps split them out by providing like a pipeline that enumerates over each entries in the bundle?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionBundleValidator.cs
```diff
@@ -68,50 +70,47 @@ private static string BuildRequestUrlForConditionalQueries(EntryComponent entry,
             return string.IsNullOrWhiteSpace(conditionalCreateQuery) ? entry.Request.Url : entry.Request.Url + "?" + conditionalCreateQuery;
         }
 
-        private async Task<string> GetResourceId(EntryComponent entry, CancellationToken cancellationToken)
+        private async Task<string> GetResourceId(EntryComponent entry, IDictionary<string, (string resourceId, string resourceType)> idDictionary, CancellationToken cancellationToken)
```
jackliums, line 73: Personal opinion: I think having this value tuple everywhere is a bit hard to read. A struct that holds ResourceType and ResourceId could be a light way to hold these values and shorten the definition a little bit?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionBundleValidator.cs
brandonpollett: I agree. I think given the time, this could be refactored to increase the readability.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionBundleValidator.cs
brandonpollett: No. There's no requirement that the `FullUrl` must start with that, it must only be a valid `URI` - https://www.hl7.org/fhir/datatypes.html#uri.

PR #1051, opened 2020-06-02
Enable Patient Export

LTA-Thinking:
## Description
Enable exporting just Patient objects.

## Testing
Unit and E2E tests were updated to cover the new scenario.

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Api/Controllers/ExportController.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Api/Controllers/ExportController.cs
@@ -82,7 +82,7 @@ namespace Microsoft.Health.Fhir.Api.Controllers
                 throw new RequestNotValidException(string.Format(Resources.OperationNotEnabled, OperationsConstants.Export));
             }
 
-            CreateExportResponse response = await _mediator.ExportAsync(_fhirRequestContextAccessor.FhirRequestContext.Uri, since, HttpContext.RequestAborted);
+            CreateExportResponse response = await _mediator.ExportAsync(_fhirRequestContextAccessor.FhirRequestContext.Uri, null, since, HttpContext.RequestAborted);
 
             var exportResult = ExportResult.Accepted();
             exportResult.SetContentLocationHeader(_urlResolver, OperationsConstants.Export, response.JobId);
@@ -94,7 +94,7 @@ namespace Microsoft.Health.Fhir.Api.Controllers
         [Route(KnownRoutes.ExportResourceType)]
         [ServiceFilter(typeof(ValidateExportRequestFilterAttribute))]
         [AuditEventType(AuditEventSubType.Export)]
-        public IActionResult ExportResourceType(string typeParameter)
+        public async Task<IActionResult> ExportResourceType([FromQuery(Name = KnownQueryParameterNames.Since)] PartialDateTime since, string typeParameter)
         {
             // Export by ResourceType is supported only for Patient resource type.
             if (!string.Equals(typeParameter, ResourceType.Patient.ToString(), StringComparison.Ordinal))
@@ -102,7 +102,17 @@ namespace Microsoft.Health.Fhir.Api.Controllers
                 throw new RequestNotValidException(string.Format(Resources.UnsupportedResourceType, typeParameter));
             }
 
-            return CheckIfExportIsEnabledAndRespond();
+            if (!_exportConfig.Enabled)
+            {
+                throw new RequestNotValidException(string.Format(Resources.OperationNotEnabled, OperationsConstants.Export));
+            }
+
+            CreateExportResponse response = await _mediator.ExportAsync(_fhirRequestContextAccessor.FhirRequestContext.Uri, typeParameter, since, HttpContext.RequestAborted);
+
+            var exportResult = ExportResult.Accepted();
+            exportResult.SetContentLocationHeader(_urlResolver, OperationsConstants.Export, response.JobId);
+
+            return exportResult;
         }
 
         [HttpGet]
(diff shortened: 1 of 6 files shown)


src/Microsoft.Health.Fhir.Shared.Api/Controllers/ExportController.cs
```diff
@@ -77,32 +77,25 @@ public class ExportController : Controller
...
         [HttpGet]
         [Route(KnownRoutes.ExportResourceType)]
         [ServiceFilter(typeof(ValidateExportRequestFilterAttribute))]
         [AuditEventType(AuditEventSubType.Export)]
-        public IActionResult ExportResourceType(string typeParameter)
+        public async Task<IActionResult> ExportResourceType([FromQuery(Name = KnownQueryParameterNames.Since)] PartialDateTime since, string typeParameter)
         {
+            CheckIfExportIsEnabled();
+
             // Export by ResourceType is supported only for Patient resource type.
```
deepak-ms, line 92: Curious, why other resource types are not supported?

deepak-ms: FHIR spec :)

LTA-Thinking: The FHIR spec only says Patient types should be supported.
https://hl7.org/fhir/uv/bulkdata/export/index.html#endpoint---all-patients

The spec doesn't say the other types shouldn't be supported, but it only lists Patient as an export operation.
If we remove this check we would support any resource type, so their is no technical lift to add that. It just a question of if we are still following the spec if we add that.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ExportController.cs
```diff
@@ -156,16 +149,31 @@ public async Task<IActionResult> CancelExport(string idParameter)
             return new ExportResult(response.StatusCode);
         }
 
+        private async Task<IActionResult> CreateExportJob(PartialDateTime since, string type = null)
```
namadabu, line 152: nit: `SendExportRequest` instead of `CreateExportJob`?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ExportTests.cs
```diff
@@ -30,7 +30,6 @@ public ExportTests(HttpIntegrationTestFixture fixture)
         }
 
         [Theory]
```
namadabu, line 32: nit: This can be change to `Fact`. And the test can also be renamed to be more specific (maybe something like WhenRequestingExportForResourceWithId)

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ExportTests.cs
```diff
@@ -41,10 +40,12 @@ public async Task GivenExportIsEnabled_WhenRequestingExportForResourceWithCorrec
             Assert.Equal(HttpStatusCode.MethodNotAllowed, response.StatusCode);
         }
 
-        [Fact]
-        public async Task GivenExportIsEnabled_WhenRequestingExportWithCorrectHeaders_ThenServerShouldReturnAcceptedAndNonEmptyContentLocationHeader()
+        [Theory]
+        [InlineData("$export")]
+        [InlineData("Patient/$export")]
```
namadabu, line 45: Would it be useful to add a test for the endpoint `Observation/$export` and validate that it returns an unsuccessful response?

namadabu:
We also have `ExportDataValidationTests` that validates the exported data also. It might be useful to add one for the `Patient` endpoint to make sure that we only export `Patient` resources.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ExportDataValidationTests.cs
```diff
@@ -58,12 +61,19 @@ public async Task GivenFhirServer_WhenDataIsExported_ThenExportedDataIsSameAsDat
...
         {
             bool result = true;
+
             if (dataFromStorageAccount.Count != dataFromServer.Count)
             {
                 _outputHelper.WriteLine($"Count differs. Exported data count: {dataFromStorageAccount.Count} Fhir Server Count: {dataFromServer.Count}");
                 return false;
             }
 
+            if (dataFromStorageAccount.Count == 0)
```
namadabu, line 71: I don't think this check is necessary. If someone runs this test (as part of running all E2E tests, say, before sending out a PR) on their localbox with an empty fhir-server, then the test will fail and might confuse them. If `dataFromStorageAccount.Count == 0` and `dataFromServer.Count == 0` the test should still pass.

src/Microsoft.Health.Fhir.Shared.Client/FhirClient.cs
```diff
@@ -298,9 +298,9 @@ public async Task<FhirResponse<Bundle>> SearchPostAsync(string resourceType, par
             return await CreateResponseAsync<Bundle>(response);
         }
 
-        public async Task<Uri> ExportAsync()
+        public async Task<Uri> ExportAsync(string path = "$export")
```
brendankowitz, line 301: To constrain this more to the export operation, would something like this work?
```csharp
public async Task<Uri> ExportAsync(string resource = "") {
// Validate Is known resource?
var message = new HttpRequestMessage(HttpMethod.Get, $"{resource}$export");
```

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ExportDataValidationTests.cs
LTA-Thinking: I commented this check out, but I'm going to leave it in as I found it useful for me while editing the tests to make sure I didn't mess something up and no data was exported.

Issue #1061, opened 2020-06-05
When searching by _id and a code search parameters Invalid Cast Exception is thrown


anyname2:
**Describe the bug**
`InvalidCastException` is thrown when searching using an `_id` search parameter combined with a search parameter of type code (e.g. `intent`).

**To Reproduce**
GET `/DeviceRequest?intent=order&_id=123`

**Expected behavior**
DeviceRequest with id=123 and intent=order to be returned if any such resources exist.

**Actual behavior**
Server returned 500 Internal Server Error with InvalidCastException.
```
System.InvalidCastException: Unable to cast object of type 'Microsoft.Health.Fhir.SqlServer.Features.Search.Expressions.Visitors.QueryGenerators.DispatchingDenormalizedSearchParameterQueryGenerator' to type 'Microsoft.Health.Fhir.SqlServer.Features.Search.Expressions.Visitors.ISqlExpressionVisitor`2[Microsoft.Health.Fhir.SqlServer.Features.Search.Expressions.Visitors.QueryGenerators.SearchParameterQueryGeneratorContext,Microsoft.Health.Fhir.SqlServer.Features.Search.Expressions.Visitors.QueryGenerators.SearchParameterQueryGeneratorContext]'.
   at Microsoft.Health.Fhir.SqlServer.Features.Search.Expressions.SqlRootExpression.AcceptVisitor[TContext,TOutput](IExpressionVisitor`2 visitor, TContext context) in C:\Users\valja\source\fhir-server\src\Microsoft.Health.Fhir.SqlServer\Features\Search\Expressions\SqlRootExpression.cs:line 59
   at Microsoft.Health.Fhir.SqlServer.Features.Search.Expressions.Visitors.QueryGenerators.SqlQueryGenerator.VisitSqlRoot(SqlRootExpression expression, SearchOptions context) in C:\Users\valja\source\fhir-server\src\Microsoft.Health.Fhir.SqlServer\Features\Search\Expressions\Visitors\QueryGenerators\SqlQueryGenerator.cs:line 113
   at Microsoft.Health.Fhir.SqlServer.Features.Search.Expressions.SqlRootExpression.AcceptVisitor[TContext,TOutput](ISqlExpressionVisitor`2 visitor, TContext context) in C:\Users\valja\source\fhir-server\src\Microsoft.Health.Fhir.SqlServer\Features\Search\Expressions\SqlRootExpression.cs:line 65
[...]

PR #1062, opened 2020-06-05
Fixed DenormalizedPredicateRewriter to set denormalized expression correctly

anyname2:
## Description
`DenormalizedPredicateRewriter` was adding the original `SqlRootExpression` to the denormalized list for each expression not extracted into a `TableExpression`, which is incorrect and was causing issue descirbed in #1061. After the fix, denormalized expressions are added to the denormalized list themselves, instead of the root. 

## Related issues
Addresses #1061

## Testing
Added unit tests for DenormalizedPredicateRewriter.
Manually executed the query from #1061 to verify the issue is fixed.


johnstairs:
This is great! Thanks a lot, @anyname2!

PR #1063, opened 2020-06-05
Detect change in export location

LTA-Thinking:
## Description
Detects if the export location has changed since the export job was made. If the job is still able to send data to the old location it will continue, otherwise the job will fail.

## Related issues
[User Story 72598: Snapshot export job record data](https://microsofthealth.visualstudio.com/Health/_workitems/edit/72598)

## Testing

### DIFF
--- a/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Export/ExportJobTaskTests.cs
+++ b/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Export/ExportJobTaskTests.cs
@@ -508,6 +508,37 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Operations.Export
         }
 
         [Fact]
+        public async Task GivenStorageAccountConnectionChanged_WhenExecuted_ThenJobStatusShouldBeUpdatedToFailed()
+        {
+            // Setup export destination client.
+            string connectionFailure = "Storage account connection string was updated during an export job.";
+
+            var exportJobRecordWithChangedConnection = new ExportJobRecord(
+                new Uri("https://localhost/ExportJob/"),
+                "Patient",
+                "hash",
+                since: PartialDateTime.MinValue,
+                storageAccountConnectionHash: ExportJobConfiguration.HashStorageAccountConnection("different connection"),
+                storageAccountUri: _exportJobConfiguration.StorageAccountUri);
+            SetupExportJobRecordAndOperationDataStore(exportJobRecordWithChangedConnection);
+
+            var exportJobTask = new ExportJobTask(
+                () => _fhirOperationDataStore.CreateMockScope(),
+                Options.Create(_exportJobConfiguration),
+                () => _searchService.CreateMockScope(),
+                _resourceToByteArraySerializer,
+                _inMemoryDestinationClient,
+                NullLogger<ExportJobTask>.Instance);
+
+            await exportJobTask.ExecuteAsync(_exportJobRecord, _weakETag, _cancellationToken);
+
+            Assert.NotNull(_lastExportJobOutcome);
+            Assert.Equal(OperationStatus.Failed, _lastExportJobOutcome.JobRecord.Status);
+            Assert.Equal(connectionFailure, _lastExportJobOutcome.JobRecord.FailureDetails.FailureReason);
+            Assert.Equal(HttpStatusCode.BadRequest, _lastExportJobOutcome.JobRecord.FailureDetails.FailureStatusCode);
+        }
+
+        [Fact]
         public async Task GivenAnExportJobToResume_WhenExecuted_ThenItShouldExportAllRecordsAsExpected()
         {
             // We are using the SearchService to throw an exception in order to simulate the export job task
--- a/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Export/ExportJobTaskTests.cs
+++ b/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Export/ExportJobTaskTests.cs
@@ -508,6 +508,48 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Operations.Export
         }
 
         [Fact]
+        public async Task GivenStorageAccountConnectionDidNotChange_WhenExecuted_ThenJobShouldBeCompleted()
+        {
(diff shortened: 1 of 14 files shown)


src/Microsoft.Health.Fhir.Core/Features/Operations/Export/CreateExportRequestHandler.cs
```diff
@@ -58,13 +67,15 @@ public async Task<CreateExportResponse> Handle(CreateExportRequest request, Canc
 
             string hash = JsonConvert.SerializeObject(hashObject).ComputeHash();
 
+            string storageAccountConnectionHash = Microsoft.Health.Core.Extensions.StringExtensions.ComputeHash(_exportJobConfiguration.StorageAccountConnection);
```
namadabu, line 70: Can we avoid computing the hash and set `storageAccountConnectionHash` to an empty string when `_exportJobConfiguration.StorageAccountConnection` is empty?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -72,8 +72,29 @@ public async Task ExecuteAsync(ExportJobRecord exportJobRecord, WeakETag weakETa
...
+                        throw new DestinationConnectionException("Storage account connection string was updated during an export job.", HttpStatusCode.BadRequest);
+                    }
+                }
+                else
+                {
+                    if (!string.Equals(exportJobRecord.StorageAccountUri, _exportJobConfiguration.StorageAccountUri, StringComparison.Ordinal))
+                    {
+                        exportJobConfiguration = new ExportJobConfiguration();
+                        exportJobConfiguration.Enabled = _exportJobConfiguration.Enabled;
+                        exportJobConfiguration.StorageAccountUri = exportJobRecord.StorageAccountUri;
```
namadabu, line 92: Any reason we don't want to update `_exportJobConfiguration.StorageAccountUri` to the correct value and pass `_exportJobConfiguration` to `ConnectAsync()`?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -72,8 +72,29 @@ public async Task ExecuteAsync(ExportJobRecord exportJobRecord, WeakETag weakETa
 
             try
             {
+                ExportJobConfiguration exportJobConfiguration = _exportJobConfiguration;
+
+                string connectionHash = Microsoft.Health.Core.Extensions.StringExtensions.ComputeHash(_exportJobConfiguration.StorageAccountConnection);
+
+                if (string.IsNullOrEmpty(exportJobRecord.StorageAccountUri) && string.IsNullOrEmpty(_exportJobConfiguration.StorageAccountUri))
+                {
+                    if (!string.Equals(exportJobRecord.StorageAccountConnectionHash, connectionHash, StringComparison.Ordinal))
```
namadabu, line 81: In this scenario we don't attempt to connect to the old "StorageAccountConnection" whereas in the below scenario we attempt to connect to the old "StorageAccountUri". It might be better to have the same behavior for both cases.

LTA-Thinking: Because we are only passing a hash of the connection string we have no way of connecting to the old storage account if the job used the connection string. I can change the behavior to be the same with the uri, but we have the ability to try to finish the job with the uri.

namadabu: Aah. Agreed.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
LTA-Thinking: I was concerned that if we changed the default than it could be used in a different job, but this always checks the uri value in the job record so it wouldn't matter. I'll change it to update `_exportJobConfiguration`

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -72,8 +72,29 @@ public async Task ExecuteAsync(ExportJobRecord exportJobRecord, WeakETag weakETa
...
+                    {
+                        throw new DestinationConnectionException("Storage account connection string was updated during an export job.", HttpStatusCode.BadRequest);
+                    }
+                }
+                else
+                {
+                    if (!string.Equals(exportJobRecord.StorageAccountUri, _exportJobConfiguration.StorageAccountUri, StringComparison.Ordinal))
```
namadabu, line 88: nit: We could potentially simplify this code since we will always use the `exportJobRecord.StorageAccountUri` value.

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Export/ExportJobTaskTests.cs
```diff
@@ -502,6 +507,75 @@ public async Task GivenConnectingToDestinationFails_WhenExecuted_ThenJobStatusSh
             Assert.Equal(HttpStatusCode.BadRequest, _lastExportJobOutcome.JobRecord.FailureDetails.FailureStatusCode);
         }
 
+        [Fact]
+        public async Task GivenStorageAccountConnectionChanged_WhenExecuted_ThenJobStatusShouldBeUpdatedToFailed()
```
namadabu, line 511: nit: Might be good to add one more test which validates that given `storageAccountConnection` has a value and remains same, then export succeeds as expected.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
LTA-Thinking: If the uri for the _exportJobConfiguration is changed here will it update the value elsewhere in the program, namely in the CreateExportRequestHandler? Looking at the documentation for IOption it is injected as a singleton, so both objects will share the same IOption for the export job configuration. I can't find anything saying if the value of an IOption is passed by reference or by value. I would assume by reference which would imply to me that updating it here would change it in CreateExportRequestHandler.

Issue #1065, opened 2020-06-08
Allow throttling based on number of concurrent requests


YazanMSFT:

brendankowitz:
Closing as this is merged

PR #1066, opened 2020-06-08
Add the ability to limit the number of concurrent requests

YazanMSFT:
## Description
Add the ability to limit the number of concurrent requests handled by the server.

## Related issues
Addresses [issue #1065 ].

## Testing
Added unit tests.

### DIFF
--- a/src/Microsoft.Health.Fhir.Api/Features/Throttling/ThrottlingMiddleware.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/Throttling/ThrottlingMiddleware.cs
@@ -4,7 +4,7 @@
 // -------------------------------------------------------------------------------------------------
 
 using System;
-using System.Linq;
+using System.Collections.Generic;
 using System.Net;
 using System.Threading;
 using System.Threading.Tasks;
@@ -23,6 +23,7 @@ namespace Microsoft.Health.Fhir.Api.Features.Throttling
     {
         private readonly ILogger<ThrottlingMiddleware> _logger;
         private readonly ThrottlingConfiguration _configuration;
+        private readonly HashSet<(string method, string path)> _excludedEndpoints = new HashSet<(string method, string path)>();
         private int _requestsInFlight = 0;
         private RequestDelegate _next;
 
@@ -31,11 +32,27 @@ namespace Microsoft.Health.Fhir.Api.Features.Throttling
             _next = EnsureArg.IsNotNull(next, nameof(next));
             _logger = EnsureArg.IsNotNull(logger, nameof(logger));
             _configuration = EnsureArg.IsNotNull(configuration?.Value, nameof(configuration));
+            _excludedEndpoints = new HashSet<(string method, string path)>(new StringTupleOrdinalIgnoreCaseEqualityComparer());
+
+            if (_configuration?.ExcludedEndpoints != null)
+            {
+                foreach (var excludedEndpoint in _configuration.ExcludedEndpoints)
+                {
+                    var mapping = excludedEndpoint.Split(separator: ':', count: 2);
+
+                    if (mapping.Length != 2)
+                    {
+                        throw new ArgumentException($"Entry '{excludedEndpoint}' is invalid. It should be of the format [request method]:[request path].", nameof(configuration));
+                    }
+
+                    _excludedEndpoints.Add((mapping[0], mapping[1]));
+                }
+            }
         }
 
         public async Task Invoke(HttpContext context)
         {
-            if (_configuration.ExcludedEndpoints.Contains($"{context.Request.Method}:{context.Request.Path.Value?.TrimEnd('/')}", StringComparer.OrdinalIgnoreCase))
+            if (_excludedEndpoints.Contains((context.Request.Method, context.Request.Path.Value)))
             {
                 // Endpoint is exempt from concurrent request limits.
                 await _next(context);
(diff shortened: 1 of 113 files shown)


src/Microsoft.Health.Fhir.Api/Features/Throttling/ThrottlingMiddleware.cs
```diff
@@ -0,0 +1,69 @@
...
+                }
+                else
+                {
+                    // Exceeded the concurrent request limit, return 429.
+                    context.Response.StatusCode = (int)HttpStatusCode.TooManyRequests;
+                    _logger.LogWarning($"{Resources.TooManyConcurrentRequests}. Limit is {_configuration.ConcurrentRequestLimit}.");
+
+                    // Output an OperationOutcome in the body.
+                    var result = TooManyRequestsActionResult.TooManyRequests;
+                    await result.ExecuteResultAsync(new ActionContext { HttpContext = context, RouteData = context.GetRouteData() ?? new RouteData() });
```
johnstairs, line 60: I feel like this is sort of mixing in MVC stuff before MVC has even been reached, but without the full behavior we would get if this were executed later in the pipeline. For instance, we return this body:
```
{
    "severity": "Error",
    "code": "Throttled",
    "diagnostics": "The maximum number of concurrent API calls was reached.",
    "location": null
}
```

(And we ignore if the request requested XML)

Whereas we return this for a 404:
```
{
    "resourceType": "OperationOutcome",
    "id": "d4803fe1-6bd9-44ea-a635-310ddb3ea92a",
    "issue": [
        {
            "severity": "error",
            "code": "not-found",
            "diagnostics": "Resource type 'Patient' with id 'sss' couldn't be found."
        }
    ]
}
```

I think it is completely fine in this case, maybe I'm making a general comment about our codebase design and the reliance on MVC.
Also, given this, I wonder if we should just write a cached  binary serialization to the response stream. I don't know what the ovehead of the `ExecuteResultAsync` call with its argument payload is.

src/Microsoft.Health.Fhir.Api/Features/Throttling/ThrottlingMiddleware.cs
```diff
@@ -0,0 +1,69 @@
...
+        public ThrottlingMiddleware(RequestDelegate next, IOptions<ThrottlingConfiguration> configuration, ILogger<ThrottlingMiddleware> logger)
+        {
+            _next = EnsureArg.IsNotNull(next, nameof(next));
+            _logger = EnsureArg.IsNotNull(logger, nameof(logger));
+            _configuration = EnsureArg.IsNotNull(configuration?.Value, nameof(configuration));
+        }
+
+        public async Task Invoke(HttpContext context)
+        {
+            if (_configuration.ExcludedEndpoints.Contains($"{context.Request.Method}:{context.Request.Path.Value?.TrimEnd('/')}", StringComparer.OrdinalIgnoreCase))
```
johnstairs, line 38: This implementation is falling back to ExcludedEndpoints as an iEnumberable, because you are passing in the comparer. The comparer should be set in the constructor instead.

johnstairs: You could also likely achieve a more efficient implementation using a ValueTuple instead of allocating a string.

brandonpollett:
Not sure the use case for this, but we might want to call out either in the error message or the configuration that this is a per-instance limit. If you scale the service out to n instances you're going to have n*limit available requests if your distribution is perfect.

src/Microsoft.Health.Fhir.Shared.Web/appsettings.json
```diff
@@ -64,6 +64,13 @@
         },
         "Bundle": {
             "EntryLimit": 500
+        },
+        "Throttling": {
+            "Enabled": true,
+            "ConcurrentRequestLimit": 100,
```
brendankowitz, line 70: This seems highly dependency on the environment, should it just be off be default?

deepak-ms:
What's the use case of this feature? And how are we going to decide value for  ConcurrentRequestLimit config?

hansenms:
This looks good overall. My only (nit picky) comment would be about naming. This is actually not about throttling (*reduce* throughput) overall. The goal is to limit the number of concurrent requests to *increase* overall throughput by avoiding pathological behavior when the service gets overwhelmed. Throttling (as in max number of requests per some period of time) is maybe a bit different.

src/Microsoft.Health.Fhir.Api/Features/Throttling/ThrottlingMiddleware.cs
```diff
@@ -65,5 +82,18 @@ public async Task Invoke(HttpContext context)
...
+        private class StringTupleOrdinalIgnoreCaseEqualityComparer : IEqualityComparer<ValueTuple<string, string>>
+        {
+            public bool Equals(ValueTuple<string, string> x, ValueTuple<string, string> y)
+            {
+                return StringComparer.OrdinalIgnoreCase.Equals(x.Item1, y.Item1) && StringComparer.OrdinalIgnoreCase.Equals(x.Item2, y.Item2);
+            }
+
+            public int GetHashCode(ValueTuple<string, string> obj)
+            {
+                return StringComparer.OrdinalIgnoreCase.GetHashCode(obj.Item1) ^ StringComparer.OrdinalIgnoreCase.GetHashCode(obj.Item2);
```
johnstairs, line 95: You can use `HashCode.Combine()`.

src/Microsoft.Health.Fhir.Api/Features/Throttling/ThrottlingMiddleware.cs
```diff
@@ -0,0 +1,99 @@
...
+using Microsoft.Health.Fhir.Api.Configs;
+using Microsoft.Health.Fhir.Api.Features.ActionResults;
+
+namespace Microsoft.Health.Fhir.Api.Features.Throttling
+{
+    public class ThrottlingMiddleware
+    {
+        private readonly ILogger<ThrottlingMiddleware> _logger;
+        private readonly ThrottlingConfiguration _configuration;
+        private readonly HashSet<(string method, string path)> _excludedEndpoints = new HashSet<(string method, string path)>();
```
johnstairs, line 26: Unnecessary initialization?

src/Microsoft.Health.Fhir.Shared.Web/appsettings.json
```diff
@@ -64,6 +64,13 @@
         },
         "Bundle": {
             "EntryLimit": 500
+        },
+        "Throttling": {
+            "Enabled": false,
+            "ConcurrentRequestLimit": 100,
+            "ExcludedEndpoints": [
+                "get:/health/check"
```
jackliums, line 72: i think we can probably exclude metadata endpoint too?

src/Microsoft.Health.Fhir.Shared.Web/appsettings.json
```diff
@@ -4,8 +4,8 @@
             "Enabled": true,
             "EnableAadSmartOnFhirProxy": true,
             "Authentication": {
-                "Audience": null,
-                "Authority": null
+                "Audience": "fhir-api",
+                "Authority": "https://localhost:44348"
```
johnstairs, line 8: Why are you making these changes?

PR #1069, opened 2020-06-09
Sort by _lastUpdated in Sql server

rbans96:
## Description
Adds Support for sorting(ASC and DESC) by _lastUpdated in SqlServer

## Related issues
Addresses #68732

## Testing
Adds e2e tests


src/Microsoft.Health.Fhir.SqlServer/Features/Search/SortUtils.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        public static SortOrder GetSortOrderByParameterName(SearchOptions searchOptions, string parameter)
+        {
+            var sortOrder = SortOrder.Ascending;
+            if (searchOptions.Sort?.Count > 0)
+            {
+                sortOrder = searchOptions
+                    .Sort
+                    .Where(x => string.Equals(x.searchParameterInfo.Name, parameter, StringComparison.OrdinalIgnoreCase))
+                    .Select(s => s.sortOrder).First();
```
johnstairs, line 22: This will throw if there are no matches

johnstairs: What if they try to order by another parameter?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/SearchTestsBase.cs
```diff
@@ -27,33 +30,83 @@ protected SearchTestsBase(TFixture fixture)
...
+            ValidateBundle(bundle, selfLink, true, expectedResources);
 
             return bundle;
         }
 
+        protected async Task<Bundle> ExecuteAndValidateBundle(string searchUrl, string selfLink, bool sort, params Resource[] expectedResources)
+        {
+            FhirResponse<Bundle> firstBundle = await Client.SearchAsync(searchUrl);
+
+            var firstExpectedResources = expectedResources.Length > 10 ? expectedResources.ToList().GetRange(0, 10).ToArray() : expectedResources;
```
brendankowitz, line 54: Can we have `10` as a variable that is also used to set the `_count` in search, just so this paging seems more obvious

brendankowitz: Alternatively, maybe pull this out into a pageSize field

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SortUtils.cs
```diff
@@ -0,0 +1,43 @@
...
+namespace Microsoft.Health.Fhir.SqlServer.Features.Search
+{
+    public static class SortUtils
+    {
+        private static readonly IList<string> SupportedParameters = new List<string>
+            {
+                KnownQueryParameterNames.LastUpdated,
+            };
+
+        public static SortOrder GetSortOrderForSupportedParam(SearchOptions searchOptions)
```
brendankowitz, line 19: I think we're good on functionality in this PR, just a few more thoughts on this class.
 - Add some documentation for what this function is doing
 - Probably a good candidate for an extension method, also rename so its more clear what its doing `public static SortOrder GetFirstSortOrderForSupportedParam(this SearchOptions searchOptions)`

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SortUtils.cs
```diff
@@ -0,0 +1,43 @@
...
+                else
+                {
+                    throw new SearchParameterNotSupportedException(string.Format(Core.Resources.SearchParameterNotSupported, sortOptions.searchParameterInfo.Name));
+                }
+            }
+
+            return sortOrder;
+        }
+
+        public static bool IsSearchParamterSupported(string parameter)
```
brendankowitz, line 38: Could also be an extension on SearchParameterInfo: `public static bool IsSearchParamterSupported(this SearchParameterInfo parameter)`

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SortUtils.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        public static SortOrder GetSortOrderForSupportedParam(SearchOptions searchOptions)
+        {
+            var sortOrder = SortOrder.Ascending;
+
+            foreach (var sortOptions in searchOptions.Sort)
+            {
+                if (IsSearchParamterSupported(sortOptions.searchParameterInfo.Name))
+                {
+                    sortOrder = sortOptions.sortOrder;
```
brendankowitz, line 27: If we indicate its only getting _first_ we can probably return here

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SortUtils.cs
rbans96: Good point. Fixed

PR #1072, opened 2020-06-09
Refactor SqlServerFhirModel initialization

rotodd:
## Description

- Updates `SqlServerFhirModel` to:
  - inherit from `IStartable` instead of `SqlServerModelInitialization`
  - start up the Schema Initializer and the Search Parameter Definition Manager
- Puts `SqlServerModelInitialization.GetStringId()` in `SqlServerFhirModel` (`SqlServerModelInitialization` will be removed in [#AB74343](https://microsofthealth.visualstudio.com/Health/_workitems/edit/74343))
- Makes`SearchParameterDefinitionManager.Start()` idempotent

## Related issues
Addresses [#AB74281](https://microsofthealth.visualstudio.com/Health/_workitems/edit/74281).

## Testing
- All tests pass as before

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Definition/ISearchParameterDefinitionManager.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Definition/ISearchParameterDefinitionManager.cs
@@ -64,11 +64,5 @@ namespace Microsoft.Health.Fhir.Core.Features.Definition
         /// <param name="componentIndex">The optional component index if the search parameter is a composite</param>
         /// <returns>The search parameter type.</returns>
         SearchParamType GetSearchParameterType(SearchParameterInfo searchParameter, int? componentIndex);
-
-        /// <summary>
-        /// Initializes the definition manager, reading in search parameter information from a JSON file.
-        /// </summary>
-        /// <remarks>This method is idempotent.</remarks>
-        void Start();
     }
 }
--- a/src/Microsoft.Health.Fhir.Core/Features/Definition/ISearchParameterDefinitionManager.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Definition/ISearchParameterDefinitionManager.cs
@@ -14,7 +14,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Definition
     /// <summary>
     /// Provides mechanism to access search parameter definition.
     /// </summary>
-    public interface ISearchParameterDefinitionManager : IStartable
+    public interface ISearchParameterDefinitionManager
     {
         public delegate ISearchParameterDefinitionManager SearchableSearchParameterDefinitionManagerResolver();
 
@@ -64,5 +64,11 @@ namespace Microsoft.Health.Fhir.Core.Features.Definition
         /// <param name="componentIndex">The optional component index if the search parameter is a composite</param>
         /// <returns>The search parameter type.</returns>
         SearchParamType GetSearchParameterType(SearchParameterInfo searchParameter, int? componentIndex);
+
+        /// <summary>
+        /// Initializes the definition manager, reading in search parameter information from a JSON file.
+        /// </summary>
+        /// <remarks>This method is idempotent.</remarks>
+        void Start();
     }
 }
--- a/src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterDefinitionManager.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterDefinitionManager.cs
@@ -7,7 +7,6 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using EnsureThat;
-using Hl7.Fhir.Serialization;
 using Microsoft.Health.Extensions.DependencyInjection;
 using Microsoft.Health.Fhir.Core.Features.Search;
 using Microsoft.Health.Fhir.Core.Models;
--- a/src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterDefinitionManager.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterDefinitionManager.cs
@@ -22,6 +22,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Definition
         private readonly IModelInfoProvider _modelInfoProvider;
 
         private IDictionary<string, IDictionary<string, SearchParameterInfo>> _typeLookup;
+        private bool _started;
 
         public SearchParameterDefinitionManager(IModelInfoProvider modelInfoProvider)
         {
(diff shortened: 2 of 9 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
```diff
@@ -47,94 +49,94 @@ public sealed class SqlServerFhirModel : SqlServerModelInitializer
...
         }
 
-        protected override async Task Initialize()
+        public void Start()
         {
-            var connectionStringBuilder = new SqlConnectionStringBuilder(SqlServerDataStoreConfiguration.ConnectionString);
+            StartAsync().GetAwaiter().GetResult();
+        }
+
+        // TODO: Add this to IStartable interface.
```
rotodd, line 127: I created [#AB74342](https://microsofthealth.visualstudio.com/Health/_workitems/edit/74342) to address this.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
```diff
@@ -47,94 +49,94 @@ public sealed class SqlServerFhirModel : SqlServerModelInitializer
...
         {
-            ThrowIfNotInitialized();
             return _quantityCodeToId.TryGetValue(code, out quantityCodeId);
         }
 
-        protected override async Task Initialize()
+        public void Start()
         {
-            var connectionStringBuilder = new SqlConnectionStringBuilder(SqlServerDataStoreConfiguration.ConnectionString);
+            StartAsync().GetAwaiter().GetResult();
```
rotodd, line 124: Pulled this from [here](https://docs.microsoft.com/en-us/archive/msdn-magazine/2015/july/async-programming-brownfield-async-development).

src/Microsoft.Health.Fhir.Core/Features/Definition/ISearchParameterDefinitionManager.cs
```diff
@@ -13,7 +14,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Definition
     /// <summary>
     /// Provides mechanism to access search parameter definition.
     /// </summary>
-    public interface ISearchParameterDefinitionManager
+    public interface ISearchParameterDefinitionManager : IStartable
```
brendankowitz, line 17: I don't think the interface ISearchParameterDefinitionManager should build on IStartable, these seem like unrelated concepts

src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterDefinitionManager.cs
```diff
@@ -36,16 +36,21 @@ public SearchParameterDefinitionManager(IModelInfoProvider modelInfoProvider)
...
+
+                builder.Build();
 
-            builder.Build();
+                _typeLookup = builder.ResourceTypeDictionary;
+                UrlLookup = builder.UriDictionary;
 
-            _typeLookup = builder.ResourceTypeDictionary;
-            UrlLookup = builder.UriDictionary;
+                List<string> list = UrlLookup.Values.Where(p => p.Type == ValueSets.SearchParamType.Composite).Select(p => string.Join("|", p.Component.Select(c => UrlLookup[c.DefinitionUrl].Type))).Distinct().ToList();
```
brendankowitz, line 50: @jackliums do you remember why we have this list that isn't used? I keep looking at it and wondering

src/Microsoft.Health.Fhir.Core/Features/Definition/ISearchParameterDefinitionManager.cs
rotodd: Let me know what you think of 138de6b. I removed `IStartable` from the interface but exposed the `Start()` method so that it could be called from `SqlServerFhirModel`.

rotodd: As discussed, I removed `Start()` from `ISearchParameterDefinitionManager` in b2f049a.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -77,7 +77,7 @@ internal class SqlServerFhirDataStore : IFhirDataStore, IProvideCapability
 
         public async Task<UpsertOutcome> UpsertAsync(ResourceWrapper resource, WeakETag weakETag, bool allowCreate, bool keepHistory, CancellationToken cancellationToken)
         {
-            await _model.EnsureInitialized();
+            await _model.StartAsync();
```
johnstairs, line 80: Why should we have to call StartAsync() from the consumer?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
```diff
@@ -47,94 +49,98 @@ public sealed class SqlServerFhirModel : SqlServerModelInitializer
...
         }
 
         public bool TryGetResourceTypeId(string resourceTypeName, out short id)
         {
-            ThrowIfNotInitialized();
             return _resourceTypeToId.TryGetValue(resourceTypeName, out id);
         }
 
         public string GetResourceTypeName(short resourceTypeId)
         {
```
johnstairs, line 81: ThrowIfNotInitialized() will fail in a predicatable way with a better error message than something like a null ref

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
```diff
@@ -189,7 +195,7 @@ COMMIT TRANSACTION
                     sqlCommand.Parameters.AddWithValue("@claimTypes", commaSeparatedClaimTypes);
                     sqlCommand.Parameters.AddWithValue("@compartmentTypes", commaSeparatedCompartmentTypes);
 
-                    using (SqlDataReader reader = await sqlCommand.ExecuteReaderAsync(CommandBehavior.SequentialAccess))
+                    using (SqlDataReader reader = await sqlCommand.ExecuteReaderAsync(CommandBehavior.SequentialAccess).ConfigureAwait(false))
```
johnstairs, line 198: in .net core, this is not needed

rotodd: Did a bit a of reading about this... is this the reason why?

"_code running in an ASP.NET Core app by default won’t see a custom SynchronizationContext, which lessens the need for ConfigureAwait(false) running in such an environment._" (taken from [here](https://devblogs.microsoft.com/dotnet/configureawait-faq/))

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
rotodd: I was under the impression that these methods would never be called before `Start()` and `StartAsync()`, so there wouldn't be a need to check initialization. Please correct me if I'm wrong!

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
johnstairs: Yeah. What I said was technically not true, but we're not publishing these as general-purpose libraries, and we don't call ConfigureAwait(false) elsewhere, because we don't create a synchonization context and neither does asp.net core. 
You could certainly make the case that we should. But then we should do it everywhere and turn on an analyzer rule to enforce it. I'm not sure the added noise would be worth it.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
johnstairs: Start _should_ be called first. This is just contract enforcement. The class cannot be used until Started, which makes it special, and it will tell you if try to use it before Started (i.e. in a unit test). (Or if ever we did publish this a general-purpose library, though I'm contradicting my comment on ConfigureAwait :)) Not a big deal either way.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
rotodd: Okay, makes sense. Thanks for the background!

rotodd: Removed in 1079d38.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
rotodd: Good point. Removed in 4e00f98 and c1c9e77.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
rotodd: You're right - I found this useful when debugging unit tests. Added the check back in in c583087.

PR #1073, opened 2020-06-10
Adds tests for SearchIndexer

brendankowitz:
## Description
SearchIndexer does not have any tests on it directly, this adds tests to verify that indexes have been extracted correctly

## Testing
Adds unit tests for SearchIndexer


src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/SearchIndexerTests.cs
```diff
@@ -0,0 +1,67 @@
...
+                () => _fixtureData.SearchDefinitionManager,
+                SearchParameterFixtureData.Manager,
+                new LightweightReferenceToElementResolver(new ReferenceSearchValueParser(new FhirRequestContextAccessor()), ModelInfoProvider.Instance),
+                NullLogger<SearchIndexer>.Instance);
+        }
+
+        [Theory]
+        [InlineData("DocumentReference-example")]
+        [InlineData("DocumentReference-example-002")]
+        [InlineData("DocumentReference-example-003")]
```
rotodd, line 51: Nit: it might make sense to indicate the different things that are getting tested in the test file names.

I don't know much about this, but, looking at the files, maybe something like this?
`DocumentReference-example-relatesTo-code-appends`
`DocumentReference-example-relatesTo-code-transforms-add-target`
`DocumentReference-example-relatesTo-code-transforms`

brendankowitz: Probably a good idea, these are the files used in the composite search tests, which I find the hardest to debug

Issue #1077, opened 2020-06-11
Container based SQL server for helm chart


hansenms:
**User story**
As a user I want to deploy the FHIR server in Kubernetes using SQL server in a container. This will enable deployment locally (minikube) or Azure Stack Hub.

**Acceptance criteria**
1. When deploying with helm chart and `--set database.dataStore=SqlContainer`, the FHIR server is deployed with SQL server running in a container.

PR #1083, opened 2020-06-14
SQL Container for k8s

hansenms:
## Description
Introduces a SQL in container option for Kubernetes deployments. This allows users to deploy on a local cluster.

## Related issues
Addresses #1077.

## Testing
Introduces E2E tests for SQL container.


LTA-Thinking:
Splitting the tests into cosmos and sql is a great idea. Once this is in I'll update the other jobs to match.

samples/kubernetes/helm/fhir-server/templates/deployment.yaml
```diff
@@ -79,7 +80,28 @@ spec:
...
+              valueFrom:
+                secretKeyRef:
+                  name: {{ $dbservicename }}
+                  key: sapassword
+            - name: DATABASENAME
+              value: {{ include "fhir-server.fullname" . }}
+            - name: DATABASESERVERNAME
+              value: {{ $dbservicename }}
+            - name: SqlServer__ConnectionString
+              value: "Server=tcp:$(DATABASESERVERNAME),1433;Initial Catalog=FHIR;Persist Security Info=False;User ID=sa;Password=$(DATABASEPASSWORD);MultipleActiveResultSets=False;Connection Timeout=30"
```
johnstairs, line 100: I think we should have `User Id` as a parameter. Also, there are other properties in the connection string above that we should probably include.

hansenms: Yes, the `User Id` is always `sa` for the official SQL container images, but I have made it configurable in case one might use a different image. Also allowed the database name to be changed for SQL container.

johnstairs: My thinking was that you should not generally run as a user with higher privileges than you need. We will eventually be using a tool for creating the schema and can create a lower-privileged user for the FHIR server to use.

PR #1086, opened 2020-06-15
Add support for Sub Searches to export

LTA-Thinking:
## Description
As part of Patient and Group exports additional resources need to be retrieved based on the primary resources being exported. This requires additional searches based on the results returned from the primary search. As each of these sub searches could have multiple pages of data, and in the case of Group export could have sub searches of their own, the continuation tokens and page counts for these searches need to be stored. This will allow the export job to resume if it is interrupted in the middle of a sub search without duplicating exported data. 
Within the ExportJobRecord the continuation token and current page of the search is stored in the ExportJobProgress object. This object just stores the continuation token and page. By adding a recursive reference to the ExportJobProgress and an Id for the resource that is currently being used for sub searches all the necessary data to resume an interrupted job would be present.

## Testing
No new functionality to test yet.

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Operations/Export/Models/ExportJobProgress.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Operations/Export/Models/ExportJobProgress.cs
@@ -10,10 +10,12 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Export.Models
 {
     public class ExportJobProgress
     {
-        public ExportJobProgress(string continuationToken, uint page)
+        public ExportJobProgress(string continuationToken, uint page, string resourceId = null, ExportJobProgress subSearch = null)
         {
             ContinuationToken = continuationToken;
             Page = page;
+            SubSearchResourceId = resourceId;
+            SubSearch = subSearch;
         }
 
         [JsonConstructor]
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Fhir.Core/Features/Operations/Export/Models/ExportJobProgress.cs
```diff
@@ -27,12 +29,30 @@ protected ExportJobProgress()
...
+
         public void UpdateContinuationToken(string continuationToken)
         {
             EnsureArg.IsNotNullOrWhiteSpace(continuationToken, nameof(continuationToken));
 
             ContinuationToken = continuationToken;
             Page++;
         }
+
+        public void NewSubSearch(string resourceId, ExportJobProgress subSearch)
```
deepak-ms, line 46: No callers?

LTA-Thinking: None yet. This story was meant to brainstorm ideas on how to cover the use case described above. The change was small enough that instead of a design doc I decided to just implement the change, and review it with this PR. I'll implement its use in the next story: User Story 74398: Add Compartment resources to Patient export

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/Models/ExportJobProgress.cs
```diff
@@ -27,12 +29,30 @@ protected ExportJobProgress()
         [JsonProperty(JobRecordProperties.Page)]
         public uint Page { get; private set; }
 
+        [JsonProperty(JobRecordProperties.SubSearchResourceId)]
+        public string SubSearchResourceId { get; private set; }
+
+        [JsonProperty(JobRecordProperties.SubSearch)]
+        public ExportJobProgress SubSearch { get; private set; }
```
brendankowitz, line 36: CompartmentSearch maybe?

LTA-Thinking: I went with sub search because this is also needed for Group export when searching for resources in nested Groups. I wanted a name that would be accurate for both scenarios.

LTA-Thinking:
Incorporated into the branch that consumed the changes being made.

PR #1087, opened 2020-06-16
Localization Update: Fix for GeoPol_ManifestFile into master branch

JoeDRowan:
## Description
GeoPol manifest file update - ,json files no longer exist at path.
@jackliums @YazanMSFT @brendankowitz - Files may have moved so may need to be re-added to exclude list in this manifest file with their correct paths

### DIFF
--- a/GeoPol.xml
+++ b/GeoPol.xml
@@ -17,7 +17,5 @@
     <ExcludeFolder>.gitignore</ExcludeFolder>
     <ExcludeFolder>GeoPol.xml</ExcludeFolder>
     <ExcludeFolder>THIRDPARTYNOTICES.md</ExcludeFolder>
-    <ExcludeFolder>src\Microsoft.Health.Fhir.Stu3.Core\Features\Definition\search-parameters.json</ExcludeFolder>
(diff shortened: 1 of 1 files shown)


GeoPol.xml
```diff
@@ -17,7 +17,5 @@
     <ExcludeFolder>.gitignore</ExcludeFolder>
     <ExcludeFolder>GeoPol.xml</ExcludeFolder>
     <ExcludeFolder>THIRDPARTYNOTICES.md</ExcludeFolder>
-    <ExcludeFolder>src\Microsoft.Health.Fhir.Stu3.Core\Features\Definition\search-parameters.json</ExcludeFolder>
-    <ExcludeFolder>src\Microsoft.Health.Fhir.R4.Core\Features\Definition\search-parameters.json</ExcludeFolder>
```
brendankowitz, line 21: These files moved. Actually the rule is probably easier now, can you exclude all json files in https://github.com/microsoft/fhir-server/tree/master/src/Microsoft.Health.Fhir.Core/Data?

JoeDRowan: Yes @brendankowitz - I'll update this pr with that exclusion.

JoeDRowan:
Awaiting re-review @brendankowitz @YazanMSFT

PR #1090, opened 2020-06-18
Update Reindex job task to query search parameters that are not yet indexed

feordin:
## Description
Add method to SearchParameterDefinitionManager to search for parameters that are in Supported state, but not enabled.

## Related issues
Addresses [issue [AB#73572](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/73572)].

## Testing
Added unit tests for SearchParameterDefinitionManager, manual end-to-end test showing reindex job with correct search parameters and resources.

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Definition/ISearchParameterDefinitionManager.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Definition/ISearchParameterDefinitionManager.cs
@@ -64,13 +64,5 @@ namespace Microsoft.Health.Fhir.Core.Features.Definition
         /// <param name="componentIndex">The optional component index if the search parameter is a composite</param>
         /// <returns>The search parameter type.</returns>
         SearchParamType GetSearchParameterType(SearchParameterInfo searchParameter, int? componentIndex);
-
-        /// <summary>
-        /// Gets list of search parameters with the given support status: <paramref name="isSupported"/> <paramref name="isSearchable"/>.
-        /// </summary>
-        /// <param name="isSupported">Indicates whether to return supported params or unsupported params.</param>
-        /// <param name="isSearchable">Indicates whether to return searchable or unsearchable params.</param>
-        /// <returns>An <see cref="IEnumerable{T}"/> that contains the matching search parameters.</returns>
-        IEnumerable<SearchParameterInfo> GetSearchByStatus(bool isSupported, bool isSearchable);
     }
 }
--- a/src/Microsoft.Health.Fhir.Core/Features/Definition/ISearchParameterDefinitionManager.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Definition/ISearchParameterDefinitionManager.cs
@@ -64,5 +64,13 @@ namespace Microsoft.Health.Fhir.Core.Features.Definition
         /// <param name="componentIndex">The optional component index if the search parameter is a composite</param>
         /// <returns>The search parameter type.</returns>
         SearchParamType GetSearchParameterType(SearchParameterInfo searchParameter, int? componentIndex);
+
+        /// <summary>
+        /// Gets list of search parameters with the given support status: <paramref name="isSupported"/> <paramref name="isSearchable"/>.
+        /// </summary>
+        /// <param name="isSupported">Indicates whether to return supported params or unsupported params.</param>
+        /// <param name="isSearchable">Indicates whether to return searchable or unsearchable params.</param>
+        /// <returns>An <see cref="IEnumerable{T}"/> that contains the matching search parameters.</returns>
+        IEnumerable<SearchParameterInfo> GetSearchByStatus(bool isSupported, bool isSearchable);
     }
 }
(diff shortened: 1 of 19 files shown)


src/Microsoft.Health.Fhir.Core/Features/Definition/ISearchParameterDefinitionManager.cs
```diff
@@ -63,5 +63,13 @@ public interface ISearchParameterDefinitionManager
...
         /// <returns>The search parameter type.</returns>
         SearchParamType GetSearchParameterType(SearchParameterInfo searchParameter, int? componentIndex);
+
+        /// <summary>
+        /// Gets list of search parameters with the given support status: <paramref name="isSupported"/> <paramref name="isSearchable"/>.
+        /// </summary>
+        /// <param name="isSupported">Indicates whether to return supported params or unsupported params.</param>
+        /// <param name="isSearchable">Indicates whether to return searchable or unsearchable params.</param>
+        /// <returns>An <see cref="IEnumerable{T}"/> that contains the matching search parameters.</returns>
+        IEnumerable<SearchParameterInfo> GetSearchByStatus(bool isSupported, bool isSearchable);
```
feordin, line 73: Perhaps this method is a bit awkward?  Looking at it again, I am thinking that perhaps we should split up the interface and include this only on the SupportedSearchParameterDefinitionManager?  Please leave your thoughts.  Thanks.

rotodd: Right, `SearchableSearchParameterDefinitionManager.AllSearchParameters` will return an `IEnumerable<SearchParameterInfo>` of all the search parameters that are marked as `isSearchable` and `SupportedSearchParameterDefinitionManager` will do the same for `isSupported`. It might be nice to piggy back off of the existing functionality or replace it. I just doubled checked, the filtering classes aren't used anywhere in the solution yet - but it might be good to check with @brendankowitz if you are thinking of removing them.

feordin: I updated the interface and added a method only to the SupportedSearchParameterDefinitionManager.  Thanks for the review.

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/SearchParameters/SearchParameterDefinitionManagerTests.cs
```diff
@@ -0,0 +1,109 @@
...
+using Microsoft.Health.Fhir.ValueSets;
+using NSubstitute;
+using Xunit;
+
+namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Search
+{
+    public class SearchParameterDefinitionManagerTests
+    {
+        private static readonly string ResourceId = "http://hl7.org/fhir/SearchParameter/Resource-id";
+        private static readonly string ResourceLastupdated = "http://hl7.org/fhir/SearchParameter/Resource-lastUpdated";
```
LTA-Thinking, line 22: Nit: Capitol U for Updated.

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/SearchParameters/SearchParameterDefinitionManagerTests.cs
```diff
@@ -0,0 +1,109 @@
...
+using Microsoft.Health.Fhir.Core.Features.Search.Parameters;
+using Microsoft.Health.Fhir.Core.Features.Search.Registry;
+using Microsoft.Health.Fhir.Core.Models;
+using Microsoft.Health.Fhir.ValueSets;
+using NSubstitute;
+using Xunit;
+
+namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Search
+{
+    public class SearchParameterDefinitionManagerTests
```
LTA-Thinking, line 19: Is the idea for this class to have tests for all the search parameter definition managers? Right now I think it makes more sense to name it for the SupportedSearchParameterDefinitionManager unless the other classes will have tests added to this file.

feordin: I think the file could be used for more SearchParameterDefinitionManager tests as it has the infrastructure in place.  I added one more test that was different to make it slight more broad.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -62,6 +69,22 @@ public async Task ExecuteAsync(ReindexJobRecord reindexJobRecord, WeakETag weakE
                 if (_reindexJobRecord.QueryList?.Count == 0)
                 {
                     // Build query based on new search params
+                    // Find supported, but not yet searchable params
+                    var notYetIndexedParams = _supportedSearchParameterDefinitionManager.GetSupportedButNotSearchableParams();
```
brendankowitz, line 73: I guess this is also the same as `_supportedSearchParameterDefinitionManager.AllSearchParameters.Where(p => !p.IsSearchable)`. All search params returned from _supportedSearchParameterDefinitionManager should be supported

feordin: It is the same.  Just hiding it behind an interface.  I suppose we could simply remove this method.

Issue #1071, opened 2020-06-09
Support for _list in search


brendankowitz:
**User story**
As a user, I want the _list search parameter to be supported so that I can filter queries by lists that I have defined.

**Acceptance criteria**
1. When _list is supplied as a search parameter, the results are filter by entities defined in the provided list

**Resources**
- http://hl7.org/fhir/search.html#list

brendankowitz:
One idea I can think of for this might be to create a pipeline behavior that intercepts [SearchResourceRequest](https://github.com/microsoft/fhir-server/blob/master/src/Microsoft.Health.Fhir.Core/Messages/Search/SearchResourceRequest.cs), fetches the list and rewrites the search parameters so that _list is replaced with an _id filter of the resources in the list. The rest of the search should work as designed.

Although not exactly the same, [ConditionalCreate](https://github.com/microsoft/fhir-server/blob/master/src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Create/ConditionalCreateResourceHandler.cs)/Update is a composite handler that executes a search before Creating/updating based on the results

brendankowitz:
A snippet for the suggestion above if someone wants to iterate on this...

```
/// <summary>
/// This behavior looks for searches that use _list then resolves the list and rewrites the query with the resource _ids
/// </summary>
public class ListSearchPreProcessor : IRequestPreProcessor<SearchResourceRequest>
{
    private const string _listParameter = "_list";
    private const string _idParameter = "_id";
    private readonly IScoped<IFhirDataStore> _dataStore;
    private readonly ResourceDeserializer _deserializer;
    private readonly IReferenceSearchValueParser _referenceSearchValueParser;

    public ListSearchBehavior(
        IScoped<IFhirDataStore> dataStore,
        ResourceDeserializer deserializer,
        IReferenceSearchValueParser referenceSearchValueParser)
    {
        EnsureArg.IsNotNull(dataStore, nameof(dataStore));
        EnsureArg.IsNotNull(deserializer, nameof(deserializer));
        EnsureArg.IsNotNull(referenceSearchValueParser, nameof(referenceSearchValueParser));

        _dataStore = dataStore;
        _deserializer = deserializer;
        _referenceSearchValueParser = referenceSearchValueParser;
    }

    public async Task Process(SearchResourceRequest request, CancellationToken cancellationToken)
    {
        EnsureArg.IsNotNull(request, nameof(request));

        Tuple<string, string> listParameter = request.Queries
[...]

PR #1091, opened 2020-06-18
Supporting _list (sql + cosmos)

eladiw:
## Description
Adding support for _list. implemented as preprocessor and by doing so solves the issue for sql and cosmos

## Related issues
Addresses [issue #1071 ].

## Testing
Tested manual + added unit tests that covers:
1. search without the _list option -> nothing is affected (checks some full list is returned)
2. search with _list='' , list is ignored
3. search with _list=<id> , when id is not an existing list, empty result set is returned
4. search with _list=<id> , when id exists, queries are updated to reflect the ids of the matched patients


src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Preprocessors/ListSearchPreProcessor.cs
```diff
@@ -0,0 +1,104 @@
...
+            if (references.Any())
+            {
+                query = query.Concat(new[] { Tuple.Create(_idParameter, string.Join(",", references.Select(x => x.ResourceId))) });
+            }
+            else
+            {
+                MakeAlwaysFalseQuery(ref query);
+            }
+
+            request.Queries = query.ToArray();
```
brendankowitz, line 92: After reading more about the specifics of List, I think we also should consider how we can implement List order (https://www.hl7.org/fhir/list.html#order)

eladiw: Shouldn't that be in a different PR, one that touches the 'List' functionality (e.g. base/List) and not base/resource?_list=xyz ?
As I understand it the order in which is irrelevant (anyway I keep the order that the list had)

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Behaviors/ListSearchPipeBehavior.cs
```diff
@@ -0,0 +1,131 @@
...
+
+namespace Microsoft.Health.Fhir.Core.Features.Search.Behavior
+{
+    /// <summary>
+    /// This behavior looks for searches that use _list then resolves the list and rewrites the query with the resource _ids
+    /// </summary>
+    public class ListSearchPipeBehavior : IPipelineBehavior<SearchResourceRequest, SearchResourceResponse>
+    {
+        private const string _listParameter = "_list";
+        private const string _idParameter = "_id";
```
tamirkamara, line 29: Should those be added to the KnownQueryParameterNames class?

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Behaviors/ListSearchPipeBehavior.cs
```diff
@@ -0,0 +1,131 @@
...
+            _deserializer = deserializer;
+            _referenceSearchValueParser = referenceSearchValueParser;
+        }
+
+        public async Task<SearchResourceResponse> Handle(SearchResourceRequest request, CancellationToken cancellationToken, RequestHandlerDelegate<SearchResourceResponse> next)
+        {
+            EnsureArg.IsNotNull(request, nameof(request));
+
+            Tuple<string, string> listParameter = request.Queries
+                .FirstOrDefault(x => string.Equals(x.Item1, _listParameter, StringComparison.Ordinal));
```
tamirkamara, line 61: Is this case-sensitive on purpose?
I see cases where other search parameters are compared without case, like [this](https://github.com/microsoft/fhir-server/blob/6e1ed88eb20a9b73a67399870963d12ed1f786b4/src/Microsoft.Health.Fhir.Core/Extensions/SearchServiceExtensions.cs#L28).

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Behaviors/ListSearchPipeBehavior.cs
```diff
@@ -0,0 +1,131 @@
...
+
+            // if _list was not requested, continue...
+            if (listParameter == null)
+            {
+                return await next();
+            }
+
+            // create an empty response to return if needed
+            SearchResourceResponse emptyResponse = CreateEmptySearchResponse(request);
```
tamirkamara, line 70: I realize this is just for code cleanliness, but is it really required? 
I'm asking since the method creating it isn't that simple so creating the object each time even if not used might be too much... IMO using `return CreateEmptySearchResponse(request)` each time isn't that much of a duplication of code.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Behaviors/ListSearchPipeBehavior.cs
eladiw: good catch.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Behaviors/ListSearchPipeBehavior.cs
eladiw: im good with both options. I will change as it will add (minor) perf upgrade

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/Behaviors/ListSearchBehaviorTests.cs
```diff
@@ -0,0 +1,216 @@
...
+
+            var so = new SearchOptions();
+            so.UnsupportedSearchParams = new Tuple<string, string>[0];
+            so.UnsupportedSortingParams = Array.Empty<(string searchParameterName, string reason)>();
+
+            _searchOptionsFactory = Substitute.For<ISearchOptionsFactory>();
+            _searchOptionsFactory.Create(Arg.Any<string>(), Arg.Any<IReadOnlyList<Tuple<string, string>>>()).Returns(so);
+
+            _resourceDeserializer = new ResourceDeserializer(
+               (FhirResourceFormat.Json, new Func<string, string, DateTimeOffset, ResourceElement>((str, version, lastUpdated) => _fhirJsonParser.Parse(str).ToResourceElement())));
```
brendankowitz, line 60: There is a test helper Microsoft.Health.Fhir.Tests.Common.ResourceDeserializer that you should be able to use here

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Behaviors/ListSearchPipeBehavior.cs
```diff
@@ -0,0 +1,131 @@
...
+            }
+
+            // Remove the 'list' params from the queries, to be later replaced with specific ids of resources
+            IEnumerable<Tuple<string, string>> query = request.Queries.Except(new[] { listParameter });
+
+            // prepare an empty response (lazy) in case we need it
+            _emptyResponse = new Lazy<SearchResourceResponse>(() => CreateEmptySearchResponse(request));
+
+            // if _list was requested with invalid value. continue
+            if (string.IsNullOrWhiteSpace(listParameter.Item2))
```
brendankowitz, line 76: Perhaps this should be merged with the `if` statement above

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/Behaviors/ListSearchBehaviorTests.cs
```diff
@@ -0,0 +1,216 @@
...
+                    },
+                },
+            }.ToResourceElement();
+
+            _bundleFactory = Substitute.For<IBundleFactory>();
+            _bundleFactory.CreateSearchBundle(Arg.Any<SearchResult>()).Returns(_nonEmptyBundle);
+        }
+
+        [Fact]
+        public async Task GivenARequest_WhenNoListQuery_QueriesUnchanged()
```
brendankowitz, line 113: Assert that QueriesAreUnchanged

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Behaviors/ListSearchPipeBehavior.cs
```diff
@@ -0,0 +1,131 @@
...
+            if (listParameter == null)
+            {
+                return await next();
+            }
+
+            // Remove the 'list' params from the queries, to be later replaced with specific ids of resources
+            IEnumerable<Tuple<string, string>> query = request.Queries.Except(new[] { listParameter });
+
+            // prepare an empty response (lazy) in case we need it
+            _emptyResponse = new Lazy<SearchResourceResponse>(() => CreateEmptySearchResponse(request));
```
brendankowitz, line 73: I don't think `_emptyResponse` should be a field unless can utilize a field or constructor initializer. To keep things simple its probably ok to just `return CreateEmptySearchResponse(request)` when needed

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Behaviors/ListSearchPipeBehavior.cs
```diff
@@ -0,0 +1,131 @@
...
+            IReadOnlyList<(string parameterName, string reason)> unsupportedSortingParameters;
+            if (searchOptions.Sort?.Count > 0)
+            {
+                // we don't currently support sort
+                unsupportedSortingParameters = searchOptions.UnsupportedSortingParams.Concat(searchOptions.Sort.Select(s => (s.searchParameterInfo.Name, Core.Resources.SortNotSupported))).ToList();
+            }
+            else
+            {
+                unsupportedSortingParameters = searchOptions.UnsupportedSortingParams;
+            }
```
brendankowitz, line 122: I think we should leave this logic out of here

eladiw:
@brendankowitz , fixed according to your comments

PR #1097, opened 2020-06-23
Update export test to cancel export request after test is complete

namadabu:
## Description
Updating the auth export test to cancel export request after the test is complete. The test only aims to validate that the export API works as expected. Cancelling the export request will prevent the export job from being picked up by the ExportJobWorker and trying to actually export data.

## Related issues
Addresses [issue [AB#74585](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/74585)].

## Testing

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Client/FhirClient.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Client/FhirClient.cs
@@ -322,6 +322,17 @@ namespace Microsoft.Health.Fhir.Client
             return response;
         }
 
+        public async Task CancelExport(Uri contentLocation, CancellationToken cancellationToken = default)
+        {
+            var message = new HttpRequestMessage
+            {
+                Method = HttpMethod.Delete,
+                RequestUri = contentLocation,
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Fhir.Shared.Client/FhirClient.cs
```diff
@@ -322,6 +322,17 @@ public async Task<HttpResponseMessage> CheckExportAsync(Uri contentLocation, Can
             return response;
         }
 
+        public async Task CancelExport(Uri contentLocation, CancellationToken cancellationToken = default)
+        {
+            var message = new HttpRequestMessage
```
brandonpollett, line 327: nit: `using var message = new HttpRequestMessage`

namadabu: Updated. Did not know that we could skip the braces for the `using` block starting from C# 8.0! Thanks :)

PR #1098, opened 2020-06-23
Split tests based on database type

LTA-Thinking:
## Description
Splits the E2E tests based on database type. Since each database is a separate service having them run in separate jobs allow for parallelization.

## Testing
Automated build


build/jobs/run-tests.yml
```diff
@@ -3,16 +3,9 @@ parameters:
   type: string
 - name: keyVaultName
   type: string
-- name: runIntegration
-  type: boolean
-  default: true
-- name: runE2E
-  type: boolean
-  default: true
```
rotodd, line 11: For my learning, why were these parameters taken out?

build/jobs/run-tests.yml
```diff
@@ -105,6 +97,67 @@ jobs:
...
+
+        dotnet dev-certs https
+
+  - task: VSTest@2
+    displayName: 'Run E2E Tests'
+    inputs:
+      testSelector: testAssemblies
+      testAssemblyVer2: |
+        **\*${{ parameters.version }}.Tests.E2E*.dll
+      testFiltercriteria: "FullyQualifiedName~SqlServer"
```
rotodd, line 160: Another question for my understanding: why was `testFiltercriteria` added?

build/jobs/run-tests.yml
LTA-Thinking: They weren't being used, so I removed them for readability. When this script was first written they were added to help with testing, but there is now no use for running only one type of test.

build/jobs/run-tests.yml
LTA-Thinking: Test filter criteria is what splits the tests into CosmosDB and SqlServer tests. This argument filters tests based on their names, and the test names have the name of the database included in them. The filter makes it so only the Cosmos tests run in one job and only the Sql tests run in the other job.

build/jobs/run-tests.yml
rotodd: Ah I see - so the conditions on line 15 and 55 would always be true?

build/jobs/run-tests.yml
rotodd: Ah I see. Thank you for the explanation.

LTA-Thinking:
That is correct. The actual improvements unfortunately haven't been as good as hoped because we have a limited number of build agents available to us (which limits the number of parallel jobs), but there has been a noticeable time improvement (~5 minutes). If we got more build agents the performance would be better.

Another benefit of this is debugging and rerunning tests. Since they are now split based on database it is easier to see if an issue is isolated to one database and allows for the rerunning of only one type of test if needed (avoids rerunning tests that passed).

build/jobs/run-tests.yml
```diff
@@ -51,8 +44,7 @@ jobs:
       'CosmosDb:Key': $(CosmosDb--Key)
       'SqlServer:ConnectionString': $(SqlServer--ConnectionString)
 
-- job: 'e2eTests'
-  condition: eq('${{ parameters.runE2E }}', true)
+- job: 'cosmosE2eTests'
```
brendankowitz, line 47: Is it possible to align the naming with the AKS version at all?
`R4 SqlServer AKS Tests e2eTests`

LTA-Thinking: The AKS tests are split at the stage level, while these are split at the job level. I can change the names, but they are already under a stage that references the version being run.

PR #1099, opened 2020-06-24
Add max resources and pages to job record

LTA-Thinking:
## Description
Adds snapshots of the MaximumNumberOfResourcesPerQuery and NumberOfPagesPerCommit settings to export job records. If the number of resources per query is changed in the middle of an export job it will invalidate the continuation token and mess up the search results.

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
@@ -111,7 +111,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Export
                 // from the search result.
                 var queryParametersList = new List<Tuple<string, string>>()
                 {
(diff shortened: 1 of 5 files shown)


src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -111,7 +111,7 @@ public async Task ExecuteAsync(ExportJobRecord exportJobRecord, WeakETag weakETa
                 // from the search result.
                 var queryParametersList = new List<Tuple<string, string>>()
                 {
```
namadabu, line 113: nit: Since we are now using `_exportJobRecord` for almost everything in `ExportJobTask`, I am wondering whether it might be better to update the `ConnectAsync` and other downstream methods to take in a single string instead of an `ExportJobConfiguration`. That way we use the `ExportJobConfiguration` only for the initial comparison and then base everything else off of `ExportJobRecord`

LTA-Thinking: I think that would be a good idea, but it would not be a small refactor. I'll make anther work item for it.

Issue #1102, opened 2020-06-25
Link Next have issue 


ddst0111:
**Describe the bug**
When a Encounter resource have more than 200+ mediciationRequest resource, query MedicationRequest using Encounter identifier cannot get all MedicationRequest

**To Reproduce**
1.1. the first query
 https://{{fhirserver}}/MedicationRequest?context:Encounter.identifier=12345673&_include=MedicationRequest:medication

will return the link like 
"link": [
        {
            "relation": "next",
            "url": "https://{{fhirserver}}/MedicationRequest?context%3AEncounter.identifier=12345673&_include=MedicationRequest%3Amedication&ct=NTA5ODI5MDk5NTI3NDUwMTM5Ng%3D%3D"
        },
        {
            "relation": "self",
            "url": "https://{{fhilserver}}/MedicationRequest?context%3AEncounter.identifier=12345673&_include=MedicationRequest%3Amedication"
        }
    ],

 
2. when use the "next" url returned from the first query, i.e., expecting the result will have a new "next" link
https://azurefhir-prod.azurewebsites.net/MedicationRequest?context%3AEncounter.identifier=12345673&_include=MedicationRequest%3Amedication&ct=NTA5ODI5MDk5NTI3NDUwMTM5Ng%3D%3D

**Expected behavior**
A clear and concise description of what you expected to happen, we expect the "next" link will have a new "ct", so can continue the navigation

**Actual behavior**
But the next link returned is the same as the 
"link": [
        {
            "relation": "next",
            "url": "https://{{fhirserver}}/MedicationRequest?_include=MedicationRequest%3Amedication&context%3AEncounter.identifier=12345673&ct=NTA5ODI5MDk5NTI3NDUwMTM5Ng%3D%3D"
        },
        {
            "relation": "self",
            "url": "https://{{fhirserver}}/MedicationRequest?_include=MedicationRequest%3Amedication&context%3AEncounter.identifier=12345673&ct=NTA5ODI5MDk5NTI3NDUwMTM5Ng%3D%3D"
        }
    ]

ddst0111:
when query with the _summary=count can see the total is: 249
{
    "resourceType": "Bundle",
    "id": "3141835d-2989-40e8-8cc0-f796b7d67f0e",
    "type": "searchset",
    "total": 249
}

johnstairs:
Potentially the same issue as #1095. Investigating....

Issue #1095, opened 2020-06-23
Page results for searches with reference chaining are incorrect after the first page (SQL)


7opf:
[repro-transaction.zip](https://github.com/microsoft/fhir-server/files/4831971/repro-transaction.zip)

**Describe the bug**

Page results for searches with reference chaining are incorrect after the first page (SQL).

**To Reproduce**
1. Create a search with `_count=x`, `_total=accurate` and a chained reference parameter e.g.
```
/Observation?patient.organization=Organization/123&_count=2&_total=accurate
```
2.  First page returns correct `Bundle.total` (ensure > x) and first page of results (x results). However following the link with relation "next" yields a page with only one entry in the bundle and no relation for "next" in `Bundle.link`.

**Expected behavior**

Following links with the continuation token should return the next set of results as per the chained query.

**Actual behavior**

Server returns a second page with one entry and no subsequent pages

7opf:
Added transaction to reproduce the issue

anyname2:
Generated SQL query for the second page has a `SELECT DISTINCT TOP (@p6)` where `@p6` is set to `_count+1` (see `cte2` in the query below), which limits the results before the continuation token is applied.
```sql
DECLARE @p0 SmallInt = 200;
DECLARE @p1 SmallInt = 96;
DECLARE @p2 SmallInt = 103;
DECLARE @p3 SmallInt = 940;
DECLARE @p4 SmallInt = 100;
DECLARE @p5 VarChar(64) = '20489ae9-ee15-4805-8749-442f409bdbfe';
DECLARE @p6 Int = 3;
DECLARE @p7 BigInt = 5098295252858560160;

SET STATISTICS IO ON;
SET STATISTICS TIME ON;

WITH cte0 AS
(
    SELECT ref.ResourceSurrogateId AS Sid1, r.ResourceSurrogateId AS Sid2
    FROM dbo.ReferenceSearchParam ref
    INNER JOIN dbo.Resource r
    ON ref.ReferenceResourceTypeId = r.ResourceTypeId
        AND ref.ReferenceResourceId = r.ResourceId
    WHERE ref.SearchParamId = @p0
        AND r.IsHistory = 0
        AND ref.IsHistory = 0
        AND ref.ResourceTypeId = @p1
        AND ref.ReferenceResourceTypeId = @p2
),
cte1 AS
(
    SELECT Sid1, ResourceSurrogateId AS Sid2
    FROM dbo.ReferenceSearchParam
[...]

PR #1106, opened 2020-06-26
Fix continuation token issues with chained searches

johnstairs:
## Description
We were not "promoting" denormalized predicates to chained table expressions. This would have the result of the continuation clause (a predicate on `ResourceSurrgodateId`) being applied after the `TOP` clause had been applied.

The fix is to promote the denomalized predicate to the chained table expression, but to not mix it in with the `DenormalizedPredicate` field -  which is used to store the denormalized predicate on chain target - but to introduce a new `DenormalizedPredicateOnChainRoot` property.

## Related issues
Addresses #1095 
Addresses #1102  

## Testing
Added E2E and unit tests.

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/TableExpression.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/TableExpression.cs
@@ -20,7 +20,8 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Search.Expressions
             Expression normalizedPredicate,
             Expression denormalizedPredicate,
             TableExpressionKind kind,
-            int chainLevel = 0)
+            int chainLevel = 0,
+            Expression denormalizedPredicateOnChainRoot = null)
         {
             switch (normalizedPredicate)
             {
@@ -39,11 +40,14 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Search.Expressions
             DenormalizedPredicate = denormalizedPredicate;
             Kind = kind;
             ChainLevel = chainLevel;
+            DenormalizedPredicateOnChainRoot = denormalizedPredicateOnChainRoot;
         }
 
         public TableExpressionKind Kind { get; }
 
         public int ChainLevel { get; }
+
+        public Expression DenormalizedPredicateOnChainRoot { get; }
 
         public NormalizedSearchParameterQueryGenerator SearchParameterQueryGenerator { get; }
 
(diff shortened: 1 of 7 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -245,43 +245,48 @@ public object VisitTable(TableExpression tableExpression, SearchOptions context)
...
-                            .Append(" = ").Append(VLatest.Resource.ResourceId, resourceTableAlias);
+                        delimited.BeginDelimitedElement().Append(VLatest.ReferenceSearchParam.ReferenceResourceId, referenceSourceTableAlias)
+                            .Append(" = ").Append(VLatest.Resource.ResourceId, referenceTargetResourceTableAlias);
                     }
 
-                    // Denormalized predicates on reverse chains need to be applied via a join to the resource table
-                    if (tableExpression.DenormalizedPredicate != null && chainedExpression.Reversed)
+                    // For reverse chaning, if there is a parameter on the _id search parameter, we need another join to get the resource ID of the reference source (all we have is the surrogate ID at this point)
+
+                    bool denormalizedHandledBySecondJoin = tableExpression.DenormalizedPredicate != null && chainedExpression.Reversed && tableExpression.DenormalizedPredicate.AcceptVisitor(DenormalizedExpressionContainsIdParameterVisitor.Instance, null);
```
johnstairs, line 274: @brandonpollett, we only need to this extra join if the expression is over the `_id` search parameter, otherwise the existing join is sufficient.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ExportDataValidationTests.cs
```diff
@@ -51,17 +51,17 @@ public async Task GivenFhirServer_WhenDataIsExported_ThenExportedDataIsSameAsDat
             IList<Uri> blobUris = await CheckExportStatus(contentLocation);
 
             // Download exported data from storage account
-            Dictionary<string, string> dataFromExport = await DownloadBlobAndParse(blobUris);
+            Dictionary<(string resourceType, string resourceId), string> dataFromExport = await DownloadBlobAndParse(blobUris);
```
johnstairs, line 54: Changes is here are unrelated, but these tests were failing when my database had resources with different types but the same id.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/TableExpression.cs
```diff
@@ -20,7 +20,8 @@ internal class TableExpression : Expression
             Expression normalizedPredicate,
             Expression denormalizedPredicate,
             TableExpressionKind kind,
-            int chainLevel = 0)
+            int chainLevel = 0,
+            Expression denormalizedPredicateOnChainRoot = null)
```
brendankowitz, line 24: It might be helpful to add some parameter documentation to some of these common types, this constructor has 3 expression parameters now :)

johnstairs: You are right!

PR #1108, opened 2020-06-29
Add export for Patient compartment data

LTA-Thinking:
## Description
Exports resources that references Patients when exporting Patients.

## Related issues
[User Story 74398: Add Compartment resources to Patient export](https://microsofthealth.visualstudio.com/Health/_workitems/edit/74398)

## Testing
Added new unit test and modified existing E2E tests


src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Export/InMemoryExportDestinationClient.cs
```diff
@@ -106,5 +106,10 @@ public string GetExportedData(Uri fileUri)
 
             return null;
         }
+
+        public int GetExportedDataFileCount()
```
brendankowitz, line 110: Looks like it should be a property
`public int ExportedDataFileCount => ...`

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -218,7 +167,127 @@ private async Task UpdateJobRecordAsync(CancellationToken cancellationToken)
...
-        private async Task ProcessSearchResultsAsync(IEnumerable<SearchResultEntry> searchResults, uint partId, CancellationToken cancellationToken)
+        private async Task RunExportSearch(
+            ExportJobConfiguration exportJobConfiguration,
+            ExportJobProgress progress,
+            List<Tuple<string, string>> sharedQueryParametersList,
+            uint pagesWaitingForCommit,
+            string batchIdPrefix,
+            CancellationToken cancellationToken,
+            bool compartmentSearch = false)
+        {
```
brendankowitz, line 178: Looks like we should have some ensureArgs in here

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/Models/ExportJobProgress.cs
```diff
@@ -27,12 +29,28 @@ protected ExportJobProgress()
         [JsonProperty(JobRecordProperties.Page)]
         public uint Page { get; private set; }
 
+        [JsonProperty(JobRecordProperties.TriggeringResourceId)]
+        public string TriggeringResourceId { get; private set; }
+
+        [JsonProperty(JobRecordProperties.SubSearch)]
+        public ExportJobProgress SubSearch { get; private set; }
```
brendankowitz, line 36: Would it make sense to call this CompartmentSearch?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
LTA-Thinking: Added

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Export/InMemoryExportDestinationClient.cs
LTA-Thinking: Changed

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/Models/ExportJobProgress.cs
LTA-Thinking: I went with sub search because this is also needed for Group export when searching for resources in nested Groups. I wanted a name that would be accurate for both scenarios.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ExportController.cs
```diff
@@ -149,9 +150,9 @@ public async Task<IActionResult> CancelExport(string idParameter)
             return new ExportResult(response.StatusCode);
         }
 
-        private async Task<IActionResult> SendExportRequest(PartialDateTime since, string type = null)
+        private async Task<IActionResult> SendExportRequest(ExportJobType exportType, PartialDateTime since, string resourceType = null)
```
namadabu, line 153: nit: Seems like `resourceType` param may no longer be needed here.

src/Microsoft.Health.Fhir.Core/Messages/Export/CreateExportRequest.cs
```diff
@@ -6,23 +6,27 @@
...
 {
     public class CreateExportRequest : IRequest<CreateExportResponse>
     {
-        public CreateExportRequest(Uri requestUri, string resourceType = null, PartialDateTime since = null)
+        public CreateExportRequest(Uri requestUri, ExportJobType requestType, string resourceType = null, PartialDateTime since = null)
         {
             EnsureArg.IsNotNull(requestUri, nameof(requestUri));
 
             RequestUri = requestUri;
+            RequestType = requestType;
```
namadabu, line 21: The reason we initially added `resourceType` here was to have support for `Patient` and other resources in the future. Since `ExportJobType` is being used to track that, it seems like we should just remove `resourceType` from here and downstream (`ExportJobRecord`). I don't think it's being used anywhere downstream either with the current change. Will make things cleaner since we can avoid confusion on whether we have to use `ExportJobType` or `resourceType`

LTA-Thinking: When support for the _type parameter is added resource type will still be needed to pass that information. It isn't being used right now, but _type is the next story to be worked on so I don't want to remove it here.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -194,7 +143,7 @@ public async Task ExecuteAsync(ExportJobRecord exportJobRecord, WeakETag weakETa
                 // Try to update the job to failed state.
                 _logger.LogError(ex, "Encountered an unhandled exception. The job will be marked as failed.");
 
-                _exportJobRecord.FailureDetails = new JobFailureDetails(Resources.UnknownError, HttpStatusCode.InternalServerError);
+                _exportJobRecord.FailureDetails = new JobFailureDetails(ex.StackTrace, HttpStatusCode.InternalServerError);
```
namadabu, line 146: This message gets returned to the customer when they request export status. So I don't think we should be sending back the stack trace.

LTA-Thinking: Ah woops, that was to help me debug

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -218,7 +167,131 @@ private async Task UpdateJobRecordAsync(CancellationToken cancellationToken)
...
+                if (searchResult.ContinuationToken == null)
+                {
+                    // No more continuation token, we are done.
+                    break;
+                }
+
+                // Update the continuation token in local cache and queryParams.
+                // We will add or udpate the continuation token to the end of the query parameters list.
+                progress.UpdateContinuationToken(searchResult.ContinuationToken);
+                pagesWaitingForCommit++;
```
namadabu, line 265: `pagesWaitingForCommit` doesn't seem to be used anywhere?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -218,7 +167,131 @@ private async Task UpdateJobRecordAsync(CancellationToken cancellationToken)
...
+                    await UpdateJobRecordAsync(cancellationToken);
+
+                    currentBatchId = batchIdPrefix + '-' + progress.Page.ToString("d6");
+                    pagesWaitingForCommit = 0;
+                }
+            }
+
+            // Commit one last time for any pending changes.
+            await _exportDestinationClient.CommitAsync(exportJobConfiguration, cancellationToken);
+            await UpdateJobRecordAsync(cancellationToken);
```
namadabu, line 291: After the method exits, we call `CompleteJobAsync()` which calls `UpdateJobRecordAsync()`. We can avoid the `UpdateJobRecordAsync()` call here I think

LTA-Thinking: I added this to account for the recursive nature of the method. Since there can now be sub searches this records the progress when a sub search ends.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
LTA-Thinking: Good catch. I was messing around with the pages to commit settings and this is a remnant.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -218,7 +167,128 @@ private async Task UpdateJobRecordAsync(CancellationToken cancellationToken)
...
+
+                if (_exportJobRecord.ExportType == ExportJobType.Patient && !compartmentSearch)
+                {
+                    uint resultIndex = 0;
+                    foreach (SearchResultEntry result in searchResult.Results)
+                    {
+                        // If a job is resumed in the middle of processing patient compartment resources it will skip patients it has already exported compartment information for.
+                        if (progress.SubSearch != null && result.Resource.ResourceId != progress.SubSearch.TriggeringResourceId)
+                        {
+                            continue;
```
namadabu, line 238: In scenarios where we hit this code, the value of `resultIndex` will be out-of sync with the iterator. We could end up overwriting existing data in the blobs.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -218,7 +167,128 @@ private async Task UpdateJobRecordAsync(CancellationToken cancellationToken)
...
+
+                // Search and process the results.
+                using (IScoped<ISearchService> searchService = _searchServiceFactory())
+                {
+                    if (compartmentSearch)
+                    {
+                        searchResult = await searchService.Value.SearchCompartmentAsync(
+                            KnownResourceTypes.Patient,
+                            progress.TriggeringResourceId,
+                            null,
```
namadabu, line 206: nit: please call out parameter names when passing in non-descriptive values (like null). Makes it easier to read the code.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -218,7 +167,128 @@ private async Task UpdateJobRecordAsync(CancellationToken cancellationToken)
...
+                        {
+                            continue;
+                        }
+
+                        if (progress.SubSearch == null)
+                        {
+                            progress.NewSubSearch(result.Resource.ResourceId);
+                        }
+
+                        await RunExportSearch(exportJobConfiguration, progress.SubSearch, sharedQueryParametersList, cancellationToken, currentBatchId + ":" + resultIndex.ToString("d6"), true);
```
namadabu, line 246: nit: I feel like using the current method recursively for the compartment search makes the code harder to read/debug. There are similarities and subtle differences in behavior. Would it be better to have something like `RunExportCompartmentSearch` that that deals only with compartment search scenario? That will also reduce the number of if conditions in the current method. Some of the common code could be extracted into helper methods if required.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -218,7 +167,128 @@ private async Task UpdateJobRecordAsync(CancellationToken cancellationToken)
...
+                        }
+                    }
+                }
+
+                if (_exportJobRecord.ExportType == ExportJobType.Patient && !compartmentSearch)
+                {
+                    uint resultIndex = 0;
+                    foreach (SearchResultEntry result in searchResult.Results)
+                    {
+                        // If a job is resumed in the middle of processing patient compartment resources it will skip patients it has already exported compartment information for.
```
namadabu, line 235: nit: There is an assumption here that search results and ordering of the results is consistent across (same) query runs. Might be helpful to call it out in the comments.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
LTA-Thinking: This is mostly my preference showing, I really like recursive methods :) 
I agree though that this isn't the best place for it, I'll refactor.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/Models/ExportJobProgress.cs
```diff
@@ -27,12 +29,28 @@ protected ExportJobProgress()
...
         {
             EnsureArg.IsNotNullOrWhiteSpace(continuationToken, nameof(continuationToken));
 
             ContinuationToken = continuationToken;
             Page++;
         }
+
+        public void NewSubSearch(string resourceId)
+        {
+            SubSearch = new ExportJobProgress(null, 0, resourceId);
```
namadabu, line 48: nit: param validation here?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ExportDataValidationTests.cs
```diff
@@ -39,24 +39,53 @@ public ExportDataValidationTests(HttpIntegrationTestFixture fixture, ITestOutput
...
+            // NOTE: Azure Storage Emulator is required to run these tests locally.
+
+            // Trigger export request and check for export status
+            Uri contentLocation = await _testFhirClient.ExportAsync("Patient/");
+            IList<Uri> blobUris = await CheckExportStatus(contentLocation);
+
+            // Download exported data from storage account
+            Dictionary<(string resourceType, string resourceId), string> dataFromExport = await DownloadBlobAndParse(blobUris);
+
+            // Download all resources from fhir server
```
namadabu, line 73: nit: comment needs to be updated to `Patient` resources

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -218,7 +167,149 @@ private async Task UpdateJobRecordAsync(CancellationToken cancellationToken)
...
+                                resourceType: null,
+                                queryParametersList,
+                                cancellationToken);
+                            break;
+                        case ExportJobType.Patient:
+                            searchResult = await searchService.Value.SearchAsync(
+                                resourceType: KnownResourceTypes.Patient,
+                                queryParametersList,
+                                cancellationToken);
+                            break;
```
namadabu, line 212: nit: do we need a default case here?

LTA-Thinking: The switch is removed in the _type PR

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ExportDataValidationTests.cs
LTA-Thinking: Fixed in the _type PR

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/Models/ExportJobProgress.cs
LTA-Thinking: Added in _type PR

PR #1113, opened 2020-06-29
Update implementation of FhirClient to use Credential providers.

brandonpollett:
## Description
This PR implements the shared credential provider in the FHIR E2E tests.

## Related issues
Addresses [AB#74688](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/74688)

## Testing
Existing E2E tests continue to pass after using new credential mechanism.


test/Microsoft.Health.Fhir.Shared.Tests.Crucible/CrucibleDataSource.cs
```diff
@@ -55,11 +55,6 @@ public static async Task<string[]> GetSupportedIdsAsync(CrucibleClient client)
                     .GetTestFhirServer(_dataStore, null)
                     .GetTestFhirClient(ResourceFormat.Json);
 
-                if (testFhirClient?.SecurityEnabled == true)
```
brandonpollett, line 58: Some form of this still needs to happen.

jackliums: when would the client be null?

test/Microsoft.Health.Fhir.Shared.Tests.E2E.Common/TestAuthenticationHttpMessageHandler.cs
```diff
@@ -0,0 +1,31 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Health.Fhir.Tests.E2E.Common
+{
+    public class TestAuthenticationHttpMessageHandler : DelegatingHandler
```
jackliums, line 13: should this be in shared componnet? Seems like this would needed anywhere that sets the token right?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Metric/MetricTests.cs
```diff
@@ -36,6 +36,9 @@ public MetricTests(MetricTestFixture fixture)
             _fixture = fixture;
             _client = fixture.TestFhirClient;
             _metricHandler = _fixture?.MetricHandler;
+
+            // Send an empty request to guarantee that there is a bearer token set and the call isn't recorded in the metric handler.
+            _client.HttpClient.SendAsync(new HttpRequestMessage(HttpMethod.Get, string.Empty)).GetAwaiter().GetResult();
```
jackliums, line 41: Can we implement IAsyncLifetime and move this there?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/TestFhirServer.cs
```diff
@@ -52,26 +69,125 @@ public TestFhirClient GetTestFhirClient(ResourceFormat format, TestApplication c
...
+                    sessionMessageHandler.InnerHandler = _authenticationHandlers[authKey];
+                }
+                else
+                {
+                    AuthenticationHttpMessageHandler authHandler;
+
+                    var authHttpClient = new HttpClient(new SessionMessageHandler(CreateMessageHandler(), _asyncLocalSessionTokenContainer)) { BaseAddress = BaseAddress };
+                    if (user == null)
+                    {
+                        string scope = clientApplication.Equals(TestApplications.WrongAudienceClient) ? clientApplication.ClientId : AuthenticationSettings.Scope;
```
jackliums, line 108: I think there are a little bit too much logic (or test specific logic?) in this method. Would it be more readable if we separate these out?

For example, there is CreateFhirClient(ResourceFormat format, DelegatingHandler authenticationHandler) that creates a FhirClient with whatever the handler that's passed in. There could be another method that reads the CapabilityStatement and use the clientApplication/user to create an appropriate handler and call the previous method. There could be another method that creates handler using an invalid application id and call the previous method. Just to separate the logic out a litte?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/TestFhirServer.cs
```diff
@@ -52,26 +69,125 @@ public TestFhirClient GetTestFhirClient(ResourceFormat format, TestApplication c
...
+                        new Lazy<TestFhirClient>(() => CreateFhirClient(tuple.format, tuple.clientApplication, tuple.user, authenticationHandler)),
                     this)
                 .Value;
         }
 
-        private TestFhirClient CreateFhirClient(ResourceFormat format, TestApplication clientApplication, TestUser user)
+        private TestFhirClient CreateFhirClient(ResourceFormat format, TestApplication clientApplication, TestUser user, DelegatingHandler authenticationHandler = null)
         {
-            var httpClient = new HttpClient(new SessionMessageHandler(CreateMessageHandler(), _asyncLocalSessionTokenContainer)) { BaseAddress = BaseAddress };
+            ConfigureSecurityOptions().GetAwaiter().GetResult();
```
jackliums, line 85: I wonder if this could just be part of the SendAsync in chained the handler?

Something like FhirClient -> HttpClient -> AuthenticationHandler -> SessionMessageHandler

AuthenticationHandler.SendAsync will do a few things. 1. Read CapatibilityStatement if it hasn't. 2. If SecurityOption is enabled, then delegate to some external class to configure and create appropriate handler/provider that can get/set token. 3. Cache the token and check expiration date and starts refresh if it's within X minutes.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/TestFhirServer.cs
```diff
@@ -52,26 +69,125 @@ public TestFhirClient GetTestFhirClient(ResourceFormat format, TestApplication c
...
+                string authKey = GenerateKey();
+                if (_authenticationHandlers.ContainsKey(authKey))
+                {
+                    sessionMessageHandler.InnerHandler = _authenticationHandlers[authKey];
+                }
+                else
+                {
+                    AuthenticationHttpMessageHandler authHandler;
+
+                    var authHttpClient = new HttpClient(new SessionMessageHandler(CreateMessageHandler(), _asyncLocalSessionTokenContainer)) { BaseAddress = BaseAddress };
```
jackliums, line 105: If we do this in the client handler, then we dont need to create new HttpClient each time. The SendAsync could just execute additional HttpRequestMessage (to get token and etc) before execute the HttpRequestMessge that was passed in?

test/Microsoft.Health.Fhir.Shared.Tests.E2E.Common/TestAuthenticationHttpMessageHandler.cs
brandonpollett: This could be in a shared component but today it's only used in two audit tests as a way to override the default implementation in tests.

Also, I don't think the code should be in the `Microsoft.Health.Client` package so it would require a different package. I think the code is simple and one off enough today to not move yet.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/TestFhirServer.cs
brandonpollett: Realistically, this only needs to be executed once per server. I'm going to look for a way to move this out of client creation and move it into when the `TestFhirServer` is realized. 

I don't think the `AuthenticationHandler` should be concerned with this because this is really a test specific problem in my opinion. In normal uses of the `Client` I would expect the developer to know how they're going to obtain a token and there would be no need to dynamically do it based on configuration exposed by the server.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/TestFhirServer.cs
brandonpollett: The `HttpClient` is used in the implementations of the `ICredentialProvider` that we're using here. There are additional `ICredentialProviders` that don't directly make `HttpClient` calls such as `ManagedIdentityCredentialProvider`. All the `AuthenticationHttpMessageHandler` knows is that it has an `ICredentialProvider` that exposes a method `GetBearerToken()`. 

Again, I think this may be a bit of a test specific thing in that in most cases you wouldn't be creating multiple `ICredentialProviders`, but only 1 of them for a life of an application. In this case, we're creating one per cached handler.

Issue #1011, opened 2020-05-07
SQL DataProvider _total=accurate does not work with _include searches


brendankowitz:
**Describe the bug**
When performing a search using both _include and _total=accurate, an exception is thrown due to incorrect SQL being generated to get the total count.

Example URL:

> https://[fhir-server]/DiagnosticReport?_include=DiagnosticReport:patient:Patient&code=429858000&_count=1&_format=json&**_total=accurate**

The same query works without the count:
> https://[fhir-server]/DiagnosticReport?_include=DiagnosticReport:patient:Patient&code=429858000&_count=1&_format=json&**_total=none**

Related E2E tests:
GivenAnIncludeSearchExpressionWithSimpleSearchAndCount_WhenSearched_ThenCorrectBundleShouldBeReturned

**Expected behavior**
A bundle is returned including the total count

**Actual behavior**
Incorrect syntax near ')'. Incorrect syntax near the keyword 'UNION'. Incorrect syntax near ')'.

```
SqlException: Incorrect syntax near ')'. Incorrect syntax near the keyword 'UNION'. Incorrect syntax near ')'.
System.Data.SqlClient.SqlConnection.OnError(SqlException exception, bool breakConnection, Action<Action> wrapCloseInAction)
System.Data.SqlClient.SqlInternalConnection.OnError(SqlException exception, bool breakConnection, Action<Action> wrapCloseInAction)
System.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, bool callerHasConnectionLock, bool asyncClose)
System.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, out bool dataReady)
System.Data.SqlClient.SqlDataReader.TryConsumeMetaData()
System.Data.SqlClient.SqlDataReader.get_MetaData()
System.Data.SqlClient.SqlCommand.FinishExecuteReader(SqlDataReader ds, RunBehavior runBehavior, string resetOptionsString)
System.Data.SqlClient.SqlCommand.CompleteAsyncExecuteReader()
System.Data.SqlClient.SqlCommand.InternalEndExecuteReader(IAsyncResult asyncResult, string endMethod)
System.Data.SqlClient.SqlCommand.EndExecuteReaderInternal(IAsyncResult asyncResult)
[...]

PR #1116, opened 2020-07-01
Fix errors when _summary=count or _total=accurate is specified with an _include parameter

johnstairs:
## Description
We generated invalid SQL when `_summary=count` and `_include` are specified together. For `total=accurate` we would either have the same error or would return an incorrect total id the results all fit in a single bundle.

The fix is to add another visitor pass and to only count match results when computing total from a bundle.

## Related issues
Addresses #1011 

## Testing
Unit and E2E tests

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/RemoveIncludesRewriter.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/RemoveIncludesRewriter.cs
@@ -1,0 +1,54 @@
+// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Health.Fhir.Core.Features.Search.Expressions;
+
+namespace Microsoft.Health.Fhir.SqlServer.Features.Search.Expressions.Visitors
+{
+    /// <summary>
+    /// A rewriter that removes <see cref="IncludeExpression">s from an expression tree.
+    /// </summary>
+    internal class RemoveIncludesRewriter : ExpressionRewriterWithInitialContext<object>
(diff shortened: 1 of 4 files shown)


brendankowitz:
Can you also update my comment here now this is fixed :) https://github.com/microsoft/fhir-server/blob/master/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/IncludeSearchTests.cs#L125

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/IncludeSearchTests.cs
```diff
@@ -44,6 +44,14 @@ public async Task GivenAnIncludeSearchExpression_WhenSearched_ThenCorrectBundleS
                 locationResponse.Resource);
 
             ValidateSearchEntryMode(bundle, ResourceType.Location);
+
+            // ensure that the included resources are not counted
+            bundle = await Client.SearchAsync(ResourceType.Location, $"{query}&_summary=count");
+            Assert.Equal(1, bundle.Total);
+
+            // ensure that the included resources are not counted when _total is specified and the results fi in a single bundle.
```
rotodd, line 52: Nit: typo "fit"

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/RemoveIncludesRewriter.cs
```diff
@@ -0,0 +1,54 @@
...
+            List<Expression> newChildExpressions = null;
+            for (int i = 0; i < expression.Expressions.Count; i++)
+            {
+                Expression childExpression = expression.Expressions[i];
+                if (childExpression is IncludeExpression)
+                {
+                    if (i == 0 && expression.Expressions.All(e => e is IncludeExpression))
+                    {
+                        return null;
+                    }
```
feordin, line 35: Is it necessary to put this check inside the for loop?  It seems that if all the Expressions in expression are IncludeExpressions we return null from the method.  Can we not check that at the begining and return null.  It seems it would simplify the logic.

PR #1128, opened 2020-07-07
Auto-update for packages related to 'Microsoft.AspNet'

MicrosoftHealthService:
Updates package 'Microsoft.Extensions.Configuration.Binder' to version '3.1.5'
Updates package 'Microsoft.Extensions.Diagnostics.HealthChecks' to version '3.1.5'
Updates package 'Microsoft.Extensions.Logging.Abstractions' to version '3.1.5'
Updates package 'Microsoft.Extensions.Options' to version '3.1.5'
Updates package 'Microsoft.Extensions.Configuration.Json' to version '3.1.5'
Updates package 'Microsoft.Extensions.Hosting.Abstractions' to version '3.1.5'
Updates package 'Microsoft.Extensions.Http' to version '3.1.5'
Updates package 'Microsoft.Extensions.Logging' to version '3.1.5'
Updates package 'Microsoft.AspNetCore.Authentication.JwtBearer' to version '3.1.5'
Updates package 'Microsoft.AspNetCore.Mvc.Razor.RuntimeCompilation' to version '3.1.5'
Updates package 'Microsoft.Extensions.FileProviders.Embedded' to version '3.1.5'
Updates package 'Microsoft.Extensions.Options.ConfigurationExtensions' to version '3.1.5'
Updates package 'Microsoft.ApplicationInsights.AspNetCore' to version '2.14.0'
Updates package 'Microsoft.Extensions.Configuration.AzureKeyVault' to version '3.1.5'
Updates package 'Microsoft.Extensions.Logging.ApplicationInsights' to version '2.14.0'
Updates package 'Microsoft.AspNetCore.Mvc.NewtonsoftJson' to version '3.1.5'
Updates package 'Microsoft.Extensions.Diagnostics.HealthChecks.Abstractions' to version '3.1.5'
Updates package 'Microsoft.AspNetCore.Mvc.Testing' to version '3.1.5'
Updates package 'Microsoft.AspNetCore.TestHost' to version '3.1.5'


brendankowitz:
We should update all references to the SDK/Runtime to support 3.1.5 in this PR (deployment templates, docker base images, global.json)

brendankowitz:
Also: https://github.com/microsoft/fhir-server/blob/master/build/docker/Dockerfile

build/docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/dotnet/core/sdk:3.1.201-bionic AS build
+FROM mcr.microsoft.com/dotnet/core/sdk:3.1-bionic AS build
```
brendankowitz, line 1: Originally the entire sdk tag was used as we were seeing build errors when it differed from the Global.json. If the docker image moves forward before our global.json we may see random breaks in the future

build/docker/Dockerfile
```diff
@@ -9,7 +9,7 @@ COPY ./ ./
 RUN dotnet build --configuration Release
 RUN dotnet publish "./src/Microsoft.Health.Fhir.${FHIR_VERSION}.Web/Microsoft.Health.Fhir.${FHIR_VERSION}.Web.csproj" -c Release -o "/build" --no-build
 
-FROM mcr.microsoft.com/dotnet/core/aspnet:3.1-alpine AS runtime
+FROM mcr.microsoft.com/dotnet/core/aspnet:3.1.6-alpine AS runtime
```
LTA-Thinking, line 12: This doesn't look like it is a supported tag. I don't think we need to change this tag.
https://hub.docker.com/_/microsoft-dotnet-core-aspnet
Linux amd64 tags

feordin: There is a version specific alpine tag, however I do see that it has already moved to 3.1.7.  You are correct that previously we simply had the 3.1 tag, so I reverted this version specific change.

PR #1131, opened 2020-07-07
Auto-update for packages related to 'MediatR'

MicrosoftHealthService:
Updates package 'MediatR' to version '8.0.2'
Updates package 'MediatR.Extensions.Microsoft.DependencyInjection' to version '8.0.1'


LTA-Thinking:
The SQL packages in the common repo need to be updated before this can go though. There is also an even newer version out (8.1.0)

PR #1139, opened 2020-07-10
Add support for _type to Export

LTA-Thinking:
## Description
Add support for _type to the $export operation

## Testing
Existing unit and e2e tests continue to pass and new tests are written to cover the new scenario


src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -186,6 +186,15 @@ private async Task UpdateJobRecordAsync(CancellationToken cancellationToken)
                 queryParametersList.Add(Tuple.Create(KnownQueryParameterNames.ContinuationToken, progress.ContinuationToken));
             }
 
+            if (_exportJobRecord.ExportType == ExportJobType.Patient)
+            {
+                queryParametersList.Add(Tuple.Create(KnownQueryParameterNames.Type, KnownResourceTypes.Patient));
+            }
+            else if (!string.IsNullOrEmpty(_exportJobRecord.ResourceType))
+            {
+                queryParametersList.Add(Tuple.Create(KnownQueryParameterNames.Type, _exportJobRecord.ResourceType));
```
namadabu, line 195: What happens when `_exportJobRecord.ResourceType` contains invalid resource type values? Do we need to do some validation on this input?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ExportDataValidationTests.cs
```diff
@@ -90,6 +90,58 @@ public async Task GivenFhirServer_WhenPatientDataIsExported_ThenExportedDataIsSa
             Assert.True(ValidateDataFromBothSources(dataFromFhirServer, dataFromExport));
         }
 
+        [Fact]
+        public async Task GiveFhirServer_WhenAllObservationDataIsExported_ThenExportedDataIsSameAsDataInFhirServer()
```
namadabu, line 94: Test name says `AllObservation` whereas `_type` param includes Observation and Patient

LTA-Thinking: Woops. I wanted to have the test show that sending multiple types worked, but forgot to update the test name.

src/Microsoft.Health.Fhir.Api/Features/Filters/ValidateExportRequestFilterAttribute.cs
```diff
@@ -32,6 +32,7 @@ public ValidateExportRequestFilterAttribute()
             _supportedQueryParams = new HashSet<string>(StringComparer.Ordinal)
             {
                 KnownQueryParameterNames.Since,
+                KnownQueryParameterNames.Type,
```
namadabu, line 35: nit: We have some UTs in `ValidateExportRequestFilterAttributeTests`. Might be useful to add one for this scenario

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
LTA-Thinking: This is a larger issue with search in general. I've opened a new bug to track it: #1151

src/Microsoft.Health.Fhir.Api/Features/Filters/ValidateExportRequestFilterAttribute.cs
LTA-Thinking: Added

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -186,6 +186,15 @@ private async Task UpdateJobRecordAsync(CancellationToken cancellationToken)
                 queryParametersList.Add(Tuple.Create(KnownQueryParameterNames.ContinuationToken, progress.ContinuationToken));
             }
 
+            if (_exportJobRecord.ExportType == ExportJobType.Patient)
+            {
+                queryParametersList.Add(Tuple.Create(KnownQueryParameterNames.Type, KnownResourceTypes.Patient));
```
namadabu, line 191: There is an assumption in `ProcessProgressChange` that the ContinuationToken is added to the end of the `queryParametersList`. Adding _type at the end will break things. We should either add the _type before the ContinuationToken. Or better use a different data structure to hold the `queryParametersList` before we pass it on to the search service.

LTA-Thinking: Good catch. I replaced the assumption with a loop to find the continuation token if it exists. Since there will never be more than 5 items in the list I don't think this will affect performance.

namadabu: I agree that looping through the parameter list will not affect performance. But I am pretty sure there will be more than 5 items in the list in the future :)

namadabu: Also, I am wondering whether it's possible to have a unit or integration test that could have caught this issue?

LTA-Thinking: Added a new unit test for the scenario

PR #1142, opened 2020-07-11
Removes remaining reference to SqlServerModelInitializer

rotodd:
## Description
`SqlServerModelInitializer.cs` has been removed from the healthcare shared components repo in [PR #50](https://github.com/microsoft/healthcare-shared-components/pull/50). This PR removes a remaining reference to the `SqlServerModelInitializer` class.

## Related issues
Addresses [#AB74855](https://microsofthealth.visualstudio.com/Health/_workitems/edit/74855).

## Testing
All tests pass as before.

PR #1148, opened 2020-07-13
Health Check Code and Nuget Version Updated

v-shaaal:
## Description
- Updated health check code after the shared piece of code was moved to healthcare shared components solution.
- Updated nuget packages to point to the latest version of healthcare shared components solution.
- Fixed an existing bug which was caused because of an update in the shared components code. Corresponding change was however not made in FHIR. This pull request fixes that issue.

## Testing
- All existing tests are passing.
- Health check tests are passing as well.


src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
```diff
@@ -330,7 +329,7 @@ private void ThrowIfNotStarted()
         {
             if (!_started)
             {
-                _logger.LogError($"The {nameof(SqlServerModelInitializer)} instance has not been initialized.");
+                _logger.LogError($"The {nameof(SqlServerFhirModel)} instance has not been initialized.");
```
brandonpollett, line 332: FYI: I think this is also being done in #1142

v-shaaal:
> Where are we calling the `app.UseHealthChecksExtension(new PathString(KnownRoutes.HealthCheck));`?

I missed that change. Will be publishing a new iteration shortly with this change included.

v-shaaal:
> > Where are we calling the `app.UseHealthChecksExtension(new PathString(KnownRoutes.HealthCheck));`?
> 
> I missed that change. Will be publishing a new iteration shortly with this change included.

I have published an iteration with the respective change.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
rotodd: Thanks for pointing this out. I will close #1142

PR #1149, opened 2020-07-13
Build pipeline fixes

brendankowitz:
## Description
- Ensures CosmosDb secrets are in KeyVault before web app starts
- Checks and removes PR keyvault if it has been soft-deleted
- Adds parameter to disable AKS

### DIFF
--- a/build/pr-pipeline.yml
+++ b/build/pr-pipeline.yml
@@ -60,6 +60,7 @@ stages:
 
 - stage: deployAksR4Cosmos
   displayName: 'Deploy R4 Cosmos in AKS'
+  condition: and(succeeded(), eq(variables.enableAks, true))
   dependsOn:
   - DockerBuild
   jobs:
@@ -77,6 +78,7 @@ stages:
 
 - stage: deployAksR4Sql
   displayName: 'Deploy R4 Sql in AKS'
+  condition: and(succeeded(), eq(variables.enableAks, true))
   dependsOn:
   - DockerBuild
   jobs:
@@ -94,6 +96,7 @@ stages:
 
 - stage: deployAksR4SqlContainer
   displayName: 'Deploy R4 SqlContainer in AKS'
+  condition: and(succeeded(), eq(variables.enableAks, true))
   dependsOn:
   - DockerBuild
   jobs:
@@ -216,6 +219,7 @@ stages:
 
 - stage: testAksR4Cosmos
   displayName: 'R4 CosmosDb AKS Tests'
+  condition: and(succeeded(), eq(variables.enableAks, true))
   dependsOn:
   - BuildUnitTests
   - aadTestEnvironment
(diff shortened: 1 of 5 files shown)


brandonpollett:
I'm not sure what the exact issue is here, but you might want to look at https://docs.microsoft.com/en-us/azure/app-service/deploy-resource-manager-template#define-dependencies and make sure our dependencies are correct. I know that applying configuration changes end up restarting the services.

brendankowitz:
I was looking at that too. This seems to work for Webapps, but same problem in AKS, happy with a stop-gap to unblock first

build/pr-variables.yml
```diff
@@ -6,3 +6,4 @@ variables:
     DeploymentEnvironmentName: '$(resourceGroupRoot)-$(prNumber)'
     TestEnvironmentName: 'OSS PR$(prNumber)'
     ImageTag: '$(build.BuildNumber)'
+    enableAks: false
```
brendankowitz, line 9: I'd like to enable this setting from DevOps eventually

samples/templates/default-azuredeploy-docker.json
```diff
@@ -241,9 +238,25 @@
                     "type": "config",
                     "dependsOn": [
                         "[variables('appServiceResourceId')]",
-                        "[if(variables('deployAppInsights'),concat('Microsoft.Insights/components/', variables('appInsightsName')),resourceId('Microsoft.KeyVault/vaults', variables('serviceName')))]"
+                        "[if(variables('deployAppInsights'),concat('Microsoft.Insights/components/', variables('appInsightsName')),resourceId('Microsoft.KeyVault/vaults', variables('serviceName')))]",
+                        "[resourceId('Microsoft.KeyVault/vaults/secrets', variables('serviceName'), 'CosmosDb--Host')]"
```
brendankowitz, line 242: Addresses the issue where the container starts up before the keyvault secrets exist

LTA-Thinking: Does this change also need to be in the non-docker deployment script?

build/pr-pipeline.yml
```diff
@@ -311,8 +317,5 @@ stages:
   - testStu3
   - testR4
   - testR5
-  - testAksR4Cosmos
-  - testAksR4Sql
-  - testAksR4SqlContainer
```
brendankowitz, line 316: These are preventing the normal cleanup step from running when skipped

LTA-Thinking, line 316: If enableAks is true the cleanup still needs to wait on these jobs.

brendankowitz: Yep, don't have a good solution for this yet. They need to be there, but also they cause the entire step to be skipped at the moment...The AKS builds need some more work

samples/templates/default-azuredeploy-docker.json
brendankowitz: Possibly, I haven't done enough testing with the other script yet

build/pr-pipeline.yml
brendankowitz: Will look at something like this: https://elanderson.net/2020/05/azure-devops-pipelines-depends-on-with-conditionals-in-yaml/

PR #1150, opened 2020-07-13
Fixes the sql server sorting by unsupported param code

rbans96:
## Description
Fixes the sorting implementation by unsupported sort search parameter code

## Related issues
Addresses Bug #74734

## Testing
Added unit test and e2e tests


test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/SortTests.cs
```diff
@@ -116,5 +135,26 @@ private void SetPatientInfo(Patient patient, string city, string family, string
 
             patient.Name = new List<HumanName> { new HumanName { Family = family }, };
         }
+
+        private async Task<Observation[]> CreateObservations(string tag)
```
LTA-Thinking, line 139: Could you change the name of the CreateResources and CreatePaginatedResources methods to show that they make Patient resources?

PR #1152, opened 2020-07-14
Re-enabled AKS builds and tests

hansenms:
## Description
Re-enables the AKS builds and tests

## Related issues
Addresses [issue #].

## Testing


brendankowitz:
I'd like to do this in the cleanup step so we can keep the aks flag: https://github.com/microsoft/fhir-server/pull/1149#discussion_r454483108

PR #1154, opened 2020-07-14
Helper method to calculate hash of search parameters

namadabu:
## Description
Added a method that generates a hash given a list of `SearchParameterInfo`. It uses the `Name` property from each object to generate a unique hash for the entire list of `SearchParameterInfo`.

## Related issues
Addresses [issue [AB#73617](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/73617)].

## Testing
Added UTs.

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Search/SearchHelperUtilities.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Search/SearchHelperUtilities.cs
@@ -8,29 +8,30 @@ using System.Linq;
 using System.Text;
 using EnsureThat;
 using Microsoft.Health.Core.Extensions;
-using Microsoft.Health.Fhir.Core.Models;
+using Microsoft.Health.Fhir.Core.Features.Search.Registry;
 
 namespace Microsoft.Health.Fhir.Core.Features.Search
 {
     public static class SearchHelperUtilities
     {
         /// <summary>
-        /// Given a list of <see cref="SearchParameterInfo"/> calculates a hash using the
-        /// <see cref="SearchParameterInfo.Name"/> values of each component.
-        /// The same collection of search parameters (irrespective of their order in the input)
+        /// Given a list of <see cref="ResourceSearchParameterStatus"/> calculates a hash using the
+        /// <see cref="ResourceSearchParameterStatus.Uri"/> and <see cref="ResourceSearchParameterStatus.LastUpdated"/>
+        /// values of each component. The same collection of search parameter status (irrespective of their order in the input)
         /// will return the same hash.
         /// </summary>
-        /// <param name="searchParamInfo">A list of <see cref="SearchSearchParameterInfo" /></param>
-        /// <returns>A hash based on the search parameter names present in the input.</returns>
-        public static string CalculateSearchParameterNameHash(IEnumerable<SearchParameterInfo> searchParamInfo)
+        /// <param name="resourceSearchParameterStatus">A list of <see cref="ResourceSearchParameterStatus" /></param>
+        /// <returns>A hash based on the search parameter uri and last updated value.</returns>
+        public static string CalculateSearchParameterNameHash(IEnumerable<ResourceSearchParameterStatus> resourceSearchParameterStatus)
         {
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Fhir.Core/Features/Search/SearchHelperUtilities.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// </summary>
+        /// <param name="searchParamInfo">A list of <see cref="SearchSearchParameterInfo" /></param>
+        /// <returns>A hash based on the search parameter names present in the input.</returns>
+        public static string CalculateSearchParameterNameHash(IEnumerable<SearchParameterInfo> searchParamInfo)
+        {
+            EnsureArg.IsNotNull(searchParamInfo, nameof(searchParamInfo));
+            EnsureArg.IsGt(searchParamInfo.Count(), 0, nameof(searchParamInfo));
+
+            StringBuilder sb = new StringBuilder();
+            foreach (string paramName in searchParamInfo.Select(x => x.Name).OrderBy(x => x))
```
brendankowitz, line 31: In other places we are tracking SearchParameters by Uri which is more unique and relevant across resource types

src/Microsoft.Health.Fhir.Core/Features/Search/SearchHelperUtilities.cs
```diff
@@ -0,0 +1,40 @@
...
+            {
+                sb.Append(paramName);
+            }
+
+            string hash = sb.ToString().ComputeHash();
```
brendankowitz, line 36: When custom search is enabled, do we want to know if Components, Expressions, Base/Target types have changed by looking at the hash? @feordin thoughts?

src/Microsoft.Health.Fhir.Core/Features/Search/SearchHelperUtilities.cs
feordin: Probably uri would be more unique, and I think that is a good idea.  I think we should include also the lastUpdate timestamp.  That way if any search param changes, the hash would change.

src/Microsoft.Health.Fhir.Core/Features/Search/SearchHelperUtilities.cs
feordin: I think the hash should represent a snapshot of the current parameters.  If anything is changed, the hash should be new.  This will mean that we may unnecessarily query and read resources for which no index needs to be updated, but it ensures a simple implementation.  The hash value represents a given set of parameters, if anything changes we get a new hash.

When we run the reindex job, we do identify only the resources types that are affected by any new search param which has not been indexed, so a new hash value would not necessary cause all resources to reindexed.

src/Microsoft.Health.Fhir.Core/Features/Search/SearchHelperUtilities.cs
namadabu: @brendankowitz  I went with `Name` because that seemed to be the only "required" parameter when creating a `SearchParameterInfo` object. The default value for the `url` in the `SearchParameterInfo` constructor is set to `null`. Practically, would it be safe to assume that url will always be non-null?
@feordin `SearchParameterInfo` does not track the lastUpdated value. For that we will have to depend on the actual `__searchparameterstatus__` documents. Would it be better to support that as input type instead of dealing with `SearchParamterInfo`?

feordin: With lastupdated not easily available, I think we should go with @brendankowitz suggestion of including the components, expression and Base/Target types.

namadabu: Changed the code to use `ResourceSearchParameterStatus` instead of `SearchParameterInfo`. This has both the uri as well as the last updated data that is required. We build a wrapper around `ResourceSearchParameterStatus` before pushing it to the data layer (thanks for pointing this out @rotodd ). So it seems like the correct class to use for our use case.

PR #1163, opened 2020-07-20
Support Group Export

LTA-Thinking:
## Description
Add support for exporting Groups of Patients.
[Group export spec](https://hl7.org/fhir/uv/bulkdata/export/index.html#endpoint---group-of-patients)

Only direct members of a Group are exported. Implicit members are ignored.

## Testing
New and existing unit and E2E tests


src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Export/GroupMemberExtractor.cs
```diff
@@ -0,0 +1,65 @@
...
+        public async Task<List<Tuple<string, string>>> GetGroupMembers(string groupId, CancellationToken cancellationToken)
+        {
+            var groupResource = await _fhirDataStore.Value.GetAsync(new ResourceKey(KnownResourceTypes.Group, groupId), cancellationToken);
+
+            var group = _resourceDeserializer.Deserialize(groupResource);
+            var groupContents = group.ToPoco<Group>().Member;
+
+            var members = new List<Tuple<string, string>>();
+
+            foreach (Group.MemberComponent member in groupContents)
```
namadabu, line 53: From my limited understanding of group, what happens if the `Group` resource does not define any members and only has `characteristic` defined? Do we have to handle that scenario by searching for resources based on the `characteristic` value?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -203,15 +208,25 @@ private async Task UpdateJobRecordAsync(CancellationToken cancellationToken)
...
+                        }
+
+                        break;
+                    case ExportJobType.Group:
+                        searchResult = await GetGroupPatients(_exportJobRecord.GroupId, queryParametersList, cancellationToken);
+                        break;
                 }
 
-                if (_exportJobRecord.ExportType == ExportJobType.Patient)
+                if (_exportJobRecord.ExportType != ExportJobType.All)
```
namadabu, line 229: nit: Might be better to check specifically for `Patient || Group`. In the future if we add or change an `ExportJobType` type and miss updating this condition, it might lead to unexpected behavior.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -390,5 +405,60 @@ private async Task ProcessSearchResultsAsync(IEnumerable<SearchResultEntry> sear
...
+        {
+            if (!queryParametersList.Exists((Tuple<string, string> parameter) => parameter.Item1 == KnownQueryParameterNames.Id))
+            {
+                var patientIds = await GetGroupPatientIds(groupId, cancellationToken);
+                queryParametersList.Add(Tuple.Create(KnownQueryParameterNames.Id, string.Join(',', patientIds)));
+            }
+
+            using (IScoped<ISearchService> searchService = _searchServiceFactory())
+            {
+                return await searchService.Value.SearchAsync(
```
namadabu, line 419: What happens here if we have more than 100 patient ids as part of the `queryParametersList`?

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Export/GroupMemberExtractor.cs
LTA-Thinking: Good catch. This is an overlook of mine, I was only considering direct group members. I'll look into this.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -390,5 +405,60 @@ private async Task ProcessSearchResultsAsync(IEnumerable<SearchResultEntry> sear
...
+        private async Task<HashSet<string>> GetGroupPatientIds(string groupId, CancellationToken cancellationToken, HashSet<string> groupsAlreadyChecked = null)
+        {
+            // need to send in the _before param so that if this restarted the same version of the group is retrieved
+
+            if (groupsAlreadyChecked == null)
+            {
+                groupsAlreadyChecked = new HashSet<string>();
+            }
+
+            groupsAlreadyChecked.Add(groupId);
```
namadabu, line 435: I am wondering whether groups can contain other groups? Looking at the Group.type definition (https://www.hl7.org/fhir/valueset-group-type.html) seems like it can contain only 6 values.

LTA-Thinking: From the reference types allowed for the group entities it seems like more is allowed than what the valueset implies. https://www.hl7.org/fhir/group.html

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -203,15 +208,25 @@ private async Task UpdateJobRecordAsync(CancellationToken cancellationToken)
...
+                        {
+                            searchResult = await searchService.Value.SearchAsync(
+                                resourceType: null,
+                                queryParametersList,
+                                cancellationToken);
+                        }
+
+                        break;
+                    case ExportJobType.Group:
+                        searchResult = await GetGroupPatients(_exportJobRecord.GroupId, queryParametersList, cancellationToken);
```
namadabu, line 225: Once we have a group of patients and we run compartment search on each of them. If some of these patients share some common resources, say a `Practitioner`, we will end up exporting duplicates of that resource.

LTA-Thinking: There is a separate story for dealing with duplicate resources [74403](https://microsofthealth.visualstudio.com/Health/_workitems/edit/74403)

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
namadabu: Aah. I see what you are referring to. Missed that

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Export/GroupMemberExtractor.cs
LTA-Thinking: Added to https://microsofthealth.visualstudio.com/Health/_workitems/edit/74401

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -390,5 +409,58 @@ private async Task ProcessSearchResultsAsync(IEnumerable<SearchResultEntry> sear
...
+
+            var patientIds = new HashSet<string>();
+
+            foreach (Tuple<string, string> entity in groupContents)
+            {
+                var resourceId = entity.Item1;
+                var resourceType = entity.Item2;
+                switch (resourceType)
+                {
+                    case KnownResourceTypes.Patient:
```
namadabu, line 449: nit: we should add a comment here explaining why we consider only Patient and Group type and skip the resource types.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -390,5 +409,58 @@ private async Task ProcessSearchResultsAsync(IEnumerable<SearchResultEntry> sear
                 await UpdateJobRecordAsync(cancellationToken);
             }
         }
+
+        private async Task<SearchResult> GetGroupPatients(string groupId, List<Tuple<string, string>> queryParametersList, DateTimeOffset groupMembershipTime, CancellationToken cancellationToken)
+        {
+            if (!queryParametersList.Exists((Tuple<string, string> parameter) => parameter.Item1 == KnownQueryParameterNames.Id))
```
namadabu, line 415: Should we also check whether there is a continuationtoken already present? If so, then we don't need to add the id's again right?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -390,5 +409,58 @@ private async Task ProcessSearchResultsAsync(IEnumerable<SearchResultEntry> sear
                 await UpdateJobRecordAsync(cancellationToken);
             }
         }
+
+        private async Task<SearchResult> GetGroupPatients(string groupId, List<Tuple<string, string>> queryParametersList, DateTimeOffset groupMembershipTime, CancellationToken cancellationToken)
```
namadabu, line 413: nit: Seems like both of these methods can move to the GroupMemberExtractor class?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -203,15 +208,29 @@ private async Task UpdateJobRecordAsync(CancellationToken cancellationToken)
...
+                        searchResult = await GetGroupPatients(
+                            _exportJobRecord.GroupId,
+                            queryParametersList,
+                            _exportJobRecord.QueuedTime,
+                            cancellationToken);
+                        break;
                 }
 
-                if (_exportJobRecord.ExportType == ExportJobType.Patient)
+                if (_exportJobRecord.ExportType == ExportJobType.Patient || _exportJobRecord.ExportType == ExportJobType.Group)
```
namadabu, line 233: In the below code, when we call `ProcessProgressChange` do we need to update the `forceCommit` flag to be set to true for `Group` export also?

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Export/GroupMemberExtractor.cs
namadabu: Do you think it might be better to keep group export disabled until we support the above scenario also?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
LTA-Thinking: GetGroupPatientIds I agree would be good in the GroupMemberExtractor class. GetGroupPatients I think is too specialized for this export use case and needs the search service which isn't in the extractor class currently.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
LTA-Thinking: I didn't realize all that information was stored in the continuation token. Good idea.

LTA-Thinking:
Group export does not include implicit group members #1179

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Export/GroupMemberExtractor.cs
```diff
@@ -0,0 +1,121 @@
...
+using Microsoft.Health.Fhir.Core.Exceptions;
+using Microsoft.Health.Fhir.Core.Extensions;
+using Microsoft.Health.Fhir.Core.Features.Persistence;
+using Microsoft.Health.Fhir.Core.Features.Search;
+using Microsoft.Health.Fhir.Core.Models;
+
+namespace Microsoft.Health.Fhir.Core.Features.Operations.Export
+{
+    /// <summary>
+    /// Class to get member ids and types out of a group. Split between common and version specifc code due to the change between Stu3 and R4 to the ResourceReference object.
```
namadabu, line 23: Second part of the comment is no longer valid

src/Microsoft.Health.Fhir.Core/Features/Conformance/Serialization/EnumLiteralJsonConverter.cs
```diff
@@ -21,7 +21,17 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
 
             if (value is Enum obj)
             {
-                serializer.Serialize(writer, obj.GetLiteral());
+                string enumString;
```
namadabu, line 24: Would be helpful to add a comment on why this change is being made. Will make it easier in the future if we have to deal with something similar or have to modify this behavior again.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/IGroupMemberExtractor.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// <returns>List of Tuples for the members of the Group. The tuples' first values are the resource ids and the second values are the resource type.</returns>
+        public Task<List<Tuple<string, string>>> GetGroupMembers(string groupId, DateTimeOffset groupMembershipTime, CancellationToken cancellationToken, bool includeInactiveMembers = false);
+
+        /// <summary>
+        /// Gets a set of Patient ids that are included in a Group, either as members or as members of nested Groups. Implicit members of a group are not included.
+        /// </summary>
+        /// <param name="groupId">The id of the Group</param>
+        /// <param name="groupMembershipTime">Only returns Patients that were in the Group at this time</param>
+        /// <param name="cancellationToken">Cancellation token</param>
+        /// <param name="groupsAlreadyChecked">Internal use</param>
```
namadabu, line 31: If this param is for internal use then I don't think it should be part of the interface (since the caller is never going to pass a value for that). It seems like we are tying the implementation details to the interface. My suggestion would be to remove this from the interface and have the implementation call a helper method which passes in the appropriate value for this param.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ExportDataValidationTests.cs
```diff
@@ -357,5 +411,73 @@ private CloudStorageAccount GetCloudStorageAccountHelper(string storageAccountNa
 
             return cloudAccount;
         }
+
+        private async Task<(Dictionary<(string resourceType, string resourceId), string> serverData, string groupId)> CreateGroupWithPatient(bool includeAllCompartmentResources)
+        {
+            // Add data for test
```
namadabu, line 417: nit: another option would be add a json file with the sample data mentioned below and then read that in before sending to the fhir-server. We use this method in other places. The files are located in Microsoft.Health.Fhir.Tests.Common . I don't have a preference for either method. Just wanted to call that out as an option.

LTA-Thinking:
Added issue for enum mapping questions: #1181

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ExportDataValidationTests.cs
LTA-Thinking: For this test I think it works better to have it in the test file. I could do this with two JSON files and a batch, but I think it works better for the minimal resources this test needs to have it in the test file.

PR #1168, opened 2020-07-22
Update Reindex task to process queries and save updates to reindex doc

feordin:
## Description
Updates the ReindexTask to read the resources, identify the search params, and then run the query against the data store.
When a query finishes, the reindex doc is updated again to reflect the finished query.

## Related issues
Addresses [issue [AB#73615](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/73615)].

## Testing
Unit tests for reindexjobtask added.  Manual E2E tests done.


src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterStatusManager.cs
```diff
@@ -88,6 +88,9 @@ public async Task EnsureInitialized()
                 }
             }
 
+            var searchParameterHashValue = SearchHelperUtilities.CalculateSearchParameterHash(parameters.Values);
```
namadabu, line 91: I am not familiar with this class/code, but seems like `parameters` can contain search params that are `Disabled`. Do we want to include those also in the has calculation? And do we need to consider values present in `updated`?

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/Queries/QueryBuilder.cs
```diff
@@ -129,6 +134,41 @@ public QueryDefinition GenerateHistorySql(SearchOptions searchOptions)
...
+                    _queryBuilder,
+                    _queryParameterManager);
+
+                if (searchOptions.Expression != null)
+                {
+                    _queryBuilder.Append("AND ");
+                    searchOptions.Expression.AcceptVisitor(expressionQueryBuilder);
+                }
+
+                AppendFilterCondition(
```
namadabu, line 162: Should we also filter for `IsDeleted` and `IsHistory` in this query?

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -85,12 +113,41 @@ public async Task ExecuteAsync(ReindexJobRecord reindexJobRecord, WeakETag weakE
...
-                await CompleteJobAsync(OperationStatus.Completed, cancellationToken);
+                    // generate and run first query
+                    var queryStatus = new ReindexJobQueryStatus(null);
+                    queryStatus.LastModified = DateTimeOffset.UtcNow;
+                    queryStatus.Status = OperationStatus.Queued;
+
+                    _reindexJobRecord.QueryList.Add(queryStatus);
+
+                    // update the complete total
+                    var countOnlyResults = await ExecuteReindexQueryAsync(queryStatus, true, cancellationToken);
```
namadabu, line 125: nit: please call out the param name explicitly for non-descriptive values (like null, true, etc). Makes it easier to read.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -99,28 +156,147 @@ public async Task ExecuteAsync(ReindexJobRecord reindexJobRecord, WeakETag weakE
...
+            if (queryStatus.ContinuationToken != null)
+            {
+                queryParametersList.Add(Tuple.Create(KnownQueryParameterNames.ContinuationToken, queryStatus.ContinuationToken));
+            }
+
+            using (IScoped<ISearchService> searchService = _searchServiceFactory())
+            {
+                try
+                {
+                    return await searchService.Value.SearchForReindexAsync(queryParametersList, _reindexJobRecord.Hash, countOnly, cancellationToken);
```
namadabu, line 272: Just to confirm, we are searching for resources that have the "existing/old" search parameter hash?

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -99,28 +156,147 @@ public async Task ExecuteAsync(ReindexJobRecord reindexJobRecord, WeakETag weakE
...
+                    nextQuery.LastModified = DateTimeOffset.UtcNow;
+                    nextQuery.Status = OperationStatus.Queued;
+                    _reindexJobRecord.QueryList.Add(nextQuery);
+                }
+
+                await UpdateJobAsync(cancellationToken);
+
+                // TODO: Release lock on job document so another thread may pick up the next query.
+
+                await _updateIndices.ProcessSearchResultsAsync(results, _reindexJobRecord.Hash, cancellationToken);
```
namadabu, line 200: Should this be the new hash or the old hash here?

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -99,28 +156,147 @@ public async Task ExecuteAsync(ReindexJobRecord reindexJobRecord, WeakETag weakE
...
+
+                await CheckJobCompletionStatus(cancellationToken);
+            }
+            catch (Exception ex)
+            {
+                query.Error = ex.Message;
+
+                query.FailureCount++;
+
+                _logger.LogError(ex, $"Encountered an unhandled exception. The qquery failure count increased to {_reindexJobRecord.FailureCount}.");
```
namadabu, line 215: nit: typo, qquery

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -99,28 +156,147 @@ public async Task ExecuteAsync(ReindexJobRecord reindexJobRecord, WeakETag weakE
...
+                // if continuation token then update next query
+                if (!string.IsNullOrEmpty(results.ContinuationToken))
+                {
+                    var nextQuery = new ReindexJobQueryStatus(results.ContinuationToken);
+                    nextQuery.LastModified = DateTimeOffset.UtcNow;
+                    nextQuery.Status = OperationStatus.Queued;
+                    _reindexJobRecord.QueryList.Add(nextQuery);
+                }
+
+                await UpdateJobAsync(cancellationToken);
```
namadabu, line 196: When we update the job here, the query status is set to `Running` and that information will be saved as part of the job record. Let's assume the process restarts here and we have to resume our re-index job. This query has not been completed yet. But we will never pick this up again since we are looking only for "Queued" queries in the `ExecuteAsync` code

src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterStatusManager.cs
feordin: This is a good point.  I am not sure I have the correct answer right now :).  I created a new work item: 75058, to track this and update it.

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/Queries/QueryBuilder.cs
feordin: I want to include the history, as we need to reindex the history as well, but I did miss the IsDeleted.  Thanks!

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
feordin: Good catch!  Looking at the code more carefully with this in mind, I think we have a similar problem in other places if the process is shutdown before we are able to complete some processing.  We do have a separate task: 73390 which is focused on how to handle a restart.  I think we will need to check timestamps on the queries and the reindexjob to see if they have become stale and then reset the status.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
feordin: This should be the new/current hash value.  I have updated the search logic to be correct.

src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterDefinitionManager.cs
```diff
@@ -6,17 +6,21 @@
...
+using Microsoft.Health.Fhir.Core.Messages.Search;
 using Microsoft.Health.Fhir.Core.Models;
 
 namespace Microsoft.Health.Fhir.Core.Features.Definition
 {
     /// <summary>
     /// Provides mechanism to access search parameter definition.
     /// </summary>
-    public class SearchParameterDefinitionManager : ISearchParameterDefinitionManager, IStartable
+    public class SearchParameterDefinitionManager : ISearchParameterDefinitionManager, IStartable, INotificationHandler<SearchParametersHashUpdated>
```
brendankowitz, line 23: You will need to be careful here as `INotificationHandler`s are registered as Transient by default

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/Models/ReindexJobQueryStatus.cs
```diff
@@ -13,8 +13,18 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Reindex.Models
...
+        {
+            ContinuationToken = continuationToken;
+        }
+
+        [JsonConstructor]
+        protected ReindexJobQueryStatus()
+        {
+        }
+
         [JsonProperty(JobRecordProperties.Query)]
```
LTA-Thinking, line 26: Should the JSON property name be updated to match?

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -72,6 +88,18 @@ public async Task ExecuteAsync(ReindexJobRecord reindexJobRecord, WeakETag weakE
...
 
+                    // if there are not any parameters which are supported but not yet indexed, then we have nothing to do
+                    if (!notYetIndexedParams.Any())
+                    {
+                        _reindexJobRecord.Error.Add(new OperationOutcomeIssue(
+                            OperationOutcomeConstants.IssueSeverity.Information,
+                            OperationOutcomeConstants.IssueType.Informational,
+                            Resources.NoSearchParametersNeededToBeIndexed));
+                        _reindexJobRecord.CanceledTime = DateTimeOffset.UtcNow;
+                        await CompleteJobAsync(OperationStatus.Canceled, cancellationToken);
```
LTA-Thinking, line 99: This seems like it would be a completed outcome. There were no parameters to reindex, so all of the reindexing got done.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -85,12 +113,41 @@ public async Task ExecuteAsync(ReindexJobRecord reindexJobRecord, WeakETag weakE
...
+                    // check to see if queries are queued
+                    // TODO: this while loop is temporary until we multithread this task so multiple threads can
+                    // processes queries
+                    while (_reindexJobRecord.QueryList.Where(q => q.Status == OperationStatus.Queued).Any())
+                    {
+                        // grab the next query from the list which is labeled as queued and run it
+                        var query = _reindexJobRecord.QueryList.Where(q => q.Status == OperationStatus.Queued).OrderBy(q => q.LastModified).FirstOrDefault();
 
-                _logger.LogTrace("Successfully completed the job.");
+                        if (query != null)
```
LTA-Thinking, line 141: Why is the if statement needed? It looks like the while loop would require query to not be null.

src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterDefinitionManager.cs
feordin: This was my problem with SearchParameterDefinitionManager getting multiple instances.  I have updated this.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
feordin: You are correct.  I removed it.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
feordin: I marked it as canceled because no changes occurred in the server.  We did not update the hash value on any resources, or touch anything else.  I do see your point that there were no errors and it that sense it completed by checking if any work needed to be done.  I am open to changing it to complete.

src/Microsoft.Health.Fhir.Core/Features/Operations/JobRecordProperties.cs
```diff
@@ -91,6 +91,8 @@ public static class JobRecordProperties
 
         public const string SearchParams = "searchParams";
 
+        public const string ContinuationToken = "ContinuationToken";
```
namadabu, line 94: Should the string follow camel case format?

src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParametersHashUpdatedHandler.cs
```diff
@@ -0,0 +1,30 @@
...
+
+using System.Threading;
+using System.Threading.Tasks;
+using EnsureThat;
+using MediatR;
+using Microsoft.Health.Fhir.Core.Messages.Search;
+
+namespace Microsoft.Health.Fhir.Core.Features.Definition
+{
+    public class SearchParametersHashUpdatedHandler : INotificationHandler<SearchParametersHashUpdated>
```
namadabu, line 14: nit: This class seems out of place in this Features.Definition namespace. Would it be better to move it to a location where other handlers are present?

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
```diff
@@ -88,6 +88,17 @@ public class FhirCosmosSearchService : SearchService
                 cancellationToken);
         }
 
+        protected override async Task<SearchResult> SearchForReindexInternalAsync(
+            SearchOptions searchOptions,
+            string searchParameterHash,
```
namadabu, line 93: nit: do we need param validation for searchParameterHash?

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Reindex/ReindexUtilities.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// Then compare those to the old values to determine if a change is needed
+        /// Needed updates will br processed in a batch
+        /// </summary>
+        /// <param name="results">The resource batch to process</param>
+        /// <param name="searchParamHash">the current hash value of the search parameters</param>
+        /// <param name="cancellationToken">Cancellation token</param>
+        /// <returns>A Task</returns>
+        public async Task ProcessSearchResultsAsync(SearchResult results, string searchParamHash, CancellationToken cancellationToken)
+        {
+            var updatedResources = new List<SearchResultEntry>();
```
namadabu, line 48: nit: this doesn't seem to be used

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/SearchServiceTests.cs
```diff
@@ -146,6 +146,11 @@ public TestSearchService(ISearchOptionsFactory searchOptionsFactory, IFhirDataSt
             {
                 return Task.FromResult(SearchImplementation(searchOptions));
             }
+
+            protected override Task<SearchResult> SearchForReindexInternalAsync(SearchOptions searchOptions, string searchParameterHash, CancellationToken cancellationToken)
```
namadabu, line 150: do we need to add any UTs for this new method?

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexUtilities.cs
```diff
@@ -0,0 +1,47 @@
...
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Health.Extensions.DependencyInjection;
+using Microsoft.Health.Fhir.Core.Features.Persistence;
+using Microsoft.Health.Fhir.Core.Features.Search;
+
+namespace Microsoft.Health.Fhir.Core.Features.Operations.Reindex
+{
+    public class ReindexUtilities
```
namadabu, line 15: There is another class called ReindexUtilities under `Shared.Core`. I am guessing one of them needs to be deleted?

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexUtilities.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        public ReindexUtilities(Func<IScoped<IFhirDataStore>> fhirDataStoreFactory)
+        {
+            _fhirDataStoreFactory = fhirDataStoreFactory;
+        }
+
+        /// <summary>
+        /// For each result in a batch of resources this will extract new search params
+        /// Then compare those to the old values to determine if a change is needed
+        /// Needed updates will br processed in a batch
```
namadabu, line 27: nit: typo - be

src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParametersHashUpdatedHandler.cs
feordin: I do see the point, we are not entirely consistent with the location of handlers.  Some other features, Compartment, Conformance, Export also contain Handlers in the feature namespace, I believe with the intent of keeping all classes relevant to a feature in one location.  And as you said, some handlers are collected together.  I do not feel strongly either way.

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
feordin: Down the call chain, I have updated the QueryBuilder.cs to check for null.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexUtilities.cs
feordin: Indeed!  Removed it.

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/SearchServiceTests.cs
feordin: I added an additional unit test.

PR #1169, opened 2020-07-23
Combines search parameter SQL tables

rotodd:
## Description

This PR merges the search parameter status registry with the existing `dbo.SearchParam` table.

Combined table:
image

Note: I'd like to do some renaming. I saved it for [#AB74985](https://microsofthealth.visualstudio.com/Health/_workitems/edit/74985).

## Related issues
Addresses [#AB74802](https://microsofthealth.visualstudio.com/Health/_workitems/edit/74802).

## Testing
Search parameter status tests pass as before.


src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/4.diff.sql
```diff
@@ -1,39 +1,31 @@
...
-(
-    Uri varchar(128) COLLATE Latin1_General_100_CS_AS NOT NULL,
-    Status varchar(10) NOT NULL,
-    IsPartiallySupported bit NOT NULL
-)
+ALTER TABLE dbo.SearchParam
+ADD
+    Status varchar(10) NULL,
+    LastUpdated datetimeoffset(7) NULL,
+    IsPartiallySupported bit NULL
```
rotodd, line 9: This is different to the change made to 4.sql - in that file, the table definition is simply modified.

rotodd: These values are set to be `NULL` until they are initialized in `SqlServerFhirModel`.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
```diff
@@ -175,13 +175,14 @@ BEGIN TRANSACTION
...
                         BEGIN
                             DECLARE @lastUpdated datetimeoffset(7) = SYSDATETIMEOFFSET()
         
-                            INSERT INTO dbo.SearchParamStatusRegistry (Uri, Status, LastUpdated, IsPartiallySupported)
-                            SELECT sps.Uri, sps.Status, @lastUpdated, sps.IsPartiallySupported
-                            FROM @searchParamStatuses AS sps
+                            UPDATE dbo.SearchParam
+                            SET Status = sps.Status, LastUpdated = @lastUpdated, IsPartiallySupported = sps.IsPartiallySupported
+                            FROM dbo.SearchParam INNER JOIN @searchParamStatuses as sps
+                            ON dbo.SearchParam.Uri = sps.Uri
```
rotodd, line 185: @johnstairs is there a nicer way to do this?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/4.sql
```diff
@@ -2154,41 +2157,27 @@ END
 GO
 
 /*************************************************************
-    Search Parameter Status Registry
+    Search Parameter Status Information
 **************************************************************/
 
-CREATE TYPE dbo.SearchParamStatusRegistryTableType_1 AS TABLE
+CREATE TYPE dbo.SearchParamTableType_1 AS TABLE
```
johnstairs, line 2163: What is the _1 suffix for?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SqlServerStatusRegistryDataStore.cs
```diff
@@ -21,46 +21,46 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Storage.Registry
...
+                            VLatest.SearchParam.IsPartiallySupported);
 
                         var status = Enum.Parse<SearchParameterStatus>(stringStatus, true);
 
                         var resourceSearchParameterStatus = new ResourceSearchParameterStatus()
                         {
                             Uri = new Uri(uri),
                             Status = status,
-                            IsPartiallySupported = isPartiallySupported,
+                            IsPartiallySupported = (bool)isPartiallySupported,
```
johnstairs, line 63: What if this is null?

johnstairs:
In CreateSqlDataRecords(), can you not use the generated TVP types?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
johnstairs: I'm not up to date on the design, but you want to update these if none of the statuses have been set?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/4.sql
rotodd: I admit I'm not super sure. Other `CREATE TYPE` statements for table valued parameters are followed by a `_1` - that being said, I can't seem to find anything about that in the [style guide](https://github.com/ktaranov/sqlserver-kit/blob/master/SQL%20Server%20Name%20Convention%20and%20T-SQL%20Programming%20Style.md).

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
rotodd: You're right. I don't think these values should ever be `null`. I made changes in 4bf9d07 to ensure this.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SqlServerStatusRegistryDataStore.cs
rotodd: Fixed in 4bf9d07.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/4.sql
johnstairs: Oh I remember now. It's because the table types are immutable and can't be altered.

rotodd:
> In CreateSqlDataRecords(), can you not use the generated TVP types? @johnstairs 

After digging into this a bit, it looks like this might require an access-level change in the healthcare shared components repo. It also seems like setting things up to achieve this might end up being less elegant than the current solution. 

I would like to look into this a bit more later, so I created [#AB75189](https://microsofthealth.visualstudio.com/Health/_workitems/edit/75189).

johnstairs:
I'm getting an error on startup trying this out (`System.InvalidOperationException: Cannot consume scoped service 'Microsoft.Health.SqlServer.Features.Client.SqlConnectionWrapperFactory' from singleton 'Microsoft.Health.Fhir.SqlServer.Features.Storage.Registry.SqlServerStatusRegistryDataStore'.`). But otherwise it looks good.

rotodd:
> I'm getting an error on startup trying this out (`System.InvalidOperationException: Cannot consume scoped service 'Microsoft.Health.SqlServer.Features.Client.SqlConnectionWrapperFactory' from singleton 'Microsoft.Health.Fhir.SqlServer.Features.Storage.Registry.SqlServerStatusRegistryDataStore'.`). But otherwise it looks good. @johnstairs 

Ah yes good catch. I have a fix in the works for this with [#AB75037](https://microsofthealth.visualstudio.com/Health/_workitems/edit/75037).

PR #1178, opened 2020-07-31
Perf improvements for get/search calls

YazanMSFT:
## Description

- VersionId and LastUpdated will be set on POSTs
- GET Resource calls will return the raw resource if VersionId and LastUpdated were already set in the raw resource. Otherwise, it will set them using System.Text.Json
- Search calls will behave similar to GET calls. Bundles will be serialized using System.Text.Json as well.

Working on updating SQL persistence layer with the new columns that specify if VersionId and LastUpdated were set on the raw resource.
## Related issues
Addresses [issue #].

## Testing
All existing tests pass. A few tests added - will add more


src/Microsoft.Health.Fhir.Core/Features/Persistence/ResourceDeserializer.cs
```diff
@@ -38,6 +40,88 @@ public ResourceElement Deserialize(ResourceWrapper resourceWrapper)
...
+        /// If the RawResource's VersionSet or LastUpdatedSet are false, then the RawResource's data will be updated
+        /// to have them set to the values in the ResourceWrapper
+        /// </summary>
+        /// <param name="resourceWrapper">Input ResourceWrapper to convert to a JsonDocument</param>
+        /// <returns>A <see cref="JsonDocument"/></returns>
+        public static JsonDocument DeserializeToJsonDocument(ResourceWrapper resourceWrapper)
+        {
+            EnsureArg.IsNotNull(resourceWrapper, nameof(resourceWrapper));
+
+            if (resourceWrapper.RawResource.LastUpdatedSet && resourceWrapper.RawResource.VersionSet)
```
deepak-ms, line 54: Is it a valid assumption that a resource is guaranteed to have versionId/lastUpdated if it was created/modified after some timestamp T (can be in near future)? If yes, is it possible/acceptable to use T (global) for decision here instead of per document flags?

src/Microsoft.Health.Fhir.Core/Features/Persistence/RawResource.cs
```diff
@@ -26,10 +28,16 @@ protected RawResource()
...
 
         [JsonProperty("format")]
         [JsonConverter(typeof(StringEnumConverter))]
         public FhirResourceFormat Format { get; protected set; }
+
+        [JsonProperty("versionSet")]
+        public bool VersionSet { get; protected set; }
+
+        [JsonProperty("lastUpdatedSet")]
+        public bool LastUpdatedSet { get; protected set; }
```
deepak-ms, line 41: If these flags will always have same value (either both true or both false), should we merge these into one?

src/Microsoft.Health.Fhir.Core/Features/Persistence/ResourceDeserializer.cs
johnstairs: That's an interesting idea. But then we would need to block writes from an earlier version of the code after that time.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleSerializer.cs
```diff
@@ -0,0 +1,100 @@
...
+            using (Utf8JsonWriter writer = new Utf8JsonWriter(outputStream, WriterOptions))
+            {
+                writer.WriteStartObject();
+
+                writer.WriteString("resourceType", bundle.ResourceType.ToString());
+                writer.WriteString("id", bundle.Id);
+
+                SerializeMetadata();
+
+                writer.WriteString("type", bundle.Type?.ToCamelCaseString());
```
brendankowitz, line 30: All the provided types have the FHIR value in EnumLiteralAttribute. It's not always true that the FHIR value will be camelcase.

src/Microsoft.Health.Fhir.Shared.Api/Features/Headers/RawFhirResultExtensions.cs
```diff
@@ -0,0 +1,67 @@
...
+using Microsoft.Health.Fhir.Api.Features.ActionResults;
+using Microsoft.Health.Fhir.Core.Features.Persistence;
+using Microsoft.Health.Fhir.Core.Features.Routing;
+using Microsoft.Net.Http.Headers;
+
+namespace Microsoft.Health.Fhir.Api.Features.Headers
+{
+    public static class RawFhirResultExtensions
+    {
+        public static ResourceWrapperFhirResult SetLocationHeader(this ResourceWrapperFhirResult rawFhirResult, IUrlResolver urlResolver)
```
brendankowitz, line 16: In general I don't really like having both ResourceWrapper and ResourceElement being passed around. ResourceWrapper was supposed to be more of a datalayer concept. I've often gone back and forward on the idea that ResourceElement should have had an interface, but I still feel that it is currently the Domain object for FHIR resource data throughout the pipeline

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Upsert/UpsertResourceHandler.cs
```diff
@@ -61,7 +61,7 @@ public async Task<UpsertResourceResponse> Handle(UpsertResourceRequest message,
             bool allowCreate = await ConformanceProvider.Value.CanUpdateCreate(resource.TypeName, cancellationToken);
             bool keepHistory = await ConformanceProvider.Value.CanKeepHistory(resource.TypeName, cancellationToken);
 
-            ResourceWrapper resourceWrapper = CreateResourceWrapper(resource, deleted: false);
+            ResourceWrapper resourceWrapper = CreateResourceWrapper(resource, deleted: false, keepMeta: false);
```
brendankowitz, line 64: In CosmosDb the new version is determined in the Stored Proc, there will likely need to be some updates here too: https://github.com/microsoft/fhir-server/blob/master/src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/StoredProcedures/Upsert/upsertWithHistory.js#L98

src/Microsoft.Health.Fhir.Core/Features/Persistence/ResourceDeserializer.cs
deepak-ms: Right, we can select T with enough buffer to account for deployment/rollback.

johnstairs: Keep in mind that we are not the only ones running this as a service. So such a solution would need to be more deterministic, where we we block usage based on an earlier version of the code

src/Microsoft.Health.Fhir.Core/Features/Persistence/RawResource.cs
johnstairs, line 41: LastUpdated should already be set

YazanMSFT: Isn't it cleared out here: https://github.com/microsoft/fhir-server/blob/6474001b582994e703506d94dbb89c80c8b82c24/src/Microsoft.Health.Fhir.Shared.Core/Features/Persistence/RawResourceFactory.cs#L49  ?

src/Microsoft.Health.Fhir.Core/Features/Persistence/ResourceDeserializer.cs
YazanMSFT: I don't think this works, as users can do a create using a PUT. With these changes, we only set the flags to true if the create came in as a POST. On a PUT, we don't know if it will be a create or update (and thus the version ID) until the call comes back from the data layer to Cosmos/SQL.

deepak-ms: Discussed with Yazan offline. Though we can make it work in case of PUT (needs extra trip to DB to fetch previous version id), plan is to not do it for now to avoid any PUT perf regression (with the understanding that future reads of this resource will have deser/ser cost).

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleSerializer.cs
YazanMSFT: Thanks, switched to using the GetLiteral() extension method

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Upsert/UpsertResourceHandler.cs
YazanMSFT: I'm only setting the VersionId for creates, not upserts. It doesn't seem like there are any necessary changes in the create case.  For creates that come in on the upsert endpoint, we would still do the deserialize/serialize to update the versionId and lastUpdated properties.

src/Microsoft.Health.Fhir.Core/Features/Persistence/ResourceDeserializer.cs
deepak-ms: BTW, can we use persisted request method (PUT vs POST) along with timestamp T to differentiate resources with/without versionId?
method == POST and lastUpdated > T  => avoid deser/ser

YazanMSFT: That could work too, but we'd have to store the date that the FHIR server was upgraded to the version that has these changes as the T value. That could get complicated when you have multiple compute deployments as you should only have T stored once all the compute instances have been upgraded to the this version of the code.

deepak-ms: Consider T is coming from configs (with default value = max timestamp, which means optimization is disabled). Now when all the clusters are upgraded, we can apply config change (T = current timestamp) to the system (this can be in the next release). Does that make sense? 
All this will help in avoiding schema migration complexities and increase in data size per document. Right?

src/Microsoft.Health.Fhir.Api/Features/Routing/UrlResolver.cs
```diff
@@ -128,6 +129,32 @@ public Uri ResolveResourceUrl(ResourceElement resource, bool includeVersion = fa
             return new Uri(uriString);
         }
 
+        public Uri ResolveResourceWrapperUrl(ResourceWrapper resource, bool includeVersion = false)
```
johnstairs, line 132: Suggest exacting a common private method and calling it instead of the duplication

johnstairs:
This scenario looks broken:
- POST new resource
- GET by ID. Everything looks good. version = 1
- PUT an update. In response, version =2. Looks good.
- GET by ID. `meta` is missing from response.

johnstairs:
On a PUT, we should optimistically set the version to 1 in the resource. In the stored procedure, if we see the actual version is not 1, then we can set `rawResource.versionSet` to `false`.

src/Microsoft.Health.Fhir.Core/Features/Persistence/RawResource.cs
johnstairs: I see, so for earlier documents we will need to patch the value. Going forward, I don't think there is a reason this field should every be left out, since the value is always computed by the FHIR server. We should document that for for backcompat, we need to check this field too. 
We could also just have one flag for both properties, which would slightly reduce storage size.

src/Microsoft.Health.Fhir.Shared.Api/Features/Headers/RawFhirResultExtensions.cs
YazanMSFT: I added an interface and a RawResourceElement class to be passed around instead, let me know if that's closer to what you were thinking.

src/Microsoft.Health.Fhir.Core/Features/Persistence/ResourceDeserializer.cs
```diff
@@ -38,6 +40,95 @@ public ResourceElement Deserialize(ResourceWrapper resourceWrapper)
...
+            {
+                return (resourceWrapper.RawResource.Data, JsonDocument.Parse(resourceWrapper.RawResource.Data));
+            }
+
+            var jsonDocument = JsonDocument.Parse(resourceWrapper.RawResource.Data);
+
+            using (var ms = new MemoryStream())
```
johnstairs, line 70: We should use a pool of these (RecycleableMemoryStream)

src/Microsoft.Health.Fhir.Core/Features/Persistence/ResourceDeserializer.cs
```diff
@@ -38,6 +40,95 @@ public ResourceElement Deserialize(ResourceWrapper resourceWrapper)
...
+        /// </summary>
+        /// <param name="resourceWrapper">Input ResourceWrapper to convert to a JsonDocument</param>
+        /// <returns>A <see cref="JsonDocument"/></returns>
+        public static (string, JsonDocument) DeserializeToJsonDocument(ResourceWrapper resourceWrapper)
+        {
+            EnsureArg.IsNotNull(resourceWrapper, nameof(resourceWrapper));
+
+            if (resourceWrapper.RawResource.LastUpdatedSet && resourceWrapper.RawResource.VersionSet)
+            {
+                return (resourceWrapper.RawResource.Data, JsonDocument.Parse(resourceWrapper.RawResource.Data));
```
johnstairs, line 65: In this case, we should be able to skip deserialization altogether.

src/Microsoft.Health.Fhir.Core/Features/Persistence/ResourceDeserializer.cs
YazanMSFT: Good call, I didn't know about this.

src/Microsoft.Health.Fhir.Core/Features/Persistence/ResourceDeserializer.cs
YazanMSFT: This is true for a GET on a single resource, but we need the JsonDocument for serializing it into a bundle response. I might be able to refactor so that we only create the JsonDocument when we know we need to

johnstairs: Even in a bundle, we're deserializing it and serializing it untouched, right?

YazanMSFT: Right, but when serializing the bundle using a Utf8JsonWriter, you can't write raw json and need to have a JsonDocument object.

An alternative that I got to work is to have a secondary writer on the stream (StreamWriter), and pass in the raw json for it to write out to the stream. It's a little ugly but I think it's worthwhile.

YazanMSFT: These are the changes I mentioned: https://github.com/microsoft/fhir-server/pull/1178/files#diff-c4e157da5e4618e892bb001771725246R102-R111

It's a bit ugly (2 stream writers, and having to write the NullValue then seek back to ovewrite it to keep the Utf8JsonWriter in a happy state, and copying a memory stream to the response stream), but it does get us out of doing that extra JsonDocument.Parse

src/Microsoft.Health.Fhir.Core/Features/Persistence/ResourceDeserializer.cs
YazanMSFT: This got moved to ResourceWrapperExtensions, but should be resolved.

src/Microsoft.Health.Fhir.Core/Features/Persistence/RawResource.cs
YazanMSFT: Updated so LastUpdated doesn't get cleared, and reduced to one flag.

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Upsert/UpsertResourceHandler.cs
YazanMSFT: I've applied John's suggestion of assuming Version = 1 on upserts, and updated the stored proc to set the rawResourceMetaSet flag to false in the stored proc when the version isn't 1.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/4.diff.sql
```diff
@@ -0,0 +1,411 @@
+
+/*************************************************************
+    Resource table
+**************************************************************/
+
+ALTER TABLE dbo.Resource
+    ADD RawResourceMetaSet bit NOT NULL DEFAULT 0
```
johnstairs, line 7: DEFAULT 0 is missing in 4.sql

src/Microsoft.Health.Fhir.Shared.Core/Features/Persistence/RawResourceFactory.cs
```diff
@@ -31,32 +31,37 @@ public RawResourceFactory(FhirJsonSerializer fhirJsonSerializer)
...
             var poco = resource.ToPoco<Resource>();
 
             var versionId = poco.Meta?.VersionId;
-            var lastUpdated = poco.Meta?.LastUpdated;
 
             try
             {
-                // Clear meta version and lastUpdated since these are set based on generated values when saving the resource
+                // Clear meta version if keepMeta is false since this is set based on generated values when saving the resource
                 if (poco.Meta != null)
```
johnstairs, line 45: why does this only done if poco.Meta != null?

src/Microsoft.Health.Fhir.Shared.Api/Features/Formatters/FhirJsonOutputFormatter.cs
```diff
@@ -61,6 +75,55 @@ public override async Task WriteResponseBodyAsync(OutputFormatterWriteContext co
...
+                if (elementsSearchParameter != null && elementsSearchParameter.Any())
+                {
+                    // _elements is not supported for a raw resource, revert to using FhirJsonSerializer
+                    resource = (context.Object as RawResourceElement).ToPoco<Resource>(_deserializer);
+                }
+                else
+                {
+                    using (TextWriter textWriter = context.WriterFactory(response.Body, selectedEncoding))
+                    {
+                        textWriter.Write((context.Object as RawResourceElement).ResourceData);
```
johnstairs, line 116: It would be more efficient to pass in the stream as a parameter to SerializeToJsonString (which would have a different name). The result does not need to be buffered as a string.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleSerializer.cs
```diff
@@ -0,0 +1,151 @@
...
+                {
+                    if (bundle.Entry?.Any() == true)
+                    {
+                        writer.WriteStartArray("entry");
+                        foreach (var entry in bundle.Entry)
+                        {
+                            var rawBundleEntry = entry as RawBundleEntryComponent;
+                            writer.WriteStartObject();
+
+                            if (!string.IsNullOrEmpty(rawBundleEntry.FullUrl))
```
johnstairs, line 96: Since you are doing a soft cast, this could null ref

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleSerializer.cs
```diff
@@ -0,0 +1,151 @@
...
+                            {
+                                writer.WriteString("fullUrl", rawBundleEntry.FullUrl);
+                            }
+
+                            writer.WritePropertyName("resource");
```
johnstairs, line 101: Can we avoid buffering by instead manually writing something like ", resource = " instead?

src/Microsoft.Health.Fhir.Core/Features/Persistence/RawResource.cs
```diff
@@ -31,5 +32,8 @@ protected RawResource()
         [JsonProperty("format")]
         [JsonConverter(typeof(StringEnumConverter))]
         public FhirResourceFormat Format { get; protected set; }
+
+        [JsonProperty("metaSet")]
+        public bool MetaSet { get; protected set; }
```
johnstairs, line 37: Minor naming suggestion: `IsMetaSet` (along with some explanatory comments).

src/Microsoft.Health.Fhir.Core/Features/Persistence/ResourceDeserializer.cs
```diff
@@ -49,5 +58,17 @@ internal ResourceElement DeserializeRaw(RawResource rawResource, string version,
 
             return deserializer(rawResource.Data, version, lastModified);
         }
+
+        internal ResourceElement DeserializeRaw(RawResourceElement rawResourceElement)
```
johnstairs, line 62: It's not immediately obvious what the difference is between this and `Deserialize()`. Both have the same signature :)

src/Microsoft.Health.Fhir.Core/Features/Persistence/RawResource.cs
YazanMSFT: Sure. Should I do this throughout (so in the sql column name too)?

src/Microsoft.Health.Fhir.R5.Client/Microsoft.Health.Fhir.R5.Client.csproj
```diff
@@ -1,19 +1,16 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
-
   <PropertyGroup>
```
johnstairs, line 2: Is there a particular reason for removing the blank lines?

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/RawBundleEntryComponent.cs
```diff
@@ -0,0 +1,37 @@
...
+            EnsureArg.IsNotNull(resourceWrapper, nameof(resourceWrapper));
+
+            ResourceElement = new RawResourceElement(resourceWrapper);
+        }
+
+        public RawResourceElement ResourceElement { get; set; }
+
+        public override IDeepCopyable DeepCopy()
+        {
+            if (Resource != null)
```
johnstairs, line 29: nit-picking here, but I guess we can't support this is ResourceElement is not null, right?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -197,7 +202,33 @@ private async Task<SearchResult> SearchImpl(SearchOptions searchOptions, bool hi
...
+                        int version;
+                        bool isDeleted;
+                        long resourceSurrogateId;
+                        string requestMethod;
+                        bool isMatch;
+                        bool isPartialEntry;
+                        bool rawResourceMetaSet = false;
+                        Stream rawResourceStream;
+
+                        if (_schemaInformation.Current <= 3)
```
johnstairs, line 216: Another approach is to always add RawResourceMetaSet to the projection, and to make that false when version is less than 4.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -180,38 +208,77 @@ public async Task<ResourceWrapper> GetAsync(ResourceKey key, CancellationToken c
...
+                        {
+                            var resourceTable = V3.Resource;
 
-                        (long resourceSurrogateId, int version, bool isDeleted, bool isHistory, Stream rawResourceStream) = sqlDataReader.ReadRow(
-                            resourceTable.ResourceSurrogateId,
-                            resourceTable.Version,
-                            resourceTable.IsDeleted,
-                            resourceTable.IsHistory,
-                            resourceTable.RawResource);
+                            (long resourceSurrogateId, int version, bool isDeleted, bool isHistory, Stream rawResourceStream) = sqlDataReader.ReadRow(
```
johnstairs, line 215: can we have different variants of this statement and reuse the rest?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/V3/CompartmentAssignmentRowGenerator.cs
```diff
@@ -0,0 +1,97 @@
...
+
+using System.Collections.Generic;
+using System.Threading;
+using EnsureThat;
+using Microsoft.Health.Fhir.Core.Models;
+using Microsoft.Health.SqlServer.Features.Schema.Model;
+
+namespace Microsoft.Health.Fhir.SqlServer.Features.Storage.TvpRowGeneration.V3
+{
+    internal class CompartmentAssignmentRowGenerator : ITableValuedParameterRowGenerator<ResourceMetadata, Schema.Model.V3.CompartmentAssignmentTableTypeRow>
```
johnstairs, line 14: Hmm, I think we need to fix our codegen, since these table types are immutable and cannot change. So we should not need to have all this copy-paste. I'll open an issue.

src/Microsoft.Health.Fhir.R5.Client/Microsoft.Health.Fhir.R5.Client.csproj
YazanMSFT: Just that I'm still using VS2019 and it did it automatically :)

src/Microsoft.Health.Fhir.Core/Features/Persistence/RawResource.cs
YazanMSFT: I made the changes in the sql column names too, made sense to have it match between data stores and match the convention of prefixing with is for bit fields.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/4.diff.sql
YazanMSFT: Good catch

src/Microsoft.Health.Fhir.Shared.Api/Features/Formatters/FhirJsonOutputFormatter.cs
YazanMSFT: Should be resolved now, but it feels more brittle and makes me a bit nervous

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleSerializer.cs
YazanMSFT: Should be resolved now. It is nicer than having to copy over to another stream, but it is more error prone and makes me feel a bit uneasy - see comment about the comma :)

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
YazanMSFT: I'm not sure I understand, I'll sync up with you tomorrow about this.

src/Microsoft.Health.Fhir.Core/Extensions/DateTimeOffsetExtensions.cs
```diff
@@ -0,0 +1,19 @@
...
+
+using System;
+
+namespace Microsoft.Health.Fhir.Core.Extensions
+{
+    public static class DateTimeOffsetExtensions
+    {
+        private const string DateTimeOffsetFormat = "yyyy-MM-dd'T'HH:mm:ss.FFFFFFFK";
+
+        public static string ToInstantString(this DateTimeOffset offset)
```
johnstairs, line 14: Might want to explain that this formats in UTC ISO 8601, and differs from standard format string "o" in that the fractional digits are omitted once zero.
(Although using "o" is ~5x faster because it is special-cased and they do not need to interpret the format string. But this is unlikely to be an issue.).

src/Microsoft.Health.Fhir.Shared.Api/Features/Formatters/FhirJsonOutputFormatter.cs
```diff
@@ -61,6 +75,51 @@ public override async Task WriteResponseBodyAsync(OutputFormatterWriteContext co
...
+                        }
+                    }
+                }
+                else if (bundle.Entry.All(x => x is RawBundleEntryComponent))
+                {
+                    await _bundleSerializer.Serialize(context.Object as Hl7.Fhir.Model.Bundle, context.HttpContext.Response.Body);
+                    return;
+                }
+            }
+            else if (context.Object is RawResourceElement)
```
johnstairs, line 105: It looks like the output stream optimization does not kick in for POST/PUT.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleSerializer.cs
```diff
@@ -0,0 +1,155 @@
...
+                }
+
+                async Task SerializeEntries()
+                {
+                    if (bundle.Entry?.Any() == true)
+                    {
+                        writer.WriteStartArray("entry");
+                        foreach (var entry in bundle.Entry)
+                        {
+                            if (!(entry is RawBundleEntryComponent))
```
johnstairs, line 88: You can combine this and the cast into one statement 
``` csharp
if (!(entry is RawBundleEntryComponent rawBundleEntry))
```

src/Microsoft.Health.Fhir.Core/Extensions/RawResourceElementExtensions.cs
```diff
@@ -0,0 +1,108 @@
...
+        {
+            EnsureArg.IsNotNull(rawResource, nameof(rawResource));
+
+            if (rawResource.RawResource.IsMetaSet)
+            {
+                using (var sw = new StreamWriter(outputStream, leaveOpen: true))
+                {
+                    await sw.WriteAsync(rawResource.RawResource.Data);
+                    await sw.FlushAsync();
+                    return;
```
johnstairs, line 35: Instead of explicit calls to FlushAsync before disposing, you can instead do:

``` csharp
                await using (var sw = new StreamWriter(outputStream, leaveOpen: true))
                {
                    await sw.WriteAsync(rawResource.RawResource.Data);
                    return;
                }
```  

which will invoke `DisposeAsync()` instead of `Dispose()`, which in turn calls `FlushAsync()`.

Taking this a step further you can use a "using declaration":
``` csharp
await using var sw = new StreamWriter(outputStream, leaveOpen: true);
await sw.WriteAsync(rawResource.RawResource.Data);
return;
```

src/Microsoft.Health.Fhir.Core/Extensions/RawResourceElementExtensions.cs
```diff
@@ -0,0 +1,108 @@
...
+        private static readonly JsonWriterOptions WriterOptions = new JsonWriterOptions { Encoder = System.Text.Encodings.Web.JavaScriptEncoder.UnsafeRelaxedJsonEscaping };
+
+        /// <summary>
+        /// Get the raw data from resourceWrapper as a string and JsonDocument.
+        /// If the RawResource's VersionSet or LastUpdatedSet are false, then the RawResource's data will be updated
+        /// to have them set to the values in the ResourceWrapper
+        /// </summary>
+        /// <param name="rawResource">Input RawResourceElement to convert to a JsonDocument</param>
+        /// <param name="outputStream">Stream to serialize to</param>
+        public static async Task SerializeToStreamAsJson(this RawResourceElement rawResource, Stream outputStream)
```
johnstairs, line 25: Maybe the name should be SerializeToStreamAsUft8Json

src/Microsoft.Health.Fhir.Core/Extensions/RawResourceElementExtensions.cs
```diff
@@ -0,0 +1,108 @@
...
+            var jsonDocument = JsonDocument.Parse(rawResource.RawResource.Data);
+
+            using (Utf8JsonWriter writer = new Utf8JsonWriter(outputStream, WriterOptions))
+            {
+                writer.WriteStartObject();
+                bool foundMeta = false;
+
+                foreach (var current in jsonDocument.RootElement.EnumerateObject())
+                {
+                    if (current.Name == "meta")
```
johnstairs, line 48: You can instead call `current.NameEquals()`, which presumably avoids the string allocation for the Name property.

src/Microsoft.Health.Fhir.Core/Extensions/RawResourceElementExtensions.cs
```diff
@@ -0,0 +1,108 @@
...
+                    {
+                        foundMeta = true;
```
johnstairs, line 50: Once we have found meta, we can skip the current.Name check and go straight to `current.WriteTo(writer)` for remaining elements

src/Microsoft.Health.Fhir.Shared.Api/Features/Formatters/FhirJsonOutputFormatter.cs
YazanMSFT: Yeah I haven't changed any of the write handlers to return using the raw resource code path. I was planning to do it as a separate work item.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleSerializer.cs
YazanMSFT: Thanks! Learned something new :)

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/4.sql
```diff
@@ -0,0 +1,2159 @@
...
+    @baseResourceSurrogateId bigint,
+    @resourceTypeId smallint,
+    @resourceId varchar(64),
+    @eTag int = NULL,
+    @allowCreate bit,
+    @isDeleted bit,
+    @keepHistory bit,
+    @requestMethod varchar(10),
+    @rawResource varbinary(max),
+    @isRawResourceMetaSet bit = 0,
```
johnstairs, line 1348: Could we skip this parameter? I think we can set the colum in the databse based solely on whether the final version is 1.

src/Microsoft.Health.Fhir.Shared.Api/Features/Formatters/FhirJsonOutputFormatter.cs
johnstairs: Sounds good

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/4.sql
johnstairs: If we skip it, can skip all the duplicate TvpRowRegenerators...

YazanMSFT: The GetResource call is a stored proc, so I don't think we can get rid of those quite yet.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/4.diff.sql
```diff
@@ -0,0 +1,410 @@
...
+AS
+    SET NOCOUNT ON
+
+    IF (@version IS NULL) BEGIN
+        SELECT ResourceSurrogateId, Version, IsDeleted, IsHistory, IsRawResourceMetaSet, RawResource
+        FROM dbo.Resource
+        WHERE ResourceTypeId = @resourceTypeId AND ResourceId = @resourceId AND IsHistory = 0
+    END
+    ELSE BEGIN
+        SELECT ResourceSurrogateId, Version, IsDeleted, IsHistory, IsRawResourceMetaSet, RawResource
```
johnstairs, line 40: I think we have a problem here. By "altering" this stored procedure in this way, where the output schema changes, we would break existing clients that are using the "V3" schema. 
(Apologies, we have not gone through all the schema upgrade scenarios yet).
I can think of two ways we could deal with this.

1.  We create a new stored procedure called something like `dbo.ReadResource_4`. Ideally, the codegen would strip off the suffix. I think the integration test that verifies upgrade schema compatibility would fail because of the extra stored procedure left over from the previous version.
2. We keep this an `ALTER PROCEDURE` but move `IsRawResourceMetaSet` to the last column in the select statement. Then the code that reads it would need to read that column after reading the RawResource stream (if of course the schema version > 3). I think we want to do that outside of the ReadRow call because we do not want the the SqlDataReader to buffer the entire RawResource.

It seems like in the general case we will need to support (1), but we can probably get away with (2) here.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/4.sql
johnstairs: If we do (2) [here](https://github.com/microsoft/fhir-server/pull/1178#discussion_r485111502) then I don't think the codegen for this stored procedure is affected.

johnstairs: Actually, we have a similar problem here. If an "old" FHIR server calls this new stored procedure, then we will incorrectly assume that the version is set in the resource.

johnstairs: Sorry for thinking out loud here. We actually require that all FHIR servers be upgraded to the new version _before_ we begin the schema upgrade, so this may not be an issue.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -198,17 +210,22 @@ public async Task<ResourceWrapper> GetAsync(ResourceKey key, CancellationToken c
...
-                            null,
-                            new DateTimeOffset(ResourceSurrogateIdHelper.ResourceSurrogateIdToLastUpdated(resourceSurrogateId), TimeSpan.Zero),
-                            isDeleted,
-                            searchIndices: null,
-                            compartmentIndices: null,
-                            lastModifiedClaims: null)
+                        key.Id,
+                        version.ToString(CultureInfo.InvariantCulture),
+                        key.ResourceType,
+                        new RawResource(rawResource, FhirResourceFormat.Json, isMetaSet: isRawResourceMetaSet),
```
johnstairs, line 222: minor thing: could you indent the arguments to `new ResourceWrapper()`?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -180,9 +185,16 @@ public async Task<ResourceWrapper> GetAsync(ResourceKey key, CancellationToken c
...
+                        int version;
+                        bool isDeleted;
+                        bool isHistory;
+                        bool isRawResourceMetaSet = false;
+                        Stream rawResourceStream;
+
                         var resourceTable = VLatest.Resource;
 
-                        (long resourceSurrogateId, int version, bool isDeleted, bool isHistory, Stream rawResourceStream) = sqlDataReader.ReadRow(
+                        (resourceSurrogateId, version, isDeleted, isHistory, rawResourceStream) = sqlDataReader.ReadRow(
```
johnstairs, line 197: Minor: we can revert back to the declarations inside of the tuple deconstruction:
``` csharp 
(long resourceSurrogateId, int version, bool isDeleted, bool isHistory, Stream rawResourceStream) =
```

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -197,7 +202,18 @@ private async Task<SearchResult> SearchImpl(SearchOptions searchOptions, bool hi
...
+                        int version;
+                        bool isDeleted;
+                        long resourceSurrogateId;
+                        string requestMethod;
+                        bool isMatch;
+                        bool isPartialEntry;
+                        bool isRawResourceMetaSet = false;
+                        Stream rawResourceStream;
+
+                        (resourceTypeId, resourceId, version, isDeleted, resourceSurrogateId, requestMethod, isMatch, isPartialEntry, isRawResourceMetaSet, rawResourceStream) = reader.ReadRow(
```
johnstairs, line 216: Same here: we can "inline" the variable declarations like we did before.

johnstairs (approved):

johnstairs:
Nice job, @YazanMSFT! This is really great.

Issue #1084, opened 2020-06-15
History and _at does not return what we expect


ellgre:
**Describe the bug**
We expect the _at search parameter in combination with _history should give us any document that was valid at the provided date. It seems like we only get the correct document if we use the lastUpdated dates and not for any date between. Qualifiers like ge, le, eq are shown in the self link, but we do not think that they have been used since we get both version regardless of which date we have in input.

**To Reproduce**
1. We have a Patient which has a version updated 2020-06-12 and a version updated 2020-06.15
2. When searching with &_at=2020-06-12 we get version 1 as expected
3. When searching with &_at=2020-06-15 we get version 2 as expected.
4. When searching with &_at=2020-06-13 we get no documents
5. When searching with &_at=2020-06-16 we get no documents

**Expected behavior**
When searching with a date between June 12 and 14 we expect to get version 1, and when searching with June 15 or later we expect to get version 2. When we search with a date before June 12 we expect to get no documents.

**Actual behavior**
Only for searches with the _at=<lastUpdated> we get the expected document in return.
If we input a date qualifier (le, ge, eq) we get both versions regardless of which date we input.

brendankowitz:
Hi @ellgre, in reading through the R4 spec against the current behavior, I think this is a bug

ellgre:
Do you know when we can expect this bug to be solved? 

We get a lot of historic events that we need to react on, and have to get the version of Patient valid at that given point in time in order to use the correct attributes.

rbans96:
@ellgre  Which database are you using? Is it SqlServer or Cosmosdb?

ellgre:
It is Cosmosdb.

CaitlinV39:
AB#74811

CaitlinV39:
AB#75005

CaitlinV39:
@ellgre given how long this has been open we're goign to close for now as won't fix. Please let us know if you are still struggling with this.

PR #1183, opened 2020-08-05
Fix history search param _at in SQL Server

rbans96:
## Description
The code changes modifies the /history/_at= behavior as expected in the bug
https://github.com/microsoft/fhir-server/issues/1084

## Related issues
Addresses [#74811](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Resolute/Stories/?workitem=74811)

## Testing
Unit test and e2e test are added

Here is the updated SQL query for /history/_at=[] search

DECLARE @p0 Int = 11;
DECLARE @p1 SmallInt = 96;
DECLARE @p2 VarChar(64) = '53e959f5-a007-4207-9728-1630f1bf3819';
DECLARE @p3 BigInt = 5098579070714800000;

SET STATISTICS IO ON;
SET STATISTICS TIME ON;

SELECT TOP (@p0) r.ResourceTypeId, r.ResourceId, r.Version, r.IsDeleted, r.ResourceSurrogateId, r.RequestMethod, CAST(1 AS bit) AS IsMatch, r.RawResource
FROM dbo.Resource r
WHERE ResourceTypeId = @p1
    AND ResourceId = @p2
    AND ResourceSurrogateId <= @p3
    AND NOT EXISTS( 
        SELECT *
        FROM dbo.Resource r2
        WHERE ResourceSurrogateId <= @p3
            AND r2.ResourceId = r.ResourceId
            AND r2.ResourceSurrogateId > r.ResourceSurrogateId
        )
OPTION(RECOMPILE)

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
@@ -137,7 +137,7 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Search.Expressions.Visitors.Q
                             foreach (var denormalizedPredicate in expression.DenormalizedExpressions)
                             {
                                 SearchParameterExpressionBase exp = denormalizedPredicate as SearchParameterExpressionBase;
-                                if (string.Equals(exp.Parameter.Name, "_resourceSurrogateId", StringComparison.Ordinal))
+                                if (string.Equals(exp.Parameter.Name, SqlSearchParameters.ResourceSurrogateIdParameterName, StringComparison.Ordinal))
                                 {
                                     denormalizedPredicate.AcceptVisitor(DispatchingDenormalizedSearchParameterQueryGenerator.Instance, GetContext());
                                 }
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
@@ -136,8 +136,8 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Search.Expressions.Visitors.Q
                             delimitedClause.BeginDelimitedElement();
                             foreach (var denormalizedPredicate in expression.DenormalizedExpressions)
                             {
-                                SearchParameterExpression exp = denormalizedPredicate as SearchParameterExpression;
-                                if (exp.IsAtParameter)
+                                SearchParameterExpression searchParameterExpression = denormalizedPredicate as SearchParameterExpression;
+                                if (searchParameterExpression != null && searchParameterExpression.IsAtParameter)
                                 {
                                     denormalizedPredicate.AcceptVisitor(DispatchingDenormalizedSearchParameterQueryGenerator.Instance, GetContext());
                                 }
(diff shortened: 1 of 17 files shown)


johnstairs:
I think your query for `history?_at=2020` should look like this:
```
SELECT TOP (@p0) r.ResourceTypeId, r.ResourceId, r.Version, r.IsDeleted, r.ResourceSurrogateId, r.RequestMethod, CAST(1 AS bit) AS IsMatch, r.RawResource
FROM dbo.Resource r
WHERE ResourceSurrogateId <= @p1
    AND NOT EXISTS (
        SELECT *
        FROM dbo.Resource r2
        WHERE r2.ResourceSurrogateId <= @p1
            AND r2.ResourceTypeId = r.ResourceTypeId
            AND r2.ResourceId = r.ResourceId
            AND r2.ResourceSurrogateId > r.ResourceSurrogateId
        )
OPTION(RECOMPILE)
```
You are missing the `r2.ResourceTypeId = r.ResourceTypeId`.

In addition to being more correct, it performs a lot faster ;)

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -120,6 +120,40 @@ public object VisitSqlRoot(SqlRootExpression expression, SearchOptions context)
...
+                    {
+                        StringBuilder.AppendLine("SELECT *");
+                        StringBuilder.Append("FROM ").Append(VLatest.Resource).Append(" ").AppendLine(resourceTableAlias2);
+                        using (var delimitedClause = StringBuilder.BeginDelimitedWhereClause())
+                        {
+                            delimitedClause.BeginDelimitedElement();
+                            foreach (var denormalizedPredicate in expression.DenormalizedExpressions)
+                            {
+                                SearchParameterExpressionBase exp = denormalizedPredicate as SearchParameterExpressionBase;
+                                if (string.Equals(exp.Parameter.Name, SqlSearchParameters.ResourceSurrogateIdParameterName, StringComparison.Ordinal))
```
johnstairs, line 140: exp could (theoretically) be null

rbans96:
> I think your query for `history?_at=2020` should look like this:
> 
> ```
> SELECT TOP (@p0) r.ResourceTypeId, r.ResourceId, r.Version, r.IsDeleted, r.ResourceSurrogateId, r.RequestMethod, CAST(1 AS bit) AS IsMatch, r.RawResource
> FROM dbo.Resource r
> WHERE ResourceSurrogateId <= @p1
>     AND NOT EXISTS (
>         SELECT *
>         FROM dbo.Resource r2
>         WHERE r2.ResourceSurrogateId <= @p1
>             AND r2.ResourceTypeId = r.ResourceTypeId
>             AND r2.ResourceId = r.ResourceId
>             AND r2.ResourceSurrogateId > r.ResourceSurrogateId
>         )
> ORDER BY r.ResourceSurrogateId ASC
> OPTION(RECOMPILE)
> ```
> 
> You are missing the `r2.ResourceTypeId = r.ResourceTypeId`.
> 
> In addition to being more correct, it performs a lot faster ;)

Good point, Fixed

johnstairs:
We throw an error when the query specifies both `_at` and `_since`/`before`, but we really shouldn't. If we support both, then the current approach of put an `AtParam` on SearchOptions and reducing both to expressions on `ResourceSurrogateId` will not work, because we need to know which expression refers to the _at so that we can include the semi-join for the right predicate.

rbans96:
> We throw an error when the query specifies both `_at` and `_since`/`before`, but we really shouldn't. If we support both, then the current approach of put an `AtParam` on SearchOptions and reducing both to expressions on `ResourceSurrogateId` will not work, because we need to know which expression refers to the _at so that we can include the semi-join for the right predicate.

I have added a flag to identify the _at expression to be able to include just that predicate in the semi-join for resourceSurrogateId. This approach seems to work for the query specifying both _at and _since parameter. Also, I will create a new user story to enable both _at and _since/_before.

rbans96:
Closing this PR as it is considerably old and there is a plan to redesign the _at a bit and create a new PR.

PR #1184, opened 2020-08-05
Initial implementation of "Sort" feature (one date parameter only)

itye-msft:
This is an initial implementation of the Sort feature, where we only support a select set of `date` search parameters. We only support search parameters that are known to have at most one value per resource. 
Performance of the queries has not yet been optimized.


src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/IExpressionVisitor.cs
```diff
@@ -74,5 +74,12 @@ public interface IExpressionVisitor<in TContext, out TOutput>
         /// <param name="expression">The expression to visit.</param>
         /// <param name="context">The input.</param>
         TOutput VisitInclude(IncludeExpression expression, TContext context);
+
+        /// <summary>
+        /// Visits the <see cref="SortParameterExpression"/>.
+        /// </summary>
+        /// <param name="expression">The expression to visit.</param>
+        /// <param name="context">The input</param>
+        TOutput VisitSortParameter(SortParameterExpression expression, TContext context);
```
eladiw, line 83: SortParameterExpression -> SortExpression ?
I see that IncludeExpression is not IncludeParameterExpression

src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/SortParameterExpression.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        public override TOutput AcceptVisitor<TContext, TOutput>(IExpressionVisitor<TContext, TOutput> visitor, TContext context)
+        {
+            EnsureArg.IsNotNull(visitor, nameof(visitor));
+            return visitor.VisitSortParameter(this, context);
+        }
+
+        public override string ToString()
+        {
+            return $"(Param {Parameter.Name})";
```
eladiw, line 29:
```suggestion
            return $"(Sort {Parameter.Name})";
```

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/ConcatenationRewriter.cs
```diff
@@ -45,7 +45,7 @@ public override Expression VisitSqlRoot(SqlRootExpression expression, object con
 
                 // The expressions contained within a ChainExpression or IncludeExpression
```
eladiw, line 46: have you promoted the sort to table expression also?
if so, update the comment

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/MissingSearchParamVisitor.cs
```diff
@@ -28,7 +28,7 @@ public override Expression VisitSqlRoot(SqlRootExpression expression, object con
             {
                 TableExpression tableExpression = expression.TableExpressions[i];
 
-                if (tableExpression.NormalizedPredicate.AcceptVisitor(Scout.Instance, null))
+                if (tableExpression.Kind != TableExpressionKind.Sort && tableExpression.NormalizedPredicate.AcceptVisitor(Scout.Instance, null))
```
eladiw, line 31: please add a comment here to explain this if statement, why sort is a special case here

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/NormalizedSearchParameterQueryGeneratorFactory.cs
```diff
@@ -177,5 +182,10 @@ public NormalizedSearchParameterQueryGenerator VisitInclude(IncludeExpression ex
         {
             return IncludeQueryGenerator.Instance;
         }
+
+        public NormalizedSearchParameterQueryGenerator VisitSortParameter(SortParameterExpression expression, object context)
+        {
+            return GetNormalizedSearchParameterQueryGenerator(expression.Parameter);
```
eladiw, line 188: please add comment

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SearchParameterQueryGenerator.cs
```diff
@@ -35,6 +35,19 @@ public override SearchParameterQueryGeneratorContext VisitSearchParameter(Search
             return expression.Expression.AcceptVisitor(this, context);
         }
 
+        public override SearchParameterQueryGeneratorContext VisitSortParameter(SortParameterExpression expression, SearchParameterQueryGeneratorContext context)
+        {
+            short searchParamId = context.Model.GetSearchParamId(expression.Parameter.Url);
+            SmallIntColumn searchParamIdColumn = VLatest.SearchParam.SearchParamId;
```
eladiw, line 41:
```suggestion
            var searchParamIdColumn = VLatest.SearchParam.SearchParamId;
```

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -238,11 +239,16 @@ public object VisitTable(TableExpression tableExpression, SearchOptions context)
 
                 case TableExpressionKind.Top:
                     var sortOrder = context.GetFirstSortOrderForSupportedParam();
+                    var (pi, sort) = context.GetFirstSortSupportedParam();
```
eladiw, line 242:
```suggestion
                    var (paramInfo, sort) = context.GetFirstSortSupportedParam();
```

same as above

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -403,6 +409,45 @@ public object VisitTable(TableExpression tableExpression, SearchOptions context)
                         StringBuilder.Append("FROM ").AppendLine(includeCte);
                     }
 
+                    break;
+                case TableExpressionKind.Sort:
+
+                    var (searchParamaterInfo, s) = context.GetFirstSortSupportedParam();
```
eladiw, line 415:
```suggestion
                    var (paramInfo, sort) = context.GetFirstSortSupportedParam();
```

same as above

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -403,6 +409,45 @@ public object VisitTable(TableExpression tableExpression, SearchOptions context)
...
+
+                    // the normalized generator holds the data for the search table. TODO: use it.
```
eladiw, line 417: 'todo' still relevant?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -403,6 +409,45 @@ public object VisitTable(TableExpression tableExpression, SearchOptions context)
...
+                    if (searchParamaterInfo.Type == ValueSets.SearchParamType.Date)
```
eladiw, line 418: if (not date) 
  return

keeps the nested code smaller

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/SortRewriter.cs
```diff
@@ -0,0 +1,44 @@
...
+                return expression;
+            }
+
+            var queryGenerator = _normalizedSearchParameterQueryGeneratorFactory.GetNormalizedSearchParameterQueryGenerator(context.Sort[0].searchParameterInfo);
+
+            var newNormalizedPredicates = new List<TableExpression>(expression.TableExpressions.Count + 1);
+            newNormalizedPredicates.AddRange(expression.TableExpressions);
+
+            newNormalizedPredicates.Add(new TableExpression(queryGenerator, new SortParameterExpression(context.Sort[0].searchParameterInfo), null, TableExpressionKind.Sort));
+            //// newNormalizedPredicates.Add(new TableExpression(queryGenerator, null, null, TableExpressionKind.Sort));
```
eladiw, line 39: remove

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/SortRewriter.cs
```diff
@@ -0,0 +1,44 @@
...
+            if (context.CountOnly)
+            {
+                return expression;
+            }
+
+            if (context.Sort.Count == 0)
+            {
+                return expression;
+            }
+
```
eladiw, line 32: if sort count > 1 throw?

eladiw: because you anyway just use sort[0]

src/Microsoft.Health.Fhir.Core/Features/KnownQueryParameterNames.cs
```diff
@@ -55,5 +55,7 @@ public static class KnownQueryParameterNames
         public const string Id = "_id";
 
         public const string Type = "_type";
+
+        public const string BirthDate = "birthdate";
```
tamirkamara, line 59: What is special about birth date?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SearchOptionsExtensions.cs
```diff
@@ -32,5 +32,24 @@ public static SortOrder GetFirstSortOrderForSupportedParam(this SearchOptions se
 
             return sortOrder;
         }
+
+        public static (Core.Models.SearchParameterInfo, SortOrder) GetFirstSortSupportedParam(this SearchOptions searchOptions)
```
rbans96, line 36: nit : GetFirstSortSupportedParam -> GetFirstSupportedSortParam

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -122,9 +122,10 @@ public object VisitSqlRoot(SqlRootExpression expression, SearchOptions context)
 
             if (!searchOptions.CountOnly)
             {
-                var sortOrder = searchOptions.GetFirstSortOrderForSupportedParam();
+                var (paramInfo, sortOrder) = searchOptions.GetFirstSortSupportedParam();
```
rbans96, line 125: paramInfo -> searchParamInfo
paramInfo was little confusing to me but may be its just me.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -403,6 +409,45 @@ public object VisitTable(TableExpression tableExpression, SearchOptions context)
...
+
+                    var (searchParamaterInfo, s) = context.GetFirstSortSupportedParam();
+
+                    // the normalized generator holds the data for the search table. TODO: use it.
+                    if (searchParamaterInfo.Type == ValueSets.SearchParamType.Date)
+                    {
+                        if (tableExpression.ChainLevel == 0)
+                        {
+                            StringBuilder.Append("SELECT ").Append(VLatest.Resource.ResourceSurrogateId, null).Append(" AS Sid1, ")
+                                .Append(VLatest.DateTimeSearchParam.StartDateTime, null).AppendLine(" as SortExpr")
```
rbans96, line 423: as SortExpr -> AS Sdt?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/SortRewriter.cs
```diff
@@ -0,0 +1,44 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+using Microsoft.Health.Fhir.Core.Features.Search;
+using Microsoft.Health.Fhir.Core.Features.Search.Expressions;
+
+namespace Microsoft.Health.Fhir.SqlServer.Features.Search.Expressions.Visitors
+{
+    internal class SortRewriter : SqlExpressionRewriter<SearchOptions>
```
rbans96, line 12: Can we add one liner comment of what this rewriter does?

src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/SortParameterExpression.cs
```diff
@@ -0,0 +1,32 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using EnsureThat;
+using Microsoft.Health.Fhir.Core.Models;
+
+namespace Microsoft.Health.Fhir.Core.Features.Search.Expressions
+{
+    /// <summary>
+    /// Represents a set of ANDed expressions over a search parameter.
```
rbans96, line 12: comment needs to be updated for sort

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
itye-msft: @rbans96 what does `Sdt` stands for?
I chose `SortExpr` so it can fit any sort type in the future (i.e. number, text etc)

src/Microsoft.Health.Fhir.Core/Features/KnownQueryParameterNames.cs
itye-msft: @brendankowitz how would I accept any supported sort parameter?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/MissingSearchParamVisitor.cs
```diff
@@ -28,7 +28,8 @@ public override Expression VisitSqlRoot(SqlRootExpression expression, object con
             {
                 TableExpression tableExpression = expression.TableExpressions[i];
 
-                if (tableExpression.NormalizedPredicate.AcceptVisitor(Scout.Instance, null))
+                // process onlu normalized predicates. Ignore Sort as it has it's own visitor.
```
eladiw, line 31:
```suggestion
                // process only normalized predicates. Ignore Sort as it has it's own visitor.
```

src/Microsoft.Health.Fhir.Core/Features/KnownQueryParameterNames.cs
johnstairs: "Supported" here means that we know that the search parameter can have at most one value, because the current implementation does not use the smallest value as the value to sort by for a resource.
We can (1) use a hard-coded list _for now_, (2) generate the set of supported search parameters by analyzing the FHIRPath expressions against the structure definitions, (3) support all search parameters, regardless of multiplicity.
I am okay with (1) for now, but I would not put the list in this file, since it is a SQL-only feature.

itye-msft:
@johnstairs , @brendankowitz  
I need some instructions regarding the Continuation Token.

cc @tamirkamara

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -241,11 +252,16 @@ public object VisitTable(TableExpression tableExpression, SearchOptions context)
 
                 case TableExpressionKind.Top:
                     var sortOrder = context.GetFirstSortOrderForSupportedParam();
+                    var (paramInfo, sort) = context.GetFirstSupportedSortParam();
```
rbans96, line 255: Since we are getting paramInfo, sort from context.GetFirstSupportedSortParam, the line above (sortOrder = context.GetFirstSortOrderForSupportedParam();) looks redundant. I think sort and sortOrder should be same here.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -125,11 +126,21 @@ public object VisitSqlRoot(SqlRootExpression expression, SearchOptions context)
 
             if (!searchOptions.CountOnly)
             {
-                var sortOrder = searchOptions.GetFirstSortOrderForSupportedParam();
-
-                StringBuilder.Append("ORDER BY ")
+                var (searchParamInfo, sortOrder) = searchOptions.GetFirstSupportedSortParam();
+                if (searchParamInfo == null || searchParamInfo.Name == KnownQueryParameterNames.LastUpdated)
+                {
+                    StringBuilder.Append("ORDER BY ")
```
rbans96, line 132: nit: this can be moved outside of if block

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/SortRewriter.cs
```diff
@@ -0,0 +1,54 @@
...
+            }
+
+            // Proceed if we sort params were requested.
+            if (context.Sort.Count == 0)
+            {
+                return expression;
+            }
+
+            // _lastUpdated sort param is handled differently than others, because it's can be
+            // infered dorectly from the resource table itself.
```
rbans96, line 38: nit: typo it's -> it,  'inferred' and 'directly'

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/SortTests.cs
```diff
@@ -90,13 +91,82 @@ public async Task GivenMoreThanTenPatients_WhenSearchedWithSortParamWithHyphen_T
...
+        public async Task GivenQueryWithMultipleFilters_WhenSearchedWithSortParamOnDatetime_ThenResourcesAreReturnedInAscendingOrder()
+        {
+            var tag = Guid.NewGuid().ToString();
+
+            var patients = await CreatePatients(tag);
+
+            // Ask to get all patient with specific tag and family name order by birthdate (timestamp)
+            // filter and sort are different based on different types
+            var filteredFamilyName = "Williamas";
+            await ExecuteAndValidateBundle($"Patient?_tag={tag}&family={filteredFamilyName}&_sort=birthdate", false, patients.Reverse().Where(x => x.Name[0].Family == filteredFamilyName).OrderBy(x => x.BirthDate).Cast<Resource>().ToArray());
```
rbans96, line 160: I am wondering, in the current bundle if we filter by family name then it will return only one patient. as there is only one Patient with family name as "Williams. And if it returns just one patient, then order doesn't really matter. Am I missing something here?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/SortTests.cs
```diff
@@ -90,13 +91,82 @@ public async Task GivenMoreThanTenPatients_WhenSearchedWithSortParamWithHyphen_T
             await ExecuteAndValidateBundle($"Patient?_tag={tag}&_sort=-_lastUpdated", false, patients.Reverse().Cast<Resource>().ToArray());
         }
 
+        // uncomment only when db cleanup happens on each run, otherwise the paging might cause expected resources to not arrive
+        /*[Fact]
+        [HttpIntegrationFixtureArgumentSets(dataStores: DataStore.SqlServer)]
+        public async Task GivenQueryWithNoFilter_WhenSearchedWithSortParam_ThenResourcesAreReturnedInAscendingOrder()
```
rbans96, line 97: Should we add tests to verify descending order too?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/SortTests.cs
itye-msft: @eladiw

itye-msft: @eladiw I fixed it.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/SortTests.cs
itye-msft: @eladiw Do you have time to add the DESC tests as well? it looks pretty quick

itye-msft: @eladiw I added these tests

src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/IExpressionVisitor.cs
eladiw: I dont see the change. still see SortParam...

eladiw (approved):

itye-msft:
@johnstairs I updated the PR with the continuation token changes

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/SortTests.cs
```diff
@@ -90,13 +91,136 @@ public async Task GivenMoreThanTenPatients_WhenSearchedWithSortParamWithHyphen_T
             await ExecuteAndValidateBundle($"Patient?_tag={tag}&_sort=-_lastUpdated", false, patients.Reverse().Cast<Resource>().ToArray());
         }
 
+        // uncomment only when db cleanup happens on each run, otherwise the paging might cause expected resources to not arrive
```
johnstairs, line 94: The idea of the tag is that your search would be scoped to it, so we do not depend on an empty database for each run.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/SearchTestsBase.cs
```diff
@@ -38,6 +39,15 @@ protected async Task<Bundle> ExecuteAndValidateBundle(string searchUrl, bool sor
             return await ExecuteAndValidateBundle(searchUrl, searchUrl, sort, expectedResources);
         }
 
+        protected async Task<Bundle> ExecuteAndValidateBundleSuperset(string searchUrl, bool sort, params Resource[] expectedResources)
```
johnstairs, line 42: Can you explain what this is for?

src/Microsoft.Health.Fhir.Core/Features/Search/ContinuationToken.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        public static ContinuationToken FromString(string json)
+        {
+            try
+            {
+                return JsonConvert.DeserializeObject<ContinuationToken>(json);
+            }
+            catch
+            {
+                return null;
```
johnstairs, line 38: I would rather we return a bad request response than assuming that the search should be started over from the beginning.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -128,20 +132,19 @@ private async Task<SearchResult> SearchImpl(SearchOptions searchOptions, bool hi
             // AND in the continuation token
             if (!string.IsNullOrWhiteSpace(searchOptions.ContinuationToken) && !searchOptions.CountOnly)
             {
-                if (long.TryParse(searchOptions.ContinuationToken, NumberStyles.None, CultureInfo.InvariantCulture, out var token))
+                var continuationToken = ContinuationToken.FromString(searchOptions.ContinuationToken);
```
johnstairs, line 135: You've introduced a breaking change where continuation tokens from a FHIR server from a previous version will not be compatible with this version

src/Microsoft.Health.Fhir.Core/Features/Search/ContinuationToken.cs
```diff
@@ -0,0 +1,42 @@
...
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Health.Fhir.Core.Features.Search
+{
+    /// <summary>
+    /// Represents the search continuation token.
+    /// </summary>
+    public class ContinuationToken
+    {
+        public long ResourceSourogateId { get; set; }
```
johnstairs, line 16: typo

src/Microsoft.Health.Fhir.Core/Features/Search/ContinuationToken.cs
```diff
@@ -0,0 +1,42 @@
...
+
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Health.Fhir.Core.Features.Search
+{
+    /// <summary>
+    /// Represents the search continuation token.
+    /// </summary>
+    public class ContinuationToken
```
johnstairs, line 14: This should instead be an array. Remember that while you are only adding support for sorting by one parameter, the spec allows sorting by any number. 
For that reason, I would make this an object array, the last entry of which would be the resource surrogate id.

src/Microsoft.Health.Fhir.Core/Features/SupportedSortParameterNames.cs
```diff
@@ -0,0 +1,15 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Fhir.Core.Features
+{
+    /// <summary>
+    /// Provides list of supported sort parameter names.
+    /// </summary>
+    public static class SupportedSortParameterNames
```
johnstairs, line 11: This is sql-specific, so please put this in the SQL project

src/Microsoft.Health.Fhir.Core/Features/SupportedSortParameterNames.cs
```diff
@@ -0,0 +1,15 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Fhir.Core.Features
+{
+    /// <summary>
+    /// Provides list of supported sort parameter names.
```
johnstairs, line 9: You need to explain why these are the only supported ones

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -455,7 +473,69 @@ public object VisitTable(TableExpression tableExpression, SearchOptions context)
                     }
 
                     break;
+                case TableExpressionKind.Sort:
```
johnstairs, line 476: It would be nice to have an optimization where we "reuse" a CTE when we are both filtering and sorting on a search parameter.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -455,7 +473,69 @@ public object VisitTable(TableExpression tableExpression, SearchOptions context)
...
+                    if (tableExpression.ChainLevel != 0)
+                    {
+                        throw new InvalidOperationException("Multiple chain level is not supported.");
```
johnstairs, line 479: More that you can't sort on a chained search parameters

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SearchOptionsExtensions.cs
```diff
@@ -32,5 +32,24 @@ public static SortOrder GetFirstSortOrderForSupportedParam(this SearchOptions se
...
+
+            foreach (var sortOptions in searchOptions.Sort)
+            {
+                if (sortOptions.searchParameterInfo.IsSortSupported())
+                {
+                    return sortOptions;
+                }
+                else
+                {
+                    throw new SearchParameterNotSupportedException(string.Format(Core.Resources.SearchParameterNotSupported, sortOptions.searchParameterInfo.Name));
```
johnstairs, line 48: It's not that the search parameter is not supported. It's that we don't support sorting by it.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SearchOptionsExtensions.cs
```diff
@@ -32,5 +32,24 @@ public static SortOrder GetFirstSortOrderForSupportedParam(this SearchOptions se
 
             return sortOrder;
         }
+
+        public static (Core.Models.SearchParameterInfo, SortOrder) GetFirstSupportedSortParam(this SearchOptions searchOptions)
+        {
+            EnsureArg.IsNotNull(searchOptions, nameof(searchOptions));
+
+            foreach (var sortOptions in searchOptions.Sort)
```
johnstairs, line 40: When there are multiple sort fields, we need to return an error saying we only support one. Unless they specify one of the supported ones and `_lastUpdated`.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SearchOptionsExtensions.cs
johnstairs: Also, we should tell them which ones are supported for the given resource type.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/MissingSearchParamVisitor.cs
johnstairs: "only" and "its"

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/SqlRootExpressionRewriter.cs
```diff
@@ -37,6 +37,10 @@ public override Expression VisitMultiary(MultiaryExpression expression, int cont
             for (var i = 0; i < expression.Expressions.Count; i++)
             {
                 Expression childExpression = expression.Expressions[i];
+
+                // The TryGetNormalizedGenerator need to figure out if we have reached a "sort" expression, and then create the relevant  TableExpression and add it to the normalizedPredicates list.
```
johnstairs, line 41: This comment seems out of place here.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SearchOptionsExtensions.cs
```diff
@@ -32,5 +32,24 @@ public static SortOrder GetFirstSortOrderForSupportedParam(this SearchOptions se
 
             return sortOrder;
         }
+
+        public static (Core.Models.SearchParameterInfo, SortOrder) GetFirstSupportedSortParam(this SearchOptions searchOptions)
```
johnstairs, line 36: Why do we have both of these methods?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -232,6 +230,34 @@ private async Task<SearchResult> SearchImpl(SearchOptions searchOptions, bool hi
...
+                                {
+                                    sortExpr = reader.GetValue("SortExpr") as DateTime?;
+                                }
+                                else if (matchCount > 0 && searchOptions.Sort?[0].sortOrder == SortOrder.Descending)
+                                {
+                                    // Otherwise keep getting this, until we got the last result for Desc
+                                    sortExpr = reader.GetValue("SortExpr") as DateTime?;
+                                }
+                            }
+                            catch
```
johnstairs, line 249: Please don't add random swallows like this.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -232,6 +230,34 @@ private async Task<SearchResult> SearchImpl(SearchOptions searchOptions, bool hi
...
 
+                        // if we have more columns, it means sort expressions were added.
+                        if (reader.FieldCount > 9)
+                        {
+                            try
+                            {
+                                // if sort is requested asc keep the first result for continuation token
+                                if (matchCount == 0 && searchOptions.Sort?[0].sortOrder == SortOrder.Ascending)
+                                {
+                                    sortExpr = reader.GetValue("SortExpr") as DateTime?;
```
johnstairs, line 241: the `if` and `else if` blocks look identical...

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -232,6 +230,34 @@ private async Task<SearchResult> SearchImpl(SearchOptions searchOptions, bool hi
                             rawResource = await streamReader.ReadToEndAsync();
                         }
 
+                        // if we have more columns, it means sort expressions were added.
+                        if (reader.FieldCount > 9)
```
johnstairs, line 234: We only need this if it's the last match row (See `if (matchCount >= searchOptions.MaxItemCount && isMatch)` above).

johnstairs: (or it was above, at least)

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -232,6 +230,34 @@ private async Task<SearchResult> SearchImpl(SearchOptions searchOptions, bool hi
...
+                            {
+                                // if sort is requested asc keep the first result for continuation token
+                                if (matchCount == 0 && searchOptions.Sort?[0].sortOrder == SortOrder.Ascending)
+                                {
+                                    sortExpr = reader.GetValue("SortExpr") as DateTime?;
+                                }
+                                else if (matchCount > 0 && searchOptions.Sort?[0].sortOrder == SortOrder.Descending)
+                                {
+                                    // Otherwise keep getting this, until we got the last result for Desc
+                                    sortExpr = reader.GetValue("SortExpr") as DateTime?;
```
johnstairs, line 246: It is a less efficient to call this instead of `reader.GetDateTime(<columnNumber>)`

src/Microsoft.Health.Fhir.Core/Features/Search/ContinuationToken.cs
itye-msft: @johnstairs I converted the logic to handle an object array as you suggested, but kept the `ContinuationToken` class to handle all the little logic around getting and setting the values in the array.
I also added backward compatibility during parsing.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/SortTests.cs
eladiw: There are other tests which includes 'sort' + 'tag filter'. in this test I wanted to check the path of sorting alone. i.e. sorting without additional filters as it creates a different query. In order to achieve this the db must be clean....

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
itye-msft: @johnstairs I have changed it now to support an object array as you requested.
I also left in tact the `BadRequestException` in case of invalid CT.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/SearchTestsBase.cs
eladiw: its a left over.. removed.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
itye-msft: @johnstairs ,
`reader.FieldCount` is about how many **columns** we got back. More than 9 means, we have `SortExpr` columns which we need to use.
However, `matchCount` and `searchOptions.MaxItemCount` relate to the **rows**, which has nothing to do with the `SortExpr` value that we are after - in this section.
Therefore, I am not sure there is much to change

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
itye-msft: They only look identical, but we are interested mainly in the `SortOrder.Ascending` and `SortOrder.Descending` which differentiates the two blocks.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
itye-msft: Done

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
itye-msft: Done

src/Microsoft.Health.Fhir.Core/Features/Search/ContinuationToken.cs
itye-msft: Done

src/Microsoft.Health.Fhir.Core/Features/SupportedSortParameterNames.cs
itye-msft: Moved. The consequence of moving this requires to initialise the supported list when the SQL service starts.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
johnstairs: I am saying that you only need to read the sortexpr column if it's the last match row and it has exceeded the row count.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
johnstairs: Not the conditions, but the statement blocks are identical (the "then" parts). Use an `||` in the condition instead of duplicating

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -232,6 +240,30 @@ private async Task<SearchResult> SearchImpl(SearchOptions searchOptions, bool hi
                             rawResource = await streamReader.ReadToEndAsync();
                         }
 
+                        // if we have more columns, it means sort expressions were added.
+                        if (reader.FieldCount > 9 && searchOptions.Sort != null && searchOptions.Sort.Count > 0)
+                        {
+                            SortOrder sortOrder = searchOptions.Sort[0].sortOrder;
+
+                            // if sort is requested asc keep the first result for continuation token
```
johnstairs, line 248: Are you using the first match? That will not have the right value for continuation.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
itye-msft: Yes, I agree, but as we have discussed before this optimization can be implemented in a future iteration.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SearchOptionsExtensions.cs
itye-msft: Agree, removed one.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
itye-msft: Done

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
itye-msft: This section was removed based on our discussion

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/SearchTestsBase.cs
itye-msft: @eladiw Can you answer?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -232,6 +248,13 @@ private async Task<SearchResult> SearchImpl(SearchOptions searchOptions, bool hi
                             rawResource = await streamReader.ReadToEndAsync();
                         }
 
+                        // See if this resource is a continuation token candidate and increase the count
+                        if (isMatch)
+                        {
+                            newContinuationId = resourceSurrogateId;
```
johnstairs, line 254: We should handle the continuation token gathering in one place, with the above.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -194,6 +208,7 @@ private async Task<SearchResult> SearchImpl(SearchOptions searchOptions, bool hi
                     long? newContinuationId = null;
                     bool moreResults = false;
                     int matchCount = 0;
+                    DateTime? sortExpr = null;
```
johnstairs, line 211: This column should not really be called `sortExpr`, because it is not an expression, it is a value. `sortValue` would be better (here and in the SQL).

Also, you should add a comment explaining why this is a `DateTime`.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -213,14 +228,15 @@ private async Task<SearchResult> SearchImpl(SearchOptions searchOptions, bool hi
...
-                        }
 
-                        // See if this resource is a continuation token candidate and increase the count
-                        if (isMatch)
-                        {
-                            newContinuationId = resourceSurrogateId;
-                            matchCount++;
+                            // At this point we are at the last row.
+                            // if we have more columns, it means sort expressions were added.
+                            if (reader.FieldCount > 9)
```
johnstairs, line 234: Be careful as a change has just gone in that added another field to the projection. Maybe it would be safer to use the column name after all.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -268,7 +291,18 @@ private async Task<SearchResult> SearchImpl(SearchOptions searchOptions, bool hi
                         unsupportedSortingParameters = searchOptions.UnsupportedSortingParams;
                     }
 
-                    return new SearchResult(resources, searchOptions.UnsupportedSearchParams, unsupportedSortingParameters, moreResults ? newContinuationId.Value.ToString(CultureInfo.InvariantCulture) : null, _isResultPartial);
+                    // Continuation token prep
+                    var continuationToken = new ContinuationToken(new object[]
```
johnstairs, line 295: We should not allocate this if `!moreResults`. 
Also, if you should allocate it you pass in either `new[] { newContinuationId }` or `new [] { sortExpr, newContinuationId }`.

src/Microsoft.Health.Fhir.Core/Features/Search/ContinuationToken.cs
```diff
@@ -0,0 +1,116 @@
...
+
+        // Currently only a single sort is implemented
+        public string SortExpr
+        {
+            get
+            {
+                return _tokens.Length > 1 ? _tokens[0] as string : null;
+            }
+
+            set
```
johnstairs, line 47: I would eliminate the setters here

src/Microsoft.Health.Fhir.Core/Features/Search/ContinuationToken.cs
```diff
@@ -0,0 +1,104 @@
...
+        {
+            return JArray.FromObject(_tokens).ToString();
+        }
+
+        public override string ToString()
+        {
+            return ToJson();
+        }
+
+        public static ContinuationToken FromString(string json)
```
johnstairs, line 58: This approach is not particularly efficient. There are two improvements that could be made.

|         Method |     Mean |     Error |    StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------------- |---------:|----------:|----------:|-------:|------:|------:|----------:|
|        Current | 4.168 us | 0.0775 us | 0.0762 us | 0.6180 |     - |     - |    3264 B |
|     Simplified | 1.811 us | 0.0297 us | 0.0445 us | 0.5398 |     - |     - |    2824 B |
| SystemTextJson | 1.140 us | 0.0219 us | 0.0269 us | 0.0458 |     - |     - |     240 B |


The "simplified" option looks like this:
``` csharp
private static JsonSerializerSettings settings = new JsonSerializerSettings
{
    DateParseHandling = DateParseHandling.None,
};


[Benchmark]
public object[] Simplified()
{
    if (long.TryParse(json, NumberStyles.None, CultureInfo.InvariantCulture, out var sid))
    {
        return new object[] { sid };
    }

    return JsonConvert.DeserializeObject<object[]>(json, settings);
}
```

Faster still, and with a lot fewer allocations, is an implementation based on System.Text.Json:
``` csharp
private static readonly JsonSerializerOptions options = new JsonSerializerOptions() { Converters = { new MyConverter() } };

[Benchmark]
public object[] SystemTextJson()
{
    if (long.TryParse(json, NumberStyles.None, CultureInfo.InvariantCulture, out var sid))
    {
        var result = new object[] { sid };
        return result;
    }

    return System.Text.Json.JsonSerializer.Deserialize<object[]>(json, options);
}

public class MyConverter : System.Text.Json.Serialization.JsonConverter<object>
{
    public override object Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options) => reader.TokenType switch
    {
        JsonTokenType.String => reader.GetString(),
        JsonTokenType.Number => reader.GetInt64(),
        _ => throw new NotSupportedException()
    };

    public override void Write(Utf8JsonWriter writer, object value, JsonSerializerOptions options) => throw new NotImplementedException();
}
```

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -274,7 +299,28 @@ private async Task<SearchResult> SearchImpl(SearchOptions searchOptions, bool hi
...
+                    // Continuation token prep
+                    ContinuationToken continuationToken = null;
+                    if (moreResults)
+                    {
+                        if (sortValue.HasValue)
+                        {
+                            continuationToken = new ContinuationToken(new object[]
+                            {
+                                newContinuationId ?? 0,
+                                string.Format("{0:o}", sortValue),
```
johnstairs, line 311: just `sortValue.ToString("o")`

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -111,7 +114,8 @@ public object VisitSqlRoot(SqlRootExpression expression, SearchOptions context)
                     StringBuilder.Append("CAST(0 AS bit) AS IsRawResourceMetaSet, ");
                 }
 
-                StringBuilder.AppendLine(VLatest.Resource.RawResource, resourceTableAlias);
+                StringBuilder.Append(VLatest.Resource.RawResource, resourceTableAlias);
+                StringBuilder.AppendLine((searchParamInfo == null || searchParamInfo.Name == KnownQueryParameterNames.LastUpdated) ? string.Empty : $", {TableExpressionName(_tableExpressionCounter)}.SortValue");
```
johnstairs, line 118: Elsewhere in the SQL codegen, we try to minimize string allocations by chaining `Append()` calls instead of using string interpolation

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -139,11 +143,21 @@ public object VisitSqlRoot(SqlRootExpression expression, SearchOptions context)
...
+                StringBuilder.Append("ORDER BY ");
+                if (searchParamInfo == null || searchParamInfo.Name == KnownQueryParameterNames.LastUpdated)
+                {
+                    StringBuilder
                     .Append(VLatest.Resource.ResourceSurrogateId, resourceTableAlias).Append(" ")
                     .AppendLine(sortOrder == SortOrder.Ascending ? "ASC" : "DESC");
+                }
+                else
+                {
+                    // TODO: should Sid1 be ordered by ASC/DESC if other sort param is requested?
```
johnstairs, line 155: Remove this todo. We will revisit when we support multiple sort parameters.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -254,12 +268,16 @@ public object VisitTable(TableExpression tableExpression, SearchOptions context)
                     break;
 
                 case TableExpressionKind.Top:
-                    var sortOrder = context.GetFirstSortOrderForSupportedParam();
+                    var (paramInfo, sortOrder) = context.GetFirstSupportedSortParam();
+                    FindRestrictingPredecessorTableExpressionIndex();
```
johnstairs, line 272: return value is discarded

src/Microsoft.Health.Fhir.Core/Features/Search/ContinuationToken.cs
itye-msft: @johnstairs ok, I used the SystemTextJson as you suggested

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
itye-msft: @johnstairs  I am not following...
Both of the return values are being used

johnstairs:
Query generation is currently broken.
```
GET /Patient/
```
``` SQL
SELECT TOP (@p0) r.ResourceTypeId, r.ResourceId, r.Version, r.IsDeleted, r.ResourceSurrogateId, r.RequestMethod, CAST(1 AS bit) AS IsMatch, CAST(0 AS bit) AS IsPartial, r.IsRawResourceMetaSet, r.RawResource, cte-1.SortValue
```

Note the `cte-1.SortValue`

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -475,7 +496,69 @@ public object VisitTable(TableExpression tableExpression, SearchOptions context)
...
 
                     break;
+                case TableExpressionKind.Sort:
+                    if (tableExpression.ChainLevel != 0)
+                    {
+                        throw new InvalidOperationException("Multiple chain level is not possible.");
+                    }
+
+                    var (searchParamInfo, searchSort) = context.GetFirstSupportedSortParam();
+                    var continuationToken = ContinuationToken.FromString(context.ContinuationToken);
```
johnstairs, line 506: Not so great that we are parsing this for a second time.

johnstairs (approved):
Approving. We will follow-up with some adjustments.

PR #1186, opened 2020-08-07
Extract new search parameter values from resources

feordin:
## Description
Adds functionality to the Reindex worker to extract the new search parameter values from the resources, then compare them to the existing values and call the persistence layer to update the values if necessary.

## Related issues
Addresses [issue [AB#73389](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/73389)].
Addresses [issue [AB#73600](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/73600)].

## Testing
Unit tests added for the method which invokes the extraction.  Also manual testing done.


src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -190,10 +194,10 @@ private async Task ProcessQueryAsync(ReindexJobQueryStatus query, CancellationTo
 
                 // TODO: Release lock on job document so another thread may pick up the next query.
 
-                await _updateIndices.ProcessSearchResultsAsync(results, _reindexJobRecord.Hash, cancellationToken);
+                var extractRequest = new ExtractReindexRequest(results, _reindexJobRecord.Hash);
+                await _mediator.Send(extractRequest, cancellationToken);
```
LTA-Thinking, line 198: Curiosity, why use the mediator instead of making an interface for the reindex utilities and calling process search results here?

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/UriSearchValue.cs
```diff
@@ -53,6 +54,23 @@ public void AcceptVisitor(ISearchValueVisitor visitor)
...
+            {
+                return false;
+            }
+
+            var uriSearchValueOther = other as UriSearchValue;
+
+            if (uriSearchValueOther == null)
+            {
+                return false;
+            }
```
LTA-Thinking, line 69: This null check and cast is used so often I feel it should be a helper method.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
feordin: That is a good suggestion.  I was simply separating the ReindexTask from the code that calls the Extractor.  You are correct an interface would be a more straightforward method of separation.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/IReindexUtilities.cs
```diff
@@ -0,0 +1,25 @@
...
+using Microsoft.Health.Fhir.Core.Features.Search;
+
+namespace Microsoft.Health.Fhir.Core.Features.Operations.Reindex
+{
+    public interface IReindexUtilities
+    {
+        /// <summary>
+        /// For each result in a batch of resources this will extract new search params
+        /// Then compare those to the old values to determine if an update is needed
+        /// Needed updates will br committed in a batch
```
LTA-Thinking, line 17: nit: typo 'br'

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Operations/Reindex/ReindexJobTaskTests.cs
```diff
@@ -8,6 +8,7 @@
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using MediatR;
```
LTA-Thinking, line 11: nit: Import is not needed

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Reindex/ReindexUtilities.cs
```diff
@@ -0,0 +1,77 @@
...
+
+            _fhirDataStoreFactory = fhirDataStoreFactory;
+            _searchIndexer = searchIndexer;
+            _deserializer = deserializer;
+        }
+
+        /// <summary>
+        /// For each result in a batch of resources this will extract new search params
+        /// Then compare those to the old values to determine if an update is needed
+        /// Needed updates will br committed in a batch
```
LTA-Thinking, line 40: nit: typo

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/UriSearchValue.cs
feordin: The equals method can be slightly different on the last line, so for simplicity I think I will leave it as is.

PR #1188, opened 2020-08-10
Support replace document in Cosmos Db

namadabu:
## Description
This PR adds a method to allow replacing an existing resource in CosmosDb with updated data. No new version of the resource is created (i.e. versionId remains the same).

TODO: Add UTs

## Related issues
Addresses [issue [AB#73394](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/73394)].

## Testing
Manual testing with local deployment and Cosmos Db emulator.


src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/StoredProcedures/Replace/replaceWithoutVersionUpdate.js
```diff
@@ -0,0 +1,94 @@
+﻿/**
+* This stored procedure can be used to replace an existing document
+* 
+* @constructor
+* @param {any} doc - The CosmosResourceWrapper to save
+* @param {string} matchVersionId - required etag to match against when replacing an existing document
+*/
+function replaceWithoutVersionUpdate(doc, matchVersionId) {
```
brendankowitz, line 8: Do we need a stored proc if we're only updating a single document?

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
```diff
@@ -236,5 +238,49 @@ public void Build(ICapabilityStatementBuilder builder)
                 builder.AddRestInteraction(SystemRestfulInteraction.Batch);
             }
         }
+
+        public async Task<ResourceWrapper> ReplaceAsync(ResourceWrapper resourceWrapper, WeakETag weakETag, CancellationToken cancellationToken)
```
brendankowitz, line 242: If in the future we can send multiple documents, should this take an IEnumerable<ResourceWrapper>?

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/StoredProcedures/Replace/replaceWithoutVersionUpdate.js
namadabu: Sorry this was supposed to be a draft PR. I am working on updating the stored procedure to update multiple docs at the same time.

src/Microsoft.Health.Fhir.Core/Features/Persistence/IFhirDataStore.cs
```diff
@@ -20,5 +20,7 @@ public interface IFhirDataStore
         Task<ResourceWrapper> GetAsync(ResourceKey key, CancellationToken cancellationToken);
 
         Task HardDeleteAsync(ResourceKey key, CancellationToken cancellationToken);
+
+        Task<ResourceWrapper> ReplaceAsync(ResourceWrapper resourceWrapper, WeakETag weakETag, CancellationToken cancellationToken);
```
feordin, line 24: For the purposes of Reindex, we don't need to replace the entire resource.  Just update the search index values, and the searchParamHash.  For CosmosDB, since it is all one document, I imagine we simply update the entire thing.  But for other persistence providers we could only send the value that we want to the DB.  Should we name this UpdateMetadataAsync?  Or UpdateSearchParams?

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
namadabu: My thought was that when we add support for updating multiple documents, we should add a separate API called `BatchReplaceAsync`? This `ReplaceAsync` method could still be used for other scenarios where we are looking to replace just a single document. I was approaching this more from a generic api for a datastore. I don't have a strong preference and am open to updating this method to take in an IEnumerable

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/StoredProcedures/Replace/replaceWithoutVersionUpdate.js
namadabu: Removed the stored procedure and updated the code to use the SDK call for replace

src/Microsoft.Health.Fhir.Core/Features/Persistence/IFhirDataStore.cs
namadabu: I think even if we do change the method name, the input parameter will still be a `ResourceWrapper` if we want to maintain consistency across the different persistence providers right? I was thinking of `Replace` as more of a generic api call that is part of the datastore. And that it could potentially be used for other use cases also. I don't have a strong opinion on this though. I noticed that there are other methods now in `IFhirDataStore` that are specific to `SearchParameter`. I will update this method name also to reflect that.

namadabu: Update name to UpdateSearchParameterForResourceAsync

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -269,5 +269,10 @@ public void Build(ICapabilityStatementBuilder builder)
                 builder.AddRestInteraction(SystemRestfulInteraction.Transaction);
             }
         }
+
+        public Task<ResourceWrapper> UpdateSearchParameterForResourceAsync(ResourceWrapper resourceWrapper, WeakETag weakETag, CancellationToken cancellationToken)
```
brendankowitz, line 273: Perhaps `UpdateSearchIndexForResourceAsync`?

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
```diff
@@ -233,6 +224,51 @@ public async Task UpdateSearchParameterIndicesBatchAsync(IReadOnlyCollection<Res
...
 
+        public async Task<ResourceWrapper> UpdateSearchParameterForResourceAsync(ResourceWrapper resourceWrapper, WeakETag weakETag, CancellationToken cancellationToken)
+        {
+            EnsureArg.IsNotNull(resourceWrapper, nameof(resourceWrapper));
+            EnsureArg.IsNotNull(weakETag, nameof(weakETag));
+
+            var cosmosWrapper = new FhirCosmosResourceWrapper(resourceWrapper);
+            ItemRequestOptions requestOptions = new ItemRequestOptions()
+            {
+                IfMatchEtag = weakETag.VersionId,
```
brendankowitz, line 235: I don't think this will work, `IfMatchEtag` refers to the `_etag` property of the Cosmos document

src/Microsoft.Health.Fhir.Core/Features/Persistence/IFhirDataStore.cs
brendankowitz: The searchIndex is what is stored on the wrapper. Good to read the context above, I'd prefer something along the lines of `UpdateSearchIndexForResourceAsync`. We also introduced a SearchParameterStatusDataStore so I'd prefer to diverge from this naming a little.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
namadabu: Updated

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
namadabu: It works currently (I have a test also that specifically tests this scenario - `GivenAnUpdatedResourceWithWrongETag_WhenUpdateSearchParameterForResourceAsync_ThenExceptionIsThrown` ). It basically depends on the caller creating the `WeakETag` with the appropriate result from the CosmosDb document. But I get your point that using the `WeakETag` class here does seem confusing (since that class is supposed to be a wrapper for version id). I have updated the interface and code to take in a string instead of `WeakETag`.

brendankowitz: Ah, ok thanks, makes sense

brendankowitz (approved):

PR #1190, opened 2020-08-10
Select container for export

LTA-Thinking:
## Description
Allows the user to select the container that data will be sent to as part of an export request.
The container is specified with the parameter _container when making the export job.

Additionally all data will now be exported to a folder in the container with the name <timestamp>-<id> to avoid overwrite issues.

## Related issues
Addresses [74055](https://microsofthealth.visualstudio.com/Health/_workitems/edit/74055)

## Testing
New unit and E2E tests have been added to cover the new scenario.

### DIFF
--- a/src/Microsoft.Health.Fhir.Api/Features/Headers/ResourceActionResultExtensions.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/Headers/ResourceActionResultExtensions.cs
@@ -29,7 +29,7 @@ namespace Microsoft.Health.Fhir.Api.Features.Headers
         public static ResourceActionResult<TResult> SetContentTypeHeader<TResult>(this ResourceActionResult<TResult> result, string contentTypeValue)
         {
             EnsureArg.IsNotNull(result, nameof(result));
-            EnsureArg.IsNotNullOrWhiteSpace(contentTypeValue);
+            EnsureArg.IsNotNullOrWhiteSpace(contentTypeValue, nameof(contentTypeValue));
 
             result.Headers.Add(HeaderNames.ContentType, contentTypeValue);
             return result;
@@ -38,7 +38,7 @@ namespace Microsoft.Health.Fhir.Api.Features.Headers
         public static ResourceActionResult<TResult> SetDateHeader<TResult>(this ResourceActionResult<TResult> result, string dateTime)
         {
             EnsureArg.IsNotNull(result, nameof(result));
-            EnsureArg.IsNotNullOrWhiteSpace(dateTime);
+            EnsureArg.IsNotNullOrWhiteSpace(dateTime, nameof(dateTime));
 
             result.Headers.Add(HeaderNames.Date, dateTime);
             return result;
--- a/src/Microsoft.Health.Fhir.Api/Features/Headers/ResourceActionResultExtensions.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/Headers/ResourceActionResultExtensions.cs
@@ -34,5 +34,14 @@ namespace Microsoft.Health.Fhir.Api.Features.Headers
             result.Headers.Add(HeaderNames.ContentType, contentTypeValue);
             return result;
         }
+
+        public static ResourceActionResult<TResult> SetDateHeader<TResult>(this ResourceActionResult<TResult> result, string dateTime)
+        {
+            EnsureArg.IsNotNull(result, nameof(result));
+            EnsureArg.IsNotNullOrWhiteSpace(dateTime);
+
+            result.Headers.Add(HeaderNames.Date, dateTime);
+            return result;
+        }
     }
 }
--- a/src/Microsoft.Health.Fhir.Api/Features/Headers/ResourceActionResultExtensions.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/Headers/ResourceActionResultExtensions.cs
@@ -34,14 +34,5 @@ namespace Microsoft.Health.Fhir.Api.Features.Headers
             result.Headers.Add(HeaderNames.ContentType, contentTypeValue);
             return result;
         }
-
-        public static ResourceActionResult<TResult> SetDateHeader<TResult>(this ResourceActionResult<TResult> result, string dateTime)
-        {
-            EnsureArg.IsNotNull(result, nameof(result));
-            EnsureArg.IsNotNullOrWhiteSpace(dateTime, nameof(dateTime));
-
-            result.Headers.Add(HeaderNames.Date, dateTime);
-            return result;
-        }
     }
 }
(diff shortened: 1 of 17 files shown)


src/Microsoft.Health.Fhir.Api/Features/Headers/ResourceActionResultExtensions.cs
```diff
@@ -34,5 +34,14 @@ public static ResourceActionResult<TResult> SetContentTypeHeader<TResult>(this R
             result.Headers.Add(HeaderNames.ContentType, contentTypeValue);
             return result;
         }
+
+        public static ResourceActionResult<TResult> SetDateHeader<TResult>(this ResourceActionResult<TResult> result, string dateTime)
+        {
+            EnsureArg.IsNotNull(result, nameof(result));
+            EnsureArg.IsNotNullOrWhiteSpace(dateTime);
```
namadabu, line 41: nit: add nameof(dateTime)

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/Models/ExportJobRecord.cs
```diff
@@ -51,6 +52,15 @@ public class ExportJobRecord : JobRecord
             Status = OperationStatus.Queued;
 
             QueuedTime = Clock.UtcNow;
+
+            if (storageAccountContainerName == null)
```
namadabu, line 56: nit: `string.IsNullOrEmpty` instead of just a null check?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ExportDataValidationTests.cs
```diff
@@ -185,6 +185,28 @@ public async Task GivenFhirServer_WhenGroupObervationDataIsExported_ThenExported
             Assert.True(ValidateDataFromBothSources(dataInFhirServer, dataFromExport));
         }
 
+        [Fact]
+        public async Task GivenFhirServer_WhenAllDataIsExportedToASpecificContainer_ThenExportedDataIsInTheSpecifiedContianer()
+        {
+            // NOTE: Azure Storage Emulator is required to run these tests locally.
+
+            string testContainer = "test-contianer";
```
namadabu, line 193: nit: typo, test-container

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -365,7 +365,10 @@ private async Task ProcessSearchResultsAsync(IEnumerable<SearchResultEntry> sear
                     else
                     {
                         // File does not exist. Create it.
-                        string fileName = resourceType + ".ndjson";
+                        string dateTime = _exportJobRecord.QueuedTime.UtcDateTime.ToString("s")
+                            .Replace("-", string.Empty, StringComparison.OrdinalIgnoreCase)
+                            .Replace(":", string.Empty, StringComparison.OrdinalIgnoreCase);
+                        string fileName = $"{dateTime}-{_exportJobRecord.Id}/{resourceType}.ndjson";
```
namadabu, line 371: Anyone who has already setup an automated workflow for export will now be broken by this change (even though they are not using the `_container` parameter for their export requests). Should we try to maintain backward compatibility and only add the folder if the customer supplies `_container` parameter? If we are fine with making this breaking change, we might want to think about how we will notify existing users.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ExportController.cs
```diff
@@ -150,13 +162,13 @@ public async Task<IActionResult> CancelExport(string idParameter)
...
-        private async Task<IActionResult> SendExportRequest(ExportJobType exportType, PartialDateTime since, string resourceType = null, string groupId = null)
+        private async Task<IActionResult> SendExportRequest(ExportJobType exportType, PartialDateTime since, string resourceType = null, string groupId = null, string containerName = null)
         {
-            CreateExportResponse response = await _mediator.ExportAsync(_fhirRequestContextAccessor.FhirRequestContext.Uri, exportType, resourceType, since, groupId, HttpContext.RequestAborted);
+            CreateExportResponse response = await _mediator.ExportAsync(_fhirRequestContextAccessor.FhirRequestContext.Uri, exportType, resourceType, since, groupId, containerName, HttpContext.RequestAborted);
 
             var exportResult = ExportResult.Accepted();
             exportResult.SetContentLocationHeader(_urlResolver, OperationsConstants.Export, response.JobId);
-
+            exportResult.SetDateHeader(response.JobTime.UtcDateTime.ToString("s"));
```
namadabu, line 171: I am wondering if there is a specific reason we want to include the job queuedtime information in this response. We already send that information as part of the response when export completes.

LTA-Thinking: That's fair. I was wanting a way for the user to get the folder name based only on the response from the create export request, but they really shouldn't be looking it up until they know it is done.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
LTA-Thinking: @brendankowitz? I think that the change is needed to keep the containers organized incase a user decides to reuse one of the GUID containers we make, and to keep the internal structure consistent between export jobs. I don't know who is the PM for OSS at this point.

ranvijaykumar: I had an offline discussion with RB on this. We can go with this: If the customer does not provide the container name, we export as per the prior behavior. If the customer provides the container name, we create folder inside the container for the export. We should document this behavior in user documentation.

src/Microsoft.Health.Fhir.Api/Features/Headers/ResourceActionResultExtensions.cs
```diff
@@ -29,10 +29,19 @@ public static ResourceActionResult<TResult> SetContentLocationHeader<TResult>(th
...
         {
             EnsureArg.IsNotNull(result, nameof(result));
-            EnsureArg.IsNotNullOrWhiteSpace(contentTypeValue);
+            EnsureArg.IsNotNullOrWhiteSpace(contentTypeValue, nameof(contentTypeValue));
 
             result.Headers.Add(HeaderNames.ContentType, contentTypeValue);
             return result;
         }
+
+        public static ResourceActionResult<TResult> SetDateHeader<TResult>(this ResourceActionResult<TResult> result, string dateTime)
```
namadabu, line 38: I don't think this is being used anywhere. Can be removed?

PR #1192, opened 2020-08-11
Updates after moving Auditing Code to Healthcare Shared Components

v-shaaal:
## Description
Initial pass after moving generic auditing code to healthcare shared components.


src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Audit/AuditEventTypeMappingTests.cs
```diff
@@ -11,8 +11,8 @@
 using Microsoft.AspNetCore.Mvc.Controllers;
 using Microsoft.AspNetCore.Mvc.Infrastructure;
 using Microsoft.AspNetCore.Mvc.RazorPages;
+using Microsoft.Health.Api.Features.Audit;
```
jackliums, line 14: These tests should move together with the code right?

v-shaaal: Yep, I will move them. Quick question, should we also keep them here in FHIR or just having them in the shared components would suffice?

jackliums: We only need these in the shared. FHIR has its own integration tests that validates the functionality already so we don't need these unit tests here.

v-shaaal: Makes sense. I will delete these tests from here. I have already pushed a new iteration to the other pull request (in healthcare shared components) with these tests added.

Issue #962, opened 2020-03-27
SearchOptions.MaxItemCountPerSearch should be configurable per server


brendankowitz:
**User story**
As a user, I want MaxItemCountPerSearch to be configurable so that the performance threshold can be determined on a per-server bases

**Acceptance criteria**
1. MaxItemCountPerSearch is a configuration setting that can be raised or lowered for a server deployment
1. If a _count is configured above this limit a BadRequest / OutOfRange is returned

**Reference**
The limit is currently coded as 100, https://github.com/microsoft/fhir-server/blob/master/src/Microsoft.Health.Fhir.Core/Features/Search/SearchOptions.cs#L55

PR #1167, opened 2020-07-22
Support _revinclude

eladiw:
## Description
1. Add support for _revinclude while limiting the number of included items
2. Added revinclude UnitTests
3. Moved some resources creations to the fixture
4. Added OperationOutcome resource if items were truncated.

## Related issues
Addresses [issue #306 ].

## Testing
E2E tests + manual

To verify I ran:

https://localhost:44348/Patient?_revinclude=Observation:subject

And verified the returned sql statement is as expected and returned result set is correct

PR #1195, opened 2020-08-13
Add MaxItemCountPerSearch  for server.

Ivanidzo4ka:
## Description
Adds MaxItemCountPerSearch configuration for server.

## Related issues
Addresses #962 .

## Testing
Through unit-tests.

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Configs/CoreFeatureConfiguration.cs
+++ b/src/Microsoft.Health.Fhir.Core/Configs/CoreFeatureConfiguration.cs
@@ -31,5 +31,10 @@ namespace Microsoft.Health.Fhir.Core.Configs
         /// Gets or sets the default value for IncludeTotal in search bundles
         /// </summary>
         public TotalType IncludeTotalInBundle { get; set; } = TotalType.None;
+
+        /// <summary>
+        /// Gets or sets the default value for maximum value for _count in search.
+        /// </summary>
+        public int MaxItemCountPerSearch { get; set; } = 100;
     }
 }
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
@@ -141,7 +141,16 @@ namespace Microsoft.Health.Fhir.Core.Features.Search
             // Check the item count.
             if (searchParams.Count != null)
             {
+                if (searchParams.Count > _featureConfiguration.MaxItemCountPerSearch)
+                {
+                    throw new BadRequestException($"_count exceeds limit configured for server ({_featureConfiguration.MaxItemCountPerSearch}).");
+                }
+
                 searchOptions.MaxItemCount = searchParams.Count.Value;
+            }
+            else
+            {
+                searchOptions.MaxItemCount = _featureConfiguration.MaxItemCountPerSearch;
             }
 
             // Check to see if only the count should be returned
--- a/src/Microsoft.Health.Fhir.Core/Features/Search/SearchOptions.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Search/SearchOptions.cs
@@ -16,7 +16,6 @@ namespace Microsoft.Health.Fhir.Core.Features.Search
     public class SearchOptions
     {
         private const int DefaultItemCountPerSearch = 10;
-        private const int MaxItemCountPerSearch = 100;
 
         private int _maxItemCount = DefaultItemCountPerSearch;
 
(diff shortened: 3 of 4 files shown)


src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
```diff
@@ -141,8 +141,17 @@ public SearchOptions Create(string compartmentType, string compartmentId, string
             // Check the item count.
             if (searchParams.Count != null)
             {
+                if (searchParams.Count > _featureConfiguration.MaxItemCountPerSearch)
+                {
+                    throw new BadRequestException($"_count exceeds limit configured for server ({_featureConfiguration.MaxItemCountPerSearch}).");
```
brendankowitz, line 146: Any error messages that could potentially be exposed to a user should be in a Resources file

Ivanidzo4ka: Thank you.
Updated.

src/Microsoft.Health.Fhir.Core/Resources.Designer.cs
```diff
@@ -816,16 +825,7 @@ internal class Resources {
                 return ResourceManager.GetString("SortNotSupported", resourceCulture);
             }
         }
-
-        /// <summary>
-        ///   Looks up a localized string similar to Include result was truncated.
-        /// </summary>
-        internal static string TruncatedIncludeMessage {
```
Ivanidzo4ka, line 823: >TruncatedIncludeMessage [](start = 31, length = 23)

@eladiw 
This resource was added in [https://github.com/microsoft/fhir-server/pull/1167](https://github.com/microsoft/fhir-server/pull/1167) but only autogenerated file was checked in without updating Resorces.resx.
I don't mind to update Resources.resx in this pr, but I'm not sure what should be the message.

Ivanidzo4ka: Since it has original message I added it to Resources.resx file

---
In reply to: [471665471](https://github.com/microsoft/fhir-server/pull/1195#discussion_r471665471) [](ancestors = 471665471)

src/Microsoft.Health.Fhir.Core/Configs/CoreFeatureConfiguration.cs
```diff
@@ -23,13 +23,23 @@ public class CoreFeatureConfiguration
...
+
+        /// <summary>
+        /// Gets or sets the default value for maximum value for _count in search.
+        /// </summary>
+        public int MaxItemCountPerSearch { get; set; } = 100;
+
+        /// <summary>
+        /// Gets or sets the default value for maximum value for _count in search.
+        /// </summary>
+        public int DefaultItemCountPerSearch { get; set; } = 10;
```
Ivanidzo4ka, line 43: >DefaultItemCountPerSearch [](start = 19, length = 25)

It wasn't part of issue, but it feels like it also should be configurable.
Also I can't find any validation logic for configuration. Is it suppose to happen during request which relies on configurations? By validation I talk about case where `MaxItemCountPerSearch` is  always not less than `DefaultItemCountPerSearch`

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -353,7 +352,7 @@ public object VisitTable(TableExpression tableExpression, SearchOptions context)
                     {
                         // In case its revinclude, we limit the number of returned items as the resultset size is potentially
                         // unbounded. we ask for +1 so in the limit expression we know if to mark at truncated...
-                        StringBuilder.Append("TOP (").Append(Parameters.AddParameter(MaxIncludedItems + 1)).Append(") ");
+                        StringBuilder.Append("TOP (").Append(Parameters.AddParameter(context.MaxItemCount + 1)).Append(") ");
```
brendankowitz, line 355: MaxItems when relating to _include_ or _revInclude_ is not the same as the number of items on the page. See: http://hl7.org/fhir/http.html#paging. E.g. you may want 10 patients (the page size) _and_ include up to n (in this case 100) of their observation records. Likely you'd need a separate configuration for this. This is one of those slightly unexpected parts of the FHIR spec.

Ivanidzo4ka: I see. Let me revert it to original code.
Do you want me to create another issue, for it, or extend current PR and introduce another option?

---
In reply to: [471918805](https://github.com/microsoft/fhir-server/pull/1195#discussion_r471918805) [](ancestors = 471918805)

brendankowitz: Since we now have the other two, I think having a separate option for this would be good

Ivanidzo4ka: Added.
I'm looking into build failures, but they test works locally, so I'm not sure what is going on.

---
In reply to: [473172836](https://github.com/microsoft/fhir-server/pull/1195#discussion_r473172836) [](ancestors = 473172836)

PR #1197, opened 2020-08-16
Add Cloud RoleName for AppInsights

tamirkamara:
## Description
When using AppInsights it is useful to know which app created different entries especially when multiple applications are taking part in the same solution (like fhir server and proxy).
This PR adds that information to the Cloud RoleName property for each tracked item sent from the server.
(please let me know if you'd like a different name)

## Related issues
Addresses #1198.

## Testing
Run locally and seen the information received in AppInsights.

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Web/CloudRoleNameTelemetryInitializer.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Web/CloudRoleNameTelemetryInitializer.cs
@@ -12,7 +12,7 @@ namespace Microsoft.Health.Fhir.Web
     {
         public void Initialize(ITelemetry telemetry)
         {
-            telemetry.Context.Cloud.RoleName = "Microsoft Fhir Server";
+            telemetry.Context.Cloud.RoleName = "Microsoft FHIR Server";
         }
     }
 }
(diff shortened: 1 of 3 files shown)


src/Microsoft.Health.Fhir.Shared.Web/CloudRoleNameTelemetryInitializer.cs
```diff
@@ -0,0 +1,18 @@
...
+using Microsoft.ApplicationInsights.Channel;
+using Microsoft.ApplicationInsights.Extensibility;
+
+namespace Microsoft.Health.Fhir.Web
+{
+    public class CloudRoleNameTelemetryInitializer : ITelemetryInitializer
+    {
+        public void Initialize(ITelemetry telemetry)
+        {
+            telemetry.Context.Cloud.RoleName = "Microsoft Fhir Server";
```
brendankowitz, line 15: The name we have in the capability statement is _Microsoft FHIR Server_. Perhaps match this string initially (https://github.com/microsoft/fhir-server/blob/master/src/Microsoft.Health.Fhir.Core/Data/R4/BaseCapabilities.json)

tamirkamara: Done @brendankowitz.

PR #1200, opened 2020-08-17
Added PathBase to allow hosting FHIR Server behind a rewriting reverse proxy on a subpath

anyname2:
## Description
When FHIR server is hosted behind a rewriting reverse proxy (on a subpath), the URLs returned by FHIR server are missing a prefix. This PR fixes this.

#### Example:
- Reverse proxy is configured on `{{host}}` to forward traffic from `/r4` to `{{fhirHost}}` (removing the `/r4` prefix). 
- `GET {{host}}/r4/Patient --> GET {{fhirHost}}/Patient`. 
- The returned URLs in `entry[].fullUrl` will be `{{host}}/Patient/{{id}}`, but should be `{{host}}/r4/Patient/{{id}}` to be valid.

This behaviour is configurable with optional configuration variable `FhirServer:PathBase`.

## Testing
All existing tests pass (Integration and E2E checked with SqlServer only). 

Starting the server with `FhirServer:PathBase`:
- Create resource request returns location header with prefix
- Create resource transaction request returns `entry[].response.location` with prefix
- Search request
  -  Returns `entry[].fullUrl` with prefix
  -  Returns `link[].url` with prefix (both `self` and `next`)

### DIFF
--- a/src/Microsoft.Health.Fhir.Api/Registration/FhirServerApplicationBuilderExtensions.cs
+++ b/src/Microsoft.Health.Fhir.Api/Registration/FhirServerApplicationBuilderExtensions.cs
@@ -7,6 +7,7 @@ using System;
 using System.Threading.Tasks;
 using EnsureThat;
 using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Options;
 using Microsoft.Health.Api.Registration;
 using Microsoft.Health.Fhir.Api.Configs;
@@ -27,13 +28,13 @@ namespace Microsoft.AspNetCore.Builder
 
             app.UseHealthChecksExtension(new PathString(KnownRoutes.HealthCheck));
 
-            var config = app.ApplicationServices.GetService(typeof(IOptions<FhirServerConfiguration>)) as IOptions<FhirServerConfiguration>;
-            EnsureArg.IsNotNull(config, nameof(config));
+            var config = app.ApplicationServices.GetRequiredService<IOptions<FhirServerConfiguration>>();
 
-            var pathBase = new PathString(config.Value.PathBase.TrimEnd('/'));
-            if (pathBase.HasValue)
+            var pathBase = config.Value.PathBase?.TrimEnd('/');
+            if (!string.IsNullOrWhiteSpace(pathBase))
             {
-                app.UseMiddleware<PathBaseMiddleware>(pathBase);
+                var pathString = new PathString(pathBase);
+                app.UseMiddleware<PathBaseMiddleware>(pathString);
             }
 
             app.UseStaticFiles();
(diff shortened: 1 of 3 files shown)


src/Microsoft.Health.Fhir.Api/Registration/FhirServerApplicationBuilderExtensions.cs
```diff
@@ -23,10 +27,56 @@ public static IApplicationBuilder UseFhirServer(this IApplicationBuilder app)
 
             app.UseHealthChecksExtension(new PathString(KnownRoutes.HealthCheck));
 
+            var config = app.ApplicationServices.GetService(typeof(IOptions<FhirServerConfiguration>)) as IOptions<FhirServerConfiguration>;
+            EnsureArg.IsNotNull(config, nameof(config));
+
+            var pathBase = new PathString(config.Value.PathBase.TrimEnd('/'));
```
anyname2, line 33: https://github.com/microsoft/fhir-server/pull/1200#issuecomment-674832683
> 
> 
> This pull request **introduces 1 alert** when merging [d904924](https://github.com/microsoft/fhir-server/commit/d9049245a5398e2e95b2450629c6434f7ea9d0f0) into [dfbb34c](https://github.com/microsoft/fhir-server/commit/dfbb34cea30bf47abdc13c175be590a37d2307c9) - [view on LGTM.com](https://lgtm.com/projects/g/microsoft/fhir-server/rev/pr-f7329449dc8d1349a6fb1a2b3f6fbec415189d7b)
> 
> **new alerts:**
> 
>     * 1 for Dereferenced variable may be null

It is a false positive (`L33`). There is a null check in `L31`

src/Microsoft.Health.Fhir.Api/Registration/FhirServerApplicationBuilderExtensions.cs
```diff
@@ -23,10 +27,56 @@ public static IApplicationBuilder UseFhirServer(this IApplicationBuilder app)
...
+            if (pathBase.HasValue)
+            {
+                app.UseMiddleware<PathBaseMiddleware>(pathBase);
```
brendankowitz, line 36: .NET also has https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.builder.usepathbaseextensions.usepathbase?view=aspnetcore-3.1, does this built-in function that looks like it does something very similar

anyname2: Correct. But it does something slightly different.
Here is an example with PathBase set to `/r4`.
| Request | PR | .NET | Comment | 
|-|-|-|-|
| `/Patient` | PathBase: `/r4`<br/>Path: `/Patient` | PathBase: `null`<br/>Path: `/Patient` | PR version is what is needed for rewriting proxy.<br/>.NET version makes no changes to this request |
| `/r4/Patient` | PathBase: `/r4`<br/>Path: `/r4/Patient`<br/>Will result in **404** | PathBase: `/r4`<br/>Path: `/Patient` | PR version assumes original request was `/r4/r4/Patient`.<br/>.NET version performs a rewrite and routes to `/Patient` route |

I can add logic to handle the second scenario the way .NET version does, but .NET version is not suitable to handle the first scenario.

Here is a link to a relevant source: https://github.com/dotnet/aspnetcore/blob/master/src/Http/Http.Abstractions/src/Extensions/UsePathBaseMiddleware.cs#L54-L74

anyname2: @brendankowitz should I make server respond on both `/` and `/{{pathBase}}`?

brendankowitz:
Sorry for leaving this open for so long, planning on looping back to this soon

brendankowitz (approved):

PR #1201, opened 2020-08-17
Update docs and combine deployment templates

LTA-Thinking:
## Description
Updates the deployment doc to explain how to deploy the FHIR server for free.
Also combines the Cosmos DB and SQL Server deployment templates into one template.

## Testing
Existing E2E tests will test both database deployment types.


samples/templates/default-azuredeploy.json
```diff
@@ -327,31 +400,47 @@
...
         },
         {
+            "condition": "[equals(parameters('solutionType'),'FhirServerCosmosDB')]",
             "type": "Microsoft.KeyVault/vaults/secrets",
             "name": "[concat(variables('serviceName'), '/CosmosDb--Host')]",
             "apiVersion": "2015-06-01",
             "properties": {
                 "contentType": "text/plain",
-                "value": "[reference(concat('Microsoft.DocumentDb/databaseAccounts/', variables('serviceName'))).documentEndpoint]"
+                "value": "[if(equals(parameters('solutionType'),'FhirServerCosmosDB'), reference(concat('Microsoft.DocumentDb/databaseAccounts/', variables('serviceName'))).documentEndpoint, '')]"
```
Ivanidzo4ka, line 409: >if(equals(parameters('solutionType'),'FhirServerCosmosDB') [](start = 27, length = 58)

nit: not an expert in templates, but would it actually be computed if condition above is false?
If not, can this line can be left as is?

Ivanidzo4ka: Same applies to other changes with condition.

---
In reply to: [516303071](https://github.com/microsoft/fhir-server/pull/1201#discussion_r516303071) [](ancestors = 516303071)

Ivanidzo4ka (approved):
:shipit:

samples/templates/default-azuredeploy.json
LTA-Thinking: The template was failing to deploy without these changes. The reason I found was that references get computed regardless of whether the section they are a part of is used or not. Since that reference was to a service that only gets deployed when Cosmos DB is being used it would fail the SQL deployment with an invalid reference error. Similar for the other changes like this.

Ivanidzo4ka:
>                 "supportsHttpsTrafficOnly": "true"

it should be bool instead of string.

---
Refers to: samples/templates/default-azuredeploy-docker.json:475 in 81591b1. [](commit_id = 81591b164561fc6003f68413614becc72b208c97, deletion_comment = False)

Ivanidzo4ka:
>                 "supportsHttpsTrafficOnly": "true"

should be bool instead of string.

---
Refers to: samples/templates/default-azuredeploy.json:457 in 81591b1. [](commit_id = 81591b164561fc6003f68413614becc72b208c97, deletion_comment = False)

Ivanidzo4ka:
I find VS code with this extension: https://marketplace.visualstudio.com/items?itemName=msazurermtools.azurerm-vscode-tools quite handy when you work with templates.

It points to errors in schema, and has some autocomplete features.

For example right now it points what app settings for web site doesn't have api-version 2015-08-01 
```
 "apiVersion": "2015-08-01",
 "name": "appsettings",
 "type": "config",
```
see https://docs.microsoft.com/en-us/azure/templates/microsoft.web/2016-08-01/sites/config-appsettings

PR #1202, opened 2020-08-19
Add capability to include supported but not fully indexed search params in search results

feordin:
## Description
Adds a header which can be included on a search request which allows searching with parameters that are not fully indexed.

## Related issues
Addresses [issue AB#73601].

## Testing
Added Unit tests.  Manually tested with following steps:
1) Run FHIR server to initialize Cosmos DB with __searchparameterstatus__ documents
2) Stop service, and find __searchparameterstatus__ for parameter with uri: "http://hl7.org/fhir/SearchParameter/Patient-identifier"
3) Update parameter status to "Supported" instead of "Enabled"
4) Search for a patient using identifier parameter.  Results should include all patients as identifier parameter is ignored as not enabled
5) Add header "x-ms-use-partial-indices" with value of "true"
6) Repeat search of patient with identifier specified.  Now only matching patient is returned as identifier parameter is honored.


src/Microsoft.Health.Fhir.Core/Features/Definition/SearchableSearchParameterDefinitionManager.cs
```diff
@@ -59,6 +72,10 @@ public SearchParameterInfo GetSearchParameter(string resourceType, string name)
             {
                 return parameter;
             }
+            else if (_fhirReqeustContextAccessor.FhirRequestContext.IncludePartiallyIndexedSearchParams && parameter.IsSupported)
```
LTA-Thinking, line 75: I think this and the one below should be 'or' cases for the original if statement sense they have the same body.

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/FhirRequestContextRouteDataPopulatingFilterAttribute.cs
```diff
@@ -20,6 +20,7 @@ public class FhirRequestContextRouteDataPopulatingFilterAttribute : ActionFilter
     {
         private readonly IFhirRequestContextAccessor _fhirRequestContextAccessor;
         private readonly IAuditEventTypeMapping _auditEventTypeMapping;
+        private const string PartiallyIndexedParamsHeaderName = "x-ms-use-partial-indices";
```
LTA-Thinking, line 23: This feels like it should be in a constants file like the known parameters.

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/SearchParameters/SearchParameterDefinitionManagerTests.cs
```diff
@@ -133,6 +139,58 @@ public async Task GivenSearchableParams_WhenGettingSearchable_ThenCorrectParamsR
...
+                    Assert.False(p3.IsSearchable);
+                });
+
+            _fhirRequestContext.IncludePartiallyIndexedSearchParams = false;
+        }
+
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public async Task GivenContextToIncludePatialIndexedParams_WhenGettingSearchableParameterByName_ThenCorrectParamReturned(bool includePartiallyIndexedParams)
```
LTA-Thinking, line 174: This feels like two separate tests since you need an if/else for the asserts.

PR #1203, opened 2020-08-19
Fix export bugs

LTA-Thinking:
## Description
Fixes two bugs in bulk export behavior:
1. Inactive members of groups are included in a group export.
2. Patients are always included in a Patient or Group export even if the _type parameter is used and doesn't specify Patient as a type to be exported.

## Related issues
Addresses [75341](https://microsofthealth.visualstudio.com/Health/_workitems/edit/75341), [75350](https://microsofthealth.visualstudio.com/Health/_workitems/edit/75350)

## Testing
New unit tests were made for scenario 1 and existing unit and E2E tests have been modified to fit the new behavior.


src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Operations/Export/GroupMemberExtractorTests.cs
```diff
@@ -236,6 +263,14 @@ private void SetUpGroupMock()
                         },
                     },
                     new Group.MemberComponent()
+                    {
+                        Entity = new ResourceReference()
+                        {
+                            Reference = _secondPatientReference,
+                        },
+                        Inactive = true,
```
namadabu, line 271: Would it be useful to add some resources/tests that also test the condition that checks for the Period Start/End element?

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Export/GroupMemberExtractor.cs
```diff
@@ -106,9 +106,9 @@ private async Task<HashSet<string>> GetGroupPatientIdsHelper(string groupId, Dat
                 if (
                     (member.Inactive == null
                     || member.Inactive == false
-                    || member.Period?.EndElement == null
-                    || member.Period?.EndElement > fhirGroupMembershipTime
                     || includeInactiveMembers)
+                    && (member.Period?.EndElement == null
```
namadabu, line 110: What's the expected behavior when includeInactiveMembers = true and the EndElements < fhirGroupMembershipTime? The current condition will exclude that member. But do they count as "inactive" member?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -294,7 +294,12 @@ private async Task UpdateJobRecordAsync(CancellationToken cancellationToken)
                     }
                 }
 
-                await ProcessSearchResultsAsync(searchResult.Results, currentBatchId, anonymizer, cancellationToken);
+                if (_exportJobRecord.ExportType == ExportJobType.All
```
namadabu, line 297: nit: would be helpful to add some comments here to explain why we don't process the search results in certain scenarios. Else it might be hard to follow what's happening.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Export/GroupMemberExtractor.cs
```diff
@@ -106,9 +106,9 @@ private async Task<HashSet<string>> GetGroupPatientIdsHelper(string groupId, Dat
                 if (
                     (member.Inactive == null
                     || member.Inactive == false
-                    || member.Period?.EndElement == null
-                    || member.Period?.EndElement > fhirGroupMembershipTime
                     || includeInactiveMembers)
```
namadabu, line 109: Also, do we need the `includeInactiveMembers` flag? Is that used anywhere?

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Export/GroupMemberExtractor.cs
LTA-Thinking: I flip-flopped on this. The spec doesn't say one way or the other, so I would say it is up to us. I left it this way to differentiate between the idea of a resource being an inactive member and a resource no longer being in the group.

namadabu: Sounds good to me.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Export/GroupMemberExtractor.cs
LTA-Thinking: No, I should remove it. I added it to give us the option in the future, but there is no planned use case for it.

PR #1204, opened 2020-08-20
Add output format to export

LTA-Thinking:
## Description
Modifies $export to support the _outputFormat parameter as per the [FHIR spec](https://hl7.org/fhir/uv/bulkdata/export/index.html#query-parameters)
This does not affect the current export functionality, no new output formats beyond ndjson are supported.

## Related issues
Addresses [68722](https://microsofthealth.visualstudio.com/Health/_workitems/edit/68722)

## Testing
New unit tests were added.


src/Microsoft.Health.Fhir.Api/Resources.resx
```diff
@@ -316,4 +316,8 @@
   <data name="TooManyConcurrentRequests" xml:space="preserve">
     <value>The maximum number of concurrent API calls on this instance was reached.</value>
   </data>
+  <data name="InvalidOutputFormat" xml:space="preserve">
+    <value>The given output format is not supported</value>
```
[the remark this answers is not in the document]
rbans96: Instead of mentioning given, can we add placeholder for the given output format?
e.g. The "{0}" output format is not supported.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ExportController.cs
```diff
@@ -199,5 +209,13 @@ private void CheckIfAnonymizedExportIsEnabled()
                 throw new RequestNotValidException(string.Format(Resources.OperationNotEnabled, OperationsConstants.AnonymizedExport));
             }
         }
+
+        private static void CheckOutputFormat(string outputFormat)
+        {
+            if (!(string.IsNullOrWhiteSpace(outputFormat) || SupportedOutputFormats.Contains(outputFormat)))
```
rbans96, line 215: Just for my understanding, are we defaulting null, empty and whitespace string to the supported output parameter?

LTA-Thinking: If a query parameter isn't given by the user it defaults to null. Whitespace would also be wrong, so I'll change it to just allow null.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ExportController.cs
```diff
@@ -199,5 +209,13 @@ private void CheckIfAnonymizedExportIsEnabled()
                 throw new RequestNotValidException(string.Format(Resources.OperationNotEnabled, OperationsConstants.AnonymizedExport));
             }
         }
+
+        private static void CheckOutputFormat(string outputFormat)
```
namadabu, line 213: nit: any reason we don't want to do the validation in the `ValidateExportRequestFilterAttribute` ? Seems like a good candidate to be part of the validation we do there.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ExportController.cs
```diff
@@ -36,6 +37,7 @@ namespace Microsoft.Health.Fhir.Api.Controllers
     [ServiceFilter(typeof(OperationOutcomeExceptionFilterAttribute))]
     public class ExportController : Controller
     {
+        private static readonly List<string> SupportedOutputFormats = new List<string>(new string[] { "application/fhir+ndjson", "application/ndjson", "ndjson" });
```
namadabu, line 40: nit: maybe HashSet instead of List? Also, should we explicitly call out the `IEqualityComparer` to determine how the items should be compared?

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ExportController.cs
LTA-Thinking: I debated this. I didn't put it in the ValidateExportRequestFilterAttribute because so far all the validation we do in there is for header values, not parameter values.

LTA-Thinking: Moved

PR #1216, opened 2020-08-21
Remove environment pipeline

LTA-Thinking:
## Description
Adds a pipeline that cleans up an environment in a bad state. This removes the need to manually delete the resource group or container if they get into a bad state.

## Testing
Run against another PR environment and it deleted both environments.

### DIFF
--- a/build/remove-environment.yml
+++ b/build/remove-environment.yml
@@ -1,0 +1,24 @@
+# DESCRIPTION: 	
+# Cleans up a PR environment if it has gotten into a bad state.
+
+trigger: none
+
+variables:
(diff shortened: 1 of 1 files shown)


LTA-Thinking:
Yes

brendankowitz (approved):

brendankowitz:
I wonder if we can trigger this from Github when a PR is closed

LTA-Thinking:
In theory we shouldn't have to. All this is doing is running the clean up steps which run when a PR build fully passes. The reason I needed it as a separate build is in case we can't get a build to pass we have a way of easily resetting the environment.

Issue #1206, opened 2020-08-21
Search Param Table can't be altered in 4.diff.sql


rotodd:
**Describe the bug**

On startup, an application startup exception is thrown:

```
Application startup exception: Microsoft.SqlServer.Management.Common.ExecutionFailureException: An exception occurred while executing a Transact-SQL statement or batch.
 ---> System.Data.SqlClient.SqlException (0x80131904): ALTER TABLE only allows columns to be added that can contain nulls, or have a DEFAULT definition specified, or the column being added is an identity or timestamp column, or alternatively if none of the prev
ious conditions are satisfied the table must be empty to allow addition of this column. Column 'Status' cannot be added to non-empty table 'SearchParam' because it does not satisfy these conditions.
   at Microsoft.SqlServer.Management.Common.ConnectionManager.ExecuteTSql(ExecuteTSqlAction action, Object execObject, DataSet fillDataSet, Boolean catchException)
   at Microsoft.SqlServer.Management.Common.ServerConnection.ExecuteNonQuery(String sqlCommand, ExecutionTypes executionType, Boolean retry)

```

The following SQL command will not work in `4.diff.sql` since non nullable columns can't be added to a table with existing data.

```
ALTER TABLE dbo.SearchParam
ADD
    Status varchar(10) NOT NULL,
    LastUpdated datetimeoffset(7) NOT NULL,
    IsPartiallySupported bit NOT NULL

```

Dev Ops bug: [#AB75439](https://microsofthealth.visualstudio.com/Health/_workitems/edit/75439)

CaitlinV39:
@rotodd was this resolved by #1219 or is there still open work on this one?

rotodd:
> @rotodd was this resolved by #1219 or is there still open work on this one?

@CaitlinV39 this was resolved in #1219. Sorry, I didn't realize the bug was still open! I will mark it as closed

PR #1219, opened 2020-08-25
SQL .diff script fix and initialization rework

rotodd:
## Description
This PR makes a few fixes to the schema version 4 files:
- Changes the columns added to the `dbo.SearchParam` to be nullable, since non-nullable columns can't be added to a table with existing rows
- Fixes entry to `dbo.SchemaVersion` to be `(4, 'started')`

This PR also reworks the initialization of the search parameter status information in `SqlServerFhirModel`. Now, search parameter status information will only be set on startup if the schema version is greater than or equal to 4.

Note: if `AutomaticUpdatesEnabled` is set to `false` and schema version 4 is applied to a fhir server running schema version 3 using the schema management tool, the initialization code in `SqlServerFhirModel` won't be run. I am planning to address this in [#AB75471](https://microsofthealth.visualstudio.com/Health/_workitems/edit/75471).

## Related issues
Addresses [#AB75439](https://microsofthealth.visualstudio.com/Health/_workitems/edit/75439) / #1206.

## Testing
Running various manual test scenarios outline [here](https://microsoft-my.sharepoint.com/:x:/g/personal/rotodd_microsoft_com1/Eed8a0F8Vi9JmQNCUNlj0dMBxRjnRcoZ7b4B_fCxX8bhkQ?e=pwMUa1).


src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SqlServerSearchParameterStatusDataStore.cs
```diff
@@ -50,7 +50,7 @@ public async Task<IReadOnlyCollection<ResourceSearchParameterStatus>> GetSearchP
                 {
                     while (await sqlDataReader.ReadAsync())
                     {
-                        (string uri, string stringStatus, DateTimeOffset lastUpdated, bool isPartiallySupported) = sqlDataReader.ReadRow(
+                        (string uri, string stringStatus, DateTimeOffset? lastUpdated, bool? isPartiallySupported) = sqlDataReader.ReadRow(
```
rbans96, line 53: Just wondering, if status in the SearchParam is null and when we try to Enum.Parse the null value it might results in 500 if not handled properly. Should we handle it as OperationOutcome?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
```diff
@@ -310,6 +315,36 @@ private static IEnumerable<SqlDataRecord> CreateSqlDataRecords(IEnumerable<Resou
...
+                        ON dbo.SearchParam.Uri = sps.Uri
+                        COMMIT TRANSACTION";
+
+                    IEnumerable<ResourceSearchParameterStatus> statuses = _filebasedSearchParameterStatusDataStore.GetSearchParameterStatuses().GetAwaiter().GetResult();
+
+                    SqlParameter tableValuedParameter = sqlCommand.Parameters.AddWithValue("@searchParamStatuses", CreateSqlDataRecords(statuses));
+                    tableValuedParameter.SqlDbType = SqlDbType.Structured;
+                    tableValuedParameter.TypeName = "dbo.SearchParamTableType_1";
+
+                    sqlCommand.ExecuteScalar();
```
rbans96, line 343: nit : Can we use ExecuteNonQuery instead of ExecuteScalar since we are not expecting any value to return

rotodd: Good catch, fixed in 67df5f3.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SqlServerSearchParameterStatusDataStore.cs
rotodd: Good question! I discussed this with @feordin a bit, and I now throw an error in 9a9827c.

These columns are nullable because they are added to `dbo.SearchParam` in V4, after there is already existing data in the table.

They should be populated as soon as they are added to the table (in `SqlServerFhirModel.Start()`). If this check fails, we should be returning a 500 because something would be broken on our end that the user could not fix.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SqlServerSearchParameterStatusDataStore.cs
```diff
@@ -23,20 +25,35 @@ internal class SqlServerSearchParameterStatusDataStore : ISearchParameterStatusD
...
 
         public async Task<IReadOnlyCollection<ResourceSearchParameterStatus>> GetSearchParameterStatuses()
         {
+            // If the search parameter table in SQL does not yet contain status columns
+            if (_schemaInformation.Current < 4)
+            {
+                // Get status information from file.
+                return await _filebasedSearchParameterStatusDataStore.GetSearchParameterStatuses();
+            }
+
```
rotodd, line 56: We need to have a  check here to support schema versions 1-3.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SqlServerSearchParameterStatusDataStore.cs
rbans96: Thanks for taking a look. It makes sense. As discussed, it is not required to throw NullReferenceException as a reduandant code in the mentioned commit.

PR #1224, opened 2020-08-26
Auditing remaining code moved to Shared Components.

v-shaaal:
## Description
Moved remaining shared auditing files to the shared components.

## Related issues
[AB#75499](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/75499)

## Testing
All the existing tests are passing.

## Todo
Once healthcare shared nugets are published with latest changes, they will be incorporated in this pull request after which this pull request will be changed from draft to a final pull request.


src/Microsoft.Health.Fhir.Core/Exceptions/RequestEntityTooLargeException.cs
```diff
@@ -10,6 +10,7 @@ namespace Microsoft.Health.Fhir.Core.Exceptions
     public class RequestEntityTooLargeException : FhirException
     {
         public RequestEntityTooLargeException()
+            : base(Resources.RequestEntityTooLargeException)
```
jackliums, line 13: I think it's fine to make the message not required to be null in the base MicrosoftHealthException. The Exception class does not require it to be non-null anyways and you don't have to repeat these messages everywhere that's not going to be used.

v-shaaal: That makes sense. I will remove the Debug.Assert from MicrosoftHealthException in shared components code.

v-shaaal: Updated.

jackliums:
I think there is no change in Resources.resx file so we should revert it.

Issue #306, opened 2018-12-27
"_include" and "_revinclude" are not supported in Search


Efreeto:
**User story**
As a server provider, I want _include and _revinclude in Search (https://www.hl7.org/fhir/search.html#revinclude) feature so that clients can run the query on the server such as "http://xyz.com/Patient?_revinclude=*".

**Acceptance criteria**
1. When clients run do "http://xyz.com/Patient?_revinclude=*", along side the resulting patients, other resource types referenced by resulting patients are returned.

Before edit:
> **Describe the bug**
> "_include" and "_revinclude" are not supported in Search
> 
> **To Reproduce**
> Steps to reproduce the behavior:
> 1. Put a Patient on server
> 2. Put an Observation that has a reference to the patient from above
> 3. Run the query on the server such as "http://xyz.com/Patient?_revinclude=*"
> 
> **Expected behavior**
> Both the Patient and the Observation are returned in a Bundle
> https://www.hl7.org/fhir/search.html#revinclude
> 
> **Actual behavior**
> Only the Patient is returned in a Bundle

hansenms:
These are features that are on our roadmap. We would probably qualify this as a "feature request" rather than a bug and add it to the backlog. Could I ask that you edit the issue to reflect this. We are working on the feature.

Efreeto:
@hansenms I have edited my original comment as you requested. Please feel free to make further edits to ease your process such as adding tags.

Efreeto:
Are we still in backlog?

ElYusubov:
@jackliums is there any progress on this feature?

hansenms:
https://github.com/microsoft/fhir-server/commit/ced75d779ce8cbb7cb90330d820f148e5cf9bd8d

hansenms:
Include for SQL server has been merged.

ElYusubov:
> Include for SQL server has been merged.

@hansenms is there a timeline for "_revinclude" support?

ellgre:
Is there a timeline for cosmosdb support?

hansenms:
`_revinclude` for SQL is on the backlog, we will pick it up after some other items (like transactions), but we have no current plans to implement `_include` or `_revinclude` for Cosmos. We could consider `_include` for SQL, but we have no current plans.

ellgre:
[...]

PR #1246, opened 2020-09-06
_include:iterate support - Phase#1

limorl:
## Description
Adding :iterate support for _include and _revinclude - Phase #1:
* Adding _include:iterate support for non recursive queries & wildcard
* Adding E2E test to cover these cases

The solution is based on using a dictionary in SqlQueryGenerator.cs that maps target resource types to ctes.
This was _include:iterate expressions selects from the relevant CTE (and not from _cteMainSelect).
A sort was added to IncludeRewrite to organize parameters so that the include:iterate expressions appear after the relevant include

It includes implementation of recursive queries, but the tests were commented out. That is sincde the _Revinclude PR introduced a new cte for IncludeLimit cte, which requires some adjustment to be discussed.

Next phases will include:
* Adding IncludeLimits to all Includes (not only to RevInclude)
* _revinclude:iterate support
* Supporting recursive iterations  

## Related issues
Addresses [issue #306 ].

## Testing
Multiple use cases were added to cover multiple cases such as:
* Simple iteration (e.g.,  /MedicationDispense?_include=MedicationDispense:prescription&_include:iterate=MedicationRequest:patient
* Multiple iterations (e.g., /MedicationDispense?_include=MedicationDispense:prescription&_include:iterate=MedicationRequest:patient&_include:iterate=Patient:general-practitioner&_include:iterate=Patient:organization)
* Wild card variations (on _include and include:iterate)
* Variations on the include source type (array to a reference of a single type, array to a reference of multiple types)

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterDefinitionManager.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterDefinitionManager.cs
@@ -107,5 +107,19 @@ namespace Microsoft.Health.Fhir.Core.Features.Definition
 
             return componentSearchParameter.Type;
         }
+
+        public bool TryGetAllReferences(string resourceType, out IEnumerable<SearchParameterInfo> references)
+        {
+            IDictionary<string, SearchParameterInfo> searchParameters = null;
+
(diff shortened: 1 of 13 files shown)


limorl:
A new PR will be submitted from the feature branch (instead of master)

PR #1247, opened 2020-09-08
Add support for Reindex request against a single resource

namadabu:
## Description
Added support for users to "reindex" a single resource. The endpoint will follow the following pattern `GET /<fhir-server>/<resourceType>/<resourceId>/$reindex`. The response will be a `Parameter` resource with the search parameters that will be supported for that resource. The actual resource itself will NOT be updated (hence the GET request). The purpose of this endpoint is for users to test that their $reindex will work as expected before kicking off an actual $reindex job.

## Related issues
Addresses [issue [AB#73618](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/73618)].

## Testing
UTs. Manual testing on local deployment.


src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Reindex/ReindexSingleResourceRequestHandler.cs
```diff
@@ -0,0 +1,93 @@
...
+            // a new search parameter has been added to the fhir server.
+            ResourceElement resourceElement = _resourceDeserializer.Deserialize(storedResource);
+            IReadOnlyCollection<SearchIndexEntry> newIndices = _searchIndexer.Extract(resourceElement);
+
+            string searchIndexString = string.Empty;
+            if (newIndices.Count > 0)
+            {
+                // A resource can have multiple values for the same search param (eg: name for Patient).
+                // Hence using a HashSet to avoid duplicate values in the response.
+                searchIndexString = string.Join(",", newIndices.Select(x => x.SearchParameter.Name).ToHashSet());
```
feordin, line 76: It appears this will return a list of the names of the search params for the resource.  I think we should return a collection of name:value pairs, so the user can see the search param and the value that it extracted from the resource.  This will help determine if the new search param is defined correctly.

feordin (commented):
I like the fact that as you stated, you can run this GET command with no side-affects and that no data on the server is altered.  Should we also have a way to run the same synchronous reindexing of one resource, and also commit the results to the server?  Perhaps a POST of the same URL?  That way a user could reindex a handful of resources and test searching against the new parameters.  We could also make this a different task/PR.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Reindex/ReindexSingleResourceRequestHandler.cs
namadabu: Updated code to return search parameter name and corresponding values for the resource as part of the `Parameter` response.

namadabu:
> I like the fact that as you stated, you can run this GET command with no side-affects and that no data on the server is altered. Should we also have a way to run the same synchronous reindexing of one resource, and also commit the results to the server? Perhaps a POST of the same URL? That way a user could reindex a handful of resources and test searching against the new parameters. We could also make this a different task/PR.

Sounds good. I'll send that out as a separate PR.

feordin (approved):

PR #1269, opened 2020-09-16
Support POST $reindex for a single resource

namadabu:
## Description
Added support for POST $reindex endpoint for a single resource where the resource will get updated in the database with the new search parameters. The endpoint will follow the following pattern POST /<fhir-server>/<resourceType>/<resourceId>/$reindex. The response will be a Parameter resource with the search parameters that will be supported for that resource. 

## Related issues
Addresses [issue [AB#75742](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/75742)].

## Testing
UTs. Manual testing on local deployment.

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Reindex/ReindexSingleResourceRequestHandler.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Reindex/ReindexSingleResourceRequestHandler.cs
@@ -28,6 +28,8 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Reindex
         private readonly ISearchIndexer _searchIndexer;
         private readonly IResourceDeserializer _resourceDeserializer;
 
+        private const string HttpPostName = "POST";
+
         public ReindexSingleResourceRequestHandler(
             IFhirAuthorizationService authorizationService,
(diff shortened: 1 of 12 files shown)


src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Reindex/ReindexSingleResourceRequestHandler.cs
```diff
@@ -28,6 +28,8 @@ public class ReindexSingleResourceRequestHandler : IRequestHandler<ReindexSingle
         private readonly ISearchIndexer _searchIndexer;
         private readonly IResourceDeserializer _resourceDeserializer;
 
+        private const string HttpPostName = "POST";
```
feordin, line 31: I think this is already defined: Microsoft.AspNetCore.Http.HttpMethods

namadabu: That's a good idea. But that would mean Core would have a dependency on AspNetCore, which we want to avoid (since this handler is in Core).

feordin (approved):

Issue #591, opened 2019-07-22
Remove the Base64 format check


jackliums:
We changed the logic to base64 encoding the continuation token to avoid some encoding issues. To maintain backwards compatibility, we added the logic to check and see if the incoming token is base 64 encoded or not and processes it differently. We should remove this logic after a few releases.

LTA-Thinking:
Additional context: Added the encoding in PR #584, the check is in SearchOptionsFactory.cs

CaitlinV39:
@LTA-Thinking is this still an issue?

LTA-Thinking:
Yes, we never came back to this. [This](https://github.com/microsoft/fhir-server/blob/5a74aa2746811e2ab53402b7d38f54f31483f158/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs#L84) section needs to be edited so that we only support base 64 encoded continuation tokens.

CaitlinV39:
AB#76206

LTA-Thinking:
In order to prevent possible interruptions in export jobs running during a system upgrade a temporary fix had to be added to #1278 so that it checks for base64 encoded CTs. This should be removed soon.

CaitlinV39:
@LTA-Thinking Is this one okay to close?

Issue #674, opened 2019-09-30
DOS on bad continuation tokens


mmacagno:
**Describe the bug**
Providing altered continuation tokens on a query (adding a few digits) will cause a Denial Of Service on the FHIR instance.

**To Reproduce**
1. Produce a FHIR query that returns a next element with continuation token, for example:

https://{{fhir_uri}}/Observation?date=2019-09-20

2. Use the next URL and verify you get additional records
3. add a few random digits at the end, for example:

https://{{fhir_uri}}/Observation?date=2019-09-20&ct=eyJ0b2tlbiI6IitSSUQ6MzRKOEFQeEJHcUpsQmdBQUFBQUFBQT09I1JUOjEjVFJDOjEwI0lTVjoyI0lFTzo2NTUzNiIsInJhbmdlIjp7Im1pbiI6IiIsIm1heCI6IkZGIn19**1122**

**Expected behavior**
A 400 error for bad continuation token

**Actual behavior**
The server will timeout and restart 
From Postman: 
error 502 >> The specified CGI application encountered an error and the server terminated the process.

I can repro consistently.

CaitlinV39:
@CaitlinV39 create an investigation story

CaitlinV39:
AB#76150

CaitlinV39:
@johnstairs can we close?

PR #1278, opened 2020-09-18
Remove base 64 check on continuation token

LTA-Thinking:
## Description
Removes the check for base 64 encoded continuation tokens as that has been the only way of getting them for over a year now.
This is technically a breaking change as anyone who uses an unencoded continuation token would now get an error response, but those tokens would have had to be cached for over a year.

## Related issues
Addresses #591 

## Testing
Existing unit and e2e tests

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
@@ -7,6 +7,7 @@ using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Net;
+using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 using EnsureThat;
@@ -27,6 +28,8 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Export
 {
     public class ExportJobTask : IExportJobTask
     {
+        private static readonly Regex Base64FormatRegex = new Regex("^([A-Za-z0-9+/]{4})*([A-Za-z0-9+/]{3}=|[A-Za-z0-9+/]{2}==)?$", RegexOptions.Compiled | RegexOptions.Singleline);
+
         private readonly Func<IScoped<IFhirOperationDataStore>> _fhirOperationDataStoreFactory;
         private readonly IScoped<IAnonymizerFactory> _anonymizerFactory;
         private readonly ExportJobConfiguration _exportJobConfiguration;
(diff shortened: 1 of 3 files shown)


brendankowitz:
Seems fine, one thought, does expose us more to #674 or other invalid CTs

LTA-Thinking:
This makes us a little more resistant to invalid CTs as they now have to be valid base64 encoded strings or an error will be thrown.

brendankowitz (approved):

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -27,6 +28,8 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Export
 {
     public class ExportJobTask : IExportJobTask
     {
+        private static readonly Regex Base64FormatRegex = new Regex("^([A-Za-z0-9+/]{4})*([A-Za-z0-9+/]{3}=|[A-Za-z0-9+/]{2}==)?$", RegexOptions.Compiled | RegexOptions.Singleline);
```
brendankowitz, line 31: @feordin since this is still behind a feature flag, do you think this check is needed?

feordin: We do not need to check any previously created continuation tokens.  The reindex functionality is not E2E functional yet (but soon!) so any created jobs would need to be re-created and run anyway.  However, We may need to add logic to encode any returned continuation tokens which come back in the search results if the search service expects them to be encoded.  I think we can simply allow this update to go through, and then test a reindex job with continuation tokens to be sure.

feordin (approved):

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -346,7 +355,13 @@ private async Task<IAnonymizer> CreateAnonymizerAsync(CancellationToken cancella
             List<Tuple<string, string>> queryParametersList = new List<Tuple<string, string>>(sharedQueryParametersList);
             if (progress.ContinuationToken != null)
             {
-                queryParametersList.Add(Tuple.Create(KnownQueryParameterNames.ContinuationToken, progress.ContinuationToken));
+                var continuationToken = progress.ContinuationToken;
+                if (!Base64FormatRegex.IsMatch(continuationToken))
+                {
+                    continuationToken = Convert.ToBase64String(System.Text.Encoding.UTF8.GetBytes(continuationToken));
```
brendankowitz, line 361: After looking through this again I'm kind of surprised we don't have a `ToBase64` string extension :)

LTA-Thinking: We could add that when we remove the check in the ExportJobRecord in a week or two.

PR #1280, opened 2020-09-18
Add multi-threading and error recovery to reindex job task

feordin:
## Description
Updates the reindex job task to created multiple threads up to a limit of max allowed concurrency.  Also adds updates to recover from a failure condition

## Related issues
Addresses [issue [AB#73390](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/73390)].

## Testing
Manual testing, and updated unit tests.

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
@@ -155,8 +155,12 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Reindex
                     {
                         lock (jobLock)
                         {
-                            var tokenSource = queryCancellationTokens[staleQuery];
-                            tokenSource.Cancel(false);
+                            // if this query has a created task, cancel it
+                            if (queryCancellationTokens.TryGetValue(staleQuery, out var tokenSource))
+                            {
+                                tokenSource.Cancel(false);
+                            }
+
                             staleQuery.Status = OperationStatus.Queued;
                             UpdateJobAsync(cancellationToken).Wait();
                         }
@@ -170,7 +174,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Reindex
                     {
                         var taskArray = queryTasks.ToArray();
                         Task.WaitAny(taskArray);
-                        var finishedTasks = queryTasks.Where(t => t.IsCompleted);
+                        var finishedTasks = queryTasks.Where(t => t.IsCompleted).ToArray();
                         foreach (var finishedTask in finishedTasks)
                         {
                             queryTasks.Remove(finishedTask);
--- a/src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
@@ -243,6 +243,10 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Reindex
                     jobSemaphore.Release();
                 }
             }
+            finally
+            {
+                jobSemaphore.Dispose();
+            }
         }
 
         private async Task<ReindexJobQueryStatus> ProcessQueryAsync(ReindexJobQueryStatus query, SemaphoreSlim jobSemaphore, CancellationToken cancellationToken)
--- a/src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
@@ -69,6 +69,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Reindex
 
             _reindexJobRecord = reindexJobRecord;
             _weakETag = weakETag;
+            object jobLock = new object();
 
             try
             {
@@ -107,7 +108,6 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Reindex
                     {
                         resourceList.UnionWith(param.TargetResourceTypes);
 
-                        // TODO: Expand the BaseResourceTypes to all child resources
                         resourceList.UnionWith(param.BaseResourceTypes);
                     }
 
@@ -126,21 +126,38 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Reindex
                     _reindexJobRecord.Count = countOnlyResults.TotalCount.Value;
 
                     // Query first batch of resources
-                    await ProcessQueryAsync(queryStatus, cancellationToken);
+                    await ProcessQueryAsync(queryStatus, jobLock, cancellationToken);
                 }
-                else
+
+                // while not all queries are finished
+                while (_reindexJobRecord.QueryList.Where(q =>
+                    q.Status == OperationStatus.Queued ||
+                    q.Status == OperationStatus.Running).Any())
                 {
-                    // check to see if queries are queued
-                    // TODO: this while loop is temporary until we multithread this task so multiple threads can
-                    // processes queries
-                    while (_reindexJobRecord.QueryList.Where(q => q.Status == OperationStatus.Queued).Any())
+                    if (_reindexJobRecord.QueryList.Where(q => q.Status == OperationStatus.Queued).Any())
                     {
                         // grab the next query from the list which is labeled as queued and run it
                         var query = _reindexJobRecord.QueryList.Where(q => q.Status == OperationStatus.Queued).OrderBy(q => q.LastModified).FirstOrDefault();
 
-                        await ProcessQueryAsync(query, cancellationToken);
+#pragma warning disable CS4014 // Suppressed as we want to continue execution and begin processing the next query while this continues to run
+                        ProcessQueryAsync(query, jobLock, cancellationToken);
+#pragma warning restore CS4014 // Because this call is not awaited, execution of the current method continues before the call is completed
+                    }
+
+                    // reset stale queries to pending
+                    var staleQueries = _reindexJobRecord.QueryList.Where(
+                        q => q.Status == OperationStatus.Running && q.LastModified < DateTimeOffset.UtcNow - _reindexJobConfiguration.JobHeartbeatTimeoutThreshold);
+                    foreach (var staleQuery in staleQueries)
+                    {
+                        lock (jobLock)
+                        {
+                            staleQuery.Status = OperationStatus.Queued;
+                            UpdateJobAsync(cancellationToken).Wait();
+                        }
                     }
                 }
+
+                await CheckJobCompletionStatus(cancellationToken);
             }
             catch (JobConflictException)
             {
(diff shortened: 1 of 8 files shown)


src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -126,96 +125,138 @@ public async Task ExecuteAsync(ReindexJobRecord reindexJobRecord, WeakETag weakE
...
+                            UpdateJobAsync(cancellationToken).Wait();
+                        }
+                    }
+
+                    await Task.Delay(_reindexJobConfiguration.QueryDelayIntervalInMilliseconds);
+
+                    if (queryTasks.Count >= reindexJobRecord.MaximumConcurrency)
+                    {
+                        var taskArray = queryTasks.ToArray();
+                        int taskIndex = Task.WaitAny(taskArray);
```
namadabu, line 166: Let's say `MaximumConcurrency = 5`. Even if all tasks are complete, we will only remove one task. That means the next loop can add only one task. Which means after this initial condition `queryTasks.Count >= reindexJobRecord.MaximumConcurrency` is met, we will only be adding/working on one task at a time.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -126,96 +125,138 @@ public async Task ExecuteAsync(ReindexJobRecord reindexJobRecord, WeakETag weakE
                     _reindexJobRecord.Count = countOnlyResults.TotalCount.Value;
 
                     // Query first batch of resources
-                    await ProcessQueryAsync(queryStatus, cancellationToken);
+                    await ProcessQueryAsync(queryStatus, jobLock, cancellationToken);
                 }
-                else
+
+                var queryTasks = new ConcurrentBag<Task>();
```
namadabu, line 131: Do we need `queryTasks` to be a ConcurrentBag? It seems like a List should be enough for the current use case.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Reindex/ReindexUtilities.cs
```diff
@@ -54,7 +57,8 @@ public async Task ProcessSearchResultsAsync(SearchResult results, string searchP
                 var resourceElement = _deserializer.Deserialize(entry.Resource);
                 var newIndices = _searchIndexer.Extract(resourceElement);
                 var newIndicesHash = new HashSet<SearchIndexEntry>(newIndices);
-                var prevIndicesHash = new HashSet<SearchIndexEntry>(entry.Resource.SearchIndices);
+                var prevIndicesHash = entry.Resource.SearchIndices != null ?
```
namadabu, line 60: `entry.Resource.SearchIndices` will never have the SearchIndexEntry information since we don't have the de-serializer for it. We can simply this code (and the below if condition) by just removing this.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -126,96 +125,138 @@ public async Task ExecuteAsync(ReindexJobRecord reindexJobRecord, WeakETag weakE
...
                         var query = _reindexJobRecord.QueryList.Where(q => q.Status == OperationStatus.Queued).OrderBy(q => q.LastModified).FirstOrDefault();
 
-                        await ProcessQueryAsync(query, cancellationToken);
+#pragma warning disable CS4014 // Suppressed as we want to continue execution and begin processing the next query while this continues to run
+                        queryTasks.Add(ProcessQueryAsync(query, jobLock, cancellationToken));
+                        _logger.LogInformation($"Reindex job task using {queryTasks.Count} threads");
+#pragma warning restore CS4014 // Because this call is not awaited, execution of the current method continues before the call is completed
+                    }
+
+                    // reset stale queries to pending
```
namadabu, line 149: We should not be resetting stale queries without having a mechanism to cancel the "stale tasks". A query could be in the `ProcessSearchResultsAsync` method and still end up being "stale". We will then update the status to Queued here and it could possibly be picked up on the next loop. Meanwhile `ProcessSearchResultsAsync` can complete and mark it as completed, thus ending up processing it twice. And we would also end up processing the continuation token from this query twice.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Reindex/ReindexUtilities.cs
feordin: That is true.  Will update this as part of the work item on possibly calculating a has of the given values.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
feordin: Updated with cancellationTokens for the created tasks.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
feordin: updated to remove multiple tasks.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
feordin: This is true, reverted to a List.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -126,96 +124,172 @@ public async Task ExecuteAsync(ReindexJobRecord reindexJobRecord, WeakETag weakE
...
-                    nextQuery.LastModified = DateTimeOffset.UtcNow;
-                    nextQuery.Status = OperationStatus.Queued;
-                    _reindexJobRecord.QueryList.Add(nextQuery);
-                }
+                    query.Status = OperationStatus.Running;
+                    query.LastModified = DateTimeOffset.UtcNow;
 
-                await UpdateJobAsync(cancellationToken);
+                    // Query first batch of resources
+                    results = ExecuteReindexQueryAsync(query, false, cancellationToken).Result;
```
namadabu, line 244: I don't think we should be combining synchronous blocks with async/await pattern. It defeats the purpose of using async/await

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -126,96 +124,172 @@ public async Task ExecuteAsync(ReindexJobRecord reindexJobRecord, WeakETag weakE
...
-                    while (_reindexJobRecord.QueryList.Where(q => q.Status == OperationStatus.Queued).Any())
+                    if (_reindexJobRecord.QueryList.Where(q => q.Status == OperationStatus.Queued).Any())
                     {
                         // grab the next query from the list which is labeled as queued and run it
                         var query = _reindexJobRecord.QueryList.Where(q => q.Status == OperationStatus.Queued).OrderBy(q => q.LastModified).FirstOrDefault();
+                        CancellationTokenSource queryTokensSource = new CancellationTokenSource();
+                        queryCancellationTokens.Add(query, queryTokensSource);
+
+#pragma warning disable CS4014 // Suppressed as we want to continue execution and begin processing the next query while this continues to run
+                        queryTasks.Add(ProcessQueryAsync(query, jobLock, queryTokensSource.Token));
```
namadabu, line 146: Why not use `queryTasks.Add(Task.Run(...))` here so that we can avoid all the synchronous blocks downstream?

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -126,96 +124,172 @@ public async Task ExecuteAsync(ReindexJobRecord reindexJobRecord, WeakETag weakE
...
             }
             catch (Exception ex)
             {
-                query.Error = ex.Message;
-
-                query.FailureCount++;
+                lock (jobLock)
+                {
+                    query.Error = ex.Message;
+                    query.FailureCount++;
```
namadabu, line 277: We seem to be double counting failures here. If `ExecuteReindexQueryAsync` throws and exception and fails, we increment the failure count there. And then when the exception is caught here, we will increment `FailureCount` again.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
feordin: Changed to use semaphore instead of lock(object) to properly enable async/await pattern.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
feordin: As ProcessQueryAsync returns a task object it should remain asynchronous now that all .Wait methods have been removed

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
feordin: Yes, removed one of the increments.

namadabu (approved):

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -126,96 +124,207 @@ public async Task ExecuteAsync(ReindexJobRecord reindexJobRecord, WeakETag weakE
                     _reindexJobRecord.Count = countOnlyResults.TotalCount.Value;
```
namadabu, line 124: We don't seem to be using this Count value anywhere. We can avoid executing a query to get the total count.

feordin: It is true, we don't use it yet.  I missed implementing an update to the progress property which uses the count to track overall total progress. For larger jobs I think it will be very useful, so I would like to keep it for now.  I will add a bug to make sure we implement the progress.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -126,96 +124,207 @@ public async Task ExecuteAsync(ReindexJobRecord reindexJobRecord, WeakETag weakE
...
                         var query = _reindexJobRecord.QueryList.Where(q => q.Status == OperationStatus.Queued).OrderBy(q => q.LastModified).FirstOrDefault();
+                        CancellationTokenSource queryTokensSource = new CancellationTokenSource();
+                        queryCancellationTokens.Add(query, queryTokensSource);
+
+#pragma warning disable CS4014 // Suppressed as we want to continue execution and begin processing the next query while this continues to run
+                        queryTasks.Add(ProcessQueryAsync(query, jobSemaphore, queryTokensSource.Token));
+#pragma warning restore CS4014 // Because this call is not awaited, execution of the current method continues before the call is completed
 
-                        await ProcessQueryAsync(query, cancellationToken);
+                        _logger.LogInformation($"Reindex job task using {queryTasks.Count} threads");
```
brendankowitz, line 149: This is not entirely true, if too many tasks are created they can be awaiting activation, before running

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -126,96 +124,207 @@ public async Task ExecuteAsync(ReindexJobRecord reindexJobRecord, WeakETag weakE
...
+                    }
+
+                    // reset stale queries to pending
+                    var staleQueries = _reindexJobRecord.QueryList.Where(
+                        q => q.Status == OperationStatus.Running && q.LastModified < DateTimeOffset.UtcNow - _reindexJobConfiguration.JobHeartbeatTimeoutThreshold);
```
brendankowitz, line 154: Prefer Clock.UtcNow for testing with times

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -126,96 +124,207 @@ public async Task ExecuteAsync(ReindexJobRecord reindexJobRecord, WeakETag weakE
...
+                    foreach (var staleQuery in staleQueries)
+                    {
+                        await jobSemaphore.WaitAsync();
+                        try
+                        {
+                            // if this query has a created task, cancel it
+                            if (queryCancellationTokens.TryGetValue(staleQuery, out var tokenSource))
+                            {
+                                tokenSource.Cancel(false);
```
brendankowitz, line 163: The the task is faulted, completed or cancelled, would this throw an exception?

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -126,96 +124,207 @@ public async Task ExecuteAsync(ReindexJobRecord reindexJobRecord, WeakETag weakE
...
+                            queryCancellationTokens.Remove(finishedTask.Result);
+                        }
+                    }
+
+                    // if our received CancellationToken is cancelled we should
+                    // pass that cancellation request onto all the cancellationTokens
+                    // for the currently executing threads
+                    if (cancellationToken.IsCancellationRequested)
+                    {
+                        foreach (var tokenSource in queryCancellationTokens.Values)
```
brendankowitz, line 196: The comment sounds like we should be using `CancellationTokenSource.CreateLinkedTokenSource` for the these

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -126,96 +124,207 @@ public async Task ExecuteAsync(ReindexJobRecord reindexJobRecord, WeakETag weakE
...
-                    while (_reindexJobRecord.QueryList.Where(q => q.Status == OperationStatus.Queued).Any())
+                    if (_reindexJobRecord.QueryList.Where(q => q.Status == OperationStatus.Queued).Any())
                     {
                         // grab the next query from the list which is labeled as queued and run it
                         var query = _reindexJobRecord.QueryList.Where(q => q.Status == OperationStatus.Queued).OrderBy(q => q.LastModified).FirstOrDefault();
+                        CancellationTokenSource queryTokensSource = new CancellationTokenSource();
+                        queryCancellationTokens.Add(query, queryTokensSource);
+
+#pragma warning disable CS4014 // Suppressed as we want to continue execution and begin processing the next query while this continues to run
+                        queryTasks.Add(ProcessQueryAsync(query, jobSemaphore, queryTokensSource.Token));
```
brendankowitz, line 146: I might have missed it, is there something limiting the max number of running tasks?

feordin: There is a check on line 179, if the count of the collection of tasks is >= the maximumconcurrency, then we Wait for any one of the tasks to complete before continuing with the loop.

PR #1281, opened 2020-09-18
Add searchParamHash to resources on upsert

namadabu:
## Description
Adding the search param hash value to resources when they are created/upserted to the store. This PR also updates the current way we calculate and store the search parameter hash. Instead of one hash for all the parameters, we are calculating and storing hashes for search parameters for each resource type. This means the current reindex query that uses the search param hash needs to be updated.

## Related issues
Addresses [issue [AB#75058](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/75058)].

## Testing
Manual testing. 
TODO: Update the reindex search query to use the new search param hash.
TODO: Add UTs


src/Microsoft.Health.Fhir.Core/Features/Definition/ISearchParameterDefinitionManager.cs
```diff
@@ -27,7 +27,7 @@ public interface ISearchParameterDefinitionManager
         /// <summary>
         /// Gets a hash value calculated from the current set of search parameters
```
LTA-Thinking, line 28: Doc need to be updated to match new behavior.

src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterDefinitionManager.cs
```diff
@@ -107,5 +113,35 @@ public ValueSets.SearchParamType GetSearchParameterType(SearchParameterInfo sear
 
             return componentSearchParameter.Type;
         }
+
+        public void UpdateSearchParameterHashMap(Dictionary<string, string> updatedSearchParamHashMap)
+        {
```
LTA-Thinking, line 118: Should there be a way to remove a hash mapping? Currently this method gives a way to add or update a mapping, but not remove one.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -291,5 +293,22 @@ private async Task UpdateJobAsync(CancellationToken cancellationToken)
...
+
+        private List<string> GetSearchParamHashesForResources(List<string> resourceList)
+        {
+            List<string> searchParamHashes = new List<string>();
+            foreach (string resource in resourceList)
+            {
+                if (!_reindexJobRecord.ResourceTypeSearchParameterHashMap.ContainsKey(resource))
+                {
+                    _logger.LogError($"Unable to find search parameter hash for resource type {resource}");
+                    throw new Exception();
```
LTA-Thinking, line 305: This should throw an exception with more information.

src/Microsoft.Health.Fhir.Core/Features/Search/SearchHelperUtilities.cs
```diff
@@ -16,22 +17,34 @@ public static class SearchHelperUtilities
...
+            EnsureArg.IsNotNull(searchParameterUris, nameof(searchParameterUris));
+            EnsureArg.IsGt(searchParameterUris.Count(), 0, nameof(searchParameterUris));
+
             StringBuilder sb = new StringBuilder();
-            foreach (ResourceSearchParameterStatus searchParameterStatus in resourceSearchParameterStatus.OrderBy(x => x.Uri.ToString()))
+            foreach (Uri uri in searchParameterUris.OrderBy(x => x.ToString()))
             {
-                sb.Append(searchParameterStatus.Uri.ToString());
-                sb.Append(searchParameterStatus.LastUpdated.ToString());
+                sb.Append(uri.ToString());
```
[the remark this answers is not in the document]
namadabu: Reverted the way we calculate the hash to the old format - with uri and lastupdated values.

src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterDefinitionManager.cs
```diff
@@ -107,5 +113,35 @@ public ValueSets.SearchParamType GetSearchParameterType(SearchParameterInfo sear
...
+            // TODO: Make this thread-safe.
+            EnsureArg.IsNotNull(updatedSearchParamHashMap, nameof(updatedSearchParamHashMap));
+
+            foreach (KeyValuePair<string, string> kvp in updatedSearchParamHashMap)
```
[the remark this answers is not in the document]
namadabu: This represents a resourceType - searchParameterHash mapping. Currently, I don't see a scenario where we will be removing supported resource types. Hence I have added only Get and Update APIs for this.

src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterDefinitionManager.cs
namadabu: The underlying dictionary represents a resourceType - searchParameterHash mapping. Currently, I don't see a scenario where we will be removing supported resource types. Hence I have added only Get and Update methods.

src/Microsoft.Health.Fhir.Core/Features/Search/SearchHelperUtilities.cs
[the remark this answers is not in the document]
feordin: That will be fine as long as changing the state from Supported to Enabled when a reindex job finishes, does not change the lastupdated value.  We can confirm this and change if necessary.

feordin (approved):

LTA-Thinking (approved):

Issue #892, opened 2020-02-19
Add support for FHIR searches of types String=>Token


brendankowitz:
**User story**
Add support for the R4 [primitive types](https://www.hl7.org/fhir/datatypes.html).

**Acceptance criteria**
1. Support is visible at the /metadata endpoint
1. Datatype is indexed correctly
1. Verify resources can be searched by version. e.g. ActivityDefinition.version with version=abc123

CaitlinV39:
AB#72424

PR #1287, opened 2020-09-21
Adds support for string to token searches

rotodd:
## Description
- Adds support for search parameters that are converted from string values to token values.

## Demos
`ActivityDefinition` resources can be searched by version:
version-search-param

Support is visible at the metadata endpoint:
version-metadata

## Related issues
Addresses [#AB72424](https://microsofthealth.visualstudio.com/Health/_workitems/edit/72424) and #892.

## Testing
Added a new unit test.


johnstairs:
Do these parameters remain disabled until reindexing is complete?

rotodd:
> Do these parameters remain disabled until reindexing is complete?

@johnstairs yes, that makes sense and I believe would be the reason for the failing tests. @brendankowitz would a good approach here be to undo changes to the `.json` files and disable the test for now?

brendankowitz:
If you update from master you shouldn't see the Crucible failures anymore.

> Do these parameters remain disabled until reindexing is complete?

@johnstairs: Yes, this is the job of the Registry. It records the state of the search parameters and tracks them in the DB. If they are supported, we will start indexing them, however they will not be searchable until the reindex job changes the status to enabled.

If you create a new FHIR Server, this support should be on from the beginning.

src/Microsoft.Health.Fhir.Core/Features/Search/Converters/StringNodeToTokenSearchValueTypeConverter.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Collections.Generic;
+using Hl7.Fhir.ElementModel;
+using Microsoft.Health.Fhir.Core.Features.Search.SearchValues;
+
+namespace Microsoft.Health.Fhir.Core.Features.Search.Converters
+{
+    /// <summary>
+    /// A converter used to convert from <see cref="string"/> to a list of <see cref="TokenSearchValue"/>.
+    /// </summary>
+    public class StringNodeToTokenSearchValueTypeConverter : FhirNodeToSearchValueTypeConverter<TokenSearchValue>
```
brendankowitz, line 16: Also until we remove them (hopefully in soon), we should implement this in the old converters too:
https://github.com/microsoft/fhir-server/tree/master/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Converters

rotodd:
@brendankowitz as discussed, I will hold off merging this PR until the SQL search parameter registry changes are merged (PR #1262).

src/Microsoft.Health.Fhir.Core/Features/Search/Converters/StringNodeToTokenSearchValueTypeConverter.cs
rotodd: Thanks for the direction. Added them in eaea80f.

feordin (approved):

PR #1288, opened 2020-09-21
Changes to kick-off, but no longer rely on Crucible test coverage

brendankowitz:
## Description
Keeps the API to kick off Crucible test runs, but no longer waits to verify coverage

## Testing
All other E2E tests should run without impact


feordin (approved):

LTA-Thinking:
Just to confirm my understanding: this will only run during the CI build and will only kick off the crucible test run, it won't wait for it to finish and won't care if it fails?

brendankowitz:
@LTA-Thinking yes. It won't wait or validate any results, but we should still be able to see them in the Crucible dashboard

LTA-Thinking (approved):

PR #1292, opened 2020-09-22
Updates history _at parameter as unsupported

rbans96:
## Description
Since the history _at parameter behavior is not as expected so we will no longer support it until the fix is made.
This PR drops the _at param from url and returns the history results.

## Related issues
Addresses [#AB75665](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Resolute/Stories/?workitem=75665)

## Testing
Added a e2e test

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Search/SearchService.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Search/SearchService.cs
@@ -139,7 +139,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Search
 
             if (at != null)
             {
-                queryParameters.Add(Tuple.Create(SearchParameterNames.LastUpdated, at.ToString()));
+                queryParameters.Add(Tuple.Create(KnownQueryParameterNames.At, at.ToString()));
             }
             else
             {
--- a/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/HistoryTests.cs
+++ b/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/HistoryTests.cs
@@ -6,6 +6,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Net;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Web;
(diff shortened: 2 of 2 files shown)


brendankowitz:
Adding a link to #1183, which will need to take this into consideration

brendankowitz:
@CaitlinV39 this will have an impact on existing functionality.

Current behavior:
_at sometimes returns the wrong results as noted in #1084

New behavior:
_at will be declared as 'unsupported' and dropped from the url

Notes:
* The fix for this in CosmosDB is more complicated and may require re-indexing/updating existing data
* We are working on a fix for SQL Server

CaitlinV39:
@brendankowitz is the first step to disable and that will go live and then we will resolve first for SQL and then for Cosmos once we have a plan of how to do that?

src/Microsoft.Health.Fhir.Core/Features/Search/SearchService.cs
```diff
@@ -139,7 +139,7 @@ protected SearchService(ISearchOptionsFactory searchOptionsFactory, IFhirDataSto
 
             if (at != null)
             {
-                queryParameters.Add(Tuple.Create(SearchParameterNames.LastUpdated, at.ToString()));
+                queryParameters.Add(Tuple.Create(KnownQueryParameterNames.At, at.ToString()));
```
LTA-Thinking, line 142: Since at will not be supported, should it still be mutually exclusive with since & before?

LTA-Thinking (approved):

src/Microsoft.Health.Fhir.Core/Features/Search/SearchService.cs
rbans96: Good question. As John has mentioned in one of my older PR, that we should support _At and _since/_before together. So I have created UserStory #[76521](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Resolute/Stories/?workitem=76521) for this. And there is an open [PR ](https://github.com/microsoft/fhir-server/pull/1183)to enable _at parameter for SqlServer. So considering these, I have not changed _at and _since/_before relation as part of this PR.

Issue #1295, opened 2020-09-23
Search with reference:missing and any other parameter results in invalid SQL query


anyname2:
**Describe the bug**
When making a search requst with multiple parameters and with `:missing` modifier on any reference search parameter, the server will return 500 with invalid SQL syntax error. 

The syntax error comes from: [SqlQueryGenerator.cs#L249](https://github.com/microsoft/fhir-server/blob/master/src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs#L249). The reference search parameter is the first CTE, and after 1 is subtracted it results in:
```sql
SELECT Sid1 FROM cte-1
```
<details>
<summary>Full query:</summary>

```sql
WITH cte0 AS
(
    SELECT Sid1 FROM cte-1
    WHERE Sid1 NOT IN
    (
        SELECT ResourceSurrogateId
        FROM dbo.ReferenceSearchParam
        WHERE IsHistory = 0
            AND ResourceTypeId = @p0
            AND SearchParamId = @p1

    )
),
cte1 AS
(
    SELECT ResourceSurrogateId AS Sid1
    FROM dbo.TokenSearchParam
    WHERE IsHistory = 0
        AND ResourceSurrogateId IN (SELECT Sid1 FROM cte0)
        AND ResourceTypeId = @p0
        AND SearchParamId = @p2
        AND SystemId IN (SELECT SystemId FROM dbo.System WHERE Value = @p3)
        AND Code = @p4

),
cte2 AS
(
    SELECT DISTINCT TOP (@p5) Sid1, 1 AS IsMatch, 0 AS IsPartial
    FROM cte1
    ORDER BY  Sid1 ASC
)
SELECT r.ResourceTypeId, r.ResourceId, r.Version, r.IsDeleted, r.ResourceSurrogateId, r.RequestMethod, CAST(IsMatch AS bit) AS IsMatch, CAST(IsPartial AS bit) AS IsPartial, r.IsRawResourceMetaSet, r.RawResource
FROM dbo.Resource r
INNER JOIN cte2
ON r.ResourceSurrogateId = cte2.Sid1
ORDER BY r.ResourceSurrogateId ASC
```

</details>

**To Reproduce**
1. `GET /Observation?code=http://loinc.org|2708-6&based-on:missing=true`

**Expected behavior**
Server should return results matching search criteria *(in the example observations with mathicng code and missing based-on property)*

[...]

PR #1296, opened 2020-09-23
Fixed missing modifier search on references

anyname2:
## Description
Changed NormalizedPredicate SortOrder to order MissingSearchParameterExpression after all other expressions.
Changed MissingSearchParamVisitor to create an `All` table expression if the MissingSearchParameterExpression is the first in the expression list (rather than the onle one)

## Related issues
Addresses #1295.

## Testing
Added unit and E2E tests

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer.UnitTests/Features/Search/Expressions/MissingSearchParamVisitorTests.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer.UnitTests/Features/Search/Expressions/MissingSearchParamVisitorTests.cs
@@ -1,0 +1,78 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+using Microsoft.Health.Fhir.Core.Features.Search.Expressions;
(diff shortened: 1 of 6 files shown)


johnstairs:
@anyname2 , thanks for fixing this! I think there is still an issue with `_include`/`_revinclude`, for example:
`/Observation?_include=Observation:patient&based-on:missing=true`

anyname2:
> @&#8203;anyname2 , thanks for fixing this! I think there is still an issue with `_include`/`_revinclude`, for example:
> `/Observation?_include=Observation:patient&based-on:missing=true`

@johnstairs, I fixed `_include` by reordering Include Expression after MissingSearchParameterExpression. But why are include expressions sorted separately ([IncludeRewriter.cs#L32-L42](https://github.com/microsoft/fhir-server/blob/master/src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/IncludeRewriter.cs#L32-L42))?

johnstairs:
> @johnstairs, I fixed `_include` by reordering Include Expression after MissingSearchParameterExpression. But why are include expressions sorted separately ([IncludeRewriter.cs#L32-L42](https://github.com/microsoft/fhir-server/blob/master/src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/IncludeRewriter.cs#L32-L42))?

I agree, it doesn't need to be that way and is confusing.

johnstairs (approved):

johnstairs:
Thank you @anyname2!

Issue #1286, opened 2020-09-21
Search with _lastUpdated less than with millisecond precision yields missing results with SQL Datastore


anyname2:
**Describe the bug**
When searching for `_lastUpdated=lt2020-09-21T14:22:37.123`, the result does not contain all resources updated 1 millisecond prior (`2020-09-21T14:22:37:122`). 

[LastUpdatedToResourceSurrogateIdRewriter.cs#L64-L70](https://github.com/microsoft/fhir-server/blob/master/src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/LastUpdatedToResourceSurrogateIdRewriter.cs#L64-L70) contains a special case for this bug. I believe the below patch would fix this, but I am not sure why `LessThanOrEqual` is used in both instances in that `case`:
```diff
case BinaryOperator.LessThan:
    if (original == truncated)
    {
-        return Expression.LessThanOrEqual(
+        return Expression.LessThan(
            SqlFieldName.ResourceSurrogateId,
            null,
-            ResourceSurrogateIdHelper.LastUpdatedToResourceSurrogateId(truncated.AddTicks(-TimeSpan.TicksPerMillisecond)));
+            ResourceSurrogateIdHelper.LastUpdatedToResourceSurrogateId(truncated));
    }
```

**To Reproduce**
1. `GET /Observation?_lastUpdated=ge2020-09-21T09:41:07.195&_lastUpdated=lt2020-09-21T09:41:07.196`
2. Observe the generated SQL query
    ```sql
    SELECT r.resource_type_id, r.resource_id, r.version, r.is_deleted, r.resource_surrogate_id, r.request_method, true AS is_match, false AS is_partial, r.raw_resource
    FROM resource r
    WHERE resource_type_id = @p0
        AND resource_surrogate_id >= @p1
        AND resource_surrogate_id <= @p1
        AND is_history = false
        AND is_deleted = false
    ORDER BY r.resource_surrogate_id ASC
    LIMIT (@p2)
    ```
    *notice the `@p1` in both comparisons*

**Expected behavior**
All Observations with last modified date equal to `2020-09-21T09:41:07.195` to be returned.

**Actual behavior**
Only a single observation (with `tick` component equal to 0) is returned.

PR #1298, opened 2020-09-24
Fix incorrect _lastUpdated to ResourceSurrogateId translation

johnstairs:
## Description
When we get a search expression over `_lastUpdated`, we turn it into an expression over the `ResourceSurrogateId` field. This field encodes the datetime truncated to millisecond precision, with lower bits used to make the value unique.

We were not always doing the translation correctly around millisecond boundaries. 

## Related issues
Addresses #1286

## Testing
Added unit tests.


brendankowitz (approved):

johnstairs:
Merging despite the R4 checks failing. When we attempt to grant the FHIR server access to the storage account in the ARM template, we get this error:
> A hash conflict was encountered for the role Assignment ID. Please use a new Guid. (Code:HashConflictOnDifferentRoleAssignmentIds)

The GUID we generate based on the PR name, when hashed, somehow collides with another role assignment.

PR #1299, opened 2020-09-24
Update search params when reindexjob completed

feordin:
## Description
Notify the searchparameterdefinition manager that the reindexjob is completed.  Set the searchparams to Enabled and searchable.  Also persist the change.

## Related issues
Addresses [issue [AB#75846](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/75846)].

## Testing
Unit tests updated, and manual testing done


src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParametersUpdatedHandler.cs
```diff
@@ -0,0 +1,58 @@
...
+            _searchParameterRegistry = searchParameterRegsitry;
+        }
+
+        public Task Handle(SearchParametersHashUpdated notification, CancellationToken cancellationToken)
+        {
+            _searchParameterDefinitionManager.SearchParametersHash = notification.HashValue;
+            return Task.CompletedTask;
+        }
+
+        public Task Handle(ReindexJobCompleted notification, CancellationToken cancellationToken)
```
namadabu, line 38: nit: I think it might be better to have a separate handler class for this message.

src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParametersUpdatedHandler.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            var searchParameterStatusList = new List<ResourceSearchParameterStatus>();
+            foreach (string uri in notification.SearchParameterUrls)
+            {
+                var searchParamUri = new Uri(uri);
+                var searchParam = _searchParameterDefinitionManager.GetSearchParameter(searchParamUri);
+                searchParam.IsSearchable = true;
```
namadabu, line 45: Should we also update `IsSupported` value or will that be done by a different workflow?

src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParametersUpdatedHandler.cs
```diff
@@ -0,0 +1,58 @@
...
+            return Task.CompletedTask;
+        }
+
+        public Task Handle(ReindexJobCompleted notification, CancellationToken cancellationToken)
+        {
+            var searchParameterStatusList = new List<ResourceSearchParameterStatus>();
+            foreach (string uri in notification.SearchParameterUrls)
+            {
+                var searchParamUri = new Uri(uri);
+                var searchParam = _searchParameterDefinitionManager.GetSearchParameter(searchParamUri);
```
namadabu, line 44: nit: Using a Get call in order to update the `SearchParameterInfo` value in `SearchParameterDefinitionManager` seems a little unintuitive. And can cause issues in the future when multiple callers are using this at the same time. Maybe we should add an API call for updating the search param info?

src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParametersUpdatedHandler.cs
```diff
@@ -0,0 +1,58 @@
...
+using MediatR;
+using Microsoft.Health.Fhir.Core.Features.Search.Registry;
+using Microsoft.Health.Fhir.Core.Messages.Reindex;
+using Microsoft.Health.Fhir.Core.Messages.Search;
+
+namespace Microsoft.Health.Fhir.Core.Features.Definition
+{
+    public class SearchParametersUpdatedHandler : INotificationHandler<SearchParametersHashUpdated>, INotificationHandler<ReindexJobCompleted>
+    {
+        private SearchParameterDefinitionManager _searchParameterDefinitionManager;
```
namadabu, line 20: can this be readonly? And any reason we don't want to use the interface here?

feordin: Yes, it can be readonly, I will update that.  I am using the concrete class rather than the interface because the interface doesn't expose the ability to set the SearchParametersHash.  I didn't want to add that to the interface, because that should only be able to be updated here, by the handler.  All other classes that access the SearchParameterDefinition manager using the interface won't have the set access.

src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParametersUpdatedHandler.cs
feordin: These search parameters should already be marked as IsSupported, or we wouldn't be running a reindex operation with them.  We could verify that is still the case before doing the update.

src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParametersUpdatedHandler.cs
feordin: I will add an API.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobCompletedHandler.cs
```diff
@@ -0,0 +1,77 @@
...
+            ISupportedSearchParameterDefinitionManager searchParameterDefinitionManager,
+            ISearchParameterRegistry searchParameterRegistry,
+            IFhirAuthorizationService fhirAuthorizationService)
+        {
+            EnsureArg.IsNotNull(searchParameterDefinitionManager, nameof(searchParameterDefinitionManager));
+            EnsureArg.IsNotNull(searchParameterRegistry, nameof(searchParameterRegistry));
+            EnsureArg.IsNotNull(fhirAuthorizationService, nameof(fhirAuthorizationService));
+
+            _searchParameterDefinitionManager = searchParameterDefinitionManager;
+            _searchParameterRegistry = searchParameterRegistry;
```
namadabu, line 38: Do we need the SearchParameterRegistry in this class?

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobCompletedHandler.cs
```diff
@@ -0,0 +1,77 @@
...
+            _authorizationService = fhirAuthorizationService;
+        }
+
+        public async Task<ReindexJobCompletedResponse> Handle(ReindexJobCompletedRequest message, CancellationToken cancellationToken)
+        {
+            if (await _authorizationService.CheckAccess(DataActions.Reindex) != DataActions.Reindex)
```
namadabu, line 44: Just curious how the authorization flow will look like here since this request is being generated internally

namadabu (approved):

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobCompletedHandler.cs
feordin: You are correct, this is a problem.  I added in the security check because we have a unit test which runs against all classes which implement IRequestHandler and it failed without the check.  However, this authorization does not work on internal call.  I had turned off security in appsettings, so I missed it.  I decided it makes more sense to change the call to a simple method call, rather than a mediatr message.  Which avoids the problem.

Issue #1276, opened 2020-09-18
Following Quickstart instructions yields non-functional deployment.


robwafle:
**Describe the bug**
Following Quickstart instructions yields non-functional deployment.

**To Reproduce**
1. Follow the quickstart instructions here: https://docs.microsoft.com/en-us/azure/healthcare-apis/fhir-oss-portal-quickstart

**Expected behavior**
https://SERVICENAME.azurewebsites.net/metadata should return a capability statement.

**Actual behavior**
https://SERVICENAME.azurewebsites.net/metadata returns ERR_INVALID_RESPONSE.

Looking inside of app insights, I see this error:
```The MetadataAddress or Authority must use HTTPS unless disabled for development by setting RequireHttpsMetadata=false```
 
But, I am using https for my request.  Does something inside the service need to use https?  If so, shouldn't https be used by default, so the template works as-is?  

I haven't made any configuration changes.

robwafle:
I was able to get a successful deployment after many tries.

It seems if I don't check the last option, "Sql Schema Auto Updates Enabled" SQL tables are missing and I get: HTTP Error 500.30 - ANCM In-Process Start Failure when trying to access the /metadata endpoint.

It seems if I specify Security Authentication Authority, Security Authentication Audience and Enable Aad Smart On Fhir Proxy, that's when I get the ERR_INVALID_RESPONSE and The MetadataAddress or Authority must use HTTPS unless disabled for development by setting RequireHttpsMetadata=false in the logs.

CaitlinV39:
AB#76138

CaitlinV39:
@robwafle thank you so much for identifying this. We are updating Sql Schema Auto Updates Enabled to be a required field and also updating our help text to address this. In a production scenario you would set this to false and use our Schema Migration tool but in a test/dev environment true should be fine. I will post here once this is resolved. We can also add more details into our OSS guide to help with the optional fields.

CaitlinV39:
@robwafle I'm updating the documentation to make sure that this is more clear. I'm going to close this as a duplicate against the other issue but will combine all doc issues from this as well into that update.

Issue #1297, opened 2020-09-24
DB schema is not creating automatically


eladiw:
**Describe the bug**
using latest build (stu3_fhir-server:build-20200912-1) on a new AKS cluster + new,empty DB does not create the DB schema and pods fail to start

**To Reproduce**
1.) Create an empty database
2.) Deploy the Fhir Server image with tag stu3_fhir-server:build-20200912-1, using environment variables:
SqlServer__AllowDatabaseCreate: true
SqlServer__ConnectionString: empty-db-conn-string
3.) The container fails to start. See screenshot attachment showing StackTrace

4.) To work around the issue, we deploy an older image tag stu3_fhir-server:build-20200819-1 and the schema is created
![image](https://user-images.githubusercontent.com/13205761/94113477-9beb9300-fe4f-11ea-880d-1e19969eda4b.png)

**Expected behavior**
DB schema is expected to be created automatically

**Actual behavior**
pods fail to start. using a workaround we managed to get the pods to work (started with lower version and upgraded)

brendankowitz:
@eladiw thanks for highlighting this. I think there is a second setting `SqlServer__SchemaOptions__AutomaticUpdatesEnabled` which is now required for the schema updates to run. We'll use this ticket to track cleaning this up

CaitlinV39:
AB#76268

balteravishay:
FYI, this issue is also affecting the standard deployment to Azure of FHIR server OSS. 
the ARM template sets this parameter to false, causing the solution to initially be at at a fault state, until the config is changed and the app is restarted.

https://github.com/microsoft/fhir-server/blob/7fab70ab5c350926a420e29f56a7046c57c50037/samples/templates/default-azuredeploy.json#L149

rbans96:
> FYI, this issue is also affecting the standard deployment to Azure of FHIR server OSS.
> the ARM template sets this parameter to false, causing the solution to initially be at at a fault state, until the config is changed and the app is restarted.
> 
> https://github.com/microsoft/fhir-server/blob/7fab70ab5c350926a420e29f56a7046c57c50037/samples/templates/default-azuredeploy.json#L149

The default value for sqlSchemaAutomaticUpdatesEnabled is false only for CosmosDB server and for SQL Server, the default value is null so that we can choose true/false value as per our requirement.

CaitlinV39:
The resolution for this has been merged so closing this item.

PR #1301, opened 2020-09-24
Updates template for automatic SQL schema updates flag

rbans96:
## Description
Updates template for automatic SQL schema updates flag to default it to empty and make mandatory to select either of the value(true/false). The description of the flag is also updated.

## Related issues
Addresses #[AB76233](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Resolute/Stories/?workitem=76233) and addresses #1276  

## Testing
Tested by customizing the template on azure portal

### DIFF
--- a/samples/templates/default-azuredeploy-docker-sql.json
+++ b/samples/templates/default-azuredeploy-docker-sql.json
@@ -147,9 +147,8 @@
         },
         "sqlSchemaAutomaticUpdatesEnabled": {
             "type": "bool",
-            "defaultValue": false,
             "metadata": {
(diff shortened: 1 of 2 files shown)


CaitlinV39 (commented):
Should we add in the help text that you should use the schema migration tool if you set to false?

CaitlinV39:
Related to #1297

brendankowitz:
Does this also effect docker container template?

rbans96:
> Does this also effect docker container template?

Thanks, updated the docker template as well.

brendankowitz (approved):

Issue #1274, opened 2020-09-17
Deployment Error following quickstart instructions


robwafle:
**Describe the bug**
Deployment Error following quickstart instructions

**To Reproduce**
1.  Follow the quickstart instructions here: https://docs.microsoft.com/en-us/azure/healthcare-apis/fhir-oss-portal-quickstart

**Expected behavior**
No deployment errors.  I'm able to start quickly following the quick start instructions.

**Actual behavior**
Deployment Error of: "Cannot find ServerFarm with name *serverfarm-name-from-template-input*.  Click here for details ->" was returned.

robwafle:
Looks like this was the issue.

App Service Plan Resource Group and App Service Plan are OPTIONAL, and could perhaps be renamed to be prefixed with "Use Existing" in the field name to make template use a bit more straightforward.   I was able to figure this out using the tool tip text, but it was not immediately obvious. 

Thx

Rob

robwafle:
Also, I'd like to mention that it has been 8 minutes since the deployment, and the /metadata endpoint is not responding.  Eight minutes seems a bit longer than "a minute or so" as described on the quickstart page.  So, either something has still gone wrong, or maybe it takes longer for this to warm up.   I deployed to the default size on the template of S1.

CaitlinV39:
@robwafle thank you for reporting this. I have started making some updates to the quickstart guide to make it more clear and will let you know once its updated.

CaitlinV39:
AB#76139

CaitlinV39:
@robwafle I have included the QuickStart guide in our OSS Documentation now and added some updates that will hopefully make it clearer. Let me know if you have any additional questions!

https://github.com/microsoft/fhir-server/blob/master/docs/QuickstartDeployPortal.md

PR #1303, opened 2020-09-25
Created Quickstart for OSS Portal Deployment and updated a couple other docs

CaitlinV39:
## Description
Moving the quickstart guides to the OSS repo and adding the images for them. Also removed some duplicate text from Bulk Export that is covered in PaaS documentation so we don't have to keep it up to date in two places and added some links to the search page.

## Related issues
Addresses #1274


CaitlinV39:
@brendankowitz or @LTA-Thinking wondering if you could take a look at my changes. Let me know if someone else should be added as a reviewer!

LTA-Thinking:
Why are we removing the documentation on connecting with Postman? I know we have a Microsoft docs page that covers that workflow, could we link to that?

LTA-Thinking:
In the default deployment doc we switch between saying a Cosmos DB or SQL instance will be deployed. The button on the page deploys SQL, but the command line instructions deploy Cosmos DB. I think it need to be clearer that there are different templates for the different data stores.

brendankowitz:
Looks good, I also have this random wiki page for getting started with dev, let me know if you wanted me to move this: https://github.com/microsoft/fhir-server/wiki/Build-and-debug-locally

brendankowitz:
> In the default deployment doc we switch between saying a Cosmos DB or SQL instance will be deployed. The button on the page deploys SQL, but the command line instructions deploy Cosmos DB. I think it need to be clearer that there are different templates for the different data stores.

I think our wording should be along the lines of we recommend starting with the sql deployment as it offers the most features. However deployments in cosmos are still available...

brendankowitz (approved):

CaitlinV39:
> > In the default deployment doc we switch between saying a Cosmos DB or SQL instance will be deployed. The button on the page deploys SQL, but the command line instructions deploy Cosmos DB. I think it need to be clearer that there are different templates for the different data stores.
> 
> I think our wording should be along the lines of we recommend starting with the sql deployment as it offers the most features. However deployments in cosmos are still available...

I updated this and also changed the powershell to reference the SQL template instead. LMK what you think

docs/DefaultDeployment.md
```diff
@@ -68,7 +70,7 @@ The template can also be deployed using PowerShell. Here is an example of how th
 $rg = New-AzResourceGroup -Name "RG-NAME" -Location westus2
 
 New-AzResourceGroupDeployment `
--TemplateUri "https://raw.githubusercontent.com/Microsoft/fhir-server/master/samples/templates/default-azuredeploy.json" `
+-TemplateUri "https://raw.githubusercontent.com/Microsoft/fhir-server/master/samples/templates/default-azuredeploy-sql.json" `
```
brendankowitz, line 73: I don't think this will just work as sql requires extra/different parameters, e.g. sql server password

LTA-Thinking: I think the two that would be required would be sql server password and sql schema automatic updates enabled.

CaitlinV39: Thanks! I reverted and just made it clear that its Cosmos

LTA-Thinking (approved):

PR #1305, opened 2020-09-29
Improve inner-loop docker build performance

johnstairs:
Improves the Dockerfile to enable faster inner-loop builds. We previously did not have a `.dockerignore` file, which can result in a large amount of data (build output) being copied around. We also did not do restores and publish in separate layers. So previously, on my box, a docker build would always take around **3 minutes, 30 seconds**. Now, the times are:

**Initial build (or with `--no-cache`):** 1 minute, 23 seconds
**Change a `.csproj` file:** 1 minute, 14 seconds
**Change a `.cs` file:** 32 seconds
**No changes:** 1 second

Note also that the Dockerfile does not include any test projects.

### DIFF
--- a/build/docker/Dockerfile
+++ b/build/docker/Dockerfile
@@ -49,9 +49,6 @@ COPY ./src/Microsoft.Health.Fhir.Shared.Web/Microsoft.Health.Fhir.Shared.Web.pro
 COPY ./src/Microsoft.Health.Fhir.Shared.Core/Microsoft.Health.Fhir.Shared.Core.projitems \
      ./src/Microsoft.Health.Fhir.Shared.Core/Microsoft.Health.Fhir.Shared.Core.projitems
 
-COPY ./src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Microsoft.Health.Fhir.Shared.Core.UnitTests.projitems \
-     ./src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Microsoft.Health.Fhir.Shared.Core.UnitTests.projitems
-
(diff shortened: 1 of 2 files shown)


build/docker/Dockerfile
```diff
@@ -2,12 +2,69 @@ FROM mcr.microsoft.com/dotnet/core/sdk:3.1.401-bionic AS build
...
+     ./src/Microsoft.Health.Fhir.${FHIR_VERSION}.Web/Microsoft.Health.Fhir.${FHIR_VERSION}.Web.csproj
+
+COPY ./src/Microsoft.Health.Fhir.Shared.Web/Microsoft.Health.Fhir.Shared.Web.projitems \
+     ./src/Microsoft.Health.Fhir.Shared.Web/Microsoft.Health.Fhir.Shared.Web.projitems
+
+COPY ./src/Microsoft.Health.Fhir.Shared.Core/Microsoft.Health.Fhir.Shared.Core.projitems \
+     ./src/Microsoft.Health.Fhir.Shared.Core/Microsoft.Health.Fhir.Shared.Core.projitems
+
+COPY ./src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Microsoft.Health.Fhir.Shared.Core.UnitTests.projitems \
+     ./src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Microsoft.Health.Fhir.Shared.Core.UnitTests.projitems
```
LTA-Thinking, line 53: Why are we bringing over the shared Core unit tests at this point?

johnstairs: Absolutely no good reason. Thanks

LTA-Thinking (approved):

Issue #1263, opened 2020-09-11
Specifying a _count greater than Max should return OperationOutcome inside search Bundle instead of a BadRequest


brendankowitz:
**Describe the bug**
Currently when specifying a _count greating than the max will result in a 400 and return an OperationOutcome. We should continue the search and put this OperationOutcome message and into the Bundle that is returned to specify that less results have been returned than requested.

**To Reproduce**
1. Specify ?_count=1000

**Expected behavior**
Search Bundle is returned that includes the OperationOutcome message

**Actual behavior**
Status code 400 with OperationOutcome

CaitlinV39:
AB#76267

PR #1315, opened 2020-09-30
Changes behavior so exceeding _count limit no longer throws a bad request

brendankowitz:
## Description
Changes behavior so exceeding _count limit no longer throws a bad request
Instead the max configured items are returned and a warning is added to the result

## Related issues
Addresses #1263, [AB#76267](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/76267)

## Testing
Adds unit and e2e tests

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Context/FhirRequestContext.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Context/FhirRequestContext.cs
@@ -8,6 +8,7 @@ using System.Collections.Generic;
 using System.Security.Claims;
 using EnsureThat;
 using Microsoft.Extensions.Primitives;
+using Microsoft.Health.Fhir.Core.Models;
 
 namespace Microsoft.Health.Fhir.Core.Features.Context
 {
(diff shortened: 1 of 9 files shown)


LTA-Thinking:
How should this interact with the total and sort parameters? The total count will now not equal the number of entries returned as it wouldn't account for the operation outcomes, and it looks like the operation outcomes will always be the first things returned regardless of sort. I feel like if we are going to add these then we should have a parameter or header flag for the level of message returned, incase a customer doesn't want to have to filter them out.

brendankowitz:
> How should this interact with the total and sort parameters? The total count will now not equal the number of entries returned as it wouldn't account for the operation outcomes, and it looks like the operation outcomes will always be the first things returned regardless of sort. I feel like if we are going to add these then we should have a parameter or header flag for the level of message returned, incase a customer doesn't want to have to filter them out.

It already doesn't since Bundles can include "Matched", "Included", and "Outcome" items. I've update that field for OperationOutcomes in this PR (https://github.com/microsoft/fhir-server/pull/1315/files#diff-959b6cc9a60449c75f30a3e2075b220eR94)

namadabu (approved):

LTA-Thinking (approved):

PR #1321, opened 2020-10-02
Updates the deserializer for ResourceWrapper to include a JsonString of the searchindices

feordin:
## Description
Adds a new property to ResourceWrapper to include a new property which contains the serialzed version of the search indices
Updates ReindexUtilities to utilize these.

## Related issues
Addresses [issue [AB#75331](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/75331)].

## Testing
Unit tests updated, and manual testing.

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Persistence/ResourceWrapper.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Persistence/ResourceWrapper.cs
@@ -119,5 +119,16 @@ namespace Microsoft.Health.Fhir.Core.Features.Persistence
         {
             SearchIndices = searchIndices;
         }
+
+        public virtual bool SearchIndicesEqual(IReadOnlyCollection<SearchIndexEntry> otherIndices)
+        {
+            var otherHashSet = otherIndices != null ?
+                    new HashSet<SearchIndexEntry>(otherIndices) : new HashSet<SearchIndexEntry>();
+
+            var thisHashSet = SearchIndices != null ?
+                    new HashSet<SearchIndexEntry>(SearchIndices) : new HashSet<SearchIndexEntry>();
+
+            return thisHashSet.SetEquals(otherHashSet);
+        }
     }
 }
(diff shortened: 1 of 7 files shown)


src/Microsoft.Health.Fhir.Core/Features/Persistence/ResourceWrapper.cs
```diff
@@ -119,5 +119,16 @@ public void UpdateSearchIndices(IReadOnlyCollection<SearchIndexEntry> searchIndi
         {
             SearchIndices = searchIndices;
         }
+
+        public virtual bool SearchIndicesEqual(IReadOnlyCollection<SearchIndexEntry> otherIndices)
```
namadabu, line 123: If one of the lists has two search index entries for "name" and the other has only one entry for "name", we will still end up returning true here. Is that the expected behavior?

src/Microsoft.Health.Fhir.Core/Features/Search/SearchIndexEntry.cs
```diff
@@ -30,11 +30,22 @@ public SearchIndexEntry(SearchParameterInfo searchParameter, ISearchValue value)
...
         /// <summary>
         /// Gets the search parameter
         /// </summary>
         public SearchParameterInfo SearchParameter { get; }
 
+        /// <summary>
+        /// Used to capture the Json representation of the search index
+        /// usually when returned from Cosmos DB
+        /// </summary>
+        public string JsonString { get; set; }
```
namadabu, line 47: I don't think we should have a data layer specific solution in the `Core` code.

src/Microsoft.Health.Fhir.Core/Features/Persistence/ResourceWrapper.cs
namadabu, line 123: nit: Do we need this as a part of ResourceWrapper class? We could just have it as a separate helper method in ReindexUtilities that does the comparison between two searchindices (since that is the only use case for now). That way we can avoid having this virtual method here.

feordin:
After discussing with @namadabu, we realize that there will be little benefit to doing the work of serializing the search Indices in order to do a comparison with the existing values.  That serialization will take work and may not save any effort on the persistence layer as we submit the entire document back to Cosmos DB in order to update the searchParameterHash value.  I will close the PR without merging it, and as we work on performance testing and continue to iterate on Reindexing we can revisit this if necessary.

PR #1322, opened 2020-10-03
Minor updates/suggestions to the Helm chart

chgl:
## Description
A few minor updates to the Helm chart.

- split container image path into registry + repository. This makes it easier to run using a private registry.
- .Values.securityContext was originally also applied as the sql container's pod security context. This is confusing and may cause the sql server to fail to start due to permission errors if the user id differs from the mssql users uid `10001`.
- removed the serviceAccount as it doesn't look like the server interacts with the Kubernetes API

## Testing

Installed the chart in a local kind cluster.


johnstairs (approved):

chgl:
sorry, forgot to add one last check to support Kubernetes versions < 1.18/networking.k8s.io/v1beta1.

chgl:
New attempt. I've replaced Capabilities.APIVersions.Has with an explicit semver comparison. I am fairly certain this used to work before but I've now double checked with k8s v1.16, 1.18 and 1.19.

PR #1323, opened 2020-10-05
Fixes case where resource content is not included in bundle output

brendankowitz:
## Description
Fixes case where resource content is not included in bundle output, this can happen when an issue is included then the "RawResource" bundle entries do not get serialized

## Related issues
Refs  #1263, [AB#76267](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/76267)

## Testing
Adds regression E2E test


YazanMSFT (approved):
:shipit:

PR #1335, opened 2020-10-08
Adds an active end attribute in the Cosmosdb

rbans96:
## Description
Adds an activeEndDateTime attribute in the Cosmosdb. It will be null for created resources. And when the resources is converted into history record, then this attribute is set to the lastModified date of the updated resource.

## Related issues
Addresses [#75664](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Resolute/Stories/?workitem=75664)

## Testing
Tested manually in the Cosmosdb emulator
image

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,5 @@
+# These owners will be the default owners for everything in
+# the repo. Unless a later match takes precedence,
+# @global-owner1 and @global-owner2 will be requested for
+# review when someone opens a pull request.
+*       @microsoft/fhir-server-team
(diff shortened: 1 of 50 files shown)


rbans96:
image
I spent sometime to identify why activePeriodEndDateTime is getting appended at the bottom just before _ts for the history record. And I think, since activePeriodEndDateTime is null for the non-history(version=1) resource and when this resource is converted into history record, few fields(e.g. isHistory) are modified which are already existing on the resource but since activePeriodEndDateTime does not exists in the document, it is created and appended at the bottom. Ideally this should not be a problem but for the readability of document, it could matter.

LTA-Thinking (approved):

namadabu (approved):

PR #1337, opened 2020-10-08
Add integration tests for Reindex job task

feordin:
## Description
Adds integration tests which create a new search param, reindex resources using it, and then search using the new parameter

## Related issues
Addresses [issue [AB#73396](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/73396)].

## Testing
Created and ran the new integration tests


namadabu:
@feordin Seems like this PR contains other edits apart from the integration tests.

feordin:
> @feordin Seems like this PR contains other edits apart from the integration tests.

Yes, it needed a rebase, should be fixed now.

src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterDefinitionManager.cs
```diff
@@ -35,6 +35,8 @@ public SearchParameterDefinitionManager(IModelInfoProvider modelInfoProvider)
 
         internal IDictionary<Uri, SearchParameterInfo> UrlLookup { get; set; }
 
+        internal IDictionary<string, IDictionary<string, SearchParameterInfo>> TypeLookup { get => _typeLookup; }
```
namadabu, line 38: Should we be returning a copy of the dictionary here so that we can prevent the caller from accidentally updating `_typeLookup`?

feordin: I want the test cases to be able to update _typeLookup so they can insert new search params.  It will be somewhat limited as it is internal, but yes it does make it possible for other classes to make changes we may not want.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Reindex/ReindexUtilities.cs
```diff
@@ -71,19 +71,12 @@ public async Task ProcessSearchResultsAsync(SearchResult results, IReadOnlyDicti
...
-                else
-                {
-                    entry.Resource.UpdateSearchIndices(newIndices);
-                    updateSearchIndices.Add(entry.Resource);
-                }
+                // TODO: If it reasonable to do so, we can compare
+                // old and new search indices to avoid unnecessarily updating search indices
+                // when not changes have been made.
+                entry.Resource.UpdateSearchIndices(newIndices);
+                updateSearchIndices.Add(entry.Resource);
```
namadabu, line 79: We can remove the call to `UpdateSearchParameterHashBatchAsync` below and also remove `updateHashValueOnly` above

test/Microsoft.Health.Fhir.Shared.Tests.Integration/TestScoped.cs
```diff
@@ -0,0 +1,19 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using Microsoft.Health.Extensions.DependencyInjection;
+
+namespace Microsoft.Health.Fhir.Shared.Tests.Integration
+{
+    public class TestScoped<T> : IScoped<T>
+    {
+        public T Value { get; set; }
```
namadabu, line 12: There is an existing helper method `CreateMockedScopeExtensions.CreateMockScope` that we use for other tests. Would it be possible to use that for the `ReindexJobTests` instead of adding a new class?

namadabu (approved):

Issue #1319, opened 2020-10-01
Incorrect code for _revinclude bad request


CaitlinV39:
**Describe the bug**
Incorrect code for _revinclude bad request

**To Reproduce**
Hit request /Observation?_revinclude=MedicationRequest:observation
OperationOutcome is returned for status - 400 Bad Request

**Expected behavior**
* "code": "invalid".
* Refer to the [spec](https://www.hl7.org/fhir/search.html#errors) for cases where the http status code should change to 403

**Actual behavior**
* "code": "forbidden".
* Status code is always 400

eladiw:
@CaitlinV39, this behavior is not revinclude specific. it is the case for every bad request.
e.g. 
/MedicationRequest?_include=MedicationRequest:Patient

brendankowitz:
Story is reopened to address status code corrections.

PR #1351, opened 2020-10-14
Fix issue 1319 - return correct code on bad / invalid requests

eladiw:
## Description
Instead of returning Forbidden. returning the correct code

## Related issues
Addresses [issue #1319].


johnstairs:
@eladiw, the problem is that the [spec](https://www.hl7.org/fhir/search.html#errors) says we should return 403 in these cases. I agree it shouldn't be this way though.
>A HTTP status code of 403 signifies that the server refused to perform the search, while other 4xx and 5xx codes signify that some sort of error has occurred

@brendankowitz, what do you think?

src/Microsoft.Health.Fhir.Core/Exceptions/CustomerManagedKeyException.cs
```diff
@@ -14,7 +14,7 @@ public CustomerManagedKeyException(string message)
         {
             Issues.Add(new OperationOutcomeIssue(
                     OperationOutcomeConstants.IssueSeverity.Error,
-                    OperationOutcomeConstants.IssueType.Forbidden,
+                    OperationOutcomeConstants.IssueType.Invalid,
```
brendankowitz, line 17: https://www.hl7.org/fhir/codesystem-issue-type.html#issue-type-no-store, thoughts?

brendankowitz:
> @eladiw, the problem is that the [spec](https://www.hl7.org/fhir/search.html#errors) says we should return 403 in these cases. I agree it shouldn't be this way though.
> 
> > A HTTP status code of 403 signifies that the server refused to perform the search, while other 4xx and 5xx codes signify that some sort of error has occurred
> 
> @brendankowitz, what do you think?

If we also don't return the correct status code, we should also fix this, but it doesn't have to be in this PR since its just changing the issue types.

brendankowitz:
I'll merge this as a partial fix for #1319. I've updating but bug with additional information about changing the status code.

brendankowitz (approved):

PR #1361, opened 2020-10-15
Remove unnecessary index on ReferenceSearchParam

johnstairs:
Small schema change to remove an unnecessary index of ReferenceSearchParam. This index also didn't have page compression, so it took up about 13% of the database size.

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/5.sql
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/5.sql
@@ -1,0 +1,1872 @@
+-- NOTE: This script DROPS AND RECREATES all database objects.
+-- Style guide: please see: https://github.com/ktaranov/sqlserver-kit/blob/master/SQL%20Server%20Name%20Convention%20and%20T-SQL%20Programming%20Style.md
+
+
+
+/*************************************************************
+    Schema Version
+**************************************************************/
+
+INSERT INTO dbo.SchemaVersion
+VALUES
+    (4, 'started')
+
+GO
+
+/*************************************************************
+    Model tables
+**************************************************************/
+
+CREATE TABLE dbo.SearchParam
+(
+    SearchParamId smallint IDENTITY(1,1) NOT NULL,
+    Uri varchar(128) COLLATE Latin1_General_100_CS_AS NOT NULL
+)
+
+CREATE UNIQUE CLUSTERED INDEX IXC_SearchParam ON dbo.SearchParam
+(
+    Uri
+)
+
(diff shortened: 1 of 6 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/5.sql
```diff
@@ -0,0 +1,1872 @@
+-- NOTE: This script DROPS AND RECREATES all database objects.
```
brendankowitz, line 1: Is this note still true? :P

brendankowitz (approved):

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/5.sql
johnstairs: Ah, it is not at all true anymore.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/5.sql
```diff
@@ -0,0 +1,1869 @@
+-- Style guide: please see: https://github.com/ktaranov/sqlserver-kit/blob/master/SQL%20Server%20Name%20Convention%20and%20T-SQL%20Programming%20Style.md
+
+/*************************************************************
+    Schema Version
+**************************************************************/
+
+INSERT INTO dbo.SchemaVersion
+VALUES
+    (4, 'started')
```
YazanMSFT, line 9: Should this be 5?

YazanMSFT: 5.diff.sql would need that too, maybe it's intentional? Just making sure

johnstairs: Good catch! Thank yo. It seems to be added by the tool when applying the diff. (I don't think should be needed for the full sql file, @rbans96, what do you think)?

rbans96: Good point! Since dropping of database objects are removed from the full sql file from version 4 and onwards, we can insert the version in the same way as we do for diff files by the tool or SchemaUpgradeRunner.cs(for automatic upgrades). Probably we can plan to do this change for the next version scripts(>=6).

johnstairs: Thanks. Opened https://github.com/microsoft/healthcare-shared-components/issues/113

Issue #1342, opened 2020-10-09
Error on _include query


johnstairs:
**Describe the bug**
Error when handling simple include query.

**Data provider?**
SQL Server

**To Reproduce**
1. `GET https://localhost:44348/MedicationRequest?_include=MedicationRequest:patient`

**Expected behavior**
A result set

**Actual behavior**
A 400 with body
``` json
{
  "resourceType": "OperationOutcome",
  "id": "581b4b6c-9eec-4700-a725-40ff58f2d4a1",
  "issue": [
    {
      "severity": "error",
      "code": "invalid",
      "diagnostics": "Expected an expression that evaluates to true. (Parameter 'expressions')"
    }
  ]
}
```
Note also that this should be a 500 and not a 400.

CaitlinV39:
@CaitlinV39 create a user story

CaitlinV39:
AB#76269

CaitlinV39:
@johnstairs I think you merged a fix for this. Is this still open?

johnstairs:
Closing.

Issue #1259, opened 2020-09-11
Malformed continuation token results in 500


johnstairs:
When using SQL Server as the backend, call

``` 
GET https://localhost:44348/Patient?_sort=birthdate&ct=bla
```

result:
```
System.Text.Json.JsonException: 'b' is an invalid start of a value. Path: $ | LineNumber: 0 | BytePositionInLine: 0.
 ---> System.Text.Json.JsonReaderException: 'b' is an invalid start of a value. LineNumber: 0 | BytePositionInLine: 0.
   at System.Text.Json.ThrowHelper.ThrowJsonReaderException(Utf8JsonReader& json, ExceptionResource resource, Byte nextByte, ReadOnlySpan`1 bytes)
   at System.Text.Json.Utf8JsonReader.ConsumeValue(Byte marker)
   at System.Text.Json.Utf8JsonReader.ReadFirstToken(Byte first)
   at System.Text.Json.Utf8JsonReader.ReadSingleSegment()
   at System.Text.Json.Utf8JsonReader.Read()
   at System.Text.Json.JsonSerializer.ReadCore(JsonSerializerOptions options, Utf8JsonReader& reader, ReadStack& readStack)
   --- End of inner exception stack trace ---
   at System.Text.Json.ThrowHelper.ReThrowWithPath(ReadStack& readStack, JsonReaderException ex)
   at System.Text.Json.JsonSerializer.ReadCore(JsonSerializerOptions options, Utf8JsonReader& reader, ReadStack& readStack)
   at System.Text.Json.JsonSerializer.ReadCore(Type returnType, JsonSerializerOptions options, Utf8JsonReader& reader)
   at System.Text.Json.JsonSerializer.Deserialize(String json, Type returnType, JsonSerializerOptions options)
   at System.Text.Json.JsonSerializer.Deserialize[TValue](String json, JsonSerializerOptions options)
   at Microsoft.Health.Fhir.Core.Features.Search.ContinuationToken.FromString(String json) in D:\repos\fhir-server\src\Microsoft.Health.Fhir.Core\Features\Search\ContinuationToken.cs:line 70
   at Microsoft.Health.Fhir.SqlServer.Features.Search.SqlServerSearchService.SearchImpl(SearchOptions searchOptions, Boolean historySearch, CancellationToken cancellationToken) in D:\repos\fhir-server\src\Microsoft.Health.Fhir.SqlServer\Features\Search\SqlServerSearchService.cs:line 144
[...]

PR #1367, opened 2020-10-16
Improve error handling for bad continuation tokens and search values.

johnstairs:
## Description
Improves handling for malformed SQL and Cosmos DB continuation tokens, and improves error handling when parsing search values. Also avoids being over broad with our error handling. For instance, `ArgumentException`s, like those raised in `EnsureArg` calls, would be translated to 400 response codes instead of 500.

## Related issues
Addresses #1259, #674.
Addresses part of #1342.

## Testing
Added E2E test cases.

### DIFF
--- a/src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
+++ b/src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
@@ -3,6 +3,7 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Net;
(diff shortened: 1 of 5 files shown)


src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
```diff
@@ -278,7 +279,15 @@ internal async Task<FeedResponse<T>> ExecuteDocumentQueryAsync<T>(QueryDefinitio
 
             var documentQuery = _cosmosQueryFactory.Create<T>(_containerScope.Value, context);
 
-            return await documentQuery.ExecuteNextAsync(cancellationToken);
+            try
+            {
+                return await documentQuery.ExecuteNextAsync(cancellationToken);
+            }
+            catch (CosmosException e) when (e.StatusCode == HttpStatusCode.BadRequest && continuationToken != null && e.ResponseBody.StartsWith("Malformed continuation token", StringComparison.OrdinalIgnoreCase))
```
brendankowitz, line 286: Did this pass through https://github.com/microsoft/fhir-server/blob/master/src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosResponseProcessor.cs#L45? Should this be in the handler?

johnstairs: Makes sense. I was thinking that we should only return a 400 if this was a CT that came from a client, not one that we used internally. But that is probably unlikely.

johnstairs: @brandonpollett, it does not pass though the handler:
image
I don't know if this is a bug in the SDK or our code.

brendankowitz: Thanks, I'll make a bug to look into this

brendankowitz: #1390

brendankowitz (approved):

Issue #1353, opened 2020-10-14
Exception when loading data into server


johnstairs:
I was loading data into the FHIR server (16 concurrent requests) and got this exception:

```
System.InvalidOperationException: Operations that change non-concurrent collections must have exclusive access. A concurrent update was performed on this collection and corrupted its state. The collection's state is no longer correct.
   at System.Collections.Generic.Dictionary`2.FindEntry(TKey key)
   at System.Collections.Generic.Dictionary`2.ContainsKey(TKey key)
   at Microsoft.Health.Fhir.Core.Features.Definition.SearchParameterDefinitionManager.UpdateSearchParameterHashMap(Dictionary`2 updatedSearchParamHashMap) in D:\repos\fhir-server\src\Microsoft.Health.Fhir.Core\Features\Definition\SearchParameterDefinitionManager.cs:line 124
   at Microsoft.Health.Fhir.Core.Features.Definition.SearchParametersHashUpdatedHandler.Handle(SearchParametersHashUpdated notification, CancellationToken cancellationToken) in D:\repos\fhir-server\src\Microsoft.Health.Fhir.Core\Features\Definition\SearchParametersHashUpdatedHandler.cs:line 26
   at MediatR.Internal.NotificationHandlerWrapperImpl`1.<>c__DisplayClass0_0.<Handle>b__1(INotification theNotification, CancellationToken theToken)
   at MediatR.Mediator.PublishCore(IEnumerable`1 allHandlers, INotification notification, CancellationToken cancellationToken)
   at Microsoft.Health.Fhir.Core.Features.Search.Registry.SearchParameterStatusManager.EnsureInitialized() in D:\repos\fhir-server\src\Microsoft.Health.Fhir.Core\Features\Search\Registry\SearchParameterStatusManager.cs:line 145
   at Microsoft.Health.Api.Modules.InitializationModule.<>c__DisplayClass2_0.<<Configure>b__0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at Microsoft.Health.Fhir.Api.Features.Context.FhirRequestContextAfterAuthenticationMiddleware.Invoke(HttpContext context, IFhirRequestContextAccessor fhirRequestContextAccessor) in D:\repos\fhir-server\src\Microsoft.Health.Fhir.Api\Features\Context\FhirRequestContextAfterAuthenticationMiddleware.cs:line 36
[...]

PR #1368, opened 2020-10-18
Handle concurrent updates to search param hash map

namadabu:
## Description
Updated code to use a ConcurrentDictionary to store `_resourceTypeSearchParameterHashMap` in `SearchParameterDefinitionManager`. This will make sure that the calls to `UpdateSearchParameterHashMap` is thread-safe.

## Related issues
Addresses [issue #1353].

## Testing
Existing tests.


src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterDefinitionManager.cs
```diff
@@ -118,19 +119,14 @@ public ValueSets.SearchParamType GetSearchParameterType(SearchParameterInfo sear
...
             {
-                if (_resourceTypeSearchParameterHashMap.ContainsKey(kvp.Key))
-                {
-                    _resourceTypeSearchParameterHashMap[kvp.Key] = kvp.Value;
-                }
-                else
-                {
-                    _resourceTypeSearchParameterHashMap.Add(kvp.Key, kvp.Value);
-                }
+                _resourceTypeSearchParameterHashMap.AddOrUpdate(
```
brendankowitz, line 126: Just for my understanding, why/where is this being called concurrently from?

feordin: Once we implement Custom Search, it would be possible for multiple requests to update/add search parameters to come in at the same time.  Right now, this is only called during start up when we build the search parameter definitions.  @namadabu please correct me if I am wrong.

namadabu: @brendankowitz +1 to what @feordin mentioned,

brendankowitz: Agree this change is needed. My question was just why its potentially being called multiple times during startup

brendankowitz (approved):

PR #1376, opened 2020-10-20
Update Reindex job progress when a query is successfully processed

feordin:
## Description
Added to the progress property when  a query is successfully processed

## Related issues
Addresses [issue AB76382#].

## Testing
Updated unit test

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Operations/Reindex/ReindexJobTaskTests.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Operations/Reindex/ReindexJobTaskTests.cs
@@ -214,6 +214,7 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Operations.Reindex
 
             Assert.Equal(OperationStatus.Completed, job.Status);
             Assert.Equal(10, job.Count);
+            Assert.Equal(4, job.Progress);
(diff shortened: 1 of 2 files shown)


namadabu (approved):

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Operations/Reindex/ReindexJobTaskTests.cs
```diff
@@ -214,6 +214,7 @@ public async Task GivenRunningJob_WhenExecuted_ThenQueuedQueryCompleted()
 
             Assert.Equal(OperationStatus.Completed, job.Status);
             Assert.Equal(10, job.Count);
+            Assert.Equal(4, job.Progress);
```
Ivanidzo4ka, line 217: >Assert.Equal(4, job.Progress); [](start = 12, length = 30)

Just to note: `PercentComplete` here returns 40 (it's defined as Progress/Count *100);
For person new to project it seems a bit odd to have operation status = Completed, but `PercentComplete` is not 100.

feordin: It is true, it is a result of the faked data for the test, where fakes are not entirely consistent.

Issue #1151, opened 2020-07-14
_type parameter does not validate resource types 


LTA-Thinking:
**Describe the bug**
When searching with the _type parameter the resource types are not validated. If a resource type is not recognized, then it is ignored. This causes a typo to return the same results as there being no matches. 

**To Reproduce**
1. Conduct a global search with _type=notreal

**Expected behavior**
An operation outcome is returned stating that the resource type is invalid.

**Actual behavior**
An empty bundle is returned.

---

**Edit:** See [below](#issuecomment-726367437), if possible, we should not throw exceptions where previously the response was allowed. This looks to be one of those cases where the error can be reported with the output

CaitlinV39:
@LTA-Thinking was this resolved with #1139 or is there more work to do?

LTA-Thinking:
@CaitlinV39 The issue was discovered while working on #1139. It has not been fixed yet.

CaitlinV39:
AB#76266

Ivanidzo4ka:
There is certain uncertainty how to handle `_type` parameter for non global case.
Like `http://server/Observation?_type=Patient` doesn't make much sense to me.
In same time `http://server/Observation/[id]/*?_type=Patient` does.

brendankowitz:
@CaitlinV39 I'm reviewing this requirement again, this could be potentially breaking for a customer, it will throw a BadRequest where the invalid type was previously ignored. Suggestion is we add the invalid type message to the bundle like here: https://github.com/microsoft/fhir-server/blob/master/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs#L150

CaitlinV39:
@brendankowitz that makes sense to me and I can document that we updated the bundle. I don't want to throw the BadRequest.

Ivanidzo4ka:
@brendankowitz @CaitlinV39 
So for the case of `?_type=Patient,foo` I think we can all agree it should do search for `Patient` only and add message what `foo` is unknow.
But what should be the case for `?type=foo`? We can ignore type and in this [case](https://www.hl7.org/fhir/search.html#_type):
`If no type is specified, the only search parameters that be can be used in global search like this are the base parameters that apply to all resources.`
Or we can don't do search at all, since we know it's pointless.

Second implementation would align with what we do right now.

Which brings my second question. In my PR @brendankowitz  mention what it would be nice to clean up  self url and next url links from invalid parameters.

[...]

PR #1385, opened 2020-10-22
Add validation for `_type` parameter.

Ivanidzo4ka:
## Description
Validate passed `_type` parameter to be known resource.

## Related issues
Addresses [#1151].

## Testing
Manually and via unit tests.


src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
```diff
@@ -102,6 +102,36 @@ public SearchOptions Create(string compartmentType, string compartmentId, string
...
+                            new OperationOutcomeIssue(
+                                OperationOutcomeConstants.IssueSeverity.Information,
+                                OperationOutcomeConstants.IssueType.Informational,
+                                string.Format(Core.Resources.InvalidTypeParameter, string.Join(",", badTypes))));
+                        if (badTypes.Count != types.Count)
+                        {
+                            searchParams.Add(KnownQueryParameterNames.Type, string.Join(',', types.Except(badTypes)));
+                            foreach (var badType in badTypes)
+                            {
+                                unsupportedSearchParameters.Add(new Tuple<string, string>(KnownQueryParameterNames.Type, badType));
```
brendankowitz, line 122: So just confirming that if I passed in `?_type=Observation,foo1,foo2` then the resulting Bundle self link would have `?_type=Observation`?

brendankowitz: It is based on routing + the behavior of dropping unsupported options, have a look at UrlResolver to see if its consistent 
https://github.com/microsoft/fhir-server/blob/f96a3bc70da1422fb8de8e0129cba7e7e3318f02/src/Microsoft.Health.Fhir.Api/Features/Routing/UrlResolver.cs#L144-L236

Ivanidzo4ka: I guess I broke contract of what unsupported parameters should mean.
One in `UrlResolver` looks for exact matching, while I put pieces instead of whole.

Example: `_type=Observation,foo,foo1` would have two unsupported parameters `_type=foo` and `_type==foo1`.
But in case if I pass `_type=foo` it would be removed from link.

What is preferred behavior here?


---
In reply to: [511157028](https://github.com/microsoft/fhir-server/pull/1385#discussion_r511157028) [](ancestors = 511157028)

brendankowitz: I think the ideal case is that the unsupported options are removed from the link

Ivanidzo4ka: I've added logic to remove unsupported types and add tests for that.

---
In reply to: [512828294](https://github.com/microsoft/fhir-server/pull/1385#discussion_r512828294) [](ancestors = 512828294)

Ivanidzo4ka: Change it to contain only valid types in self link

---
In reply to: [511147213](https://github.com/microsoft/fhir-server/pull/1385#discussion_r511147213) [](ancestors = 511147213,511142603)

src/Microsoft.Health.Fhir.Core/Features/Search/StringExtensions.cs
```diff
@@ -60,7 +60,7 @@ internal static class StringExtensions
         /// <param name="s">The string to be parsed.</param>
         /// <returns>An <see cref="IList{T}"/> that contains strings split by using token separator.</returns>
         /// <remarks>Since the token separator character has special rule defined, if the character appears in the actual value,
-        /// it needs to be escaped. More detail can be found here: http://hl7.org/fhir/STU3/search.html#escaping</remarks>
+        /// it needs to be escaped. More detail can be found here: http://hl7.org/fhir/STU3/search.html#escaping </remarks>
```
brendankowitz, line 63: Prefer links to latest spec e.g. http://hl7.org/fhir/search.html#escaping

src/Microsoft.Health.Fhir.Core/Resources.resx
```diff
@@ -450,4 +450,7 @@
     <value>Reindex job id {0} is in state {1} and cannot be cancelled.</value>
     <comment>{0} is the id of the reindex which is a a GUID.  {1} is the state, an enumerated set of state values.</comment>
   </data>
-</root>
\ No newline at end of file
+  <data name="InvalidTypeParameter" xml:space="preserve">
+    <value>The '_type' parameter contains unknown resource(s): {0}</value>
```
LTA-Thinking, line 454: We need a comment for what {0} will be replaced with.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
```diff
@@ -106,6 +106,32 @@ public SearchOptions Create(string compartmentType, string compartmentId, string
...
+                                OperationOutcomeConstants.IssueSeverity.Warning,
+                                OperationOutcomeConstants.IssueType.NotSupported,
+                                string.Format(Core.Resources.InvalidTypeParameter, badTypes.OrderBy(x => x).Select(type => $"'{type}'").JoinByOrSeparator())));
+                        if (badTypes.Count != types.Count)
+                        {
+                            searchParams.Add(KnownQueryParameterNames.Type, types.Except(badTypes).JoinByOrSeparator());
+                        }
+                        else
+                        {
+                            searchParams.Add(KnownQueryParameterNames.Type, query.Item2);
```
LTA-Thinking, line 127: Could we add a comment for why we are including the bad types if there are only bad types. I know we discussed this but it looks like an area that we will forget what our reasoning was in a few months.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/BasicSearchTests.cs
```diff
@@ -188,6 +188,75 @@ public async Task GivenVariousTypesOfResources_WhenSearchingAcrossAllResourceTyp
...
+
+            bundle = await Client.SearchAsync("?_type=Patient1,Patient2");
+            Assert.Contains("_type=Patient1,Patient2", bundle.Link[0].Url);
+            outcome = GetOutcome(bundle);
+            ValidateOperationOutcome(expectedDiagnosticsMultipleWrongTypes, expectedIssueSeverities, expectedCodeTypes, outcome);
+
+            bundle = await Client.SearchPostAsync(null, default, ("_type", "Patient1"));
+            Assert.Contains("_type=Patient1", bundle.Link[0].Url);
+            outcome = GetOutcome(bundle);
+            ValidateOperationOutcome(expectedDiagnosticsOneWrongType, expectedIssueSeverities, expectedCodeTypes, outcome);
```
LTA-Thinking, line 213: Nit: Linebreak between test sections

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/BasicSearchTests.cs
```diff
@@ -188,6 +188,75 @@ public async Task GivenVariousTypesOfResources_WhenSearchingAcrossAllResourceTyp
...
+            ValidateBundle(bundle, patients.AsEnumerable<Resource>().Append(outcome).ToArray());
+            ValidateOperationOutcome(expectedDiagnostics, expectedIssueSeverities, expectedCodeTypes, outcome);
+
+            bundle = await Client.SearchPostAsync(null, default, ("_type", "Patient1,Patient"));
+            Assert.Contains("_type=Patient1,Patient", bundle.Link[0].Url);
+            outcome = GetOutcome(bundle);
+            ValidateBundle(bundle, patients.AsEnumerable<Resource>().Append(outcome).ToArray());
+            ValidateOperationOutcome(expectedDiagnostics, expectedIssueSeverities, expectedCodeTypes, outcome);
+
+            static OperationOutcome GetOutcome(Bundle bundle)
```
LTA-Thinking, line 250: Since this local method is in multiple test cases could it be moved to be a class method so there isn't duplicate code.

LTA-Thinking (approved):

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/BundleFactory.cs
```diff
@@ -83,6 +83,7 @@ private ResourceElement CreateBundle(SearchResult result, Bundle.BundleType type
             {
                 var operationOutcome = new OperationOutcome
                 {
+                    Id = _fhirRequestContextAccessor.FhirRequestContext.CorrelationId,
```
brendankowitz, line 86: I don't think this makes sense when the bundle will have the same id set below (https://github.com/microsoft/fhir-server/pull/1385/files#diff-038e862a9e90e7e3da159fb2485bc1017c36fbc313bca2f4bd8d2c3052ad2f88R116)

Ivanidzo4ka: [https://www.hl7.org/fhir/resource-definitions.html#Resource.id](https://www.hl7.org/fhir/resource-definitions.html#Resource.id)

```
Comments 
The only time that a resource does not have an id is when it is being submitted to the server using a create operation.
```

My interpretation of standard - we need to return id for every resource. I know it's same as id in bundle, (I'm actually not sure is it ok for two resources have same id or not, couldn't find clarification in standard), but I think it's better than return nothing.

---
In reply to: [525361013](https://github.com/microsoft/fhir-server/pull/1385#discussion_r525361013) [](ancestors = 525361013)

brendankowitz: I think that part is ok, for example `Patient/example` and `Observation/example` are both valid with the same "id". I guess under that pretense it makes sense to leave it in

brendankowitz (approved):

Issue #1382, opened 2020-10-21
Concurrent conditional creates results in duplicate data


CaitlinV39:
**Describe the bug**
When doing two conditional creates simultaneously (or near simultaneously) for the same resource, both are created, causing duplicate data to be stored in the system

**FHIR Version?**
Stu3/R4

**Data provider?**
CosmosDB/SQL Server

**To Reproduce**
Following are the steps the customer is following
1 Delete all existing  patient data from their instance.
2. Make concurrent ‘Create Patient’ requests with ‘If-None-Exist:’ with a patient Identifier as  a header parameter.

**Expected behavior**
Receive the 412 response for the 2nd create patient request that hits the Azure API for FHIR instance. 

**Actual behavior**
2nd patient is created as a duplicate

CaitlinV39:
AB#76902 and AB#76903 to address in SQL and Cosmos

anyname2:
I have just ran into the same issue. I can see there was a PR to fix it: #1386.
@brendankowitz,  what was the reason it has been closed and not merged?

CraigP68:
AB#87262

moloneyd-draeger:
We have noticed the same issue with conditional creates in our environment. A quick python script to recreate the issue is attached - please note this is "flaky" as the server sometimes does only create one resource! 

Is there any way of configuring the system or sending a request so that the server will apply a "stronger" uniqueness condition ?

We have also noticed that the issue occurs much more frequently if the server is under more load, and especially when the CosmosDB is reaching the throttled/configured RU limit. Does anyone have any recommended mitigations for this issue?

FHIR Version?
R4

Data provider?
CosmosDB

Recreate script
```python
# requirements
# pip install requests fhir.resources==6.1.0

# Recreate conditional create issue with Azure API for FHIR.
# tested with
#   python 3.9
#   Azure API for FHIR: fhir-R4, CosmosDB (1000 RUs)

import time
import uuid

from datetime import datetime, timezone
from concurrent.futures import ThreadPoolExecutor
from urllib.parse import urlencode

import requests

from fhir.resources.bundle import Bundle
from fhir.resources.identifier import Identifier
from fhir.resources.patient import Patient

# configure environment and auth
FHIR_SERVER_ADDRESS = ''
AZURE_TENANT_ID = ''
OAUTH_CLIENT_ID = ''
OAUTH_CLIENT_SECRET = ''

# configure request flooding
[...]

PR #1386, opened 2020-10-23
Adds locking behavior to ConditionalCreate pipeline

brendankowitz:
## Description
Exploring solutions for adding concurrency safety around Conditional Creates/Updates.

- CosmosDB uses a lock document with `ResourceType:ConditionalParameters` as the key
- SQL wraps the operations in a `Serializable` transaction

One thread will succeed, other competing requests will result in a 409 conflict when competing with the same Condition.

### Known issues

- CosmosDB - Does not solve for the case were a resource could use a normal Create call which results in a resource that should be picked up by Conditional Create
- CosmosDB - Does not solve for the case where another concurrent Conditional Create uses a different search parameter that would result in the creation of the same resource

### When does this work?

- Conditional Create is highly utilized with the `identifier` parameter when importing or replicating data from another system. Under these circumstances, this solution would reject concurrent conditional creates with a 409.

Refs #1382

### DIFF
--- a/src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/ConditionalCreateLockingBehavior.cs
+++ b/src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/ConditionalCreateLockingBehavior.cs
@@ -1,0 +1,84 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using EnsureThat;
+using MediatR;
+using Microsoft.Health.Core.Extensions;
+using Microsoft.Health.Fhir.Core.Features.Persistence;
+using Microsoft.Health.Fhir.Core.Messages.Create;
+using Microsoft.Health.Fhir.Core.Messages.Upsert;
+
+namespace Microsoft.Health.Fhir.CosmosDb.Features.Storage
+{
+    /// <summary>
+    /// Intercepts ConditionalCreateResourceRequests and wraps them in a CosmosDbDistributedLock
+    /// </summary>
+    public class ConditionalCreateLockingBehavior
+        : IPipelineBehavior<ConditionalCreateResourceRequest, UpsertResourceResponse>,
+            IPipelineBehavior<ConditionalUpsertResourceRequest, UpsertResourceResponse>
+    {
+        private readonly ICosmosDbDistributedLockFactory _lockFactory;
+
+        public ConditionalCreateLockingBehavior(ICosmosDbDistributedLockFactory lockFactory)
+        {
+            EnsureArg.IsNotNull(lockFactory, nameof(lockFactory));
+
+            _lockFactory = lockFactory;
+        }
+
+        public async Task<UpsertResourceResponse> Handle(ConditionalCreateResourceRequest request, CancellationToken cancellationToken, RequestHandlerDelegate<UpsertResourceResponse> next)
+        {
+            EnsureArg.IsNotNull(request, nameof(request));
+            EnsureArg.IsNotNull(next, nameof(next));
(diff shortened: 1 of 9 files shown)


test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Metric/MetricTests.cs
```diff
@@ -85,7 +85,7 @@ public async Task GivenABatch_WhenInvokedAtCosmosDb_MetricNotificationsShouldBeE
             await ExecuteAndValidate(
                 () => _client.PostBundleAsync(Samples.GetDefaultBatch().ToPoco()),
                 (type: typeof(ApiResponseNotification), count: 1, resourceType: Samples.GetDefaultBatch().ToPoco().ResourceType.ToString()),
-                (type: typeof(CosmosStorageRequestMetricsNotification), count: 10, resourceType: Samples.GetDefaultBatch().ToPoco().ResourceType.ToString()));
+                (type: typeof(CosmosStorageRequestMetricsNotification), count: 14, resourceType: Samples.GetDefaultBatch().ToPoco().ResourceType.ToString()));
```
brendankowitz, line 88: Lock document accounts for an extra 2 requests each Conditional Create/Update

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/ConditionalCreateLockingBehavior.cs
```diff
@@ -0,0 +1,84 @@
...
+            EnsureArg.IsNotNull(conditions, nameof(conditions));
+            EnsureArg.IsNotNull(next, nameof(next));
+
+            var hashBuilder = new StringBuilder();
+            conditions.Aggregate(hashBuilder, (builder, tuple) => builder.AppendFormat("{0}={1}&", tuple.Item1, tuple.Item2));
+
+            var conditionalHash = hashBuilder.ToString().ComputeHash();
+            ICosmosDbDistributedLock lockDocument = _lockFactory.Create($"ConditionalCreate:{resourceType}:{conditionalHash}");
+
+            if (await lockDocument.TryAcquireLock() == false)
```
feordin, line 65: Is there any need to place TryAcquireLock inside the try below?  If this returns false, or throws, then we won't dispose the lockDocument.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/ConditionalCreateTransactionBehavior.cs
```diff
@@ -0,0 +1,96 @@
...
+                        SqlTransaction trans = wrapper.SqlConnection.BeginTransaction(IsolationLevel.Serializable);
+
+                        transactionScope.SqlConnection = wrapper.SqlConnection;
+                        transactionScope.SqlTransaction = trans;
+                    }
+                }
+
+                UpsertResourceResponse result = await next();
+
+                transactionScope?.Complete();
```
feordin, line 83: Should these be in a finally?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ConditionalCreateTests.cs
```diff
@@ -104,5 +106,35 @@ public async Task GivenAResource_WhenCreatingConditionallyWithEmptyIfNoneHeader_
 
             Assert.Equal(HttpStatusCode.BadRequest, exception.Response.StatusCode);
         }
+
+        [Fact]
+        [Trait(Traits.Priority, Priority.One)]
+        public async Task GivenConcurrentResources_WhenCreatingConditionallyWithNoIdAndNoExisting_TheServerShouldReturnOneResourceSuccessfully()
```
feordin, line 112: Not sure if it is possible to create a test which replicates the resources locked case and we see the resource conflict result?

brendankowitz: When I run this locally I see 1 created and 19 conflicted. I added `|| x == HttpStatusCode.OK` which I believe is also valid where the initial request finished before others had started.

Given the attempt to run 20 requests at once, I'm not sure how to isolate any further to guarantee Conflicts.

PR #1388, opened 2020-10-23
Fixes R5 test failures

johnstairs:
## Description
All the R5 E2E tests were failing for me when run locally with Unauthenticated. This exception was being thrown when validating the token:

```
System.MissingMethodException: Method not found: 'Void Microsoft.IdentityModel.JsonWebTokens.JwtTokenUtilities.ValidateTokenType(System.String, Microsoft.IdentityModel.Tokens.TokenValidationParameters)'.
   at System.IdentityModel.Tokens.Jwt.JwtSecurityTokenHandler.ValidateTokenPayload(JwtSecurityToken jwtToken, TokenValidationParameters validationParameters)
   at System.IdentityModel.Tokens.Jwt.JwtSecurityTokenHandler.ValidateToken(String token, TokenValidationParameters validationParameters, SecurityToken& validatedToken)
   at Microsoft.AspNetCore.Authentication.JwtBearer.JwtBearerHandler.HandleAuthenticateAsync()
```

The root cause was a that we did had not updated the package versions in a couple of R5 project files. (The versions we have the the R4 and STU3 files were not the same as in the R5 files)


feordin (approved):

brendankowitz (approved):
Thanks @johnstairs

PR #1314, opened 2020-09-30
Supporting _include:iterate and _revinclude:iterate 

limorl:
## Description
The :iterate modifier (:recurse in STU3)  allows to query iteratively for resources.
  For example, returning MedicationRequest ‘mr123’, and the Patient resources it references and those patient's Practitioner:
`/MedicationRequest?_include=MedicationRequest:patient&_include:iterate=Patient:general-practitioner&_id=mr123`

The solution is based mainly on updating SqlQueryGenerator to store a dictionary of result resource type and the cte names associated with them. This way, when iterating over results, the selection is done from the relevant cte (or ctes in case of multiple results sets).

For this to work, the IncludeRewriter should have been updated to sort the search parameters so that an _include/_revinclude parameter will appear after all the expressions yielding the result sets it selects from. 

The feature doesn't fully implement the spec and some capabilities were left out following a discussion with the PG. 

**In scope of this PR:**
- **Supporting _include:iterate expressions, with single/multiple iterations and with wildcards.** 
   For example:
   - `/MedicationRequest?_include=MedicationRequest:patient&_include:iterate=Patient:general-practitioner&_id=mr123`
   - `MedicationRequest?_include:MedicationRequest:*&_include:iterate=Patient:general-practitioner `
   -  `MedicationRequest&_include:MedicationRequest:patient&_include:iterate=Patient:* MedicationRequest?_include:MedicationRequest:*&_include:iterate=Patient:*`

- **Supporting _revinclude:iterate expressions with single/multiple iterations and with wildcard.** 
  -  This is similar to the _include queries. However, for _revinclude we require that if the TargetType can be of multiple types, a resource type must be specified. Other a BadRequest is returned. For example:
  - ` Practitioner?_revinclude=Patient:*&_revinclude:iterate=MedicationRequest:patient`
  -  ` Practitioner?_revinclude:iterate=MedicationRequest:*&_revinclude=Patient:general-practitioner `
  -  `Patient?_revinclude:iterate=MedicationDispense:*&_revinclude=MedicationRequest:*`

**Out of scope of this PR:**
* #1310  
  Recursive iterate expressions are expression where the Source and the Target are of the same type. For example:  
  `Organization?_include:iterate=Organization:partof Patient&_include=Patient:organization&_include:iterate=Organization:partof `

* Supporting _revinclude:iterate expressions where there is more than a single target type and the target type is not specified:   
[...]

Issue #1313, opened 2020-09-30
Support _include:iterate and _revinclude:iterate search queries


limorl:
**User story**
As a user, I want to use :iterate modifier for _include and _revinclude so that I can iteratively include results by executing a single query.

**Acceptance criteria**
1. When :iterate modifier is used, the iterative results are included in the results following the FHIR spec
2. When using :recurse modifier, it returns equal results to when using :iterate, on all versions, and considered an alias
3. When using :iterate (:recurse) each iterative result set should be limited by a max nu,ber of results
4. When executing recursive queries (circular references) , the recursion depth needs to be limited to avoid performance issues

PR #1391, opened 2020-10-24
Feature/iterate cyclic unsupported

limorl:
## Description
This PR is a continuation of https://github.com/microsoft/fhir-server/pull/1314.
It addresses two cases:

1. Cyclic include iterate queries - which are not supported.
   That is, include iterate queries that include a (transitive) cycle. For example, in this query the parameter _include:iterate=DiagnosticReport:patient produces what Patient:general-practitioner consumes: 
`/MedicationDispense?_include=MedicationDispense:prescription&_include:iterate=MedicationRequest:patient&_include:iterate=Patient:general-practitioner&_revinclude:iterate=DiagnosticReport:performer:Practitioner&_include:iterate=DiagnosticReport:patient&_id=12345`
2. Include Iterate parameters with Circular Reference - which according to the spec results in multiple iterations and up to the implementer to decide how many. We implement it with one iteration and the  For example, in this query, 'partof' reference is os type Organization:
`Organization?_include:iterate=Organization:partof`
  
## Related issues
Addresses #1313

## Testing
1. Unit tests in IncludeRewriterTests to assert that exception is thrown for cyclic queries
2. E2E tests in IncludeSearchTests to assert that for a include iterate parameters with circular reference, the iteration is executed once and an informational issue is added to indicate that

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/IncludeExpression.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/IncludeExpression.cs
@@ -51,7 +51,8 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Expressions
             WildCard = wildCard;
             Reversed = reversed;
             Iterate = iterate;
-            CircularReference = SourceResourceType == TargetResourceType || (ReferenceSearchParameter?.TargetResourceTypes != null && ReferenceSearchParameter.TargetResourceTypes.Contains(sourceResourceType));
+            CircularReference = TargetResourceType != null ? SourceResourceType == TargetResourceType
+                                                           : ReferenceSearchParameter?.TargetResourceTypes != null && ReferenceSearchParameter.TargetResourceTypes.Contains(sourceResourceType);
         }
 
         /// <summary>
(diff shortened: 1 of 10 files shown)


limorl:
@johnstairs I've given it some thought,  I think we should consider relaxing the exception on Cyclic include iterate queries and take a similar approach to circular reference. That is if a cyclic (transitive) query exists, we can execute is it once, the order of the loop is based on the original query, and then provide informational issue as an operation outcome.
It's basically similar to circular reference, only a transitive recursion, so why not execute one iteration?
WDYT?

limorl:
@LTA-Thinking Looks like Cosmos E2E tests fail, even though the changes are related to SQL only.
There are also failures due to tests running more than 60 minutes.
Is this an environment issue?

LTA-Thinking:
@limorl It looks like some flaky tests. I'm rerunning them and will keep an eye on it.

limorl:
> @limorl It looks like some flaky tests. I'm rerunning them and will keep an eye on it.

Thanks a lot!

johnstairs (approved):

johnstairs:
Thanks @limorl!

PR #1396, opened 2020-10-26
Implement IHostedService instead of IStartable and avoid blocking on async calls

johnstairs:
## Description
- Incorporates shared components build 1.0.0-master-20201022-1
- Implements `IHostedService` instead of `IStartable`
- Eliminates calls to `Task.Result`, `Task.Wait()`, and `Task.GetAwaiter().GetResult()` in test code which can result in deadlocks with xUnit's custom `SynchronizationContext`. (Some are still required, ironically, in our xUnit extension, because of the xunit design)
- Fixed test instabilities resulting from modification of `SearchParameterDefinitionManager`s that were shared across test classes.

### DIFF
--- a/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Security/RoleLoaderTests.cs
+++ b/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Security/RoleLoaderTests.cs
@@ -8,6 +8,8 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
 using Microsoft.Extensions.FileProviders;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Health.Fhir.Core.Configs;
@@ -156,7 +158,7 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Security
         }
 
         [Fact]
-        public void GivenValidRoles_WhenLoaded_AreProperlyTransformed()
+        public async Task GivenValidRoles_WhenLoaded_AreProperlyTransformed()
         {
             var roles = new
             {
@@ -172,7 +174,7 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Security
                 },
             };
 
-            AuthorizationConfiguration authConfig = Load(roles);
+            AuthorizationConfiguration authConfig = await Load(roles);
 
             Role actualRole = Assert.Single(authConfig.Roles);
             Assert.Equal(roles.roles.First().name, actualRole.Name);
@@ -180,7 +182,7 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Security
         }
 
         [Fact]
-        public void GivenValidDataActions_WhenSpecifiedAsRoleActions_AreRecognized()
+        public async Task GivenValidDataActions_WhenSpecifiedAsRoleActions_AreRecognized()
         {
             var actionNames = Enum.GetValues(typeof(DataActions)).Cast<DataActions>()
                 .Where(a => a != DataActions.All && a != DataActions.None);
@@ -197,7 +199,7 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Security
                     }).ToArray(),
             };
 
-            AuthorizationConfiguration authConfig = Load(roles);
+            AuthorizationConfiguration authConfig = await Load(roles);
 
             Assert.All(
                 actionNames.Zip(authConfig.Roles.Select(r => r.AllowedDataActions)),
@@ -206,13 +208,13 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Security
 
         [Theory]
         [MemberData(nameof(GetInvalidRoles))]
-        public void GivenInvalidRoles_WhenLoaded_RaiseValidationErrors(string description, object roles)
+        public async Task GivenInvalidRoles_WhenLoaded_RaiseValidationErrors(string description, object roles)
         {
             Assert.NotEmpty(description);
-            Assert.Throws<InvalidDefinitionException>(() => Load(roles));
+            await Assert.ThrowsAsync<InvalidDefinitionException>(() => Load(roles));
         }
 
-        private static AuthorizationConfiguration Load(object roles)
+        private static async Task<AuthorizationConfiguration> Load(object roles)
         {
             IFileProvider fileProvider = Substitute.For<IFileProvider>();
             var hostEnvironment = Substitute.For<IHostEnvironment>();
@@ -223,7 +225,7 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Security
 
             var authConfig = new AuthorizationConfiguration();
             var roleLoader = new RoleLoader(authConfig, hostEnvironment);
-            roleLoader.Start();
+            await roleLoader.StartAsync(CancellationToken.None);
             return authConfig;
         }
     }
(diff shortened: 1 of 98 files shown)


src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosContainerProvider.cs
```diff
@@ -73,14 +74,19 @@ public Container Container
         /// <summary>
         /// Starts the initialization of the document client and cosmos data store.
         /// </summary>
-        public void Start()
+        /// <param name="cancellationToken">The cancellation token</param>
```
Ivanidzo4ka, line 77: >token</ [](start = 61, length = 7)

super nit:
would be nice to add `.` in the end of sentence :)

src/Microsoft.Health.Fhir.Core/Features/Search/CodeSystemResolver.cs
```diff
@@ -44,11 +46,18 @@ public string ResolveSystem(string shortPath)
...
+        public Task StartAsync(CancellationToken cancellationToken)
         {
             using Stream file = _modelInfoProvider.OpenVersionedFileStream("resourcepath-codesystem-mappings.json");
             using var reader = new StreamReader(file);
             _dictionary = JsonConvert.DeserializeObject<Dictionary<string, string>>(reader.ReadToEnd());
+
+            return Task.CompletedTask;
+        }
+
+        Task IHostedService.StopAsync(CancellationToken cancellationToken)
```
brendankowitz, line 58: Some thought for future, it might have been nice to have an abstract base so the empty `StopAsync` this is not needed everywhere

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/SearchParameters/SearchParameterFixtureData.cs
```diff
@@ -23,25 +24,35 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Search
...
-        public static SupportedSearchParameterDefinitionManager SupportedSearchDefinitionManager { get; }
+        public async Task<SearchParameterDefinitionManager> GetSearchDefinitionManager()
+        {
+            return _searchDefinitionManager ??= await CreateSearchParameterDefinitionManager(new VersionSpecificModelInfoProvider());
+        }
 
-        public static FhirNodeToSearchValueTypeConverterManager Manager { get; } = CreateFhirElementToSearchValueTypeConverterManager();
+        public async Task<SupportedSearchParameterDefinitionManager> GetSupportedSearchDefinitionManager()
+        {
+            return _supportedSearchDefinitionManager ??= new SupportedSearchParameterDefinitionManager(await GetSearchDefinitionManager());
```
Ivanidzo4ka, line 47: >GetSearchDefinitionManager [](start = 109, length = 26)

Changes here happened mostly because it's impossible to have async constructor.
I understand what as soon as you go async, everything upstream has to be async as well. But if downstream code you using remain sync, you can continue to be sync in upstream until you reach async behavior.

I've tried to track classes, and they mostly look like they remain sync, just wrapped in async logic. (can be easily wrong and maybe there is dependency on async code somewhere at the bottom)
Is it possible to have two interfaces and have `IAsyncHostedService` and `IHostedService` and use them appropriately?

Yes, this would make people question which interface to use, and maybe eventually everything would converge to async version, but on plus side, some code can be simplified, like here, or other places where all is changed is `async` added with Task.Complete.

brendankowitz (approved):
A lot of good clean up in there, thanks for taking this on

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/SearchParameters/SearchParameterFixtureData.cs
brendankowitz: We had `IStartable` that was synchronous, this PR was to move everything to IHostedService which provides the Async Start method. I am interested to learn your experience here, what do you see as the biggest downsides/concerns for having only the async version? Is it the sync=>async wrapping logic/mismatch, or performance or other?

Ivanidzo4ka: Never run benchmarks to compare how sync vs async code would perform (if both act in a sync way) (I run tonight, for my own education), but I doubt performance wise it's stark contrast.

My main concern is mental overload which comes with async. Why be complex if you can be simple.

That been said, I just notice GetAwaiter() in line 79 (left) which mean this whole thing need to be async anyway.

---
In reply to: [512256266](https://github.com/microsoft/fhir-server/pull/1396#discussion_r512256266) [](ancestors = 512256266)

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/SearchParameters/SearchConverterForAllSearchTypes.cs
```diff
@@ -154,7 +161,14 @@ public void ListAllUnsupportedTypes()
 
         public static IEnumerable<object[]> GetAllSearchParameters()
         {
-            var manager = SearchParameterFixtureData.CreateSearchParameterDefinitionManager(ModelInfoProvider.Instance);
+            Task<SearchParameterDefinitionManager> searchParameterDefinitionManagerTask = SearchParameterFixtureData.CreateSearchParameterDefinitionManager(ModelInfoProvider.Instance);
+
+            // XUnit does not currently support async signatures for MemberDataAttributes. Until it does we need to block on
+            // this task, which could cause a deadlock, but we know that the task should have completed synchronously,
+            // so there should not be a problem.
```
rotodd, line 168: +1 it's nice to have this explanation here

rotodd (approved):

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Security/RoleLoaderTests.cs
```diff
@@ -206,13 +208,13 @@ public void GivenValidDataActions_WhenSpecifiedAsRoleActions_AreRecognized()
...
-        public void GivenInvalidRoles_WhenLoaded_RaiseValidationErrors(string description, object roles)
+        public async Task GivenInvalidRoles_WhenLoaded_RaiseValidationErrors(string description, object roles)
         {
             Assert.NotEmpty(description);
-            Assert.Throws<InvalidDefinitionException>(() => Load(roles));
+            await Assert.ThrowsAsync<InvalidDefinitionException>(() => Load(roles));
         }
 
-        private static AuthorizationConfiguration Load(object roles)
+        private static async Task<AuthorizationConfiguration> Load(object roles)
```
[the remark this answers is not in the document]
johnstairs: A crude regex search (`async\s+Task\S*\s+(?![^_(]*_)(?![^(]+Async\()[^(]+\(`) suggest there might be something like 193 such methods. I'm not sure this is really all that necessary. One convention is to add Async only when there is a synchronous version of the method. But we're not very consistent (sometimes we use the suffix, sometimes we don't)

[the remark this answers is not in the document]
johnstairs: I see the asp.net team prefers the suffix: https://github.com/dotnet/aspnetcore/wiki/Engineering-guidelines#async-method-patterns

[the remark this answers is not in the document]
rotodd: I agree that it didn't make sense to rename 193 methods in this PR!

To your point on consistency - it might be nice to create a story to look into making our async naming a bit more uniform (whether that's following the asp.net team's standard or only adding "async" if there is a synchronous version). Thoughts?

PR #1403, opened 2020-10-28
Improve E2E test stability

johnstairs:
- Preserve session tokens across test classes, so that we can read our own writes. This is important for the (definitely suboptimal) approach we take of deleting all existing patients when initializing test class.
- Changed the DeleteAllResources implementation to follow continuation tokens. With Cosmos, you can get an empty bundle with a continuation token.
- Added retrying when we get a 429.
- Changed HistoryTests to use an idempotent create via `PUT` calls instead of `POST`. With POST, in the event of a retry because of a network failure, we could end up with several resources being created when we intended to just have one.


test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/HistoryTests.cs
```diff
@@ -236,38 +243,38 @@ public async Task GivenAQueryThatReturnsMoreThan10Results_WhenGettingSystemHisto
...
 
             using FhirResponse<Bundle> firstBatch = await _client.SearchAsync($"_history?_tag={tag}&_since={sinceUriString}&_before={beforeUriString}");
 
-            Assert.Equal(10, firstBatch.Resource.Entry.Count);
+            AssertCount(10, firstBatch.Resource.Entry);
 
             var secondBatch = await _client.SearchAsync(firstBatch.Resource.NextLink.ToString());
 
-            Assert.Single(secondBatch.Resource.Entry);
+            AssertCount(1, secondBatch.Resource.Entry);
```
rbans96, line 261: Just for my knowldege, Does AssertCount performs better than the Assert.Single?

rbans96 (approved):

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/HistoryTests.cs
johnstairs: No, and the implementation is below. It just includes more information when the failure occurs.

Issue #1389, opened 2020-10-23
Comma escaping not working per spec


CaitlinV39:
**Describe the bug**
Comma escaping not working per spec: https://www.hl7.org/fhir/search.html#escaping 

**FHIR Version?**
Stu3/R4

**Data provider?**
SQL (may also be in Cosmos)

**To Reproduce**
https://{{fhirserver}}/Practitioner?name=KEBER MD`\`,BARBARA A ; 

**Expected behavior**
Should return **Keber MD, Barbara A**

**Actual behavior**
It is searching for **Kerber MD** and **Barbara A** as two entries

CaitlinV39:
@brendankowitz  - FYI

CaitlinV39:
@Ivanidzo4ka  - Created AB#77071 for this for you

brendankowitz:
Verified using a few scenarios using the following, it looks to be generating the correct queries in these situations and returning the expected results.
-  `/Patient?name:exact=KEBER MD\, BARBARA A`
-  `/Patient?name:exact=KEBER MD, BARBARA A`
-  `/Patient?name:exact=KEBER MD`
-  `/Patient?name=KEBER MD`
-  `/Patient?name=KEBER MD, BARBARA A`

PR #1406, opened 2020-10-28
Escape string search values.

Ivanidzo4ka:
## Description
Make sure strings get escaped before we perform search on them.

## Related issues
Addresses [#1389].

## Testing
Manually and Unit-tests.

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/ReferenceSearchValueParser.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/ReferenceSearchValueParser.cs
@@ -40,7 +40,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.SearchValues
         public ReferenceSearchValue Parse(string s)
         {
             EnsureArg.IsNotNullOrWhiteSpace(s, nameof(s));
-
+            s = s.UnescapeSearchParameterValue();
             Match match = ReferenceRegex.Match(s);
 
             if (match.Success)
--- a/src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/ReferenceSearchValueParser.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/ReferenceSearchValueParser.cs
@@ -40,7 +40,6 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.SearchValues
         public ReferenceSearchValue Parse(string s)
         {
             EnsureArg.IsNotNullOrWhiteSpace(s, nameof(s));
-            s = s.UnescapeSearchParameterValue();
             Match match = ReferenceRegex.Match(s);
 
             if (match.Success)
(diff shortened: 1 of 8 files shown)


johnstairs (approved):

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/ReferenceSearchValueParser.cs
```diff
@@ -40,7 +40,7 @@ public ReferenceSearchValueParser(IFhirRequestContextAccessor fhirRequestContext
         public ReferenceSearchValue Parse(string s)
         {
             EnsureArg.IsNotNullOrWhiteSpace(s, nameof(s));
-
+            s = s.UnescapeSearchParameterValue();
```
brendankowitz, line 43: I'm a little surprised this didn't break a unit test also :) i.e. https://github.com/microsoft/fhir-server/blob/master/src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/SearchValues/ReferenceSearchValueParserTests.cs

Ivanidzo4ka: I've look and it seems like there is regex in reference search parameter which cuts values and doesn't let any special characters like ',' or `|` or `$`
I'm a bit scary to modify that regex, so I'm gonna be defensive and remove my changes in reference search param

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/StringSearchValue.cs
```diff
@@ -42,7 +42,7 @@ public static StringSearchValue Parse(string s)
         {
             EnsureArg.IsNotNullOrWhiteSpace(s, nameof(s));
 
-            return new StringSearchValue(s);
+            return new StringSearchValue(s.UnescapeSearchParameterValue());
```
brendankowitz, line 45: Does/should this have an effect on https://github.com/microsoft/fhir-server/blob/master/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Search/SearchValues/StringSearchValueTests.cs#L65?

Can we add a test for `StringSearchValue.String` to verify this behavior?

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/ReferenceSearchValueParser.cs
brendankowitz: I think that sounds ok. The validation for an id (https://www.hl7.org/fhir/datatypes.html) looks like it can't include these

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/StringSearchValue.cs
Ivanidzo4ka: Added.

brendankowitz: Thanks, and good point. Is there anything we can do that would make this more obvious? Would updating the comment on `String` and `ToString` help?

Ivanidzo4ka: I've update comment on top of `String` property. Not sure that changes saved me in past of been confused, but that confusion is purely based on reusing word `String` to many times in class :)

PR #1412, opened 2020-11-02
Updates package 'Microsoft.Azure.Cosmos' to version '3.15.0'

MicrosoftHealthService:


src/Microsoft.Health.Fhir.CosmosDb/Features/Queries/CosmosQuery.cs
```diff
@@ -91,6 +95,11 @@ public async Task<FeedResponse<T>> ExecuteNextAsync(CancellationToken token = de
                 // The SDK wraps exceptions we throw in handlers with a CosmosException.
                 Exception fhirException = ex.InnerException as FhirException ?? ex.InnerException as MicrosoftHealthException;
 
+                if (fhirException == null)
+                {
+                    _processor.ProcessErrorResponse(ex.StatusCode, ex.Headers, ex.Message);
```
johnstairs, line 100: Is there a specific scenario we're targeting here? Like the exception was thrown by the SDK but did not make it to our handlers? In that case, we would have similar issues with other SDK entrypoints, like when we execute a stored proc, do a direct read, etc.

brendankowitz: I have a more generic fix I was playing around with: https://github.com/microsoft/fhir-server/compare/personal/bkowitz/cosmos-mediatr-errors?expand=1#diff-62aca314a00664c19a0bd48fbf4bbc40a989f3cccffaa05abc4f452a60a31b9cR32 which would work for any request.

The current downside is that the exception was still getting wrapped and required an additional modification: https://github.com/microsoft/fhir-server/compare/personal/bkowitz/cosmos-mediatr-errors?expand=1#diff-58efbf37ff862f5393d897e95e8f52af324d9775b8a2ca673c2682b4259c77e4R52.

I wasn't sure that would was too much change to also introduce here.

johnstairs: I like that approach, but I might suggest we stick to just the nuget upgrade in this PR.

brendankowitz: @johnstairs, to answer your original question, this catches an Invalid Continuation Token exception

johnstairs: Then we should try to remove this catch block: https://github.com/Microsoft/fhir-server/blob/7c01a811956f2f55c9e30223cf6824ec13399598/src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs#L358-L362

brendankowitz: Thanks, this might not be needed. I think it would be more ideal to keep all these checks in the error processor

johnstairs (approved):

Issue #1115, opened 2020-07-01
Paged search results made with POST don't support subsequent GET  


holpit:
**Describe the bug**
I'm using the https://github.com/FirelyTeam
When the result of search done with a POST request has more pages, a subsequent call to the next link raise an error

**To Reproduce**
1. Make a search with POST https://fhirserverxyz/ClinicalImpression/_search
with x-www-form-urlencoded parameters:
_include ClinicalImpression:subject
_count  5

2. The bundle has the next link:  "relation": "next", "url": "https://fhirserverxyz/ClinicalImpression/_search?ct=NTA5ODI4ODQ3NTQ4MzY4MDAyMA%3D%3D" 
3. Do a GET with the link to the next page returned throw the error :
`{
    "resourceType": "OperationOutcome",
    "id": "362fda35-053a-4d90-a2e8-9a67838bd245",
    "issue": [
        {
            "severity": "error",
            "code": "invalid",
            "diagnostics": "Id must be any combination of upper or lower case ASCII letters ('A'..'Z', and 'a'..'z'), numerals ('0'..'9'), '-' and '.', with a length limit of 64 characters. (This might be an integer, an un-prefixed OID, UUID, or any other identifier pattern that meets these constraints.)",
            "location": [
                "ResourceKey.Id"
            ]
        }
    ]
}`

**Expected behavior**
The next link must be executed with a GET request as reported in documentation:
http://hl7.org/fhir/http.html#paging
In the case of a search, the initial request may be made via a POST, but the follow up page requests will be made via GET requests. However servers SHOULD allow for a client to convert the follow up requests to be made via a POST.

**Actual behavior**
The next link works only if made with POST to the next url

CaitlinV39:
AB#76144

feordin:
Verified as fixed.

PR #1414, opened 2020-11-02
Add "POST" search rerouting middleware.

Ivanidzo4ka:
## Description
Adds middleware which reroute all "post" search requests to "get" searches.

## Related issues
Addresses [#1115 ].

## Testing
Manually and via unit tests


src/Microsoft.Health.Fhir.Api/Features/Routing/SearchPostReroutingMiddleware.cs
```diff
@@ -0,0 +1,48 @@
...
+
+namespace Microsoft.Health.Fhir.Api.Features.Routing
+{
+    public class SearchPostReroutingMiddleware
+    {
+        private readonly RequestDelegate _next;
+
+        public SearchPostReroutingMiddleware(RequestDelegate next)
+        {
+            _next = next;
```
brendankowitz, line 18: Add `EnsureArg.IsNotNull` for parameters for consistency

src/Microsoft.Health.Fhir.Api/Features/Routing/SearchPostReroutingMiddleware.cs
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        public async Task Invoke(HttpContext context)
+        {
+            var request = context.Request;
+
+            if (request != null
+                && request.Method == "POST"
+                && request.Path.Value.EndsWith(KnownRoutes.Search, System.StringComparison.OrdinalIgnoreCase)
+                && request.ContentType == "application/x-www-form-urlencoded")
```
LTA-Thinking, line 28: If a user makes a request with the wrong/missing content type they get a 404 back. I feel this should be a 400 informing them of the wrong content type.

src/Microsoft.Health.Fhir.Api/Features/Routing/SearchPostReroutingMiddleware.cs
```diff
@@ -0,0 +1,60 @@
...
+                    }
+
+                    request.Path = request.Path.Value.Substring(0, request.Path.Value.Length - KnownRoutes.Search.Length);
+                    request.Method = "GET";
+                }
+                else
+                {
+                    context.Response.Clear();
+                    context.Response.StatusCode = (int)HttpStatusCode.BadRequest;
+                    await context.Response.WriteAsync(Resources.ContentTypeFormUrlEncodedExpected);
```
LTA-Thinking, line 52: I think we should have an operation outcome wrapping the error message like we do elsewhere. Example: Post to /Patient with a Content-Type of application/text

Ivanidzo4ka: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/middleware/index/_static/middleware-pipeline.svg?view=aspnetcore-3.1
image

I'm using custom middleware which is happening before request hitting endpoint.
So far most of exception conversion happens inside filter on top of controller/endpoint.

Right now main reason why I don't return OperationOutcome is because it's not accessible for me here.
I include this middleware to non-spec specific project, and operation outcome is part of standard.

Considering it's a bit hard for me to figure out how to make it work, I would argue what we don't have operation outcome result for every error, there is still plenty of 404 or 401 which are not wrapped I believe.

brendankowitz (approved):

LTA-Thinking (approved):

Issue #1395, opened 2020-10-26
Improve parser error messages


brendankowitz:
**Describe the bug**
The following questionnaire resource has two problems:

1. The parser is throwing a format exception when converting `00` to a decimal
2. The exception message that is thrown is generic and does not indicate the source of the error

**FHIR Version?**
Stu3/R4/R5

**Data provider?**
All

**To Reproduce**
1. Post the resource below

**Expected behavior**
Resource should be accepted successfully

**Actual behavior**
`00` causes an exception that is returned in a non-detailed message, giving no indication of the problem

```
{
    "resourceType": "Questionnaire",
    "id": "0001",
    "meta": {
      "versionId": "1",
      "lastUpdated": "2020-09-27T02:33:59.000-04:00",
      "profile": [
        "http://hl7.org/fhir/uv/sdc/StructureDefinition/sdc-questionnaire|2.7"
      ],
      "tag": [
        {
          "code": "lformsVersion: 25.1.5"
        }
      ]
    },
    "extension": [
      {
        "url": "http://hl7.org/fhir/StructureDefinition/variable",
        "valueExpression": {
          "name": "weight",
          "language": "text/fhirpath",
          "expression": "%resource.item.where(linkId='/29463-7').answer.valueDecimal"
        }
      },
      {
        "url": "http://hl7.org/fhir/StructureDefinition/variable",
        "valueExpression": {
          "name": "height",
          "language": "text/fhirpath",
          "expression": "%resource.item.where(linkId='/8302-2').answer.valueDecimal*0.0254"
        }
      }
    ],
    "url": "https://lforms-smart-fhir.nlm.nih.gov/v/r4/fhir/Questionnaire/55418-8-x",
    "identifier": [
      {
        "system": "http://loinc.org",
        "value": "55418-8"
      }
    ],
    "name": "weight_height_tracking_panel",
    "title": "Weight & Height tracking panel",
    "status": "draft",
    "subjectType": [
      "Patient"
    ],
    "date": "2018-09-12T18:03:40-04:00",
    "code": [
      {
        "system": "http://loinc.org",
        "code": "55418-8",
        "display": "Weight and Height tracking panel"
      }
    ],
    "item": [
      {
        "extension": [
          {
            "url": "http://hl7.org/fhir/StructureDefinition/questionnaire-unit",
            "valueCoding": {
              "system": "http://unitsofmeasure.org",
              "code": "kg"
            }
          },
          {
[...]

PR #1417, opened 2020-11-02
Add underlying parsing error to meassage.

Ivanidzo4ka:
## Description
Improve error message in case of parsing of document failed.

## Related issues
Addresses [#1395].

## Testing
TBD

### DIFF
--- a/src/Microsoft.Health.Fhir.Api/Resources.resx
+++ b/src/Microsoft.Health.Fhir.Api/Resources.resx
@@ -287,7 +287,7 @@
     <comment>{0} is the number of entries allowed.</comment>
   </data>
   <data name="ParsingError" xml:space="preserve">
-    <value>Error occurred when parsing model.</value>
+    <value>Error occurred when parsing model: '{0}'.</value>
   </data>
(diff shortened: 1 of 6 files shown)


brendankowitz:
I think there was some history with this also, i.e. it used to give an unhelpful error message when the resourceType property was missing or blank

Ivanidzo4ka:
And solution for that was to hide error message completely and replace it with static message. I prefer to go hybrid route, still have static message indicating it's something wrong during parsing, but also include parsing error, so that can give idea to user what is wrong.

---
In reply to: [720791503](https://github.com/microsoft/fhir-server/pull/1417#issuecomment-720791503) [](ancestors = 720791503)

brendankowitz:
Sounds good. 
One day I'd like to invest more in the lower level parsers like (https://github.com/microsoft/fhir-server/blob/personal/bkowitz/json-pipeline-optimizations/src/Microsoft.Health.Fhir.Shared.Api/Features/Formatters/FhirJsonInputFormatter.cs#L84), but its a bit of a concept shift through the entire pipelines

brendankowitz (approved):

PR #1420, opened 2020-11-04
Delete cleanup action

LTA-Thinking:
## Description
The action won't work until we are able to pass parameters to the DevOps job.

## Testing


Ivanidzo4ka (approved):
:shipit:

PR #1422, opened 2020-11-05
Add export formatting

LTA-Thinking:
## Description
Adds a setting for configuring the file structure format for export jobs. 

$export now supports the _format parameter that lets a user select a format for the file structure the export job creates. Different formats can be defined in the appSettings by combining constants, folder level breaks ('/'), and known tags. The tags will be replaced with different pieces of data when the job is run. Currently three tags are supported: resource type, date time, and id. Resource type is replaced with the resource type currently being exported. Date time is replaced with a timestamp of the job's queried time. Id is replaced with the GUID of the export job. 

Two export formats are included with this change to prevent regression. One is for when a container is specified, the other is for when one isn't. These create the same expected behavior as currently exists. 

## Related issues
This PR removes a check in the ExportJobTask for old continuation tokens that was added as part of #591. As the change as been live for several weeks now no old continuation tokens should still be present in export jobs.

## Testing
Existing and new unit and e2e tests added.


src/Microsoft.Health.Fhir.Core/Features/Operations/Export/Models/ExportJobRecord.cs
```diff
@@ -20,6 +20,7 @@ public class ExportJobRecord : JobRecord
         public ExportJobRecord(
             Uri requestUri,
             ExportJobType exportType,
+            string exportFormat,
```
Ivanidzo4ka, line 23: >exportFormat [](start = 19, length = 12)

Is it ok to pass null here? if yes, why it's not optional parameter like others below.
If it's no ok, can EnsureArg be added?

src/Microsoft.Health.Fhir.Core/Configs/ExportJobFormatConfiguration.cs
```diff
@@ -0,0 +1,34 @@
...
+        public string Name { get; set; } = string.Empty;
+
+        /// <summary>
+        /// The format definition string. An export job's format is used to create the folder stucture inside the container.
+        /// The format is defined by tags and characters. Supported tags are defined below. The / character is used to indicate a subfolder.
+        /// <timestamp> - Places a timestamp corisponding to the time the export job was enqueued.
+        /// <resourcename> - The name of the resource currently being exported.
+        /// <id> - The GUID id of the export job.
+        /// </summary>
+        public string Format { get; set; }
```
Ivanidzo4ka, line 22: >Format [](start = 22, length = 6)

Is it worth to default to past behavior of '<resourcename>'?

src/Microsoft.Health.Fhir.Core/Configs/ExportJobConfiguration.cs
```diff
@@ -45,5 +46,12 @@ public class ExportJobConfiguration
         /// Number of pages to be iterated before committing the export progress.
         /// </summary>
         public uint NumberOfPagesPerCommit { get; set; } = 10;
+
+        /// <summary>
+        /// Formats for export jobs.
+        /// </summary>
+#pragma warning disable CA2227 // Collection properties should be read only
+        public IList<ExportJobFormatConfiguration> ExportJobFormats { get; set; }
```
Ivanidzo4ka, line 54: >ExportJobFormats [](start = 51, length = 16)

There is no validation for this collection, so there is chance of having two jobs formats with same name, and first one would prevail. It's probably ok, just want to point that out.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/CreateExportRequestHandler.cs
```diff
@@ -76,11 +76,45 @@ public async Task<CreateExportResponse> Handle(CreateExportRequest request, Canc
...
+
+            if (request.Format != null)
+            {
+                formatConfiguration = _exportJobConfiguration.ExportJobFormats.FirstOrDefault(
+                (ExportJobFormatConfiguration formatConfig) => formatConfig.Name.Equals(request.Format, StringComparison.OrdinalIgnoreCase));
+            }
+
+            if (formatConfiguration == null)
+            {
+                if (request.ContainerName != null)
```
Ivanidzo4ka, line 89: >request.ContainerName != null [](start = 20, length = 29)

I think you can push this condition inside FirstOrDefault and that would make code a bit less bloated.
`formatConfiguration = _exportJobConfiguration.ExportJobFormats.FirstOrDefault((ExportJobFormatConfiguration formatConfig) => request.ContainerName!=null? formatConfig.ContainerDefault:formatConfig.NoContainerDefault);`

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/CreateExportRequestHandler.cs
```diff
@@ -76,11 +76,45 @@ public async Task<CreateExportResponse> Handle(CreateExportRequest request, Canc
...
+                    formatConfiguration = _exportJobConfiguration.ExportJobFormats.FirstOrDefault();
+                }
+
+                if (formatConfiguration == null)
+                {
+                    formatConfiguration = new ExportJobFormatConfiguration()
+                    {
+                        Format = ExportFormatTags.ResourceName,
+                    };
+                }
```
Ivanidzo4ka, line 109: If you interested. there is [https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-coalescing-operator](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-coalescing-operator) which allow to compact this code a little bit.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -228,13 +225,7 @@ private async Task UpdateJobRecordAsync(CancellationToken cancellationToken)
...
             if (progress.ContinuationToken != null)
             {
-                var continuationToken = progress.ContinuationToken;
-                if (!Base64FormatRegex.IsMatch(continuationToken))
-                {
-                    continuationToken = Convert.ToBase64String(System.Text.Encoding.UTF8.GetBytes(continuationToken));
-                }
-
-                queryParametersList.Add(Tuple.Create(KnownQueryParameterNames.ContinuationToken, continuationToken));
+                queryParametersList.Add(Tuple.Create(KnownQueryParameterNames.ContinuationToken, progress.ContinuationToken));
```
Ivanidzo4ka, line 228: >progress.ContinuationToken [](start = 97, length = 26)

Can I ask why?
Was decoding added accidently?

src/Microsoft.Health.Fhir.Core/Configs/ExportJobConfiguration.cs
LTA-Thinking: That is a concern, but since these are user controlled formats if we added any duplicate prevention (like appending something to the name) the user wouldn't know how to request it from the _format parameter. I'll add a note to the documentation in ExportJobFormatConfiguration to not have duplicate names, but I don't think there is a good way to avoid this.

src/Microsoft.Health.Fhir.Core/Configs/ExportJobFormatConfiguration.cs
LTA-Thinking: My concern with setting a default value here is that we currently have two default values: one for when a container is set and one for when one isn't set. I added both to the CreateExportRequestHandler as fallbacks.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
LTA-Thinking: It was added as part of #591. Now that the change has been live for a few weeks there shouldn't be any old continuation tokens still floating around. This is just the first change in this area since then, so I'm cleaning it up.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/Models/ExportJobRecord.cs
LTA-Thinking: Good catch, this shouldn't allow null here. I'll add an EnsureArg.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
Ivanidzo4ka: Any chance you can add description of that change and reasoning for that in PR description?
In case if someone have to blame file, and check changes, going through comments in PR is a bit too much for ask...

src/Microsoft.Health.Fhir.Core/Messages/Export/CreateExportRequest.cs
```diff
@@ -13,7 +13,16 @@ namespace Microsoft.Health.Fhir.Core.Messages.Export
...
     {
-        public CreateExportRequest(Uri requestUri, ExportJobType requestType, string resourceType = null, PartialDateTime since = null, string groupId = null, string containerName = null, string anonymizationConfigurationLocation = null, string anonymizationConfigurationFileETag = null)
+        public CreateExportRequest(
+            Uri requestUri,
+            ExportJobType requestType,
+            string resourceType = null,
+            PartialDateTime since = null,
+            string groupId = null,
+            string containerName = null,
+            string format = null,
```
Ivanidzo4ka, line 23: >format [](start = 19, length = 6)

If we decided to not allow format to be `null` in `ExportJobRecord` we probably shouldn't allow it be null here as well.

src/Microsoft.Health.Fhir.Core/Configs/ExportJobFormatConfiguration.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        /// <summary>
+        /// Whether the format is the default format for when no container is specified by the user.
+        /// </summary>
+        public bool NoContainerDefault { get; set; } = false;
+
+        /// <summary>
+        /// Whether the format is the default format for when a container is specified by the user.
+        /// </summary>
+        public bool ContainerDefault { get; set; } = false;
```
Ivanidzo4ka, line 32: From how it's used, it feels more like a bool switch to determine is this container specific or not.
Any chance this can be reduced to just one variable? (Container/NotContainer)
Or even moved it out, and having two configurations one container specific and other non container?

I understand desire to have default, but nothing stops people to define multiple default configurations since there is no validation, so maybe just document what first configuration is default one if no name specified?

Ivanidzo4ka: Ok, maybe two collections for container specified/not specified is a bit of overkill, it doesn't make sense from user perspective to have to different collections.

I would probably go following route:
If user didn't specify format, just don't pick configuration and leave old code in place.
If user specify format, and it's not found, I would fail with error to user.
If user specify format and it's found, use it despite is it container value.

That would allow you to preserve old behavior, and would remove this two flags, which from user standpoint only bring confusion. 

---
In reply to: [518329969](https://github.com/microsoft/fhir-server/pull/1422#discussion_r518329969) [](ancestors = 518329969)

Ivanidzo4ka: That been said I actually start questioning why we even have this collection in first place?

Why user can't just pass us `Format` string?

User need to learn how `Format` works anyway and in current moment he need to do roundtrip of going to app settings and looking for what was the format name for this specific format.

---
In reply to: [518341229](https://github.com/microsoft/fhir-server/pull/1422#discussion_r518341229) [](ancestors = 518341229,518329969)

LTA-Thinking: I agree that if a user enters a name that isn't found an error should be returned, I'll add that. The thought behind having two defaults is because currently we have two different behaviors depending on if a container is specified or not. The reason I added both was to minimize the hardcoded values we would need. 

As for just passing in the format, I wanted the users to have the ability to set up a standard that they could specify. That way not all users needed to know about the export format. The server admins would set up any that the organization needed and then the format could just be referenced by name. I'm open to just having them pass in the format. @CaitlinV39 thoughts?

CaitlinV39: @LTA-Thinking  and @Ivanidzo4ka unless its too inefficient, I like that the format could be setup and then other users would be able to just leverage it without needing to know the full format.

Ivanidzo4ka: Fine by me. My main concern is that two extra properties in configuration.
So here is my reasoning why I would prefer to hardcode them in code rather than expose as part of configuration.

I don't think there is much value in them other than back compatibility.  User already specifying name of format, it can be confusing if his format would change to something else depending on passed container value.
It's hard to server admin to configure, because this two new configuration bring more options, and they confusing.
Why presence of container is important? It used to be, but with passing format name, should it still be important?
I think it's better to have some hardcoded code for back-compat in code base, than force users to scratch their heads in order to figure out what they should put in this fields.

Just in case, I'm not saying there is no need to keep old code, I'm in favor in full export configuration change, just in place where we choose configuration we can check is format passed or not, and pass configuration 
format based on container presence if there is no format.

---
In reply to: [518920366](https://github.com/microsoft/fhir-server/pull/1422#discussion_r518920366) [](ancestors = 518920366)

src/Microsoft.Health.Fhir.Core/Messages/Export/CreateExportRequest.cs
LTA-Thinking: Here I think it is ok for it to be null because this is what the user has passed in the _format parameter. I'll rename this to formatName to make it more clear this is not the same value.

src/Microsoft.Health.Fhir.Core/Configs/ExportJobFormatConfiguration.cs
LTA-Thinking: I removed the two tags and replaced them with a single "Default" property. Neither of the included formats have the property set. If there is no default format and no format is entered the behavior will be identical to current behavior.

Ivanidzo4ka (approved):
:shipit:

src/Microsoft.Health.Fhir.Core/Resources.resx
```diff
@@ -453,4 +453,8 @@
     <value>Reindex job id {0} is in state {1} and cannot be cancelled.</value>
     <comment>{0} is the id of the reindex which is a a GUID.  {1} is the state, an enumerated set of state values.</comment>
   </data>
+  <data name="ExportFormatInvalid" xml:space="preserve">
```
Ivanidzo4ka, line 456: >ExportFormatInvalid [](start = 14, length = 19)

nit: Invalid in my head and couldn't be found is a bit different categories.

src/Microsoft.Health.Fhir.Core/Configs/ExportJobConfiguration.cs
```diff
@@ -45,5 +46,11 @@ public class ExportJobConfiguration
         /// Number of pages to be iterated before committing the export progress.
         /// </summary>
         public uint NumberOfPagesPerCommit { get; set; } = 10;
+
+        /// <summary>
+        /// Formats for export jobs.
+        /// </summary>
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Usage", "CA2227:Collection properties should be read only", Justification = "This is a configuration class")]
+        public IList<ExportJobFormatConfiguration> ExportJobFormats { get; set; }
```
Ivanidzo4ka, line 54: >ExportJobFormats [](start = 51, length = 16)

nit: This is already part of ExportJobConfiguration, so putting ExportJob in front of Formats sound a bit excessive for me.

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Features/Operations/Export/CreateExportRequestHandlerTests.cs
```diff
@@ -171,6 +180,90 @@ public async Task GivenThereIsAMatchingJob_WhenRequestorClaimsInDifferentOrder_T
...
+        }
+
+        [Fact]
+        public async Task GivenARequestWithAnNonexistantFormatName_WhenConverted_ThenABadRequestIsReturned()
+        {
+            var request = new CreateExportRequest(RequestUrl, ExportJobType.All, formatName: "invalid");
+            await Assert.ThrowsAsync<BadRequestException>(() => _createExportRequestHandler.Handle(request, _cancellationToken));
+        }
+
+        private T AddListener<T>(T baseObject)
```
Ivanidzo4ka, line 246: >AddListener [](start = 18, length = 11)

nit: I can checkout branch and put breakpoints inside Invoke method to find out what is going on here, but any chance some description can be added for this black magic reflection? Mainly what's it for?

PR #1425, opened 2020-11-05
Fixed search by _id + reference

### LINKED ISSUE #1423: _id search combined with reference does not filter by reference
**Describe the bug**
When searching by `_id` and reference type parameter in a single query, results are filtered by `_id` only, reference is not checked.

```curl
/Observation?subject=Patient/567&_id=123
```

Looking at the generated SQL query, reference parameter is in `cte0`, which is not referenced by any of the following ctes.
```sql
WITH cte0 AS
(
    FROM dbo.ReferenceSearchParam
    WHERE IsHistory = 0
        AND ResourceTypeId = @p0
        AND SearchParamId = @p1
        AND ReferenceResourceTypeId = @p2
        AND ReferenceResourceId = @p3

),
cte1 AS
(
    FROM dbo.Resource
    WHERE IsHistory = 0
        AND IsDeleted = 0
        AND ResourceTypeId = @p0
        AND ResourceId = @p4

),
cte2 AS
(
    SELECT DISTINCT TOP (@p5) Sid1, 1 AS IsMatch, 0 AS IsPartial
    FROM cte1
    ORDER BY  Sid1 ASC
)
SELECT DISTINCT r.ResourceTypeId, r.ResourceId, r.Version, r.IsDeleted, r.ResourceSurrogateId, r.RequestMethod, CAST(IsMatch AS bit) AS IsMatch, CAST(IsPartial AS bit) AS IsPartial, r.IsRawResourceMetaSet, r.RawResource
FROM dbo.Resource r
INNER JOIN cte2
ON r.ResourceSurrogateId = cte2.Sid1
```

**FHIR Version?**
R4 (tested), likely all

**Data provider?**
SQL Server

**To Reproduce**
1. Create Observation
    ```json
    {
        "id": "123",
        "status": "final",
        "code": {
            "coding": [
                {
                    "system": "http://loinc.org",
                    "code": "60832-3",
                    "display": "Room temperature"
                }
            ]
        },
        "subject": {
            "reference": "Patient/234",
            "type": "Patient"
        }
    }
    ```
2. Query by `_id` and reference not matching the observation created
    ```curl
    /Observation?subject=Patient/567&_id=123
    ```

**Expected behavior**
No results to be returned, because reference search parameter does not match

**Actual behavior**
Results matching `_id` are returned no matter if reference search parameter mathes the resource

anyname2:
## Description
Fixed Normalized Predicate Reorderer to sort `_id` expression first.

## Related issues
Fixes #1423.

## Testing
Added unit and E2E tests.

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/NormalizedPredicateReorderer.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/NormalizedPredicateReorderer.cs
@@ -26,6 +26,12 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Search.Expressions.Visitors
 
             List<TableExpression> reorderedExpressions = expression.TableExpressions.OrderByDescending(t =>
             {
+                // Sort _id (denormalized) expression to the front
+                if (t.NormalizedPredicate == null && t.DenormalizedPredicate != null && t.Kind == TableExpressionKind.All)
+                {
+                    return 20;
+                }
+
                 if (t.NormalizedPredicate is MissingSearchParameterExpression)
                 {
                     return -10;
--- a/src/Microsoft.Health.Fhir.SqlServer.UnitTests/Features/Search/Expressions/NormalizedPredicateReordererTests.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer.UnitTests/Features/Search/Expressions/NormalizedPredicateReordererTests.cs
@@ -18,6 +18,8 @@ namespace Microsoft.Health.Fhir.SqlServer.UnitTests.Features.Search.Expressions
 {
     public class NormalizedPredicateReordererTests
     {
+        private static readonly Expression NormalExpression = new SearchParameterExpression(new SearchParameterInfo("TestParam"), Expression.Equals(FieldName.TokenCode, null, "TestValue"));
+
         [Fact]
         public void GivenExpressionWithSingleTableExpression_WhenReordered_ReturnsOriginalExpression()
         {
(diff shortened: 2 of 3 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/NormalizedPredicateReorderer.cs
```diff
@@ -26,6 +26,12 @@ public override Expression VisitSqlRoot(SqlRootExpression expression, object con
 
             List<TableExpression> reorderedExpressions = expression.TableExpressions.OrderByDescending(t =>
             {
+                // Sort _id (denormalized) expression to the front
```
Ivanidzo4ka, line 29: >// Sort _id (denormalized) expression to the front [](start = 16, length = 50)

Took me a while to figure out why it's working.

There is this code:
[https://github.com/microsoft/fhir-server/blob/1cb5628235aff99285e82a7ef193c153ad3c04ac/src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/DenormalizedPredicateRewriter.cs#L62](https://github.com/microsoft/fhir-server/blob/1cb5628235aff99285e82a7ef193c153ad3c04ac/src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/DenormalizedPredicateRewriter.cs#L62)
which does denormalization and put Table expression "All" on top of list of all expressions.

Which happened to be reordered here. That "All" expression doesn't have `SearchParameterQueryGenerator` so it get 0 from default statement in switch, and other table expression has ReferenceSearchParameterQueryGenerator which pushes it to the top of expression.

This weird condition `t.NormalizedPredicate == null && t.DenormalizedPredicate != null && t.Kind == TableExpressionKind.All`

comes directly from file which I linked above, but to be fair it looks like this is common pattern to create All table expression (at least in current code) so maybe it can be partially truncated.

Ivanidzo4ka (approved):
:shipit:

brendankowitz:
@Ivanidzo4ka can you merge this

PR #1426, opened 2020-11-05
Implements _include and _revinclude for Cosmos DB

johnstairs:
## Description
Implements _include and _revinclude for the Cosmos DB provider. This is done by issuing multiple queries to the database. 

Does **not** include support for `_include:iterate` and `_revinclude:iterate`.

## Related issues
Addresses [AB#77190](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/77190).

## Testing
Existing E2E tests and manual tests.


src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
```diff
@@ -138,10 +203,156 @@ public class FhirCosmosSearchService : SearchService
...
+                }
+            }
+
+            if (revIncludeExpressions.Count > 0 && !includesTruncated)
+            {
+                // fetch in the resources to include from _revinclude parameters.
+
+                foreach (IncludeExpression revIncludeExpression in revIncludeExpressions)
+                {
+                    // construct the expression resourceType = <SourceResourceType> AND (referenceSearchParameter = <MatchResourceType1, MatchResourceId1> OR referenceSearchParameter = <MatchResourceType2, MatchResourceId2> OR ...)
```
brendankowitz, line 280: Is there some max number of conditions we need to check for?

johnstairs: Good point. Query size is [limited to 256KB](https://docs.microsoft.com/en-us/azure/cosmos-db/concepts-limits#sql-query-limits). We may need to partition in the case of forward include, where there are a lot of references.

johnstairs: For _revinclude, with the page size limit we have, we should be fine.

brendankowitz (approved):

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
johnstairs: Added partitioning

PR #1429, opened 2020-11-06
Reorder background workers

rbans96:
## Description
Fixes the issue on server startup when the database does not exists.
Prior to the fix, ExportJobWorker is starting up before the SchemaInitializer is run and throws the exception
 Microsoft.Health.Fhir.Core.Features.Operations.Export.ExportJobWorker[0]
      ExportJobWorker: Cancellation requested.
Unhandled exception. Microsoft.Data.SqlClient.SqlException (0x80131904): Cannot open database "FHIR_R4_50" requested by the login. The login failed.

## Related issues
Addresses [issue #].

## Testing
Tested manually with standing up a new database and server is able to start.

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Web/Startup.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Web/Startup.cs
@@ -29,7 +29,6 @@ namespace Microsoft.Health.Fhir.Web
             services.AddDevelopmentIdentityProvider(Configuration);
 
             Core.Registration.IFhirServerBuilder fhirServerBuilder = services.AddFhirServer(Configuration)
-                .AddBackgroundWorkers()
                 .AddAzureExportDestinationClient()
                 .AddAzureExportClientInitializer(Configuration);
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.Health.Fhir.Shared.Web/Startup.cs
```diff
@@ -43,6 +42,12 @@ public virtual void ConfigureServices(IServiceCollection services)
                 fhirServerBuilder.AddSqlServer(Configuration);
             }
 
+            /*
+            IHostedServices are executed depends on the order they are added to the DI Container.
+            So need to ensure that the schema is initialized before the background workers are started
+            */
```
rotodd, line 48: Nit: slight rewording - what do you think of this?

`The execution of IHostedServices depends on the order they are added to the dependency injection container, so we need to ensure that the schema is initialized before the background workers are started.`

I usually opt for typing out abbreviations instead of shortening them for clarity. Ideally someone who is looking at `Startup.cs` would know that DI = "dependency injection", but I still think it's nice to avoid short forms.

rotodd (approved):
Thank you for fixing this!

brendankowitz (approved):

src/Microsoft.Health.Fhir.Shared.Web/Startup.cs
Ivanidzo4ka: I also like `.` in the end of sentence :)

---
In reply to: [518921356](https://github.com/microsoft/fhir-server/pull/1429#discussion_r518921356) [](ancestors = 518921356)

Ivanidzo4ka (approved):
:shipit:

Ivanidzo4ka:
Just curious, is it possible to add test for such case so we can catch that in future? Doesn't have to be this PR.

rbans96:
> Just curious, is it possible to add test for such case so we can catch that in future? Doesn't have to be this PR.

Created #[77230](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Resolute/Stories/?workitem=77230) to address this.

PR #1374, opened 2020-10-20
Removes SqlServerFhirModel initialization redundancies

rotodd:
## Description

Note: this PR replaces #1262. I created new dev and feature branches that are up-to-date with master.

Description from #1262:

Currently (in my feature branch), the `Start()` method in `SqlServerFhirModel.cs` is being called on startup and every time the schema is upgraded.

This PR:
- Updates `SqlServerFhirModel` to no longer implement `IStartable` because it's initialization is now invoked on every schema upgrade
- Modularizes `SqlServerFhirModel`'s initialization code by schema version, allowing us to invoke schema version-specific initialization on schema upgrade.

## Related issues
Addresses [AB#75557](https://microsofthealth.visualstudio.com/Health/_workitems/edit/75557) and [#AB75558](https://microsofthealth.visualstudio.com/Health/_workitems/edit/75558).

## Testing
Local testing to check that the correct information is being passed when the schema is upgraded.

Followed the manual test scenarios that are outlined [here](https://microsoft-my.sharepoint.com/:x:/g/personal/rotodd_microsoft_com1/EZf3FpQrmZxHldZDZBjWS0UBK4SgbEFu9eZet1UwvyU67g?e=P3UGTN).

PR #1187, opened 2020-08-07
Registers SqlConnectionWrapperFactory service as a IScoped factory

rotodd:
## Description
A new SQL service is being added to the FHIR server to manage search parameter statuses (please see PR #1144). A change is required to the service registration of the `SqlConnectionWrapperFactory` in order for everything to successfully start up.

## Related issues
Addresses [#AB75037](https://microsofthealth.visualstudio.com/Health/_workitems/edit/75037).

## Testing
Manual testing (services start up without error).

PR #1430, opened 2020-11-07
Implements SQL search parameter status data layer

rotodd:
## Description
Merges in the feature branch for the SQL-side implementation of the search parameter status data store layer. 

This feature branch:

- Creates a new schema version and adds a SQL service to track search parameter status info in the SQL database (PR #1144)
- Adds the stored search parameter status info to the existing `dbo.SearchParam` table (PR #1169)
- Updates service registration to support the new SQL service (PR #1187)
- Removes "registry" from search parameter status class naming (PR #1189)
- Makes the columns added to `dbo.SearchParam` nullable (#1219)
- Handles schema upgrade notifications (#1233)
- Refactors how `SqlServerFhirModel` starts up to handle new dependencies and initialization requirements (#1374)

Some refactoring/setup has already been merged into master to enable these changes:

- `SqlServerModelInitializer` was removed from the healthcare-shared-components repo (PR [#50](https://github.com/microsoft/healthcare-shared-components/pull/50))
- `SchemaUpgradeRunner` and `SchemaJobWorker` were modified to fire notifications on schema upgrade (PR [#87](https://github.com/microsoft/healthcare-shared-components/pull/87) and PR [#92](https://github.com/microsoft/healthcare-shared-components/pull/92))
- `SqlServerFhirModel` was refactored to no longer rely on the `SqlServerModelInitializer` (PR #1072)

## Related issues
Addresses [#AB75420](https://microsofthealth.visualstudio.com/Health/_workitems/edit/75420).

## Testing

Adds integration tests for the new SQL service, `SqlServerSearchParameterStatusDataStore`.

Followed the manual test scenarios that are outlined [here](https://microsoft-my.sharepoint.com/:x:/g/personal/rotodd_microsoft_com1/EZf3FpQrmZxHldZDZBjWS0UBK4SgbEFu9eZet1UwvyU67g?e=P3UGTN).

## Future related work items
- Add test for incremental schema upgrades with data ([#AB75738](https://microsofthealth.visualstudio.com/Health/_workitems/edit/75738))
- Investigate starting up `SqlServerFhirModel` later than sever startup but before the first request ([#AB77009](https://microsofthealth.visualstudio.com/Health/_workitems/edit/77009))
- Refactor `SqlServerFhirModel` to use `SqlConnectionWrapperFactory` (this will be required for Managed Identity) ([#AB77229](https://microsofthealth.visualstudio.com/Health/_workitems/edit/77229))


brendankowitz (approved):

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SqlServerSearchParameterStatusDataStore.cs
```diff
@@ -0,0 +1,115 @@
...
+
+            using (IScoped<SqlConnectionWrapperFactory> scopedSqlConnectionWrapperFactory = _scopedSqlConnectionWrapperFactory())
+            using (SqlConnectionWrapper sqlConnectionWrapper = await scopedSqlConnectionWrapperFactory.Value.ObtainSqlConnectionWrapperAsync(CancellationToken.None, true))
+            using (SqlCommandWrapper sqlCommandWrapper = sqlConnectionWrapper.CreateSqlCommand())
+            {
+                VLatest.GetSearchParamStatuses.PopulateCommand(sqlCommandWrapper);
+
+                var parameterStatuses = new List<ResourceSearchParameterStatus>();
+
+                // TODO: Make cancellation token an input.
```
rbans96, line 67: Is ithis still a TODO? I see that in below statement we are passing CancellationToken

src/Microsoft.Health.Fhir.SqlServer/Resources.resx
```diff
@@ -126,4 +126,10 @@
   <data name="CyclicIncludeIterateNotSupported" xml:space="preserve">
     <value>Cyclic include iterate queries are not supported.</value>
   </data>
-</root>
\ No newline at end of file
+  <data name="SchemaVersionNeedsUpgrading" xml:space="preserve">
```
rbans96, line 129: nit: probably SchemaVersionNeedsUpgrading -> SchemaVersionNeedsUpgrade

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
```diff
@@ -48,238 +51,314 @@ public sealed class SqlServerFhirModel : IHostedService
...
-            var searchParamUriToId = new Dictionary<Uri, short>();
-            var systemToId = new ConcurrentDictionary<string, int>(StringComparer.OrdinalIgnoreCase);
-            var quantityCodeToId = new ConcurrentDictionary<string, int>(StringComparer.OrdinalIgnoreCase);
-            var claimNameToId = new Dictionary<string, byte>(StringComparer.Ordinal);
-            var compartmentTypeToId = new Dictionary<string, byte>();
-
-            // result set 1
-            while (await reader.ReadAsync(cancellationToken))
+            // If we are apply a full snap shot schema file, or if the server is just starting up
+            if (runAllInitialization || _highestInitializedVersion == 0)
```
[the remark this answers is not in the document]
rotodd: Ah, yes good question. I found this edge case recently while doing some testing locally. Here is the scenario:

1) Run the FHIR server on schema version 5 (I did this by setting `SchemaVersionConstants.Max = SchemaVersion.V5`)
2) Stop the FHIR server
3) Run the FHIR server with schema version 6 and automatic updates enabled

In this case, `runAllInitialization` will be `false` because we are applying a diff schema, but we want to call `InitializeBase()` to initialize the C# dictionaries. `InitializeBase()` won't be triggered on first request either because `_highestInitializedVersion` is already set to 6 at that point.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SqlServerSearchParameterStatusDataStore.cs
```diff
@@ -0,0 +1,115 @@
...
+            _scopedSqlConnectionWrapperFactory = scopedSqlConnectionWrapperFactory;
+            _updateSearchParamsTvpGenerator = updateSearchParamsTvpGenerator;
+            _filebasedSearchParameterStatusDataStore = filebasedRegistry.Invoke();
+            _schemaInformation = schemaInformation;
+        }
+
+        public async Task<IReadOnlyCollection<ResourceSearchParameterStatus>> GetSearchParameterStatuses()
+        {
+            // If the search parameter table in SQL does not yet contain status columns
+            if (_schemaInformation.Current < SchemaVersionConstants.SearchParameterStatusSchemaVersion)
```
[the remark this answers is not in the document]
rotodd: Good point, and I would agree if `SearchParameterStatusSchemaVersion` were only used it this file - but it is also used in `SqlServerFhirModel` to determine if `dbo.SearchParameter` should be populated with search parameter status information.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/6.diff.sql
```diff
@@ -0,0 +1,77 @@
...
+    Search Parameter Status Information
+**************************************************************/
+
+ALTER TABLE dbo.SearchParam
+ADD
+    Status varchar(10) NULL,
+    LastUpdated datetimeoffset(7) NULL,
+    IsPartiallySupported bit NULL
+
+CREATE TYPE dbo.SearchParamTableType_1 AS TABLE
```
Ivanidzo4ka, line 11: >_1 [](start = 36, length = 2)

is it normal have type with `_` in it?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/6.sql
```diff
@@ -0,0 +1,1944 @@
...
+
+/*************************************************************
+    Model tables
+**************************************************************/
+
+CREATE TABLE dbo.SearchParam
+(
+    SearchParamId smallint IDENTITY(1,1) NOT NULL,
+    Uri varchar(128) COLLATE Latin1_General_100_CS_AS NOT NULL,
+    Status varchar(10) NULL,
```
Ivanidzo4ka, line 21: >Status [](start = 4, length = 6)

Is it mapped to `SearchParameterStatus` ?
Wouldn't be better to map it to int instead of string?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SearchParameterStatusCollection.cs
```diff
@@ -0,0 +1,32 @@
...
+using Microsoft.Health.Fhir.Core.Features.Search.Registry;
+
+namespace Microsoft.Health.Fhir.SqlServer.Features.Storage.Registry
+{
+    public class SearchParameterStatusCollection : List<ResourceSearchParameterStatus>, IEnumerable<SqlDataRecord>
+    {
+        IEnumerator<SqlDataRecord> IEnumerable<SqlDataRecord>.GetEnumerator()
+        {
+            var sqlRow = new SqlDataRecord(
+                new SqlMetaData("Uri", SqlDbType.NVarChar, 128),
```
Ivanidzo4ka, line 18: >NVarChar [](start = 49, length = 8)

Isn't that are VarChar? At least it's varchar in 6.sql.
Same for Status.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SearchParameterStatusCollection.cs
```diff
@@ -0,0 +1,32 @@
...
+                new SqlMetaData("Status", SqlDbType.NVarChar, 10),
+                new SqlMetaData("IsPartiallySupported", SqlDbType.Bit));
```
Ivanidzo4ka, line 20: >IsPartiallySupported [](start = 33, length = 20)

How is `LastUpdated` field get updated?

src/Microsoft.Health.Fhir.SqlServer/Resources.resx
rotodd: Reworded to "SchemaVersionNeedsToBeUpgraded" in 229f90c.

src/Microsoft.Health.Fhir.SqlServer/Resources.resx
```diff
@@ -126,4 +126,10 @@
   <data name="CyclicIncludeIterateNotSupported" xml:space="preserve">
     <value>Cyclic include iterate queries are not supported.</value>
   </data>
-</root>
\ No newline at end of file
+  <data name="SchemaVersionNeedsToBeUpgraded" xml:space="preserve">
+    <value>Search parameter status information should not be null.</value>
```
Ivanidzo4ka, line 130: >Search parameter status information should not be null [](start = 11, length = 54)

Name and content is sound misalign for me.
How status is null for search param and Schema needs to be upgraded related to eachover?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
```diff
@@ -48,238 +51,314 @@ public sealed class SqlServerFhirModel : IHostedService
...
             {
-                (short id, string resourceTypeName) = reader.ReadRow(VLatest.ResourceType.ResourceTypeId, VLatest.ResourceType.Name);
+                // Run schema initialization required for [SchemaVersion.Min, SchemaVersion.Current]
```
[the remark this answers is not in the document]
rotodd: Haha yes fair point, math set notation does not belong in comments... wrote this out in 229f90c.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
```diff
@@ -48,238 +51,314 @@ public sealed class SqlServerFhirModel : IHostedService
...
-            var resourceTypeIdToTypeName = new Dictionary<short, string>();
-            var searchParamUriToId = new Dictionary<Uri, short>();
-            var systemToId = new ConcurrentDictionary<string, int>(StringComparer.OrdinalIgnoreCase);
-            var quantityCodeToId = new ConcurrentDictionary<string, int>(StringComparer.OrdinalIgnoreCase);
-            var claimNameToId = new Dictionary<string, byte>(StringComparer.Ordinal);
-            var compartmentTypeToId = new Dictionary<string, byte>();
-
-            // result set 1
-            while (await reader.ReadAsync(cancellationToken))
+            // If we are apply a full snap shot schema file, or if the server is just starting up
```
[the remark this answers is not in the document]
rotodd: Thanks, good catch. Fixed the grammar here in 229f90c.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/6.diff.sql
```diff
@@ -0,0 +1,77 @@
...
+
+/*************************************************************
+    Stored procedures for search parameter information
+**************************************************************/
+--
+-- STORED PROCEDURE
+--     GetSearchParamStatuses
+--
+-- DESCRIPTION
+--     Gets all the search parameters and their statuses
```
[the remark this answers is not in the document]
rotodd: Yes! Added periods in 8e7a49d and 5be0cef.

A bit more info: I was working off of `UpsertResource`, which seems to also be missing a period at the end of the sentence. I am hesitant to change the comment there though, since I think it will break the comment-sensitive schema upgrade tests. Perhaps going back and fixing `UpsertResource` in all schema files can be a change for another PR (..on a slow day haha).

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SqlServerSearchParameterStatusDataStore.cs
rotodd: Ah, I see how this is unclear. By this, I intended to indicate that a cancellation token should be an input to `GetSearchParameterStatuses()` and `UpsertStatuses()`. I moved the comment to be above those methods in 229f90c so that it's more clear.

I remember creating a user story for this at one point. I can't seem to find it anymore, so I created a new one: [#AB77269](https://microsofthealth.visualstudio.com/Health/_workitems/edit/77269)

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
```diff
@@ -48,238 +51,314 @@ public sealed class SqlServerFhirModel : IHostedService
...
+            ThrowIfCurrentSchemaVersionIsNull();
 
-            if (_searchParameterDefinitionManager is IHostedService hostedService)
+            // If the fhir-server is just starting up, synchronize the fhir-server dictionaries with the SQL database
+            Initialize((int)_schemaInformation.Current, true);
+
+            return Task.CompletedTask;
+        }
+
+        public void Initialize(int version, bool runAllInitialization)
```
Ivanidzo4ka, line 153: >Initialize [](start = 20, length = 10)

Just curious, how it's protected from been called multiple times from different threads?
Are we just claim it's not thread-safe and know it would never be called concurrently?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SqlServerSearchParameterStatusDataStore.cs
```diff
@@ -0,0 +1,116 @@
...
+                }
+
+                return parameterStatuses;
+            }
+        }
+
+        // TODO: Make cancellation token an input.
+        public async Task UpsertStatuses(IEnumerable<ResourceSearchParameterStatus> statuses)
+        {
+            EnsureArg.IsNotNull(statuses, nameof(statuses));
```
Ivanidzo4ka, line 99: >statuses [](start = 49, length = 8)

I would check is this collection empty or not, before doing all this SQL with it.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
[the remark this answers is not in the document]
rbans96: Makes sense!

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/6.sql
```diff
@@ -0,0 +1,1944 @@
...
+    SET Status = sps.Status, LastUpdated = @lastUpdated, IsPartiallySupported = sps.IsPartiallySupported
+    FROM dbo.SearchParam INNER JOIN @searchParams as sps
+    ON dbo.SearchParam.Uri = sps.Uri
+
+    INSERT INTO dbo.SearchParam
+        (Uri, Status, LastUpdated, IsPartiallySupported)
+    SELECT sps.Uri, sps.Status, @lastUpdated, sps.IsPartiallySupported
+    FROM @searchParams AS sps
+    WHERE sps.Uri NOT IN
+        (SELECT Uri FROM dbo.SearchParam)
```
Ivanidzo4ka, line 1941: Is there are reason to not use MERGE?
[https://docs.microsoft.com/en-us/sql/t-sql/statements/merge-transact-sql?view=sql-server-ver15#c-using-merge-to-do-update-and-insert-operations-on-a-target-table-by-using-a-derived-source-table](https://docs.microsoft.com/en-us/sql/t-sql/statements/merge-transact-sql?view=sql-server-ver15#c-using-merge-to-do-update-and-insert-operations-on-a-target-table-by-using-a-derived-source-table)

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/6.sql
```diff
@@ -0,0 +1,1944 @@
...
+-- STORED PROCEDURE
+--     GetSearchParamStatuses
+--
+-- DESCRIPTION
+--     Gets all the search parameters and their statuses.
+--
+-- RETURN VALUE
+--     The search parameters and their statuses.
+--
+CREATE PROCEDURE dbo.GetSearchParamStatuses
```
Ivanidzo4ka, line 1900: >GetSearchParamStatuses [](start = 21, length = 22)

I'm looking into other store procedures, and they at least have some other logic other than just plain select.
I'm curious, why it's a store procedure rather than just plain sql code?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SearchParameterStatusCollection.cs
rotodd: Good question. `LastUpdated` is set in the SQL schema code (see line 60 of `6.diff.sql`).

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/6.diff.sql
rotodd: We adopted this naming convention for table-valued parameters because they cannot be updated, deleted or inserted into (more info [here](https://docs.microsoft.com/en-us/sql/relational-databases/tables/use-table-valued-parameters-database-engine?view=sql-server-ver15#Restrictions)).

You aren't the first person to ask this question (see John's comment [here](https://github.com/microsoft/fhir-server/pull/1169#discussion_r465217000)) - maybe we should consider documenting this somewhere? @johnstairs

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
rotodd: Yes, I believe that we are claiming that this is not thread-safe. I'm not sure if we have plans to look into this - @brendankowitz do you know if this is something on the roadmap?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/6.sql
rotodd: Good point. I think this could go either way since the performance is the same (had to double check this and ended up reading through a post about it [here](https://docs.microsoft.com/en-us/previous-versions/dotnet/articles/ms973918(v=msdn.10))). I think it's nice to have this as a stored procedure because it makes the `SqlServerSearchParameterStatusDataStore` file a bit cleaner/more consistent. What do you think?

Ivanidzo4ka: I think amount lines of code in `SqlServerSearchParameterStatusDataStore` would be the same, it's just instead of populating command via generated code you directly specify command as string. But that would also reduce amount of code everywhere else.

If I was a DBA, I would probably advocate what everything need to be store procedure and triggers, but as developer I don't like idea of splitting business logic and pushing it into database layer. It can be often justified by performance reasons, but that's not the case.

Technically I would ask some of the elders in the project, what is common strategy for store procedures vs sql code. I don't have much background knowledge regarding data access layer to be right person to make a call, but I can express opinions :)

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SqlServerSearchParameterStatusDataStore.cs
rotodd: Good idea, added in 12919af.

src/Microsoft.Health.Fhir.SqlServer/Resources.resx
rotodd: Thanks for catching this. Fixed in 5a32607.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/6.sql
rotodd: Thanks for the suggestion! I'm currently investigating this a bit further.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SqlServerSearchParameterStatusDataStore.cs
```diff
@@ -0,0 +1,120 @@
...
+
+            if (!statuses.Any())
```
Ivanidzo4ka, line 100: >statuses.Any() [](start = 17, length = 14)

`IEnumerable` is a promise to provide something on request.
Every time it's requested it's has to evaluate itself. 
If `List<>` passed it wouldn't calculate anything, and just enumerate list, but if instead of `List<>` some real `IEnumerable` passed which actually does calculation or issue network requests, it has to perform it for `Any` and  below for `statuses.ToList()`

So I would either cache `IEnumerable` into `List` collection immediately and then check Count here , and pass that list to `Generate`, or changed signature of function to actually accept `List<>`.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/6.sql
rotodd: Nice, you make a good point. @brendankowitz thoughts?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/6.sql
rotodd: Good question, I agree that an `int` would be smaller and more efficient space-wise. I based this implementation off of the stored procedures that I wrote for SQL export operations, where status information is stored as a `string`.

I originally had export status info serializing to an `int`, but the point was made that we serialize status values as a `string` in Cosmos DB and it would be better to do that for SQL as well. [Here ](https://github.com/microsoft/fhir-server/pull/812#discussion_r368082167)is a link to the review comment about this -- let me know what you think.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SqlServerSearchParameterStatusDataStore.cs
rotodd: Ah, thank you for the explanation. I modified the method to accept a `List` instead of `IEnumberable` in 307a244.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SearchParameterStatusCollection.cs
rotodd: Thanks, fixed in 12d56d7.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/6.sql
Ivanidzo4ka: I'm not quite understand reasoning, to be honest, mostly because I'm not sure what is 'BE' mean in that comment.
Overall, personally in spherical horse in a vacuum situation, I would prefer to have ints, but 
a) it's not that important, there is not much savings from db side considering search params would be max 10k records or so
b) you have precedent on your side, and I like to keep things align, so it's better to keep code align than do micro optimizations.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/6.diff.sql
Ivanidzo4ka: Ok, this is part of accepted convention.
I still not sure I understand why this convention and not some other, but that's irrelevant to my question :)

---
In reply to: [520081613](https://github.com/microsoft/fhir-server/pull/1430#discussion_r520081613) [](ancestors = 520081613)

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SqlServerSearchParameterStatusDataStore.cs
```diff
@@ -0,0 +1,120 @@
...
+
+                using (SqlDataReader sqlDataReader = await sqlCommandWrapper.ExecuteReaderAsync(CommandBehavior.SequentialAccess, CancellationToken.None))
+                {
+                    while (await sqlDataReader.ReadAsync())
+                    {
+                        (string uri, string stringStatus, DateTimeOffset? lastUpdated, bool? isPartiallySupported) = sqlDataReader.ReadRow(
+                            VLatest.SearchParam.Uri,
+                            VLatest.SearchParam.Status,
+                            VLatest.SearchParam.LastUpdated,
+                            VLatest.SearchParam.IsPartiallySupported);
```
Ivanidzo4ka, line 75: From schema:
```
  Status varchar(10) NULL,
    LastUpdated datetimeoffset(7) NULL,
    IsPartiallySupported bit NULL
```
I think nullness need to be handled here, or maybe schema need to be changed.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SqlServerSearchParameterStatusDataStore.cs
```diff
@@ -0,0 +1,120 @@
...
+
+                        var status = Enum.Parse<SearchParameterStatus>(stringStatus, true);
+
+                        var resourceSearchParameterStatus = new ResourceSearchParameterStatus()
+                        {
+                            Uri = new Uri(uri),
+                            Status = status,
+                            IsPartiallySupported = (bool)isPartiallySupported,
```
Ivanidzo4ka, line 83: >(bool)isPartiallySupported [](start = 51, length = 26)

if isPartiallySupported is null, it would throw error. Same for DateTimeOffset.
Is expected?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SqlServerSearchParameterStatusDataStore.cs
```diff
@@ -0,0 +1,120 @@
...
+                {
+                    while (await sqlDataReader.ReadAsync())
+                    {
+                        (string uri, string stringStatus, DateTimeOffset? lastUpdated, bool? isPartiallySupported) = sqlDataReader.ReadRow(
+                            VLatest.SearchParam.Uri,
+                            VLatest.SearchParam.Status,
+                            VLatest.SearchParam.LastUpdated,
+                            VLatest.SearchParam.IsPartiallySupported);
+
+                        var status = Enum.Parse<SearchParameterStatus>(stringStatus, true);
```
Ivanidzo4ka, line 77: >stringStatus [](start = 71, length = 12)

Can be null, would throw exception.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/6.sql
rotodd: Updated the stored proc to use `MERGE` - thanks for the tip. Let me know how 9d6e015 looks.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SqlServerSearchParameterStatusDataStore.cs
rotodd: Ah, yes we have discussed this before [here](https://github.com/microsoft/fhir-server/pull/1219#discussion_r479576698).

> These columns are nullable because they are added to `dbo.SearchParam` in V4, after there is already existing data in the table.
> 
> They should be populated as soon as they are added to the table (in `SqlServerFhirModel.Start()`). If this check fails, we should be returning a 500 because something would be broken on our end that the user could not fix.

I do think it makes sense to explicitly throw a `NullReferenceException` to make the code more clear (we used to have this - I'm not sure when/why this explicit check got removed). Thanks for catching it! Added it back in in 14c333b.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SqlServerSearchParameterStatusDataStore.cs
rotodd: Fixed in 14c333b (please see above).

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/6.diff.sql
rotodd: Noticed @brendankowitz also brought this up [here](https://github.com/microsoft/fhir-server/pull/1191#discussion_r473159762). Would be good to have a short chat on the naming convention for table-valued params.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/6.sql
rotodd: Thanks for the reply - I agree with both your points a) and b).

Double checked with @jackliums, and by "BE" he just meant the FHIR service.

One thing I might add is if you look at the commit where I made the change to store enum values as `strings` ([link](https://github.com/microsoft/fhir-server/pull/812/commits/255ef52dc506b5e94a8fd95d21f8792de331940d)), someone reading the SQL schema file won't need to know that the `int` value `2` maps to the second `OperationStatus` enum value, `Running`.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/6.diff.sql
rotodd: Not sure this really needs a meeting, but I created a note to set up a Teams chat about this tomorrow morning.

rotodd: Added a comment about this in 577f673.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/6.sql
rotodd: One thing to note: the `GetExportJobById` is another stored proc that only contains a single `SELECT` statement

Ivanidzo4ka: Fair enough, precedent law is on your side 😄

Ivanidzo4ka (approved):
:shipit:

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SqlServerSearchParameterStatusDataStore.cs
```diff
@@ -0,0 +1,127 @@
...
+            if (_schemaInformation.Current < SchemaVersionConstants.SearchParameterStatusSchemaVersion)
+            {
+                throw new BadRequestException(Resources.SchemaVersionNeedsToBeUpgraded);
+            }
+
+            using (IScoped<SqlConnectionWrapperFactory> scopedSqlConnectionWrapperFactory = _scopedSqlConnectionWrapperFactory())
+            using (SqlConnectionWrapper sqlConnectionWrapper = await scopedSqlConnectionWrapperFactory.Value.ObtainSqlConnectionWrapperAsync(CancellationToken.None, true))
+            using (SqlCommandWrapper sqlCommandWrapper = sqlConnectionWrapper.CreateSqlCommand())
+            {
+                VLatest.UpsertSearchParams.PopulateCommand(sqlCommandWrapper, _updateSearchParamsTvpGenerator.Generate(statuses.ToList()));
```
Ivanidzo4ka, line 121: >statuses.ToList() [](start = 119, length = 17)

nit: It's a list already, no need to cast it again.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
```diff
@@ -48,238 +51,314 @@ public sealed class SqlServerFhirModel : IHostedService
...
+                        
+                        COMMIT TRANSACTION
+    
+                        -- result set 5
+                        SELECT Value, SystemId from dbo.System;
+
+                        -- result set 6
+                        SELECT Value, QuantityCodeId FROM dbo.QuantityCode";
+
+                    string searchParametersJson = JsonConvert.SerializeObject(_searchParameterDefinitionManager.AllSearchParameters.Select(p => new { Name = p.Name, Uri = p.Url }));
```
Ivanidzo4ka, line 237: > Name = p.Name, [](start = 149, length = 15)

What exactly the point of passing name if we extract only Uri?
```
 SELECT * FROM  OPENJSON (@searchParams)
                        WITH (Uri varchar(128) '$.Uri')
```

I know it's not your code, but if it can be cleaned, why not

Ivanidzo4ka (approved):
:shipit:

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SqlServerSearchParameterStatusDataStore.cs
rotodd: Good catch, thanks. Fixed in 6bb60d6.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
rotodd: Fixed in dffb0c6.

PR #1439, opened 2020-11-11
Avoid returning empty search bundles with next links

johnstairs:
There are several situations in which Cosmos DB will return an empty page of results with a continuation token. It could be because the query had to halted because of RU exhaustion, because the query had not completed after 5 seconds, or because the SDK executed the query on only one of several physical partitions. 
Up until now, we simply returned the first page of results that the SDK gave us, but customers have rightly complained that this is both surprising behavior and suboptimal (when would you want an empty page with a next link?) .

In this PR, we attempt to fill up a bundle with at least half of the the items requested. We give up after 30s or if we get a 429, in which case we return whatever we had accumulated. If the `_count` parameter is specified, the bundle will not have more items than specified. If omitted, bundles may have up to 14 entries, up from the usual default of 10 if the first page is filled up.

At the moment, we are not executing cross-partition queries in parallel.


src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
```diff
@@ -151,6 +151,8 @@ public SearchOptions Create(string compartmentType, string compartmentId, string
             // Check the item count.
             if (searchParams.Count != null)
             {
+                searchOptions.MaxItemCountSpecifiedByClient = true;
```
LTA-Thinking, line 154: Should this only be true if we use the value for Count?

LTA-Thinking:
Nit: In the PR description you say the max number returned with the default page size is 19, but I think it would be 14. First page has 4 results, which is less than half of the requested 10, then the second page has 10 results.

src/Microsoft.Health.Fhir.CosmosDb.UnitTests/Features/Storage/CosmosFhirDataStoreTests.cs
```diff
@@ -0,0 +1,204 @@
...
+                new QueryDefinition("abc"),
+                new QueryRequestOptions { MaxItemCount = 10 },
+                mustNotExceedMaxItemCount: false);
+
+            Assert.Equal(Enumerable.Range(0, 2), results);
+            Assert.Null(continuationToken);
+        }
+
+        [Fact]
+        public async Task GivenAQueryWhereItemCountMustNotBeExceeded_WhenExecuted_FetchesSubsequentPagesByIssuingNewQueries()
```
LTA-Thinking, line 130: I don't think this test shows the desired behavior. The second half of the first CreateResponses is the same as the second CreateResponses. It looks like this test would pass whether it was calling the new query or the old one.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
johnstairs: It should be true iff `_count` is specified in the search.

johnstairs:
> Nit: In the PR description you say the max number returned with the default page size is 19, but I think it would be 14. First page has 4 results, which is less than half of the requested 10, then the second page has 10 results.

Good point! Will correct.

src/Microsoft.Health.Fhir.CosmosDb.UnitTests/Features/Storage/CosmosFhirDataStoreTests.cs
johnstairs: Good catch. I've updated the test.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
LTA-Thinking: Sorry I was unclear. Below this we ignore the _count if it is higher than the MaxItemCountPerSearch. Should this still be true if we ignore the _count?

johnstairs: Oh I see what you mean. When this is true, we treat the MaxItemCount as a hard limit. When _count is higher than MaxItemCountPerSearch and we set it to MaxItemCountPerSearch, I think we should treat it as a hard limit.

brendankowitz (approved):

PR #1444, opened 2020-11-12
Updates package 'Microsoft.Azure.Cosmos.Direct' to version '3.14.2'

MicrosoftHealthService:


johnstairs:
Upgrading this causes problems. We should be able to remove this reference in the future.

PR #1445, opened 2020-11-12
Fix Cosmos DB error handling

johnstairs:
## Description
Exceptions that we throw inside of a `RequestHandler` are wrapped by the Cosmos DB SDK in a `CosmosException` with status code 500. This means that our error handling for 429s and other conditions where we throw a specific exception have resulted in 500s since we moved to v3 of the Cosmos DB SDK.

The change here is to recognize that the `RequestRateExceededException` and others we throw from within the `RequestHandler` will appear as an inner exception.

## Related issues
Addresses [AB#76850](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/76850)

## Testing
Added unit tests and performed manual testing

### DIFF
--- a/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Reindex/ReindexHandlerTests.cs
+++ b/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Reindex/ReindexHandlerTests.cs
@@ -87,11 +87,12 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Operations.Reindex
 
             var jobRecord = new ReindexJobRecord(_resourceTypeSearchParameterHashMap, 1);
             var jobWrapper = new ReindexJobWrapper(jobRecord, WeakETag.FromVersionId("id"));
-            _fhirOperationDataStore.GetReindexJobByIdAsync("id", CancellationToken.None).Throws(new RequestRateExceededException(TimeSpan.FromMilliseconds(100)));
+            _fhirOperationDataStore.GetReindexJobByIdAsync("id", CancellationToken.None).Throws(new Exception(null, new RequestRateExceededException(TimeSpan.FromMilliseconds(100))));
 
             var handler = new GetReindexRequestHandler(_fhirOperationDataStore, DisabledFhirAuthorizationService.Instance);
 
-            await Assert.ThrowsAsync<RequestRateExceededException>(() => handler.Handle(request, CancellationToken.None));
+            Exception thrownException = await Assert.ThrowsAsync<Exception>(() => handler.Handle(request, CancellationToken.None));
+            Assert.IsType<RequestRateExceededException>(thrownException.InnerException);
         }
 
         [Fact]
--- a/src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Operations/CosmosFhirOperationDataStore.cs
+++ b/src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Operations/CosmosFhirOperationDataStore.cs
@@ -16,6 +16,7 @@ using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Microsoft.Health.Abstractions.Exceptions;
 using Microsoft.Health.Extensions.DependencyInjection;
+using Microsoft.Health.Fhir.Core.Exceptions;
 using Microsoft.Health.Fhir.Core.Features.Operations;
 using Microsoft.Health.Fhir.Core.Features.Operations.Export.Models;
 using Microsoft.Health.Fhir.Core.Features.Operations.Reindex.Models;
@@ -101,9 +102,9 @@ namespace Microsoft.Health.Fhir.CosmosDb.Features.Storage.Operations
             }
             catch (CosmosException dce)
             {
-                if (dce.StatusCode == HttpStatusCode.TooManyRequests)
+                if (dce.InnerException is RequestRateExceededException)
                 {
-                    throw new RequestRateExceededException(dce.RetryAfter);
+                    throw;
                 }
 
                 _logger.LogError(dce, "Failed to create an export job.");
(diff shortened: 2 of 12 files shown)


src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Filters/OperationOutcomeExceptionFilterTests.cs
```diff
@@ -217,6 +217,40 @@ public void GivenATransactionFailedExceptionException_WhenExecutingAnAction_Then
...
+            Assert.False(_context.ExceptionHandled);
+            Assert.Null(_context.Result);
+            Assert.Same(exception, _context.Exception); // ensure state is restored
+        }
+
+        [Fact]
+        public void GivenARequestRateExceededExceptionAsAnInnerException_WhenExecutingAnAction_ThenTheResponseShouldBeAnOperationOutcome()
+        {
+            ValidateOperationOutcome(new Exception(null, new RequestRateExceededException(null)), HttpStatusCode.TooManyRequests);
+        }
```
LTA-Thinking, line 252: Is it worth having a test to show that the inner exception is only used if the outer exception couldn't be parsed?

johnstairs: Makes sense.

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Reindex/ReindexHandlerTests.cs
```diff
@@ -87,11 +87,12 @@ public async Task GivenAGetRequest_WhenTooManyRequestsThrown_ThenTooManyRequests
 
             var jobRecord = new ReindexJobRecord(_resourceTypeSearchParameterHashMap, 1);
             var jobWrapper = new ReindexJobWrapper(jobRecord, WeakETag.FromVersionId("id"));
-            _fhirOperationDataStore.GetReindexJobByIdAsync("id", CancellationToken.None).Throws(new RequestRateExceededException(TimeSpan.FromMilliseconds(100)));
+            _fhirOperationDataStore.GetReindexJobByIdAsync("id", CancellationToken.None).Throws(new Exception(null, new RequestRateExceededException(TimeSpan.FromMilliseconds(100))));
```
Ivanidzo4ka, line 90: >new Exception(null [](start = 96, length = 18)

I find this route is a bit counterintuitive.
Why can't we peel cosmos exception and pop to the top it's inner exception rather than bend rest of code for it?

johnstairs: The issue is that there are many entrypoints to the cosmos SDK (basically the methods on `Container`), and therefore many places where we would have to make sure the exception is mapped properly. Prior to adopting the v3 SDK, we used to have a generated wrapper to the SDK that could forward calls to the and would do common pre/post processing. We could do something similar again.

Ivanidzo4ka: Ok, it's probably fine as long as it's not leaking into non cosmos specific code.

namadabu (approved):

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Filters/OperationOutcomeExceptionFilterTests.cs
johnstairs: There's actually no recognized exception that has a constructor that takes an inner exception, so we currently can't get get into the situation to test (without reflection)

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Operations/CosmosFhirOperationDataStore.cs
```diff
@@ -169,9 +171,9 @@ public async Task<ExportJobOutcome> GetExportJobByHashAsync(string hash, Cancell
             }
             catch (CosmosException dce)
             {
-                if (dce.StatusCode == HttpStatusCode.TooManyRequests)
+                if (dce.InnerException is RequestRateExceededException)
```
brendankowitz, line 174: Perhaps we can smooth this over a little with some extensions...
`bool dce.IsRateExceededException()`?
Or similar, I guess the point might be to unwrap inside an extension, keeps the logic consistent everywhere

brendankowitz (approved):

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Operations/CosmosFhirOperationDataStore.cs
johnstairs: There is something about accepting this to be the inner exception that just doesn't feel right. And I don't know that adding an extension method helps that much (why should write `catch (Exception e) when (e.IsRequestRateExceededException())` instead of `catch (RequestRateExceededException)`?). I feel like this fix is a stop-gap, but I don't find it very satisfying.

johnstairs: Should another data provider throw RequestRateExceededException as an inner exception too? Seems like the answer should be no.

johnstairs: ...But I've added the extension methods :)

brendankowitz: Agree, it seems like the simplest fix, yet unsatisfying. Though if it improves the experience we have time to find a nicer solution

PR #1449, opened 2020-11-13
Add support to POST a new SearchParameter

feordin:
## Description
When a SearchParameter resource is POST'ed to the server it will be validated and added to the list of "supported" search parameters, and available to be applied during reindexing.

## Related issues
Addresses [issue [AB#76345](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/76345)].

## Testing
Unit tests and E2E tests added.  Manual testing done as well.

## FHIR Team Checklist

When Custom Search becomes fully available, will tag with feature, but for now no semver updates as this is only the initial commit for Custom Search.


src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Reindex/ReindexUtilities.cs
```diff
@@ -91,30 +91,14 @@ public async Task ProcessSearchResultsAsync(SearchResult results, IReadOnlyDicti
...
-                {
-                    LastUpdated = DateTimeOffset.UtcNow,
-                    Status = SearchParameterStatus.Enabled,
-                    Uri = searchParamUri,
-                });
+                await _searchParameterStatusManager.UpdateSearchParameterStatus(searchParameterUris, SearchParameterStatus.Enabled);
+            }
+            catch (SearchParameterNotSupportedException)
+            {
+                return (false, string.Format(Core.Resources.SearchParameterNoLongerSupported));
```
Ivanidzo4ka, line 100: >SearchParameterNoLongerSupported [](start = 60, length = 32)

Can we take name from SearchParameterNotSupportedException and pass it to error message? It can be confusing from user perspective to do binary search on list of params to find out which one of them is not working.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Parameters/SearchParameterValidator.cs
```diff
@@ -0,0 +1,120 @@
...
+            {
+                throw new UnauthorizedFhirActionException();
+            }
+
+            // check if reindex job is running
+            using (IScoped<IFhirOperationDataStore> fhirOperationDataStore = _fhirOperationDataStoreFactory())
+            {
+                if (await fhirOperationDataStore.Value.CheckActiveReindexJobsAsync(cancellationToken))
+                {
+                    throw new JobConflictException(Resources.ChangesToSearchParametersNotAllowedWhileReindexing);
```
Ivanidzo4ka, line 62: Can we pass address or id of reindex job to user so user can terminate or monitor that job?

src/Microsoft.Health.Fhir.Core/Features/Definition/ISearchParameterDefinitionManager.cs
```diff
@@ -70,17 +71,17 @@ public interface ISearchParameterDefinitionManager
...
         /// Gets the hash of the current search parameters that are supported for the given resource type.
         /// </summary>
         /// <param name="resourceType">Resource type for which we need the hash of search parameters.</param>
         /// <returns>A string representing a hash of the search parameters.</returns>
         public string GetSearchParameterHashForResourceType(string resourceType);
+
+        /// <summary>
+        /// Allows addition of a new search parameters at runtime
+        /// </summary>
+        /// <param name="searchParamBundle">A bundle containing SearchParameter resources</param>
```
Ivanidzo4ka, line 84: super nit: `.` in the end of sentences.

src/Microsoft.Health.Fhir.Core/Features/Definition/ISearchParameterDefinitionManager.cs
```diff
@@ -70,17 +71,17 @@ public interface ISearchParameterDefinitionManager
...
+        public void AddNewSearchParameters(BundleWrapper searchParamBundle);
```
Ivanidzo4ka, line 85: >BundleWrapper searchParamBundle [](start = 43, length = 31)

I'm a bit puzzled by fact it's a bundle.
Is there are reason why it's not just `List<SearchParameterInfo>` ?

src/Microsoft.Health.Fhir.Core/Features/Search/SearchHelperUtilities.cs
```diff
@@ -8,30 +8,45 @@
...
-            foreach (ResourceSearchParameterStatus searchParameterStatus in resourceSearchParameterStatus.OrderBy(x => x.Uri.ToString()))
+            foreach (SearchParameterInfo searchParamInfo in searchParamaterInfos.OrderBy(x => x.Url.ToString()))
             {
-                sb.Append(searchParameterStatus.Uri.ToString());
-                sb.Append(searchParameterStatus.LastUpdated.ToString());
+                sb.Append(searchParamInfo.Url.ToString());
+                sb.Append(searchParamInfo.Type.ToString());
+                sb.Append(searchParamInfo.Expression);
+
+                if (searchParamInfo.TargetResourceTypes != null &&
```
Ivanidzo4ka, line 39: It's a bit strange what Equals and GetHashCode in SearchParameterInfo  don't use `TargetResourceTypes` and `BaseResourceTypes` but this method does.

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/SearchParameterFilterAttribute.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            EnsureArg.IsNotNull(context, nameof(context));
+
+            if (context.ActionArguments.TryGetValue(KnownActionParameterNames.Resource, out var parsedModel))
+            {
+                if (parsedModel is SearchParameter)
+                {
+                    // wait for the validation checks to pass before allowing the FHIRController action to continue
+                    await _searchParameterValidator.ValidateSearchParamterInput(
+                        parsedModel as SearchParameter,
```
Ivanidzo4ka, line 42: >parsedModel as SearchParameter [](start = 24, length = 30)

nit:
you can do instead of 
```
if (parsedModel is SearchParameter)
```
this:
```
if (parsedModel is SearchParameter searchParameter)
```
and use `searchParameter` variable instead of `parsedModel as SearchParameter` here and few lines below.

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/SearchParameterFilterAttribute.cs
```diff
@@ -0,0 +1,67 @@
...
+
+                    // wait for the Action to execute
+                    await next();
+
+                    if (HttpMethods.IsPost(context.HttpContext.Request.Method))
+                    {
+                        // Once the SearchParameter resource is committed to the data store, we can update the in
+                        // memory SearchParameterDefinitionManager, and persist the status to the data store
+                        await _searchParameterEditor.AddSearchParameterAsync(parsedModel as SearchParameter, context.HttpContext.RequestAborted);
+                    }
```
Ivanidzo4ka, line 54: I'm not sure this is right place to do that.
Haven't went through whole codereview, so I don't know workflow yet, but I assume Put/Delete should do something with in-memory collection of search parameters as well, and if it's so, why POST have special treatment here.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Parameters/SearchParameterEditor.cs
```diff
@@ -0,0 +1,48 @@
...
+using Hl7.Fhir.ElementModel;
+using Hl7.Fhir.Model;
+using Microsoft.Health.Fhir.Core.Extensions;
+using Microsoft.Health.Fhir.Core.Features.Definition;
+using Microsoft.Health.Fhir.Core.Features.Definition.BundleWrappers;
+using Microsoft.Health.Fhir.Core.Features.Search.Registry;
+
+namespace Microsoft.Health.Fhir.Shared.Core.Features.Search.Parameters
+{
+    public class SearchParameterEditor : ISearchParameterEditor
```
Ivanidzo4ka, line 17: >Editor [](start = 32, length = 6)

DefinitionManager, DefinitionBuilder and Editor walk into the bar...

So there is StatusManager as well which looks like contains reference on DefinitionManager.

I can understand model where you have layer to access DB (DataStore)
There should be collection which contains all current search params and push updates to DB. (is it StatusManager? Or DefinitionManager?)

Builder I guess is helper class which builds that collection.

I'm sure there is reason for all this classes to exist, it just it's not clear from names and documentation aren't great. And that raises question, what this class does, and why it's not StatusManager?

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/SearchParameterFilterAttribute.cs
feordin: You are correct, PUT/DELETE will also act in this same place.  This PR is only for the POST and I haven't done the PUT/DELETE yet.  Plan is to have a check for each HttpMethod.  Does it make more sense then?

src/Microsoft.Health.Fhir.Core/Features/Definition/ISearchParameterDefinitionManager.cs
feordin: When we read in the static file of the search parameters from the spec, they come in as a bundle.  Also, when we query for SearchParameter resources from the datastore, they should come as a bundle.  We could change it, read the bundle and get the list of entries.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Parameters/SearchParameterEditor.cs
feordin: Good question :).  StatusManager, as well as DefinitionManager are in Core, while SearchParameterEditor is in Shared.Core.  Shared.Core is dependent on a specific version of FHIR and it's main job is grab the version specifc SearchParameter resource and put it into a bundle which is then wrapped in the version agnostic bundleWrapper.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Parameters/SearchParameterValidator.cs
```diff
@@ -0,0 +1,120 @@
...
+                    }
+                }
+                catch (SearchParameterNotSupportedException)
+                {
+                    // if thrown, then the search parameter is not found
+                    // if a PUT, then we should be updating an exsting paramter, but it was not found
+                    if (method.Equals(HttpPutName, StringComparison.OrdinalIgnoreCase) ||
+                        method.Equals(HttpDeleteName, StringComparison.OrdinalIgnoreCase))
+                    {
+                        // if no exception is thrown, then the search parameter with the same Uri was found
```
namadabu, line 98: nit: comment needs to be updated.

src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterDefinitionManager.cs
```diff
@@ -48,14 +50,23 @@ public SearchParameterDefinitionManager(IModelInfoProvider modelInfoProvider)
...
 
-            builder.Build();
+                List<string> list = UrlLookup.Values.Where(p => p.Type == ValueSets.SearchParamType.Composite).Select(p => string.Join("|", p.Component.Select(c => UrlLookup[c.DefinitionUrl].Type))).Distinct().ToList();
 
-            _typeLookup = builder.ResourceTypeDictionary;
-            UrlLookup = builder.UriDictionary;
+                CalculateSearchParameterHashAsync();
+
+                _started = true;
+            }
```
rotodd, line 69: I think some old changes that have been removed may have sneaked back in here. I don't think we need the `_started` flag or the url lookup values list anymore.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Reindex/ReindexUtilities.cs
feordin: Updated.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Parameters/SearchParameterValidator.cs
feordin: Updated.

src/Microsoft.Health.Fhir.Core/Features/Definition/ISearchParameterDefinitionManager.cs
feordin: Updated.

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/SearchParameterFilterAttribute.cs
feordin: Updated.

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/SearchParameterFilterAttribute.cs
```diff
@@ -0,0 +1,67 @@
...
+                        context.HttpContext.RequestAborted);
+
+                    // wait for the Action to execute
+                    await next();
+
+                    if (HttpMethods.IsPost(context.HttpContext.Request.Method))
+                    {
+                        // Once the SearchParameter resource is committed to the data store, we can update the in
+                        // memory SearchParameterDefinitionManager, and persist the status to the data store
+                        await _searchParameterEditor.AddSearchParameterAsync(parsedModel as SearchParameter, context.HttpContext.RequestAborted);
```
Ivanidzo4ka, line 53: >parsedModel as SearchParameter [](start = 77, length = 30)

I think you should be able to use `searchParameter` variable here instead of casting.

src/Microsoft.Health.Fhir.Core/Features/Definition/ISearchParameterDefinitionManager.cs
Ivanidzo4ka: I'm just looking on content of this file, and it's called `SearchParameter`DefinitionManager and in all other methods it operates and produce `SearchParameterInfo` (with exception of Dictionary<string,string>, string and SearchParameterType).
If I need to perform singular Get I will get back `SearchParameterInfo` if I do all , I get back list of `SearchParameterInfo`

If I look into code which uses this method  we also wrap single resource into bundle, so later it can be unpacked which I find a bit convoluted.

I think it would be nice to have separation between how things passed to us and how we operate on them.  It's better to operate on one entity and have converters before we pass information, than support multiple things at once. 

---
In reply to: [532759584](https://github.com/microsoft/fhir-server/pull/1449#discussion_r532759584) [](ancestors = 532759584)

src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterStatusManager.cs
```diff
@@ -88,63 +88,43 @@ public async Task EnsureInitialized()
...
+            await _searchParameterStatusDataStore.UpsertStatuses(searchParameterStatusList);
 
             await _mediator.Publish(new SearchParametersUpdated(updated));
         }
+
+        public async Task AddSearchParameterStatus(string searchParamUri)
+        {
+            // new search parameters are added as supported, until reindexing occurs, when
+            // they will be fully enabled
+            await UpdateSearchParameterStatus(new List<string>() { searchParamUri }, SearchParameterStatus.Supported);
```
rotodd, line 127: Silly question: would there ever be a time that a user would want to add a search parameter that is not supported?

feordin: I think there could be a case when you might want to.  Perhaps you are adding a profile or implementation guide, such as US Core, and it just happens that we don't support one of the parameters.  You may not want the entire profile to be rejected because of one unsupported parameter, and perhaps you want the parameter stored for reference, even if you can't search on it.  However, for the moment, to get to an MVP of custom search parameters, we are going to reject any unsupported parameters.

src/Microsoft.Health.Fhir.Core/Features/Search/SearchHelperUtilities.cs
feordin: It is a good point.  I am not sure I want to add the target or base types to the Equals method or the Hash calculation as the Uri must be unique in any case.  For the overall calculation here, it is a hash per resource type, so perhaps it makes a bit more sense to include it.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Parameters/SearchParameterValidator.cs
feordin: updated.

src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterDefinitionManager.cs
feordin: Removed _started flag and unnecessary list.

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/SearchParameterFilterAttribute.cs
feordin: Updated

src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterStatusManager.cs
rotodd: Makes sense, thanks for the info!

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlServerFhirStorageTestHelper.cs
```diff
@@ -50,6 +50,7 @@ public class SqlServerFhirStorageTestHelper : IFhirStorageTestHelper, ISqlServer
             _searchParameterDefinitionManager = searchParameterDefinitionManager;
             _sqlServerFhirModel = sqlServerFhirModel;
             _sqlConnectionFactory = sqlConnectionFactory;
+            _searchParameterDefinitionManager.StartAsync(CancellationToken.None).Wait();
```
rotodd, line 53: +1 nice work around moving this to the constructor

rotodd, line 53: Nit: I think this is only used in one place now - can we remove the private variable?

Ivanidzo4ka: I think it's actually somewhat antipattern to switch from async/await to Wait method.
Can be wrong but Wait() has blocking behavior while await is not.

---
In reply to: [533650616](https://github.com/microsoft/fhir-server/pull/1449#discussion_r533650616) [](ancestors = 533650616)

feordin: In generate, this is definitely true.  In this case downstream test resources need this method to finish, and I don't think the blocking behavior will negatively affect the test execution.

src/Microsoft.Health.Fhir.Core/Features/Definition/ISearchParameterDefinitionManager.cs
Ivanidzo4ka: As discussed, reason behind this method accepting bundle, is because this is shared among 3 different specifications code and there is no way to pick right SearchParam. SearchParamInfo is something we use internally to represent SearchParam and it has limited coverage compared to any implmentation of SearchParam so we can't use it either.

---
In reply to: [532826560](https://github.com/microsoft/fhir-server/pull/1449#discussion_r532826560) [](ancestors = 532826560,532759584)

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/SearchParameterFilterAttribute.cs
feordin: After more discussion, @Ivanidzo4ka pointed out that the filter is perhaps not the best place to invoke the add logic.  I have moved it to the Handler.  Refactoring some code along the way.

src/Microsoft.Health.Fhir.Core/Features/Definition/ISearchParameterDefinitionManager.cs
feordin: After refactoring to move the call to addsearchparameters out of the filter, it was easier to also move away from using Bundle and switch to ITypedElement.

src/Microsoft.Health.Fhir.Core/Resources.resx
```diff
@@ -372,22 +375,28 @@
...
   </data>
   <data name="SearchParameterDefinitionInvalidExpression" xml:space="preserve">
-    <value>bundle.entry[{0}].resource.expression is null or empty.</value>
+    <value>SearchParameter[{0}].resource.expression is null or empty.</value>
+    <comment>{0} Is the Uri Identifying the SearchParameter</comment>
   </data>
   <data name="SearchParameterDefinitionInvalidResource" xml:space="preserve">
-    <value>bundle.entry[{0}].resource is not a SearchParameter resource.</value>
+    <value>Entry {0} is not a SearchParameter resource.</value>
+    <comment>{0} Index of the processed element in the ennumeration</comment>
```
Ivanidzo4ka, line 399: >nn [](start = 56, length = 2)

super nit: `enumeration` (above in line 391 as well)

src/Microsoft.Health.Fhir.Core/Resources.resx
```diff
@@ -446,13 +455,23 @@
     <value>WeakETag must be in the weak ETag format.</value>
   </data>
   <data name="SearchParameterNoLongerSupported" xml:space="preserve">
-    <value>Search parameter {0} is no longer supported and cannot be marked enabled.</value>
-    <comment>{0}; the name of the search parameter which is no longer supported</comment>
+    <value>{0} So it cannot be marked enabled.</value>
```
Ivanidzo4ka, line 458: >{0} So it cannot be marked enabled. [](start = 11, length = 35)

nit: maybe `{0}, so it cannot ...`

src/Microsoft.Health.Fhir.Core/Resources.resx
```diff
@@ -446,13 +455,23 @@
...
   </data>
   <data name="ReindexJobInCompletedState" xml:space="preserve">
     <value>Reindex job id {0} is in state {1} and cannot be cancelled.</value>
     <comment>{0} is the id of the reindex which is a a GUID.  {1} is the state, an enumerated set of state values.</comment>
   </data>
+  <data name="SearchParameterDefinitionInvalidMissingUri" xml:space="preserve">
+    <value>The search parameter defintion is missing the Uri.</value>
+  </data>
+  <data name="SearchParameterDefinitionNotFound" xml:space="preserve">
+    <value>A search parameter with Uri '{0}' was not found.</value>
```
Ivanidzo4ka, line 469: >{0} [](start = 40, length = 3)

nit: <comment> section?

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Parameters/SearchParameterValidator.cs
```diff
@@ -0,0 +1,121 @@
...
+                                nameof(searchParam.Url),
+                                string.Format(Resources.SearchParameterDefinitionNotFound, searchParam.Url)));
+                    }
+                }
+            }
+
+            // validate that the url does not correspond to a search param from the spec
+            // TODO: still need a method to determine spec defined search params
+
+            // validation of the fhir path
```
johnstairs, line 112: For SQL, we'll also need additional validation for composite parameters, ensuring that it's one of the existing type combinations.

src/Microsoft.Health.Fhir.Core/Features/Definition/ISearchParameterDefinitionManager.cs
```diff
@@ -70,17 +71,17 @@ public interface ISearchParameterDefinitionManager
...
         /// Gets the hash of the current search parameters that are supported for the given resource type.
         /// </summary>
         /// <param name="resourceType">Resource type for which we need the hash of search parameters.</param>
         /// <returns>A string representing a hash of the search parameters.</returns>
         public string GetSearchParameterHashForResourceType(string resourceType);
+
+        /// <summary>
+        /// Allows addition of a new search parameters at runtime.
+        /// </summary>
+        /// <param name="searchParameters">An ennumeration containing SearchParameter resources.</param>
```
Ivanidzo4ka, line 84: >nn [](start = 47, length = 2)

nit: `enumeration`

src/Microsoft.Health.Fhir.Core/Features/Search/SearchHelperUtilities.cs
```diff
@@ -8,30 +8,45 @@
...
+                    searchParamInfo.TargetResourceTypes.Any())
+                {
+                    sb.Append(string.Join(null, searchParamInfo.TargetResourceTypes.OrderBy(s => s)));
+                }
+
+                if (searchParamInfo.BaseResourceTypes != null &&
+                    searchParamInfo.BaseResourceTypes.Any())
+                {
+                    sb.Append(string.Join(null, searchParamInfo.BaseResourceTypes.OrderBy(s => s)));
```
johnstairs, line 48: Did we consider not allocating for the hash code? Using [`HashCode`](https://docs.microsoft.com/en-us/dotnet/api/system.hashcode.combine?view=dotnet-plat-ext-5.0)'s methods perhaps?

src/Microsoft.Health.Fhir.Core/Features/Definition/ISearchParameterDefinitionManager.cs
```diff
@@ -70,17 +71,17 @@ public interface ISearchParameterDefinitionManager
...
+        void AddNewSearchParameters(IList<ITypedElement> searchParameters);
```
Ivanidzo4ka, line 85: >IList [](start = 36, length = 5)

nit: `IList` has `Insert` and `RemoveAt` methods which allow to change list content. I would change it to IEnumerable, since maybe point of this collection is to pass something which can be enumerated but not specifically modifed.

src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterDefinitionBuilder.cs
```diff
@@ -48,119 +38,90 @@ internal class SearchParameterDefinitionBuilder
...
             {
-                if (_resourceTypeDictionary.TryGetValue(resourceType, out IDictionary<string, SearchParameterInfo> _))
-                {
-                    // The list has already been built, move on.
-                    continue;
-                }
-
                 // Recursively build the search parameter definitions. For example,
                 // Appointment inherits from DomainResource, which inherits from Resource
                 // and therefore Appointment should include all search parameters DomainResource and Resource supports.
```
Ivanidzo4ka, line 65: nit: can we move this description to function summary/remarks? I find myself head scratching on recursive call inside this function, so this comment is really helpful, I just had to scroll all the way up to find it.

src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterDefinitionManager.cs
```diff
@@ -33,11 +32,13 @@ public SearchParameterDefinitionManager(IModelInfoProvider modelInfoProvider)
...
             _modelInfoProvider = modelInfoProvider;
             _resourceTypeSearchParameterHashMap = new ConcurrentDictionary<string, string>();
+            TypeLookup = new Dictionary<string, IDictionary<string, SearchParameterInfo>>();
+            UrlLookup = new Dictionary<Uri, SearchParameterInfo>();
         }
 
         internal IDictionary<Uri, SearchParameterInfo> UrlLookup { get; set; }
 
-        internal IDictionary<string, IDictionary<string, SearchParameterInfo>> TypeLookup { get => _typeLookup; }
+        internal IDictionary<string, IDictionary<string, SearchParameterInfo>> TypeLookup { get; }
```
Ivanidzo4ka, line 41: >TypeLookup [](start = 79, length = 10)

nit: can we add comment what are the keys in inner and current dictionary?

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Parameters/SearchParameterValidator.cs
feordin: New work item created to track this validation need (78059).

src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterDefinitionManager.cs
```diff
@@ -138,5 +127,28 @@ public string GetSearchParameterHashForResourceType(string resourceType)
...
+            SearchParameterDefinitionBuilder.Build(
+                searchParameters,
+                UrlLookup,
+                TypeLookup,
+                _modelInfoProvider);
+
+            CalculateSearchParameterHash();
+        }
+
+        private void CalculateSearchParameterHash()
```
Ivanidzo4ka, line 142: >CalculateSearchParameterHash [](start = 21, length = 28)

I would add filter to this function. Calculating all hashes during start up is right thing to do, but recalculate them all again if only we update one search parameter is sound a bit overkill.

src/Microsoft.Health.Fhir.Core/Features/Definition/BundleWrappers/BundleEntryWrapper.cs
```diff
@@ -11,7 +11,7 @@
 
 namespace Microsoft.Health.Fhir.Core.Features.Definition.BundleWrappers
 {
-    internal class BundleEntryWrapper
+    public class BundleEntryWrapper
```
Ivanidzo4ka, line 14: >BundleEntryWrapper [](start = 17, length = 18)

Are change here and BundleWrapper still need considering we switch to `IList<ITypedElement>` away from `BundleWrapper`

src/Microsoft.Health.Fhir.Core/Features/Operations/IFhirOperationDataStore.cs
```diff
@@ -96,7 +96,7 @@ public interface IFhirOperationDataStore
         /// Queries the datastore for any reindex job documents with a status of running, queued or paused
         /// </summary>
         /// <param name="cancellationToken">The cancellation token</param>
-        /// <returns>True if any found</returns>
-        Task<bool> CheckActiveReindexJobsAsync(CancellationToken cancellationToken);
+        /// <returns>True if any found, along with id of the job</returns>
+        Task<(bool, string)> CheckActiveReindexJobsAsync(CancellationToken cancellationToken);
```
Ivanidzo4ka, line 100: >CheckActiveReindexJobsAsync [](start = 29, length = 27)

Discussion:
Is it worth to just return collection of currently running jobs instead of this limited tuple?
I personally find it's been a bit more cleaner in terms of how interface looks and how to work with it.

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/ISearchParameterUtilities.cs
```diff
@@ -0,0 +1,15 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System.Threading;
+using Hl7.Fhir.ElementModel;
+
+namespace Microsoft.Health.Fhir.Core.Features.Search.Parameters
+{
+    public interface ISearchParameterUtilities
+    {
+        System.Threading.Tasks.Task AddSearchParameterAsync(ITypedElement searchParam, CancellationToken cancellationToken);
```
Ivanidzo4ka, line 13: >System.Threading.Tasks. [](start = 8, length = 23)

super nit: can this be moved to using?

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/SearchParameterUtilities.cs
```diff
@@ -0,0 +1,40 @@
...
+            ISearchParameterDefinitionManager searchParameterDefinitionManager)
+        {
+            EnsureArg.IsNotNull(searchParameterStatusManager, nameof(searchParameterStatusManager));
+            EnsureArg.IsNotNull(searchParameterDefinitionManager, nameof(searchParameterDefinitionManager));
+
+            _searchParameterStatusManager = searchParameterStatusManager;
+            _searchParameterDefinitionManager = searchParameterDefinitionManager;
+        }
+
+        public async System.Threading.Tasks.Task AddSearchParameterAsync(ITypedElement searchParam, CancellationToken cancellationToken)
```
Ivanidzo4ka, line 32: >cancellationToken [](start = 118, length = 17)

nit: I find it's a bit strange to have this dead parameter in function.

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/SearchParameterUtilities.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+            _searchParameterDefinitionManager.AddNewSearchParameters(new List<ITypedElement>() { searchParam });
+
+            var searchParameterWrapper = new SearchParameterWrapper(searchParam);
+            await _searchParameterStatusManager.AddSearchParameterStatusAsync(new List<string>() { searchParameterWrapper.Url });
```
Ivanidzo4ka, line 37: >AddSearchParameterStatusAsync [](start = 48, length = 29)

Just curious, do we plan in future to work with multiple parameters? Right now it feels like it's always one parameter only, and if that the case, we can simplified code if we move away from collection to single entity.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Reindex/ReindexUtilities.cs
```diff
@@ -91,30 +92,15 @@ public async Task ProcessSearchResultsAsync(SearchResult results, IReadOnlyDicti
 
         public async Task<(bool, string)> UpdateSearchParameters(IReadOnlyCollection<string> searchParameterUris, CancellationToken cancellationToken)
```
Ivanidzo4ka, line 93: >UpdateSearchParameters [](start = 42, length = 22)

nit: I think name is a bit misleading, it's enabling search parameters rather than updating them (update for me sound like more generic operation)

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/SearchParameterUtilities.cs
feordin: We do plan to handle batches of SearchParameters soon.  Just starting with single add for the moment.

src/Microsoft.Health.Fhir.Core/Features/Operations/IFhirOperationDataStore.cs
feordin: For now just one job can be running, and I do like having the bool for a simple true/false comparison.

src/Microsoft.Health.Fhir.Core/Features/Search/SearchHelperUtilities.cs
feordin: I had not previously considered it, but I after looking at the Combine method, I am not sure it is a good match, as we have a varying number of items to combine, and the length can be more than what the different signatures of Combine allow for.

johnstairs: For those scenarios, you can call HashCode.Add() like [here](https://source.dot.net/#System.Diagnostics.DiagnosticSource/System/Diagnostics/ActivityLink.netcoreapp.cs,24). Another approach is to call Combine() successively, but Add() likely gives better results.

src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterDefinitionManager.cs
feordin: This is a good idea.  After a quick look, we need to do some additional refactoring to get the necessary info to be able to filter correctly.  The Build method needs to return a list of affected resource types.  I think we should split this to a separate work item, as this PR is big enough already :).  Added item #78077.

src/Microsoft.Health.Fhir.Core/Features/Search/SearchHelperUtilities.cs
```diff
@@ -3,39 +3,52 @@
...
+        public static string CalculateSearchParameterHash(IEnumerable<SearchParameterInfo> searchParamaterInfos)
         {
-            EnsureArg.IsNotNull(resourceSearchParameterStatus, nameof(resourceSearchParameterStatus));
-            EnsureArg.IsGt(resourceSearchParameterStatus.Count(), 0, nameof(resourceSearchParameterStatus));
+            EnsureArg.IsNotNull(searchParamaterInfos, nameof(searchParamaterInfos));
+            EnsureArg.IsGt(searchParamaterInfos.Count(), 0, nameof(searchParamaterInfos));
 
-            StringBuilder sb = new StringBuilder();
-            foreach (ResourceSearchParameterStatus searchParameterStatus in resourceSearchParameterStatus.OrderBy(x => x.Uri.ToString()))
+            HashCode hashCode = default;
```
Ivanidzo4ka, line 31: >HashCode [](start = 12, length = 8)

Just want to note what by using HashCode object you heavily reduce dimensionality of hash (it was 256 bit, now it would be 32).
Which is probably fine, considering low amount of objects we need to hash.

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Create/CreateResourceHandler.cs
```diff
@@ -70,7 +76,16 @@ public async Task<UpsertResourceResponse> Handle(CreateResourceRequest message,
...
             resource.VersionId = result.Wrapper.Version;
 
-            return new UpsertResourceResponse(new SaveOutcome(new RawResourceElement(result.Wrapper), SaveOutcomeType.Created));
+            var response = new UpsertResourceResponse(new SaveOutcome(new RawResourceElement(result.Wrapper), SaveOutcomeType.Created));
+
+            if (message.Resource.InstanceType.Equals(KnownResourceTypes.SearchParameter, StringComparison.Ordinal))
+            {
+                // Once the SearchParameter resource is committed to the data store, we can update the in
+                // memory SearchParameterDefinitionManager, and persist the status to the data store
+                await _searchParameterUtitliies.AddSearchParameterAsync(message.Resource.Instance);
```
Ivanidzo4ka, line 85: >await _searchParameterUtitliies.AddSearchParameterAsync(message.Resource.Instance); [](start = 16, length = 83)

So if this one fail (let's say on status update for search parameter, since it's a work with external resource), what exactly we gonna return to user? I'm sure we catch exception in filter, but would it be some generic thing or something more specific.
What kind of actions user need to perform to recover from such error? Does it need to just run reindex?

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Operations/Reindex/ReindexUtilitiesTests.cs
```diff
@@ -60,7 +67,7 @@ public async Task GivenResourcesWithUnchangedOrChangedIndices_WhenResultsProcess
             var resultList = new List<SearchResultEntry>();
             resultList.Add(entry1);
             resultList.Add(entry2);
-            var result = new SearchResult(resultList, "token", null, new List<Tuple<string, string>>());
+            var result = new SearchResult(resultList, "token", new List<(SearchParameterInfo, SortOrder)>(), new List<Tuple<string, string>>());
```
Ivanidzo4ka, line 70: >new List<(SearchParameterInfo, SortOrder)>() [](start = 63, length = 44)

Just curious, is this change because it unblocks something, or it's just for how nice code look this way?

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/Expressions/Parsers/ExpressionParserTests.cs
```diff
@@ -302,6 +302,7 @@ private SearchParameterInfo SetupSearchParameter(ResourceType resourceType, stri
             {
                 Name = paramName,
                 Type = SearchParamType.String,
+                Url = $"http://testparameter/{resourceType}-{paramName}",
```
Ivanidzo4ka, line 305: >Url = $"http://testparameter/{resourceType}-{paramName}", [](start = 16, length = 57)

Do we need to update `SearchParameterInfo` constructor to ensure Url is present?

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/SearchParameters/SearchParameterDefinitionManagerTests.cs
```diff
@@ -40,6 +43,8 @@ public class SearchParameterDefinitionManagerTests : IAsyncLifetime
         private readonly ISearchParameterSupportResolver _searchParameterSupportResolver;
         private readonly IFhirRequestContextAccessor _fhirRequestContextAccessor;
         private readonly IFhirRequestContext _fhirRequestContext = new DefaultFhirRequestContext();
+        private readonly SearchParameterStatusManager _searchParameterStatusManager;
```
Ivanidzo4ka, line 46: >SearchParameterStatusManager [](start = 25, length = 28)

why there is two of them? 
`private readonly SearchParameterStatusManager _manager;`
`private readonly SearchParameterStatusManager _searchParameterStatusManager;`

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/SearchParameters/SearchParameterValidatorTests.cs
```diff
@@ -0,0 +1,105 @@
...
+        [Theory]
+        [MemberData(nameof(InvalidSearchParamData))]
+        public async Task GivenInvalidSearchParam_WhenValidatingSearchParam_ThenResourceNotValidExceptionThrown(SearchParameter searchParam, string method)
+        {
+            var validator = new SearchParameterValidator(() => _fhirOperationDataStore.CreateMockScope(), _authorizationService, _searchParameterDefinitionManager);
+            await Assert.ThrowsAsync<ResourceNotValidException>(() => validator.ValidateSearchParamterInput(searchParam, method, CancellationToken.None));
+        }
+
+        [Theory]
+        [MemberData(nameof(ValidSearchParamData))]
```
Ivanidzo4ka, line 49: >MemberData [](start = 9, length = 10)

That's a cool trick to pass parameters which are not const objects. I need to remember that.

src/Microsoft.Health.Fhir.Shared.Web/appsettings.json
```diff
@@ -59,7 +59,7 @@
                 ]
             },
             "Reindex": {
-                "Enabled": false,
+                "Enabled": true,
```
Ivanidzo4ka, line 62: >true [](start = 27, length = 4)

Is this change for tests purposes? I'm not sure it's a good idea to enable functionality by default which works only in Cosmos.

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Operations/Reindex/ReindexUtilitiesTests.cs
feordin: I think it was just for nice looking code :).

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/Expressions/Parsers/ExpressionParserTests.cs
feordin: Looking through the code, currently there are quite a few instances where SearchParamterInfo is instantiated with an empty URL.  Mostly test code.  For custom search parameters, we definitely need it populated, but we check for that case.  I am thinking to leave as is for now.

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Create/CreateResourceHandler.cs
feordin: Good point!  After checking, depending on the exception it will fail with a 500 and give a full stack trace with no formatted exception at all.  Updated with better exception handling in the SearchParameterUtilities.

build/jobs/provision-deploy.yml
```diff
@@ -22,6 +22,9 @@ parameters:
 - name: schemaAutomaticUpdatesEnabled
   type: boolean
   default: false
+- name: reindexEnabled
+  type: boolean
+  default: false
```
LTA-Thinking, line 27: I don't think this is needed as a parameter. Is there a case where we would want to deploy to our test servers without reindex enabled?

build/jobs/deploy-aks.yml
```diff
@@ -21,6 +21,9 @@ parameters:
 - name: schemaAutomaticUpdatesEnabled
   type: boolean
   default: false
+- name: reindexEnabled
+  type: boolean
+  default: false
```
LTA-Thinking, line 26: I don't think this is needed as a parameter. Is there a case where we would want to deploy to our test servers without reindex enabled?

LTA-Thinking (approved):
YML files look good.

samples/templates/default-azuredeploy.json
```diff
@@ -176,6 +176,13 @@
             "metadata": {
                 "description": "Determines whether the $convert-data operation will be enabled for this fhir instance. If true, an Azure container registry will be created as part of the deployment. You will need owner or user-administrator permissions for this."
             }
+        },
+        "enableReindex": {
+            "type": "bool",
+            "defaultValue": false,
```
brendankowitz, line 182: Any reason why we wouldn't now default this to on?

src/Microsoft.Health.Fhir.Core/Features/Operations/IFhirOperationDataStore.cs
brendankowitz: nit: Can give these names: `(bool Found, string JobId)`. 

Though, having written that is feels more like the pattern:
```
Task<bool> TryGetReindexJobsAsync(out string jobId, CancellationToken cancellationToken);
```

src/Microsoft.Health.Fhir.Core/Features/Search/ConfigureCustomSearchException.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// The exception that is thrown when the search parameter is not supported.
+    /// </summary>
+    public class ConfigureCustomSearchException : FhirException
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ConfigureCustomSearchException"/> class.
+        /// </summary>
+        /// <param name="error">The error message to include in the operation outcome issues list.</param>
+        public ConfigureCustomSearchException(string error)
+        {
```
brendankowitz, line 21: The pattern we landed on here is that exceptions should be using `Debug.Assert(!string.IsNullOrEmpty(message), "Exception message should not be empty");` to indicate a problem with the message.
I see some other exceptions using EnsureArg, which also isn't great as that will throw another exception from the exception 😕

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/SearchParameterUtilities.cs
```diff
@@ -0,0 +1,65 @@
...
+using Hl7.Fhir.ElementModel;
+using Microsoft.Health.Fhir.Core.Exceptions;
+using Microsoft.Health.Fhir.Core.Features.Definition;
+using Microsoft.Health.Fhir.Core.Features.Definition.BundleWrappers;
+using Microsoft.Health.Fhir.Core.Features.Search.Registry;
+using Microsoft.Health.Fhir.Core.Models;
+
+namespace Microsoft.Health.Fhir.Core.Features.Search.Parameters
+{
+    public class SearchParameterUtilities : ISearchParameterUtilities
```
brendankowitz, line 19: Can `Utilities` be internal? Or could this be an extension on ISearchParameterDefinitionManager?

src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterStatusManager.cs
```diff
@@ -88,63 +88,43 @@ public async Task EnsureInitialized()
...
+                searchParameterStatusList.Add(new ResourceSearchParameterStatus()
+                {
+                    LastUpdated = Clock.UtcNow,
+                    Status = status,
+                    Uri = searchParamUri,
+                });
             }
 
-            await _mediator.Publish(new SearchParametersHashUpdated(resourceHashMap));
+            await _searchParameterStatusDataStore.UpsertStatuses(searchParameterStatusList);
```
brendankowitz, line 118: We had a bug in the beginning with multiple instances over-writing each other, I haven't traced the calls to this, just FYI

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Create/CreateResourceHandler.cs
brendankowitz: This is when I'd separate with INotification, could be a separate PR, but something like `ResourceCreatedNotification : INotification`. These will be useful and probably required for things like subscriptions.

Alternatively, this hook is also available with `IPipelineBehavior<CreateResourceRequest, UpsertResourceResponse>>` or `RequestPostProcessorBehavior<CreateResourceRequest, UpsertResourceResponse>`

samples/templates/default-azuredeploy.json
brendankowitz: Ah, reason above, waiting for SQL 👍  #closed

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Create/CreateResourceHandler.cs
feordin: This is good info @brendankowitz , I added a story to review and possibly reduce the coupling using mediatr to hook up this behavior.  AB#78200

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/SearchParameterUtilities.cs
feordin: With your other comment about hooking this functionality up using a different mediatr method, I would review the changes here along with that AB#78200.

johnstairs: I agree, and Utilities sounds more like a drawer of helper methods :)

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/SearchParameterUtilities.cs
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        public async Task AddSearchParameterAsync(ITypedElement searchParam)
+        {
+            try
+            {
+                _searchParameterDefinitionManager.AddNewSearchParameters(new List<ITypedElement>() { searchParam });
+
+                var searchParameterWrapper = new SearchParameterWrapper(searchParam);
+                await _searchParameterStatusManager.AddSearchParameterStatusAsync(new List<string>() { searchParameterWrapper.Url });
```
johnstairs, line 42: What does it mean if this line fails but the `AddNewSearchParameters` call succeeded? Will we be in an inconsistent state?

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/SearchParameterFilterAttribute.cs
```diff
@@ -0,0 +1,47 @@
...
+        public override async Task OnActionExecutionAsync(ActionExecutingContext context, ActionExecutionDelegate next)
+        {
+            EnsureArg.IsNotNull(context, nameof(context));
+
+            if (context.ActionArguments.TryGetValue(KnownActionParameterNames.Resource, out var parsedModel))
+            {
+                if (parsedModel is SearchParameter searchParameter)
+                {
+                    // wait for the validation checks to pass before allowing the FHIRController action to continue
+                    await _searchParameterValidator.ValidateSearchParamterInput(
```
johnstairs, line 37: We typically do validation in the mediator handlers. I wonder if we should do the same in the cases too.

src/Microsoft.Health.Fhir.Core/Features/Search/SearchHelperUtilities.cs
johnstairs: Yes. When using this kind of hash code, it's usually followed up by a full equality check, which we are not doing. It seems like we should be doing a proper cryptographic hash here.

johnstairs: Also, it we are persisting these hashes, then this approach is not deterministic

feordin: Reverted the changes to the hash calculation, because as John mentioned the new approach would be deterministic.

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/SearchParameterFilterAttribute.cs
feordin: Discussed with John, and for now we will continue with this validation here in the filter.  It will allow us to quickly abort processing the SearchParameter with some initial validation, such as a repeated URL.

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/SearchParameterUtilities.cs
feordin: If AddSearchParameterStatusAsync fails after the SearchParameter is added to the definition manager, then the parameter would not be usable at all.  No newly posted resources would use it for extraction, and no reindex job would pick it up either.   The client would need to PUT an update to the parameter to fix cause it to be reprocessed, or restart the FHIR service.   When the code is implemented to poll for new changes in SearchParameters, it could also fix issues such as this.  I will make a note in that user story.

src/Microsoft.Health.Fhir.Core/Features/Search/SearchHelperUtilities.cs
johnstairs: This was not a good suggestion. I did not realize that this was meant to be a proper cryptographic hash and not a HashCode implementation. Apologies for the detour.

johnstairs (approved):

PR #1450, opened 2020-11-13
Changed _include search to return referenced version instead of latest

valeneiko:
## Description
When a resource contains a versioned reference, if querying with `_include` previously the latest version of the resource was returned, with this PR referenced version of the resource is returned. 

From the spec (https://www.hl7.org/fhir/search.html#versions):

> If a resource has a reference that is versioned and _include is performed, the specified version SHOULD be provided.

#### Given the resource:
```json
{
    "id": "123",
    "status": "final",
    "code": {
        "coding": [
            {
                "system": "http://loinc.org",
                "code": "60832-3",
                "display": "Room temperature"
            }
        ]
    },
    "subject": {
        "reference": "Patient/456/_history/1",
        "type": "Patient"
    }
}
```

#### If queried:
```curl
GET /Observation?_include=Observation:subject
```

Previously latest version of the `Patient` resource was included, after the PR referenced version is included.

## Testing
Unit and E2E tests.


src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -811,4 +838,41 @@ private DenormalizedExpressionContainsIdParameterVisitor()
...
+                    sb.Append(delimiter);
+                },
+                sb =>
+                {
+                    sb.AppendLine();
+                    sharedScope?.Dispose();
+                    sb.Append(")");
+                });
+        }
+    }
```
anyname2, line 877: This extension method can be removed after this PR is merged: https://github.com/microsoft/healthcare-shared-components/pull/124

johnstairs:
@anyname2, thanks for the PR! I think if we make this change that chained searches should also follow the same behavior. This worries me a little bit because the nonclustered indexes are all filtered by `IsHistory=0`, but but inner loop joins should still be efficient, provided they use the clustered index for the references (but this may not always be optimal).
Also, this change requires re-indexing (cc @brendankowitz) (and we should also implement the same include behavior for Cosmos DB). 
Thoughts?

anyname2:
> @anyname2, thanks for the PR! I think if we make this change that chained searches should also follow the same behavior. This worries me a little bit because the nonclustered indexes are all filtered by `IsHistory=0`, but but inner loop joins should still be efficient, provided they use the clustered index for the references (but this may not always be optimal).
> Also, this change requires re-indexing (cc @brendankowitz) (and we should also implement the same include behavior for Cosmos DB).
> Thoughts?

@johnstairs , the spec is quite permissive with regards to chained searches:
> * If a resource has a reference that is versioned, and chaining is performed, the criteria should ideally be evaluated against the version referenced, but most systems will not be capable of this because search is only defined to function against the current version of a resource
> * Where a search does not act on the referenced version, search results SHOULD contain a OperationOutcome with a warning that indicates the discrepancy

I have the same concerns about the performance if we start searching with versioned references. Maybe returning a warning would be better? *(and implement chained searches with verisoned references later, possibly with a feature flag to allow users to choose the behaviour)*

rbans96:
Closing out since its already been referenced for API versioning milestone.

PR #1451, opened 2020-11-13
Support null Format config

LTA-Thinking:
## Description
If the Format config isn't included in settings calls to create an export job will return 500.

## Testing
Existing unit and e2e tests.


Ivanidzo4ka (approved):
:shipit:

brendankowitz (approved):

Issue #1421, opened 2020-11-04
Rename app roles


pjirsa:
https://github.com/microsoft/fhir-server/blob/034aea8a6ab183c00a0cc2beed4d8b8a8f27cc26/src/Microsoft.Health.Fhir.Shared.Web/roles.json#L39

Role names including term "global" (i.e. globalAdmin) create ambiguity with other systems or built-in Azure AD role names such as "Global Administrator". This naming convention potentially raises security concerns during IT SecOps audits.

Consider renaming to avoid confusion.

Suggestions:
- fhirReader
- fhirWriter
- fhirExporter
- fhirAdmin

CaitlinV39:
@pjirsa thank you so much for submitting this. We have labeled this up for grabs and are happy to have someone from the community to make this change. We simply have these as placeholder names.

CaitlinV39:
@pjirsa after discussing further, we would not want to change these in our repo. The reason for the global prefix is taht the roles are scoped to the entire FHIR server. The plan, one day, is to be able to define "slices" that represent a scoped view of a FHIR service. We would also want to be cautious of making this change as it would be breaking for anyone that has already deployed.

Definitely feel free to update for your own config, but we will leave these names as is for now.

PR #1457, opened 2020-11-17
Update role names

pjirsa:
## Description
Update app role names to avoid collision with other 'global' role types. (i.e. AAD GlobalAdmin, etc.)

## Related issues
Addresses #1421.


CaitlinV39:
@pjirsa - I'm so sorry for the timespan between my two comments but for now, we would not like to bring this change into our repo.

CaitlinV39:
@pjirsa per the issue, we won't make this change to the repo for now to avoid breaking changes and based on future plans we have for RBAC.

PR #1462, opened 2020-11-17
Add support for _typeFilter to export

LTA-Thinking:
## Description
Add support for the _typeFilter parameter when creating an export job.
[Fhir spec](https://hl7.org/fhir/uv/bulkdata/export/index.html#experimental-query-parameters)

## Testing
Manually tested. New and existing unit and e2e tests run.


src/Microsoft.Health.Fhir.Core/Features/KnownQueryParameterNames.cs
```diff
@@ -69,5 +69,9 @@ public static class KnownQueryParameterNames
         public const string AnonymizationConfigurationFileEtag = "_anonymizationConfigEtag";
 
         public const string OutputFormat = "_outputFormat";
+
+        public const string TypeFilter = "_typeFilter";
+
+        public const string IncludeAssociatedData = "includeAssociatedData";
```
Ivanidzo4ka, line 75: > public const string IncludeAssociatedData = "includeAssociatedData"; [](start = 7, length = 69)

I can't find any usage of this.

src/Microsoft.Health.Fhir.Core/Features/Operations/JobRecordProperties.cs
```diff
@@ -99,6 +99,12 @@ public static class JobRecordProperties
 
         public const string StorageAccountContainerName = "storageAccountContainerName";
 
+        public const string Filters = "filters";
+
+        public const string Filter = "filter";
```
Ivanidzo4ka, line 104: One for JobRecord, and another is for JobProgress, but what is the difference?
Also in JobProgress:
why one property is plural and other is singular?
```
      [JsonProperty(JobRecordProperties.Filter)]
        public ExportJobFilter Filter { get; private set; }

        [JsonProperty(JobRecordProperties.FinishedFilters)]
        public bool FinishedFilters { get; private set; }
```
Can this properties be distinguished by more than just a plurality (not sure it's a right word, basically on attribute of how many)

src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/Parsers/SearchValueExpressionBuilderHelper.cs
```diff
@@ -271,6 +271,34 @@ void ISearchValueVisitor.Visit(TokenSearchValue token)
...
+                {
+                    // If the system is not supplied, then the token code is matched irrespective of the value of system.
+                    _outputExpression = Expression.NotContains(FieldName.TokenCode, _componentIndex, token.Code, false);
+                }
+                else if (token.System.Length == 0)
+                {
+                    // If the system is empty, then the token is matched if there is no system property.
+                    _outputExpression = Expression.And(
+                        Expression.Missing(FieldName.TokenSystem, _componentIndex),
+                        Expression.NotContains(FieldName.TokenCode, _componentIndex, token.Code, false));
```
Ivanidzo4ka, line 288: All other expression is do opposite of absence of modifier.
but not this one.
`not (a and b)` should be: `not(a) or not(b)`
but in this case it becomes: `(a) and not(b)`

shouldn't this be Expression.Or (Expression.Exist(), Expression.NotContains)?
I also surprised absence of `StringOperator.NotEquals`

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/CreateExportRequestHandler.cs
```diff
@@ -77,6 +77,29 @@ public async Task<CreateExportResponse> Handle(CreateExportRequest request, Canc
...
+                    var parameterTupleList = new List<Tuple<string, string>>();
+
+                    foreach (string parameter in filterParameters)
+                    {
+                        var keyValue = parameter.Split("=");
+                        parameterTupleList.Add(new Tuple<string, string>(keyValue[0], keyValue[1]));
+                    }
+
+                    filters.Add(new ExportJobFilter(filterType, parameterTupleList));
+                }
```
Ivanidzo4ka, line 100: This looks quite complex, can it be make a separate class/method which can be unit tested?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -237,8 +234,76 @@ private async Task UpdateJobRecordAsync(CancellationToken cancellationToken)
...
+                            progress.SetFilter(null);
+                        }
+                    }
+                }
+
+                progress.MarkFiltersFinished();
+                await UpdateJobRecordAsync(cancellationToken);
+            }
+
+            if (_exportJobRecord.ExportType == ExportJobType.All || !filteredResources.Contains(KnownResourceTypes.Patient))
```
Ivanidzo4ka, line 283: >Contains [](start = 87, length = 8)

nit: shouldn't this use  StringComparison.Ordinal as in loop above?

Ivanidzo4ka, line 283: >_exportJobRecord.ExportType == ExportJobType.All [](start = 16, length = 48)

Took me awhile to figure out what is going on here.
Basically you split types by having filter and not having filter, and then first run for types which have filter parameters by doing `/search/TypeA/?{filter}` and then you run `/search/?_type:not={typeA,typeB}`

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/CreateExportRequestHandler.cs
```diff
@@ -77,6 +77,29 @@ public async Task<CreateExportResponse> Handle(CreateExportRequest request, Canc
...
 
+            var filters = new List<ExportJobFilter>();
+
+            if (!string.IsNullOrWhiteSpace(request.Filters))
+            {
+                var filterArray = request.Filters.Split(",");
+                foreach (string filter in filterArray)
+                {
+                    var parameterIndex = filter.IndexOf("?", StringComparison.Ordinal);
+                    var filterType = filter.Substring(0, parameterIndex);
```
Ivanidzo4ka, line 88: This can be -1, which will throw ArgumentOutOfRangeException.
This whole method need to be wrapped in try/catch and handled properly in case of bad input.

src/Microsoft.Health.Fhir.Core/Features/KnownQueryParameterNames.cs
LTA-Thinking: Sorry, left over from testing a different feature. Will remove.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/CreateExportRequestHandler.cs
LTA-Thinking: I can move this to a private method, but I don't think it should be public since it is only used here. There are some new unit tests for this code in CreateExportRequestHandlerTests.cs

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
LTA-Thinking: I don't know if the HashSet can take a StringComparison for its contains method. I'll look into it.

LTA-Thinking: Patient is a special case because if a Patient or Group level export is run then the first search is just for patients and then there is a compartment search for associated resources. We only want to apply filters for the first search if they relate to patient resources in that case. I'll add some comments to explain what is going on.

src/Microsoft.Health.Fhir.Core/Features/Operations/JobRecordProperties.cs
LTA-Thinking: Filters is for the entire list of filters for the job. Filter is for the filter that is currently being evaluated, so we can restart in the case of an interrupted job. FinishedFilters is a flag that indicates that all filters have been evaluated and we are now doing the search for any resources that didn't have filters.

LTA-Thinking: I'll add comments and see what I can do to change the names.

src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/Parsers/SearchValueExpressionBuilderHelper.cs
LTA-Thinking: I wish there was a StringOperator.NotEquals. You're right, this should have been fully reversed, I'll fix that.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/CreateExportRequestHandler.cs
LTA-Thinking: I'm adding in a catch for a badly formatted parameter. So long as the format is correct (x?y=z) I think it should allow the input, even if it will fail the search latter on because of an invalid resource type or search parameters.

src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/Parsers/SearchValueExpressionBuilderHelper.cs
Ivanidzo4ka: Actually, disregard my comment, it's valid for one dimensional case, but here you have multidimensional case for system and code.
I probably shouldn't but out of curiosity I start implementing NotEquals, and I'll try to post PR with changes within day.

---
In reply to: [526329666](https://github.com/microsoft/fhir-server/pull/1462#discussion_r526329666) [](ancestors = 526329666)

LTA-Thinking: I think your comment is still accurate for this case. For the case of not "|<code>" it is asking for if it isn't equal to the code or it has a system. Thanks for looking into the NotEquals expression.

LTA-Thinking: Replace the entire section with the new Not expression

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/CreateExportRequestHandler.cs
```diff
@@ -122,5 +106,72 @@ public async Task<CreateExportResponse> Handle(CreateExportRequest request, Canc
...
+        /// <returns>A list of <see cref="ExportJobFilter"/></returns>
+        private IList<ExportJobFilter> ParseFilter(string filterString)
+        {
+            var filters = new List<ExportJobFilter>();
+
+            try
+            {
+                if (!string.IsNullOrWhiteSpace(filterString))
+                {
+                    var filterArray = filterString.Split(",");
```
namadabu, line 123: Is it possible that one of the `_typefilter` queries could contain a parameter that contains a comma separated list of values? In that case splitting using `,` would cause incorrect splitting.

LTA-Thinking: This is a good point, but I'm unsure how to handle it. The comma is specified by the [spec](https://hl7.org/fhir/uv/bulkdata/export/index.html#example-request-with-_typefilter). I could make the parsing guess at what roll the comma has based on the existence of other characters (specifically an equals sign). I'll see what I can do.

Ivanidzo4ka: I would suggest what in that case inside comma need to be escaped.
Also just to note, there is SplitByOrSeparator() extension on top of String type, would be better to use it rather than custom split function.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -354,6 +416,69 @@ private async Task<IAnonymizer> CreateAnonymizerAsync(CancellationToken cancella
...
+                {
+                    foreach (var filter in _exportJobRecord.Filters)
+                    {
+                        index++;
+                        if (progress.CurrentFilter != null && filter != progress.CurrentFilter)
+                        {
+                            continue;
+                        }
+
+                        if (_exportJobRecord.ExportType == ExportJobType.All || filter.Type.Equals(KnownResourceTypes.Patient, StringComparison.Ordinal))
```
namadabu, line 441: `RunExportCompartmentSearch` is only executed for export job types `Patient` and `Group`. This check for `ExportJobType.All` is redundant.

namadabu, line 441: `RunExportCompartmentSearch` is only executed for export job types `Patient` and `Group`. This check for `ExportJobType.All` is redundant. Here and below

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/Models/ExportJobFilter.cs
```diff
@@ -0,0 +1,35 @@
...
+            Parameters = parameters;
+        }
+
+        [JsonConstructor]
+        public ExportJobFilter()
+        {
+        }
+
+        [JsonProperty(JobRecordProperties.ResourceType)]
+        public string Type { get; private set; }
```
namadabu, line 30: nit: maybe rename `Type` to `ResourceType` to make it more descriptive

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -354,6 +416,69 @@ private async Task<IAnonymizer> CreateAnonymizerAsync(CancellationToken cancella
...
+                            await SearchCompartmentWithFilter(exportJobConfiguration, progress, filterQueryParametersList, batchIdPrefix + "-filter" + index, cancellationToken);
+                        }
+                    }
+                }
+
+                progress.MarkFiltersFinished();
+                await UpdateJobRecordAsync(cancellationToken);
+            }
+
+            if (_exportJobRecord.ExportType == ExportJobType.All || !filteredResources.Contains(KnownResourceTypes.Patient))
```
namadabu, line 461: If there are 2 filters - one for `Patient` and another for `MedicationRequest` - it looks like we will never use the `MedicationRequest` filter while searching the compartment. `!filteredResources.Contains(KnownResourceTypes.Patient)` will return true and the previous if condition runs the search only with `Patient` filters.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
namadabu, line 441: Wouldn't we have already filtered for the `Patients` by the time we are here?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
LTA-Thinking: Wow, this was really broken. Thank you for catching this. I'll fix this and add some new unit tests.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -354,6 +418,66 @@ private async Task<IAnonymizer> CreateAnonymizerAsync(CancellationToken cancella
...
+                        index++;
+                        if (progress.CurrentFilter != null && filter != progress.CurrentFilter)
+                        {
+                            continue;
+                        }
+
+                        progress.SetFilter(filter);
+
+                        List<Tuple<string, string>> filterQueryParametersList = new List<Tuple<string, string>>(queryParametersList);
+                        filterQueryParametersList.Add(Tuple.Create(KnownQueryParameterNames.Type, filter.ResourceType));
```
namadabu, line 446: We are adding the resourceType to the query parameters here but also passing the filter.ResourceType to search below (where it gets added as a "higher" level filter). Seems like we can remove one of them.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -354,6 +418,66 @@ private async Task<IAnonymizer> CreateAnonymizerAsync(CancellationToken cancella
...
+                    }
+                }
+
+                progress.MarkFiltersFinished();
+                await UpdateJobRecordAsync(cancellationToken);
+            }
+
+            foreach (var resource in filteredResources)
+            {
+                queryParametersList.Add(Tuple.Create(KnownQueryParameterNames.Type + ":not", resource));
```
namadabu, line 464: Assuming a sample request: `Patient/$export?_type=MedicationRequest&_typeFilter=MedicationRequest%3Fstatus%3Dactive`. Here we end up doing a search that has both _type=MedicationRequest and _type:not=MedicationRequest. This is an unnecessary search. Granted the export results won't be incorrect since we got the data that was requested (in the previous filtered search). But we should still avoid doing this search.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -237,8 +234,78 @@ private async Task UpdateJobRecordAsync(CancellationToken cancellationToken)
...
+                        {
+                            progress.SetFilter(filter);
+
+                            List<Tuple<string, string>> filterQueryParametersList = new List<Tuple<string, string>>(queryParametersList);
+                            foreach (var param in filter.Parameters)
+                            {
+                                filterQueryParametersList.Add(param);
+                            }
+
+                            await SearchWithFilter(exportJobConfiguration, progress, filter.ResourceType, filterQueryParametersList, sharedQueryParametersList, anonymizer, "filter" + index + "-", cancellationToken);
```
namadabu, line 272: Similarly, here also we are including `_type` as part of the queryparameters (if _exportJobRecord.ResourceType != null) and passing a higher level resource type filter in `filter.ResourceType`.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -237,8 +234,78 @@ private async Task UpdateJobRecordAsync(CancellationToken cancellationToken)
...
+                await UpdateJobRecordAsync(cancellationToken);
+            }
+
+            // This runs if the filteredResources doesn't contain Patient so that when a Patient or Group export is run it will run a full search of Patients.
+            // If Patients was filtered the search would already have been run above.
+            if (_exportJobRecord.ExportType == ExportJobType.All || !filteredResources.Contains(KnownResourceTypes.Patient))
+            {
+                foreach (var resource in filteredResources)
+                {
+                    queryParametersList.Add(Tuple.Create(KnownQueryParameterNames.Type + ":not", resource));
```
namadabu, line 289: I am wondering whether the following might be easier to do compared to using `:not`. After running all the searches for the filtered resources, we can then run a single search for the rest of the resource types present in `_type`. That would mean we will have to parse the `_type` parameter values, compare with `filteredResources` (which we already have) and remove the common ones. We would get two advantages though:
1. No potential empty searches (if _type=MedicationRequest&_typeFilter=MedicationRequest%3Fstatus%3Dactive)
2. Should work for SQL as well since we don't have to use the `:not` modifier

src/Microsoft.Health.Fhir.Api/Features/Filters/ValidateExportRequestFilterAttribute.cs
```diff
@@ -45,6 +45,7 @@ public ValidateExportRequestFilterAttribute()
                 KnownQueryParameterNames.Type,
                 KnownQueryParameterNames.Container,
                 KnownQueryParameterNames.Format,
+                KnownQueryParameterNames.TypeFilter,
```
namadabu, line 48: Do we want to validate that when `TypeFilter` is present, `Type` should also be part of the query parameters?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -237,8 +234,78 @@ private async Task UpdateJobRecordAsync(CancellationToken cancellationToken)
...
+            {
+                foreach (var filter in _exportJobRecord.Filters)
+                {
+                    filteredResources.Add(filter.ResourceType);
+                }
+            }
+
             IAnonymizer anonymizer = IsAnonymizedExportJob() ? await CreateAnonymizerAsync(cancellationToken) : null;
 
+            if (!progress.RunFilteredSearches)
```
namadabu, line 248: nit: The name `RunFilteredSearches` seems to convey that we should run filtered searches when it is `true`. But we do the opposite here :). Do you think renaming it to `FilteredSearchesComplete` or something similar can be better?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/Models/ExportJobProgress.cs
```diff
@@ -10,12 +10,15 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Export.Models
...
-        public ExportJobProgress(string continuationToken, uint page, string resourceId = null, ExportJobProgress subSearch = null)
+        public ExportJobProgress(string continuationToken, uint page, string resourceId = null, ExportJobProgress subSearch = null, ExportJobFilter filter = null)
         {
             ContinuationToken = continuationToken;
             Page = page;
             TriggeringResourceId = resourceId;
             SubSearch = subSearch;
+            CurrentFilter = filter;
+
+            RunFilteredSearches = false;
```
namadabu, line 21: Can we set `RunFilteredSearches = true` if `filter == null`? This we can avoid the initial check in `ExportJobTask` `if (!progress.RunFilteredSearches)` where we have to do an extra `UpdateJobRecordAsync` to set this to `true`. This way we can avoid the update call for requests that do not have any filters.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
LTA-Thinking: Here I think it is needed. By my reading of the spec if the user enters the _type parameter and _typeFilter parameter the _type can override the _typeFiler. So if they have `_type=Observation&_typeFilter=MedicationRequest%3Fstatus%3Dactive` no MedicationRequests will be returned because the _type parameter isn't set to return any.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
LTA-Thinking: The issue I see is that the user isn't required to have the _type parameter if they have _typeFilter. I wish we didn't need `:not` since this story can't finish until `:not` is implemented.

src/Microsoft.Health.Fhir.Api/Features/Filters/ValidateExportRequestFilterAttribute.cs
LTA-Thinking: The spec doesn't state that `Type` is required if `TypeFilter` is present. In theory you could have `TypeFilter` without `Type`.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
namadabu: I agree that `_type` overrides whatever is in `_typefilter`. But I also feel that ideally we shouldn't execute the above query since it's a no-op.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -227,25 +228,97 @@ private async Task UpdateJobRecordAsync(CancellationToken cancellationToken)
...
+                            List<Tuple<string, string>> filterQueryParametersList = new List<Tuple<string, string>>(queryParametersList);
+                            foreach (var param in filter.Parameters)
+                            {
+                                filterQueryParametersList.Add(param);
+                            }
+
+                            await SearchWithFilter(exportJobConfiguration, progress, filter.ResourceType, filterQueryParametersList, sharedQueryParametersList, anonymizer, "filter" + index + "-", cancellationToken);
+
+                            progress.SetFilter(null);
+                            await UpdateJobRecordAsync(cancellationToken);
```
namadabu, line 278: If the worker crashes and we have to restart export from this state, `progress.CurrentFilter` will be `null`. This means we will end up re-processing all the filters that we had already completed. We should maybe have a different way of tracking completed filters. Could be a different list or remove from existing list?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -227,25 +228,97 @@ private async Task UpdateJobRecordAsync(CancellationToken cancellationToken)
...
+            {
+                if (_exportJobRecord.ExportType == ExportJobType.Patient)
+                {
+                    queryParametersList.Add(Tuple.Create(KnownQueryParameterNames.Type, KnownResourceTypes.Patient));
+                }
+                else if (_exportJobRecord.ExportType == ExportJobType.All && requestedResourceTypes != null)
+                {
+                    foreach (var resource in requestedResourceTypes)
+                    {
+                        if (!filteredResources.Contains(resource.ToUpper(CultureInfo.InvariantCulture)))
```
namadabu, line 299: >ToUpper [](start = 65, length = 7)

Curious why we are doing `ToUpper()` here

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -227,25 +228,97 @@ private async Task UpdateJobRecordAsync(CancellationToken cancellationToken)
...
 
-            if (_exportJobRecord.ExportType == ExportJobType.Patient)
+            var requestedResourceTypes = _exportJobRecord.ResourceType?.Split(',');
+            var filteredResources = new HashSet<string>();
+            if (_exportJobRecord.Filters != null)
             {
-                queryParametersList.Add(Tuple.Create(KnownQueryParameterNames.Type, KnownResourceTypes.Patient));
+                foreach (var filter in _exportJobRecord.Filters)
+                {
+                    filteredResources.Add(filter.ResourceType.ToUpper(CultureInfo.InvariantCulture));
```
namadabu, line 243: >ToUpper [](start = 62, length = 7)

Curious why we are doing ToUpper() here

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -227,25 +228,97 @@ private async Task UpdateJobRecordAsync(CancellationToken cancellationToken)
...
             }
-            else if (_exportJobRecord.ExportType == ExportJobType.All && !string.IsNullOrEmpty(_exportJobRecord.ResourceType))
+
+            IAnonymizer anonymizer = IsAnonymizedExportJob() ? await CreateAnonymizerAsync(cancellationToken) : null;
+
+            if (!progress.FilteredSearchesComplete)
             {
-                queryParametersList.Add(Tuple.Create(KnownQueryParameterNames.Type, _exportJobRecord.ResourceType));
+                var index = -1;
+                if (_exportJobRecord.Filters != null)
```
namadabu, line 252: nit: Can we flip the if condition checks? Since `progress.FilteredSearchesComplete` is `false `by default, we will call `UpdateJobRecordAsync` in order to mark that we have completed filtered searches. If we check for `_exportJobRecord.Filters != null` first, we could avoid that.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
LTA-Thinking: It was to avoid issues with users entering the resource type in different cases. I would use StringComparison but the List Contains method doesn't allow you to pass one so I need to convert them to a consistent case.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
LTA-Thinking: Yea, this method just keeps having holes. I'll make a list of completed filters since each progress record needs its own list.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
namadabu: There is an option to set the comparer when instantiating the HashSet. Using that we should be able to avoid the `ToUpper()` conversions `var filteredResources = new HashSet<string>(StringComparer.OrdinalIgnoreCase)`

---
In reply to: [537724269](https://github.com/microsoft/fhir-server/pull/1462#discussion_r537724269) [](ancestors = 537724269)

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -227,25 +228,109 @@ private async Task UpdateJobRecordAsync(CancellationToken cancellationToken)
...
+
+            // The unfiltered search should be run if there were no filters specified, there were types requested that didn't have filters for them, or if a Patient/Group level export didn't have filters for Patients.
+            // Examples:
+            // If a patient/group export job with type and type filters is run, but patients aren't in the types requested, the search should be run here but no patients printed to the output
+            // If a patient/group export job with type and type filters is run, and patients are in the types requested and filtered, the search should not be run as patients were searched above
+            // If an export job with type and type filters is run, the search should not be run if all the types were searched above.
+            if (_exportJobRecord.Filters == null ||
+                _exportJobRecord.Filters.Count == 0 ||
+                (_exportJobRecord.ExportType == ExportJobType.All &&
+                !requestedResourceTypes.All(resourceType => filteredResources.Contains(resourceType.ToUpper(CultureInfo.InvariantCulture)))) ||
```
namadabu, line 277: >!requestedResourceTypes.All [](start = 16, length = 27)

Should we check whether `requestedResourceTypes` can be null here?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -227,25 +228,109 @@ private async Task UpdateJobRecordAsync(CancellationToken cancellationToken)
...
+                        await ProcessFilter(exportJobConfiguration, progress, queryParametersList, sharedQueryParametersList, anonymizer, "filter", cancellationToken);
+                    }
+                }
+            }
+
+            // The unfiltered search should be run if there were no filters specified, there were types requested that didn't have filters for them, or if a Patient/Group level export didn't have filters for Patients.
+            // Examples:
+            // If a patient/group export job with type and type filters is run, but patients aren't in the types requested, the search should be run here but no patients printed to the output
+            // If a patient/group export job with type and type filters is run, and patients are in the types requested and filtered, the search should not be run as patients were searched above
+            // If an export job with type and type filters is run, the search should not be run if all the types were searched above.
```
namadabu, line 273: These comments are very helpful :)

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -348,20 +433,88 @@ private async Task<IAnonymizer> CreateAnonymizerAsync(CancellationToken cancella
...
+            if (_exportJobRecord.Filters != null && _exportJobRecord.Filters.Any(filter => !progress.CompletedFilters.Contains(filter)))
+            {
+                foreach (var filter in _exportJobRecord.Filters)
+                {
+                    if (!progress.CompletedFilters.Contains(filter) &&
+                        requestedResourceTypes != null &&
+                        requestedResourceTypes.Contains(filter.ResourceType, StringComparison.OrdinalIgnoreCase))
+                    {
+                        progress.SetFilter(filter);
+                        await ProcessFilterForCompartment(exportJobConfiguration, progress, queryParametersList, batchIdPrefix + "-filter", cancellationToken);
```
namadabu, line 466: >batchIdPrefix + "-filter" [](start = 113, length = 25)

Should this change per filter?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -348,20 +433,88 @@ private async Task<IAnonymizer> CreateAnonymizerAsync(CancellationToken cancella
...
+                    }
+                }
+            }
+
+            if (_exportJobRecord.Filters == null ||
+                _exportJobRecord.Filters.Count == 0 ||
+                !requestedResourceTypes.All(resourceType => filteredResources.Contains(resourceType.ToUpper(CultureInfo.InvariantCulture))))
```
namadabu, line 473: >!requestedResourceTypes.All [](start = 16, length = 27)

Should we check whether `requestedResourceTypes` can be null here?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -348,20 +433,88 @@ private async Task<IAnonymizer> CreateAnonymizerAsync(CancellationToken cancella
...
+                }
+            }
+
+            if (progress.CurrentFilter != null)
             {
-                queryParametersList.Add(Tuple.Create(KnownQueryParameterNames.Type, _exportJobRecord.ResourceType));
+                await ProcessFilterForCompartment(exportJobConfiguration, progress, queryParametersList, batchIdPrefix + "-filter", cancellationToken);
             }
 
+            if (_exportJobRecord.Filters != null && _exportJobRecord.Filters.Any(filter => !progress.CompletedFilters.Contains(filter)))
```
namadabu, line 457: >&& _exportJobRecord.Filters.Any(filter => !progress.CompletedFilters.Contains(filter)) [](start = 49, length = 86)

nit: It seems like we are checking for the same condition in the for-each loop below. Can we remove this check (since functionality will not change) in order to avoid looping through the filters twice?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/CreateExportRequestHandler.cs
namadabu: Bumping up this thread to see whether we are planning to handle this @LTA-Thinking

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
LTA-Thinking: It does inside the ProcessFilter method. It adds the index of the filter to the batch id.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
LTA-Thinking: Added exception check

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/CreateExportRequestHandler.cs
LTA-Thinking: Thanks for the reminder, I'll take a look.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
namadabu: Ah okay. Agreed. I was confused because we have a `requestedResourceTypes != null` check below.

---
In reply to: [541091271](https://github.com/microsoft/fhir-server/pull/1462#discussion_r541091271) [](ancestors = 541091271)

namadabu (approved):
:shipit:

Issue #1452, opened 2020-11-13
500 when SQL Timeout


CaitlinV39:
**Describe the bug**
A customer reported that they got a 500 when running a search query, which ended up being a timeout. 

**FHIR Version?**
Stu3 (likely other versions)

**Data provider?**
SQL Server

**To Reproduce**
1. Load 100,000 providers into an S1 SQL FHIR Server
2. Run a search to pull all providers
3. Receive a 500

**Expected behavior**
Ideally you wouldn't receive a timeout but if they do, it shouldn't be a 500

**Actual behavior**
Receive a 500

**Example Trace**

```
Microsoft.Data.SqlClient.SqlException (0x80131904): Execution Timeout Expired.  The timeout period elapsed prior to completion of the operation or the server is not responding.

 ---> System.ComponentModel.Win32Exception (258): The wait operation timed out.

   at Microsoft.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)

   at Microsoft.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)

   at Microsoft.Data.SqlClient.SqlCommand.InternalEndExecuteReader(IAsyncResult asyncResult, Boolean isInternal, String endMethod)

   at Microsoft.Data.SqlClient.SqlCommand.EndExecuteReaderInternal(IAsyncResult asyncResult)

   at Microsoft.Data.SqlClient.SqlCommand.EndExecuteReaderAsync(IAsyncResult asyncResult)

   at System.Threading.Tasks.TaskFactory`1.FromAsyncCoreLogic(IAsyncResult iar, Func`2 endFunction, Action`1 endAction, Task`1 promise, Boolean requiresSynchronization)

--- End of stack trace from previous location ---

   at Polly.Retry.AsyncRetryEngine.ImplementationAsync[TResult](Func`3 action, Context context, CancellationToken cancellationToken, ExceptionPredicates shouldRetryExceptionPredicates, ResultPredicates`1 shouldRetryResultPredicates, Func`5 onRetryAsync, Int32 permittedRetryCount, IEnumerable`1 sleepDurationsEnumerable, Func`4 sleepDurationProvider, Boolean continueOnCapturedContext)

[...]

PR #1463, opened 2020-11-17
Add index hint to SQL query

johnstairs:
Adds an index hint to the SQL we generate for what it probably the simplest search: find resources of a particular type, like `GET /Observation`. We've seen instances where the optimizer decides to perform a clustered index scan over the `Resource` table, which will likely result in very poor performance. For these exact queries, we add a hint that will force it to use an existing filtered nonclustered index that exists for this very scenario:

```sql
CREATE UNIQUE NONCLUSTERED INDEX IX_Resource_ResourceTypeId_ResourceSurrgateId ON dbo.Resource
(
    ResourceTypeId,
    ResourceSurrogateId
)
WHERE IsHistory = 0 AND IsDeleted = 0
```

Address part of #1452 (does not address timeout handling)

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
@@ -133,20 +133,23 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Search.Expressions.Visitors.Q
 
             StringBuilder.Append("FROM ").Append(VLatest.Resource).Append(" ").Append(resourceTableAlias);
 
-            if (expression.TableExpressions.Count > 0)
+            if (expression.TableExpressions.Count == 0 &&
+                !_isHistorySearch &&
+                expression.DenormalizedExpressions.Any(e => e.AcceptVisitor(ExpressionContainsParameterVisitor.Instance, SearchParameterNames.ResourceType)))
             {
-                StringBuilder.AppendLine().Append("INNER JOIN ").AppendLine(TableExpressionName(_tableExpressionCounter));
-                StringBuilder.Append("ON ").Append(VLatest.Resource.ResourceSurrogateId, resourceTableAlias).Append(" = ").Append(TableExpressionName(_tableExpressionCounter)).AppendLine(".Sid1");
-            }
-            else if (!_isHistorySearch)
-            {
-                // If not joining with any other CTEs and this is not a history search, make sure the optimizer does not decide to do a scan on the clustered index
+                // If this is a simple search over a resource type (like GET /Observation)
+                // make sure the optimizer does not decide to do a scan on the clustered index, since we have an index specifically for this common case
                 StringBuilder.Append(" WITH(INDEX(").Append(VLatest.Resource.IX_Resource_ResourceTypeId_ResourceSurrgateId).AppendLine("))");
             }
             else
             {
-                // History search will likely resort to a table scan until we add ResourceTypeId as the leading column to the clustered index Resource table.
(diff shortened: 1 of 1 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -131,13 +131,23 @@ public object VisitSqlRoot(SqlRootExpression expression, SearchOptions context)
...
             if (expression.TableExpressions.Count > 0)
             {
-                StringBuilder.Append("INNER JOIN ").AppendLine(TableExpressionName(_tableExpressionCounter));
+                StringBuilder.AppendLine().Append("INNER JOIN ").AppendLine(TableExpressionName(_tableExpressionCounter));
                 StringBuilder.Append("ON ").Append(VLatest.Resource.ResourceSurrogateId, resourceTableAlias).Append(" = ").Append(TableExpressionName(_tableExpressionCounter)).AppendLine(".Sid1");
             }
+            else if (!_isHistorySearch)
+            {
+                // If not joining with any other CTEs and this is not a history search, make sure the optimizer does not decide to do a scan on the clustered index
+                StringBuilder.AppendLine(" WITH(INDEX(IX_Resource_ResourceTypeId_ResourceSurrgateId))");
```
johnstairs, line 144: Waiting on https://github.com/microsoft/healthcare-shared-components/pull/128 to avoid magic strings here

Ivanidzo4ka, line 144: >IX_Resource_ResourceTypeId_ResourceSurrgateId [](start = 54, length = 45)

This feels like shooting bazooka to kill a mosquito... There should be a reason why SQL server pick clustered index over this one. Any chance we can have access to server on which this occurs or emulate that? I would love to see statistics and index fragmentation. Maybe that index is so fragmented what it doesn't yield much different time in execution.

johnstairs: You can repro it on a small database too. My hypothesis is that it deems the clustered index preferable so that it does not have to do a key lookup to retrieve the raw resource column and others.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -131,13 +131,23 @@ public object VisitSqlRoot(SqlRootExpression expression, SearchOptions context)
...
 
             if (expression.TableExpressions.Count > 0)
             {
-                StringBuilder.Append("INNER JOIN ").AppendLine(TableExpressionName(_tableExpressionCounter));
+                StringBuilder.AppendLine().Append("INNER JOIN ").AppendLine(TableExpressionName(_tableExpressionCounter));
                 StringBuilder.Append("ON ").Append(VLatest.Resource.ResourceSurrogateId, resourceTableAlias).Append(" = ").Append(TableExpressionName(_tableExpressionCounter)).AppendLine(".Sid1");
             }
+            else if (!_isHistorySearch)
+            {
+                // If not joining with any other CTEs and this is not a history search, make sure the optimizer does not decide to do a scan on the clustered index
```
johnstairs, line 143: We also don't necessarily want this if it's a system-wide search, unless we change the default sorting to `ResourceTypeId, ResourceSurrogateId`

feordin (approved):
I tested it on my DB with 10000+ practitioner resources and observed the change in the query plan.

PR #1465, opened 2020-11-18
Create Register-Resource-Application.md

CaitlinV39:
## Description
Moving the resource application information to OSS Repo from Azure Docs

## Related issues
Addresses https://github.com/MicrosoftDocs/azure-docs/issues/66056


CaitlinV39:
@johnstairs I made the changes to the resource application here and I'm going to remove it and reference this file on the Azure Docs site. Let me know if you see any issues with this one and then once this is live I'll fix the Azure Docs site.

brendankowitz (approved):

johnstairs:
Sorry am late to this. The roles.json file is not what does inside the application manifest file. Rather each of the roles in the manifest should correspond to a role in the roles.json file. Also, the portal now has UI support for add app roles instead of editing the manifest:
image
In this UI, the `Value` should correspond to the `name` property of the role in roles.json

PR #1471, opened 2020-11-19
Adds support for `first()` in SearchParameterToTypeResolver

brendankowitz:
## Description
Adds support for `first()` function when resolving supported types in SearchParameterToTypeResolver

## Testing
Adds unit test


brendankowitz:
@feordin while testing reindexing with this search parameter, it fails because `first()` is not supported when determining if we have the converter

johnstairs (approved):

Issue #891, opened 2020-02-19
Add support for FHIR searches of types Money => Quantity


brendankowitz:
**User story**
Add support for the R4 primitive type [Money](https://www.hl7.org/fhir/datatypes.html).

**Acceptance criteria**
1. Support is visible at the /metadata endpoint
1. Canonical urls are indexed correctly
1. Search mapped from Money=>Quantity. e.g. Account.Balance

CaitlinV39:
AB#72427

PR #1475, opened 2020-11-20
Adds support for money to quantity searches

rotodd:
## Description

Adds support for search parameters that are converted from [Money](https://www.hl7.org/fhir/datatypes.html#Money) values to Quantity search values.

There are a lot of duplicate files in this PR!

One reason for this is that we still have old converters (which can be turned on by changing [this setting](https://github.com/microsoft/fhir-server/pull/1304/files) to `false`). Once the old converters are removed, we will no longer need these files:

`src/Microsoft.Health.Fhir.R4.Core/Features/Search/Converters/MoneyToQuantitySearchValueTypeConverter.cs`
`src/Microsoft.Health.Fhir.R5.Core/Features/Search/Converters/MoneyToQuantitySearchValueTypeConverter.cs`
`src/Microsoft.Health.Fhir.Stu3.Core/Features/Search/Converters/MoneyToQuantitySearchValueTypeConverter.cs`

`src/Microsoft.Health.Fhir.R4.Core.UnitTests/Features/Search/Converters/MoneyToQuantitySearchValueTypeConverterTests.cs`
`src/Microsoft.Health.Fhir.R5.Core.UnitTests/Features/Search/Converters/MoneyToQuantitySearchValueTypeConverterTests.cs`
`src/Microsoft.Health.Fhir.Stu3.Core.UnitTests/Features/Search/Converters/MoneyToQuantitySearchValueTypeConverterTests.cs`

Another reason is that the `Money` data type is defined differently for STU3 (see [spec](https://www.hl7.org/fhir/STU3/datatypes.html#money)). If/when we choose to remove `STU3` support, we could remove this file:

`src/Microsoft.Health.Fhir.Stu3.Core.UnitTests/Features/Search/Converters/NodeConverterTests/MoneyNodeToQuantitySearchValueTypeConverterTests.cs`

And replace these two files with one file in the Shared project:

`src/Microsoft.Health.Fhir.R4.Core.UnitTests/Features/Search/Converters/NodeConverterTests/MoneyNodeToQuantitySearchValueTypeConverterTests.cs`
`src/Microsoft.Health.Fhir.R4.Core.UnitTests/Features/Search/Converters/NodeConverterTests/MoneyNodeToQuantitySearchValueTypeConverterTests.cs`

## Demo

totalnet

## Related issues
Addresses [#AB72427](https://microsofthealth.visualstudio.com/Health/_workitems/edit/72427) and #891.

## Testing
Added unit tests.


src/Microsoft.Health.Fhir.Core/Features/Search/Converters/MoneyNodeToQuantitySearchValueTypeConverter.cs
```diff
@@ -0,0 +1,61 @@
...
+
+                yield return new QuantitySearchValue(
+                    system,
+                    code,
+                    decimalValue.GetValueOrDefault());
+            }
+            else
+            {
+                yield return new QuantitySearchValue(
+                    CurrencyValues.System,
```
rotodd, line 55: @brendankowitz I noticed that in `CodeNodeToTokenSearchValueTypeConverter.cs`, `system` is pulled from the `codeSystemResolver`. That doesn't seem to have "urn:iso:std:iso:4217", though. Is there a nicer way to pull the system info from the POCO instead of creating a new class?

src/Microsoft.Health.Fhir.Stu3.Core.UnitTests/Features/Search/Converters/NodeConverterTests/MoneyNodeToQuantitySearchValueTypeConverterTests.cs
```diff
@@ -0,0 +1,71 @@
...
+            const decimal value = 0.125m;
+            const string code = "USD";
+
+            await Test(
+                m =>
+                {
+                    m.Value = value;
+                    m.Code = code;
+                    m.System = null;
+                },
```
rotodd, line 66: `MoneyNodeToQuantitySearchValueTypeConverter` is independent of the `Money` type, but its tests are not - is this expected? Is there a way around creating a separate test file for `STU3` here?

src/Microsoft.Health.Fhir.Core/Features/Search/Converters/MoneyNodeToQuantitySearchValueTypeConverter.cs
brendankowitz: These Converters don't use the POCOs. I think we should be aiming to have code that operates more generically but can pull in the metadata that it needs

src/Microsoft.Health.Fhir.Stu3.Core.UnitTests/Features/Search/Converters/NodeConverterTests/MoneyNodeToQuantitySearchValueTypeConverterTests.cs
LTA-Thinking: Something you could do is make this class abstract and leave the TTypeConverter as a generic, then have two empty classes that just specify the generic. Personally I think in a use case like that it would be ok to have all three classes in one file as two of the classes are empty.

src/Microsoft.Health.Fhir.ValueSets/CurrencyValues.cs
```diff
@@ -0,0 +1,12 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Fhir.ValueSets
+{
+    public static class CurrencyValues
```
LTA-Thinking, line 8: I'm wondering if these names should be flipped so that this file could be reused if needed. Something like the class being SystemValues and the const being Currency. As it is I don't know what we could use this for outside this PR.

brendankowitz: There are a lot of valuesets http://www.hl7.org/implement/standards/fhir/terminologies-valuesets.html.
The difference here is that most things in assembly are defining the ValueSet values, whereas this is the valueset system

brendankowitz: I think I was thrown with `SystemValues` :P There is the valueset system and the valueset values. SystemValues made me think twice

rotodd: Good point! I originally decided on this way to parallel the implementation of [SpecialValues](https://github.com/microsoft/fhir-server/blob/master/src/Microsoft.Health.Fhir.ValueSets/SpecialValues.cs). I think you're right though - it would be nice to put these in one file.

I'll wait to see what feedback I get on [this comment](https://github.com/microsoft/fhir-server/pull/1475#discussion_r527322562) before committing a change (I want to confirm we actually need this hard coded string).

src/Microsoft.Health.Fhir.Core/Features/Search/Converters/MoneyNodeToQuantitySearchValueTypeConverter.cs
rotodd: Oops, sorry. By POCO, I meant the `ITypedElement` value that is passed into the `MoneyNodeToQuantitySearchValueTypeConverter`.

While debugging, I can see that "urn:iso:std:iso:4217" gets populated in the `ITypedElement` object. I'm not sure if this is just debugger magic or if there is a way to pull out the value.

src/Microsoft.Health.Fhir.Stu3.Core.UnitTests/Features/Search/Converters/NodeConverterTests/MoneyNodeToQuantitySearchValueTypeConverterTests.cs
rotodd: Awesome, thanks for the rec. Looking into this now.

src/Microsoft.Health.Fhir.ValueSets/CurrencyValues.cs
rotodd: @brendankowitz sorry, I didn't see your comment before adding my reply (GitHub has been weird today). By `SystemValues` do you mean `SpecialValues.System`?

src/Microsoft.Health.Fhir.Core/Features/Search/Converters/MoneyNodeToQuantitySearchValueTypeConverter.cs
rotodd: As discussed in scrum, Brendan will check with Gino to see if there is a way to pull the "urn:iso:std:iso:4217" system value from FHIR libraries (@brendankowitz please correct me if I misinterpreted this).

brendankowitz: A bit more context: The value you see there, is from the Poco objects, however, it is also possible to create an ITypedElement using JsonNode, which will not have these values embedded.

I was mainly referring to Gino's codegen as a better way of creating the codesystem mappings over the powershell that currently does it. I think that those values should be there, and if they are not, then I'm interested to find out why not.

src/Microsoft.Health.Fhir.ValueSets/CurrencyValues.cs
brendankowitz: Perhaps something like CurrencyValueSet.System is fine.

rotodd: Good idea, thanks. Renamed in 02c118d.

Summarizing our earlier discussion: the `ValueSet` uri is "http://hl7.org/fhir/ValueSet/currencies" and the `CodeSystem` is "urn:iso:std:iso:4217", so it would be good to avoid "SystemValues" ([here](https://www.hl7.org/fhir/valueset-currencies.html) is the relevant part of the spec).

src/Microsoft.Health.Fhir.Core/Features/Search/Converters/MoneyNodeToQuantitySearchValueTypeConverter.cs
rotodd: Thanks, that is helpful 👍 

As discussed, it doesn't make sense to wait for the switch to codegen to get this change in. I added a comment in 4827b96.

src/Microsoft.Health.Fhir.Stu3.Core.UnitTests/Features/Search/Converters/NodeConverterTests/MoneyNodeToQuantitySearchValueTypeConverterTests.cs
rotodd: Making the change to an abstract class with a generic type doesn't end up reducing duplicate code, since `Money` has different properties in `Stu3` (all the tests would end up needing to have their version-specific implementation of the abstract class).

src/Microsoft.Health.Fhir.Core/Features/Search/Converters/MoneyNodeToQuantitySearchValueTypeConverter.cs
```diff
@@ -0,0 +1,61 @@
...
+            var decimalValue = (decimal?)value.Scalar("value");
+
+            if (decimalValue == null)
+            {
+                yield break;
+            }
+
+            var currency = value.Scalar("currency")?.ToString();
+
+            // Currency information is specified differently if we are running STU3.
```
rotodd, line 36: Currently looking into adding an `#if Stu3` conditional compile constant here - I think it might add a bit of clarity.

brendankowitz: You can't add it in Core, it's only available in Shared (these projects compile into FHIR specific assemblies). In core you can use ModelInfoProvider.Version

rotodd: Ah, thank you. Added in 0ed9dcc

src/Microsoft.Health.Fhir.Stu3.Core.UnitTests/Features/Search/Converters/NodeConverterTests/MoneyNodeToQuantitySearchValueTypeConverterTests.cs
LTA-Thinking: Sorry, I was unclear. I meant you could combine the MoneyNode and Money test files into an abstract class and two child classes. There would still be different versions for R4, R5, and Stu3 but only one file for each instead of two.

LTA-Thinking (approved):

brendankowitz (approved):

src/Microsoft.Health.Fhir.Stu3.Core.UnitTests/Features/Search/Converters/NodeConverterTests/MoneyNodeToQuantitySearchValueTypeConverterTests.cs
rotodd: As discussed, we will keep the duplicate converter tests as is for now, simply because we should hopefully be removing the old converters soon

PR #1476, opened 2020-11-20
Improve export task telemetry

tongwu-msft:
## Description
Improve export operation telemetry
- Add export task status & committed data size in logging message
- Add Notification after export task completed.

## Related issues
Addresses issue #77082.

## Testing
We updated anonymized export and basic export tests with notification verification.

### DIFF
--- a/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Export/ExportJobTaskTests.cs
+++ b/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Export/ExportJobTaskTests.cs
@@ -12,6 +12,7 @@ using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Hl7.Fhir.ElementModel;
+using MediatR;
 using Microsoft.Extensions.Logging.Abstractions;
 using Microsoft.Extensions.Options;
 using Microsoft.Health.Core.Internal;
@@ -72,6 +73,7 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Operations.Export
                 _inMemoryDestinationClient,
                 _resourceDeserializer,
                 null,
+                Substitute.For<IMediator>(),
                 NullLogger<ExportJobTask>.Instance);
         }
 
@@ -496,6 +498,7 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Operations.Export
                 mockExportDestinationClient,
                 _resourceDeserializer,
                 null,
+                Substitute.For<IMediator>(),
                 NullLogger<ExportJobTask>.Instance);
 
             await exportJobTask.ExecuteAsync(_exportJobRecord, _weakETag, _cancellationToken);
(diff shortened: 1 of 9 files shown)


src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Export/ExportJobTaskTests.cs
```diff
@@ -12,6 +12,7 @@
 using System.Threading;
 using System.Threading.Tasks;
 using Hl7.Fhir.ElementModel;
+using MediatR;
```
LTA-Thinking, line 15: Can you add an assert to at least one of these tests that the Mediator substitute is being called. I saw the tests in the anonymizer test file, but since this isn't something specifically for anonymizer I would like to see it test here too.

skywalker-wu: Thanks, updated.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -489,6 +497,18 @@ private bool IsAnonymizedExportJob()
             return !string.IsNullOrEmpty(_exportJobRecord.AnonymizationConfigurationLocation);
         }
 
+        private string ExtractExportTaskLoggingMessage()
+        {
+            string id = _exportJobRecord?.Id ?? string.Empty;
```
namadabu, line 502: nit: Are the null checks for the `_exportJobRecord` necessary since it will never be null.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportTaskMetricsNotification.cs
```diff
@@ -0,0 +1,28 @@
...
+            ResourceType = null;
+
+            ExportJobRecord = exportJobRecord;
+        }
+
+        public string FhirOperation { get; }
+
+        public string ResourceType { get; }
+
+        public ExportJobRecord ExportJobRecord { get; }
```
namadabu, line 26: Do we really want to publish the entire `ExportJobRecord` object as part of the notification? It contains a lot of information that are completely unnecessary for metrics. I think it might be better to include only the required information.

skywalker-wu: Thanks, I updated to use only few metrics from record.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
skywalker-wu: Remove check on the record. Thanks!

namadabu (approved):

LTA-Thinking (approved):

PR #1477, opened 2020-11-20
Cleanup old images

LTA-Thinking:
## Description
Adds a job to cleanup old docker images. The script breaks images into three categories: PR, CI, and Release. PR images are deleted after 1 week. CI images are deleted after 60 days. Release images are never deleted. 

The script works by removing the tags of images that are out of date, and then deleting any images that have no tags. This prevents the issue of an image being in multiple categories being deleted improperly. This was a concern as Release images are just retagged CI images.  

## Related issues
Addresses [User Story 73115](https://microsofthealth.visualstudio.com/Health/_workitems/edit/73115)

## Testing
The script was run manually with a modification to print a file of images that would have been affected instead of deleting them.


build/cleanup-images.yml
```diff
@@ -0,0 +1,55 @@
...
+      scriptLocation: InlineScript
+      inlineScript: |
+        $versions = "stu3","r4","r5"
+        $tagList = az acr repository show-tags --name $(azureContainerRegistryName) --repository r4_fhir-server
+
+        foreach ($tag in $tagList) {
+          $tagParts = $tag.Split('-')
+          if ($tagParts.Count -gt 1) {
+            try {
+              $imageDate = [datetime]($tagParts[1].Substring(0,4) + '/' + $tagParts[1].Substring(4,2) + '/' + $tagParts[1].Substring(6,2))
```
brendankowitz, line 30: Is there another date we could use? Ideally I'd eventually want to switch to semver for tags too

build/cleanup-images.yml
```diff
@@ -0,0 +1,55 @@
...
+  steps:
+  - task: AzureCLI@2
+    displayName: 'Delete Images'
+    imputs:
+      azureSubscription: $(ConnectedServiceName)
+      scriptType: ps
+      scriptLocation: InlineScript
+      inlineScript: |
+        $versions = "stu3","r4","r5"
+        $tagList = az acr repository show-tags --name $(azureContainerRegistryName) --repository r4_fhir-server
```
brendankowitz, line 24: Is this supposed to be `r4_fhir-server`?

build/jobs/cleanup.yml
```diff
@@ -54,3 +54,19 @@ jobs:
...
+    vmImage: $(WindowsVmImage)
+  steps:
+  - task: AzureCLI@2
+    displayName: 'Delete Image'
+    imputs:
+      azureSubscription: $(ConnectedServiceName)
+      scriptType: ps
+      scriptLocation: InlineScript
+      inlineScript: |
+        az acr repository delete --name $(azureContainerRegistryName) --image 'r4_fhir-server:$(ImageTag)'
```
brendankowitz, line 70: So we just remove the image at the end of the build, makes sense

build/cleanup-images.yml
LTA-Thinking: We can look up the image's last updated date, it just takes a few more steps.

build/cleanup-images.yml
LTA-Thinking: Yes. Since each repo has the same images sent to them I was using one of them as the source of truth for the tags to avoid redundant requests.

build/jobs/cleanup.yml
LTA-Thinking: I added this to help with clean up, but it isn't perfect as it only runs if the build succeeds. The other script will clean up any from failed builds.

LTA-Thinking:
I really don't get how Devops picks the pipelines to run. It ran the new cleanup script instead of the pr-pipeline script. Thankfully I had a typo that stopped it from running, which I will leave in until I figure out how have it not run the cleanup script.

build/cleanup-images.yml
LTA-Thinking: Changed it to look up the list of tags for each version separately.

build/cleanup-images.yml
LTA-Thinking: Changed to use the last updated date.

feordin (approved):

Issue #1384, opened 2020-10-22
_elements and _summary should return Bad Request


CaitlinV39:
**Describe the bug**
In OSS when making a search request with both the _elements and _summary request an internal server error is returned as FHIR doesn't support this combination. This should return a bad request.

**FHIR Version?**
Stu3/R4/R5

**Data provider?**
CosmosDB/SQL Server

**To Reproduce**
Make a search request with both _elements & _summary

**Expected behavior**
This should return a bad request.

**Actual behavior**
an internal server error is returned

CaitlinV39:
AB#76836

PR #1480, opened 2020-11-21
Returns Bad Request when a request has both _elements and _summary parameters

rotodd:
## Description
Prior to this fix, when making a search request with both the `_elements` and `_summary` parameters, an internal server error was returned. Now, an operation outcome is returned specifying a bad request and an error message.

image

## Related issues
Addresses [#AB76836](https://microsofthealth.visualstudio.com/Health/_workitems/edit/76836) and #1384.

## Testing
Added two E2E tests to cover this case.

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Resources.resx
+++ b/src/Microsoft.Health.Fhir.Core/Resources.resx
@@ -461,4 +461,7 @@
     <value>The '_type' parameter contains unknown resource(s): {0}</value>
     <comment>{0}: Single or list of unknown resources.</comment>
   </data>
+  <data name="IncompatibleSearchParameters" xml:space="preserve">
+    <value>The parameters '{0}' and '{1}' are not compatible.</value>
+  </data>
 </root>
--- a/src/Microsoft.Health.Fhir.Core/Resources.resx
+++ b/src/Microsoft.Health.Fhir.Core/Resources.resx
@@ -461,7 +461,7 @@
     <value>The '_type' parameter contains unknown resource(s): {0}</value>
     <comment>{0}: Single or list of unknown resources.</comment>
   </data>
-  <data name="IncompatibleSearchParameters" xml:space="preserve">
-    <value>The parameters '{0}' and '{1}' are not compatible.</value>
+  <data name="ElementsAndSummaryParametersAreIncompatible" xml:space="preserve">
+    <value>The '_elements' parameter is supported only when '_summary' is SummaryType.False or '_summary' is not specified at all.</value>
   </data>
 </root>
--- a/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/BasicSearchTests.cs
+++ b/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/BasicSearchTests.cs
@@ -8,11 +8,13 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 using System.Net;
+using System.Threading.Tasks;
 using Hl7.Fhir.ElementModel;
 using Hl7.Fhir.FhirPath;
 using Hl7.Fhir.Model;
 using Microsoft.Health.Fhir.Client;
 using Microsoft.Health.Fhir.Core.Extensions;
+using Microsoft.Health.Fhir.Core.Features;
 using Microsoft.Health.Fhir.Core.Features.Search;
 using Microsoft.Health.Fhir.Core.Models;
 using Microsoft.Health.Fhir.Tests.Common;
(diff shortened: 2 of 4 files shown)


test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/BasicSearchTests.cs
```diff
@@ -425,22 +427,10 @@ public async Task GivenResources_WhenSearchedWithIncorrectSummaryParams_ThenExce
...
-            var patients = new Patient[numberOfResources];
-
-            for (int i = 0; i < numberOfResources; i++)
-            {
-                patient.Meta = new Meta();
-                patient.Meta.Tag.Add(tag);
-
-                FhirResponse<Patient> createdPatient = await Client.CreateAsync(patient);
-                patients[i] = MaskingNode.ForElements(new ScopedNode(createdPatient.Resource.ToTypedElement()), elements).ToPoco<Patient>();
-            }
```
rotodd, line 443: Moved common code to a shared method: `CreatePatientsWithSpecifiedElements()`

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/BasicSearchTests.cs
```diff
@@ -476,24 +466,44 @@ public async Task GivenListOfResources_WhenSearchedWithEmptyElements_ThenAllProp
...
+            Assert.Equal(expectedStatusCode, ex.StatusCode);
+
+            var expectedErrorMessage = $"{expectedStatusCode}: " + string.Format(Core.Resources.IncompatibleSearchParameters, KnownQueryParameterNames.Summary, KnownQueryParameterNames.Elements);
+
+            Assert.Equal(expectedErrorMessage, ex.Message);
+        }
+
+        [Fact]
+        [Trait(Traits.Priority, Priority.One)]
+        public async Task GivenListOfResources_WhenSearchedWithElementsAndSummarySetToFalse_ThenOnlySpecifiedPropertiesShouldBeReturned()
```
rotodd, line 499: The `_summary` parameter could still be included in a request that has the `_elements` parameter, as long as it is set to `false`.

src/Microsoft.Health.Fhir.Core/Resources.resx
```diff
@@ -461,4 +461,7 @@
     <value>The '_type' parameter contains unknown resource(s): {0}</value>
     <comment>{0}: Single or list of unknown resources.</comment>
   </data>
+  <data name="IncompatibleSearchParameters" xml:space="preserve">
+    <value>The parameters '{0}' and '{1}' are not compatible.</value>
+  </data>
```
rotodd, line 466: Open to different wording - is it worth making it clear that `_summary=false` is permitted?

Ivanidzo4ka: [https://github.com/FirelyTeam/firely-net-sdk/blob/03c257328631716fb335b634396d711e0b0e22eb/src/Hl7.Fhir.Core/Serialization/BaseFhirSerializer.cs#L30](https://github.com/FirelyTeam/firely-net-sdk/blob/03c257328631716fb335b634396d711e0b0e22eb/src/Hl7.Fhir.Core/Serialization/BaseFhirSerializer.cs#L30)

Is there are reason to not just resurface that exception to user?

---
In reply to: [528067067](https://github.com/microsoft/fhir-server/pull/1480#discussion_r528067067) [](ancestors = 528067067)

rotodd: Good question. I originally decided to throw the exception here instead of using this existing exception message thrown by `_fhirJsonSerializer.Serialize()` in `FhirXmlOutputFormatter.cs` or `FhirJsonOutputFormatter.cs` because the `BadRequestException` is automatically wrapped in an `OperationOutcome` if I throw it from `SearchOptionsFactory.cs`.

That being said, I'm not sure if it would actually be ugly to manually add my own `OperationOutcome` in the formatters. Looking into this a bit more now... thanks!

Ivanidzo4ka: On a second thought, it's probably better to wrap this in our exception which would be localizable (hope that's a word). But we can use they wording if you like how they formulate it.

---
In reply to: [528436000](https://github.com/microsoft/fhir-server/pull/1480#discussion_r528436000) [](ancestors = 528436000)

rotodd: Updated the error message in 514a89e.

Ivanidzo4ka (approved):
:shipit:

Issue #1448, opened 2020-11-13
Migrate to .NET 5


brendankowitz:
**User story**
Migrate codebase to use .NET 5

- https://docs.microsoft.com/en-us/aspnet/core/migration/31-to-50?view=aspnetcore-5.0&tabs=visual-studio
- https://docs.microsoft.com/en-us/dotnet/core/compatibility/3.1-5.0

CaitlinV39:
AB#77730

brendankowitz:
Requires support from https://github.com/microsoft/healthcare-shared-components/pull/133

chgl:
Do you have/plan on adding performance benchmarks for the server? A before/after comparison for this migration would be neat.

CaitlinV39:
This has been completed so closing this issue.

PR #1487, opened 2020-11-26
Moving to .net 5

Ivanidzo4ka:
## Description
Moves projects to .net 5 framework. (it's still uses C#8 by default, not sure why)

## Related issues
Addresses [#1448].


Ivanidzo4ka:
Microsoft.Health.Extensions.BuildTimeCodeGenerator nuget package from https://github.com/microsoft/healthcare-shared-components requires .netcore 3.1 sdk to be present on machine. (it basically run publish command on dll to make it selfcontained resource and push it inside nuget, and nuget has trigger to be run during compilation https://github.com/microsoft/healthcare-shared-components/blob/b0736351fc578bebbdd5f45d3f5ddc2a93326104/tools/Microsoft.Health.Extensions.BuildTimeCodeGenerator/Microsoft.Health.Extensions.BuildTimeCodeGenerator.targets#L6 in which path with sdk is hardcoded.)

So it basically asks for two SDKs to be present 5.0 and 3.1, which is doable for PR Build & Deploy, but not the case for LGTM (which is solely rely on global.json (AFAIK).

Easiest way is to move health-shared-components to 5.0, but that would require lot of coordination between teams.

Considering what we work with generator through dotnet CLI I'm somewhat doubt about need for generator to be published, maybe if it was proper dll and not pushed to tools folder it would allow it to work on multiple platforms. (or at least it would allow us to create normal nuget which support multiple platforms and targets file could use $TargetFramework property instead of hardcoding 3.1.

Anyway, all roads leads to ~~Rome~~ health-shared-component repo which need to be updated first before anything can be done here.

johnstairs:
Yes, we should move the shared components to .net 5. Also the codegen stuff should probably be ported to source generators in the future.

PR #1491, opened 2020-12-01
Codegen for table types in separate file

johnstairs:
Updates shared components package to include https://github.com/microsoft/healthcare-shared-components/pull/130, where the generated types for table types are no longer nested within `VLastest`.


LTA-Thinking:
Why do we check in the generated files? I've had to delete them locally several times because of branch checkout issues and they are regenerated automatically as part of the build.

johnstairs:
> Why do we check in the generated files? I've had to delete them locally several times because of branch checkout issues and they are regenerated automatically as part of the build.

@LTA-Thinking, the idea was that it's easier to understand the effect changes have to the generated code (either the code that generates the code or its input).
Since the code generator is not in this repo, maybe it's less relevant here (I would keep the codegen part of of the shared components repo), but we could remove it from the repo here, if people would prefer that.

johnstairs:
> > Why do we check in the generated files? I've had to delete them locally several times because of branch checkout issues and they are regenerated automatically as part of the build.
> 
> @LTA-Thinking, the idea was that it's easier to understand the effect changes have to the generated code (either the code that generates the code or its input).
> Since the code generator is not in this repo, maybe it's less relevant here (I would keep the codegen part of of the shared components repo), but we could remove it from the repo here, if people would prefer that.

@LTA-Thinking, I've now removed the generated files from the repo!

src/Microsoft.Health.Fhir.SqlServer/Microsoft.Health.Fhir.SqlServer.csproj
```diff
@@ -20,18 +20,23 @@
...
+      <Generator>ImmutableSqlModelGenerator</Generator>
+      <Namespace>Microsoft.Health.Fhir.SqlServer.Features.Schema.Model</Namespace>
+      <Args>"$([System.IO.Path]::GetFullPath('$(MSBuildThisFileDirectory)\Features\Schema\Migrations\6.sql'))"</Args>
+    </Generated>
     <Generated Include="Features\Schema\Model\VLatest.Generated.cs">
-      <Generator>SqlModelGenerator</Generator>
+      <Generator>MutableSqlModelGenerator</Generator>
       <Namespace>Microsoft.Health.Fhir.SqlServer.Features.Schema.Model</Namespace>
-      <Args>$([System.IO.Path]::GetFullPath('$(MSBuildThisFileDirectory)\Features\Schema\Migrations\6.sql'))</Args>
+      <Args>"$([System.IO.Path]::GetFullPath('$(MSBuildThisFileDirectory)\Features\Schema\Migrations\6.sql'))"</Args>
```
brendankowitz, line 31: Would it make sense to now create this path in an msbuild variable to keep these in sync?

brendankowitz (approved):

src/Microsoft.Health.Fhir.SqlServer/Microsoft.Health.Fhir.SqlServer.csproj
johnstairs: Reduced a lot of the duplication [here](https://github.com/microsoft/fhir-server/pull/1491/commits/c5f7c2b18b7973e8861c48ea46988b810f210316)

PR #1495, opened 2020-12-03
Prevent SmartLauncher sample being packed

brendankowitz:
## Description
This project doesn't need to be packaged as a nuget

### DIFF
--- a/samples/apps/SmartLauncher/SmartLauncher.csproj
+++ b/samples/apps/SmartLauncher/SmartLauncher.csproj
@@ -4,6 +4,7 @@
     <TargetFramework>netcoreapp3.1</TargetFramework>
     <GenerateEmbeddedFilesManifest>true</GenerateEmbeddedFilesManifest>
     <UserSecretsId>SmartLauncher_3d9536f4-982c-4334-83df-890fb8ae70e6</UserSecretsId>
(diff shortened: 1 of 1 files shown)


samples/apps/SmartLauncher/SmartLauncher.csproj
```diff
@@ -4,6 +4,8 @@
     <TargetFramework>netcoreapp3.1</TargetFramework>
     <GenerateEmbeddedFilesManifest>true</GenerateEmbeddedFilesManifest>
     <UserSecretsId>SmartLauncher_3d9536f4-982c-4334-83df-890fb8ae70e6</UserSecretsId>
+    <IsPackable>false</IsPackable>
+    <PackageId>Microsoft.Internal.SmartLauncher</PackageId>
```
Ivanidzo4ka, line 8: If we don't want to pack it, what's the point of specifying package name?

Ivanidzo4ka (approved):
:shipit:

samples/apps/SmartLauncher/SmartLauncher.csproj
brendankowitz: True, if packed it should have *.internal.* but I don't think it needs to be packed at all. I'll confirm what happens in the build output and update

Issue #1458, opened 2020-11-17
500 on Conditional Update


CaitlinV39:
**Describe the bug**
A customer reported that they are getting a 500 on a conditional update. 

**FHIR Version?**
Stu3

**Data provider?**
SQL Server

**To Reproduce**

The first PUT returns a 201 but subsequent PUTS are returning a 500

**Expected behavior**
Expect for the conditional update to work, or at least a better error describing the issue

**Actual behavior**
Receive a 500

CaitlinV39:
@brendankowitz & @feordin FYI

CaitlinV39:
[stacktrace.txt](https://github.com/microsoft/fhir-server/files/5555926/stacktrace.txt)

CaitlinV39:
[task_example.txt](https://github.com/microsoft/fhir-server/files/5555940/task_example.txt)

This is a non-PHI example of the task

CaitlinV39:
@johnstairs I think this was resolved in your PR. Can I close?

PR #1497, opened 2020-12-04
Handle references with missing type

johnstairs:
## Description
This change fixes an issue with the SQL data provider where we would null-ref when handling a request like
```
POST https://localhost:44348/Patient
Content-Type: application/json

{
    "resourceType": "Patient",
    "managingOrganization": {
        "reference": "abc"
    }
}
```

where the type of the `managingOrganization` reference is missing.

This requires a change to the schema because a table column needs to be made nullable.

In addition, in cases where the a reference type can only be a single type - like in the above example, which can only be of type `Organization` - we automatically add the type to the search parameter index in the DB.

Finally, for the SQL provider, for searches like `Patient?organization=abc`, turn this into `Patient?organization=Organization/abc`.

## Related issues
Addresses #1458


src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterDefinitionBuilder.cs
```diff
@@ -244,6 +246,14 @@ private List<(string ResourceType, SearchParameterInfo SearchParameter)> Validat
...
                             continue;
                         }
+
+                        resolvedComponents.Add(componentSearchParameter);
+                    }
+
+                    if (resolvedComponents.Count > 0)
+                    {
+                        // this is not actually creating the search parameter, it will return the existing one.
+                        CreateSearchParameterInfo(searchParameter).ResolvedComponents = resolvedComponents;
```
Ivanidzo4ka, line 256: >CreateSearchParameterInfo [](start = 24, length = 25)

nit: this reads like some kind of black magic. Can we rename to `GetOrCreateSearchParameterInfo`

Ivanidzo4ka: or just do `_urlDictionary.GetValue(searchParameter.Uri).ResolvedComponents = resolvedComponents`


---
In reply to: [538846387](https://github.com/microsoft/fhir-server/pull/1497#discussion_r538846387) [](ancestors = 538846387)

src/Microsoft.Health.Fhir.Core/Models/SearchParameterInfo.cs
```diff
@@ -95,5 +95,7 @@ public SearchParameterInfo(string name)
         public bool IsPartiallySupported { get; set; }
 
         public IReadOnlyList<SearchParameterComponentInfo> Component { get; }
+
+        public IReadOnlyList<SearchParameterInfo> ResolvedComponents { get; set; } = Array.Empty<SearchParameterInfo>();
```
Ivanidzo4ka, line 99: Can we add comments to this properites?
<summary>
All possible components of composite search parameters. Should be present only if Type = SearchParamType.Composite
</summary>
 public IReadOnlyList<SearchParameterComponentInfo> Component { get; }

<summary>
Populated by query composite search parameters
</summary>
public IReadOnlyList<SearchParameterInfo> ResolvedComponents { get; set; } = Array.Empty<SearchParameterInfo>();

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/UntypedReferenceRewriter.cs
```diff
@@ -0,0 +1,108 @@
...
+                // nothing to do for this expression
+                return expression;
+            }
+
+            // now see which components have a expression on the reference type
+            int componentsPresent = expression.Expression.AcceptVisitor(ParameterPredicateVisitor.Instance, null);
+
+            // now determine which ones we get a type expression
+
+            int componentsToFill = componentCandidates & ~componentsPresent;
```
Ivanidzo4ka, line 51: >componentCandidates & ~componentsPresent [](start = 35, length = 40)

I'm not afraid to say: "It's hard for me to understand bitwise operations"
Can we add comment about what is going on here?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/UntypedReferenceRewriter.cs
```diff
@@ -0,0 +1,108 @@
...
+    /// of only one type, rewrites the expression to specify the single target type.
+    /// </summary>
+    internal class UntypedReferenceRewriter : ExpressionRewriterWithInitialContext<object>
+    {
+        public static readonly UntypedReferenceRewriter Instance = new UntypedReferenceRewriter();
+
+        public override Expression VisitSearchParameter(SearchParameterExpression expression, object context)
+        {
+            // Handle reference search parameters as well as composite search parameters with one or more reference components.
+            // We fist create a bitmask with bits set representing the component indexes that are candidates for this rule.
```
Ivanidzo4ka, line 25: >bitmask with bits set representing the component indexes that are candidates for this rule [](start = 32, length = 90)

Any chance I can persuade choice of BitArray over int with bit operations?
I understand BitArray would bring performance downgrade, but there is chance it would be easier to read...

src/Microsoft.Health.Fhir.SqlServer/Microsoft.Health.Fhir.SqlServer.csproj
```diff
@@ -11,9 +11,13 @@
     <EmbeddedResource Include="Features\Schema\Migrations\5.sql" />
     <EmbeddedResource Include="Features\Schema\Migrations\6.diff.sql" />
     <EmbeddedResource Include="Features\Schema\Migrations\6.sql">
-      <InputToMutableSqlGenerator>true</InputToMutableSqlGenerator>
       <InputToImmutableSqlGenerator>true</InputToImmutableSqlGenerator>
```
Ivanidzo4ka, line 14: >true [](start = 36, length = 4)

is it because we don't include ReferenceTokenCompositeSearchParamTableType_1 in 7.sql, but it's still used somehow in code?

johnstairs: Yes, for backwards compat. First, we upgrade the code, which needs to continue to work against the V6 schema. Then, after we upgrade the schema, the code can start using the V7 constructs. 
So the code needs to have access to the older types.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/UntypedReferenceRewriter.cs
johnstairs: I don't know. Would `componentCandidates.And(componentsPresent.Not())` really improve readability :)?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/UntypedReferenceRewriter.cs
johnstairs: Yeah, you are right. Definitely needs a comment.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/UntypedReferenceRewriter.cs
```diff
@@ -0,0 +1,108 @@
...
+            // Bit 0 is for reference, non-composite parameters where the reference if of one type
```
namadabu, line 26: >if [](start = 84, length = 2)

nit: typo, if -> is

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/7.sql
```diff
@@ -0,0 +1,1943 @@
+-- Style guide: please see: https://github.com/ktaranov/sqlserver-kit/blob/master/SQL%20Server%20Name%20Convention%20and%20T-SQL%20Programming%20Style.md
+
+/*************************************************************
+    Schema Version
+**************************************************************/
+
+INSERT INTO dbo.SchemaVersion
+VALUES
+    (6, 'started')
```
johnstairs, line 9: Needs to be 7

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchIndexer.cs
```diff
@@ -234,12 +234,32 @@ private IEnumerable<SearchIndexEntry> ProcessNonCompositeSearchParameter(SearchP
...
+                IEnumerable<ISearchValue> searchValues = converter.ConvertTo(extractedValue);
 
-                results.AddRange(converter.ConvertTo(extractedValue) ?? Enumerable.Empty<ISearchValue>());
+                if (searchValues != null)
+                {
+                    if (searchParameterType == SearchParamType.Reference && allowedReferenceResourceTypes?.Count == 1)
+                    {
+                        // For references, if the type is not specified in the reference string, we can set the type on the search value because
+                        // in this case it can only be of one type.
+                        string singleAllowedResourceType = null;
```
Ivanidzo4ka, line 245: >null [](start = 59, length = 4)

Are we doing late binding via `??==` because of optimization of the case when we don't have ReferenceSearchValue in searchValues and allowedReferenceResourceTypes.Single() is costly operation?

johnstairs: Actually, I put the Single() call in before switching to IReadOnlyCollection, so this is not necessary anymore. Thank you for catching that.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/UntypedReferenceRewriter.cs
Ivanidzo4ka: Not that case, but rest of them, I personally always forget which shift left or shift right is increasing value, and which one is decreasing (without context of overflow), access to bit by index is much easier in case of BitArray.
But someone can argue should I even be allowed to read code if I don't have such a basic bit operation understanding :D 
Again, it's just an opinion.

---
In reply to: [538869133](https://github.com/microsoft/fhir-server/pull/1497#discussion_r538869133) [](ancestors = 538869133)

Ivanidzo4ka:
>                 allowedReferenceResourceTypes.Any())

This can be changed to allowedReferenceResourceTypes?.Count >0 I would assume


---
Refers to: src/Microsoft.Health.Fhir.Core/Features/Search/TypedElementSearchIndexer.cs:203 in 4640916. [](commit_id = 464091685ef57cc91d8e167b164e6b4297cc64f3, deletion_comment = False)

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/7.sql
```diff
@@ -0,0 +1,1943 @@
...
+(
+    ResourceTypeId smallint NOT NULL,
+    ResourceSurrogateId bigint NOT NULL,
+    SearchParamId smallint NOT NULL,
+    BaseUri1 varchar(128) COLLATE Latin1_General_100_CS_AS NULL,
+    ReferenceResourceTypeId1 smallint NULL,
+    ReferenceResourceId1 varchar(64) COLLATE Latin1_General_100_CS_AS NOT NULL,
+    ReferenceResourceVersion1 int NULL,
+    SystemId2 int NULL,
+    Code2 varchar(128) COLLATE Latin1_General_100_CS_AS NOT NULL,
```
Ivanidzo4ka, line 705: Pure curiosity, why they have `2` in the end? They are not reserved words, they belong to scope of the table, is it because somewhere later they got joined and it's easier to write `System2` than `ReferenceToken.SystemId` ?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/ReferenceTokenCompositeSearchParameterV2RowGenerator.cs
```diff
@@ -0,0 +1,47 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using Microsoft.Health.Fhir.Core.Features.Search.SearchValues;
+using Microsoft.Health.Fhir.SqlServer.Features.Schema.Model;
+
+namespace Microsoft.Health.Fhir.SqlServer.Features.Storage.TvpRowGeneration
+{
+    internal class ReferenceTokenCompositeSearchParameterV2RowGenerator : CompositeSearchParameterRowGenerator<(ReferenceSearchValue component1, TokenSearchValue component2), ReferenceTokenCompositeSearchParamTableTypeV2Row>
```
Ivanidzo4ka, line 11: >ReferenceTokenCompositeSearchParameterV2RowGenerator [](start = 19, length = 52)

They consumed via reflection, and this is reason why `ReferenceTokenCompositeSearchParamTableType_1` got removed from sql script, is this correct assumption?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/7.diff.sql
```diff
@@ -0,0 +1,397 @@
...
+
+GO
+
+/*************************************************************
+    Stored procedures for creating and deleting
+**************************************************************/
+
+--
+-- STORED PROCEDURE
+--     UpsertResource
```
Ivanidzo4ka, line 35: >UpsertResource [](start = 7, length = 14)

should we change name to `UpsertResource_2` ?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/7.sql
johnstairs: This is a convention for composite search parameter tables. Composites are made up of several search parameters. In this case (Reference, Token) all field names would unique without adding component index suffix. But `TokenTokenCompositeSearchParam` some kind of differentiation is required. It also makes it clearer (IMO) which component a field belongs to, which is useful for reasoning about indexes.
And we adopt the convention across all composite search parameters tables because it makes it easier for the SQL generation code to not have to special-case the logic per-table.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/ReferenceTokenCompositeSearchParameterV2RowGenerator.cs
johnstairs: We removed `ReferenceTokenCompositeSearchParamTableType_1` from the "snapshot" 7.sql SQL script because it has been superseded by `ReferenceTokenCompositeSearchParamTableType_2`. 

We codegen a class and a struct from the SQL DDL files for each table type, in this case `ReferenceTokenCompositeSearchParamTableTypeV2TableValuedParameterDefinition` and `ReferenceTokenCompositeSearchParamTableTypeV2Row` respectively.

The row generator in this file is registered as a service in the dependency injection container and consumed by the generated `UpsertResourceTvpGenerator` class.

src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterDefinitionBuilder.cs
johnstairs: Renamed.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/UntypedReferenceRewriter.cs
johnstairs: Added an explanation. I hope it's clearer now.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/UntypedReferenceRewriter.cs
johnstairs: Added more comments.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/7.diff.sql
johnstairs: Good catch. Renamed the name in the the comments.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/UntypedReferenceRewriter.cs
```diff
@@ -0,0 +1,112 @@
...
+
+            int componentsToFill = componentCandidates & ~componentsPresent;
+
+            if (componentsToFill == 0)
+            {
+                return expression;
+            }
+
+            List<Expression> newExpressionsToBeAnded;
+            if (expression.Expression is MultiaryExpression me)
```
Ivanidzo4ka, line 63: >MultiaryExpression me [](start = 41, length = 21)

Would it worth to add some Ensure or Debug.Assert here what me MultiaryOperation is actually And?
Could go terribly wrong if we AND an OR expression (which I assume never the case because that's not how code operates right now, but i think it's better be safe.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/UntypedReferenceRewriter.cs
```diff
@@ -0,0 +1,112 @@
...
+using Microsoft.Health.Fhir.ValueSets;
+
+namespace Microsoft.Health.Fhir.SqlServer.Features.Search.Expressions.Visitors
+{
+    /// <summary>
+    /// For expressions over a reference search parameter where the type is not specified
+    /// (e.g. "abc" instead of "Patient/123") AND the search parameter is constrained to be
+    /// of only one type, rewrites the expression to specify the single target type.
+    /// </summary>
+    internal class UntypedReferenceRewriter : ExpressionRewriterWithInitialContext<object>
```
Ivanidzo4ka, line 18: >UntypedReferenceRewriter [](start = 19, length = 24)

nit: I would consider adding `SearchParameter` to name since it's operates only on search parameters.
Or maybe it's too mouthful...

src/Microsoft.Health.Fhir.Core/Models/SearchParameterInfo.cs
```diff
@@ -94,6 +94,14 @@ public SearchParameterInfo(string name)
...
 
+        /// <summary>
+        /// The component definitions if this is a composite search parameter (<see cref="Type"/> is <see cref="SearchParamType.Composite"/>)
+        /// </summary>
         public IReadOnlyList<SearchParameterComponentInfo> Component { get; }
+
+        /// <summary>
+        /// The resolved <see cref="SearchParameterInfo"/>s for each component if this is a composite search parameter (<see cref="Type"/> is <see cref="SearchParamType.Composite"/>)
+        /// </summary>
+        public IReadOnlyList<SearchParameterInfo> ResolvedComponents { get; set; } = Array.Empty<SearchParameterInfo>();
```
Ivanidzo4ka, line 105: >ResolvedComponents [](start = 50, length = 18)

nit: I would consider to move `SearchParameterInfo.Type` and `SearchParameterInfo.TargetResourceTypes` into `SearchParameterComponentInfo` to unify them. Having two, somewhat similar entities, can be a bit confusing (at least for me)

Ivanidzo4ka (approved):
:shipit:

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/UntypedReferenceRewriter.cs
johnstairs: Good idea

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/UntypedReferenceRewriter.cs
johnstairs: All parsed expressions are either a single SearchParameterExpression or ANDed SearchParameterExpressions, so I'm not sure how helpful the addind that to the name would be. (A later rewriter creates a SqlRootExpression, but ultimately, any expression over any field is always going to be under a SearchParameterExpression).

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/UntypedReferenceRewriter.cs
johnstairs: ...And in so doing I completely missed ORed expressions like Patient?Organization=a,b 
😱 
Fixed that now. Thanks for the comment. I added the check to `ParameterPredicateVisitor.VisitMultiary()`

src/Microsoft.Health.Fhir.Core/Models/SearchParameterInfo.cs
johnstairs: Thank you for the suggestion. I've added a `ResolvedSearchParameter` property to `SearchParameterComponentInfo` and removed `ResolvedComponents` from `SearchParameterInfo`

Ivanidzo4ka:
>             _searchParameterDefinitionManager = searchParameterDefinitionManagerResolver();

nit: looks like this can be removed as well.

---
Refers to: src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/Parsers/SearchParameterExpressionParser.cs:41 in a8d18a5. [](commit_id = a8d18a5ac250807fb5c1fb2b9bd6570f136584da, deletion_comment = False)

johnstairs:
Thank you for the review and feedback , @Ivanidzo4ka!

PR #1498, opened 2020-12-06
HealthCheck need to handle error due to customer CMK settings

shmartel:
## Description
- Update the HealthCheck to not report unhealthy when the error received is expected due to a customer's CMK settings.
 - Also, remove 4007 from the known list of Cosmos DB CMK sub status value.  After reading https://docs.microsoft.com/en-us/rest/api/cosmos-db/http-status-codes-for-cosmosdb#http-substatus-codes, this is not a CMK-related issue

## Related issues
Addresses [#77750](https://microsofthealth.visualstudio.com/Health/_workitems/edit/77750)

## Testing
All existing unit tests passed, and I added new unit test coverage

### DIFF
--- a/src/Microsoft.Health.Fhir.CosmosDb/Features/Health/CosmosHealthCheck.cs
+++ b/src/Microsoft.Health.Fhir.CosmosDb/Features/Health/CosmosHealthCheck.cs
@@ -64,6 +64,10 @@ namespace Microsoft.Health.Fhir.CosmosDb.Features.Health
 
                 return HealthCheckResult.Healthy("Successfully connected to the data store.");
             }
+            catch (CosmosException ex) when (ex.IsCmkClientError())
+            {
+                return HealthCheckResult.Healthy($"Connection to the data store was unsuccesful, however this is expected because the client's customer-managed key is not available. Error: {ex.Message}");
+            }
             catch (Exception ex) when (ex.IsRequestRateExceeded())
             {
                 return HealthCheckResult.Healthy("Connection to the data store was successful, however, the rate limit has been exceeded.");
--- a/src/Microsoft.Health.Fhir.CosmosDb/Features/Health/CosmosHealthCheck.cs
+++ b/src/Microsoft.Health.Fhir.CosmosDb/Features/Health/CosmosHealthCheck.cs
@@ -4,6 +4,7 @@
 // -------------------------------------------------------------------------------------------------
 
 using System;
+using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 using EnsureThat;
@@ -66,7 +67,10 @@ namespace Microsoft.Health.Fhir.CosmosDb.Features.Health
             }
             catch (CosmosException ex) when (ex.IsCmkClientError())
             {
-                return HealthCheckResult.Healthy($"Connection to the data store was unsuccesful, however this is expected because the client's customer-managed key is not available. Error: {ex.Message}");
+                return HealthCheckResult.Unhealthy(
+                    "Connection to the data store was unsuccesful because of the client's customer-managed key is not available.",
+                    exception: ex,
+                    new Dictionary<string, object>() { { "IsCustomerManagedKeyError", true } });
             }
             catch (Exception ex) when (ex.IsRequestRateExceeded())
             {
--- a/src/Microsoft.Health.Fhir.CosmosDb/Features/Health/CosmosHealthCheck.cs
+++ b/src/Microsoft.Health.Fhir.CosmosDb/Features/Health/CosmosHealthCheck.cs
@@ -68,7 +68,7 @@ namespace Microsoft.Health.Fhir.CosmosDb.Features.Health
             catch (CosmosException ex) when (ex.IsCmkClientError())
             {
                 return HealthCheckResult.Unhealthy(
-                    "Connection to the data store was unsuccesful because of the client's customer-managed key is not available.",
+                    "Connection to the data store was unsuccesful because the client's customer-managed key is not available.",
                     exception: ex,
                     new Dictionary<string, object>() { { "IsCustomerManagedKeyError", true } });
             }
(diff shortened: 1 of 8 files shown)


src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/KnownCosmosDbCmkSubStatusValue.cs
```diff
@@ -11,15 +11,14 @@ namespace Microsoft.Health.Fhir.CosmosDb.Features.Storage
     public enum KnownCosmosDbCmkSubStatusValue
```
shmartel, line 11: FYI: I broke out this enum into "server-side" and "client side" issues based on the categorizations listed in https://docs.microsoft.com/en-us/rest/api/cosmos-db/http-status-codes-for-cosmosdb#http-substatus-codes

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosResponseProcessor.cs
```diff
@@ -163,9 +163,6 @@ private string GetCustomerManagedKeyErrorMessage(int subStatusCode)
                 case KnownCosmosDbCmkSubStatusValue.InvalidKeyVaultKeyUri:
                     errorMessage = Resources.InvalidKeyVaultKeyUri;
                     break;
-                case KnownCosmosDbCmkSubStatusValue.InvalidInputBytes:
```
shmartel, line 166: FYI: I removed this value.  Looking at new docs from https://docs.microsoft.com/en-us/rest/api/cosmos-db/http-status-codes-for-cosmosdb#http-substatus-codes, it does not seem to be related to CMK.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosExceptionExtensions.cs
```diff
@@ -54,5 +55,16 @@ public static class CosmosExceptionExtensions
...
+
+        /// <summary>
+        /// Determines if the error is due to a client customer-managed key error
+        /// </summary>
+        /// <param name="exception">The exception object</param>
+        /// <returns>True iff the error is due to client CMK setting.</returns>
+        public static bool IsCmkClientError(this CosmosException exception)
+        {
+            return exception.StatusCode == HttpStatusCode.Forbidden
+                && (Enum.IsDefined(typeof(KnownCosmosDbCmkSubStatusValueClientIssue), exception.SubStatusCode) || exception.SubStatusCode == 3);
```
shmartel, line 67: For a substatus code, I expected values from KnownCosmosDbCmkSubStatusValueClientIssue.  However, when testing this scenario, I noticed that Cosmos DB is returning substatus=3.  I think this is an issue and filed incident 217300148.  However, to make this scenario work work today, I have allowed for this unexpected status code.

src/Microsoft.Health.Fhir.CosmosDb/Features/Health/CosmosHealthCheck.cs
```diff
@@ -64,6 +64,10 @@ public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context
 
                 return HealthCheckResult.Healthy("Successfully connected to the data store.");
             }
+            catch (CosmosException ex) when (ex.IsCmkClientError())
+            {
+                return HealthCheckResult.Healthy($"Connection to the data store was unsuccesful, however this is expected because the client's customer-managed key is not available. Error: {ex.Message}");
```
brendankowitz, line 69: Just following up here, did we decide to change this to Unhealthy?

shmartel: Yes.  I will change that in my next iteration.  Additionally, I plan on including more info in the health check to denote this is a client CMK error, I will includer the Exception itself, or a dictionary entry like "isClientCustomerManagedKeyError"=true (or something like that), or possibly both.  I'm leaning towards both.

shmartel: The dictionary entry will denote that the error has been classified as a client CMK error, and the exception will contain an error message saying what the error actually is.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosExceptionExtensions.cs
shmartel: Per the incident, there are no current plans to update this behavior.  So, I will keep the clause to handle substatus=3

LTA-Thinking: Should we add it as a value of the enum if it is "expected"? Either way I think a comment in the code explaining why this value is here would be valuable.

src/Microsoft.Health.Fhir.CosmosDb/Features/Health/CosmosHealthCheck.cs
```diff
@@ -64,6 +65,13 @@ public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context
 
                 return HealthCheckResult.Healthy("Successfully connected to the data store.");
             }
+            catch (CosmosException ex) when (ex.IsCmkClientError())
+            {
+                return HealthCheckResult.Unhealthy(
+                    "Connection to the data store was unsuccesful because of the client's customer-managed key is not available.",
```
LTA-Thinking, line 71: Nit: 'of' isn't needed after 'because'

LTA-Thinking (approved):

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosExceptionExtensions.cs
shmartel: I would like to add it to the enum of expected values, but it's not one that is actually expected, per documentation.  It's one that has been observed, and then acknowledged by Cosmos DB.  So, I have added a comment here.  Let me know if you prefer some other wordsmithing.

PR #1505, opened 2020-12-09
Fixes smartlauncher references for e2e tests

brendankowitz:
## Description
- Fixes SMART sample app to be packaged with internal namespace
- Moves SMART E2E tests to a shared project so they are run in all versions of FHIR (was previously only in the STU3 assembly)

## Testing
SMART E2E tests now run in all versions of FHIR


Ivanidzo4ka (approved):
:shipit:

Issue #1503, opened 2020-12-09
ArgumentNullException thrown on GET request


rotodd:
**Describe the bug**
We are seeing [this](https://jarvis-west.dc.ad.msft.net/ADD23BA6) exception in the logs.

The end of the stack trace points to [here](https://github.com/microsoft/fhir-server/blob/master/src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/ResourceTypeAndId.cs), which indicates this could be due to a bug in how the query is constructed for the _include functionality that was added in #1426.

**FHIR Version?**
Stu3/R4/R5

**Data provider?**
CosmosDB

CaitlinV39:
@johnstairs I think you resolved this. Can we close?

PR #1508, opened 2020-12-10
Handle _include in Cosmos DB with untyped references

johnstairs:
## Description
Fixes a bug with in the `_include` implementation for Cosmos DB, where we were not accounting for untyped references in the database like `"123"` instead of `"Patient/123"`. We currently null-ref when we encounter this kind of reference in an `_include`. With this change, we will ignore these kinds of references, which aligns with the current SQL server behavior.

## Related issues
Addresses #1503

## Testing
E2E tests.


johnstairs:
This is currently dependent on #1497. I will change the target branch one it has merged.

feordin (approved):
LGTM :)

PR #1509, opened 2020-12-10
PR resource names include build number

LTA-Thinking:
## Description
Changes the names of the resources created for testing a PR to include the build number in addition to the PR number. 

## Related issues
Previously we have encountered an error where resource names were conflicting with existing resources. This would require a new PR to be created as the names of the resources was based only on the PR number. This change will change the names of the resources on each build, allowing a rerun of the build to fix this issue.

## Testing
Running the PR.

## FHIR Team Checklist
- [x] Tag the PR with the type of update: **Bug**, **Enhancement**, or **New-Feature**

### DIFF
--- a/build/pr-pipeline.yml
+++ b/build/pr-pipeline.yml
@@ -18,7 +18,7 @@ stages:
     steps:
     - template: ./jobs/update-semver.yml
     - bash: |
-        envName = "${$(GitVersion.semver)//\./}"
+        envName = "${$(GitVersion.semver)//.}"
         echo "##vso[task.setvariable variable=environmentName]$envName"
     - script: echo %Action%%BuildVersion%
       displayName: 'Set build version'
--- a/build/pr-pipeline.yml
+++ b/build/pr-pipeline.yml
@@ -21,7 +21,7 @@ stages:
         $buildNumber = "$(GitVersion.semVer)" -replace "\.", "" 
         Write-Host "##vso[build.updatebuildnumber]$buildNumber" 
         Write-Host "Updated  build number to '$buildNumber"
-      name: "Set build version"
+      name: SetBuildVersion
 
 - stage: BuildUnitTests
   displayName: 'Build and run unit tests'
--- a/build/pr-pipeline.yml
+++ b/build/pr-pipeline.yml
@@ -17,11 +17,6 @@ stages:
       vmImage: 'ubuntu-latest'
     steps:
     - template: ./jobs/update-semver.yml
-    - powershell: |
-        $buildNumber = "$(GitVersion.semVer)" -replace "\.", "" 
-        Write-Host "##vso[build.updatebuildnumber]$buildNumber" 
-        Write-Host "Updated  build number to '$buildNumber"
-      name: SetBuildVersion
 
 - stage: BuildUnitTests
   displayName: 'Build and run unit tests'
--- a/build/pr-pipeline.yml
+++ b/build/pr-pipeline.yml
@@ -18,7 +18,8 @@ stages:
     steps:
     - template: ./jobs/update-semver.yml
     - bash: |
-        envName = "${$(GitVersion.semver)//.}"
+        semver = $(GitVersion.semver)
+        envName = "${semver//.}"
         echo "##vso[task.setvariable variable=environmentName]$envName"
     - script: echo %Action%%BuildVersion%
       displayName: 'Set build version'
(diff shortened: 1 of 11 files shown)


build/pr-pipeline.yml
```diff
@@ -64,7 +65,13 @@ stages:
...
         ScriptType: inlineScript
         Inline: |
-          New-AzResourceGroup -Name "$(DeploymentEnvironmentName)" -Location "$(ResourceGroupRegion)" -Force
+          try
+          {
+            Get-AzResourceGroup -Name $(ResourceGroupName) | Remove-AzResourceGroup -Verbose -Force
+          }
+          catch
+          {}
+          New-AzResourceGroup -Name "$(ResourceGroupName)" -Location "$(ResourceGroupRegion)" -Force
```
Ivanidzo4ka, line 74: >New-AzResourceGroup [](start = 10, length = 19)

Aren't this operation idempotent? Why we want drop resource group and everything in it?

LTA-Thinking: If the previous build didn't successfully complete the resource group from the last build will still be around. Before this change the old resources would be overwritten by the new ones since they would have the same name. Now new resources with different names will be added. To avoid pilling up a lot of resources in the resource group I added this to clean it up before the new resources are made.

build/pr-variables.yml
```diff
@@ -3,7 +3,8 @@ variables:
     resourceGroupRoot: 'msh-fhir-pr'
     appServicePlanName: '$(resourceGroupRoot)-$(prNumber)-asp'
     prNumber: $(system.pullRequest.pullRequestNumber)
-    DeploymentEnvironmentName: '$(resourceGroupRoot)-$(prNumber)'
+    ResourceGroupName: '$(resourceGroupRoot)-$(prNumber)'
+    DeploymentEnvironmentName: '$(build.BuildNumber)'
```
Ivanidzo4ka, line 7: >. [](start = 39, length = 1)

looks like `.` is not allowed for keyvault name

brendankowitz: One of the other variables replaces them with dashes, I think it was $(GitVersion.NuGetVersionV2)
For reference: https://github.com/GitTools/actions/blob/master/gitversion/execute/action.yml

build/pr-pipeline.yml
Ivanidzo4ka: Make sense, on the same note, what happens to resources if someone closes PR for failed build? Are they automatically cleaned?

---
In reply to: [540447194](https://github.com/microsoft/fhir-server/pull/1509#discussion_r540447194) [](ancestors = 540447194)

Ivanidzo4ka (approved):
:shipit:

build/pr-pipeline.yml
brendankowitz: This keeps changing though, e.g.
If I have `1.0.1-pr1.1`
I add a commit to fix my build => `1.0.1-pr1.2`
Someone checks into master then I add a commit => `1.0.2-pr1.2`

build/pr-variables.yml
LTA-Thinking: That is super helpful! Thank you!

build/pr-pipeline.yml
LTA-Thinking: If someone closes a PR with a failed build the resource group is not cleaned up. This is an open issue for us. I had added a script to cleanup a resource group when a PR was closed, but it wasn't working because the GithubAction  to run AzureDevOps pipelines doesn't have a feature we need (https://github.com/Azure/pipelines/issues/17)

build/pr-variables.yml
```diff
@@ -3,7 +3,8 @@ variables:
     resourceGroupRoot: 'msh-fhir-pr'
     appServicePlanName: '$(resourceGroupRoot)-$(prNumber)-asp'
     prNumber: $(system.pullRequest.pullRequestNumber)
-    DeploymentEnvironmentName: '$(resourceGroupRoot)-$(prNumber)'
+    ResourceGroupName: '$(resourceGroupRoot)-$(prNumber)'
+    DeploymentEnvironmentName: 'f$(build.BuildNumber)'
```
Ivanidzo4ka, line 7: >f [](start = 32, length = 1)

why plain `f` is better than `fhir`?

LTA-Thinking: It is shorter. I'm running into issues with the full build number being too long for some resource types. There is a max name length of 24 characters.

feordin (approved):

PR #1516, opened 2020-12-15
sqlSchemaAutomaticUpdatesEnabled now set to valid bool

alsami:
## Description
When using `solutionType` other than `FhirServerCosmosDB` (currently only other option is `FhirServerSqlServer`) then the flag `sqlSchemaAutomaticUpdatesEnabled` shall be set to true, so that the schema will be updated and the templates are valid.

I am not a hundred percent sure if that flag has been set as empty on purpose, given that one might not want the schema to be automatically applied / patched / whatever. If so, please let me know, so I can at least set it to `false`.

## Related issues
There is no related issue, I stumbled on that problem while trying to deploy the environment.

## Testing
I used Azure-CLI to validate that the template-files are still working.

For the docker-template:
`az deployment group validate --template-file default-azuredeploy-docker.json -g <the-resource-group>`

For the default-template:
`az deployment group validate --template-file default-azuredeploy.json -g <the-resource-group>`

It would ask for a service-name after, which I set to `fire-server`.

## FHIR Team Checklist
- [ ] Review [squash-merge requirements](https://github.com/microsoft/fhir-server/blob/master/SquashMergeRequirements.md)

Patch

### DIFF
--- a/samples/templates/default-azuredeploy-docker.json
+++ b/samples/templates/default-azuredeploy-docker.json
@@ -139,7 +139,7 @@
         },
         "sqlSchemaAutomaticUpdatesEnabled": {
             "type": "bool",
-            "defaultValue": "[if(equals(parameters('solutionType'),'FhirServerCosmosDB'), bool('false'), '')]",
+            "defaultValue": "[if(equals(parameters('solutionType'),'FhirServerCosmosDB'), bool('false'), bool('true'))]",
             "metadata": {
                 "description": "Determine whether the sql schema should be automatically upgraded on server startup. If false, sql schema will not be initialized or upgraded on the server startup. The schema migration tool will be required to perform initialize or upgrade. If true, sql schema will be upgraded to the maximum supported version."
             }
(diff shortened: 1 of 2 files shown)


samples/templates/default-azuredeploy-docker.json
```diff
@@ -139,7 +139,7 @@
         },
         "sqlSchemaAutomaticUpdatesEnabled": {
             "type": "bool",
-            "defaultValue": "[if(equals(parameters('solutionType'),'FhirServerCosmosDB'), bool('false'), '')]",
+            "defaultValue": "[if(equals(parameters('solutionType'),'FhirServerCosmosDB'), bool('false'), bool('true'))]",
```
rbans96, line 142: @alsami  - Thanks for checking. We have added 'sqlSchemaAutomaticUpdatesEnabled' as an empty intentionally to let users decide consciously if they want to enable or disable automatic schema initialization based on their requirement. If we default to any(true\false), this may get overlooked and not get the desired behavior.

alsami: @rbans96 I thought so but wasn't sure. Problem I see is that the error-messages are not very descriptive, something like 'expected type System.Bool but got System.String'.

Thanks for the heads up anyway!

PR #1519, opened 2020-12-15
Fix image cleanup script

LTA-Thinking:
## Description
Fixes a breaking typo and other issues in the docker image cleanup script preventing it from running

## Related issues
The image cleanup can't run

## Testing
The script will be run from the branch

## FHIR Team Checklist

Skip (build infrastructure)

### DIFF
--- a/build/cleanup-images.yml
+++ b/build/cleanup-images.yml
@@ -32,7 +32,7 @@ jobs:
                 $difference = New-TimeSpan -Start $imageDate.lastUpdateTime -End $currentDate
 
                 if (($tagParts[0] -notlike "*build*") -and (($difference.totalDays -gt 60) -or (($difference.totalDays -gt 7) -and ($tagParts[0] -like "pr*")))) {
-                  az acr repository untag --name $(azureContai | nerRegistryName) --image '${version}_fhir-server:${tag}'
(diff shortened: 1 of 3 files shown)


namadabu (dismissed):

build/cleanup-images.yml
```diff
@@ -4,45 +4,26 @@
...
-        foreach ($version in $versions) {
-          $repository = "${version}_fhir-server"
-          $tagList = az acr repository show-tags --name $(azureContainerRegistryName) --repository $repository | ConvertFrom-Json
+stages:
+- stage: CleanupStu3
+  displayName: 'Cleanup Stu3'
+  jobs:
+  - template: ./jobs/cleanup-version-images.yml
+    parameters: 
+      version: 'stu3'
```
namadabu, line 13: >stu3 [](start = 16, length = 4)

nit: does the case matter here? In other yml files we seem to pass in Stu3, R4 and R5 for the version parameters.

namadabu (approved):
:shipit:

build/cleanup-images.yml
LTA-Thinking: In this case it does because it has to match the image name, which I'm pretty sure is case sensitive. We use the uppercase ones when it is display or documentation and lowercase for code uses.

PR #1525, opened 2020-12-18
Support splitting of exported files by size

namadabu:
## Description
Currently when we export data from the fhir-server we create a separate file for each resource type and export all the resources into the corresponding files. When exporting huge amounts of data this could cause a single file to become multiple gigabytes in size. This PR updates the export code to limit the size of a single exported file to a configurable maximum (default value of 64 MB). As an example, instead of a single Patient.ndjson file, we will now have Patient-1.ndjson, Patient-2.ndjson, … etc. 

## Related issues
Addresses [AB#78415](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/78415)

## Testing
Existing tests. Added UTs. Manual E2E test on local deployment.

## FHIR Team Checklist

Patch


src/Microsoft.Health.Fhir.Shared.Web/appsettings.json
```diff
@@ -47,6 +47,7 @@
                 "JobPollingFrequency": "00:00:10",
                 "MaximumNumberOfResourcesPerQuery": 100,
                 "NumberOfPagesPerCommit": 10,
+                "MaxFileSizeInMB": 64,
                 "Formats": [
                     {
                         "Name": "resourcename_only",
```
LTA-Thinking, line 53: Should we add \<sequence\> to these Formats as well?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportFileManager.cs
```diff
@@ -0,0 +1,134 @@
...
+                await Initialize(cancellationToken);
+            }
+
+            ExportFileInfo fileInfo = await GetFile(resourceType, cancellationToken);
+            await _exportDestinationClient.WriteFilePartAsync(fileInfo.FileUri, partId, data, cancellationToken);
+
+            fileInfo.IncrementCount(data.Length);
+
+            // We need to create a new file if the current file has exceeded a certain limit.
+            if (fileInfo.CommittedBytes >= _maxFileSizeInBytes)
```
LTA-Thinking, line 74: This makes me concerned about calling the variable maxFileSizeInBytes. Most files will be slightly larger than this value before a new one is made, so if a user sets the value to the max size a storage account file can be we could see the same error we currently get. Would it be possible to check if the next data write would put the file over the limit, and make a new file before we go above this point?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportFileManager.cs
```diff
@@ -0,0 +1,134 @@
...
+            string fileName = _exportJobRecord.ExportFormat + ".ndjson";
+
+            string dateTime = _exportJobRecord.QueuedTime.UtcDateTime.ToString("s")
+                    .Replace("-", string.Empty, StringComparison.OrdinalIgnoreCase)
+                    .Replace(":", string.Empty, StringComparison.OrdinalIgnoreCase);
+
+            fileName = fileName.Replace(ExportFormatTags.Timestamp, dateTime, StringComparison.OrdinalIgnoreCase);
+            fileName = fileName.Replace(ExportFormatTags.Id, _exportJobRecord.Id, StringComparison.OrdinalIgnoreCase);
+            fileName = fileName.Replace(ExportFormatTags.ResourceName, resourceType, StringComparison.OrdinalIgnoreCase);
+            fileName = fileName.Replace(ExportFormatTags.Sequence, fileSequence.ToString(), StringComparison.OrdinalIgnoreCase);
```
LTA-Thinking, line 105: Having the sequence as part of the format concerns me since a user could make a format that doesn't contain a sequence and that would break our ability to export large data sets. Should we make sequence something we automatically append to the end of a file name?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportFileManager.cs
```diff
@@ -0,0 +1,134 @@
...
+            _maxFileSizeInBytes = _exportJobRecord.MaxFileSizeInMB * 1024 * 1024;
+            _resourceTypeToFileInfoMapping = new Dictionary<string, ExportFileInfo>();
+        }
+
+        private async Task Initialize(CancellationToken cancellationToken)
+        {
+            // Each resource type can have multiple files. We need to keep track of the latest file.
+            foreach (KeyValuePair<string, List<ExportFileInfo>> output in _exportJobRecord.Output)
+            {
+                ExportFileInfo latestFile = output.Value[output.Value.Count - 1];
```
LTA-Thinking, line 47: I'm concerned about the order the files would be in after a restart, whether the one with the highest sequence will always be the last one in the array.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportFileManager.cs
namadabu: That's a good point. I misunderstood the way export format is used. I have updated the code so that we add a sequence to the end of the file irrespective of what format we are using (since that is what I was aiming for).

src/Microsoft.Health.Fhir.Shared.Web/appsettings.json
namadabu: Like discussed below, have updated code to add sequence irrespective of format.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportFileManager.cs
namadabu: My understanding was that since we are storing the ExportFileInfo as a list, the order should be preserved when de-serializing the json into an object. But I looked at the documentation and it does not specifically mention that order will be preserved (though practically that seems to be the case). To be on the safer side, I have updated the code to loop through the export files and find the most recent one.

src/Microsoft.Health.Fhir.Core/Configs/ExportJobConfiguration.cs
```diff
@@ -52,5 +52,10 @@ public class ExportJobConfiguration
         /// </summary>
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Usage", "CA2227:Collection properties should be read only", Justification = "This is a configuration class")]
         public IList<ExportJobFormatConfiguration> Formats { get; set; }
+
+        /// <summary>
+        /// Determines max size for an exported file.
+        /// </summary>
+        public uint MaxFileSizeInMB { get; set; } = 64;
```
Ivanidzo4ka, line 59: That whole name and description reads for me:
`We not gonna export more than this limit`
while your intent is more a rolling file size.

So I would suggest to rename it RollingMaxFileSize.

Also I somewhat feel it needs to be part of the `ExportJobFormatConfiguration` since you define your file name pattern via format

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportFileManager.cs
namadabu: This is intended to be more of an "approximate" max file size. My plan was to call this out when we add documentation for this configuration. 

One issue that could happen if we check for file size before adding is that we have the potential to end up in an infinite loop without exporting any data. This could happen if the user sets a very low limit value by mistake. Or if the data that has to be added is greater than the limit set by the user (say 1MB limit and the resource data that needs to be uploaded is 2MB). To handle this scenario we have to add more logic to "force" add files when we encounter such situations. We finally end up with a file size that is greater than the specified limit. We will have to call this out in the documentation either way.

I'll update the code to check before adding.

LTA-Thinking: That is a good point. I'm ok with the current approach and handling this in documentation.

src/Microsoft.Health.Fhir.Core/Configs/ExportJobConfiguration.cs
namadabu: I have updated the name to `RollingFileSizeInMB`. But I have kept the config location in `Export` since I have updated the code to use this outside the export format. The sequence will now be added to every file irrespective of the Export format.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/Models/ExportJobRecordOutputConverter.cs
```diff
@@ -0,0 +1,64 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Health.Fhir.Core.Features.Operations.Export.Models
```
LTA-Thinking, line 11: Is it possible to have unit tests for this class?

LTA-Thinking (approved):

LTA-Thinking:
Would this be considered a breaking change since it changes the names of the outputted files?

CaitlinV39:
@namadabu and @LTA-Thinking I added the KI-warning label. I'm on the fence of whether we would say this is a breaking change, but at a minimum we should have a warning. FYI @brendankowitz

namadabu:
@LTA-Thinking That's a good point regarding the file name change. The most common scenario for using these files in an `$export` workflow would be to read the filenames from the response to the GetExportStatus request. It could be that users might have hardcoded their tools/scripts for the current pattern of file names. But I feel it would be a less common scenario. 

I usually like to err on the safer side and hence would prefer marking it as a breaking change (to make sure users are aware). Thoughts @CaitlinV39  @brendankowitz  ?

LTA-Thinking (approved):

PR #1549, opened 2020-12-23
Added support for type modifier on reference search parameter

anyname2:
## Description
Added support for type modifer when searching by reference. It was previously only supported for chained reference, but not when reference is the last search parameter.

#### Example query
```http
GET /Encounter?subject:Patient=Patient/123
```

## Testing
Unit + manual (SqlServer)

## FHIR Team Checklist
- [ ] Review [squash-merge requirements](https://github.com/microsoft/fhir-server/blob/master/SquashMergeRequirements.md)

Patch (Modifier was previously supported if it was in the middle of the chain, but not the end)

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/Parsers/ExpressionParser.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/Parsers/ExpressionParser.cs
@@ -265,20 +265,31 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Expressions.Parsers
 
         private Expression ParseSearchValueExpression(SearchParameterInfo searchParameter, string modifier, string value)
         {
-            SearchModifierCode? parsedModifier = ParseSearchParamModifier();
+            SearchModifierCode? parsedModifier = null;
+            string targetTypeModifier = null;
+            if (!string.IsNullOrEmpty(modifier))
+            {
+                parsedModifier = ParseSearchParamModifier();
 
-            return _searchParameterExpressionParser.Parse(searchParameter, parsedModifier, value);
+                if (parsedModifier == null)
+                {
+                    targetTypeModifier = modifier;
+                }
+            }
+
+            return _searchParameterExpressionParser.Parse(searchParameter, parsedModifier, targetTypeModifier, value);
 
             SearchModifierCode? ParseSearchParamModifier()
             {
-                if (string.IsNullOrEmpty(modifier))
-                {
-                    return null;
-                }
-
                 if (SearchParamModifierMapping.TryGetValue(modifier, out SearchModifierCode searchModifierCode))
                 {
                     return searchModifierCode;
+                }
+
+                // Modifier on a Reference Search Parameter can be used to restrict target type
+                if (searchParameter.Type == SearchParamType.Reference && searchParameter.TargetResourceTypes.Contains(modifier, StringComparer.OrdinalIgnoreCase))
+                {
+                    return null;
                 }
 
                 throw new InvalidSearchOperationException(
--- a/src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/Parsers/ExpressionParser.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/Parsers/ExpressionParser.cs
@@ -265,31 +265,25 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Expressions.Parsers
 
         private Expression ParseSearchValueExpression(SearchParameterInfo searchParameter, string modifier, string value)
         {
-            SearchModifierCode? parsedModifier = null;
-            string targetTypeModifier = null;
-            if (!string.IsNullOrEmpty(modifier))
-            {
-                parsedModifier = ParseSearchParamModifier();
+            SearchModifier parsedModifier = ParseSearchParamModifier();
+            return _searchParameterExpressionParser.Parse(searchParameter, parsedModifier, value);
 
-                if (parsedModifier == SearchModifierCode.Type)
+            SearchModifier ParseSearchParamModifier()
+            {
+                if (string.IsNullOrEmpty(modifier))
                 {
-                    targetTypeModifier = modifier;
+                    return null;
                 }
-            }
 
-            return _searchParameterExpressionParser.Parse(searchParameter, parsedModifier, targetTypeModifier, value);
-
-            SearchModifierCode? ParseSearchParamModifier()
-            {
                 if (SearchParamModifierMapping.TryGetValue(modifier, out SearchModifierCode searchModifierCode))
                 {
-                    return searchModifierCode;
+                    return new SearchModifier(searchModifierCode);
                 }
 
                 // Modifier on a Reference Search Parameter can be used to restrict target type
                 if (searchParameter.Type == SearchParamType.Reference && searchParameter.TargetResourceTypes.Contains(modifier, StringComparer.OrdinalIgnoreCase))
                 {
-                    return SearchModifierCode.Type;
+                    return new SearchModifier(SearchModifierCode.Type, modifier);
                 }
 
                 throw new InvalidSearchOperationException(
--- a/src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/Parsers/ExpressionParser.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/Parsers/ExpressionParser.cs
@@ -271,7 +271,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Expressions.Parsers
             {
                 parsedModifier = ParseSearchParamModifier();
 
-                if (parsedModifier == null)
+                if (parsedModifier == SearchModifierCode.Type)
                 {
                     targetTypeModifier = modifier;
                 }
@@ -289,7 +289,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Expressions.Parsers
                 // Modifier on a Reference Search Parameter can be used to restrict target type
                 if (searchParameter.Type == SearchParamType.Reference && searchParameter.TargetResourceTypes.Contains(modifier, StringComparer.OrdinalIgnoreCase))
                 {
-                    return null;
+                    return SearchModifierCode.Type;
                 }
 
                 throw new InvalidSearchOperationException(
--- a/src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/Parsers/ISearchParameterExpressionParser.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/Parsers/ISearchParameterExpressionParser.cs
@@ -13,6 +13,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Expressions.Parsers
         Expression Parse(
             SearchParameterInfo searchParameter,
             SearchModifierCode? modifier,
+            string targetTypeModifier,
             string value);
     }
 }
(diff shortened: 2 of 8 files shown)


src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/Parsers/ExpressionParser.cs
```diff
@@ -265,20 +265,31 @@ private Expression ParseChainedExpression(string resourceType, SearchParameterIn
...
-                    return null;
+                    return searchModifierCode;
                 }
 
-                if (SearchParamModifierMapping.TryGetValue(modifier, out SearchModifierCode searchModifierCode))
+                // Modifier on a Reference Search Parameter can be used to restrict target type
+                if (searchParameter.Type == SearchParamType.Reference && searchParameter.TargetResourceTypes.Contains(modifier, StringComparer.OrdinalIgnoreCase))
                 {
-                    return searchModifierCode;
+                    return null;
```
brendankowitz, line 292: Should we return `SearchModifierCode.Type` here? It seems to match this the [description](http://hl7.org/fhir/R4/codesystem-search-modifier-code.html) 
> The search parameter only applies to the Resource Type specified as a modifier (e.g. the modifier is not actually :type, but :Patient etc.).

src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/Parsers/ISearchParameterExpressionParser.cs
```diff
@@ -13,6 +13,7 @@ public interface ISearchParameterExpressionParser
         Expression Parse(
             SearchParameterInfo searchParameter,
             SearchModifierCode? modifier,
+            string targetTypeModifier,
             string value);
```
Ivanidzo4ka, line 17: nit: Since you touching this method....
I would consider to change order of these parameters and provide default values for optional parameters.
Make necessary parameters first, and put all optional parameters in the back.

Code like that: `Parse(searchParameter, null, null, value)` looks much better when it's just `Parse(searchParameter, value);`

It's a bit of extra refactoring, but i think it code would look cleaner and it's worth it.

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/Expressions/Parsers/ExpressionParserTests.cs
```diff
@@ -329,7 +329,7 @@ private Expression SetupExpression(SearchParameterInfo searchParameter, string v
         {
             Expression expectedExpression = Substitute.For<Expression>();
 
-            _searchParameterExpressionParser.Parse(searchParameter, null, value).Returns(expectedExpression);
+            _searchParameterExpressionParser.Parse(searchParameter, null, null, value).Returns(expectedExpression);
```
Ivanidzo4ka, line 332: Any chance E2E test can be added?
Either to [https://github.com/microsoft/fhir-server/blob/main/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/BasicSearchTests.cs](https://github.com/microsoft/fhir-server/blob/main/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/BasicSearchTests.cs)
or [https://github.com/microsoft/fhir-server/blob/main/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/ReferenceSearchTests.cs](https://github.com/microsoft/fhir-server/blob/main/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/ReferenceSearchTests.cs)
(later sound more appropriate, but I can be wrong)

src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/Parsers/SearchValueExpressionBuilderHelper.cs
```diff
@@ -20,6 +20,7 @@ internal class SearchValueExpressionBuilderHelper : ISearchValueVisitor
 
         private string _searchParameterName;
         private SearchModifierCode? _modifier;
+        private string _targetTypeModifier;
```
Ivanidzo4ka, line 23: >_targetTypeModifier [](start = 23, length = 19)

I'm not quite understand why we need this field. All we do is assign value to it, but I can't find code which would read it.

src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/Parsers/SearchParameterExpressionParser.cs
```diff
@@ -202,17 +208,53 @@ public SearchParameterExpressionParser(IReferenceSearchValueParser referenceSear
...
                         searchValue);
                 }).ToArray();
 
                 return Expression.Or(expressions);
             }
+
+            ISearchValue ApplyTargetTypeModifier(ISearchValue source)
+            {
+                var referenceSearchValue = source as ReferenceSearchValue;
+                if (referenceSearchValue == null || string.IsNullOrEmpty(targetTypeModifier))
```
Ivanidzo4ka, line 228: >targetTypeModifier [](start = 73, length = 18)

Took me a while to figure out this is local function which doing variable capturing from another upper function.
It's reads a bit weird, to pass one value as parameter, but capture another one by context.
Maybe for readability reasons, can we pass  `targetTypeModifier` as parameter as well?

src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/Parsers/SearchValueExpressionBuilderHelper.cs
anyname2: My thinking was: `_modifier` does not represent the complete state, so we would need this field if we wanted to handle modifier in some way. But I guess it could be added later if we need it. What do you think?

Ivanidzo4ka: Maybe we can alter `void ISearchValueVisitor.Visit(ReferenceSearchValue reference)` method to check what `reference.ResourceType` is equal to `_targetTypeModifier` to be more safe...

src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/Parsers/SearchParameterExpressionParser.cs
brendankowitz: Is there any guidance on this type of thing in the .net docs? Else, I would say this is a nit, it is a design feature of local functions

src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/Parsers/SearchValueExpressionBuilderHelper.cs
brendankowitz: It almost feels like SearchModifierCode should be inside a new container class or tuple, i.e. `SearchModifier(SearchModifierCode, ResourceType = null)`, thoughts?

anyname2: > It almost feels like SearchModifierCode should be inside a new container class or tuple, i.e. `SearchModifier(SearchModifierCode, ResourceType = null)`, thoughts?

Good idea! Will add a class

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/Expressions/Parsers/ExpressionParserTests.cs
anyname2: Added in e041735

src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/Parsers/ISearchParameterExpressionParser.cs
anyname2: After adding `SearchModifier` class in f4409b5 the `modifier` is only used as an optional parameter in a few unit tests.
I think, I prefer them in the current order (which is the order they come in Http query).

src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/Parsers/ExpressionParser.cs
```diff
@@ -265,11 +265,10 @@ private Expression ParseChainedExpression(string resourceType, SearchParameterIn
 
         private Expression ParseSearchValueExpression(SearchParameterInfo searchParameter, string modifier, string value)
         {
-            SearchModifierCode? parsedModifier = ParseSearchParamModifier();
-
+            SearchModifier parsedModifier = ParseSearchParamModifier();
             return _searchParameterExpressionParser.Parse(searchParameter, parsedModifier, value);
 
-            SearchModifierCode? ParseSearchParamModifier()
+            SearchModifier ParseSearchParamModifier()
```
Ivanidzo4ka, line 271: >ParseSearchParamModifier [](start = 27, length = 24)

nit: Maybe it's worth to move this function which creates `SearchModifier` to `SearchModifier` itself?

Ivanidzo4ka: actually, disregard that comment, it would be much easier to convert that class to record as we moved to .net 5, and it a only place where we do parsing anyway.

Ivanidzo4ka (approved):
:shipit:

src/Microsoft.Health.Fhir.Core/Features/Search/SearchModifier.cs
```diff
@@ -0,0 +1,77 @@
...
+                SearchModifierCode == other.SearchModifierCode &&
+                string.Equals(ResourceType, other.ResourceType, StringComparison.OrdinalIgnoreCase);
+        }
+
+        public override int GetHashCode()
+        {
+            int h1 = SearchModifierCode.GetHashCode();
+            int h2 = ResourceType?.GetHashCode(StringComparison.OrdinalIgnoreCase) ?? 0;
+            return ((h1 << 5) + h1) ^ h2;
+        }
```
Ivanidzo4ka, line 70: nit: I'm not sure all this code is necessary. It's probably nice to have, but I don't think it's been used anywhere.

anyname2: The equals override is used in tests (by mock argument matchers). The rest is just compiler warnings/best practices for overriding Equals. We can remove most of it in .Net 5 by making this class a record.

src/Microsoft.Health.Fhir.Core/Features/Search/SearchModifier.cs
```diff
@@ -0,0 +1,77 @@
...
+    public sealed class SearchModifier : IEquatable<SearchModifier>
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="SearchModifier"/> class.
+        /// </summary>
+        /// <param name="searchModifierCode"><see cref="SearchModifierCode"/></param>
+        /// <param name="resourceType">Resource type used to constrain abstract modifier code (e.g. <see cref="SearchModifierCode.Type"/>)</param>
+        public SearchModifier(SearchModifierCode searchModifierCode, string resourceType = null)
+        {
+            EnsureArg.EnumIsDefined(searchModifierCode, nameof(searchModifierCode));
```
johnstairs, line 25: I would drop this check (a simple BenchmarkDotNet test where I comment it out reveals its overhead):

|         Method |       Mean |     Error |    StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------------- |-----------:|----------:|----------:|-------:|------:|------:|----------:|
| WithValidation | 193.409 ns | 3.6804 ns | 5.1595 ns | 0.0105 |     - |     - |      56 B |
|   NoValidation |   7.225 ns | 0.2155 ns | 0.2016 ns | 0.0061 |     - |     - |      32 B |

johnstairs: Or make it debug-only.

johnstairs (approved):

johnstairs:
Thanks @anyname2!

PR #1557, opened 2020-12-28
Disable image tag support for template collection reference

moria97:
## Description
Support digest only in template collection reference as image tags are mutable and can lead to unexpected results:

- We have added the pulled image to local cache for 30 minutes and when user modified the same tag, the image will not be refreshed until cache expires
- When another admin/user changes the image tag, the conversion result would be strange. 

The decision is to remove support for image tags for now and request with payload of image tags will result in 400 Bad Request ```Digest must be provided in image reference```

## Related issues
Addresses [AB#78458](https://microsofthealth.visualstudio.com/Health/_workitems/edit/78485).

## Testing
Add cases for image tags and digest.

## FHIR Team Checklist

Skip (fix bug)

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Messages/ConvertData/ConvertDataRequest.cs
+++ b/src/Microsoft.Health.Fhir.Core/Messages/ConvertData/ConvertDataRequest.cs
@@ -58,8 +58,8 @@ namespace Microsoft.Health.Fhir.Core.Messages.ConvertData
 
         /// <summary>
         /// Reference for template collection.
-        /// The format is "<registryServer>/<imageName>:<imageTag>" for template collection stored in container registries.
-        /// Also supports image digest as reference. Will use 'latest' if no tag or digest present.
+        /// It can be the default templates "microsofthealth/fhirconverter:default", or an image on Azure Container Registry with digest "<registryServer>/<imageName>@<imageDigest>".
+        /// Currently we don't support image tags because the content is mutable and may lead to unexpected results.
         /// </summary>
         public string TemplateCollectionReference { get; }
 
--- a/src/Microsoft.Health.Fhir.Shared.Web/appsettings.json
+++ b/src/Microsoft.Health.Fhir.Shared.Web/appsettings.json
@@ -71,7 +71,6 @@
                 "Enabled": false,
                 "ContainerRegistryServers": [],
                 "TemplateCollectionOptions": {
-                    "ShortCacheTimeSpan": "00:30:00",
                     "LongCacheTimeSpan": "01:00:00",
                     "TemplateCollectionSizeLimitMegabytes": 10
                 },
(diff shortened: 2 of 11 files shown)


src/Microsoft.Health.Fhir.Shared.Web/appsettings.json
```diff
@@ -71,7 +71,6 @@
                 "Enabled": false,
                 "ContainerRegistryServers": [],
                 "TemplateCollectionOptions": {
-                    "ShortCacheTimeSpan": "00:30:00",
```
LTA-Thinking, line 74: Why are we removing this setting?

LTA-Thinking:
I'm concerned about this being a breaking change. I don't know how many users have custom image repos set up, but this would be a breaking change for them.

LTA-Thinking:
As a bug fix I think not having a semver command is more appropriate than a skip command.

src/Microsoft.Health.Fhir.Core/Messages/ConvertData/ConvertDataRequest.cs
```diff
@@ -58,8 +58,8 @@ public class ConvertDataRequest : IRequest<ConvertDataResponse>
 
         /// <summary>
         /// Reference for template collection.
-        /// The format is "<registryServer>/<imageName>:<imageTag>" for template collection stored in container registries.
-        /// Also supports image digest as reference. Will use 'latest' if no tag or digest present.
+        /// It can be the default templates "microsofthealth/fhirconverter:default", or an image on Azure Container Registry with digest "<registryServer>/<imageName>@<imageDigest>".
```
brendankowitz, line 61: Is `:default` a tag? Is this the only tag that is supported?

brendankowitz: Oh I think I get it. It must be exactly `microsofthealth/fhirconverter:default` or a reference with an image digest.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ConvertDataTests.cs
```diff
@@ -135,9 +135,31 @@ public async Task GivenAValidRequest_ButTemplateReferenceIsInvalid_WhenConvertDa
...
         [InlineData("fakeacr.azurecr.io/template:default")]
         [InlineData("test.azurecr-test.io/template:default")]
-        [InlineData("test.azurecr.com/template@sha256:592535ef52d742f81e35f4d87b43d9b535ed56cf58c90a14fc5fd7ea0fbb8696")]
         [InlineData("*****####.com/template:default")]
         [InlineData("¶Š™œãý£¾.com/template:default")]
+        public async Task GivenAValidRequest_WithoutDigest_WhenConvertData_ShouldReturnBadRequest(string templateReference)
+        {
+            if (!_convertDataEnabled)
+            {
+                return;
```
brendankowitz, line 144: We can now use [SkippableTheory/Facts](https://github.com/AArnott/Xunit.SkippableFact/blob/master/test/Xunit.SkippableFact.Tests/SampleTests.cs#L51)

moria97: Awesome! Thanks!

moria97:
Close this pull request as we may want to keep tags.

PR #1560, opened 2020-12-29
Add bundle issues to export record

LTA-Thinking:
## Description
Currently an export job fails if an issue is added to the search bundle as the export job doesn't have a bundle to return. This PR adds a stand in for the bundle that can have issues added to it. These issues are then added to the export job record and can be seen by getting the export job status.

## Related issues
Addresses 77673.

## Testing
New and existing unit tests

## FHIR Team Checklist

Patch (Bug fix)

### DIFF
--- a/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Export/ExportJobTaskTests.cs
+++ b/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Export/ExportJobTaskTests.cs
@@ -18,6 +18,7 @@ using Microsoft.Extensions.Options;
 using Microsoft.Health.Core.Internal;
 using Microsoft.Health.Fhir.Core.Configs;
 using Microsoft.Health.Fhir.Core.Exceptions;
+using Microsoft.Health.Fhir.Core.Features.Context;
 using Microsoft.Health.Fhir.Core.Features.Operations;
 using Microsoft.Health.Fhir.Core.Features.Operations.Export;
 using Microsoft.Health.Fhir.Core.Features.Operations.Export.ExportDestinationClient;
@@ -74,6 +75,7 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Operations.Export
                 _resourceDeserializer,
                 null,
                 Substitute.For<IMediator>(),
+                Substitute.For<IFhirRequestContextAccessor>(),
                 NullLogger<ExportJobTask>.Instance);
         }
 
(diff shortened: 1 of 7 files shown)


src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Export/ExportJobTaskTests.cs
```diff
@@ -1607,6 +1620,7 @@ public async Task GivenAnUnKnownExceptionThrowFromAnonymizer_WhenExecuted_ThenJo
                 _resourceDeserializer,
                 factory.CreateMockScope(),
                 Substitute.For<IMediator>(),
+                Substitute.For<IFhirRequestContextAccessor>(),
```
brendankowitz, line 1623: There constructor looks like it is repeated a lot

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -123,6 +129,17 @@ public async Task ExecuteAsync(ExportJobRecord exportJobRecord, WeakETag weakETa
...
+                // Add a request context so that bundle issues can be added by the SearchOptionFactory
+                var fhirRequestContext = new FhirRequestContext(
+                method: "Export",
+                uriString: "$export",
+                baseUriString: "$export",
+                correlationId: _exportJobRecord.Id,
+                requestHeaders: new Dictionary<string, StringValues>(),
+                responseHeaders: new Dictionary<string, StringValues>());
+
+                _contextAccessor.FhirRequestContext = fhirRequestContext;
```
brendankowitz, line 141: Should we only set this if its null?

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Export/ExportJobTaskTests.cs
LTA-Thinking: Yea, I can replace that with a method call. Most of them are identical.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
LTA-Thinking: It has always been null in my testing. I'm concerned about adding a null check as this works on the assumption that the request context doesn't matter outside of recording issues.

brendankowitz: My general rule of thumb/concern is mutating state on an object thats outside of the creating or owning component. In this case _contextAccessor is not set because it's outside of the http pipeline, but the point being ExportJobTask has no visibility if other components require this component in this scope. I'd say to keep the assumptions of other components (that you can't see), the right pattern here is to save the existing value, set your new one, then set it back after.

brendankowitz (approved):

PR #1569, opened 2021-01-06
SQL search renaming/refactoring

johnstairs:
## Description
This PR intends to simplify the code for SQL search by eliminating "Normalized" and "Denormalized" concepts. These were meant to represent columns that were on the search parameter tables and the Resource table, respectively, but created confusion because some (but not all) denormalized columns were also on the search parameter tables.

- Adding a `SearchParameterColumnLocation` enum to identify where the columns for a search parameter reside. 
- `SqlRootExpression` has properties `TableExpressions` and `ResourceExpressions` instead of `TableExpressions` and `DenormalizedExpressions`.
- `TableExpression` consolidates `NormalizedPredicate` and `DenormalizedPredicate` properties into a single `Predicate` property.
- Introducing `SqlChainLinkExpression`, which is created from `ChainExpressions` but includes a `ExpressionOnSource` property. This means we can get rid of `TableExpression.DenormalizedPredicateOnChainRoot`, which polluted the `TableExpression` class with a concept that only applies to chaining.

There should not be any functional changes with in this PR. It is strictly a refactoring (well, mostly a renaming).


CustomAnalysisRules.ruleset
```diff
@@ -5,7 +5,7 @@
   <Rules AnalyzerId="Microsoft.CodeAnalysis.CSharp.Features" RuleNamespace="Microsoft.CodeAnalysis.CSharp.Features">
     <Rule Id="IDE0003" Action="Warning" />
     <Rule Id="IDE0004" Action="Warning" />
-    <Rule Id="IDE0005" Action="Warning" />
```
Ivanidzo4ka, line 8: >Warning [](start = 31, length = 7)

I actually like this rule, it keep your using statement clean. Any reason to ignore it?

johnstairs: I the rule too, but I don't like the way it shows up in your IDE sporadically based on whether or not you have the file open. I would prefer it to run on build, for for that it appears we need to turn XML documentation generation. Maybe there is another rule that covers this?

johnstairs: And if we turn on XML documentation generation, we get a bunch of errors because of malformed xml, missing documentation, etc. I am also concerned that this will hurt compilation speed.
If you want to keep this, I can just put back in IDE0005  (at least for now).

Here is the compiler bug: https://github.com/dotnet/roslyn/issues/41640

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SearchParameterColumnLocation.cs
```diff
@@ -0,0 +1,19 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+
+namespace Microsoft.Health.Fhir.SqlServer.Features.Search
+{
+    /// <summary>
+    /// Indicates where the column(s) of for a search parameter are located.
```
Ivanidzo4ka, line 11: >of for [](start = 38, length = 6)

super nit: two prepositions together look weird for me, but English is not my native.

johnstairs: Ha ha, its definitely not proper English! Will fix.

CustomAnalysisRules.ruleset
johnstairs: I've reverted the change.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/SqlRootExpression.cs
```diff
@@ -13,43 +13,41 @@
...
             EnsureArg.IsNotNull(tableExpressions, nameof(tableExpressions));
-            EnsureArg.IsNotNull(denormalizedExpressions, nameof(denormalizedExpressions));
+            EnsureArg.IsNotNull(resourceExpressions, nameof(resourceExpressions));
 
             TableExpressions = tableExpressions;
-            DenormalizedExpressions = denormalizedExpressions;
+            ResourceExpressions = resourceExpressions;
         }
 
         public IReadOnlyList<TableExpression> TableExpressions { get; }
```
Ivanidzo4ka, line 31: >Table [](start = 46, length = 5)

nit: maybe it's worth to rename this one as well. Table is such a generic name, plus Resource table is also technically table.
`SearchParamExpressions` maybe?

I would also add summary 
```
<summary>
List of expression performed on various search parameter tables (e.g. TokenSearchParam, NumberSearchParam, etc)
</summary>
``` for this property and
```
<summary>
List of expression applied to directly to Resource table.
<summary>
``` for ResourceExpressions

File: src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/ChainFlatteningRewriter.cs [hunk 4]
```diff
@@ -19,65 +17,12 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Search.Expressions.Visitors
     /// </summary>
     internal class ChainFlatteningRewriter : SqlExpressionRewriterWithInitialContext<(TableExpression containingTableExpression, int chainLevel)>
```
Ivanidzo4ka, line 18: >(TableExpression containingTableExpression, int chainLevel) [](start = 85, length = 59)

Probably not related to PR, but I'm scratching my head when I look on this type.

I've just change it to object and one test which I ran which hits this part is passing. 
What's the point of this complex type if it doesn't used anywhere?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/SqlRootExpression.cs
johnstairs: Ok and keep the `TableExpression` type as-is?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/ChainFlatteningRewriter.cs
johnstairs: Good catch (I assume you mean the `<(TableExpression containingTableExpression, int chainLevel)>` type parameter). It was used before, but not anymore. It can be reverted to object.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/SqlRootExpression.cs
Ivanidzo4ka: Naming is tricky. -_-"
I wouldn't mind to rename it to `SearchParamTableExpression` as type and `SearchParamTableExpressions`, but there is comment on top of `TableExpression`: `An expression over a search param or **compartment** table`.

It's ok-ish the way it is right now, but I would still add summary comments for public properties.


---
In reply to: [553618664](https://github.com/microsoft/fhir-server/pull/1569#discussion_r553618664) [](ancestors = 553618664)

johnstairs: Yeah. TableExpressions match 1:1 with the CTEs that we generate, so that's where the type name comes from. But I think `SearchParamTableExpressions` and `ResourceTableExpressions` could work. (I think we can address the compartment exception with a comment). What do you think?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/ChainFlatteningRewriter.cs
Ivanidzo4ka: Yep, sorry for confusion, I was under impression I selected part of line with type, not whole line, but I guess I fat fingered it. (or fat moused if that is the thing)

---
In reply to: [553619485](https://github.com/microsoft/fhir-server/pull/1569#discussion_r553619485) [](ancestors = 553619485)

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/SqlRootExpression.cs
Ivanidzo4ka: Sounds good to me.

---
In reply to: [553631371](https://github.com/microsoft/fhir-server/pull/1569#discussion_r553631371) [](ancestors = 553631371)

johnstairs: Ok, I've done this in https://github.com/microsoft/fhir-server/pull/1569/commits/5eebddf2a917f25193a5a0f755e003ccf78e7855

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/ChainFlatteningRewriter.cs
johnstairs: Fixed in https://github.com/microsoft/fhir-server/pull/1569/commits/427a33aba193c6f5c6c9fe659774fc90da51e64f

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SearchParameterQueryGenerator.cs
```diff
@@ -91,6 +108,21 @@ public override SearchParameterQueryGeneratorContext VisitMultiary(MultiaryExpre
             return context;
         }
 
+        protected static SearchParameterQueryGenerator GetSearchParameterQueryGeneratorIfResourceColumnSearchParameter(SearchParameterExpressionBase searchParameter)
+        {
+            switch (searchParameter.Parameter.Name)
```
Ivanidzo4ka, line 113: >searchParameter.Parameter.Name [](start = 20, length = 30)

This look very similar to `ColumnLocation` in `SearchParameterInfoExtensions`.
Is lastUpdated got omitted here because it has special handling case somewhere else?

johnstairs: At this point, LastUpdated should have been replaced by ResourceSurrogateId. I could add a Debug.Assert or throw if we encounter LastUpdated.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/ResourceColumnPredicatePushdownRewriter.cs
```diff
@@ -0,0 +1,100 @@
...
+        public static readonly ResourceColumnPredicatePushdownRewriter Instance = new ResourceColumnPredicatePushdownRewriter();
+
+        public override Expression VisitSqlRoot(SqlRootExpression expression, object context)
+        {
+            if (expression.SearchParamTableExpressions.Count == 0 || expression.ResourceTableExpressions.Count == 0 ||
+                expression.SearchParamTableExpressions.All(e => e.Kind == SearchParamTableExpressionKind.Include))
+            {
+                // if only Include expressions, the case is handled in IncludeMatchSeedRewriter
+                return expression;
+            }
```
Ivanidzo4ka, line 30: nit: maybe it's worth to push that check directly into `IncludeMatchSeedRewriter` as a static method?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/IncludeMatchSeedRewriter.cs
```diff
@@ -0,0 +1,44 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Health.Fhir.Core.Features.Search.Expressions;
+
+namespace Microsoft.Health.Fhir.SqlServer.Features.Search.Expressions.Visitors
+{
+    /// <summary>
+    /// Rewriter used to add an All SearchParamTableExpression to serve as a seed for match results when the SearchParamTableExpressions in a SqlRootExpression
```
Ivanidzo4ka, line 14: >seed for match [](start = 77, length = 14)

I see that 'seed' in few other places, but i'm not sure I quite understand idea behind it. Any chance you can comment what exactly seed is?

johnstairs: Yeah, we should improve the comments here. 

Suppose you have an include query like `Observation?code=abc&_include=Observation:subject`. The include will be based on the CTE that gets the IDs of the first 10 observations where the code is `abc`.

Now suppose you have simply `Observation?_include=Observation:subject`. Here there is no CTE to base the include on, so we inject one that yields all Observations (well, the first 10). This is what I'm calling "seeding" the include.

Ivanidzo4ka (approved):
:shipit:

PR #1571, opened 2021-01-07
Improves performace of count queries

johnstairs:
For the SQL provider, we were adding an unnecessary `JOIN` with the `Resource` table in many `_summary=count` queries. Removing the join improves the efficiency in a significant way. Here are two examples:

|                                                                      | Count  | BEFORE_READS | AFTER_READS |   |
|----------------------------------------------------------------------|--------|--------------|-------------|---|
| /Observation?status=final&_summary=count                             | 543671 | 20334        | 3170        |   |
| /Patient?name=john&_summary=count                                    | 26     | 108          | 4           |   |

For the first query, the difference between the SQL is:
image

(We would get further improvements by eliminating `DISTINCT` when possible)

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
@@ -102,7 +102,7 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Search.Expressions.Visitors.Q
                     selectingFromResourceTable = false;
                     StringBuilder.AppendLine("SELECT COUNT(DISTINCT Sid1)");
                 }
-                else if (expression.TableExpressions.Count == 0)
+                else
                 {
                     // We will be counting over the Resource table.
                     selectingFromResourceTable = true;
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -131,59 +151,68 @@ public object VisitSqlRoot(SqlRootExpression expression, SearchOptions context)
...
-            }
+                StringBuilder.Append("FROM ").Append(VLatest.Resource).Append(" ").Append(resourceTableAlias);
 
-            using (var delimitedClause = StringBuilder.BeginDelimitedWhereClause())
-            {
-                foreach (var denormalizedPredicate in expression.DenormalizedExpressions)
+                if (expression.TableExpressions.Count == 0 &&
+                    !_isHistorySearch &&
+                    expression.DenormalizedExpressions.Any(e => e.AcceptVisitor(ExpressionContainsParameterVisitor.Instance, SearchParameterNames.ResourceType)) &&
+                    !expression.DenormalizedExpressions.Any(e => e.AcceptVisitor(ExpressionContainsParameterVisitor.Instance, SearchParameterNames.Id)))
```
johnstairs, line 161: This is not the index to use for this kind of query:
`GET https://localhost:44348/Observation?_id=a937ba06-da1e-4f37-bbb4-848d60f26aed` because `ResourceId` is not part of this index.

namadabu (approved):
:shipit:

PR #1577, opened 2021-01-08
Makes reindex job tests independent of data store

rotodd:
## Description
The reindex job tests were setup to use the Cosmos DB search service. This PR moves the search service setup logic to the datastore-specific test fixtures, which will allow the tests to run for SQL as well.

These tests are not yet enabled for SQL - they should be enabled and should pass once SQL reindexing is fully implemented.

## Related issues
Addresses [AB#78696](https://microsofthealth.visualstudio.com/Health/_workitems/edit/78696).

It might be nice to rethink how our integration tests are setup, since manually instantiating services is not very pretty or scalable. Created [AB#78697](https://microsofthealth.visualstudio.com/Health/_workitems/edit/78697) to track this.

## Testing
Manual testing:
- Ensured that the tests continue to pass for Cosmos DB
- Temporarily enabled the tests for SQL to ensure that the tests run without fixture-related issues (although some tests fail, as expected)

## FHIR Team Checklist

### DIFF
--- a/test/Microsoft.Health.Fhir.Shared.Tests.Integration/Features/Operations/Reindex/ReindexJobTests.cs
+++ b/test/Microsoft.Health.Fhir.Shared.Tests.Integration/Features/Operations/Reindex/ReindexJobTests.cs
@@ -6,19 +6,18 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Net;
 using System.Threading;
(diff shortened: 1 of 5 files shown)


johnstairs (approved):
Agree that we could do better than setting up all the services manually!

feordin (approved):

PR #1581, opened 2021-01-12
Add support to DELETE custom search parameters

feordin:
## Description
Adds the support to the DeleteResourceHandler to remove custom search parameter from the SearchParameterDefinitionManager collections.

## Related issues
Addresses [issue [AB#76347](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/76347)].

## Testing
Unit test added. E2E Test updated with DELETE.  Manual testing performed.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Extensions/RawResourceExtensions.cs
+++ b/src/Microsoft.Health.Fhir.Core/Extensions/RawResourceExtensions.cs
@@ -1,0 +1,51 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.IO;
+using EnsureThat;
+using Hl7.Fhir.ElementModel;
+using Hl7.Fhir.Serialization;
+using Microsoft.Health.Fhir.Core.Exceptions;
+using Microsoft.Health.Fhir.Core.Features.Persistence;
+using Microsoft.Health.Fhir.Core.Models;
+using Newtonsoft.Json;
+
+namespace Microsoft.Health.Fhir.Core.Extensions
+{
+    public static class RawResourceExtensions
+    {
+        /// <summary>
+        /// Converts the RawResource to an ITypedElement
+        /// </summary>
+        /// <param name="rawResource">The RawResource to convert</param>
+        /// <param name="modelInfoProvider">The IModelInfoProvider to use when converting the RawResource</param>
+        /// <returns>An ITypedElement of the RawResource</returns>
+        public static ITypedElement ToITypedElement(this RawResource rawResource, IModelInfoProvider modelInfoProvider)
+        {
+            EnsureArg.IsNotNull(rawResource, nameof(rawResource));
+            EnsureArg.IsNotNull(modelInfoProvider, nameof(modelInfoProvider));
+
+            using TextReader reader = new StringReader(rawResource.Data);
+            using JsonReader jsonReader = new JsonTextReader(reader);
+            try
+            {
+                ISourceNode sourceNode = FhirJsonNode.Read(jsonReader);
+                return modelInfoProvider.ToTypedElement(sourceNode);
+            }
+            catch (FormatException ex)
+            {
+                var issue = new OperationOutcomeIssue(
+                    OperationOutcomeConstants.IssueSeverity.Fatal,
+                    OperationOutcomeConstants.IssueType.Invalid,
+                    ex.Message);
+
+                throw new InvalidDefinitionException(
+                    Core.Resources.SearchParameterDefinitionContainsInvalidEntry,
+                    new OperationOutcomeIssue[] { issue });
+            }
+        }
+    }
+}
--- a/src/Microsoft.Health.Fhir.Core/Features/Definition/BundleWrappers/SearchParameterWrapper.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Definition/BundleWrappers/SearchParameterWrapper.cs
@@ -14,7 +14,7 @@ using Microsoft.Health.Fhir.Core.Models;
 
 namespace Microsoft.Health.Fhir.Core.Features.Definition.BundleWrappers
 {
-    public class SearchParameterWrapper
+    internal class SearchParameterWrapper
     {
         private readonly Lazy<string> _url;
         private readonly Lazy<IReadOnlyList<ITypedElement>> _component;
--- a/src/Microsoft.Health.Fhir.Core/Features/Definition/BundleWrappers/SearchParameterWrapper.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Definition/BundleWrappers/SearchParameterWrapper.cs
@@ -45,7 +45,9 @@ namespace Microsoft.Health.Fhir.Core.Features.Definition.BundleWrappers
 
         public string Description => _description.Value;
 
+#pragma warning disable CA1056 // Uri properties should not be strings
         public string Url => _url.Value;
+#pragma warning restore CA1056 // Uri properties should not be strings
 
         public string Type => _type.Value;
 
--- a/src/Microsoft.Health.Fhir.Core/Features/Definition/BundleWrappers/SearchParameterWrapper.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Definition/BundleWrappers/SearchParameterWrapper.cs
@@ -14,7 +14,7 @@ using Microsoft.Health.Fhir.Core.Models;
 
 namespace Microsoft.Health.Fhir.Core.Features.Definition.BundleWrappers
 {
-    internal class SearchParameterWrapper
+    public class SearchParameterWrapper
     {
         private readonly Lazy<string> _url;
         private readonly Lazy<IReadOnlyList<ITypedElement>> _component;
--- a/src/Microsoft.Health.Fhir.Core/Features/Definition/BundleWrappers/SearchParameterWrapper.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Definition/BundleWrappers/SearchParameterWrapper.cs
@@ -45,9 +45,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Definition.BundleWrappers
 
         public string Description => _description.Value;
 
-#pragma warning disable CA1056 // Uri properties should not be strings
         public string Url => _url.Value;
-#pragma warning restore CA1056 // Uri properties should not be strings
 
         public string Type => _type.Value;
 
--- a/src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterDefinitionManager.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterDefinitionManager.cs
@@ -13,6 +13,7 @@ using System.Threading.Tasks;
 using EnsureThat;
 using Hl7.Fhir.ElementModel;
 using Microsoft.Extensions.Hosting;
+using Microsoft.Health.Fhir.Core.Features.Definition.BundleWrappers;
 using Microsoft.Health.Fhir.Core.Features.Search;
 using Microsoft.Health.Fhir.Core.Models;
 
@@ -150,6 +151,28 @@ namespace Microsoft.Health.Fhir.Core.Features.Definition
                     searchParamHash,
                     (resourceType, existingValue) => searchParamHash);
             }
+        }
+
+        public void DeleteSearchParameter(ITypedElement searchParam)
+        {
+            var searchParamWrapper = new SearchParameterWrapper(searchParam);
+
+            if (!UrlLookup.Remove(new Uri(searchParamWrapper.Url)))
+            {
+                throw new Exception(string.Format(Resources.CustomSearchParameterNotfound, searchParamWrapper.Url));
+            }
+
+            var searchParameterInfo = TypeLookup[searchParamWrapper.Type][searchParamWrapper.Name];
+            var allResourceTypes = searchParameterInfo.TargetResourceTypes.Union(searchParameterInfo.BaseResourceTypes);
+            foreach (var resourceType in allResourceTypes)
+            {
+                if (TypeLookup.ContainsKey(resourceType))
+                {
+                    TypeLookup[resourceType].Remove(searchParameterInfo.Name);
+                }
+            }
+
+            CalculateSearchParameterHash();
         }
     }
 }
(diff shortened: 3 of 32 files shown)


src/Microsoft.Health.Fhir.Core/Extensions/RawResourceExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+{
+    public static class RawResourceExtensions
+    {
+        /// <summary>
+        /// Converts the RawResource to an ITypedElement
+        /// </summary>
+        /// <param name="rawResource">The RawResource to convert</param>
+        /// <param name="modelInfoProvider">The IModelInfoProvider to use when converting the RawResource</param>
+        /// <returns>An ITypedElement of the RawResource</returns>
+        public static ITypedElement ToITypedElement(this RawResource rawResource, IModelInfoProvider modelInfoProvider)
```
Ivanidzo4ka, line 26: >ToITypedElement [](start = 36, length = 15)

This looks quite similar to `ReadEmbeddedSearchParameters` in `SearchParameterDefinitionBuilder`
Is it worth to consider using this method in that part of code?

src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterDefinitionManager.cs
```diff
@@ -151,5 +152,27 @@ private void CalculateSearchParameterHash()
...
+                throw new Exception(string.Format(Resources.CustomSearchParameterNotfound, searchParamWrapper.Url));
+            }
+
+            var searchParameterInfo = TypeLookup[searchParamWrapper.Type][searchParamWrapper.Name];
+            var allResourceTypes = searchParameterInfo.TargetResourceTypes.Union(searchParameterInfo.BaseResourceTypes);
+            foreach (var resourceType in allResourceTypes)
+            {
+                if (TypeLookup.ContainsKey(resourceType))
+                {
+                    TypeLookup[resourceType].Remove(searchParameterInfo.Name);
```
Ivanidzo4ka, line 171: >.Remove [](start = 44, length = 7)

How we protect ourselves from concurrency issues?
From what I can see in code, that `IDictionary<string, SearchParameterInfo>` is just a regular Dictionary, and it's ok if you instantiate it once during startup and just do lookup for it, but now we adding and removing elements from it, and I'm not sure is it locked for one thread in upstream or just not locked at at all.

src/Microsoft.Health.Fhir.Core/Resources.resx
```diff
@@ -515,4 +515,14 @@
...
   </data>
-</root>
+  <data name="CustomSearchDeleteError" xml:space="preserve">
+    <value>An error occurred deleting the custom search parameter.  The issue must be resolved and the delete request resubmitted to remove the parameter.</value>
+  </data>
+  <data name="CustomSearchParameterNotfound" xml:space="preserve">
+    <value>Custom Search parameter with Uri: {0} was not found.</value>
+    <comment>{0} the uri identifying the custom search parameter.</comment>
+  </data>
+  <data name="NoResourcesNeedbeReindexed" xml:space="preserve">
```
Ivanidzo4ka, line 525: >be [](start = 29, length = 2)

super nit: typo?

src/Microsoft.Health.Fhir.Core/Features/Definition/BundleWrappers/SearchParameterWrapper.cs
```diff
@@ -14,7 +14,7 @@
 
 namespace Microsoft.Health.Fhir.Core.Features.Definition.BundleWrappers
 {
-    internal class SearchParameterWrapper
+    public class SearchParameterWrapper
```
Ivanidzo4ka, line 17: >public [](start = 4, length = 6)

Is this change necessary?

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/DeleteSearchParameterBehavior.cs
```diff
@@ -0,0 +1,54 @@
...
+            {
+                searchParamResource = await _fhirDataStore.GetAsync(deleteRequest.ResourceKey, cancellationToken);
+            }
+
+            var response = await next();
+
+            if (searchParamResource != null)
+            {
+                // Once the SearchParameter resource is removed from the data store, we can update the in
+                // memory SearchParameterDefinitionManager, and persist the deleted status to the data store
```
Ivanidzo4ka, line 47: >persist the deleted status to the data store [](start = 64, length = 44)

Didn't we remove it from data store already?
This part a bit confusing.

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/DeleteSearchParameterBehavior.cs
```diff
@@ -0,0 +1,54 @@
...
+        {
+            var deleteRequest = request as DeleteResourceRequest;
+            ResourceWrapper searchParamResource = null;
+
+            if (deleteRequest.ResourceKey.ResourceType.Equals(KnownResourceTypes.SearchParameter, StringComparison.Ordinal))
+            {
+                searchParamResource = await _fhirDataStore.GetAsync(deleteRequest.ResourceKey, cancellationToken);
+            }
+
+            var response = await next();
```
Ivanidzo4ka, line 42: >next [](start = 33, length = 4)

nit:
Part of me want to argue about order of things.
So if correctly understand there is two parts, storing search parameter in data store, and storing search parameter in memory of the server.
By having search parameter in memory, we apply it to all incoming data, and i think we should purge it from memory first and then try to remove it from server. If someone try to delete search parameter i would assume they don't want it to be applied anymore, so it sound a bit unnatural for me to remove it from memory as second step.
Also if I look into adding search parameter order is : add to data store, then add to memory.
Considering what delete operation is opposite of adding,  it doesn't feel like it should follow same order, and should be opposite.

src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterStatusManager.cs
```diff
@@ -126,5 +126,11 @@ public async Task AddSearchParameterStatusAsync(IReadOnlyCollection<string> sear
             // they will be fully enabled
             await UpdateSearchParameterStatusAsync(searchParamUris, SearchParameterStatus.Supported);
         }
+
+        internal async Task DeleteSearchParameterStatusAsync(string url)
```
Ivanidzo4ka, line 130: >internal [](start = 8, length = 8)

nit: it's a bit strange to see Delete as `internal` and Add as `public`.
Can they be synced?

src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterDefinitionManager.cs
```diff
@@ -142,14 +143,33 @@ public void AddNewSearchParameters(IReadOnlyCollection<ITypedElement> searchPara
...
+
+            if (!UrlLookup.TryRemove(new Uri(searchParamWrapper.Url), out searchParameterInfo))
+            {
+                throw new Exception(string.Format(Resources.CustomSearchParameterNotfound, searchParamWrapper.Url));
+            }
+
+            var allResourceTypes = searchParameterInfo.TargetResourceTypes.Union(searchParameterInfo.BaseResourceTypes);
+            foreach (var resourceType in allResourceTypes)
+            {
+                TypeLookup[resourceType].TryRemove(searchParameterInfo.Name, out var removedParam);
```
Ivanidzo4ka, line 169: >TryRemove [](start = 41, length = 9)

if all `TryRemove` return false, it's probably doesn't make sense to do `CalculateSearchParameterHash`

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/CustomSearchParamTests.cs
```diff
@@ -44,7 +44,25 @@ public async Task GivenANewSearchParam_WhenReindexingComplete_ThenResourcesSearc
...
+                // if the SearchParameter exists, we should delete it and recreate it
+                var searchParamBundle = await Client.SearchAsync(ResourceType.SearchParameter, $"url={searchParam.Url}");
+                if (searchParamBundle.Resource?.Entry[0] != null && searchParamBundle.Resource?.Entry[0].Resource.ResourceType == ResourceType.SearchParameter)
+                {
+                    await DeleteSearchParameterAndVerify(searchParamBundle.Resource?.Entry[0].Resource as SearchParameter);
+                    searchParamPosted = await Client.CreateAsync(searchParam);
+                }
+                else
+                {
+                    throw ex;
```
Ivanidzo4ka, line 63: >throw ex [](start = 20, length = 8)

nit: I think this is antipattern.
`ex` is not used anywhere and just got rethrown which would lead to lost of original stack. Not that important in tests, but i think it's nice thing to avoid overall.

here some details:
[https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2200](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2200)

src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterDefinitionManager.cs
feordin: This is true, but if we passed the first attempt to remove a SearchParameter from the UrlLookup, we should have at least one entry in the TypeLookup to remove.

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/DeleteSearchParameterBehavior.cs
feordin: I had the same thought about order and considered the alternate approach.  Currently the call to DeleteSearchParameterStatusAsync uses the metadata stored in the SearchParameterDefinitionManager to then go and remove the status from the data store.  So, because of dependencies, it does need to happen in this order.

Ivanidzo4ka: Would it worth to add comment regarding order? I think we both agree it's a bit counterintuitive, so maybe it's worth to leave some information here.

---
In reply to: [560619433](https://github.com/microsoft/fhir-server/pull/1581#discussion_r560619433) [](ancestors = 560619433)

Ivanidzo4ka:
>                 Assert.True(delayCount <= 9);

I think there is change to not run job if it doesn't have any matching resources, and fails here.

Probably according to new contract this job would be considered `Cancelled`
```
 await CompleteJobAsync(OperationStatus.Canceled, cancellationToken);
```
instead of `Completed`

---
Refers to: test/Microsoft.Health.Fhir.Shared.Tests.Integration/Features/Operations/Reindex/ReindexJobTests.cs:224 in f87a5ff. [](commit_id = f87a5ff33e317f75d90b5f4cbe1803f9353ca66f, deletion_comment = False)

Ivanidzo4ka (approved):
:shipit:

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/CreateSearchParameterBehavior.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Threading.Tasks;
+using EnsureThat;
+using MediatR;
+using Microsoft.Health.Fhir.Core.Messages.Create;
+using Microsoft.Health.Fhir.Core.Models;
+
+namespace Microsoft.Health.Fhir.Core.Features.Search.Parameters
+{
+    public class CreateSearchParameterBehavior<TCreateResourceRequest, TUpsertResourceResponse> : IPipelineBehavior<TCreateResourceRequest, TUpsertResourceResponse>
+        where TCreateResourceRequest : CreateResourceRequest
```
Ivanidzo4ka, line 17: >TCreateResourceRequest [](start = 14, length = 22)

Silly question, but should it work with conditional update/create?

src/Microsoft.Health.Fhir.Core/Extensions/RawResourceExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+            }
+            catch (FormatException ex)
+            {
+                var issue = new OperationOutcomeIssue(
+                    OperationOutcomeConstants.IssueSeverity.Fatal,
+                    OperationOutcomeConstants.IssueType.Invalid,
+                    ex.Message);
+
+                throw new InvalidDefinitionException(
+                    Core.Resources.SearchParameterDefinitionContainsInvalidEntry,
```
brendankowitz, line 46: This error message doesn't seem correct here

src/Microsoft.Health.Fhir.Core/Features/Definition/BundleWrappers/SearchParameterWrapper.cs
```diff
@@ -45,7 +45,9 @@ public SearchParameterWrapper(ITypedElement searchParameter)
 
         public string Description => _description.Value;
 
+#pragma warning disable CA1056 // Uri properties should not be strings
```
brendankowitz, line 48: Disable with attribute instead

src/Microsoft.Health.Fhir.Core/Extensions/RawResourceExtensions.cs
brendankowitz, line 26: By switching this around it could become an overload, that might make discovery easier, thoughts?

```suggestion
        public static ITypedElement ToTypedElement(this IModelInfoProvider modelInfoProvider, RawResource rawResource)
```

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/CreateSearchParameterBehavior.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using EnsureThat;
+using MediatR;
+using Microsoft.Health.Fhir.Core.Messages.Create;
+using Microsoft.Health.Fhir.Core.Models;
+
+namespace Microsoft.Health.Fhir.Core.Features.Search.Parameters
+{
+    public class CreateSearchParameterBehavior<TCreateResourceRequest, TUpsertResourceResponse> : IPipelineBehavior<TCreateResourceRequest, TUpsertResourceResponse>
```
brendankowitz, line 16: Unless the generic is needed for derived types, I think this should still work with:
```csharp
CreateSearchParameterBehavior : IPipelineBehavior<CreateResourceRequest, UpsertResourceResponse>
```

feordin: I may want to update this for conditionalcreate, so I wanted to leave it generic for now.

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/CreateSearchParameterBehavior.cs
feordin: I hope so :), I have not net yet tried it.

src/Microsoft.Health.Fhir.Core/Extensions/RawResourceExtensions.cs
feordin: I agree, I have updated it as an overload on ModelInfoProvider.

Issue #1596, opened 2021-01-14
Export fails when multiple types are given


LTA-Thinking:
**Describe the bug**
When an export job with the _type parameter and multiple types the job will export no resources.

**FHIR Version?**
Stu3/R4/R5

**Data provider?**
CosmosDB/SQL Server

**To Reproduce**
1. <server>/$export?_type=Observation,Encounter

**Expected behavior**
All the listed resource types are exported.

**Actual behavior**
No resources are exported.

LTA-Thinking:
This was caused by the typeFilter feature which changed the way _type was parsed.

PR #1597, opened 2021-01-14
Fix export _type

LTA-Thinking:
## Description
When using _type with multiple resources no resources are exported.

## Related issues
Addresses #1596 

## Testing
New unit tests

## FHIR Team Checklist

Patch (Bug)


namadabu:
Should we add an E2E test for this scenario? Seems like we could use a modification of `GivenFhirServer_WhenGroupObervationDataIsExported_ThenExportedDataIsSameAsDataInFhirServer` with multiple types instead of only `Encounter`. That should cover the above scenario

brendankowitz (approved):
Looks good, though add a unit or regression test if possible

namadabu (approved):

brendankowitz (approved):

Issue #1566, opened 2021-01-05
Add support for R4 type Canonical => Reference Searches


brendankowitz:
**User story**
Add support for the R4 primitive type [Canonical](https://www.hl7.org/fhir/datatypes.html).

**Acceptance criteria**
1. Support is visible at the /metadata endpoint
1. Canonical urls are indexed correctly
1. Verify resources can be searched by QuestionnaireResponse.questionnaire, including ?questionnaire=http://url|version

brendankowitz:
Related to #882

CaitlinV39:
AB#76717

kyle-mwnz:
Hi there, are there any updates in regards to this?

This feature is a dealbreaker for us moving from an AWS solution over to Azure, as our API clients must be able to search for certain resources by their canonical identifier (e.g. Questionnaire resources by canonical url). At present attempting this on a Saas health data services, an OperationOutcome with

```
        {
            "resource": {
                "resourceType": "OperationOutcome",
                "id": "574e8fd763dac8a410252a26cfe07e7d",
                "issue": [
                    {
                        "severity": "warning",
                        "code": "not-supported",
                        "diagnostics": "The search parameter 'questionnaire' is not supported for resource type 'QuestionnaireResponse'."
                    }
                ]
            },
            "search": {
                "mode": "outcome"
            }
        },
```

Is returned

mauriceli-unsw:
Does the following Canonical search work now?

- `definition` in `PlanDefinition`
- `instantiates-canonical` in `CarePlan`

evachen96:
@mauriceli-unsw  This is still scheduled in our backlog.

Issue #882, opened 2020-02-12
Add support for R4 type Canonical => URI searches


brendankowitz:
**User story**
Add support for the R4 primitive type [Canonical](https://www.hl7.org/fhir/datatypes.html).

**Acceptance criteria**
1. Support is visible at the /metadata endpoint
1. Canonical urls are indexed correctly
1. Verify resources can be searched by _profile, including _profile=http://url|version
1. Verify resources can be searched by Uri, e.g. ConceptMap.source(uri)

CaitlinV39:
AB#69418

CaitlinV39:
AB#76717

CaitlinV39:
AB#76852

CaitlinV39:
AB#76852

brendankowitz:
Spec issue that may affect this in future: https://jira.hl7.org/browse/FHIR-30514

PR #1605, opened 2021-01-20
Adds Canonical, Url, Uri search support

brendankowitz:
## Description

- Canonical => Uri Support
- Url => Uri support
- Uri/Url => Reference support (note: not canonical)
- Adds ability for tests to detect if a Search Param is supported in CapabilityStatement

### Note on `_profile` in R4:
In reference to the [spec](https://www.hl7.org/fhir/search.html#uri) where `_profile` is a canonical converting to a `Uri` search type, in this case we will remove the canonical components (if present) and index the Uri only. When searching with _profile the same behavior applies where the canonical components will not interact with the search. We have raised ticket [FHIR-30514](https://jira.hl7.org/browse/FHIR-30514) to clarify this.

Reasoning for ignoring the canonical components when converting from Canonical => Uri revolves around the terminology:

>  SHOULD support automatically detecting a |[version] portion as part of the search parameter, and interpreting that portion as a search on the version

In future work (#1566), using a canonical with a `reference` search type, where the [target entity](https://www.hl7.org/fhir/references.html#canonical) is one that exists as part of the Definition pattern, the `|4` component will automatically be converted to the search term `&version=4` [as designed](https://build.fhir.org/resource.html#versions).

## Related issues
Refs #882

## Testing
Updates unit and integration tests

Adds the following searches (R4):
- /Observation?_profile=https://uri
- /ConceptMap?source=ValueSet/1 (note: Canonical => Reference is not implemented yet)
- /ConceptMap?target=ValueSet/1 (note: Canonical => Reference is not implemented yet)
- /DocumentReference?location=http://attatchment-uri
- /MessageHeader?destination=http://uri
- /MessageHeader?source=http://uri
- /PlanDefinition?definition=PlanDefinition/1 (note: Canonical => Reference is not implemented yet)
- /Subscription?url=http://channel-uri

## FHIR Team Checklist

Patch


brendankowitz:
@feordin, @johnstairs thanks for reviewing this work previously. I've updated the approach in the PR description, let me know if this makes sense or you have questions?

//cc @CaitlinV39

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/UriSearchValue.cs
```diff
@@ -38,43 +88,58 @@ public UriSearchValue(string uri)
...
-
             var uriSearchValueOther = other as UriSearchValue;
 
             if (uriSearchValueOther == null)
             {
                 return false;
             }
 
-            return Uri == uriSearchValueOther.Uri;
+            return ToString() == uriSearchValueOther.ToString();
```
feordin, line 124: Are the canonical components case sensitive for the comparison?  Should this be a case insensitive comparison?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/CanonicalSearchTests.cs
```diff
@@ -0,0 +1,115 @@
...
+using Microsoft.Health.Fhir.Core.Models;
+using Microsoft.Health.Fhir.Tests.Common.Extensions;
+using Microsoft.Health.Fhir.Tests.Common.FixtureParameters;
+using Xunit;
+using Task = System.Threading.Tasks.Task;
+
+namespace Microsoft.Health.Fhir.Tests.E2E.Rest.Search
+{
+    [HttpIntegrationFixtureArgumentSets(DataStore.All, Format.Json)]
+    public class CanonicalSearchTests : SearchTestsBase<CanonicalSearchTestFixture>
```
feordin, line 19: I was looking for a test case with a canonical style URL that was being treated as a reference and then used in a search.  Perhaps we have it, and I just can't find it?

brendankowitz: This PR doesn't cover Canonical => Reference, I plan to address this separately in #1566.

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/UriSearchValue.cs
brendankowitz: It does say: `URLs are always considered to be case-sensitive` (https://www.hl7.org/fhir/references.html#literal)
However, it probably helps to be explicit about this.
```suggestion
            return string.Equals(ToString(), uriSearchValueOther.ToString(), StringComparison.Ordinal);
```

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/CanonicalSearchTests.cs
feordin: Ah yes! sound good.

feordin (approved):

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/UriSearchValue.cs
brendankowitz: Updated

johnstairs (approved):

Issue #1594, opened 2021-01-14
Backwards compat broken with earlier schema versions.


johnstairs:
**Describe the bug**

We've added new versions of the `UpsertResource` stored procedure (_2, more in open PRs), but we are always invoking VLatest, which would be _2 in your PR. And that procedure won't be there if we are running schema version, say, 6.

Regression added my me in https://github.com/microsoft/fhir-server/pull/1497

**FHIR Version?**
All

**Data provider?**
SQL Server

**To Reproduce**
1. Create a database that is running V6 of the schema
2. Run the FHIR server with the latest from main, but do not upgrade the schema
3. POST a new resource
4. Notice the failure to resolve `UpsertResource_2`

CaitlinV39:
@johnstairs  - Can I close this issue?

johnstairs:
@CaitlinV39 yes. Closing.

Issue #1538, opened 2020-12-22
Issue with custom IPipelineBehavior


giacomodeliberali:
**Describe the bug**

We added a custom `IPipelineBehavior` that appends the _tenantId_ in the `meta.security` property.

```csharp
public class AddTenantIdBehavior : IPipelineBehavior<CreateResourceRequest, UpsertResourceResponse>
    {
        private readonly IMediator _mediator;
        private readonly ResourceDeserializer _deserializer;

        public AddTenantIdBehavior(
            IMediator mediator,
            ResourceDeserializer deserializer)
        {
            _mediator = mediator;
            _deserializer = deserializer;
        }

        public async Task<UpsertResourceResponse> Handle(CreateResourceRequest request, CancellationToken cancellationToken, RequestHandlerDelegate<UpsertResourceResponse> next)
        {
            // just a simplification of what we are doing
            var deviceId = observation.Device.Reference.Split("/")[1];
            var resourceElement = await _mediator.GetResourceAsync(new ResourceKey(ModelInfo.GetFhirTypeNameForType(typeof(Device)), deviceId), cancellationToken);
            var device = resourceElement.ToPoco<Device>(_deserializer);

            var tenant = device.Meta.Security.Find(c => c.System == "our-tenant-system");
            observation.Meta.Security.Add(tenant);

            return await next();
        }
    }
```

The creation of the resource randomly fails throwing a null pointer exception at line 53 of `BaseResourceHandler` when trying to generate a new GUID:

https://github.com/microsoft/fhir-server/blob/77dd1cc76ca7e3fea0756631589832b7e5af1aae/src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/BaseResourceHandler.cs#L49-L54

The `Create` property that should return the lambda that generates the GUID does seem to return null:

![create-null](https://user-images.githubusercontent.com/4717227/102875361-fe026700-4443-11eb-83aa-203b1774864a.png)

If we replace the DI registration type of the `ResourceIdProvider` 

https://github.com/microsoft/fhir-server/blob/77dd1cc76ca7e3fea0756631589832b7e5af1aae/src/Microsoft.Health.Fhir.Shared.Api/Modules/PersistenceModule.cs#L28

to `AddTransient` the issue never happend again. 

**FHIR Version?**
R4

**Data provider?**
SQL Server

**To Reproduce**
It randomly happens after we added the new pipeline middleware 😕 

**Expected behavior**
The resource should awlays be created.

[...]

PR #1606, opened 2021-01-20
Fix UpsertResource backwards compatibility

johnstairs:
## Description
Fixes an earlier compatibility break where the code was only able to insert into a database with the latest schema version.

Added an integration smoke test for this scenario. The test refactoring to accomplish this is a bit messy.

## Related issues
Addresses #1594.

## Testing
Introduced a new integration test.

Patch

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Conformance/ConformanceProviderBase.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Conformance/ConformanceProviderBase.cs
@@ -21,6 +21,11 @@ namespace Microsoft.Health.Fhir.Core.Features.Conformance
 
         public abstract Task<ResourceElement> GetCapabilityStatementAsync(CancellationToken cancellationToken = default(CancellationToken));
 
+        internal void ClearCache()
+        {
+            _evaluatedQueries.Clear();
+        }
+
         public async Task<bool> SatisfiesAsync(IEnumerable<CapabilityQuery> queries, CancellationToken cancellationToken = default(CancellationToken))
         {
             EnsureArg.IsNotNull(queries, nameof(queries));
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
@@ -39,6 +39,7 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Storage
         private readonly SqlServerDataStoreConfiguration _configuration;
         private readonly SqlServerFhirModel _model;
         private readonly SearchParameterToSearchValueTypeMap _searchParameterTypeMap;
+        private readonly V6.UpsertResourceTvpGenerator<ResourceMetadata> _upsertResourceTvpGeneratorV6;
         private readonly VLatest.UpsertResourceTvpGenerator<ResourceMetadata> _upsertResourceTvpGeneratorVLatest;
         private readonly RecyclableMemoryStreamManager _memoryStreamManager;
         private readonly CoreFeatureConfiguration _coreFeatures;
@@ -50,6 +51,7 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Storage
             SqlServerDataStoreConfiguration configuration,
             SqlServerFhirModel model,
             SearchParameterToSearchValueTypeMap searchParameterTypeMap,
+            V6.UpsertResourceTvpGenerator<ResourceMetadata> upsertResourceTvpGeneratorV6,
             VLatest.UpsertResourceTvpGenerator<ResourceMetadata> upsertResourceTvpGeneratorVLatest,
             IOptions<CoreFeatureConfiguration> coreFeatures,
             SqlConnectionWrapperFactory sqlConnectionWrapperFactory,
@@ -59,6 +61,7 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Storage
             EnsureArg.IsNotNull(configuration, nameof(configuration));
             EnsureArg.IsNotNull(model, nameof(model));
             EnsureArg.IsNotNull(searchParameterTypeMap, nameof(searchParameterTypeMap));
+            EnsureArg.IsNotNull(upsertResourceTvpGeneratorV6, nameof(upsertResourceTvpGeneratorV6));
             EnsureArg.IsNotNull(upsertResourceTvpGeneratorVLatest, nameof(upsertResourceTvpGeneratorVLatest));
             EnsureArg.IsNotNull(coreFeatures, nameof(coreFeatures));
             EnsureArg.IsNotNull(sqlConnectionWrapperFactory, nameof(sqlConnectionWrapperFactory));
@@ -68,6 +71,7 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Storage
             _configuration = configuration;
             _model = model;
             _searchParameterTypeMap = searchParameterTypeMap;
+            _upsertResourceTvpGeneratorV6 = upsertResourceTvpGeneratorV6;
             _upsertResourceTvpGeneratorVLatest = upsertResourceTvpGeneratorVLatest;
             _coreFeatures = coreFeatures.Value;
             _sqlConnectionWrapperFactory = sqlConnectionWrapperFactory;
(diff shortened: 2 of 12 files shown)


src/Microsoft.Health.Fhir.SqlServer/Microsoft.Health.Fhir.SqlServer.csproj
```diff
@@ -12,11 +12,14 @@
     <EmbeddedResource Include="Features\Schema\Migrations\6.diff.sql" />
     <EmbeddedResource Include="Features\Schema\Migrations\6.sql">
       <InputToImmutableSqlGenerator>true</InputToImmutableSqlGenerator>
+      <InputToMutableSqlGenerator>true</InputToMutableSqlGenerator>
+      <MutableClassVersion>6</MutableClassVersion>
```
brendankowitz, line 16: In future, should we do this in versions where we add immutable types that feed into the Upsert Proc?

johnstairs: Yes, I think this would become the pattern

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlServerSqlCompatibilityTests.cs
```diff
@@ -0,0 +1,73 @@
...
+using Xunit;
+
+namespace Microsoft.Health.Fhir.Tests.Integration.Persistence
+{
+    public class SqlServerSqlCompatibilityTests
+    {
+        /// <summary>
+        /// A basic smoke test verifying that the code is compatible with schema versions
+        /// all the way back to <see cref="SchemaVersionConstants.Min"/>. Ensures that the code can
+        /// insert insert a resource using every schema version, and that we can read resources
```
rotodd, line 23: Nit: there is one extra "insert" here

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlServerSqlCompatibilityTests.cs
```diff
@@ -0,0 +1,73 @@
...
+        /// that were inserted into with earlier schemas.
+        /// </summary>
+        [Fact]
+        public async Task GivenADatabaseWithAnEarlierSupportedSchema_WhenUpserting_OperationSucceeds()
```
rotodd, line 27: +1 this test is awesome, thanks for adding it.

My understanding is, if the current schema version is `x` and version `x+1` is added, this would test that `x+1.diff.sql` can be applied on top of schema version `x`. 

Would it also be possible to test the scenario when the snapshot file, `x+1.sql`, is run? I believe this has a different initialization path in `SqlServerFhirModel.cs` that would be good to verify (np if we want to make a work item to add this in later).

rotodd (approved):

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlServerSqlCompatibilityTests.cs
rotodd: Dug up [AB#78858](https://microsofthealth.visualstudio.com/Health/_workitems/edit/78858/) which we could use to track this?

johnstairs: Ok, yes, I see. We can expand the test to also ensure we create the database from each snapshot version, and then upgrade from there. Is that what you had in mind?

src/Microsoft.Health.Fhir.Core/Features/Conformance/ConformanceProviderBase.cs
```diff
@@ -21,6 +21,11 @@ public abstract class ConformanceProviderBase : IConformanceProvider
 
         public abstract Task<ResourceElement> GetCapabilityStatementAsync(CancellationToken cancellationToken = default(CancellationToken));
 
+        internal void ClearCache()
+        {
+            _evaluatedQueries.Clear();
```
brendankowitz, line 26: @feordin I forgot about this, we should also clear this when we reflect changes in the capability statement to do with search params

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -148,6 +138,43 @@ public async Task<UpsertOutcome> UpsertAsync(ResourceWrapper resource, WeakETag
             }
         }
 
+        private void PopulateUpsertResourceCommand(SqlCommandWrapper sqlCommandWrapper, ResourceWrapper resource, ResourceMetadata resourceMetadata, bool allowCreate, bool keepHistory, int? eTag, RecyclableMemoryStream stream)
+        {
+            long baseResourceSurrogateId = ResourceSurrogateIdHelper.LastUpdatedToResourceSurrogateId(resource.LastModified.UtcDateTime);
+            short resourceTypeId = _model.GetResourceTypeId(resource.ResourceTypeName);
+
+            if (_schemaInformation.Current >= 7)
```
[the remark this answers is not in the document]
brendankowitz: I was trying to think of what this might be called, VersionWithLatestUpsertResourceStoredProc :P

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/FhirStorageTestsFixture.cs
```diff
@@ -4,36 +4,74 @@
...
+        {
+        }
+
+        internal FhirStorageTestsFixture(IServiceProvider fixture)
+        {
+            _fixture = fixture;
+
+            // This step has to be done in the constructor because it uses an AsyncLocal and the tests run with the same
+            // execution context as the fixture constructor, but not the same as InitializeAsync().
+            _resourceIdProvider = new ResourceIdProvider();
```
brendankowitz, line 62: Adding a ref to this bug #1538

brendankowitz (approved):

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlServerSqlCompatibilityTests.cs
rotodd: Yep, I think even simply creating the database from each snapshot version and posting a resource would be good, since we are already testing the upgrade paths in this test.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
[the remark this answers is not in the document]
rotodd: Haha yeah it is not an easy one to name. `7.sql` was added in #1497 in order to add support for references with a missing type. Maybe we could incorporate that in the naming and have something like `SupportForReferencesWithMissingTypeVersion`? (just an idea)

We will soon be adding another schema version which modifies the upsert proc to support SQL reindexing, which will redefine what the "latest" will be.

src/Microsoft.Health.Fhir.Core/Features/Conformance/ConformanceProviderBase.cs
johnstairs: Separate bug?

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlServerSqlCompatibilityTests.cs
johnstairs: Added in https://github.com/microsoft/fhir-server/pull/1606/commits/c09b5d12d283143adcb7981f5f7797479fa7f72e

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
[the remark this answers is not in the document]
johnstairs: Added in https://github.com/microsoft/fhir-server/pull/1606/commits/c09b5d12d283143adcb7981f5f7797479fa7f72e

PR #1579, opened 2021-01-11
Update string search table schema

namadabu:
## Description
Adding two columns - IsMin and IsMax - to the string search parameter table. This will help us keep track of "min" and "max" values for parameters that can have multiple values. Part of the work required to support _sort. 

The two columns aren't being used for now (I have added a "dummy" Row Generator as a placeholder). As part of the next PR I will add code that pipes the required values and sets the columns to the correct values.

This change is being merged into a separate feature branch.

## Related issues
Addresses [AB#78588](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/78588)

## Testing
Manually tested to confirm new columns are present in the database and upserting a resource works.

## FHIR Team Checklist
- [x] **Update the title** of the PR to be succinct and less than 50 characters
- [x] **Add a milestone** to the PR for the sprint that it is merged (i.e. add S47)
- [x] Tag the PR with the type of update: **Bug**, **Dependencies**, **Enhancement**, or **New-Feature**
- [x] Tag the PR with **Azure API for FHIR** if this will release to the managed service
- Review [squash-merge requirements](https://github.com/microsoft/fhir-server/blob/master/SquashMergeRequirements.md)

### Semver Change ([docs](https://github.com/microsoft/fhir-server/blob/master/docs/Versioning.md))
Patch|Skip|Feature|Breaking (reason)

PR #1607, opened 2021-01-20
Extract min/max values for string search param

namadabu:
## Description
Some search parameters can have multiple values. We want to keep track of which amongst these values will be first and last when sorting them. The plan is to store them in the corresponding search parameter table so that we can retrieve them when required.

This PR adds code to calculate and keep track of which values for a given search parameter are "min" and "max". I have added fields to the ISearchValue objects to keep track of this information. The TVP generator then takes care of adding this to the search parameter table. Currently the code only handles this for String Search Parameter types.

This PR builds on top of the schema changes introduced as part of PR #1579 . Currently I have this merging against the branch in PR #1579 . Will rebase and merge this to feature/sort-sql once the other PR gets checked in.

## Related issues
Addresses [issue #78593].

## Testing
Added UTs. Did E2E testing on local deployment.

## FHIR Team Checklist


src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/StringSearchValue.cs
```diff
@@ -53,6 +60,23 @@ public void AcceptVisitor(ISearchValueVisitor visitor)
...
+                throw new ArgumentException("Value to be compared to cannot be null");
+            }
+
+            var otherValue = searchValue as StringSearchValue;
+            if (otherValue == null)
+            {
+                throw new ArgumentException($"Value to be compared should be of type {typeof(StringSearchValue)}");
+            }
+
+            return string.Compare(ToString(), otherValue.ToString(), StringComparison.CurrentCultureIgnoreCase);
```
namadabu, line 77: A little unsure about whether `StringComparison.CurrentCultureIgnoreCase` is the current comparison mode here.

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/ISearchValue.cs
```diff
@@ -17,6 +17,30 @@ public interface ISearchValue : IEquatable<ISearchValue>
...
+        /// Compares two ISearchValue objects of the same type and returns an integer that
+        /// indicates their relative position in the sort order.
+        /// </summary>
+        /// <param name="otherValue">The value to compare against.</param>
+        /// <returns>
+        /// -1 if the current value comes before the given search value.
+        /// 0 if both values will occur at the same position.
+        /// 1 if the current value comes after the given search value.
+        /// </returns>
+        int Compare(ISearchValue otherValue);
```
feordin, line 42: Should we update this to IComparable interface? and rename to CompareTo?

Ivanidzo4ka:
>     public interface ISearchValue : IEquatable<ISearchValue>

I would create ISupportMinMaxSearchValue interface instead of extending this one.
At least for how things look right now.
I'm not sure it's worth to plumb code to 8 classes if only one class support it.
Can be wrong since I don't know what plans for future are. Are we plan to support min max in all ISearchValues?

---
Refers to: src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/ISearchValue.cs:13 in 014e34e. [](commit_id = 014e34eccfcfb9113f2da0f39a70bc007b117f24, deletion_comment = False)

Ivanidzo4ka:
>     public interface ISearchValue : IEquatable<ISearchValue>

From code design perspective I think it's better to have separate interface mostly because of this line in ResourceWrapperFactory:

` if (currentEntry.SearchParameter.Type != ValueSets.SearchParamType.String)`

It basically contains knowledge about which ISearchValue implemented min/max support and which not, and it can be hard to support in future because of disconnect of classes and code.

(Another option is to introduce flag here : bool SupportMinMax, but it's feels a bit too much code)

---
In reply to: [765616568](https://github.com/microsoft/fhir-server/pull/1607#issuecomment-765616568) [](ancestors = 765616568)

---
Refers to: src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/ISearchValue.cs:13 in 014e34e. [](commit_id = 014e34eccfcfb9113f2da0f39a70bc007b117f24, deletion_comment = False)

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/ISearchValue.cs
namadabu: @feordin  I tried this out after you mentioned it. One issue is that we not only have to implement `CompareTo`, but also override all equality operations and `GetHashCode()`. We won't be using some of that code (at least not right now). So I am thinking it might be okay to have `Compare` as a standalone method. And if things change in the future we can change the code to implement `IComparable`. Thoughts?

namadabu:
> > ```
> > public interface ISearchValue : IEquatable<ISearchValue>
> > ```
> 
> From code design perspective I think it's better to have separate interface mostly because of this line in ResourceWrapperFactory:
> 
> ` if (currentEntry.SearchParameter.Type != ValueSets.SearchParamType.String)`
> 
> It basically contains knowledge about which ISearchValue implemented min/max support and which not, and it can be hard to support in future because of disconnect of classes and code.
> 
> (Another option is to introduce flag here : bool SupportMinMax, but it's feels a bit too much code)
> 
> In reply to: [765616568](https://github.com/microsoft/fhir-server/pull/1607#issuecomment-765616568) [](ancestors = 765616568)
> 
> Refers to: src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/ISearchValue.cs:13 in [014e34e](https://github.com/microsoft/fhir-server/commit/014e34eccfcfb9113f2da0f39a70bc007b117f24). [](commit_id = [014e34e](https://github.com/microsoft/fhir-server/commit/014e34eccfcfb9113f2da0f39a70bc007b117f24), deletion_comment = False)

Have added a new interface and moved `IsMin`, `IsMax` and `Compare()` to the new interface. Only StringSearchValue implements this new interface for now.

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/ISearchMinMaxValue.cs
```diff
@@ -0,0 +1,38 @@
...
+        bool IsMin { get; set; }
+
+        /// <summary>
+        /// Determines whether this current value is the maximum when compared to
+        /// a collection of search values for the same paramter.
+        /// </summary>
+        bool IsMax { get; set; }
+
+        /// <summary>
+        /// Compares two ISearchValue objects of the same type and returns an integer that
```
Ivanidzo4ka, line 27: >ISearchValue [](start = 25, length = 12)

super nit: <see cref="ISearchMinMaxValue">

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/ISearchMinMaxValue.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// indicates their relative position in the sort order.
+        /// </summary>
+        /// <param name="otherValue">The value to compare against.</param>
+        /// <returns>
+        /// -1 if the current value comes before the given search value.
+        /// 0 if both values will occur at the same position.
+        /// 1 if the current value comes after the given search value.
+        /// </returns>
+        int Compare(ISearchMinMaxValue otherValue);
```
Ivanidzo4ka, line 36: >Compare [](start = 12, length = 7)

[https://docs.microsoft.com/en-us/dotnet/api/system.icomparable-1.compareto?view=net-5.0](https://docs.microsoft.com/en-us/dotnet/api/system.icomparable-1.compareto?view=net-5.0)
I think you can just do

public interface ISearchMinMaxValue: IComparable<ISearchMinMaxValue>

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/ISearchValue.cs
Ivanidzo4ka: >but also override all equality operations and GetHashCode()
Why tho? IComparable<T> require you to only provide CompareTo method.

namadabu: Yes, the interface only asks for `CompareTo`. But I ran into the following warning when doing so - [CA1036](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1036?view=vs-2019).

Ivanidzo4ka: Well, no one can stop you from disabling that warning, as long as you know what you doing...

namadabu: I usually prefer to avoid suppressing warnings. I have updated the class to implement `IComparable` and suppressed the warning.

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/ISearchMinMaxValue.cs
namadabu: Updated to implement that `IComparable` interface

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/StringSearchValue.cs
brendankowitz: Perhaps Invariant would be better. Though how does `Muller` compare to `Müller`? In FHIR they should be equal

namadabu: Good point. Updated the code to use `CompareOptions.IgnoreNonSpace` to ignore accents. Added a test case for the above strings as well.

brendankowitz (approved):

PR #1610, opened 2021-01-22
Add x-ms-retry-after-ms header when throttling requests

johnstairs:
## Description

When we throttle requests based on the maximum number of concurrent requests, we should return a response header indicating how long the client should wait before retrying.

Note that we are not using the standard `Retry-After` header, but instead `x-ms-retry-after-ms`, which we already return when Cosmos DB throttles.

The approach is very simple. It is based on observing the success rate of requests over time and adjusting it up or down to try to keep the success rate at 99% without returning an unnecessarily large value. Growth and decay are both exponential at different rates, with the growth rate being faster.

In a future, change ([AB#79045](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/79045)), we'll add queuing before rejecting requests to increase throughput.

## Related issues
Addresses [AB#78885](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/78885) .

## Testing
Testing was based on a simulator where I adjusted various constants.

Patch

### DIFF
--- a/src/Microsoft.Health.Fhir.Api/Features/Throttling/ThrottlingMiddleware.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/Throttling/ThrottlingMiddleware.cs
@@ -26,6 +26,7 @@ namespace Microsoft.Health.Fhir.Api.Features.Throttling
     /// </summary>
     public sealed class ThrottlingMiddleware : IDisposable
     {
+        private const double TargetSuccessPercentage = 99;
         private const int MinRetryAfterMilliseconds = 20;
         private const int MaxRetryAfterMilliseconds = 60000;
         private const double RetryAfterGrowthRate = 1.2;
@@ -78,8 +79,8 @@ namespace Microsoft.Health.Fhir.Api.Features.Throttling
         /// <summary>
         /// Samples the success rate (i.e. requests not throttled) over a period, and adjusts the value we return as the retry after header.
         /// This is an extremely simple approach that exponentially grows or decays the value depending on whether the success rate is
-        /// less than or greater than 99%, respectively.
-        /// The value as high as necessary to keep the success rate over 99%, but we want it to decrease when the request rate backs off,
+        /// less than or greater than TargetSuccessPercentage, respectively.
+        /// The value grows to be as high as necessary to keep the success rate over 99%, but we want it to decrease when the request rate backs off,
         /// so overall latency for clients is not unnecessarily high.
         /// </summary>
         public async Task SamplingLoop()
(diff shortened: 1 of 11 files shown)


src/Microsoft.Health.Fhir.Api/Features/Throttling/ThrottlingMiddleware.cs
```diff
@@ -55,6 +67,50 @@ public class ThrottlingMiddleware
...
+
+                var totalCount = successCount + failureCount;
+                double successRate = totalCount == 0 ? 100.0 : successCount * 100.0 / totalCount;
+
+                if (successRate >= 99)
+                {
+                    // see if we should lower the value
+                    if (_currentRetryAfterMilliseconds > MinRetryAfterMilliseconds)
+                    {
+                        _currentRetryAfterMilliseconds = (int)(_currentRetryAfterMilliseconds / RetryAfterDecayRate);
```
LTA-Thinking, line 102: Nit: This could end up with a value below the MinRetryAfterMilliseconds. With the current values it couldn't go below 19ms though.

src/Microsoft.Health.Fhir.Api/Features/Throttling/ThrottlingMiddleware.cs
```diff
@@ -55,6 +67,50 @@ public class ThrottlingMiddleware
...
+                    }
+                }
+                else
+                {
+                    // see if we should raise the value.
+                    if (_currentRetryAfterMilliseconds < MaxRetryAfterMilliseconds)
+                    {
+                        _currentRetryAfterMilliseconds = (int)(_currentRetryAfterMilliseconds * RetryAfterGrowthRate);
```
LTA-Thinking, line 110: Nit: Similarly this could be above MaxRetryAfterMilliseconds. The max value with current settings is 71998ms.

src/Microsoft.Health.Fhir.Api/Features/Throttling/ThrottlingMiddleware.cs
```diff
@@ -78,14 +134,19 @@ public async Task Invoke(HttpContext context)
                 if (Interlocked.Increment(ref _requestsInFlight) <= _configuration.ConcurrentRequestLimit)
                 {
                     // Still within the concurrent request limit, let the request through.
+                    Interlocked.Increment(ref _currentPeriodSuccessCount);
```
LTA-Thinking, line 137: When do these values reset? It looks like they would be remembered as long as the server is running, so this count would be less useful the longer the server is up.

brendankowitz: I think with the above it resets to 0 when the SamplingLoop runs

johnstairs: Yes, these are reset with every sample loop iteration
```csharp
var successCount = Interlocked.Exchange(ref _currentPeriodSuccessCount, 0);
var failureCount = Interlocked.Exchange(ref _currentPeriodRejectedCount, 0);
```

src/Microsoft.Health.Fhir.Api/Features/Throttling/ThrottlingMiddleware.cs
johnstairs: Yeah, it's not super important that we stay exactly within these limits, more that the header doesn't reach some outrageously large number or less than 3, where because of truncation it will never grow.
But I can cap the results so they never surpass the min/max

johnstairs: Capped in https://github.com/microsoft/fhir-server/pull/1610/commits/30820bc71b481457f431941e342ad86aa61634b4

LTA-Thinking (approved):

src/Microsoft.Health.Fhir.Api/Features/Throttling/ThrottlingMiddleware.cs
```diff
@@ -55,6 +68,39 @@ public class ThrottlingMiddleware
                     _excludedEndpoints.Add((excludedEndpoint.Method, excludedEndpoint.Path));
                 }
             }
+
+#pragma warning disable 4014 // Intentional fire and forget behavior
```
brendankowitz, line 72: Would the way around this be to assign the task as a class variable and cleanup in dispose?

johnstairs: Yes. I was reluctant to do because of the noise it introduces, but I've added it in https://github.com/microsoft/fhir-server/pull/1610/commits/9966d1236676c9584b92daf0251a80542b1861e9.

Issue #1474, opened 2020-11-19
When using $export?_format=foo an unformatted error message is returned


brendankowitz:
**Describe the bug**
When using $export?_format=foo an unformatted error message is returned.

**To Reproduce**
1. Use a _format option that is not valid, e.g. `$export?_format=foo`

**Expected behavior**
OperationOutcome with error message of missing format

**Actual behavior**
The OperationOutcome still has text placeholder
```
  "resourceType": "OperationOutcome",
  "id": "4248cdfb-e657-4b26-affa-67aff4a01b22",
  "issue": [
    {
      "severity": "error",
      "code": "invalid",
      "diagnostics": "The format '{0}' could not be found."
    }
  ]
}
```

Related to #1422

CaitlinV39:
AB#77963

CaitlinV39:
AB#77963

PR #1611, opened 2021-01-22
Fix exception messages in export create request.

LTA-Thinking:
## Description
Fix bad exception messages from export create request.

## Related issues
Addresses #1474

## Testing
Manual testing

## FHIR Team Checklist

Patch


namadabu:
Can we add a UT to validate this change?

namadabu (approved):

LTA-Thinking:
> Can we add a UT to validate this change?

Both of the changes have UTs for them, but they didn't catch the issue because we don't check the exact exception text in our UTs and the status codes were correct.

namadabu (approved):

PR #1616, opened 2021-01-28
Implements initial SQL reindexing prototype

rotodd:
## Description
This PR implements an initial prototype of SQL reindexing. More, specifically, it:
- Adds the `SearchParamHash` column to the `dbo.Resource` table in order to keep track of each resource's latest indexed search parameters 
    - This value is initially `NULL`
    - It will be populated for a given resource when it is involved in a reindexing operation
- Creates the `UpsertResource_3` procedure which takes in the hash value as input
     - Note: this is subject to change (please see Open TODOs A)
- Modifies the `ReadResource` stored procedure to return the search parameter hash value
- Updates the SQL server search functionality to return the search parameter hash value
    - There is more to be done here (please see Open TODOs B)
- Removes checks for the schema version being greater than or equal to V4, since V4 is now the minimum version 
- Returns the newly added search parameters in the `UpsertSearchParam` procedure to keep a C#-side dictionary up to date

Open TODOs:

A) Currently, we are creating a new version of a resource when its search parameter hash is updated. This should not be the case and will be fixed when the methods to update search indicies are further developed in [AB#79121](https://microsofthealth.visualstudio.com/Health/_workitems/edit/79121) and [AB#73395](https://microsofthealth.visualstudio.com/Health/_workitems/edit/73395).
B) The search that is executed during reindexing operations needs to be modified to return resources independent of if `IsHistory` is `true` or `false`. This will be addressed in [AB#73616](https://microsofthealth.visualstudio.com/Health/_workitems/edit/73616).

## Demo

In this demo, we follow the steps outline [here](https://microsoft.sharepoint.com/teams/msh/_layouts/OneNote.aspx?id=%2Fteams%2Fmsh%2FShared%20Documents%2FNotebooks%2FResolute%20Engineering&wd=target%28Features%2FTesting.one%7CB6E4751D-7CD8-4E7C-9B23-A05EC3773E96%2FCustom%20Search%20Manual%20Test%7C48BFFBB0-7E99-4C22-A179-7EA1C0C573DB%2F%29).

**Part 1**: creating a patient resource with the new search parameter, creating the new search parameter and triggering the reindex operation
reindex_1

**Part 2**: querying with the new search parameter
reindex_2

## Related issues
Addresses [AB#78839](https://microsofthealth.visualstudio.com/Health/_workitems/edit/78839).

## Testing
Expands existing Cosmos reindexing integration tests to cover SQL as well. 

## FHIR Team Checklist


src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -132,17 +133,15 @@ public override object VisitSqlRoot(SqlRootExpression expression, SearchOptions
...
-                {
-                    // IsRawResourceMetaSet column was added in V4
-                    StringBuilder.Append(VLatest.Resource.IsRawResourceMetaSet, resourceTableAlias).Append(", ");
-                }
-                else
+                StringBuilder.Append(VLatest.Resource.IsRawResourceMetaSet, resourceTableAlias).Append(", ");
+
+                if (_schemaInfo.Current >= SchemaVersionConstants.SearchParameterHashSchemaVersion)
                 {
-                    StringBuilder.Append("CAST(0 AS bit) AS IsRawResourceMetaSet, ");
```
Ivanidzo4ka, line 142: >StringBuilder.Append("CAST(0 AS bit) AS IsRawResourceMetaSet, "); [](start = 20, length = 65)

Why we getting rid of this line?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/8.diff.sql
```diff
@@ -1,3 +1,476 @@
+/*************************************************************
+    Resource table
```
Ivanidzo4ka, line 2: >Resource table [](start = 4, length = 14)

shouldn't this be 9th version?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
```diff
@@ -106,6 +106,13 @@ public short GetSearchParamId(Uri searchParamUri)
             return _searchParamUriToId[searchParamUri];
         }
 
+        public void SetSearchParamId(string searchParamUri, short searchParamId)
```
Ivanidzo4ka, line 109: >SetSearchParamId [](start = 20, length = 16)

It's a bit weird to see how these values been set, but no changes related to how they consumed (or maybe I'm missing them). Was consume part already implemented?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
rotodd: Good question, there are two things going on here. The first thing is I'm removing the check for the schema version being greater than or equal to V4, since V4 is now the minimum schema version.

The second thing is I am adding a check to see if the schema version is greater than or equal to V8, since the query will differ depending on if there is a column for the search parameter hash value in `dbo.Resource`.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
rotodd: Yes, you're right - the `UpsertSearchParams` stored procedure was added in #1430. This change would have been better suited to be part of that PR, but the need to sync up the C# dictionary with the SQL table was only brought to light now that we are relying on the dictionary for reindexing.

All this being said, the dictionary could be updated at different point in the creation of the new search parameter resource (right now `SetSearchParamId()` is being called in `SqlServerSearchParameterStatusDataStore`'s `UpsertStatuses()` method). Very open to feedback on this.

Tagging @feordin here in case you have thoughts/ideas.

feordin: I think the method makes sense to SetSearchParamId.  I understand @Ivanidzo4ka's comment on it seeming out of place.  Pieces of functionality are coming together over time here.  I do wonder if we need to check for the Uri already existing in the _searchParamUriToId dictionary?  Will the .Add method fail if the key already exists?  As mentioned this method is called from an UpsertStatus method.  Is it possible that could be called more than once for a given SearchParam?

rotodd: Ah yes, good question. I originally had a `.TryAdd()` here but then made modifications to `UpsertSearchParams` in `8.sql` to only return the parameters that were inserted and not updated. Happy to put the `TryAdd()` back in though if it makes more sense to have this logic here.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -238,7 +256,8 @@ public async Task<ResourceWrapper> GetAsync(ResourceKey key, CancellationToken c
                             isDeleted,
                             searchIndices: null,
                             compartmentIndices: null,
-                            lastModifiedClaims: null)
+                            lastModifiedClaims: null,
+                            searchParamHash)
```
johnstairs, line 260: What is the reason for including this in ResourceWrapper?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
feordin: I see the changes in the stored proc.  Just as a question, will there ever be a case when this mapping needs to be updated?  Could the id in Sql ever change for a search parameter uri?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
rotodd: Ah, good question. From what I understand, it's a logical place to keep it because reindexing operations involve searching for and updating resources, which ingest and output `ResourceWrapper` objects. During a reindexing operation, we need to keep track of the resources that have been indexed as well as their updated hash values, and the resource wrapper is a nice way to bundle things together.

@feordin am I missing anything?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
rotodd: I believe this mapping will need to be updated whenever we allow a user to modify/add/delete a `SearchParameter` resource. In the cases of modifying and deleting a search parameter, the flow will be the same as adding one: the request is made by the user, the SQL stored procedure is called, then the C# dictionary is updated to match the search parameter registry in the database.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
feordin: @rotodd That is correct. It is another key piece of metadata for the resource, and directly related to the SearchIndices collection which are also a property on the ResourceWrapper.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
feordin: Sounds good.  So perhaps you will have a separate method to update the mapping?  Or will this method be used with an AddOrUpdate?

rotodd: Ah, I see what you're saying - this method is called "SetSearchParamId" but it only adds new ones right now. It might be better suited to be named "AddSearchParamIdMapping" for now, then maybe it could be modified and renamed once update is enabled. Is that correct?

feordin: Yes, feel free to change it when you add that functionality in the future..

feordin (approved):

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
rotodd: Sounds good, renamed to "AddSearchParamIdMapping" in 7c5c5c2, and I plan to modify it once we add the update functionality for SQL in a future PR.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -248,7 +252,7 @@ private async Task<SearchResult> SearchImpl(SearchOptions searchOptions, bool hi
 
                             // At this point we are at the last row.
                             // if we have more columns, it means sort expressions were added.
-                            if (reader.FieldCount > 10)
+                            if (reader.FieldCount > _resourceTableColumnCount)
```
Ivanidzo4ka, line 255: >_resourceTableColumnCount [](start = 52, length = 25)

rant: If you will ever have chance to touch sql query generator  I think this should be static property of class. (VLatest.Resource.ColumnCount)
It's too easy to forget to update this field in the future.

(As option you can calculate via reflection how many properties of class derived from `Column` base class, but I think it's too much)

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -284,7 +288,8 @@ private async Task<SearchResult> SearchImpl(SearchOptions searchOptions, bool hi
                                 isDeleted,
                                 null,
                                 null,
-                                null),
+                                null,
+                                searchParameterHash),
```
Ivanidzo4ka, line 292: super nit: it's a bit weird to see hash populated and `SearchIndices` is set to null.
Any chance we can document in future what is `SearchIndices` and `SearchParameterHash` is?

Ivanidzo4ka (approved):
:shipit:

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
rotodd: Good idea. I originally attempted to implement reflection to pull in the number of properties, but, as you said, it was a little over the top. I created the work item [AB#79261](https://microsofthealth.visualstudio.com/Health/_workitems/edit/79261) to track this.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
rotodd: Not a nit! I think this is really important to understand. I am currently updating the reindexing documentation to clarify this, and I'll post a link to it here once I'm done.

rotodd: A PaaS documentation PR is up [here](https://microsofthealth.visualstudio.com/Health/_git/health-paas-docs/pullrequest/14652?path=%2Fspecs%2FFHIR%2FCustomSearch%2FReindexing.md) 😊

PR #1617, opened 2021-01-28
Add queuing to throttling middleware

johnstairs:
## Description
Two changes in this PR: (1) adds optional request queuing to ThrollingMiddleware instead of immediately rejecting requests when the max count concurrent requests has been reached and (2) improves the performance of the 429 response.

The queue has a configurable max size and a configurable max duration that requests are allowed to stay in the queue.

I ran a little simulation to show the potential benefits. Suppose you have configured 5 maximum concurrent requests, a queue size of 100, and a maximum queue duration of 100ms. Also suppose there is a client latency of 6 ms. If clients respect the `x-ms-retry-after-ms` header, and you have 10 clients repeatedly calling the the server with an operation that takes 20ms, you might expect something like:

|            | successful RPS | 429 Rate | AVG latency |
|------------|----------------|----------|-------------|
| no queue   | 105            | 0.78%    | 94ms        |
| with queue | 157        | 0%       | 64ms        |

The latency is here includes all retries before a request succeeds.

Next, this improves the performance of returning a 429 response by bypassing the MVC output formatted and just writing to the response stream directly. (This improves 429 throughput by around 3x)

## Related issues
Addresses [AB#79045](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/79045)

## Testing
Local stress testing.

Patch

### DIFF
--- a/src/Microsoft.Health.Fhir.Api/Configs/ThrottlingConfiguration.cs
+++ b/src/Microsoft.Health.Fhir.Api/Configs/ThrottlingConfiguration.cs
@@ -8,16 +8,38 @@ using Microsoft.Health.Fhir.Api.Features.Throttling;
 
 namespace Microsoft.Health.Fhir.Api.Configs
 {
+    /// <summary>
+    /// Configuration values for optional request throttling.
+    /// </summary>
     public class ThrottlingConfiguration
     {
+        /// <summary>
+        /// Whether to enable request throttling.
+        /// </summary>
         public bool Enabled { get; set; }
 
+        /// <summary>
+        /// The maximum number of requests that are allowed to execute concurrently.
+        /// If queueing is disabled, the server responds with error code 429.
+        /// If queueing is enabled, the request is queued for up to be executed once there is capacity.
+        /// </summary>
         public int ConcurrentRequestLimit { get; set; }
 
+        /// <summary>
+        /// Endpoints that are excluded from throttling
+        /// </summary>
         public HashSet<ExcludedEndpoint> ExcludedEndpoints { get; } = new HashSet<ExcludedEndpoint>();
 
+        /// <summary>
+        /// The maximum number of requests that can be queued up at a time.
+        /// If 0, queueing is disabled.
+        /// </summary>
         public int MaxQueueSize { get; set; }
 
+        /// <summary>
+        /// The total number of milliseconds a request can sit in the queue. If this time
+        /// elapses before the request is picked up, the server responds with a 429.
+        /// </summary>
         public int MaxMillisecondsInQueue { get; set; }
     }
 }
--- a/src/Microsoft.Health.Fhir.Api/Configs/ThrottlingConfiguration.cs
+++ b/src/Microsoft.Health.Fhir.Api/Configs/ThrottlingConfiguration.cs
@@ -15,5 +15,9 @@ namespace Microsoft.Health.Fhir.Api.Configs
         public int ConcurrentRequestLimit { get; set; }
 
         public HashSet<ExcludedEndpoint> ExcludedEndpoints { get; } = new HashSet<ExcludedEndpoint>();
+
+        public int MaxQueueSize { get; set; }
+
+        public int MaxMillisecondsInQueue { get; set; }
     }
 }
--- a/src/Microsoft.Health.Fhir.Api/Features/Throttling/ThrottlingMiddleware.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/Throttling/ThrottlingMiddleware.cs
@@ -5,17 +5,14 @@
 
 using System;
 using System.Collections.Generic;
-using System.Net;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using EnsureThat;
 using Microsoft.AspNetCore.Http;
-using Microsoft.AspNetCore.Mvc;
-using Microsoft.AspNetCore.Routing;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Microsoft.Health.Fhir.Api.Configs;
-using Microsoft.Health.Fhir.Api.Features.ActionResults;
 using Microsoft.Health.Fhir.Api.Features.Headers;
 using Microsoft.Health.Fhir.Core.Configs;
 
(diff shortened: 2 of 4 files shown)


src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Throttling/ThrottlingMiddlewareTests.cs
```diff
@@ -109,60 +153,66 @@ public async Task GivenRequestToExcludedEndpoint_WhenAlreadyThrottled_Succeeds()
             Assert.Contains(mapping, x => x.httpContext.Response.StatusCode == 429);
         }
 
-        private List<(Task task, HttpContext httpContext)> SetupPreexistingRequests(int numberOfConcurrentRequests, string path = "")
+        [Fact]
+        public async Task GivenARequestWhenMaxRequestsAlreadyInFlightAndQueueingEnabled_WhenExistingRequestCompletes_TheQueuedRequestCompletes()
```
LTA-Thinking, line 157: Could you add a test for if the max queue size is exceeded?

src/Microsoft.Health.Fhir.Api/Features/Throttling/ThrottlingMiddleware.cs
```diff
@@ -33,19 +32,25 @@ public sealed class ThrottlingMiddleware : IAsyncDisposable, IDisposable
         private const double RetryAfterDecayRate = 1.1;
         private const int SamplePeriodMilliseconds = 500;
 
+        // hard-coding these to minimize resource consumption when throttling
+        private const string ThrottledContentType = "application/json; charset=utf-8";
+        private static readonly ReadOnlyMemory<byte> _throttledBody = Encoding.UTF8.GetBytes($@"{{""severity"":""Error"",""code"":""Throttled"",""diagnostics"":""{Resources.TooManyConcurrentRequests}"",""location"":null}}").AsMemory();
```
brendankowitz, line 37: nit: perhaps remove `""location"":null` since it is null, and also deprecated in R4

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Throttling/ThrottlingMiddlewareTests.cs
johnstairs: The other tests hit that code path because the queue size is 0 😄 . I can add a test.

src/Microsoft.Health.Fhir.Api/Configs/ThrottlingConfiguration.cs
```diff
@@ -15,5 +15,9 @@ public class ThrottlingConfiguration
         public int ConcurrentRequestLimit { get; set; }
 
         public HashSet<ExcludedEndpoint> ExcludedEndpoints { get; } = new HashSet<ExcludedEndpoint>();
+
+        public int MaxQueueSize { get; set; }
+
+        public int MaxMillisecondsInQueue { get; set; }
```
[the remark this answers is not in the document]
johnstairs: That's a good point. When we first brought this in, we decided not to include it in the appsettings because we didn't want to make recommendations on what the value should be. But I think we should document that this functionality exists. @brendankowitz what do you think?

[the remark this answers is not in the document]
brendankowitz: At first the numbers were fairly arbitrary and I think specific to some of the setups we had. I think I'd be ok with introducing a QoS guide. @CaitlinV39 thoughts?

src/Microsoft.Health.Fhir.Api/Features/Throttling/ThrottlingMiddleware.cs
johnstairs: Done

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Throttling/ThrottlingMiddlewareTests.cs
johnstairs: Added

src/Microsoft.Health.Fhir.Api/Configs/ThrottlingConfiguration.cs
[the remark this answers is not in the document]
johnstairs: I've added comments on the configuration properties and included throttling in appsettings.json, with it disabled.

LTA-Thinking (approved):

src/Microsoft.Health.Fhir.Api/Features/Throttling/ThrottlingMiddleware.cs
```diff
@@ -127,35 +137,137 @@ public async Task Invoke(HttpContext context)
...
+                // we do two loop iterations only when we need to queue up the request.
+                lock (_queue)
                 {
-                    // Still within the concurrent request limit, let the request through.
-                    Interlocked.Increment(ref _currentPeriodSuccessCount);
-                    await _next(context);
+                    if (_requestsInFlight < _concurrentRequestLimit)
+                    {
+                        canRun = true;
+                        _requestsInFlight++;
```
brendankowitz, line 152: Does this variable also need to be changed with Interlocked or volatile?

brendankowitz (approved):

src/Microsoft.Health.Fhir.Api/Features/Throttling/ThrottlingMiddleware.cs
```diff
@@ -55,21 +60,26 @@ public sealed class ThrottlingMiddleware : IAsyncDisposable, IDisposable
...
             {
-                foreach (var excludedEndpoint in _configuration.ExcludedEndpoints)
+                foreach (var excludedEndpoint in configuration.ExcludedEndpoints)
                 {
                     _excludedEndpoints.Add((excludedEndpoint.Method, excludedEndpoint.Path));
                 }
             }
 
+            // snapshot the configuration values to reduce the number of instructions that need to execute in the lock.
+            _concurrentRequestLimit = configuration.ConcurrentRequestLimit;
```
namadabu, line 79: Would it be useful to validate here that `_concurrentRequestLimit` is > 0 (to avoid returning 429s for all requests)?

namadabu (approved):
:shipit:

src/Microsoft.Health.Fhir.Api/Features/Throttling/ThrottlingMiddleware.cs
johnstairs: We only access in the the scope of the lock, so no need.

Issue #1500, opened 2020-12-08
Provenance in header for PUT and POST


ellgre:
**Question**

According to the hl7 documentation it should be possible to use the header of a PUT or POST in order to create a provenance resource linked to the updated resource see https://www.hl7.org/fhir/provenance.html#header. I have tried to use the x-provenance as described but I cannot find a provenance being generated.

Is this functionality supported in the OSS cosmos solution? If yes could you please provide me with an example of how to do it? If no, do you have any plans to implement the functionality?

brendankowitz:
Hi @ellgre, sorry we don't currently support this header

CaitlinV39:
AB#79231

CaitlinV39:
@ellgre - We recently merged support for this header. Please let us know if you have any additional questions!

PR #1618, opened 2021-01-28
Provenance header support

Ivanidzo4ka:
## Description
Adds support for "x-provenance" header for POST/PUT operations 

## Related issues
Addresses [#1500].

## Testing
Unit tests and manual testing

## FHIR Team Checklist

+semver: feature


src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/ProvenanceHeaderConditionalCreateHandler.cs
```diff
@@ -0,0 +1,76 @@
...
+        public ProvenanceHeaderConditionalCreateHandler(FhirJsonParser fhirJsonParser, IMediator mediator)
+        {
+            EnsureArg.IsNotNull(fhirJsonParser, nameof(fhirJsonParser));
+            EnsureArg.IsNotNull(mediator, nameof(mediator));
+
+            _fhirJsonParser = fhirJsonParser;
+            _mediator = mediator;
+        }
+
+        public async Task<UpsertResourceResponse> Handle(ProvenanceHeaderConditionalCreateRequest request, CancellationToken cancellationToken)
```
brendankowitz, line 38: I would have probably explored a completely different approach :P, something along the lines of:

- Create a new PipelineBehavior that intercepts Create/Upsert requests
- Inject IMediator, IRequestContext 
- Check IRequestContext for ProvenanceHeader - then this similar logic

Ivanidzo4ka: Yeah, I'm quite unhappy with 4 different handles which all do the same thing.
My thoughts was to unite all 4 original requests classes into one with parameter which specifies behavior, and have one handler to rule them all and one behavior pipeline.

IRequestContext should do the trick, but it's somehow not pure enough for my taste (coming from person who copypaste 4 classes) it's some side thing related to asp.net pipeline rather than pure classes which we send to mediator.
I'll give it a try and update PR.

src/Microsoft.Health.Fhir.Tests.Common/TestFiles/R4/ProvenanceHeader.json
```diff
@@ -0,0 +1 @@
+﻿{"resourceType":"Provenance","id":"8a42c91e-dd16-4a9c-aed5-637076d8d7de","meta":{"versionId":"1","lastUpdated":"2021-01-28T21:22:47.984+00:00"},"text":{"status":"generated","div":"<div xmlns=\"http:\/\/www.w3.org\/1999\/xhtml\">procedure record authored on 27-June 2015 by Harold Hippocrates, MD Content extracted from XDS managed CDA Referral received 26-June as authorized by a referenced Consent.<\/div>"},"occurredPeriod":{"start":"2015-06-27","end":"2015-06-28"},"recorded":"2015-06-27T08:39:24+10:00","policy":["http:\/\/acme.com\/fhir\/Consent\/25"],"location":{"reference":"Location\/1"},"reason":[{"coding":[{"system":"http:\/\/snomed.info\/sct","code":"3457005","display":"Referral"}]}],"agent":[{"type":{"coding":[{"system":"http:\/\/terminology.hl7.org\/CodeSystem\/v3-ParticipationType","code":"AUT"}]},"who":{"reference":"Practitioner\/xcda-author"}},{"id":"a1","type":{"coding":[{"system":"http:\/\/terminology.hl7.org\/CodeSystem\/v3-ParticipationType","code":"DEV"}]},"who":{"reference":"Device\/software"}}],"entity":[{"role":"source","what":{"reference":"DocumentReference\/example","display":"CDA Document in XDS repository"}}]}
```
brendankowitz, line 1: Its hard to see from comparison, but if these are all the same you can just add a single file to `TestFiles/Normative/ProvenanceHeader.json` and it load for all versions

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/ProvenanceHeaderBehavior.cs
```diff
@@ -0,0 +1,112 @@
...
+            {
+                // Set target to provided resource.
+                provenance.Target = new System.Collections.Generic.List<ResourceReference>()
+                {
+                    new ResourceReference($"{response.Outcome.RawResourceElement.InstanceType}/{response.Outcome.RawResourceElement.Id}"),
+                };
+
+                // Create Provenance resource.
+                // TODO: It should probaby go through controller to trigger audit events, but it's quite tricky to do now.
+                await _mediator.Send<UpsertResourceResponse>(new CreateResourceRequest(provenance.ToResourceElement()), cancellationToken);
```
feordin, line 81: Is it possible that this could fail, perhaps with a more in depth validation beyond what the FhirJsonParser performs?  If it did fail, should we catch and throw a ProvenanceHeaderException here as well?

Ivanidzo4ka: I don't think so.
So there is lot of things can happen here, persistence storage is down, validation failure, maybe some weird misconfiguration with search parameters, who knows... And all this cases will throw thier own exception which I need to catch and wrap in ProvenanceHeaderException? I would rather just let them go to the top, I'm not sure it's there is much benefit I would gain from throwing specific exception.

Ivanidzo4ka:
Closing since I can't make figure out why PR Internal Checks is not working

PR #1630, opened 2021-02-02
Turn off splitting canonical for Uri search queries

brendankowitz:
## Description

Splitting Canonical Components for incoming URI queries may break existing searches.

The example for the provided E2E tests is that the following search should continue to work:
```
/ValueSet?url=http://example.org/rdf#54135-9
```
With the previous code the `#54135-9` would be removed from the search query.

## Related issues
Refs #882

## Testing
Updates tests

## FHIR Team Checklist

Patch


src/Microsoft.Health.Fhir.Core/Data/R4/unsupported-search-parameters.json
```diff
@@ -55,6 +55,7 @@
         "http://hl7.org/fhir/SearchParameter/OperationDefinition-output-profile",
         "http://hl7.org/fhir/SearchParameter/PlanDefinition-composed-of",
         "http://hl7.org/fhir/SearchParameter/PlanDefinition-depends-on",
+        "http://hl7.org/fhir/SearchParameter/PlanDefinition-definition",
```
brendankowitz, line 58: Moving back to unsupported as this has Canonical=>Reference as well

feordin (approved):

PR #1643, opened 2021-02-05
Fix relaying x-ms-retry-after-ms header from Cosmos DB

johnstairs:
## Description
We were not relaying any retry delay information when getting throttled by Cosmos DB because we were looking for the wrong header name.

## Testing
Unit and manual testing

Patch


namadabu (approved):
:shipit:

brendankowitz (approved):

PR #1646, opened 2021-02-05
Cosmos sort mvp [WIP]

brendankowitz:
## Description

## Related issues
Addresses [issue #].

## Testing

## FHIR Team Checklist


src/Microsoft.Health.Fhir.Core/Features/Search/SearchIndexEntryExtensions.cs
```diff
@@ -0,0 +1,75 @@
...
+    internal static class SearchIndexEntryExtensions
+    {
+        public static void SetMinMaxValues(this IEnumerable<SearchIndexEntry> entries)
+        {
+            // determine min/max values
+            foreach (var item in entries.GroupBy(x => x.SearchParameter.Name))
+            {
+                switch (item.First().SearchParameter.Type)
+                {
+                    case SearchParamType.String:
```
johnstairs, line 23: I think we should only handle type we know we can/should support. For example, what does it mean to sort by quantity when there are different units, systems, etc. 
I would do string  and date for now

Issue #1564, opened 2020-12-31
Reindex fails on bad type


LTA-Thinking:
**Describe the bug**
If a reindexed search parameter references a FHIR type that doesn't exist when the reindex job runs it will get a null ref error in the search option factory line 116. This is the same issue as export ran into in #1560. 

**FHIR Version?**
Stu3/R4/R5

**Data provider?**
CosmosDB/SQL Server

**To Reproduce**
1. Create a SearchParameter for an invalid type
2. Run a reindex job

**Expected behavior**
Gracefully handle the issue. Either ignore it and continue with other search parameters, or record a useful error in the reindex job record.

**Actual behavior**
An unhelpful error of 'Object reference not set to an instance of an object.' is recorded to the job record.

CaitlinV39:
AB#78774

CaitlinV39:
AB#78774

PR #1647, opened 2021-02-05
CosmosDB configuration-based _sort

brendankowitz:
## Description

* Sort index affects search parameter hash, so reindexing will pick this up
* Enables string and date sort in CosmosDb
* Provides a list of example sort parameters that can be enabled on new dbs

Usage:
```
@host = localhost:44348

###
# test

https://{{host}}/Patient?_sort=family

###

https://{{host}}/Patient?_sort=-birthdate
```

## Related issues
Addresses [AB#79408](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/79408).

## Testing

## FHIR Team Checklist

Patch


src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -115,8 +109,8 @@ public async Task ExecuteAsync(ReindexJobRecord reindexJobRecord, WeakETag weakE
                         }
                     }
 
-                    _reindexJobRecord.Resources.AddRange(resourceList);
-                    _reindexJobRecord.SearchParams.AddRange(notYetIndexedParams.Select(p => p.Url.ToString()));
+                    _reindexJobRecord.Resources.AddRange(resourceList.Where(x => ModelInfoProvider.IsKnownResource(x)));
```
brendankowitz, line 112: Related to #1564

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/SortValue.cs
```diff
@@ -0,0 +1,34 @@
...
+            SearchParameterUri = searchParameterUri;
+        }
+
+        public SortValue()
+        {
+        }
+
+        public ISearchValue Low { get; }
+
+        public ISearchValue Hi { get; }
```
namadabu, line 30: nit: maybe `High` instead of `Hi`?

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
```diff
@@ -364,6 +375,37 @@ internal async Task<(IReadOnlyList<T> results, string continuationToken)> Execut
             return (results, page.ContinuationToken);
         }
 
+        private void UpdateSortIndex(FhirCosmosResourceWrapper cosmosWrapper)
+        {
+            var searchParameters = _supportedSearchParameters.Value.GetSearchParameters(cosmosWrapper.ResourceTypeName)
```
namadabu, line 380: nit: Would be easier to read the code if we used the explicit type instead of var here.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
```diff
@@ -364,6 +375,37 @@ internal async Task<(IReadOnlyList<T> results, string continuationToken)> Execut
...
+                        cosmosWrapper.SortValues.Remove(item.Key);
+                    }
+                }
+
+                foreach (SearchParameterInfo field in searchParameters.Values)
+                {
+                    if (cosmosWrapper.SortValues.All(x => x.Value.SearchParameterUri != field.Url))
+                    {
+                        // Ensure sort property exists
+                        cosmosWrapper.SortValues.Add(field.Code, new SortValue());
```
namadabu, line 399: >cosmosWrapper.SortValues.Add(field.Code, new SortValue()); [](start = 24, length = 58)

nit: Would it be safer to add the Uri to this empty SortValue? I know we are not going to de-serialize the searchindexentry when reading resources. But I am afraid that somewhere in some change we are going to pass a FhirCosmosWrapper that has existing sort values and then this part will blow up because we already have an entry for the Code (and it was not caught because we were comparing only Uri).

src/Microsoft.Health.Fhir.Shared.Web/appsettings.json
```diff
@@ -116,7 +116,13 @@
         "IndividualBatchActionRetryOptions" : {
             "MaxNumberOfRetries": 18,
             "MaxWaitTimeInSeconds": 90
-        }
+        },
+        "SortConfigOperationEnabled": true,
```
namadabu, line 120: This does not seemed to be used anywhere in the code.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Search/SortValueJObjectGenerator.cs
```diff
@@ -0,0 +1,101 @@
...
+            return CurrentEntry;
+        }
+
+        public void Visit(CompositeSearchValue composite)
+        {
+        }
+
+        public void Visit(DateTimeSearchValue dateTime)
+        {
+            AddProperty(_prefix, dateTime.Start.ToString("o", CultureInfo.InvariantCulture));
```
namadabu, line 59: Should this be` dateTime.End` when _prefix is `SearchValueConstants.SortHighValueFieldName`?

src/Microsoft.Health.Fhir.Core/Features/Search/SearchHelperUtilities.cs
```diff
@@ -29,13 +30,18 @@ public static string CalculateSearchParameterHash(IEnumerable<SearchParameterInf
...
+            var sb = new StringBuilder();
             foreach (SearchParameterInfo searchParamInfo in searchParamaterInfos.OrderBy(x => x.Url.ToString()))
             {
-                sb.Append(searchParamInfo.Url.ToString());
-                sb.Append(searchParamInfo.Type.ToString());
+                sb.Append(searchParamInfo.Url);
+                sb.Append(searchParamInfo.Type);
                 sb.Append(searchParamInfo.Expression);
 
+                if (searchParamInfo.SortStatus != SortParameterStatus.Disabled)
```
namadabu, line 40: Would be useful to add some UTs for this code change.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
brendankowitz: Lol, Resharper makes me forget this rule :P
image

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Search/SortValueJObjectGenerator.cs
brendankowitz: Thanks, good call. I reverted this when I changed to a single field

src/Microsoft.Health.Fhir.Core/Features/Definition/SupportedSearchParameterDefinitionManager.cs
```diff
@@ -74,7 +75,7 @@ public SearchParameterInfo GetSearchParameter(Uri definitionUri)
 
         public IEnumerable<SearchParameterInfo> GetSupportedButNotSearchableParams()
         {
-            return _inner.AllSearchParameters.Where(p => p.IsSearchable == false && p.IsSupported == true);
+            return AllSearchParameters.Where(p => p.IsSearchable == false || p.SortStatus == SortParameterStatus.Supported);
```
feordin, line 78: Is the case of IsSearchable == false, and SortStatus == NotSupported and p.IsSupported == true, possible?

brendankowitz: In CosmosDB, SortStatus will only selectively be enabled on Search parameters we can both index and types we can sort on

feordin: Yep, that makese sense.  Not all parameters are sortable. 
A SearchParameter which is supported, but not sortable would have the following properties:
p.IsSearchable == false
p.IsSupported == true
p.SortStatus == SortParameterStatus.Disabled
In which case I believe it should be included in the returned list of params, but as is would not be.

Also, we may have the case that we have a parameter which is possible to sort, and also already Searchable:
p.IsSearchable == true
p.SortStatus == Supported
This parameter would be included in the results, but should not be.

feordin: Discussed with Brendan, and realized that I missed the change to using the AllSearchParams rather than _inner. which applies the IsSupported filter.  Looks good.

feordin (approved):

src/Microsoft.Health.Fhir.Core/Features/Search/Registry/ISearchParameterStatusDataStore.cs
```diff
@@ -12,6 +14,8 @@ public interface ISearchParameterStatusDataStore
     {
         Task<IReadOnlyCollection<ResourceSearchParameterStatus>> GetSearchParameterStatuses();
 
+        Task<ResourceSearchParameterStatus> GetSearchParameterStatus(Uri searchParameterUri, CancellationToken cancellationToken);
```
johnstairs, line 17: Looks like this is not used?

brendankowitz: I can revert from this PR if it makes sense, I was it to load a single status with the $configure-sort operation

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Search/SearchHelperUtilitiesTests.cs
```diff
@@ -14,20 +14,35 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Search
 {
     public class SearchHelperUtilitiesTests
```
namadabu, line 15: >SearchHelperUtilitiesTests [](start = 17, length = 26)

Test class name needs to be updated

src/Microsoft.Health.Fhir.Core/Features/Search/Registry/ISearchParameterStatusDataStore.cs
brendankowitz: Done

src/Microsoft.Health.Fhir.Core/Models/SortParameterStatus.cs
```diff
@@ -0,0 +1,14 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Fhir.Core.Models
+{
+    public enum SortParameterStatus
```
johnstairs, line 8: I suggest putting some comments on these to explain what thy mean (especially Supported vs Enabled)

src/Microsoft.Health.Fhir.CosmosDb/Configs/CosmosDataStoreConfiguration.cs
```diff
@@ -46,5 +46,7 @@ public class CosmosDataStoreConfiguration
         /// This time includes the time to fetch the first page.
         /// </summary>
         public int SearchEnumerationTimeoutInSeconds { get; set; } = 30;
+
+        public HashSet<string> SortSearchParameters { get; } = new HashSet<string>();
```
johnstairs, line 50: Suggest adding a comment that these are expected to be the URIs of the search parameters

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Search/SortValueJObjectGenerator.cs
```diff
@@ -0,0 +1,131 @@
...
+                    break;
+                case SearchValueConstants.SortHighValueFieldName:
+                    AddProperty(_prefix, number.High ?? number.Low);
+                    break;
+                default:
+                    throw new NotSupportedException();
+            }
+        }
+
+        public void Visit(QuantitySearchValue quantity)
```
johnstairs, line 87: I'm hesitant to support Quantity since we don't do any unit normalization

brendankowitz: I removed this. It isn't able to be used until `QuantitySearchValue` implements `ISupportSortSearchValue` anyway

johnstairs:
I'm getting this exception:
>System.ArgumentException: Can not add property l to Newtonsoft.Json.Linq.JObject. Property with the same name already exists on object.
   at Newtonsoft.Json.Linq.JObject.ValidateToken(JToken o, JToken existing)
   at Newtonsoft.Json.Linq.JContainer.InsertItem(Int32 index, JToken item, Boolean skipParentCheck)
   at Microsoft.Health.Fhir.CosmosDb.Features.Storage.Search.SortValueJObjectGenerator.AddProperty(String name, Object value) in D:\repos\fhir-server\src\Microsoft.Health.Fhir.CosmosDb\Features\Storage\Search\SortValueJObjectGenerator.cs:line 128

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Search/SortValueJObjectGenerator.cs
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        public void Visit(UriSearchValue uri)
+        {
+            AddProperty(_prefix, uri.Uri);
+        }
+
+        private void CreateEntry()
+        {
+            CurrentEntry = new JObject();
```
johnstairs, line 111: This instance seems to be reused across threads? How is this thread-safe?

src/Microsoft.Health.Fhir.Core/Models/SortParameterStatus.cs
brendankowitz: Done

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Search/SortValueJObjectGenerator.cs
brendankowitz: `SortValueJObjectGenerator` is not thread-safe. I've updated `SortEntryConverter` to manage access

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
```diff
@@ -364,6 +375,37 @@ internal async Task<(IReadOnlyList<T> results, string continuationToken)> Execut
             return (results, page.ContinuationToken);
         }
 
+        private void UpdateSortIndex(FhirCosmosResourceWrapper cosmosWrapper)
+        {
+            Dictionary<string, SearchParameterInfo> searchParameters = _supportedSearchParameters.Value.GetSearchParameters(cosmosWrapper.ResourceTypeName)
```
johnstairs, line 380: Can we avoid recomputing this dictionary with every request?

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
```diff
@@ -364,6 +375,37 @@ internal async Task<(IReadOnlyList<T> results, string continuationToken)> Execut
...
+                .Where(x => x.SortStatus != SortParameterStatus.Disabled)
+                .ToDictionary(x => x.Code);
+
+            if (searchParameters.Any())
+            {
+                foreach (KeyValuePair<string, SortValue> item in cosmosWrapper.SortValues.ToArray())
+                {
+                    if (!searchParameters.ContainsKey(item.Key))
+                    {
+                        cosmosWrapper.SortValues.Remove(item.Key);
```
johnstairs, line 390: Seems like a shame that we are finding min/max for search parameters where sort is not even enabled. Can we prune this uptream instead of here?

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Search/SortValueJObjectGenerator.cs
```diff
@@ -0,0 +1,119 @@
...
+
+namespace Microsoft.Health.Fhir.CosmosDb.Features.Storage.Search
+{
+    public class SortValueJObjectGenerator : ISearchValueVisitor
+    {
+        private string _prefix;
+
+        private JObject CurrentEntry { get; set; }
+
+        public JObject Generate(SortValue entry)
```
johnstairs, line 24: I know this is how it's done today, but one day, we'll need to reimplement this this using Utf8JsonWriter/Utf8JsonWriter instead of allocating all these intermediate objects.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Search/SortEntryConverter.cs
```diff
@@ -0,0 +1,51 @@
...
+
+namespace Microsoft.Health.Fhir.CosmosDb.Features.Storage.Search
+{
+    public class SortEntryConverter : JsonConverter
+    {
+        private static readonly ConcurrentQueue<SortValueJObjectGenerator> _cachedGenerators = new();
+
+        public override bool CanConvert(Type objectType)
+        {
+            return objectType == typeof(SortValue);
```
johnstairs, line 20: Can we use this converter for the dictionary instead of each individual sort value (so newtonsoft doesn't call this method repeatedly and we don't have to rent an return a cached generator for each sortvalue)? Maybe put an attribute on the property?

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/FhirCosmosResourceWrapper.cs
```diff
@@ -105,5 +112,26 @@ internal string GetETagOrVersion()
...
+
+        public override void UpdateSearchIndices(IReadOnlyCollection<SearchIndexEntry> searchIndices)
+        {
+            base.UpdateSearchIndices(searchIndices);
+
+            UpdateSortIndex(searchIndices);
+        }
+
+        private void UpdateSortIndex(IReadOnlyCollection<SearchIndexEntry> searchIndices)
+        {
```
johnstairs, line 124: Given that there is one min and one max per search parameter, I would do a linear scan over the indexes as build up the dictionary as you encounter mins and maxes

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
brendankowitz: This is a little trickier. This should probably be pushed back up into SupportedSearchParameterDefinitionManager since the states in SearchParameterInfo are loaded from the db and change over time.

The dictionary at worst is being created on only the search params for this resource type. I'll look into this further, but the change carries more risk

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Search/SortEntryConverter.cs
brendankowitz: This can potentially be avoided by setting the itemtypeconverter directly (`[JsonProperty("sort", ItemConverterType = typeof(SearchIndexEntryConverter))]`)

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/ISupportSortSearchValue.cs
```diff
@@ -17,12 +17,12 @@ public interface ISupportSortSearchValue : IRangedComparable
         /// Determines whether this current value is the minimum when compared to
         /// a collection of search values for the same parameter.
         /// </summary>
-        bool IsMin { get; set; }
+        bool? IsMin { get; set; }
```
Ivanidzo4ka, line 20: >bool? [](start = 8, length = 5)

Can I ask what does null value means?
Right now it just answer question: is this value are minimal value or not, and there is yes or no answer. What would null add to that statement?

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/CosmosDbSortingValidator.cs
```diff
@@ -26,7 +36,16 @@ public bool ValidateSorting(IReadOnlyList<(SearchParameterInfo searchParameter,
                     errorMessages = Array.Empty<string>();
                     return true;
                 case { Count: 1 }:
-                    errorMessages = new[] { string.Format(CultureInfo.InvariantCulture, Core.Resources.SearchSortParameterNotSupported, sorting[0].searchParameter.Code) };
+                    (SearchParameterInfo searchParameter, SortOrder sortOrder) parameter = sorting[0];
+
+                    if (parameter.searchParameter.SortStatus == SortParameterStatus.Enabled ||
+                        (parameter.searchParameter.SortStatus == SortParameterStatus.Supported && _contextAccessor.FhirRequestContext?.RequestHeaders.TryGetValue(KnownHeaders.PartiallyIndexedParamsHeaderName, out StringValues _) == true))
```
johnstairs, line 42: ContainsKey()? Or do we want to check for a specific value?

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/ISupportSortSearchValue.cs
brendankowitz: Not set, if the search parameter is not supported (i.e. by config or the database) then the work to figure out min/max is skipped. We then cannot trust that it is the minimum which is the default

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/CosmosDbSortingValidator.cs
brendankowitz: Assuming `_contextAccessor.FhirRequestContext?.RequestHeaders.TryGetValue(KnownHeaders.PartiallyIndexedParamsHeaderName, out StringValues _)`? This is the same check used elsewhere for the search parameters, I'll group this code into an extension so it remains consistent

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/FhirCosmosResourceWrapper.cs
```diff
@@ -105,5 +113,52 @@ internal string GetETagOrVersion()
...
+            SortValues = new Dictionary<string, SortValue>();
+
+            if (searchIndices == null)
+            {
+                return;
+            }
+
+            foreach (SearchIndexEntry entry in searchIndices)
+            {
+                if (entry.Value is ISupportSortSearchValue sortValue && (sortValue.IsMax.HasValue || sortValue.IsMin.HasValue))
```
johnstairs, line 135: Don't you want to check `sortValue.IsMax == true || sortValue.IsMin == true`?

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/ISupportSortSearchValue.cs
Ivanidzo4ka: >it is the minimum which is the default.
Default for bool is false, so I assume you talking about code which set this values. Can we modify that code rather than expand this concept?


---
In reply to: [585764243](https://github.com/microsoft/fhir-server/pull/1647#discussion_r585764243) [](ancestors = 585764243)

johnstairs (approved):

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/FhirCosmosResourceWrapper.cs
```diff
@@ -105,5 +113,52 @@ internal string GetETagOrVersion()
...
+            {
+                if (sortValue.IsMax)
+                {
+                    item.High = entry.Value;
+                }
+
+                if (sortValue.IsMin)
+                {
+                    item.Low = entry.Value;
+                }
```
Ivanidzo4ka, line 160: super nit:
There is check before we call this method `(sortValue.IsMax || sortValue.IsMin)` so I I would went with simple `if else` statement rather than two `if`

src/Microsoft.Health.Fhir.Core/Features/Definition/SupportedSearchParameterDefinitionManager.cs
```diff
@@ -74,7 +74,7 @@ public SearchParameterInfo GetSearchParameter(Uri definitionUri)
 
         public IEnumerable<SearchParameterInfo> GetSupportedButNotSearchableParams()
```
Ivanidzo4ka, line 75: >GetSupportedButNotSearchableParams [](start = 48, length = 34)

Can we rename it to `GetRequiredReindexingSearchParams`

PR #1649, opened 2021-02-09
Add separate pipeline for export tests

namadabu:
## Description
Added a new pipeline to run some of our long running export e2e tests. These had been disabled because there were issues running them as part of CI. I have setup the pipeline to run once every night (to prevent overloading build agents). 

The pipeline builds the latest branch (to get the tests), deploys the most recent docker image (there is a race condition here where the tests and the image might be of different versions, but the probability of hitting it is very low) and runs the ExportLongRunningtests against STU3 and R4 services. I haven't added R5 since since I couldn't find a way to generate R5 specific data to load onto the server. 

There is still some room for improvement in the pipeline:
1. It would be ideal to download the artifacts from the CI pipeline instead of building again. To do that we need to define id's for the pipelines first.
2. Avoid using a hardcoded nuget version.

## Related issues
Addresses [issue #].

## Testing
Manually ran the pipeline to validate that only expected tests were run.

## FHIR Team Checklist


test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ExportLongRunningTests.cs
```diff
@@ -0,0 +1,462 @@
...
+
+            // Download all resources from fhir server
+            Dictionary<(string resourceType, string resourceId), Resource> dataFromFhirServer = await GetResourcesFromFhirServer(_testFhirClient.HttpClient.BaseAddress);
+
+            // Assert both data are equal
+            Assert.True(ValidateDataFromBothSources(dataFromFhirServer, dataFromExport));
+            Assert.True(blobUris.All((url) => url.OriginalString.Contains(testContainer)));
+        }
+
+        private bool ValidateDataFromBothSources(Dictionary<(string resourceType, string resourceId), Resource> dataFromServer, Dictionary<(string resourceType, string resourceId), Resource> dataFromStorageAccount)
```
LTA-Thinking, line 174: Could we put these helper methods in an ExportTestHelpers file so we don't have duplicate code?

build/export-variables.yml
```diff
@@ -0,0 +1,9 @@
+variables:
+    ResourceGroupRegion: 'southcentralus'
+    resourceGroupRoot: 'msh-fhir-export'
+    appServicePlanName: '$(resourceGroupRoot)-linux'
+    DeploymentEnvironmentName: '$(resourceGroupRoot)'
+    ResourceGroupName: '$(resourceGroupRoot)'
+    TestEnvironmentName: 'OSS Export'
+    ImageTag: 'master'
+    nuGetVersion: '5.8.1'
```
LTA-Thinking, line 9: Where is this variable used?

namadabu: This is used in `package.yml`

LTA-Thinking: Ah, this is one of the variables from Semver. Why don't we use the update-semver script to get this variable like in the ci-pipeline?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ExportLongRunningTests.cs
namadabu: Moved common methods to a helper class - `ExportTestHelper`

build/export-variables.yml
namadabu: I initially had it as part of the pipeline. But I removed it since we aren't really going to use that version information anywhere else. Hence decided to use a hardcoded variable for nuget version. In the future I would like to remove the build step completely and download the artifacts from the CI pipeline (using the `DownloadPipelineArtifact` task). But there are a couple of other changes that need to be done before we can do that.

LTA-Thinking (approved):

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ExportLongRunningTests.cs
```diff
@@ -0,0 +1,177 @@
...
+using Microsoft.Health.Fhir.Tests.Common.FixtureParameters;
+using Microsoft.Health.Fhir.Tests.E2E.Common;
+using Microsoft.Health.Test.Utilities;
+using Xunit;
+using Xunit.Abstractions;
+using Task = System.Threading.Tasks.Task;
+
+namespace Microsoft.Health.Fhir.Tests.E2E.Rest
+{
+    [Trait(Traits.Category, Categories.ExportLongRunning)]
```
LTA-Thinking, line 22: We'll want to add this to the list of tests to skip in PaaS.

PR #1650, opened 2021-02-10
Reindex bug fixes

feordin:
## Description
Fix reindex behavior to properly update parameters when a job has no resources to reindex.  Handle null refs and update SearchParameters.json test file

## Related issues
Addresses [issue [AB#79555](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/79555)].

## Testing
E2E tests and integeration tests were updated, as well as manual tests.

## FHIR Team Checklist

Skip, a bug fix

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Core/Extensions/ReindexJobRecordExtensions.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Extensions/ReindexJobRecordExtensions.cs
@@ -26,6 +26,17 @@ namespace Microsoft.Health.Fhir.Core.Extensions
             parametersResource.Parameter = new List<Parameters.ParameterComponent>();
             parametersResource.Parameter.Add(new Parameters.ParameterComponent() { Name = JobRecordProperties.Id, Value = new FhirString(job.Id) });
 
+            if (job.Error != null && job.Error.Count > 0)
+            {
+                var outputMessages = new List<Parameters.ParameterComponent>();
+                foreach (var error in job.Error)
+                {
+                    outputMessages.Add(new Parameters.ParameterComponent() { Name = error.Code, Value = new FhirString(error.Diagnostics) });
+                }
+
+                parametersResource.Parameter.Add(new Parameters.ParameterComponent() { Name = JobRecordProperties.Output, Part = outputMessages });
+            }
+
             if (job.StartTime.HasValue)
             {
                 parametersResource.Parameter.Add(new Parameters.ParameterComponent() { Name = JobRecordProperties.StartTime, Value = new FhirDateTime(job.StartTime.Value) });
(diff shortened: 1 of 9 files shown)


src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
```diff
@@ -118,7 +118,7 @@ public SearchOptions Create(string compartmentType, string compartmentId, string
 
                     if (badTypes.Count != 0)
                     {
-                        _contextAccessor.FhirRequestContext.BundleIssues.Add(
+                        _contextAccessor.FhirRequestContext?.BundleIssues.Add(
```
brendankowitz, line 121: I wonder if SearchOptionsFactory should demand that `_contextAccessor.FhirRequestContext` is not null, thoughts? 
 Should have a standard way to call this from background jobs? @LTA-Thinking

src/Microsoft.Health.Fhir.Tests.Common/TestFiles/Normative/SearchParameter.json
```diff
@@ -32,7 +32,7 @@
         }
     ],
     "description": "A server defined search that may match any of the string fields in the HumanName, including family, give, prefix, suffix, suffix, and/or text",
-    "code": "name",
+    "code": "foo",
```
brendankowitz, line 35: 🚀

src/Microsoft.Health.Fhir.Shared.Core/Extensions/ReindexJobRecordExtensions.cs
```diff
@@ -26,6 +26,17 @@ public static ResourceElement ToParametersResourceElement(this ReindexJobWrapper
             parametersResource.Parameter = new List<Parameters.ParameterComponent>();
             parametersResource.Parameter.Add(new Parameters.ParameterComponent() { Name = JobRecordProperties.Id, Value = new FhirString(job.Id) });
 
+            if (job.Error != null && job.Error.Count > 0)
+            {
+                var outputMessages = new List<Parameters.ParameterComponent>();
+                foreach (var error in job.Error)
+                {
+                    outputMessages.Add(new Parameters.ParameterComponent() { Name = error.Code, Value = new FhirString(error.Diagnostics) });
```
brendankowitz, line 34: If this is a flat structure there is a helper method: `parametersResource.Add(error.Code, new FhirString(rerror.Diagnostics));`

brendankowitz (approved):

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
feordin: I found this issue because the reindex operation runs a search.  In the case where an operation is using the search service, will there be any FhirRequestContext ?  Should we create one for background workers?

brendankowitz: I noticed we started using it in the export task: https://github.com/microsoft/fhir-server/blob/main/src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs#L140

LTA-Thinking: Sorry I'm late on this. The request context being null was causing exceptions in Export when we ran into bundle issues. There is a similar problem in Reindex: #1564.

LTA-Thinking: This look like it will prevent the exceptions, but the user won't be told that there was an issue.

PR #1658, opened 2021-02-11
Change 429 handling for bundles

johnstairs:
## Description

When a client issues a batch request, and one of those the actions in the bundle fails because of database throttling (a 429), the client does not get a `Retry-After` value to wait before retying the action. Furthermore, it doesn't really make sense to continue processing actions in a bundle if one has already failed due to a 429.

In this change we are:

- Increasing the retries for database operations that occur within a bundle
- If an action in a bundle does result in a 429, all subsequent actions will be skipped and will have the same 429 response.
- `Retry-After`, `x-ms-retry-after-ms`, `x-ms-session-token` and `x-ms-request-charge` will be added as headers to the bundle response if added by individual actions
- Making `x-ms-request-charge` a sum of all request charges for a request
- Fixing error handling for 429s when run outside of an HTTP request (like during export)

## Related issues
[AB#79403](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/79403)

## Testing
Manual testing with a loaded Cosmos DB collection.

Patch


src/Microsoft.Health.Fhir.Core/Features/Context/IFhirRequestContext.cs
```diff
@@ -19,5 +19,10 @@ public interface IFhirRequestContext : IRequestContext
         IList<OperationOutcomeIssue> BundleIssues { get; }
 
         bool IncludePartiallyIndexedSearchParams { get; set; }
+
+        /// <summary>
+        /// Indicates whether this request is part of the execution of a batch or transaction request.
+        /// </summary>
+        bool ExecutingBatchOrTransaction { get; set; }
```
brendankowitz, line 26: @Ivanidzo4ka can this now replace [IProvenanceHeaderState.cs](https://github.com/microsoft/fhir-server/blob/main/src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/IProvenanceHeaderState.cs)?

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/RetryExceptionPolicyFactory.cs
```diff
@@ -14,25 +19,66 @@ namespace Microsoft.Health.Fhir.CosmosDb.Features.Storage
...
+                : (AsyncPolicy)Policy.NoOpAsync();
+
+            _backgroundJobRetryPolicy = CreateExtendedRetryPolicy(3, configuration.RetryOptions.MaxWaitTimeInSeconds * 3);
         }
 
-        public AsyncRetryPolicy CreateRetryPolicy()
+        public AsyncPolicy GetRetryPolicy() =>
+            _requestContextAccessor.FhirRequestContext switch
+            {
+                null => _backgroundJobRetryPolicy,
```
brendankowitz, line 46: Some background jobs are setting this now https://github.com/microsoft/fhir-server/blob/main/src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs#L140

src/Microsoft.Health.Fhir.R4.Web/Properties/launchSettings.json
```diff
@@ -27,7 +27,7 @@
                 "SqlServer:AllowDatabaseCreation": "true",
                 "SqlServer:SchemaOptions:AutomaticUpdatesEnabled": "true",
                 "TestAuthEnvironment:FilePath": "..//..//testauthenvironment.json",
-                "SqlServer:ConnectionString": "server=(local);Initial Catalog=FHIR_R4;Integrated Security=true",
+                "SqlServer:ConnectionString": "server=(local);Initial Catalog=FHIR_R4_100K;Integrated Security=true",
```
brendankowitz, line 30: Revert?

johnstairs: Yup!

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/RetryExceptionPolicyFactory.cs
johnstairs: Thanks for flagging!

johnstairs: addressed here: https://github.com/microsoft/fhir-server/pull/1658/commits/8d1068f1cda03085489c099533dd0cc60d687812

brendankowitz (approved):

PR #1659, opened 2021-02-11
Converts more build tasks to use Linux agents

brendankowitz:
## Description
Moves most build steps to use linux agents.

- Changes GitVersion to use .NET Tool
- Changes E2E tests to use dotnet test
- Changes docker cleanup to use pscore
- Added check to end of deployment step to wait for containers to load
- Increases initial RUs of Integration test Cosmos DB
- Specifies --no-build in package and public steps
- Changes to Linux build for creating packages
- Workaround to provide Storage Account test credentials to concating an 'AllStorageAccounts' string.

Steps not converted:

- Windows Build :)
- Setup AAD
- Cleanup AAD

These changes save a few minutes on the build (~4 mins in the Build and Package Steps). Also 2-3mins in the E2E test steps.

## FHIR Team Checklist

### DIFF
--- a/build/jobs/package.yml
+++ b/build/jobs/package.yml
@@ -23,18 +23,9 @@ steps:
       zipAfterPublish: false
 
   # Package nugets
-  - task: DotNetCoreCLI@2
-    displayName: 'dotnet pack nugets'
-    inputs:
-      command: pack
-      configuration: '$(buildConfiguration)'
-      packDirectory: '$(build.artifactStagingDirectory)/nupkgs'
-      nobuild: true
-      versioningScheme: byEnvVar
-      versionEnvVar: 'nuget_version'
-      zipAfterPublish: true
-    env:	
-      nuget_version: $(nuGetVersion)
+  - powershell: |
+      & dotnet pack $(Build.SourcesDirectory) --output $(Build.ArtifactStagingDirectory)/nupkgs --no-build --configuration=Release -p:PackageVersion=$(nuGetVersion)
+    name: PackNugets
 
   # Publish artifacts
   - task: PublishBuildArtifacts@1
@@ -68,7 +59,7 @@ steps:
   - task: PublishBuildArtifacts@1
     displayName: 'publish test configuration jsons'
     inputs:
-      pathToPublish: './test/configuration/'
+      pathToPublish: './test/Configuration/'
       artifactName: 'deploy'
       artifactType: 'container'
 
--- a/build/jobs/provision-deploy.yml
+++ b/build/jobs/provision-deploy.yml
@@ -29,7 +29,8 @@ parameters:
 jobs:
 - job: provisionEnvironment
   pool:
-    vmImage: $(WindowsVmImage)
+    name: '$(DefaultLinuxPool)'
+    vmImage: '$(LinuxVmImage)'
   steps:
   - task: AzureKeyVault@1
     displayName: 'Azure Key Vault: resolute-oss-tenant-info'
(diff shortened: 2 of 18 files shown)


build/jobs/package.yml
```diff
@@ -92,3 +83,5 @@ steps:
       pathToPublish: '$(build.binariesdirectory)/IntegrationTests'
       artifactName: 'IntegrationTests'
       artifactType: 'container'
+      parallel: true
+      parallelCount: 32
```
LTA-Thinking, line 87: This doesn't look like it is making this step faster. The time seems to be ~5 minutes with or without it.

build/jobs/run-tests.yml
```diff
@@ -32,14 +33,12 @@ jobs:
       azureSubscription: $(ConnectedServiceName)
       KeyVaultName: '${{ parameters.keyVaultName }}-sql'
 
-  - task: VSTest@2
+  - task: DotNetCoreCLI@2
```
LTA-Thinking, line 36: What is the benefit of this switch? The test run times look about the same and we've lost the test run titles.

LTA-Thinking:
Do we want to make the same changes to the CI pipeline? I'm debating whether it is better to be consistent or to have the Window's based build to cover more ground.

build/jobs/package.yml
brendankowitz: Worth a shot, it didn't, I'll revert

build/jobs/run-tests.yml
brendankowitz: `VSTest` is windows only. Do you mean the test names in the output?

brendankowitz:
> Do we want to make the same changes to the CI pipeline? I'm debating whether it is better to be consistent or to have the Window's based build to cover more ground.

Possibly. The advantage here is we can change the Pipeline to a self-hosted VMSS Cluster that can spin with demand (i.e. 50+ agents, the subscription is the limit), or we can use the hosted agents

build/jobs/run-tests.yml
LTA-Thinking: Yes. Previously we had output that looked like this:
image
But now it looks like this:
image

Doesn't effect the outcome, but makes it harder to debug.

brendankowitz: Thanks for the pickup, looks like I accidentally just removed this line

LTA-Thinking (approved):

build/jobs/provision-deploy.yml
```diff
@@ -88,3 +89,21 @@ jobs:
         Write-Host "Setting Key Vault access"
         Set-AzKeyVaultAccessPolicy -VaultName $webAppName -ObjectId 4d4d503d-9ca8-462e-9e18-b35fc8b5285b -PermissionsToSecrets list,get
 
+        $healthCheckUrl = "https://${{ parameters.webAppName }}.azurewebsites.net/health/check"
```
namadabu, line 92: Would it be useful to pull this health check into a separate job so that we can call this from redeploy-webapp.yml as well?

namadabu (approved):
:shipit:

build/jobs/package.yml
```diff
@@ -23,18 +23,9 @@ steps:
...
-      configuration: '$(buildConfiguration)'
-      packDirectory: '$(build.artifactStagingDirectory)/nupkgs'
-      nobuild: true
-      versioningScheme: byEnvVar
-      versionEnvVar: 'nuget_version'
-      zipAfterPublish: true
-    env:	
-      nuget_version: $(nuGetVersion)
+  - powershell: |
+      & dotnet pack $(Build.SourcesDirectory) --output $(Build.ArtifactStagingDirectory)/nupkgs --no-build --configuration=Release -p:PackageVersion=$(nuGetVersion)
```
Ivanidzo4ka, line 27: >--no-build [](start = 96, length = 10)

Is it worth to add `--no-restore` here?

PR #1663, opened 2021-02-12
Upgrade to latest StyleCop for C# 9 support

johnstairs:
## Description
Upgrade to StyleCop.Analyzers 1.2.0-beta.321 to enable C# 9.0 support. Needed to fix up a few small issues it started flagging. 
Also opted to suppress [SA1316TupleElementNamesShouldUseCorrectCasing](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1316.md) because we haven't been very consistent with the casing of tuple fields.

Patch


src/Microsoft.Health.Fhir.Core/Features/Definition/CompartmentDefinitionManager.cs
```diff
@@ -91,7 +91,7 @@ internal void Build(BundleWrapper bundle)
             _compartmentSearchParamsLookup = BuildResourceTypeLookup(compartmentLookup.Values);
         }
 
-        private static Dictionary<CompartmentType, (CompartmentType, Uri, IList<(string, IList<string>)>)> ValidateAndGetCompartmentDict(BundleWrapper bundle)
+        private static Dictionary<CompartmentType, (CompartmentType Code, Uri Url, IList<(string Resource, IList<string> Params)> Resources)> ValidateAndGetCompartmentDict(BundleWrapper bundle)
```
Ivanidzo4ka, line 94: >Code [](start = 68, length = 4)

super nit: I think you going with lowercasing in all other cases. Probably artefact from times when SA1316 wasn't disabled.

Ivanidzo4ka (approved):
:shipit:

src/Microsoft.Health.Fhir.Core/Features/Definition/CompartmentDefinitionManager.cs
johnstairs: I copied from [line 174](https://github.com/microsoft/fhir-server/blob/5f29ada1ffd8997987bb5bf6ac747ab28fdfc759/src/Microsoft.Health.Fhir.Core/Features/Definition/CompartmentDefinitionManager.cs#L174) in this file :)
I wasn't sure what we prefer. I've used camelCasing

brendankowitz: So which is preferred?
I always thought `tuple.Prop` seemed more expected than `tuple.prop`?

johnstairs: In the ASP.net core repo, they use `tuple.prop` ([example](https://github.com/aspnet/AspNetCore/blob/b93bc433db66175d2b07b128ec9990f7a4dd7e1b/src/Mvc/Mvc.Core/src/ApplicationModels/ActionAttributeRouteModel.cs#L144-L144))

PR #1666, opened 2021-02-15
Restrict resources related to profiles behind role

Ivanidzo4ka:
## Description
Add new role and add checks what Create Update Delete operations for related to profiles resources is hidden behind it.

## Testing
Manual and unit tests.

## FHIR Team Checklist

Feature (it's a extension of existing functionality)


src/Microsoft.Health.Fhir.Core/Features/Security/DataActions.cs
```diff
@@ -23,8 +23,9 @@ public enum DataActions : ulong
         ResourceValidate = 1 << 5,
         Reindex = 1 << 6,
         ConvertData = 1 << 7,
+        ProfileAdmin = 1 << 8, // Allows to Create/Update/Delete resources related to profile's resources.
 
         [EnumMember(Value = "*")]
-        All = (ConvertData << 1) - 1,
+        All = (ProfileAdmin << 1) - 1,
```
Ivanidzo4ka, line 29: Happy to hear better name suggestions :)

namadabu: What about `ProfileEditor` ?

---
In reply to: [576425352](https://github.com/microsoft/fhir-server/pull/1666#discussion_r576425352) [](ancestors = 576425352)

src/Microsoft.Health.Fhir.Shared.Web/roles.json
```diff
@@ -40,7 +40,8 @@
                 "*"
             ],
             "notDataActions": [
-                "hardDelete"
+                "hardDelete",
+                "profileAdmin"
```
namadabu, line 44: Would it be better to add a separate role for ProfileAdmin so that a user can be given profile editing permissions without having to be given global admin role?

src/Microsoft.Health.Fhir.Core/Features/Security/DataActions.cs
Ivanidzo4ka: Sounds quite the same to me. I'm mostly unhappy with word `Profile` since it's not exactly profile, it's just bunch of resources used during profile validation

src/Microsoft.Health.Fhir.Shared.Web/roles.json
Ivanidzo4ka: Maybe... I actually not sure, is this roles.json set in stone for PaaS, or they can be modified, and this is just sample of possible assignment?

src/Microsoft.Health.Fhir.Core/Features/Security/DataActions.cs
namadabu: I feel that `Admin` usually refers to the person/role who is in charge of the whole server. Hence the preference for `Editor`. Also, is it expected that a user who does CUD operations on profile also follows it up with a reindex operation? If so, would it make sense to combine the roles into a single one instead of multiple roles?

---
In reply to: [577313196](https://github.com/microsoft/fhir-server/pull/1666#discussion_r577313196) [](ancestors = 577313196)

src/Microsoft.Health.Fhir.Shared.Web/roles.json
namadabu: Not sure. I think @johnstairs would know the answer to that.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/ProfileResourcesBehaviour.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        public async Task<UpsertResourceResponse> Handle(CreateResourceRequest request, CancellationToken cancellationToken, RequestHandlerDelegate<UpsertResourceResponse> next)
+            => await GenericHandle(request.Resource.InstanceType, next);
+
+        public async Task<DeleteResourceResponse> Handle(DeleteResourceRequest request, CancellationToken cancellationToken, RequestHandlerDelegate<DeleteResourceResponse> next)
+            => await GenericHandle(request.ResourceKey.ResourceType, next);
+
+        private async Task<TResponse> GenericHandle<TResponse>(string resourceType, RequestHandlerDelegate<TResponse> next)
+        {
+            if (_supportedTypes.Contains(resourceType) && await _authorizationService.CheckAccess(DataActions.ProfileAdmin) != DataActions.ProfileAdmin)
```
brendankowitz, line 55: @CaitlinV39 we should definitely document this. Editing `"ValueSet", "StructureDefinition", "CodeSystem", "ConceptMap"` required this new permission.

src/Microsoft.Health.Fhir.Core/Features/Security/roles.schema.json
```diff
@@ -12,7 +12,8 @@
                 "export",
                 "resourceValidate",
                 "reindex",
-                "convertData"
+                "convertData",
+                "profileAdmin"
```
brendankowitz, line 16: Code in the managed service will need to be updated to support this.

src/Microsoft.Health.Fhir.Shared.Api/Modules/FhirModule.cs
```diff
@@ -159,6 +159,10 @@ ResourceElement SetMetadata(Resource resource, string versionId, DateTimeOffset
             services.Add<ProvenanceHeaderBehavior>().Scoped().AsSelf().AsImplementedInterfaces();
             services.Add<ProvenanceHeaderState>().Scoped().AsSelf().AsImplementedInterfaces();
 
+            // Register pipeline behavior to check service permission for CUD actions on StructuredDefinition,ValueSet,CodeSystem, ConceptMap.
+
+            services.Add<ProfileResourcesBehaviour>().Scoped().AsSelf().AsImplementedInterfaces();
```
brendankowitz, line 164: These will be picked up by convention as transient. We don't have a "ReplaceAsImplementedInterfaces" function so it might be better to individually use:

```suggestion
            services.Add<ProfileResourcesBehaviour>()
            .Scoped()
            .AsSelf()
            .ReplaceService<IPipelineBehavior<CreateResourceRequest, UpsertResourceResponse>>()
            .ReplaceService<IPipelineBehavior<UpsertResourceRequest, UpsertResourceResponse>>()
            .ReplaceService<IPipelineBehavior<ConditionalCreateResourceRequest, UpsertResourceResponse>>()
            .ReplaceService<IPipelineBehavior<ConditionalUpsertResourceRequest, UpsertResourceResponse>>()
            .ReplaceService<IPipelineBehavior<DeleteResourceRequest, DeleteResourceResponse>>();
```

brendankowitz: Perhaps we should consider also removing this default?

https://github.com/microsoft/fhir-server/blob/d40d65ba376db41fb4dd0616b3428961aea8107b/src/Microsoft.Health.Fhir.Shared.Api/Modules/MediationModule.cs#L28-L32

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/BasicAuthTests.cs
```diff
@@ -251,5 +251,37 @@ public async Task GivenAUserWithConvertDataPermissions_WhenConvertData_TheServer
...
+        {
+            TestFhirClient tempClient = _client.CreateClientForUser(TestUsers.ReadWriteUser, TestApplications.NativeClient);
+            var resource = Samples.GetJsonSample("ValueSet").ToPoco<ValueSet>();
+            FhirException fhirException = await Assert.ThrowsAsync<FhirException>(async () => await tempClient.CreateAsync<ValueSet>(resource));
+            Assert.Equal(ForbiddenMessage, fhirException.Message);
+            Assert.Equal(HttpStatusCode.Forbidden, fhirException.StatusCode);
+
+            fhirException = await Assert.ThrowsAsync<FhirException>(async () => await tempClient.UpdateAsync<ValueSet>(resource));
+            Assert.Equal(ForbiddenMessage, fhirException.Message);
+            Assert.Equal(HttpStatusCode.Forbidden, fhirException.StatusCode);
```
brendankowitz, line 267: nit: I think these additional calls/asserts should be in different tests

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/ProfileResourcesBehaviour.cs
Ivanidzo4ka: I need to remove "ConceptMap" from this list (I actually have that list as static collection in my other PR, just need to merge it and reuse), but, it's need to be documented, I agree.

src/Microsoft.Health.Fhir.Shared.Api/Modules/FhirModule.cs
Ivanidzo4ka: I removed that.
I also not sure why did I choose them to be scoped and not singleton. It's a common behavior for all requests, what's the point to recreate it all the time?


---
In reply to: [583244843](https://github.com/microsoft/fhir-server/pull/1666#discussion_r583244843) [](ancestors = 583244843)

Ivanidzo4ka:
@brendankowitz any chance you can have review it again?

src/Microsoft.Health.Fhir.Shared.Core/Features/Validation/ServerProvideProfileValidation.cs
```diff
@@ -42,7 +42,7 @@ public ServerProvideProfileValidation(Func<IScoped<ISearchService>> searchServic
             _validateOperationConfig = options.Value;
         }
 
-        public static IEnumerable<string> GetProfilesTypes => _supportedTypes;
+        public HashSet<string> GetProfilesTypes() => _supportedTypes;
```
brendankowitz, line 45: nit: Why not use a property of return type `IReadOnlySet<string>`?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/ProfileResourcesBehaviour.cs
```diff
@@ -0,0 +1,63 @@
...
+        public async Task<DeleteResourceResponse> Handle(DeleteResourceRequest request, CancellationToken cancellationToken, RequestHandlerDelegate<DeleteResourceResponse> next)
+            => await GenericHandle(request.ResourceKey.ResourceType, next);
+
+        private async Task<TResponse> GenericHandle<TResponse>(string resourceType, RequestHandlerDelegate<TResponse> next)
+        {
+            var resources = _profilesResolver.GetProfilesTypes();
+            if (resources.Contains(resourceType) && await _authorizationService.CheckAccess(DataActions.ProfileAdmin) != DataActions.ProfileAdmin)
+            {
+                throw new UnauthorizedFhirActionException();
+            }
```
brendankowitz, line 58: It looks like introducing this could potentially be breaking to existing customers, as in, before they could use these resource types, and now we will return an error.

Is there a plan for how can a FHIR Instance opt-in to this behavior, or migrate to keep current behavior?

src/Microsoft.Health.Fhir.Shared.Api/Modules/FhirModule.cs
brendankowitz: I'd say this purely depends on the expected lifetime within the PipelineBehavior

src/Microsoft.Health.Fhir.Shared.Web/roles.json
Ivanidzo4ka, line 44: >profileAdmin [](start = 17, length = 12)

move to globalAdmins

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/BasicAuthTests.cs
```diff
@@ -240,5 +240,54 @@ public async Task GivenAUserWithConvertDataPermissions_WhenConvertData_TheServer
             var result = await tempClient.ConvertDataAsync(parameters);
             Assert.NotEmpty(result);
         }
+
+        [Fact]
+        [Trait(Traits.Priority, Priority.One)]
+        public async Task GivenUserWithNoProfileAdminPermission_WhenCreate_ThenServerShouldReturnForbidden()
```
LTA-Thinking, line 246: Should the test name reference that the user is trying to create (or other CUD action) a resource type used by profile? Right now it reads as if a user needs profile admin permission to perform CUD actions on any resource type.

LTA-Thinking, line 246: Should there be tests for conditional create and conditional update as well?

src/Microsoft.Health.Fhir.Shared.Api/Modules/MediationModule.cs
```diff
@@ -25,12 +25,6 @@ public void Load(IServiceCollection services)
 
             services.AddMediatR(KnownAssemblies.All);
 
-            Predicate<Type> isPipelineBehavior = y => y.IsGenericType && y.GetGenericTypeDefinition() == typeof(IPipelineBehavior<,>);
-
-            services.TypesInSameAssembly(KnownAssemblies.All)
-                .Transient()
-                .AsImplementedInterfaces(isPipelineBehavior);
-
```
LTA-Thinking, line 33: What did this do? It doesn't look like it fits with the rest of the code review.

src/Microsoft.Health.Fhir.Core/Features/Security/DataActions.cs
LTA-Thinking: How long do we want these names to be? Something like `ProfileDefinitionsEditor` adds more explanation.

src/Microsoft.Health.Fhir.Tests.Common/TestFiles/Normative/ValueSet.json
```diff
@@ -0,0 +1,45 @@
...
+        {
+            "coding": [
+                {
+                    "system": "urn:iso:std:iso:3166",
+                    "code": "US"
+                }
+            ]
+        }
+    ],
+    "copyright": "CPT Copyright � 2014 American Medical Association. All rights reserved.",
```
LTA-Thinking, line 37: I don't think we want to be using a copyright for someone else in our test data. Should this file be edited so it isn't a direct copy?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/ProfileResourcesBehaviour.cs
```diff
@@ -0,0 +1,63 @@
...
+        public async Task<UpsertResourceResponse> Handle(CreateResourceRequest request, CancellationToken cancellationToken, RequestHandlerDelegate<UpsertResourceResponse> next)
+            => await GenericHandle(request.Resource.InstanceType, next);
+
+        public async Task<DeleteResourceResponse> Handle(DeleteResourceRequest request, CancellationToken cancellationToken, RequestHandlerDelegate<DeleteResourceResponse> next)
+            => await GenericHandle(request.ResourceKey.ResourceType, next);
+
+        private async Task<TResponse> GenericHandle<TResponse>(string resourceType, RequestHandlerDelegate<TResponse> next)
+        {
+            var resources = _profilesResolver.GetProfilesTypes();
+            if (resources.Contains(resourceType) && await _authorizationService.CheckAccess(DataActions.ProfileAdmin) != DataActions.ProfileAdmin)
```
LTA-Thinking, line 55: Would this catch an action that is part of a transaction or batch request?

Ivanidzo4ka: Yes, it should, bundle just breaks itslef into resources and submit them separatly.

---
In reply to: [598005262](https://github.com/microsoft/fhir-server/pull/1666#discussion_r598005262) [](ancestors = 598005262)

LTA-Thinking (approved):

PR #1668, opened 2021-02-16
Implements search for SQL reindexing

rotodd:
## Description

During a reindexing operation, we carry out a set of searches to retrieve the resources that need to be reindexed. These queries rely on the search parameter hash value to determine if a resource's search indices are up to date or not. 

This PR adds the ability to search for resources that need to be reindexed, or, in other words, resources that have a hash value that differs from the latest hash for their resource type.

## Example

Here is an example of a SQL query that is generated during reindexing:
```
SET STATISTICS IO ON;
SET STATISTICS TIME ON;

SELECT DISTINCT TOP (101) r.ResourceTypeId, r.ResourceId, r.Version, r.IsDeleted, r.ResourceSurrogateId, r.RequestMethod, CAST(1 AS bit) AS IsMatch, CAST(0 AS bit) AS IsPartial, r.IsRawResourceMetaSet, r.SearchParamHash, r.RawResource
FROM dbo.Resource r WITH(INDEX(IX_Resource_ResourceTypeId_ResourceSurrgateId))
WHERE ResourceTypeId = 103
    AND IsHistory = 0
    AND IsDeleted = 0
    AND SearchParamHash != 2F5C6AB1905CB44F005FCB608BE0274A1160D02DC27BF6846460402CDEB27723
OPTION(RECOMPILE)

```

## Notes

Note 1: reindexing search parameters that impact multiple resource types is currently broken. There is a bug to track this (please see [AB#79638](https://microsofthealth.visualstudio.com/Health/_workitems/edit/79638)).

Note 2: this is being merged into a feature branch. There are still work items TBD before this is merged to main:
- Reindexing should not update the version of the resources that are reindexed (this will be fixed in [AB#73395](https://microsofthealth.visualstudio.com/Health/_workitems/edit/73395) and [AB#79121](https://microsofthealth.visualstudio.com/Health/_workitems/edit/79121))
- Storing a resource's search indices in an efficient way could give us performance improvements (this is being tracked in [AB#79648](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Resolute/Stories/?workitem=79648))

## Related issues
Addresses [AB#73616](https://microsofthealth.visualstudio.com/Health/_workitems/edit/73616).

## Testing
Adds new search specific integration tests.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
@@ -859,7 +859,7 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Search.Expressions.Visitors.Q
             {
                 delimited.BeginDelimitedElement();
 
-                StringBuilder.Append(VLatest.Resource.SearchParamHash, tableAlias).Append(" != ").Append(Parameters.AddParameter(_searchParameterHash));
+                StringBuilder.Append("(").Append(VLatest.Resource.SearchParamHash, tableAlias).Append(" != ").Append(Parameters.AddParameter(_searchParameterHash)).Append(" OR ").Append(VLatest.Resource.SearchParamHash, tableAlias).Append(" IS NULL)");
             }
         }
 
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
@@ -31,13 +31,22 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Search.Expressions.Visitors.Q
 
         private int _curFromCteIndex = -1;
         private readonly bool _isHistorySearch;
+        private readonly bool _isReindexSearch;
+        private readonly string _searchParameterHash;
         private int _tableExpressionCounter = -1;
         private SqlRootExpression _rootExpression;
         private readonly SchemaInformation _schemaInfo;
         private bool _sortVisited = false;
         private HashSet<int> _cteToLimit = new HashSet<int>();
 
-        public SqlQueryGenerator(IndentedStringBuilder sb, SqlQueryParameterManager parameters, SqlServerFhirModel model, bool isHistorySearch, SchemaInformation schemaInfo)
+        public SqlQueryGenerator(
+            IndentedStringBuilder sb,
+            SqlQueryParameterManager parameters,
+            SqlServerFhirModel model,
+            bool isHistorySearch,
+            SchemaInformation schemaInfo,
+            bool isReindexSearch,
+            string searchParameterHash)
         {
             EnsureArg.IsNotNull(sb, nameof(sb));
             EnsureArg.IsNotNull(parameters, nameof(parameters));
@@ -49,6 +58,8 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Search.Expressions.Visitors.Q
             Model = model;
             _isHistorySearch = isHistorySearch;
             _schemaInfo = schemaInfo;
+            _isReindexSearch = isReindexSearch;
+            _searchParameterHash = searchParameterHash;
         }
 
         public IndentedStringBuilder StringBuilder { get; }
@@ -185,6 +196,7 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Search.Expressions.Visitors.Q
                     {
                         AppendHistoryClause(delimitedClause);
                         AppendDeletedClause(delimitedClause);
+                        AppendSearchParameterHashClause(delimitedClause);
                     }
                 }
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -124,18 +124,18 @@ protected override async Task<SearchResult> SearchInternalAsync(SearchOptions se
...
             return searchResult;
         }
 
         protected override async Task<SearchResult> SearchHistoryInternalAsync(SearchOptions searchOptions, CancellationToken cancellationToken)
         {
-            return await SearchImpl(searchOptions, true, cancellationToken);
+            return await SearchImpl(searchOptions, true, false, null, cancellationToken);
         }
 
-        private async Task<SearchResult> SearchImpl(SearchOptions searchOptions, bool historySearch, CancellationToken cancellationToken)
```
Ivanidzo4ka, line 138: >bool historySearch [](start = 81, length = 18)

Maybe it's worth to introduce enum: normal, history, reindex instead of adding new bool value. At least if the mutually exclusive.

rotodd: Ah yes, that would look a bit cleaner - but you're right that it would give the impression that `isHistorySearch` and `isReindexSearch` are mutually exclusive, which they are right now but will not always be.

A bit more info: we have (far off) plans to reindex resource history once we enable fine-grained access control policies on historical data. To support that, we will need to have both of these flags set to true for reindex searches.

Ivanidzo4ka: Well, even if they not mutually excusive, you still can go with enum, just make sure it's marked with `[Flags]` attribute [https://docs.microsoft.com/en-us/dotnet/api/system.flagsattribute?view=net-5.0](https://docs.microsoft.com/en-us/dotnet/api/system.flagsattribute?view=net-5.0)

But I don't insist :)

---
In reply to: [577119723](https://github.com/microsoft/fhir-server/pull/1668#discussion_r577119723) [](ancestors = 577119723)

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -844,6 +856,16 @@ private void AppendHistoryClause(in IndentedStringBuilder.DelimitedScope delimit
             }
         }
 
+        private void AppendSearchParameterHashClause(in IndentedStringBuilder.DelimitedScope delimited, string tableAlias = null)
+        {
+            if (_isReindexSearch)
+            {
+                delimited.BeginDelimitedElement();
+
+                StringBuilder.Append(VLatest.Resource.SearchParamHash, tableAlias).Append(" != ").Append(Parameters.AddParameter(_searchParameterHash));
```
Ivanidzo4ka, line 865: >_searchParameterHash [](start = 129, length = 20)

Can you be null, and would `AddParameter` correctly handle null here?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
rotodd: Nice didn't know we could add that attribute to enums. Let me know how ef75010 looks.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
rotodd: Yes you're right - thanks for catching this, added a null check in 5e584d8.

Here is an example of the updated query:

```
SELECT DISTINCT TOP (101) r.ResourceTypeId, r.ResourceId, r.Version, r.IsDeleted, r.ResourceSurrogateId, r.RequestMethod, CAST(1 AS bit) AS IsMatch, CAST(0 AS bit) AS IsPartial, r.IsRawResourceMetaSet, r.SearchParamHash, r.RawResource
FROM dbo.Resource r WITH(INDEX(IX_Resource_ResourceTypeId_ResourceSurrgateId))
WHERE ResourceTypeId = 81
    AND IsHistory = 0
    AND IsDeleted = 0
    AND (SearchParamHash != '4A2D391F6FCCDF97C5FF67C6B79DCABED44A521FE40F17FA9F0F44DEA6435CBD' OR SearchParamHash IS NULL)
OPTION(RECOMPILE)
```

Ivanidzo4ka: My cosmos db emulator returns true for this ``SELECT NULL != 'hash'`, but I haven't looked into documentation regarding null comparison so maybe I'm raising false hope.

(All I'm saying, maybe that OR addition is not necessary, but I'm not sure, also my original comment was somewhat generic, I see object passing, I'm curious how it would handle null, I didn't imply something is wrong here, sorry if I confused you.)

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlSearchType.cs
```diff
@@ -0,0 +1,28 @@
...
+    /// With the exception of None, the flags attribute requires enumeration constants to be powers of two, that is, 1, 2, 4, 8.
+    /// </remarks>
+    [Flags]
+    internal enum SqlSearchType
+    {
+        // Default value, set if we do not need to consider history or reindexing
+        None = 0,
+
+        // Set if we are including previous resource versions or deleted resources
+        History = 1,
```
Ivanidzo4ka, line 23: >1 [](start = 18, length = 1)

I think it's quite common practice to specify them via power of 2 like: 1<<0. 1<<1, 1<<2 and so on. Probably not necessary for 3 values enum.

Ivanidzo4ka (approved):
:shipit:

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlSearchType.cs
rotodd: Oh nice, I like this idea - I think it would prevent someone from specifying `NewSearchType = 3` in the future (I put a comment in 9cdd203 but I think this is even more clear). Added in 1bbad9e!

PR #1672, opened 2021-02-18
Personal/quwan/feature anonymized patient group export

QuanWanxx:
## Description
Support De-identified export for [Patient](https://hl7.org/Fhir/uv/bulkdata/export/index.html#endpoint---all-patients) and [Group](https://hl7.org/Fhir/uv/bulkdata/export/index.html#endpoint---group-of-patients) export.

The behaviors can be viewed at [Using $export command](https://docs.microsoft.com/en-us/azure/healthcare-apis/export-data#using-export-command).

## Examples

 `GET https://<<FHIR service base URL>>/Patient/$export?_container=<<container_name>>&_anonymizationConfig=<<config file name>>&_anonymizationConfigEtag=<<ETag on storage>> `

 `GET https://<<FHIR service base URL>>/Group/[ID]/$export?_container=<<container_name>>&_anonymizationConfig=<<config file name>>&_anonymizationConfigEtag=<<ETag on storage>> `

## Testing
Manual and unit test

## FHIR Team Checklist

Feature (An extension of De-identified export)

### DIFF
--- a/src/Microsoft.Health.Fhir.Api/Features/Filters/ValidateExportRequestFilterAttribute.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/Filters/ValidateExportRequestFilterAttribute.cs
@@ -120,7 +120,10 @@ namespace Microsoft.Health.Fhir.Api.Features.Filters
 
         private bool IsValidAnonymizedExportRequestParam(HttpRequest request, string paramName)
         {
-            if (AnonymizedExportRequestPathRegexs.Any(pathRegex => Regex.IsMatch(request.Path, pathRegex)))
+            if (AnonymizedExportRequestPathRegexs.Any(pathRegex => Regex.IsMatch(
+                input: request.Path,
+                pattern: pathRegex,
+                options: RegexOptions.IgnoreCase | RegexOptions.CultureInvariant)))
             {
                 return _supportedQueryParamsForAnonymizedExport.Contains(paramName);
             }
--- a/src/Microsoft.Health.Fhir.Api/Features/Filters/ValidateExportRequestFilterAttribute.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/Filters/ValidateExportRequestFilterAttribute.cs
@@ -29,9 +29,6 @@ namespace Microsoft.Health.Fhir.Api.Features.Filters
             "ndjson",
         };
 
-        // Support anonymization for basic and patient/group export.
-        // The regex of <id> in group export is [A-Za-z0-9\-\.]{1,64}, from HL7 website https://www.hl7.org/fhir/datatypes.html#id.
-
         private const string PreferHeaderName = "Prefer";
         private const string PreferHeaderExpectedValue = "respond-async";
         private readonly HashSet<string> _supportedQueryParams;
--- a/src/Microsoft.Health.Fhir.Api/Features/Filters/ValidateExportRequestFilterAttribute.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/Filters/ValidateExportRequestFilterAttribute.cs
@@ -5,8 +5,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
-using System.Text.RegularExpressions;
 using EnsureThat;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc.Filters;
@@ -33,17 +31,10 @@ namespace Microsoft.Health.Fhir.Api.Features.Filters
 
         // Support anonymization for basic and patient/group export.
         // The regex of <id> in group export is [A-Za-z0-9\-\.]{1,64}, from HL7 website https://www.hl7.org/fhir/datatypes.html#id.
-        private static readonly List<string> AnonymizedExportRequestPathRegexs = new List<string>()
-        {
-            @"^/[$]export",
-            @"^/Patient/[$]export",
-            @"^/Group/[A-Za-z0-9\-\.]{1,64}/[$]export",
-        };
 
         private const string PreferHeaderName = "Prefer";
         private const string PreferHeaderExpectedValue = "respond-async";
         private readonly HashSet<string> _supportedQueryParams;
-        private readonly HashSet<string> _supportedQueryParamsForAnonymizedExport;
 
         public ValidateExportRequestFilterAttribute()
         {
@@ -55,10 +46,6 @@ namespace Microsoft.Health.Fhir.Api.Features.Filters
                 KnownQueryParameterNames.Container,
                 KnownQueryParameterNames.Format,
                 KnownQueryParameterNames.TypeFilter,
-            };
-
-            _supportedQueryParamsForAnonymizedExport = new HashSet<string>(StringComparer.Ordinal)
-            {
                 KnownQueryParameterNames.AnonymizationConfigurationLocation,
                 KnownQueryParameterNames.AnonymizationConfigurationFileEtag,
             };
@@ -86,7 +73,7 @@ namespace Microsoft.Health.Fhir.Api.Features.Filters
             IQueryCollection queryCollection = context.HttpContext.Request.Query;
             foreach (string paramName in queryCollection?.Keys)
             {
-                if (IsValidBasicExportRequestParam(paramName) || IsValidAnonymizedExportRequestParam(context.HttpContext.Request, paramName))
+                if (IsValidBasicExportRequestParam(paramName))
                 {
                     continue;
                 }
@@ -116,19 +103,6 @@ namespace Microsoft.Health.Fhir.Api.Features.Filters
         private bool IsValidBasicExportRequestParam(string paramName)
         {
             return _supportedQueryParams.Contains(paramName);
-        }
-
-        private bool IsValidAnonymizedExportRequestParam(HttpRequest request, string paramName)
-        {
-            if (AnonymizedExportRequestPathRegexs.Any(pathRegex => Regex.IsMatch(
-                input: request.Path,
-                pattern: pathRegex,
-                options: RegexOptions.IgnoreCase | RegexOptions.CultureInvariant)))
-            {
-                return _supportedQueryParamsForAnonymizedExport.Contains(paramName);
-            }
-
-            return false;
         }
     }
 }
--- a/src/Microsoft.Health.Fhir.Api/Features/Filters/ValidateExportRequestFilterAttribute.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/Filters/ValidateExportRequestFilterAttribute.cs
@@ -5,6 +5,8 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
+using System.Text.RegularExpressions;
 using EnsureThat;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc.Filters;
@@ -29,10 +31,17 @@ namespace Microsoft.Health.Fhir.Api.Features.Filters
             "ndjson",
         };
 
+        // Support anonymization for basic and patient/group export.
+        // The regex of <id> in group export is [A-Za-z0-9\-\.]{1,64}, from HL7 website https://www.hl7.org/fhir/datatypes.html#id.
+        private static readonly List<string> AnonymizedExportRequestPathRegexs = new List<string>()
+        {
+            @"^/[$]export",
+            @"^/Patient/[$]export",
+            @"^/Group/[A-Za-z0-9\-\.]{1,64}/[$]export",
+        };
+
         private const string PreferHeaderName = "Prefer";
         private const string PreferHeaderExpectedValue = "respond-async";
-        private const string DefaultExportRequestPath = "/$export";
-
         private readonly HashSet<string> _supportedQueryParams;
         private readonly HashSet<string> _supportedQueryParamsForAnonymizedExport;
 
@@ -111,7 +120,7 @@ namespace Microsoft.Health.Fhir.Api.Features.Filters
 
         private bool IsValidAnonymizedExportRequestParam(HttpRequest request, string paramName)
         {
-            if (request.Path.StartsWithSegments(DefaultExportRequestPath, StringComparison.InvariantCultureIgnoreCase))
+            if (AnonymizedExportRequestPathRegexs.Any(pathRegex => Regex.IsMatch(request.Path, pathRegex)))
             {
                 return _supportedQueryParamsForAnonymizedExport.Contains(paramName);
             }
--- a/src/Microsoft.Health.Fhir.Shared.Api/Controllers/ExportController.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Api/Controllers/ExportController.cs
@@ -111,9 +111,17 @@ namespace Microsoft.Health.Fhir.Api.Controllers
             [FromQuery(Name = KnownQueryParameterNames.Container)] string containerName,
             [FromQuery(Name = KnownQueryParameterNames.TypeFilter)] string typeFilter,
             [FromQuery(Name = KnownQueryParameterNames.Format)] string formatName,
+            [FromQuery(Name = KnownQueryParameterNames.AnonymizationConfigurationLocation)] string anonymizationConfigLocation,
+            [FromQuery(Name = KnownQueryParameterNames.AnonymizationConfigurationFileEtag)] string anonymizationConfigFileETag,
             string typeParameter)
         {
             CheckIfExportIsEnabled();
+
+            if (!string.IsNullOrWhiteSpace(anonymizationConfigFileETag) || !string.IsNullOrWhiteSpace(anonymizationConfigFileETag))
+            {
+                CheckIfAnonymizedExportIsEnabled();
+                CheckContainerNameForAnonymizedExport(containerName);
+            }
 
             // Export by ResourceType is supported only for Patient resource type.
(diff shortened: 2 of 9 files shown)


src/Microsoft.Health.Fhir.Api/Features/Filters/ValidateExportRequestFilterAttribute.cs
```diff
@@ -111,7 +120,10 @@ private bool IsValidBasicExportRequestParam(string paramName)
 
         private bool IsValidAnonymizedExportRequestParam(HttpRequest request, string paramName)
         {
-            if (request.Path.StartsWithSegments(DefaultExportRequestPath, StringComparison.InvariantCultureIgnoreCase))
+            if (AnonymizedExportRequestPathRegexs.Any(pathRegex => Regex.IsMatch(
+                input: request.Path,
+                pattern: pathRegex,
+                options: RegexOptions.IgnoreCase | RegexOptions.CultureInvariant)))
```
LTA-Thinking, line 126: This check isn't needed anymore since anonymized export is now available for all the types of export. The route filter will have removed any that don't fit this pattern.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/AnonymizedExportTests.cs
```diff
@@ -55,8 +59,10 @@ public AnonymizedExportTests(ExportTestFixture fixture)
             _exportConfiguration = ((IOptions<ExportJobConfiguration>)(fixture.TestFhirServer as InProcTestFhirServer)?.Server?.Services?.GetService(typeof(IOptions<ExportJobConfiguration>)))?.Value;
         }
 
-        [Fact]
-        public async Task GivenAValidConfigurationWithETag_WhenExportingAnonymizedData_ResourceShouldBeAnonymized()
+        [Theory]
```
LTA-Thinking, line 62: Can we change this to be a SkippableTheory and use the Skip.If call instead of just returning? It makes it easier to see when the test is actually running. There are some examples in CustomSearchParamTests.

src/Microsoft.Health.Fhir.Api/Features/Filters/ValidateExportRequestFilterAttribute.cs
```diff
@@ -111,7 +120,10 @@ private bool IsValidBasicExportRequestParam(string paramName)
 
         private bool IsValidAnonymizedExportRequestParam(HttpRequest request, string paramName)
         {
-            if (request.Path.StartsWithSegments(DefaultExportRequestPath, StringComparison.InvariantCultureIgnoreCase))
+            if (AnonymizedExportRequestPathRegexs.Any(pathRegex => Regex.IsMatch(
```
Ivanidzo4ka, line 123: >AnonymizedExportRequestPathRegexs [](start = 16, length = 33)

Can we not use regex here?
if you pass `ActionExecutingContext context` here, you can inspect context.ActionArguments.Parameters for a presence of expected parameter for action.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ExportController.cs
```diff
@@ -134,18 +142,26 @@ public class ExportController : Controller
...
+            }
+
             // Export by ResourceTypeId is supported only for Group resource type.
             if (!string.Equals(typeParameter, ResourceType.Group.ToString(), StringComparison.Ordinal) || string.IsNullOrEmpty(idParameter))
             {
                 throw new RequestNotValidException(string.Format(Resources.UnsupportedResourceType, typeParameter));
             }
 
-            return await SendExportRequest(ExportJobType.Group, since, typeFilter, resourceType, idParameter, containerName: containerName, formatName: formatName);
+            return await SendExportRequest(ExportJobType.Group, since, typeFilter, resourceType, idParameter, containerName: containerName, formatName: formatName, anonymizationConfigLocation: anonymizationConfigLocation, anonymizationConfigFileETag: anonymizationConfigFileETag);
```
Ivanidzo4ka, line 164: >SendExportRequest [](start = 25, length = 17)

nit: can we break this line? It's bit too big for my taste and average developer monitor

Ivanidzo4ka:
>             return !string.IsNullOrEmpty(_exportJobRecord.AnonymizationConfigurationLocation);

I'm suprised we don't check `AnonymizationConfigFileETag`

---
Refers to: src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs:664 in cc63cc2. [](commit_id = cc63cc20944c881f98c275ddc4d795380870b8d6, deletion_comment = False)

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ExportController.cs
```diff
@@ -111,17 +111,25 @@ public class ExportController : Controller
             [FromQuery(Name = KnownQueryParameterNames.Container)] string containerName,
             [FromQuery(Name = KnownQueryParameterNames.TypeFilter)] string typeFilter,
             [FromQuery(Name = KnownQueryParameterNames.Format)] string formatName,
+            [FromQuery(Name = KnownQueryParameterNames.AnonymizationConfigurationLocation)] string anonymizationConfigLocation,
+            [FromQuery(Name = KnownQueryParameterNames.AnonymizationConfigurationFileEtag)] string anonymizationConfigFileETag,
             string typeParameter)
         {
             CheckIfExportIsEnabled();
 
+            if (!string.IsNullOrWhiteSpace(anonymizationConfigFileETag) || !string.IsNullOrWhiteSpace(anonymizationConfigFileETag))
```
Ivanidzo4ka, line 120: >if (!string.IsNullOrWhiteSpace(anonymizationConfigFileETag) || !string.IsNullOrWhiteSpace(anonymizationConfigFileETag)) [](start = 12, length = 119)

Nit: I see three instances of this code in file. Can we make function like one above?

src/Microsoft.Health.Fhir.Api/Features/Filters/ValidateExportRequestFilterAttribute.cs
QuanWanxx: Got it, removed this check.

Ivanidzo4ka:
>     public class ExportJobTask : IExportJobTask

Design question.
Why we have to pass so many parameters to private functions in this class?

If I correctly understand this object represent specific state of export job.
Why can't we have state of the job inside object instead of passing it around through functions?
For example `progress`, why it's not just a private field? Or `anonymizer`? Or cancellation token. Considering what they are same through whole export job.

---
Refers to: src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs:33 in cc63cc2. [](commit_id = cc63cc20944c881f98c275ddc4d795380870b8d6, deletion_comment = False)

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/AnonymizedExportTests.cs
QuanWanxx: Updated anonymized export tests as skippable and use Skip.If to mark if tests is not actually running.

src/Microsoft.Health.Fhir.Api/Features/Filters/ValidateExportRequestFilterAttribute.cs
QuanWanxx: Thanks for reminding, now not need this regex to do the anonymized export params check anymore.

Ivanidzo4ka:
>             // Validate that the request does not contain query parameters that are not supported.

It's a bit tricky to support action filter and controller filter in sync, so can I suggest to switch to this approach instead of hardcoding "expected" parameters?

[https://stackoverflow.com/questions/55451408/reject-requests-containing-unknown-query-parameters](https://stackoverflow.com/questions/55451408/reject-requests-containing-unknown-query-parameters)


---
Refers to: src/Microsoft.Health.Fhir.Api/Features/Filters/ValidateExportRequestFilterAttribute.cs:72 in 300e9ab. [](commit_id = 300e9abf5bc2b7fbfb4dccfbe0b744944b6b761e, deletion_comment = False)

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ExportController.cs
```diff
@@ -134,18 +142,26 @@ public class ExportController : Controller
...
+            {
+                CheckIfAnonymizedExportIsEnabled();
+                CheckContainerNameForAnonymizedExport(containerName);
+            }
+
             // Export by ResourceTypeId is supported only for Group resource type.
             if (!string.Equals(typeParameter, ResourceType.Group.ToString(), StringComparison.Ordinal) || string.IsNullOrEmpty(idParameter))
```
tongwu-msft, line 159: Test

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ExportController.cs
QuanWanxx: Now separated the parameters into multiple lines.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ExportController.cs
QuanWanxx: Extract a unified ValidateForAnonymizedExport function. 
And I add a check to ensure that when only _anonymizationConfigEtag is given, will return bad request with "Missing '_anonymizationConfig' for anonymized export"

QuanWanxx:
> > ```
> > public class ExportJobTask : IExportJobTask
> > ```
> 
> Design question.
> Why we have to pass so many parameters to private functions in this class?
> 
> If I correctly understand this object represent specific state of export job.
> Why can't we have state of the job inside object instead of passing it around through functions?
> For example `progress`, why it's not just a private field? Or `anonymizer`? Or cancellation token. Considering what they are same through whole export job.
> 
> Refers to: src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs:33 in [cc63cc2](https://github.com/microsoft/fhir-server/commit/cc63cc20944c881f98c275ddc4d795380870b8d6). [](commit_id = [cc63cc2](https://github.com/microsoft/fhir-server/commit/cc63cc20944c881f98c275ddc4d795380870b8d6), deletion_comment = False)

Thanks for advice. In fact there's only one ExportJobTask instance, and each time a new export request received, [ExportJobWorker](https://github.com/microsoft/fhir-server/blob/personal/quwan/Feature_anonymized_Patient_Group_export/src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobWorker.cs) will create a ExecuteAsync task to do the export job. So I feel like maybe those related instances (anonymizer, progress and cancellation token) better not be the private fields. 
I thought a potential optimization here is to create ExportJobTask for each export job, but that can be another refine work.

QuanWanxx:
> > ```
> >         // Validate that the request does not contain query parameters that are not supported.
> > ```
> 
> It's a bit tricky to support action filter and controller action in sync, so can I suggest to switch to this approach instead of hardcoding "expected" parameters?
> 
> https://stackoverflow.com/questions/55451408/reject-requests-containing-unknown-query-parameters
> 
> Refers to: src/Microsoft.Health.Fhir.Api/Features/Filters/ValidateExportRequestFilterAttribute.cs:72 in [300e9ab](https://github.com/microsoft/fhir-server/commit/300e9abf5bc2b7fbfb4dccfbe0b744944b6b761e). [](commit_id = [300e9ab](https://github.com/microsoft/fhir-server/commit/300e9abf5bc2b7fbfb4dccfbe0b744944b6b761e), deletion_comment = False)

That's sounds good, I think this approach would be more flexible. 
Seems like this part is written by @namadabu , do you have any concerns if I update the supported parameters in this way?
From my experiment, the conditions will be like 

`
IList<ParameterDescriptor> parameters = context.ActionDescriptor.Parameters == null ?
                                                                   new List<ParameterDescriptor>() :
                                                                   context.ActionDescriptor.Parameters;
..
if (parameters.Any(parm => string.Equals(parm.BindingInfo?.BinderModelName, paramName, StringComparison.Ordinal)))
{
 continue;
 }
`

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ExportController.cs
```diff
@@ -221,6 +247,15 @@ private void CheckIfExportIsEnabled()
...
+        private void ValidateForAnonymizedExport(string containerName, string anonymizationConfigLocation, string anonymizationConfigFileETag)
+        {
+            if (!string.IsNullOrWhiteSpace(anonymizationConfigLocation) || !string.IsNullOrWhiteSpace(anonymizationConfigFileETag))
+            {
+                CheckIfAnonymizedExportIsEnabled();
+                CheckContainerNameAndConfigLocationForAnonymizedExport(containerName, anonymizationConfigLocation);
+            }
+        }
+
         private void CheckIfAnonymizedExportIsEnabled()
```
Ivanidzo4ka, line 259: >CheckIfAnonymizedExportIsEnabled [](start = 21, length = 32)

nit: you can probably put this logic into `ValidateForAnonymizedExpor` and have less functions in a class

Ivanidzo4ka:
I was under impression what ExportJobTask got created via DI, and it's registered as transient object so it would be new for each incoming request.
It's perfectly fine to have current code, it was more question towards other people in codereview (which I forgot to mention, so sorry for my sloppiness).
I still think it would be nice to refactor code a bit, but it's too much to ask from current PR, it serves different goal. Sorry for distracting you.

---
In reply to: [781932463](https://github.com/microsoft/fhir-server/pull/1672#issuecomment-781932463) [](ancestors = 781932463)

Ivanidzo4ka (approved):
:shipit:

LTA-Thinking (approved):

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ExportController.cs
```diff
@@ -229,8 +264,13 @@ private void CheckIfAnonymizedExportIsEnabled()
             }
         }
 
-        private static void CheckContainerNameForAnonymizedExport(string containerName)
+        private static void CheckContainerNameAndConfigLocationForAnonymizedExport(string containerName, string anonymizationConfigLocation)
         {
+            if (string.IsNullOrWhiteSpace(anonymizationConfigLocation))
```
namadabu, line 269: nit: Would it be helpful to add UTs for this part of the code? Or are they covered by E2E tests?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/AnonymizedExportTests.cs
```diff
@@ -168,13 +220,31 @@ public async Task GivenInvalidConfiguration_WhenExportingAnonymizedData_ThenBadR
             Assert.Single(_metricHandler.NotificationMapping[typeof(ExportTaskMetricsNotification)]);
         }
 
-        [Fact]
+        [SkippableFact]
+        public async Task GivenAGroupIdNotExisted_WhenExportingGroupAnonymizedData_ThenBadRequestShouldBeReturned()
+        {
+            Skip.If(!_isUsingInProcTestServer, "Not using in-process fhir server.");
```
namadabu, line 226: nit: Curious to know if there is any specific reason we don't want to execute these tests against an actual fhir-server instance?

Ivanidzo4ka: I think because it's hard to turn on Anonymized Export. Right now we do it in fixture, with interface substitution, but you can't do that for real server.
Plus MetricHandler, there is no way to easy way to check metric handler execution on server side (it's easy if your server is in same process, you share memory, but when server is hosted somewhere in azure, it's a bit hard).

---
In reply to: [580684676](https://github.com/microsoft/fhir-server/pull/1672#discussion_r580684676) [](ancestors = 580684676)

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ExportController.cs
QuanWanxx: Add unit tests for there 2 Exception in ExportControllerTests.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ExportController.cs
QuanWanxx: Sure, moved this into the ValidateForAnonymizedExport.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/AnonymizedExportTests.cs
namadabu: If we want to enable anonymizedExport only for our test servers, we should be able to update our deployment steps to provision them with the config enabled. If I remember right, we do something similar for cors settings by having a corstestconfiguration.json file that we read from and deploy to our servers. 

Agreed with the MetricHandler point. But I feel that we don't need to test for the MetricHandler notification in every test. We just need one or two tests to make sure it works as expected.

---
In reply to: [580687236](https://github.com/microsoft/fhir-server/pull/1672#discussion_r580687236) [](ancestors = 580687236,580684676)

tongwu-msft (approved):

namadabu (approved):

PR #1676, opened 2021-02-18
Added logs for cancel export request flow

namadabu:
## Description
Added some logging in the CancelExportRequest handler.

## Related issues
Addresses [issue [AB#79190](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/79190)].

## Testing
No functionality change. Existing tests.

## FHIR Team Checklist

Patch

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Operations/Export/CancelExportRequestHandler.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Operations/Export/CancelExportRequestHandler.cs
@@ -88,7 +88,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Export
             }
             catch (Exception ex)
             {
-                _logger.LogWarning($"Unable to cancel export job {request.JobId}. Exception: {ex}");
+                _logger.LogError(ex, $"Unable to cancel export job {request.JobId}");
                 throw;
             }
 
--- a/src/Microsoft.Health.Fhir.Core/Features/Operations/Export/CancelExportRequestHandler.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Operations/Export/CancelExportRequestHandler.cs
@@ -9,6 +9,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using EnsureThat;
 using MediatR;
+using Microsoft.Extensions.Logging;
 using Microsoft.Health.Core;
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Fhir.Core/Features/Operations/Export/CancelExportRequestHandler.cs
```diff
@@ -56,26 +61,38 @@ public async Task<CancelExportResponse> Handle(CancelExportRequest request, Canc
...
+                    _logger.LogInformation($"Attempting to cancel export job {request.JobId}");
+                    await _fhirOperationDataStore.UpdateExportJobAsync(outcome.JobRecord, outcome.ETag, cancellationToken);
 
-                await _fhirOperationDataStore.UpdateExportJobAsync(outcome.JobRecord, outcome.ETag, cancellationToken);
+                    return new CancelExportResponse(HttpStatusCode.Accepted);
+                });
+            }
+            catch (Exception ex)
+            {
+                _logger.LogWarning($"Unable to cancel export job {request.JobId}. Exception: {ex}");
```
Ivanidzo4ka, line 91: >LogWarning [](start = 24, length = 10)

That would only give you last exception in retry cycle( at least if I correctly understand how retry policy works)
There is chance you run retry few times and it fail with actual problem, but last time you would get just 429 because you send to much requests to DB.

Also LogError can take your exception and maybe format it in better way. Plus exception feels like error thingy than just a warning.

namadabu: In this scenario, retry only kicks in if a `JobConflictException` is thrown. So it should be enough if we log the exception that gets us out of the retry loop. Also, we'll know if we are retrying because I added a log for each attempt. 

I'll update the code to use `LogError` instead of `LogWarning`.

Ivanidzo4ka: Well, you call `GetExportJobByIdAsync` and `UpdateExportJobAsync` and from my perspective both of them can throw something, but I'm not expert in export jobs workflow :)

Ivanidzo4ka (approved):
:shipit:

Issue #1674, opened 2021-02-18
Fix Nuget package metadata


brendankowitz:
**Describe the bug**
When inspecting our current packages with Nuget Package Explorer, some metadata is not being set properly.

![image](https://user-images.githubusercontent.com/197221/108406258-045c5780-71d7-11eb-8ad4-f5e6edc74881.png)

**To Reproduce**
1. Open a FHIR package with Nuget Package Explorer

**Expected behavior**
Metadata should be valid, this was recently reviewed/fixed in https://github.com/microsoft/healthcare-shared-components/blob/master/Directory.Build.props
![image](https://user-images.githubusercontent.com/197221/108406461-4eddd400-71d7-11eb-9f76-e65c17487da2.png)

**Actual behavior**
Metadata is missing or not linked properly

Issue #262, opened 2018-12-07
error MSB4018: The "Microsoft.Build.Tasks.Git.LocateRepository" task failed unexpectedly.


mattkoch614:
**Describe the bug**
After cloning this project locally, when building the ``Microsoft.Health.Fhir.Web`` project on a Mac using Visual Studio 2019 preview, I receive the following error:

``error MSB4018: The "Microsoft.Build.Tasks.Git.LocateRepository" task failed unexpectedly.``

**To Reproduce**
1. Clone Repo
2. Build
3. Error occurs

**Expected behavior**
No build errors occur.

**Actual behavior**
A build error occurs.

I'm not exactly sure what this error refers to.

brendankowitz:
This is caused by the `Microsoft.SourceLink.GitHub` package, there are references to an open issue around: https://github.com/dotnet/sourcelink/issues/155

PR #1678, opened 2021-02-18
Enable source link in nuget packages

Ivanidzo4ka:
## Description
It was disabled in #262, but instead of been disabled only for OSx it was never set to be true at first place, so it was off for all platforms.
Regarding OSx it looks like bug was fixed somewhere around 2019 (https://github.com/dotnet/sourcelink/pull/288)

## Related issues
Addresses [#1674 ].

## Testing
in progress
image

## FHIR Team Checklist

Patch (to regenerate nuget with right information)

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -12,20 +12,15 @@
     <DebugType>Portable</DebugType>
     <LangVersion>9.0</LangVersion>
     <HighEntropyVA>true</HighEntropyVA>
-    <EnableSourceLink Condition="$([MSBuild]::IsOSPlatform('osx'))">false</EnableSourceLink>
-    <EnableSourceControlManagerQueries>$(EnableSourceLink)</EnableSourceControlManagerQueries>
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
+    <EmbedUntrackedSources>true</EmbedUntrackedSources>
+    <AllowedOutputExtensionsInPackageBuildOutputFolder>$(AllowedOutputExtensionsInPackageBuildOutputFolder);.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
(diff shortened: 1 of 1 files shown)


Directory.Build.props
```diff
@@ -12,8 +12,13 @@
     <DebugType>Portable</DebugType>
     <LangVersion>9.0</LangVersion>
     <HighEntropyVA>true</HighEntropyVA>
-    <EnableSourceLink Condition="$([MSBuild]::IsOSPlatform('osx'))">false</EnableSourceLink>
-    <EnableSourceControlManagerQueries>$(EnableSourceLink)</EnableSourceControlManagerQueries>
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
```
brendankowitz, line 15: This is also above

Ivanidzo4ka: indeed...

Ivanidzo4ka: One part of me want's to sort them, another part want's to break them into different property groups (not sure I can do clear cut by nuget related/ vs everything else), but there also third part which is just lazy, and whispering into my ear: "No one cares". If you think any of approaches worth it, I'll do it, I just not sure is it really worth it or not.

Directory.Build.props
```diff
@@ -19,7 +19,6 @@
     <Company>Microsoft Corporation</Company>
     <Copyright>Copyright © Microsoft Corporation. All rights reserved.</Copyright>
     <Deterministic>true</Deterministic>
-    <GenerateDocumentationFile>true</GenerateDocumentationFile>
```
brendankowitz: Can we create a bug to fix the docs generation if this doesn't work?

Ivanidzo4ka: https://github.com/microsoft/fhir-server/issues/1681

brendankowitz (approved):

PR #1682, opened 2021-02-19
Reindexes a single resource

rotodd:
## Description

Prior to this, we were relying on the `UpsertResource` stored procedure to update a resource's search indices in SQL. This created a new version of every resource that was reindexed.

This PR creates a new stored proc for reindexing that:

1. Updates the search parameter hash value in the dbo.Resource table
2. Deletes all of the resource's entries in the search parameter index tables
3. Inserts new entries into the search parameter index tables

Note: we could improve the performance of this by only updating the search indices that are out of date. [AB#79648](https://microsofthealth.visualstudio.com/Health/_workitems/edit/79648) is being used to track this.

## Related issues

Addresses: [AB#79121](https://microsofthealth.visualstudio.com/Health/_workitems/edit/79121)
Coming up next: [AB#73395](https://microsofthealth.visualstudio.com/Health/_workitems/edit/73395)

## Testing

Adds a check to ensure that the version of resource does not change in existing reindex tests.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Persistence/IFhirDataStore.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Persistence/IFhirDataStore.cs
@@ -22,8 +22,6 @@ namespace Microsoft.Health.Fhir.Core.Features.Persistence
 
         Task HardDeleteAsync(ResourceKey key, CancellationToken cancellationToken);
 
-        Task UpdateSearchParameterHashBatchAsync(IReadOnlyCollection<ResourceWrapper> resources, CancellationToken cancellationToken);
-
         Task UpdateSearchParameterIndicesBatchAsync(IReadOnlyCollection<ResourceWrapper> resources, CancellationToken cancellationToken);
 
         Task<ResourceWrapper> UpdateSearchIndexForResourceAsync(ResourceWrapper resourceWrapper, WeakETag weakETag, CancellationToken cancellationToken);
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/8.diff.sql
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/8.diff.sql
@@ -27,6 +27,8 @@ GO
 --         * The ID of the resource type (See ResourceType table)
 --     @resourceId
 --         * The resource ID (must be the same as the in the resource itself)
+--     @etag
+--         * If specified, the version of the resource to update
 --     @allowCreate
 --         * If false, an error is thrown if the resource does not already exist
 --     @isDeleted
@@ -703,6 +705,8 @@ GO
 --         * The ID of the resource type (See ResourceType table)
 --     @resourceId
 --         * The resource ID (must be the same as the in the resource itself)
+--     @etag
+--         * If specified, the version of the resource to update
 --     @searchParamHash
 --          * A hash of the resource's latest indexed search parameters
 --     @resourceWriteClaims
@@ -741,6 +745,7 @@ GO
 CREATE PROCEDURE dbo.ReindexResource
     @resourceTypeId smallint,
     @resourceId varchar(64),
+    @eTag int = NULL,
     @searchParamHash varchar(64),
     @resourceWriteClaims dbo.ResourceWriteClaimTableType_1 READONLY,
     @compartmentAssignments dbo.CompartmentAssignmentTableType_1 READONLY,
@@ -765,16 +770,21 @@ AS
     BEGIN TRANSACTION
 
     DECLARE @resourceSurrogateId bigint
+    DECLARE @version bigint
 
     -- This should place a range lock on a row in the IX_Resource_ResourceTypeId_ResourceId nonclustered filtered index
     -- TODO: Should we be locking all index tables too?
-    SELECT @resourceSurrogateId = ResourceSurrogateId
+    SELECT @resourceSurrogateId = ResourceSurrogateId, @version = Version
     FROM dbo.Resource WITH (UPDLOCK, HOLDLOCK)
     WHERE ResourceTypeId = @resourceTypeId AND ResourceId = @resourceId AND IsHistory = 0
 
+    IF (@etag IS NOT NULL AND @etag <> @version) BEGIN
+        THROW 50412, 'Precondition failed', 1;
+    END
+
     IF (@resourceSurrogateId IS NULL) BEGIN
         -- You can't reindex a resource if the resource does not exist
-        THROW 50404, 'Resource with not found', 1;
+        THROW 50404, 'Resource not found', 1;
     END
 
     UPDATE dbo.Resource
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/8.diff.sql
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/8.diff.sql
@@ -31,8 +31,6 @@ GO
 --         * If false, an error is thrown if the resource does not already exist
 --     @isDeleted
 --         * Whether this resource marks the resource as deleted
---     @updatedDateTime
---         * The last modified time in the resource
 --     @keepHistory
 --         * Whether the existing version of the resource should be preserved
 --     @requestMethod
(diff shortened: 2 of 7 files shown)


src/Microsoft.Health.Fhir.Core/Features/Persistence/IFhirDataStore.cs
```diff
@@ -22,8 +22,6 @@ public interface IFhirDataStore
 
         Task HardDeleteAsync(ResourceKey key, CancellationToken cancellationToken);
 
-        Task UpdateSearchParameterHashBatchAsync(IReadOnlyCollection<ResourceWrapper> resources, CancellationToken cancellationToken);
-
```
rotodd, line 26: @feordin I removed this, since it has no usages and it is easier to update the search parameter hash at the same time as the indices - is there anything I'm missing/any reason to keep it?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/8.diff.sql
```diff
@@ -692,3 +692,234 @@ AS
...
+    SET NOCOUNT ON
+
+    SET XACT_ABORT ON
+    BEGIN TRANSACTION
+
+    DECLARE @resourceSurrogateId bigint
+    DECLARE @version bigint
+
+    -- This should place a range lock on a row in the IX_Resource_ResourceTypeId_ResourceId nonclustered filtered index
+    -- TODO: Should we be locking all index tables too?
```
rotodd, line 776: @johnstairs this lock is the same one that we are using in the `UpsertResource` stored proc. 

To confirm my understanding, this is only locking certain access to the `dbo.Resource` table, right? If so, is it alright that we are not blocking read/write access to all the index tables during an upsert or reindex operation?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/8.diff.sql
```diff
@@ -692,3 +692,234 @@ AS
...
+    SELECT @resourceSurrogateId = ResourceSurrogateId, @version = Version
+    FROM dbo.Resource WITH (UPDLOCK, HOLDLOCK)
+    WHERE ResourceTypeId = @resourceTypeId AND ResourceId = @resourceId AND IsHistory = 0
+
+    IF (@etag IS NOT NULL AND @etag <> @version) BEGIN
+        THROW 50412, 'Precondition failed', 1;
+    END
```
rotodd, line 783: @feordin I am passing the etag into the `ReindexResource` stored proc to perform an additional version check here - but it doesn't seem all that necessary. Is there a reason why this is needed on the Cosmos side of things?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/8.diff.sql
```diff
@@ -692,3 +692,234 @@ AS
...
+        -- You can't reindex a resource if the resource does not exist
+        THROW 50404, 'Resource not found', 1;
+    END
+
+    UPDATE dbo.Resource
+    SET SearchParamHash = @searchParamHash
+    WHERE ResourceSurrogateId = @resourceSurrogateId
+
+    -- First, delete all the resource's indices.
+    DELETE FROM dbo.ResourceWriteClaim
```
johnstairs, line 795: Did you consider the `MERGE` statement instead of `DELETE` followed by `INSERT`? In most reindexing cases, most of the existing indexes would end up being the same. A merge _might_ be more efficient here.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -319,9 +313,46 @@ public void Build(ICapabilityStatementBuilder builder)
             }
         }
 
-        public Task<ResourceWrapper> UpdateSearchIndexForResourceAsync(ResourceWrapper resourceWrapper, WeakETag weakETag, CancellationToken cancellationToken)
+        public async Task<ResourceWrapper> UpdateSearchIndexForResourceAsync(ResourceWrapper resource, WeakETag weakETag, CancellationToken cancellationToken)
         {
-            throw new NotImplementedException();
+            int? eTag = weakETag == null
+                ? (int?)null
+                : (int.TryParse(weakETag.VersionId, out var parsedETag) ? parsedETag : -1); // Set the etag to a sentinel value to enable expected failure paths when updating with both existing and nonexistent resources.
```
rotodd, line 320: Remove this if the version check isn't needed [here](https://github.com/microsoft/fhir-server/pull/1682/files#r579401925).

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/8.diff.sql
johnstairs: All reads and writes that you do within a transaction will place different kinds of locks.

Since we are using RCSI, reads will not be blocked by this operation. For writes, concurrent calls to this sproc and  UpsertResource would block each other if accessing the same resource, but that should be by design. Updating separate resources should be fine.

src/Microsoft.Health.Fhir.Core/Features/Persistence/IFhirDataStore.cs
feordin: We will only end up using this is if we implement the performance enhancement which compares preivous indices to the newly extracted indices, and then determine that the only update required is the hash value.  We can add this method back in if we ever implement that.  So it is fine to remove it.

rotodd: Ah, I see - thank you! I'll make a note about it in [AB#79648](https://microsofthealth.visualstudio.com/Health/_workitems/edit/79648).

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/8.diff.sql
rotodd: Good suggestion - looking into this now. Thanks!

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/8.diff.sql
feordin: I believe it is here in case the resource is changed between the time it was initially queried by the reindexjob, and the time the process finishes extracting the new search indices and then calls this stored procedure to update the values.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/8.diff.sql
rotodd: As discussed, I will consider this when writing the stored proc for batch updates in [AB#73395](https://microsofthealth.visualstudio.com/Health/_workitems/edit/73395), since that is where we will benefit most from performance gains.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/8.diff.sql
rotodd: As discussed, it would be good if the reindex job worker could pick up these resources and attempt to reindex them again instead of returning an error. Created [AB#79925](https://microsofthealth.visualstudio.com/Health/_workitems/edit/79925) and added some missing exception handling in 9122322 in the meantime.

feordin (approved):

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/8.diff.sql
rotodd: Thanks for the reply!

I felt like I still needed a refresher on this, so I read through [this doc](https://docs.microsoft.com/en-us/dotnet/framework/data/adonet/sql/snapshot-isolation-in-sql-server). Jotting down my takeaways here for my future self:

- Read Committed Snapshot Isolation (RCSI) ensures that data read within a transaction will never reflect changes made by other simultaneous transactions
- With RSCI, the data row versions are recorded at the beginning of a transaction, and if a transaction attempts to commit an update to a row that was changed after the transaction began, the transaction is rolled back, and an error is raised
- Lock hints like `UPDLOCK` can reduce the chance of update conflicts in long running transactions by blocking other transactions from attempting to update a specific set of rows before the transaction completes

The stored proc above begins a new transaction which has RSCI enabled. The lock is used to block other transactions from updating the row in `dbo.Resource` for the resource being reindexed. In theory, the data in the index tables relevant to this resource could be modified over the course of the transaction. If that were the case, the transaction would fail to commit, and an error would be returned.

@johnstairs please let me know if I have misunderstood something here
@feordin it sounds like this would be another similar scenario where we would like to reattempt to reindex a resource

Issue #838, opened 2020-01-17
R4 should support scientific notation in decimal


brandonpollett:
**Describe the bug**
Per the [R4 Spec](https://www.hl7.org/fhir/datatypes.html#decimal), the decimal has a regex that allows scientific notation `-?(0|[1-9][0-9]*)(\.[0-9]+)?([eE][+-]?[0-9]+)?`. Additionally, some [search examples](https://www.hl7.org/fhir/search.html#quantity) make use of this functionality. When you send a search such as `GET [base]/Observation?value-quantity=5.40e-3|http://unitsofmeasure.org|g` the server responds with an operation outcome with the message `"Malformed search value '5.40e-3'."`

**To Reproduce**
1. Send search for `Observation?value-quantity=5.40e-3|http://unitsofmeasure.org|g`

**Expected behavior**
Search completes and the self link includes the search parameters

**Actual behavior**
Operation outcome is returned with the message `"Malformed search value '5.40e-3'."`

albamati:
I think this issue can be closed now

CaitlinV39:
Thank you so much @albamati for picking this one up!

PR #1683, opened 2021-02-19
Support scientific notation for Quantity in Search

albamati:
## Description
When converting a quantity in the search parameter (from string to decimal), now also the scientific notation can be used.

## Related issues
Addresses [#838]

## Testing
- Tested manually
- Added an unit test

## FHIR Team Checklist


src/Microsoft.Health.Fhir.Core.UnitTests/Features/Search/SearchValues/QuantitySearchValueTests.cs
```diff
@@ -81,6 +81,7 @@ public void GivenAStringContainingMoreThanTwoTokenSeparators_WhenParsing_ThenExc
         [InlineData(@"234.5|system|", "system", "", 234.5)]
         [InlineData(@"0|system|code", "system", "code", 0)]
         [InlineData(@"98.565656||code", "", "code", 98.565656)]
+        [InlineData(@"5.40e-3|system|code", "system", "code", 0.00540)]
```
LTA-Thinking, line 84: Could you also add a test case to the QuanitySearchTests/QuantitySearchTestFixture files.

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/QuantitySearchValue.cs
```diff
@@ -89,7 +89,7 @@ public static QuantitySearchValue Parse(string s)
             }
 
             decimal quantity;
-            if (!decimal.TryParse(parts[0], NumberStyles.Number, CultureInfo.InvariantCulture, out quantity))
+            if (!decimal.TryParse(parts[0], NumberStyles.Number | NumberStyles.AllowExponent, CultureInfo.InvariantCulture, out quantity))
```
Ivanidzo4ka, line 92: https://www.hl7.org/fhir/STU3/datatypes.html#decimal
>decimals may not use exponents, and leading 0 digits are not allowed

Technically STU3 shouldn't support that notation, but in same time I'm not sure how strict we should be about that.
Any thoughts? @brendankowitz

albamati: Good point! 
When I was testing my changes with Postman, I noticed that the POST on Observation is working also with the exponential format - I tried with both R4 and STU3. So if the final decision will be not to accept it in the GET for the STU3... well IMHO then probably it should not be accepted in the POST.

chgl:
This may be slightly out of scope, but what happens if the number doesn't fit inside a decimal, let's say `6.5742407583884805e+84` (https://github.com/FirelyTeam/firely-net-sdk/discussions/1635)? Can such numbers be persisted to the FHIR server in the first place?

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/QuantitySearchValue.cs
brendankowitz: It might be fine to allow this, since the community serializes already allow this. The behavior is additive rather than altering an expected outcome.
//cc @CaitlinV39

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Search/SearchValues/QuantitySearchValueTests.cs
albamati: Done!

CaitlinV39:
@albamati just a heads up that we require signing our contributor license agreement to be able to commit this PR. Can you review that (above in the comments)? Thanks!

albamati:
> This may be slightly out of scope, but what happens if the number doesn't fit inside a decimal, let's say `6.5742407583884805e+84` ([FirelyTeam/firely-net-sdk#1635](https://github.com/FirelyTeam/firely-net-sdk/discussions/1635))? Can such numbers be persisted to the FHIR server in the first place?

IMHO the code in this PR should work in all those cases when the POST for the resource is working.
I tried to POST an observation with `6.5742407583884805e+84`, I got a Bad Request:
2021-03-05_14h21_07
... so I think it's ok if the search Observation for that quantity returns also an error:
2021-03-05_14h17_56

In my opinion introducing support for big numbers like `6.5742407583884805e+84` probably requires a bigger effort. It's look like a "task" more than a "bug fix"  - Probably it is out of scope trying to fix it in this PR

brendankowitz (approved):

LTA-Thinking (approved):

Issue #1690, opened 2021-02-23
Date queries are not handled efficiently with Cosmos DB provider


johnstairs:
**Describe the bug**

**FHIR Version?**
Stu3/R4/R5

**Data provider?**
CosmosDB/SQL Server

**To Reproduce**
1. 
2. 
3. 

**Expected behavior**

**Actual behavior**
A clear and concise description of what actually happened.

PR #1691, opened 2021-02-23
Improve performance of DateTime searches in Cosmos DB

johnstairs:
## Description
Searches like `/Patient?birthdate=1948-11-22` currently perform poorly with the Cosmos DB provider. In this change, we are making the following change to the query we generate this this kind of search:
image

Today, the database is forced to intersect all patients where the birthdate's start is > queryStart and birthdate's end is < queryEnd, but because there is no correlation between these fields, this results in all patients being processed to perform the intersection. 

The solution is to add an upper bound to start, the same as the end time. 

Since we already do this for the SQL implementation, we are reusing the same visitor.

This makes a major difference. For a database with 44K Synthea patients, here are some before and after numbers, searching for the first ~10 results:

|                                          | Before RUs                                               | After RUs |
|------------------------------------------|----------------------------------------------------------|-----------|
| Observation?date=2011-08-17              | 143283                                                   | 18        |
| /Patient?birthdate=1948-11-22            | 1911                                                     | 15        |
| Observation?date=2007-08-16 (no matches) | Many empty pages with continuation and 5-10K RU per page | 7.85      |

## Related issues
Addresses #1690 

## Testing
Added unit tests and manual testing against a large-ish database

Patch

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/DateTimeEqualityRewriter.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/DateTimeEqualityRewriter.cs
@@ -4,10 +4,10 @@
 // -------------------------------------------------------------------------------------------------
 
 using System.Collections.Generic;
-using Microsoft.Health.Fhir.Core.Features.Search.Expressions;
+using System.Linq;
 using Microsoft.Health.Fhir.ValueSets;
 
-namespace Microsoft.Health.Fhir.SqlServer.Features.Search.Expressions.Visitors
+namespace Microsoft.Health.Fhir.Core.Features.Search.Expressions
 {
     /// <summary>
     /// Rewrites (And (FieldGreaterThanOrEqual DateTimeStart x) (FieldLessThanOrEqual DateTimeEnd y)) to
@@ -17,14 +17,12 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Search.Expressions.Visitors
     /// </summary>
     internal class DateTimeEqualityRewriter : ExpressionRewriterWithInitialContext<object>
     {
-        internal static readonly DateTimeEqualityRewriter Instance = new DateTimeEqualityRewriter();
+        internal static readonly DateTimeEqualityRewriter Instance = new();
 
         public override Expression VisitSearchParameter(SearchParameterExpression expression, object context)
         {
-            // include composites because they may contain dates.
-
             if (expression.Parameter.Type == SearchParamType.Date ||
-                expression.Parameter.Type == SearchParamType.Composite)
+                (expression.Parameter.Type == SearchParamType.Composite && expression.Parameter.ResolvedComponents.Any(c => c.Type == SearchParamType.Date)))
             {
                 return base.VisitSearchParameter(expression, context);
             }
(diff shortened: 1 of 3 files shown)


brendankowitz (approved):

LTA-Thinking:
Is it worth also adding the constraint that end time has to be greater than or equal to parameter 3?

src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/DateTimeEqualityRewriter.cs
```diff
@@ -0,0 +1,93 @@
...
+
+namespace Microsoft.Health.Fhir.Core.Features.Search.Expressions
+{
+    /// <summary>
+    /// Rewrites (And (FieldGreaterThanOrEqual DateTimeStart x) (FieldLessThanOrEqual DateTimeEnd y)) to
+    /// (And (FieldGreaterThanOrEqual DateTimeStart x) (FieldLessThanOrEqual DateTimeEnd y) (FieldLessThanOrEqual DateTimeStart y)).
+    /// It looks specifically for this pattern because it is what we emit in the core layer for an equals search over a date parameter.
+    /// This rewriting constrains the range scan over the index (DateTimeStart, DateTimeEnd).
+    /// </summary>
+    internal class DateTimeEqualityRewriter : ExpressionRewriterWithInitialContext<object>
```
Ivanidzo4ka, line 18: >DateTimeEqualityRewriter [](start = 19, length = 24)

nit: is there are reason to move it out of `Visitors` folder? All other rewriters are remain in that folder. Also it's a bit tricky to see changes in file.

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
```diff
@@ -57,6 +57,12 @@ public class FhirCosmosSearchService : SearchService
             SearchOptions searchOptions,
             CancellationToken cancellationToken)
         {
+            // we're going to mutate searchOptions, so clone it first so the caller of this method does not see the changes.
+            searchOptions = searchOptions.Clone();
```
feordin, line 61: This looks like a shallow clone.  If so any reference properties would be shared between the original object and the clone.  Do we alter the expression object that is passed to the Visitor, or de we generate a new object which is assigned to the clone?

johnstairs: The expressions are all immutable so this should be safe

src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/DateTimeEqualityRewriter.cs
```diff
@@ -0,0 +1,93 @@
...
+                return expression;
+            }
+
+            List<Expression> newExpressions = null;
+            int i = 0;
+            for (; i < expression.Expressions.Count - 1; i++)
+            {
+                switch (MatchPattern(expression.Expressions[i], expression.Expressions[i + 1]))
+                {
+                    case ({ } low, { } high):
```
Ivanidzo4ka, line 47: >{ } low, { } high [](start = 26, length = 17)

I never saw this syntax before. What these curly braces do?

johnstairs: It matches non-null values. Alternative would be `(BinaryExpression low, BinaryExpression high)`. But `(var low, var high)` matches null values which is pretty surprising design decision IMO.

johnstairs:
> Is it worth also adding the constraint that end time has to be greater than or equal to parameter 3?

@LTA-Thinking. unless the data is invalid in the database, a query looking for endtime < starttime would return no results. Without this optimization, it would require a table scan to realize that. With this, it will find that out faster :)

PR #1695, opened 2021-02-24
Limited chained search in Cosmos provider

brendankowitz:
## Description
Adds limited support for Chained and Reverse Chained FHIR Search in CosmosDB.
Adds a feature flag to enable the feature selectively for more feedback.

**Approach:**
In a chained expression, e.g. `Observation?subject:Patient.name='Jim'` will result in a query run against Patient where name = 'Jim', these ids will then be used to filter Observations by search parameter references in the `Observation.subject` index.

In a reverse chained expression, e.g. `Patient?_has:Group:member:_id=group1` this will result in a query against Group where _id=group1, yielding out the search indexes of `Group.member`. These ids will then be used to filter the parent type `Patient`.

**Limitations:**
Due to not being able to join to other documents, this implementation walks down the search expression and issues sub-queries to resolve the matched resources, this is done for each level of the expression. If any query returns more than 100 results an error will be thrown.

**Examples:**

* Simple chained query, observations who have a Patient with name Smith:
  * `/Observation?subject:Patient.name=Smith`
* Reverse chaining example, Practitioners who have Encounters with Claims created on 2020-04-01
  * `/Practitioner?_has:Encounter:practitioner:_has:Claim:encounter:created=eq2020-04-01`
* Chaining with reverse chaining example, Observations for patients who are in a Group with _id=group1
  * `/Observation?subject:Patient._has:Group:member:_id=group1`

**Questions/considerations:**
- What performance should we expect? 
  - The RUs for all sub-queries will be totaled and returned with the response.
- Should we limit the number and levels in chained search? Each level of each chained query will add a sub-query.  
  - This might be ok, if the sub-queries fail or return 0 results, the chain is short-circuited 

## Related issues
Addresses [issue #].

## Testing
Enables existing tests against cosmosdb

image

## FHIR Team Checklist

Patch

### DIFF
--- a/src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
+++ b/src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
@@ -11,6 +11,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using EnsureThat;
 using Microsoft.Azure.Cosmos;
+using Microsoft.Extensions.Primitives;
 using Microsoft.Health.Fhir.Core.Extensions;
 using Microsoft.Health.Fhir.Core.Features.Context;
 using Microsoft.Health.Fhir.Core.Features.Definition;
@@ -18,6 +19,7 @@ using Microsoft.Health.Fhir.Core.Features.Persistence;
 using Microsoft.Health.Fhir.Core.Features.Search;
 using Microsoft.Health.Fhir.Core.Features.Search.Expressions;
 using Microsoft.Health.Fhir.Core.Models;
+using Microsoft.Health.Fhir.CosmosDb.Configs;
 using Microsoft.Health.Fhir.CosmosDb.Features.Search.Queries;
 using Microsoft.Health.Fhir.CosmosDb.Features.Storage;
 using Microsoft.Health.Fhir.ValueSets;
@@ -32,8 +34,10 @@ namespace Microsoft.Health.Fhir.CosmosDb.Features.Search
         private readonly CosmosFhirDataStore _fhirDataStore;
         private readonly IQueryBuilder _queryBuilder;
         private readonly IFhirRequestContextAccessor _requestContextAccessor;
+        private readonly CosmosDataStoreConfiguration _cosmosConfig;
         private readonly SearchParameterInfo _resourceTypeSearchParameter;
         private readonly SearchParameterInfo _resourceIdSearchParameter;
+        public const string HeaderEnableChainedSearch = "x-ms-enable-chained-search";
         private const int _chainedSearchMaxSubqueryItemLimit = 100;
 
         public FhirCosmosSearchService(
@@ -41,17 +45,20 @@ namespace Microsoft.Health.Fhir.CosmosDb.Features.Search
             CosmosFhirDataStore fhirDataStore,
             IQueryBuilder queryBuilder,
             ISearchParameterDefinitionManager searchParameterDefinitionManager,
-            IFhirRequestContextAccessor requestContextAccessor)
+            IFhirRequestContextAccessor requestContextAccessor,
+            CosmosDataStoreConfiguration cosmosConfig)
             : base(searchOptionsFactory, fhirDataStore)
         {
             EnsureArg.IsNotNull(fhirDataStore, nameof(fhirDataStore));
             EnsureArg.IsNotNull(queryBuilder, nameof(queryBuilder));
             EnsureArg.IsNotNull(searchParameterDefinitionManager, nameof(searchParameterDefinitionManager));
             EnsureArg.IsNotNull(requestContextAccessor, nameof(requestContextAccessor));
+            EnsureArg.IsNotNull(cosmosConfig, nameof(cosmosConfig));
 
             _fhirDataStore = fhirDataStore;
             _queryBuilder = queryBuilder;
             _requestContextAccessor = requestContextAccessor;
+            _cosmosConfig = cosmosConfig;
             _resourceTypeSearchParameter = searchParameterDefinitionManager.GetSearchParameter(KnownResourceTypes.Resource, SearchParameterNames.ResourceType);
             _resourceIdSearchParameter = searchParameterDefinitionManager.GetSearchParameter(KnownResourceTypes.Resource, SearchParameterNames.Id);
         }
@@ -85,15 +92,6 @@ namespace Microsoft.Health.Fhir.CosmosDb.Features.Search
                 }
             }
 
-            if (searchOptions.CountOnly)
-            {
-                int count = await ExecuteCountSearchAsync(
-                    _queryBuilder.BuildSqlQuerySpec(searchOptions, includes: Array.Empty<IncludeExpression>()),
-                    cancellationToken);
-
-                return new SearchResult(count, searchOptions.UnsupportedSearchParams);
-            }
-
             if (hasIncludeOrRevIncludeExpressions && chainedExpressions.Count > 0)
             {
                 List<Expression> chainedReferences = await PerformChainedSearch(searchOptions, chainedExpressions, cancellationToken);
@@ -105,12 +103,22 @@ namespace Microsoft.Health.Fhir.CosmosDb.Features.Search
                 }
                 else
                 {
-                    return CreateSearchResult(
-                        searchOptions,
-                        ArraySegment<SearchResultEntry>.Empty,
-                        null,
-                        false);
+                    if (searchOptions.CountOnly)
+                    {
+                        return new SearchResult(0, searchOptions.UnsupportedSearchParams);
+                    }
+
+                    return CreateSearchResult(searchOptions, ArraySegment<SearchResultEntry>.Empty, continuationToken: null);
                 }
+            }
+
+            if (searchOptions.CountOnly)
+            {
+                int count = await ExecuteCountSearchAsync(
+                    _queryBuilder.BuildSqlQuerySpec(searchOptions, includes: Array.Empty<IncludeExpression>()),
+                    cancellationToken);
+
+                return new SearchResult(count, searchOptions.UnsupportedSearchParams);
             }
 
             (IReadOnlyList<FhirCosmosResourceWrapper> results, string continuationToken) = await ExecuteSearchAsync<FhirCosmosResourceWrapper>(
@@ -156,12 +164,19 @@ namespace Microsoft.Health.Fhir.CosmosDb.Features.Search
 
         private async Task<List<Expression>> PerformChainedSearch(SearchOptions searchOptions, IReadOnlyList<ChainedExpression> chainedExpressions, CancellationToken cancellationToken)
         {
+            if (!_cosmosConfig.EnableChainedSearch && _requestContextAccessor.FhirRequestContext?.RequestHeaders.TryGetValue(HeaderEnableChainedSearch, out StringValues _) != true)
+            {
+                throw new SearchOperationNotSupportedException(Resources.ChainedExpressionNotSupported);
+            }
+
             var chainedReferences = new List<Expression>();
             SearchOptions chainedOptions = searchOptions.Clone();
             chainedOptions.CountOnly = false;
             chainedOptions.MaxItemCount = _chainedSearchMaxSubqueryItemLimit;
             chainedOptions.MaxItemCountSpecifiedByClient = false;
 
+            // Local function for recursion, this lets us walk to the end of the expression to start filtering,
+            // the results are used to filter against the parent layer.
             async Task<Expression> RecursiveLookup(ChainedExpression expression)
             {
                 Expression criteria = expression.Expression;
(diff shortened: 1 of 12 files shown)


src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
```diff
@@ -90,6 +94,25 @@ public class FhirCosmosSearchService : SearchService
                 return new SearchResult(count, searchOptions.UnsupportedSearchParams);
             }
 
+            if (hasIncludeOrRevIncludeExpressions && (chainedExpressions.Count > 0))
+            {
+                List<Expression> chainedReferences = await PerformChainedSearch(searchOptions, chainedExpressions, cancellationToken);
+
+                if (chainedReferences.Any())
```
brendankowitz, line 101: TODO: Check if resulting `chainedReferences.Count` should eq `chainedExpressions.Count`. i.e. all chained queries should result in a filter

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
```diff
@@ -131,6 +154,115 @@ public class FhirCosmosSearchService : SearchService
...
+                            false,
+                            false,
+                            false));
+                    }
+
+                    // When reversed the ids from the sub-query will match the base resource type
+                    filteredType = expression.ResourceTypes.FirstOrDefault();
+                }
+
+                if (expression.Expression is ChainedExpression innerChained)
```
brendankowitz, line 192: Move to top of method

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
```diff
@@ -90,6 +94,25 @@ public class FhirCosmosSearchService : SearchService
                 return new SearchResult(count, searchOptions.UnsupportedSearchParams);
             }
 
+            if (hasIncludeOrRevIncludeExpressions && (chainedExpressions.Count > 0))
```
brendankowitz, line 97: nit: unneeded `(`

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
```diff
@@ -131,6 +162,128 @@ public class FhirCosmosSearchService : SearchService
             return searchResult;
         }
 
+        private async Task<List<Expression>> PerformChainedSearch(SearchOptions searchOptions, IReadOnlyList<ChainedExpression> chainedExpressions, CancellationToken cancellationToken)
+        {
+            if (!_cosmosConfig.EnableChainedSearch && _requestContextAccessor.FhirRequestContext?.RequestHeaders.TryGetValue(HeaderEnableChainedSearch, out StringValues _) != true)
```
Ivanidzo4ka, line 167: >&& [](start = 51, length = 2)

Shouldn't this be `||`
It's strange(at least for me ) to hide feature behind flag and header in same time.
I think what header should override server request in order to have granular control approach.

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
```diff
@@ -131,6 +162,128 @@ public class FhirCosmosSearchService : SearchService
...
+
+            var chainedReferences = new List<Expression>();
+            SearchOptions chainedOptions = searchOptions.Clone();
+            chainedOptions.CountOnly = false;
+            chainedOptions.MaxItemCount = _chainedSearchMaxSubqueryItemLimit;
+            chainedOptions.MaxItemCountSpecifiedByClient = false;
+
+            // Local function for recursion, this lets us walk to the end of the expression to start filtering,
+            // the results are used to filter against the parent layer.
+            async Task<Expression> RecursiveLookup(ChainedExpression expression)
```
Ivanidzo4ka, line 180: >async Task<Expression> RecursiveLookup(ChainedExpression expression) [](start = 12, length = 68)

nit: Can it be normal function rather than local? (For better readability?)
It's too big to be local, at least for my taste and size of my monitor. I can't even see where it ends.

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
```diff
@@ -131,6 +162,128 @@ public class FhirCosmosSearchService : SearchService
...
+                if (!expression.Reversed)
+                {
+                    // When normal chained expression we can filter using references in the parent object. e.g. Observation.subject
+                    // The following expression constrains "subject" references on "Observation" with the ids that have matched the sub-query
+                    chainedExpressionReferences = chainedResults.results.Select(x =>
+                            Expression.SearchParameter(
+                                expression.ReferenceSearchParameter,
+                                Expression.And(
+                                    Expression.Equals(FieldName.ReferenceResourceId, null, x.Id),
+                                    Expression.Equals(FieldName.ReferenceResourceType, null, filteredType))))
```
Ivanidzo4ka, line 247: >filteredType [](start = 93, length = 12)

Considering this can be null (it's a `FirstOrDefault()` and default is null for string) would it handle null correctly? Or we shouldn't add expression if it's a null?

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
```diff
@@ -131,6 +162,128 @@ public class FhirCosmosSearchService : SearchService
...
+
+                return chainedExpressionReferences.Length > 1 ? Expression.Or(chainedExpressionReferences) : chainedExpressionReferences.FirstOrDefault();
+            }
+
+            // Loop over all chained expressions in the search query and pass them into the local recursion function
+            foreach (ChainedExpression item in chainedExpressions)
+            {
+                Expression recursiveLookup = await RecursiveLookup(item);
+
+                // When no results are returned from a chain we can break since all operands in the search should be met
```
Ivanidzo4ka, line 275: >we can break  [](start = 61, length = 13)

Can I have situation where ReverseLookup on chainedExpression would yeild (something, null, something)? In this case we would ignore second something

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/Queries/ExpressionQueryBuilder.cs
```diff
@@ -198,13 +198,11 @@ public object VisitBinary(BinaryExpression expression, Context context)
 
         public object VisitChained(ChainedExpression expression, Context context)
         {
-            // TODO: This will be removed once it's implemented.
             throw new SearchOperationNotSupportedException(Resources.ChainedExpressionNotSupported);
```
Ivanidzo4ka, line 201: >SearchOperationNotSupportedException [](start = 22, length = 36)

should this be update to `NotSupportedException`?
Same for `VisitSortParameter`

Ivanidzo4ka: Maybe add comment to NotSupported what chained get handled by SearchService itself and we shouldn't reach this code at all.

---
In reply to: [585065737](https://github.com/microsoft/fhir-server/pull/1695#discussion_r585065737) [](ancestors = 585065737)

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/Queries/QueryBuilder.cs
```diff
@@ -47,14 +48,18 @@ public QueryBuilderHelper()
                 _queryHelper = new QueryHelper(_queryBuilder, _queryParameterManager, SearchValueConstants.RootAliasName);
             }
 
-            public QueryDefinition BuildSqlQuerySpec(SearchOptions searchOptions, IReadOnlyList<IncludeExpression> includes)
+            public QueryDefinition BuildSqlQuerySpec(SearchOptions searchOptions, IReadOnlyList<IncludeExpression> includes, bool idOnly)
```
Ivanidzo4ka, line 51: >idOnly [](start = 130, length = 6)

I would push that into SearchOptions class as property.
I think it looks better and you don't have to stretch function parameters.

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
```diff
@@ -131,6 +162,128 @@ public class FhirCosmosSearchService : SearchService
...
+                }
+
+                MultiaryExpression filterExpression = Expression.And(
+                    Expression.SearchParameter(_resourceTypeSearchParameter, Expression.StringEquals(FieldName.TokenCode, null, filteredType, false)),
+                    criteria);
+
+                chainedOptions.Expression = filterExpression;
+
+                var chainedResults = await ExecuteSearchAsync<FhirCosmosResourceWrapper>(
+                    _queryBuilder.BuildSqlQuerySpec(chainedOptions, includeExpressions, idsOnly: true),
```
Ivanidzo4ka, line 226: >idsOnly: true [](start = 88, length = 13)

It looks like you use ids only if `!expression.Reversed` so no point to fetch them, so maybe this can be enum: `Default`, `Id`, `Nothing`(IncludeOnly)?

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
brendankowitz: So treat this example as we're all learning together :) So, given the case:

```
GET Patient?general-practitioner.name=Joe&general-practitioner.address-state=MN&general-practitioner.address-postcode=1234
```

https://www.hl7.org/fhir/search.html#chaining
> Note that chained parameters are applied independently to the target resource
> may return Patients cared for by Joe from CA and Jane from MN [and Smith from 1234]: no one practitioner need satisfy [all] conditions

---

So in the case above if the filtered results returned (Joe#123, null, Smith#789) we could not satisfy the second condition which means pseudo logic would be: 

```sql
SELECT Patient WHERE Patient.GeneralPractitioner IN (Joe#123)
AND Patient.GeneralPractitioner IN ()
AND Patient.GeneralPractitioner IN (Smith#789)
```

I think this then makes sense as soon as a condition cannot be met to short-circuit and return, thoughts / different interpretation?

Ivanidzo4ka: Thanks for great explanation.
If that's the case, why not actually do proper short-circuit and do `return null` instead of the `break` ?

---
In reply to: [585140401](https://github.com/microsoft/fhir-server/pull/1695#discussion_r585140401) [](ancestors = 585140401)

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
brendankowitz: Changed to `First()`, it shouldn't be possible without a type

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/Queries/ExpressionQueryBuilder.cs
brendankowitz: Added a comment. Keeping the existing exception for now

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/Queries/QueryBuilder.cs
brendankowitz: I had this though also, but `SearchOptions` is in Core and applicable to all DBs, this seemed specific to this implementation. I'll think about it some more

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/Queries/QueryBuilderOptions.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.Collections.Generic;
+using Microsoft.Health.Fhir.Core.Features.Search.Expressions;
+
+namespace Microsoft.Health.Fhir.CosmosDb.Features.Search.Queries
+{
+    internal class QueryBuilderOptions
+    {
+        public QueryBuilderOptions(IReadOnlyList<IncludeExpression> includes = null, QueryProjection projection = QueryProjection.Default)
+        {
+            Includes = includes;
```
Ivanidzo4ka, line 15: >includes [](start = 23, length = 8)

nit: Includes = includes?? Array.Empty<IncludeExpression>()
and you don't have to write that part of code in service.
A bit hacky, but I think it would make code more cleaner.

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/Queries/QueryBuilder.cs
brendankowitz: Added `QueryBuilderOptions` to encapsulate these

Ivanidzo4ka (approved):
:shipit:

PR #1696, opened 2021-02-25
Split out variable for e2e tests agent

brendankowitz:
## Description
E2E tests consume 9 agents (Stu3, R4, R5) x (Integration, Sql, Cosmos) configurations. We can run these on a separate shared pool that can scale beyond 10 parallel jobs.

After observing this for a while, I think this would be the best balance of the following:

* Free hosted agents have a max parallel job count of 10, but are always available
* Shared agents take ~3-4 mins to spin up on demand, but scale well beyond 10.

By offloading the largest bottleneck we can keep the compilation fast and scale out the running of tests.

### DIFF
--- a/build/build-variables.yml
+++ b/build/build-variables.yml
@@ -26,6 +26,7 @@ variables:
   LinuxVmImage: 'ubuntu-latest'
   # The following is set by a build Pipeline variable:
   # DefaultLinuxPool: 'Azure Pipelines'
+  # SharedLinuxPool: 'Azure Pipelines'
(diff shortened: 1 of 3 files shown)


LTA-Thinking:
Might want to also switch over the AKS E2E tests too since once we start running them again they will be at the same point in the bottleneck.

LTA-Thinking (approved):

brendankowitz:
Thanks RB, I'm going to leave the AKS ones for now since I can't test it. But, the pools and variables are there, so it should be fine when we look at it again

Issue #1660, opened 2021-02-12
Unable to deploy ARM template using FhirServerSqlServer


cashman04:
If I use the template found [here](https://portal.azure.com/#create/Microsoft.Template/uri/https%3A%2F%2Fraw.githubusercontent.com%2FMicrosoft%2Ffhir-server%2Fmaster%2Fsamples%2Ftemplates%2Fdefault-azuredeploy.json) either through powershell or manually in the portal, I get the following error if I set the solutionType to FhirServerSqlServer:

    Deployment template validation failed: 'Template parameter JToken type is not valid. Expected 'Boolean'. Actual 'String'. Please see https://aka.ms/resource-manager-parameter-files for usage details.'. (Code: InvalidTemplate)

This used to work and let me deploy with sqlserver and not cosmosdb.

cashman04:
After more testing: This seems to be an issue with the latest version of the ARM template when CosmosDB was added.  I can go back to an earlier template before CosmosDB and it works fine.

brendankowitz:
I think there is a bug here, it looks like it might be caused by one of the required parameters not having the value set properly:
![image](https://user-images.githubusercontent.com/197221/108130444-8710d580-7064-11eb-8579-c9508f663b65.png)

cashman04:
Yes, that is the offending line.  It seems like the first half of an if statement was set that returns a bool of false, but the else isn't set which then returns a string.  I have fixed this by running a local copy of the template and just hard coding a false value.

CaitlinV39:
AB#79894

Ivanidzo4ka:
In case of the SQL deployment we want you to make deliberate choice of how to handle schema updates, that's why it fails during validation.

CaitlinV39:
@cashman04 let us know if you have any additional questions. I'm going to close for now but we can reopen if there are still open items.

PR #1700, opened 2021-02-26
Fix template samples

Ivanidzo4ka:
## Description
Mixing bool and string force template engine to throw error.
So switch to 'auto' / 'tool' string values

## Related issues
Addresses [#1660].

## Testing
Manually

## FHIR Team Checklist

Skip (it's a change in samples)

### DIFF
--- a/samples/templates/default-azuredeploy-docker.json
+++ b/samples/templates/default-azuredeploy-docker.json
@@ -147,7 +147,6 @@
         "sqlSchemaAutomaticUpdatesEnabled": {
             "type": "string",
 			"allowedValues":["false", "true"],
-			"minLength":"1",
             "defaultValue": "[if(equals(parameters('solutionType'),'FhirServerCosmosDB'), 'false', '')]",
             "metadata": {
                 "description": "Determine whether the sql schema should be automatically upgraded on server startup. If false, sql schema will not be initialized or upgraded on the server startup. The schema migration tool will be required to perform initialize or upgrade. If true, sql schema will be upgraded to the maximum supported version."
--- a/samples/templates/default-azuredeploy-docker.json
+++ b/samples/templates/default-azuredeploy-docker.json
@@ -146,8 +146,8 @@
         },
         "sqlSchemaAutomaticUpdatesEnabled": {
             "type": "string",
-            "allowedValues":["false", "true"],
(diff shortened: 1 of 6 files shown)


LTA-Thinking:
The issue here is that we have two conflicting requirements for this parameter:
* When Cosmos is the datastore the user shouldn't have to put anything in the field.
* When SQL is the datastore there shouldn't be a default value so the user has to manually pick if they want automatic updates or not.

This was why I had initially tried to make it so the default value was false when using Cosmos, or not set if using SQL.
@rbans96, what is your thought?

LTA-Thinking:
Relevant PR: #1301

samples/templates/default-azuredeploy-docker.json
```diff
@@ -146,7 +146,7 @@
         },
         "sqlSchemaAutomaticUpdatesEnabled": {
             "type": "bool",
-            "defaultValue": "[if(equals(parameters('solutionType'),'FhirServerCosmosDB'), bool('false'), '')]",
+            "defaultValue": "[if(equals(parameters('solutionType'),'FhirServerCosmosDB'), false(), true())]",
```
brendankowitz, line 149: I think the intended behavior was to have no default value when sql to force a selection.

- The above is only true when the datalayer targets SQL.
- The value `true` is know to have problems when running on a webfarm (should be dev/test only for now)

Ivanidzo4ka:
How about now?

Pass validation if it's cosmos, and complains about incorrect value if it's sql. (I wish I could copy error message, but for some reason azure localize error string for me, so I don't think it would be useful for English speaking people)


---
In reply to: [786959225](https://github.com/microsoft/fhir-server/pull/1700#issuecomment-786959225) [](ancestors = 786959225)

Ivanidzo4ka:
From what I see it's either this, and I need to change yml file for pipelines (to change it from boolean to string) which is somewhat ugly from developer point of view, but I can't find any other way to have unset boolean value, or we can set default value to false for all cases.

---
In reply to: [786974696](https://github.com/microsoft/fhir-server/pull/1700#issuecomment-786974696) [](ancestors = 786974696,786959225)

LTA-Thinking (approved):
This looks like the best option we have to me, but I'd like @brendankowitz opinion as well.

LTA-Thinking (approved):

PR #1702, opened 2021-02-27
Updated Azure Storage libraries to v12

valeneiko:
## Description
Updated Azure Storage library to v12, because v11 is deprecated

```
Microsoft.Azure.Storage.Blob (v11.2.2) ==> Azure.Storage.Blobs (v12.8.0)
```

## Related issues
Addresses [issue #].

## Testing
Existing tests pass.

## FHIR Team Checklist


src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/AzureAccessTokenClientInitializer.cs
```diff
@@ -7,42 +7,37 @@
...
-        private readonly IAccessTokenProvider _accessTokenProvider;
+        private readonly TokenCredential _credential;
         private readonly ExportJobConfiguration _exportJobConfiguration;
-        private readonly ILogger<AzureAccessTokenClientInitializer> _logger;
 
         public AzureAccessTokenClientInitializer(
-            IAccessTokenProvider accessTokenProvider,
-            IOptions<ExportJobConfiguration> exportJobConfiguration,
-            ILogger<AzureAccessTokenClientInitializer> logger)
+            TokenCredential credential,
```
namadabu, line 25: Since we do not register a service of type `TokenCredential` the framework will fail to resolve it and hence cannot create an `AzureAccessTokenClientInitializer` object.

src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/AzureAccessTokenClientInitializer.cs
```diff
@@ -7,42 +7,37 @@
...
+            IOptions<ExportJobConfiguration> exportJobConfiguration)
         {
-            EnsureArg.IsNotNull(accessTokenProvider, nameof(accessTokenProvider));
+            EnsureArg.IsNotNull(credential, nameof(credential));
             EnsureArg.IsNotNull(exportJobConfiguration?.Value, nameof(exportJobConfiguration));
-            EnsureArg.IsNotNull(logger, nameof(logger));
 
-            _accessTokenProvider = accessTokenProvider;
```
namadabu, line 35: Is there a specific reason we do not want to use this interface anymore. The idea behind it was to allow an easy way to change the logic of acquiring access tokens by injecting a different implementation of `IAccessTokenProvider`.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/GetExportRequestHandler.cs
```diff
@@ -58,7 +58,7 @@ public async Task<GetExportResponse> Handle(GetExportRequest request, Cancellati
                 var jobResult = new ExportJobResult(
                     outcome.JobRecord.QueuedTime,
                     outcome.JobRecord.RequestUri,
-                    requiresAccessToken: false,
+                    requiresAccessToken: true,
```
anyname2, line 61: This should not be part of the PR

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -635,7 +635,7 @@ private async Task ProcessSearchResultsAsync(IEnumerable<SearchResultEntry> sear
         {
             // Update the continuation token in local cache and queryParams.
             // We will add or udpate the continuation token in the query parameters list.
-            progress.UpdateContinuationToken(Convert.ToBase64String(System.Text.Encoding.UTF8.GetBytes(continuationToken)));
+            progress.UpdateContinuationToken(continuationToken);
```
anyname2, line 638: This should not be part of the PR

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Export/ExportJobTaskTests.cs
```diff
@@ -1259,7 +1259,7 @@ public async Task GivenAGroupExportJobToResume_WhenExecuted_ThenAllPatientResour
                     {
                         // The ids aren't in the query parameters because of the reset
                         ids = new string[] { "1", "2", "3" };
-                        continuationTokenIndex = int.Parse(Encoding.UTF8.GetString(Convert.FromBase64String(x.ArgAt<IReadOnlyList<Tuple<string, string>>>(1)[2].Item2)).Substring(2));
+                        continuationTokenIndex = int.Parse(x.ArgAt<IReadOnlyList<Tuple<string, string>>>(1)[2].Item2.Substring(2));
```
anyname2, line 1262: This should not be part of the PR

src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/AzureAccessTokenClientInitializer.cs
anyname2: Good catch! I forgot to include that change in the PR.

src/Microsoft.Health.Fhir.Azure/ExportDestinationClient/AzureAccessTokenClientInitializer.cs
anyname2: AFAIK, all the azure libraries are now using `Azure.Core` for auth, hence `TokenCredential` is accepted as a parameter. I didn't see an easy way to use `IAccessTokenProvider` here.
Also `TokenCredential` is an abstract class, so in the sense of injecting different implementation, you can still do that.

anyname2: Actually, there is even a diclaimer that it should not be used anymore: https://docs.microsoft.com/en-us/dotnet/api/overview/azure/service-to-service-authentication

namadabu:
@anyname2 Apologies for the late response. Thanks for bringing this up. This SDK update is something we definitely need to do. But this update also introduces breaking changes and we have to update our build/test pipelines and our managed service to handle this change. We are planning on working on this later rather than now. @CaitlinV39 FYI

rbans96:
Closing this out as currently we don't have any plan in near future to update azure storage libraries but we can definitely refer this PR whenever we plan this in long run.

PR #1703, opened 2021-03-01
Cleanup aad in beginning of build

brendankowitz:
## Description
Performs the AAD cleanup of test applications at the start of the build

## Related issues
Addresses the role conflict exception that occasionally occurs in builds:
image

## Testing

## FHIR Team Checklist

### DIFF
--- a/build/pr-pipeline.yml
+++ b/build/pr-pipeline.yml
@@ -83,6 +83,8 @@ stages:
           catch
           {}
           New-AzResourceGroup -Name "$(ResourceGroupName)" -Location "$(ResourceGroupRegion)" -Force
+  - template: ./jobs/cleanup-aad.yml
+
 
 - stage: aadTestEnvironment
   displayName: Setup AAD Test Environment
--- a/build/pr-pipeline.yml
+++ b/build/pr-pipeline.yml
@@ -90,7 +90,6 @@ stages:
   displayName: Setup AAD Test Environment
   dependsOn:
   - provisionEnvironment
-  - DockerBuild
   jobs:
   - template: ./jobs/add-aad-test-environment.yml
 
@@ -99,6 +98,7 @@ stages:
   condition: and(succeeded(), eq(variables.enableAks, 'true'))
   dependsOn:
   - DockerBuild
+  - aadTestEnvironment
   jobs:
   - template: ./jobs/deploy-aks.yml
     parameters: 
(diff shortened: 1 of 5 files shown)


Ivanidzo4ka (approved):
:shipit:

build/pr-pipeline.yml
```diff
@@ -83,6 +83,8 @@ stages:
           catch
           {}
           New-AzResourceGroup -Name "$(ResourceGroupName)" -Location "$(ResourceGroupRegion)" -Force
+  - template: ./jobs/cleanup-aad.yml
```
Ivanidzo4ka, line 86: >cleanup [](start = 21, length = 7)

I find it's a bit strange to be part of this stage, and not one below (aadTestEnvironment)
In my head it feels more natural to bundle together work with AAD and put cleanup right before we setup test environment.

brendankowitz: It needs to happen before the Template deployment. I'll see if it makes sense to move the setup step here too

LTA-Thinking: Could you add it to the aadTestEnvironment stage as a second template before the add template? That way all the aad steps would be together.

brendankowitz: I changed some of the dependencies so I now have this, I think this looks better:
image

Ivanidzo4ka: >It needs to happen before the Template deployment.

Any chance you can elaborate? I'm not sure what Template you talking about.

Looking on that picture, why not put Setup AAD test as job into Provision Environment stage? (and give it better name than setup :D )

LTA-Thinking (approved):

build/pr-pipeline.yml
```diff
@@ -83,20 +83,22 @@ stages:
...
           New-AzResourceGroup -Name "$(ResourceGroupName)" -Location "$(ResourceGroupRegion)" -Force
-
-- stage: aadTestEnvironment
-  displayName: Setup AAD Test Environment
-  dependsOn:
-  - provisionEnvironment
-  - DockerBuild
-  jobs:
-  - template: ./jobs/add-aad-test-environment.yml
+  - template: ./jobs/cleanup-aad.yml
```
LTA-Thinking, line 86: The provision environment stage now needs to depend on the SemVer stage since add AAD uses the environment name.

LTA-Thinking (changes requested):

build/pr-pipeline.yml
brendankowitz: Thanks! I think I've keep the provision stage separate and move the aad cleanup/create to its own stage

brendankowitz: Dependencies now look like this. 
image

LTA-Thinking (approved):

PR #1704, opened 2021-03-01
Move away from using Cosmos DB upsert stored procedure

johnstairs:
## Description
The stored procedure that we use for upsertings resources limits our ability to get full provisioned RU utilization. This is because the stored procedure always runs on the write replica of a physical partition and script RU usage (the javascript itself) and its reads are all charged to the write replica, which means that in practice, with many concurrent calls to the sproc, we observe the write partition getting saturated before the read replicas and the DB throttling us before we reach our full provisioned throughput.

The reason for using the stored procedure was for its transactional semantics, but Cosmos now support ["transactional batches"](https://devblogs.microsoft.com/cosmosdb/introducing-transactionalbatch-in-the-net-sdk/), which are also transactional and more efficient. 

The downside is that this approach is chattier in update scenarios, but latency is still better.

Here is a performance comparison of crate and update operations:

image

One caveat is that we are no longer preserving search indexes in historical documents (though we weren't using those anyway).

## Related issues
Addresses [AB#79744](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/79744)

## Testing

Manual and perf tests

Patch


src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
```diff
@@ -40,9 +44,9 @@ public sealed class CosmosFhirDataStore : IFhirDataStore, IProvideCapability
         private readonly RetryExceptionPolicyFactory _retryExceptionPolicyFactory;
         private readonly ILogger<CosmosFhirDataStore> _logger;
 
-        private static readonly UpsertWithHistory _upsertWithHistoryProc = new UpsertWithHistory();
         private static readonly HardDelete _hardDelete = new HardDelete();
         private static readonly ReplaceSingleResource _replaceSingleResource = new ReplaceSingleResource();
+        private static readonly RecyclableMemoryStreamManager _recyclableMemoryStreamManager = new();
```
YazanMSFT, line 49: That's _new_ to me!

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
```diff
@@ -93,53 +97,125 @@ public sealed class CosmosFhirDataStore : IFhirDataStore, IProvideCapability
...
+                    return new UpsertOutcome(cosmosWrapper, SaveOutcomeType.Created);
+                }
+                catch (CosmosException e) when (e.StatusCode == HttpStatusCode.Conflict)
+                {
+                    // this means there is already an existing version of this resource
+                }
             }
-            catch (CosmosException exception)
+
+            while (true)
```
YazanMSFT, line 125: Could you change this to while (!cancellationToken.IsCancellationRequested), or add a counter to limit retries? If the Cosmos DB SDK had an update with a bug that stopped it from respecting cancellation tokens, or some other bug is introduced that would ignore them, we could have an infinite loop here. Those are never fun to chase down in prod :)

YazanMSFT (approved):

namadabu (approved):
:shipit:

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
```diff
@@ -93,53 +97,125 @@ public sealed class CosmosFhirDataStore : IFhirDataStore, IProvideCapability
...
+                {
+                    await _retryExceptionPolicyFactory.GetRetryPolicy().ExecuteAsync(
+                        async ct => await _containerScope.Value.CreateItemAsync(
+                            cosmosWrapper,
+                            partitionKey,
+                            cancellationToken: ct,
+                            requestOptions: new ItemRequestOptions { EnableContentResponseOnWrite = false }),
+                        cancellationToken);
+
+                    return new UpsertOutcome(cosmosWrapper, SaveOutcomeType.Created);
```
brendankowitz, line 117: I'm sure its in there somewhere, where does the Version get set now?

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
```diff
@@ -93,53 +97,125 @@ public sealed class CosmosFhirDataStore : IFhirDataStore, IProvideCapability
...
-                _logger.LogError(exception, "Unhandled Document Client Exception");
+                if (existingItemResource.IsDeleted && cosmosWrapper.IsDeleted)
+                {
+                    return null;
+                }
 
-                throw;
+                existingItemResource.IsHistory = true;
+                existingItemResource.ActivePeriodEndDateTime = cosmosWrapper.LastModified;
+                existingItemResource.SearchIndices = null;
```
brendankowitz, line 165: Should this be moved into `if (keepHistory)`?

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
```diff
@@ -93,53 +97,125 @@ public sealed class CosmosFhirDataStore : IFhirDataStore, IProvideCapability
...
+                if (keepHistory)
+                {
+                    TransactionalBatchResponse transactionalBatchResponse = await _containerScope.Value.CreateTransactionalBatch(partitionKey)
+                        .ReplaceItem(cosmosWrapper.Id, cosmosWrapper, new TransactionalBatchItemRequestOptions { EnableContentResponseOnWrite = false, IfMatchEtag = existingItemResource.ETag })
+                        .CreateItem(existingItemResource, new TransactionalBatchItemRequestOptions { EnableContentResponseOnWrite = false })
+                        .ExecuteAsync(cancellationToken);
+
+                    if (!transactionalBatchResponse.IsSuccessStatusCode)
+                    {
+                        if (transactionalBatchResponse.StatusCode == HttpStatusCode.PreconditionFailed)
```
brendankowitz, line 196: Does the batch roll-up the status code? I think it also reports the status of each batch item

johnstairs: Yes, this is a roll-up. Because there are only two operations, the status code will reflect the one that failed (the other's status will be dependency failed)

Ivanidzo4ka:
>Here is a performance comparison of crate and update operations:
Can you give a bit more information how these metric got produced? Is it single machine - local cosmos DB? Or it's been measured in cloud? And is it 95% percentile, or it's just one run? It's hard to grasp from just numbers.

johnstairs:
> > Here is a performance comparison of crate and update operations:
> > Can you give a bit more information how these metric got produced? Is it single machine - local cosmos DB? Or it's been measured in cloud? And is it 95% percentile, or it's just one run? It's hard to grasp from just numbers.

Good question. I measured this in the cloud with one DB with 20K RUs and one client sequentially calling the server. (I also did runs with concurrent clients and servers, but results were noisier because of throttling). Numbers in the charts above are averages. 

Percentiles here:

|  perfRunId | requestCount | avg_duration | avg_RequestCharge | percentile_duration_50 | percentile_duration_90 | percentile_duration_99 | percentile_duration_99_9 | percentile_duration_99_99 | percentile_RequestCharge_50 | percentile_RequestCharge_90 | percentile_RequestCharge_99 | percentile_RequestCharge_99_9 | percentile_RequestCharge_99_99 |
|:----------:|:------------:|:------------:|:-----------------:|:----------------------:|:----------------------:|:----------------------:|:------------------------:|:-------------------------:|:---------------------------:|:---------------------------:|:---------------------------:|:-----------------------------:|:------------------------------:|
| new-update | 18132        | 17.87581     | 31.98333          | 17.26048               | 20.43645               | 27.3726                | 62.867                   | 203.3592                  | 31.76                       | 32.8                        | 38.48                       | 47.67                         | 47.67                          |
| new-create | 39153        | 10.43389     | 25.20107          | 9.98491                | 12.46816               | 19.59241               | 39.8585                  | 148.6354                  | 24.19                       | 28.38                       | 32.57                       | 32.57                         | 39.81                          |
| old-uptate | 17934        | 18.78817     | 49.06412          | 18.24657               | 21.70533               | 28.2897                | 65.2809                  | 209.4772                  | 48.95088                    | 52.85604                    | 58.43733                    | 68.27                         | 71.49                          |
| old-create | 24204        | 13.00499     | 28.46857          | 12.4819                | 15.55195               | 22.00875               | 48.0029                  | 200.1187                  | 27.70067                    | 32.11379                    | 35.818                      | 37.95                         | 44.34                          |

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
johnstairs: Good point. Added a ThowIfCancellationRequested() in the beginning of the loop.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
johnstairs: [Here](https://github.com/microsoft/fhir-server/blob/64b665a561caa15420c558c2614e3d0d28f2ba54/src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs#L169-L169) on the wrapper
[Here](https://github.com/microsoft/fhir-server/blob/a70c3c745d0d01975ae113dedb7a15b3665bacb5/src/Microsoft.Health.Fhir.Core/Extensions/RawResourceElementExtensions.cs#L65-L69) in the raw resource

brendankowitz (approved):

PR #1709, opened 2021-03-02
Reindexes resources in bulk in SQL

rotodd:
## Description
This PR implements reindexing resources in bulk in SQL by:

- Defining table-valued parameters (TVPs) that hold the resource ID and resource type ID
- Adding a stored procedure ` dbo.BulkReindexResources` that takes in the new TVPs
- Creating row generators for the new TVPs

## Related issues
Addresses [AB#73395](https://microsofthealth.visualstudio.com/Health/_workitems/edit/73395).

## Testing
Extends existing Cosmos DB tests to SQL and adds new `FhirStorageTests` that call the bulk reindexing method. This PR also pulls out the `SearchParameterStatusManager` from the `ReindexJobTests` and puts it in the test fixture to allow it to be used in the `FhirStorageTests`.

## FHIR Team Checklist


src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -283,12 +287,31 @@ public async Task HardDeleteAsync(ResourceKey key, CancellationToken cancellatio
...
+                    {
+                        // TODO: we should attempt to reindex resources that failed to be reindexed
+                        case SqlErrorCodes.PreconditionFailed:
+                            throw new PreconditionFailedException(string.Format(Core.Resources.ReindexingResourceVersionConflict));
+                        case SqlErrorCodes.NotFound:
+                            throw new ResourceNotFoundException(string.Format(Core.Resources.ReindexingResourceNotFound));
+                        default:
+                            _logger.LogError(e, "Error from SQL database on reindex");
+                            throw;
+                    }
```
rotodd, line 313: Note that this will be updated once we reattempt to reindex resources that fail here ([AB#79925](https://microsofthealth.visualstudio.com/Health/_workitems/edit/79925) is tracking this).

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/FhirStorageTests.cs
```diff
@@ -597,93 +604,321 @@ public async Task GivenATransactionHandler_WhenATransactionFailsFailedRequest_Th
         }
 
         [Fact]
-        [FhirStorageTestsFixtureArgumentSets(DataStore.CosmosDb)]
-        public async Task GivenAnUpdatedResource_WhenUpdateSearchIndexForResourceAsync_ThenResourceGetsUpdated()
+        public async Task GivenAnUpdatedResource_WhenUpdatingSearchParameterIndexAsync_ThenResourceMetadataIsUnchanged()
```
rotodd, line 607: @feordin I updated this test so that it can be used for both Cosmos and SQL, and, as a result, I had to remove the check that the Cosmos ETag changed.

I think this should be alright, since we check that a reindexed resource is returned in searches with the new search parameter in `ReindexJobTests`. I could be missing something though, so please let me know if this looks alright!

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -283,12 +287,31 @@ public async Task HardDeleteAsync(ResourceKey key, CancellationToken cancellatio
 
         public async Task UpdateSearchParameterIndicesBatchAsync(IReadOnlyCollection<ResourceWrapper> resources, CancellationToken cancellationToken)
```
rotodd, line 288: I plan to rename this method along with `UpdateSearchIndexForResourceAsync` in [AB#80106](https://microsofthealth.visualstudio.com/Health/_workitems/edit/80106).

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/8.diff.sql
```diff
@@ -922,3 +922,494 @@ AS
...
+        ON searchIndex.ResourceTypeId = resourceToReindex.ResourceTypeId
+            AND searchIndex.ResourceId = resourceToReindex.ResourceId
+
+    INSERT INTO dbo.ReferenceSearchParam
+        (ResourceTypeId, ResourceSurrogateId, SearchParamId, BaseUri, ReferenceResourceTypeId, ReferenceResourceId, ReferenceResourceVersion, IsHistory)
+    SELECT DISTINCT searchIndex.ResourceTypeId, resourceToReindex.ResourceSurrogateId, searchIndex.SearchParamId, searchIndex.BaseUri, searchIndex.ReferenceResourceTypeId, searchIndex.ReferenceResourceId, searchIndex.ReferenceResourceVersion, 0
+    FROM @bulkReferenceSearchParams searchIndex
+    INNER JOIN @computedValues resourceToReindex
+        ON searchIndex.ResourceTypeId = resourceToReindex.ResourceTypeId
+            AND searchIndex.ResourceId = resourceToReindex.ResourceId
```
rotodd, line 1308: @johnstairs this follows the logic in `UpsertResource` when we extract `@previousResourceSurrogateId`, and it filters on both `ResourceTypeId` and `ResourceId`. I'm wondering, is it necessary to have `ResourceTypeId`?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/8.diff.sql
```diff
@@ -922,3 +922,494 @@ AS
...
+--     @bulkTokenDateTimeCompositeSearchParams
+--         * Extracted token$datetime search params
+--     @bulkTokenQuantityCompositeSearchParams
+--         * Extracted token$quantity search params
+--     @bulkTokenStringCompositeSearchParams
+--         * Extracted token$string search params
+--     @bulkTokenNumberNumberCompositeSearchParams
+--         * Extracted token$number$number search params
+--
+CREATE PROCEDURE dbo.BulkReindexResources
```
rotodd, line 1149: @johnstairs I created [AB#80109](https://microsofthealth.visualstudio.com/Health/_workitems/edit/80109) to investigate the potential performance gains of using `MERGE` instead of `DELETE` + `INSERT`.

src/Microsoft.Health.Fhir.Core/Resources.Designer.cs
```diff
@@ -745,6 +745,24 @@ internal class Resources {
...
+        }
+        
+        /// <summary>
+        ///   Looks up a localized string similar to The version of one or more resources did not match, indicating that they have been updated since the reindex job kicked off..
+        /// </summary>
+        internal static string ReindexingResourceVersionConflict {
+            get {
+                return ResourceManager.GetString("ReindexingResourceVersionConflict", resourceCulture);
+            }
+        }
```
rotodd, line 764: These will likely be updated once we catch these errors and attempt to reindex again (please see [AB#79925](https://microsofthealth.visualstudio.com/Health/_workitems/edit/79925)).

feordin (approved):
I reviewed the changes in the test fixtures, and they look good.

johnstairs:
Can we use the bulk TVPs for the UpsertResource stored procedure, so we can phase out the old ones?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/8.diff.sql
johnstairs: Yes, because the resourceId is not necessarily unique across types

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/8.diff.sql
```diff
@@ -922,3 +922,494 @@ AS
...
+    DELETE searchIndex FROM dbo.TokenNumberNumberCompositeSearchParam searchIndex
+    INNER JOIN @computedValues resourceToReindex
+        ON searchIndex.ResourceSurrogateId = resourceToReindex.ResourceSurrogateId
+
+    -- Next, insert all the new indices.
+    INSERT INTO dbo.ResourceWriteClaim
+        (ResourceSurrogateId, ClaimTypeId, ClaimValue)
+    SELECT DISTINCT resourceToReindex.ResourceSurrogateId, searchIndex.ClaimTypeId, searchIndex.ClaimValue
+    FROM @bulkResourceWriteClaims searchIndex
+    INNER JOIN @computedValues resourceToReindex
```
johnstairs, line 1290: An alternate approach would be to pass in the ResourceSurrogateId, which would be preseved from when the resource was read. It would mean fewer joins. Probably not a big change in performance though.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/8.sql
```diff
@@ -133,6 +133,14 @@ CREATE TYPE dbo.ResourceWriteClaimTableType_1 AS TABLE
     ClaimValue nvarchar(128) NOT NULL
 )
 
+CREATE TYPE dbo.BulkResourceWriteClaimTableType_1 AS TABLE
+(
+    ResourceTypeId smallint NOT NULL,
+    ResourceId varchar(64) COLLATE Latin1_General_100_CS_AS NOT NULL,
```
johnstairs, line 139: I'd like for a surrogate id field (or an offset, which is what I think I called it in my branch) to be included in these so that the table types can be re-used for bulk upserting, even if we ignore the field when reindexing.

rotodd: Replaced `ResourceId` and `ResourceTypeId` with `Offset` in 46be3d9.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/8.diff.sql
rotodd: I was able to `JOIN` on `Offset` in 46be3d9. Let me know what you think.

rotodd:
> Can we use the bulk TVPs for the UpsertResource stored procedure, so we can phase out the old ones?

Good idea - fixed in 29e3e52.

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlServerSchemaUpgradeTests.cs
```diff
@@ -50,7 +50,7 @@ public async Task GivenTwoSchemaInitializationMethods_WhenCreatingTwoDatabases_B
             }
         }
 
-        [Fact]
+        [Fact(Skip = "Disable until AB80636")]
```
rotodd, line 53: The plan is to complete [AB#80636](https://microsofthealth.visualstudio.com/Health/_workitems/edit/80636) before I merge the `feature/sql-reindex` feature branch.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/BulkTokenNumberNumberCompositeSearchParameterV1RowGenerator.cs
```diff
@@ -0,0 +1,54 @@
...
+
+        internal override bool TryGenerateRow(
+            int offset,
+            short searchParamId,
+            (TokenSearchValue component1, NumberSearchValue component2, NumberSearchValue component3) searchValue,
+            out BulkTokenNumberNumberCompositeSearchParamTableTypeV1Row row)
+        {
+            if (_tokenRowGenerator.TryGenerateRow(offset, default, searchValue.component1, out var token1Row) &&
+                _numberV1RowGenerator.TryGenerateRow(offset, default, searchValue.component2, out var token2Row) &&
+                _numberV1RowGenerator.TryGenerateRow(offset, default, searchValue.component3, out var token3Row))
```
rotodd, line 31: @johnstairs, if you look at the diff between this file and `BulkTokenNumberNumberCompositeSearchParameterRowGenerator.cs` from the [bulk upsert branch](https://github.com/microsoft/fhir-server/compare/personal/johnstairs/prtotype-optimized-bundle-transactions), you'll see that `default` is being passed in as the first arg instead of `offset`. Is this because the token and number rows generated here are scoped to this method (ie only used in the creation of the row on line 34)?

johnstairs: Yeah, we don't use the offset that is passed in to the call to _tokenRowGenerator.TryGenerateRow (we don't use token1Row.Offset), so I just made the argument default.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -294,12 +298,31 @@ public async Task HardDeleteAsync(ResourceKey key, CancellationToken cancellatio
...
+                catch (SqlException e)
+                {
+                    switch (e.Number)
+                    {
+                        // TODO: we should attempt to reindex resources that failed to be reindexed
+                        case SqlErrorCodes.PreconditionFailed:
+                            throw new PreconditionFailedException(string.Format(Core.Resources.ReindexingResourceVersionConflict));
+                        case SqlErrorCodes.NotFound:
+                            throw new ResourceNotFoundException(string.Format(Core.Resources.ReindexingResourceNotFound));
+                        default:
```
johnstairs, line 321: You could probably also end up with deadlocking as well

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/BulkTokenNumberNumberCompositeSearchParameterV1RowGenerator.cs
rotodd: Makes sense, thanks for checking. Changed in 0c52f29.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/8.sql
```diff
@@ -212,6 +226,16 @@ CREATE TYPE dbo.ReferenceSearchParamTableType_2 AS TABLE
     ReferenceResourceVersion int NULL
 )
 
+CREATE TYPE dbo.BulkReferenceSearchParamTableType_1 AS TABLE
```
johnstairs, line 229: I believe you can remove the non-bulk types now (like dbo.ReferenceSearchParamTableType_2 above)

rotodd: Oops, good point. Removed the old types in a080589.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
rotodd: I can see how the `UPDLOCK` on the index `IX_Resource_ResourceTypeId_ResourceId` could block other transactions, and I can see transactions being rolled back because we have RCSI enabled - but what is the second resource that would lead to a deadlock?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/8.diff.sql
```diff
@@ -8,6 +8,42 @@ ADD
 
 GO
 
+DROP PROCEDURE IF EXISTS dbo.UpsertResource;
+
+DROP PROCEDURE IF EXISTS dbo.UpsertResource_2;
+
+DROP TYPE IF EXISTS dbo.ResourceWriteClaimTableType_1;
```
johnstairs, line 15: I think it's too early to drop these. Consider this scenario:
- Server is running latest version of code (this) and on V7 of the schema. So it uses UpsertResource_2 to insert resources.
- We run 8.diff.
- Then UpsertResource2 is gone before the server knows to use the V8 schema.

rotodd: As discussed, I reverted the changes to drop the types in 8113277 and created the story [AB#80754](https://microsofthealth.visualstudio.com/Health/_workitems/edit/80754) to make sure we drop the unused types in a later schema version.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
rotodd: As discussed, it makes sense to add retry logic at this layer for both precondition failed exceptions and exceptions generated by SQL Server. 

Taking a second look at this, I'll plan to keep it as-is for now, since the next step seems to be to add the retry logic. 

I updated these two stories to reflect what we chatted about:

[AB#79925](https://microsofthealth.visualstudio.com/Health/_workitems/edit/79925) (which is SQL specific)
[AB#80638](https://microsofthealth.visualstudio.com/Health/_workitems/edit/80638) (which is for both Cosmos and SQL)

johnstairs (approved):

PR #1710, opened 2021-03-03
Validate SearchParameter Expression

feordin:
## Description
Validation of SearchParameter expression, base and type before committing the parameter to the server

## Related issues
Addresses [issue [AB#76756](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/76756)].

## Testing
E2E tests added

## FHIR Team Checklist

Skip, once we complete CustomSearchParams we should set semver to Feature


src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterDefinitionBuilder.cs
```diff
@@ -96,6 +96,37 @@ private static bool ShouldExcludeEntryStu3(string resourceType, string searchPar
             return new BundleWrapper(modelInfoProvider.ToTypedElement(rawResource));
         }
 
+        internal static SearchParameterInfo GetOrCreateSearchParameterInfo(SearchParameterWrapper searchParameter, IDictionary<Uri, SearchParameterInfo> uriDictionary = null)
```
Ivanidzo4ka, line 99: >GetOrCreateSearchParameterInfo [](start = 44, length = 30)

From how it's used, I would break it into two things:
a) something what creates SearchParameterInfo out of SearchParameterWrapper. (If it's possbile I would just add constructor to SearchParameterInfo which accepts wrapper
b) private function, or maybe just don't have function at all.. 
```
uriDictionary.TryGetValue(new Uri(searchParameter.Url), out var spi) ? spi : new SearchParameterInfo(searchParameter)
```
doesn't like lot of code to be honest...

src/Microsoft.Health.Fhir.Shared.Client/FhirClient.cs
```diff
@@ -291,6 +291,19 @@ public async Task<FhirResponse<Bundle>> SearchPostAsync(string resourceType, Can
             return await CreateResponseAsync<Bundle>(response);
         }
 
+        public async Task<FhirResponse<Resource>> PostAsync(string resourceType, string body, CancellationToken cancellationToken = default)
```
Ivanidzo4ka, line 294: >PostAsync [](start = 50, length = 9)

```
var fhirSource = Samples.GetJson("Profile-Patient-uscore");
var parser = new FhirJsonParser();
var patient = parser.Parse<Resource>(fhirSource).ToTypedElement().ToResourceElement();
Patient createdResource = await _client.CreateAsync(patient.ToPoco<Patient>());
```
I went this route to post resources to server which are in string format, gives me some guarantee what I don't mess up json file and it's a valid resource before i post it. On downside you have to do double serialization/deserialization

src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterStatusManager.cs
```diff
@@ -103,6 +103,20 @@ public async Task UpdateSearchParameterStatusAsync(IReadOnlyCollection<string> s
                 var searchParamUri = new Uri(uri);
 
                 var paramInfo = _searchParameterDefinitionManager.GetSearchParameter(searchParamUri);
+
+                if (checkSupported)
```
Ivanidzo4ka, line 107: >checkSupported [](start = 20, length = 14)

Correct me if I wrong, but this is called only if we add new search parameter.
Why we can't do that validation during adding search parameter to definition manager?

This thing is status manager, feels weird to force it do validation of something unrelated to status.

src/Microsoft.Health.Fhir.Shared.Client/FhirClient.cs
feordin: I did start with the _client.CreateAsync.  In this case, because I am testing with some invalid resources, I cannot successfully parse them into a resource object before I POST them.  I need to POST as raw JSON.

src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterStatusManager.cs
feordin: You are correct.  This check now occurs in SearchParameterOperations, and is a better fit.

src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterDefinitionBuilder.cs
feordin: Constructor added, I like that change as well.

src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterStatusManager.cs
```diff
@@ -122,8 +123,8 @@ public async Task UpdateSearchParameterStatusAsync(IReadOnlyCollection<string> s
 
         internal async Task AddSearchParameterStatusAsync(IReadOnlyCollection<string> searchParamUris)
         {
-            // new search parameters are added as supported, until reindexing occurs, when
-            // they will be fully enabled
+            // new search parameters are added as supported, however the UpdateSearchParameterStatus will
+            // check the search parameter to ensure the fhirpath is valid and supported
```
Ivanidzo4ka, line 127: nit: should we just revert this file? I don't think this new comment is right anymore.

Ivanidzo4ka (approved):
:shipit:

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/SearchParameters/SearchParameterDefinitionManagerTests.cs
```diff
@@ -282,11 +282,15 @@ public async Task GivenASearchParameterDefinitionManager_WhenGettingSearchParame
...
                 Base = new List<ResourceType?>() { ResourceType.Patient },
-                Expression = "expression",
+                Expression = "Patient.Name",
                 Name = "test",
                 Code = "test",
             };
 
+            _searchParameterSupportResolver
+                .IsSearchParameterSupported(Arg.Is<SearchParameterInfo>(p => p.Name == "test"))
+                .Returns((true, false));
```
Ivanidzo4ka, line 292: >true [](start = 26, length = 4)

Any chance you want to have tests where searchParameter is unsupported, or you prefer to cover it in e2e tests?

src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterStatusManager.cs
feordin: Yep!  I think it is back to way it was.

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/SearchParameters/SearchParameterDefinitionManagerTests.cs
feordin: As I am reusing the Parse and IsSupported logic that was already existing, just calling it in a new place, I added the E2E tests.  I think that is sufficient.

PR #1716, opened 2021-03-04
Gracefully handle Cosmos DB throttling on startup

johnstairs:
## Description
Improves the handling of Cosmos DB throttling when the service starts up. 
Adding retries, avoiding logging RequestRateExceededExceptions as critical, and handling this exception in ThrottlingMiddleware.

## Testing
Manual testing with an added unit test.

Patch

### DIFF
--- a/src/Microsoft.Health.Fhir.Api/Features/Throttling/ThrottlingMiddleware.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/Throttling/ThrottlingMiddleware.cs
@@ -292,10 +292,7 @@ namespace Microsoft.Health.Fhir.Api.Features.Throttling
 
             context.Response.StatusCode = StatusCodes.Status429TooManyRequests;
 
-            if (exception.RetryAfter.HasValue)
-            {
-                context.Response.Headers.AddRetryAfterHeaders(exception.RetryAfter.Value);
-            }
+            context.Response.Headers.AddRetryAfterHeaders(exception.RetryAfter);
 
             Memory<byte> body = CreateThrottledBody(exception.Message);
 
--- a/src/Microsoft.Health.Fhir.Api/Features/Throttling/ThrottlingMiddleware.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/Throttling/ThrottlingMiddleware.cs
@@ -13,6 +13,7 @@ using EnsureThat;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
+using Microsoft.Health.Abstractions.Exceptions;
 using Microsoft.Health.Fhir.Api.Configs;
 using Microsoft.Health.Fhir.Api.Features.Headers;
 using Microsoft.Health.Fhir.Core.Configs;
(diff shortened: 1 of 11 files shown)


src/Microsoft.Health.Fhir.Api/Features/Throttling/ThrottlingMiddleware.cs
```diff
@@ -267,6 +286,27 @@ private async Task Return429(HttpContext context)
...
 
+        private async Task Return429FromRequestRateExceededException(RequestRateExceededException exception, HttpContext context)
+        {
+            _logger.LogWarning($"Returning 429 from unhandled {nameof(RequestRateExceededException)}");
+
+            context.Response.StatusCode = StatusCodes.Status429TooManyRequests;
+
+            if (exception.RetryAfter.HasValue)
+            {
+                context.Response.Headers.AddRetryAfterHeaders(exception.RetryAfter.Value);
```
LTA-Thinking, line 297: If the exception doesn't have a retry after value, should we add one for consistency?

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CollectionInitializer.cs
```diff
@@ -3,68 +3,104 @@
...
-                    _collectionId,
-                    $"/{KnownDocumentProperties.PartitionKey}",
-                    _initialCollectionThroughput);
+                    existingContainer = await containerClient.ReplaceContainerAsync(containerResponse);
 
-                containerResponse.Resource.DefaultTimeToLive = -1;
+                    if (_cosmosCollectionConfiguration.InitialCollectionThroughput.HasValue)
+                    {
+                        ThroughputProperties throughputProperties = ThroughputProperties.CreateManualThroughput(_cosmosCollectionConfiguration.InitialCollectionThroughput.Value);
+                        await containerClient.ReplaceThroughputAsync(throughputProperties);
```
brendankowitz, line 80: Do you think this entire block should be retried or each individual call?
Assuming something fails it would start again be reading the collection, if it previously failed on `ReplaceThroughputAsync`, this setting might not be applied. Thoughts?

johnstairs: I guess putting retries around each statement lowers the chances that the throughput is not applied...

src/Microsoft.Health.Fhir.Api/Features/Throttling/ThrottlingMiddleware.cs
johnstairs: I guess we can...

brendankowitz (approved):

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CollectionInitializer.cs
```diff
@@ -3,64 +3,94 @@
...
-                    _initialCollectionThroughput);
+                    _cosmosCollectionConfiguration.InitialCollectionThroughput));
 
-                containerResponse.Resource.DefaultTimeToLive = -1;
+            containerResponse.Resource.DefaultTimeToLive = -1;
 
-                existingContainer = await containerClient.ReplaceContainerAsync(containerResponse);
+            existingContainer = await retryPolicy.ExecuteAsync(async () => await containerClient.ReplaceContainerAsync(containerResponse));
+
+            if (_cosmosCollectionConfiguration.InitialCollectionThroughput.HasValue)
```
brendankowitz, line 75: @johnstairs, sorry I missed this. We can only do this if the collection is new, else we risk overriding the existing settings. This is currently crashing CI as they are set to Autoscale

PR #1730, opened 2021-03-09
Update datetime param table schema

namadabu:
## Description
Update the schema by for dbo.DateTimeSearchParam table to add two bit columns - IsMin and IsMax. This is similar to #1579 which was for the StringSearchParam table.

I am reusing the schema version 8 since this is being merged into a feature branch for now.

## Related issues
Addresses [AB#80140](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/80140)

## Testing
Existing tests. Tested E2E flow on local deployment.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/8.diff.sql
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/8.diff.sql
@@ -17,8 +17,8 @@ CREATE TYPE dbo.StringSearchParamTableType_2 AS TABLE
 ALTER TABLE dbo.StringSearchParam
 ADD IsMin bit NOT NULL,
     IsMax bit NOT NULL,
-    CONSTRAINT IsMin_Constraint DEFAULT 0 FOR IsMin,
-    CONSTRAINT IsMax_Constraint DEFAULT 0 FOR IsMax;
+    CONSTRAINT string_IsMin_Constraint DEFAULT 0 FOR IsMin,
+    CONSTRAINT string_IsMax_Constraint DEFAULT 0 FOR IsMax;
 
 CREATE NONCLUSTERED INDEX IX_StringSearchParam_SearchParamId_Text
 ON dbo.StringSearchParam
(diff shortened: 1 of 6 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/8.diff.sql
```diff
@@ -17,8 +17,8 @@ CREATE TYPE dbo.StringSearchParamTableType_2 AS TABLE
 ALTER TABLE dbo.StringSearchParam
 ADD IsMin bit NOT NULL,
     IsMax bit NOT NULL,
-    CONSTRAINT IsMin_Constraint DEFAULT 0 FOR IsMin,
-    CONSTRAINT IsMax_Constraint DEFAULT 0 FOR IsMax;
+    CONSTRAINT string_IsMin_Constraint DEFAULT 0 FOR IsMin,
```
Ivanidzo4ka, line 20: >string_IsMin_Constraint [](start = 15, length = 23)

Shouldn't these changes be part of 8.sql as well?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/8.diff.sql
```diff
@@ -61,6 +61,105 @@ WITH
...
+    EndDateTime datetimeoffset(7) NOT NULL,
+    IsLongerThanADay bit NOT NULL,
+    IsMin bit NOT NULL,
+    IsMax bit NOT NULL
+);
+
+ALTER TABLE dbo.DateTimeSearchParam
+ADD IsMin bit NOT NULL,
+    IsMax bit NOT NULL,
+    CONSTRAINT date_IsMin_Constraint DEFAULT 0 FOR IsMin,
```
Ivanidzo4ka, line 77: > CONSTRAINT date_IsMin_Constraint DEFAULT 0 FOR IsMin, [](start = 3, length = 54)

I've tried to find what Default does for column, but I don't see that option in docs: [https://docs.microsoft.com/en-us/sql/t-sql/statements/alter-table-column-constraint-transact-sql?view=sql-server-ver15](https://docs.microsoft.com/en-us/sql/t-sql/statements/alter-table-column-constraint-transact-sql?view=sql-server-ver15)

Can you comment what it does and in case this to setup default value for added column why not: `ADD IsMin bit NOT NULL DEFAULT(0)`

namadabu: I did attempt that approach as well and there is no difference in functionality. In both scenarios, the end result is that we create a `DefaultConstraint` object on the table. The advantage of specifically adding a `CONSTRAINT` is that we can set the name. This makes it cleaner to test (we can specifically exclude the named constraint in our schema difference test - `GivenTwoSchemaInitializationMethods_WhenCreatingTwoDatabases_BothSchemasShouldBeEquivalent()`). If we use `DEFAULT(0)`, the constraint doesn't have a name.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/8.diff.sql
namadabu: We need to only set the default here because we are adding two new non-null columns to an existing table/database. For brand new databases, we don't want to set the default value since the code will take care of that part. Hence the constraints are not added for 8.sql and only for 8.diff.sql

Ivanidzo4ka (approved):
:shipit:

Issue #1050, opened 2020-06-02
Allow client to specify how unsupported search parameters are handled


brendankowitz:
**User story**
As a User I'd like to be able to specify the "Prefer: handling" header so that I can receive an error when I use an unsupported search parameter

**Acceptance criteria**
1. When I specify "handling=strict" and use a search parameter that is not supported by the server than an error is thrown
2. When I specify "handling=lenient" and use a search parameter that is not supported by the server, it is dropped from the self link and no error is reported
3. When I specify no perfer: handling header and use a search parameter that is not supported by the server, it is dropped from the self link and no error is reported

**Reference**
- https://www.hl7.org/fhir/search.html#errors
- https://github.com/microsoft/fhir-server/blob/main/src/Microsoft.Health.Fhir.Core/Features/Context/FhirRequestContext.cs#L64

brendankowitz:
Related to ab#73598

Issue #612, opened 2019-08-13
Search results should include warnings for ignored parameters


johnstairs:
**User story**

As a client, I want to clearly see which search parameters were ignored in my request.

For example, when searching using an invalid search parameter in Vonk, cush as `https://vonk.fire.ly/Observation?foo=bar`, you get a helpful `OperationOutcome` in the search results

```json
        {
            "fullUrl": "https://vonk.fire.ly/OperationOutcome/d5c6d473-b1ef-4692-8b19-709c8bdc3b83",
            "search": {
                "mode": "outcome"
            },
            "resource": {
                "resourceType": "OperationOutcome",
                "id": "d5c6d473-b1ef-4692-8b19-709c8bdc3b83",
                "meta": {
                    "versionId": "4d208d26-49cc-453d-bf07-94e38c817ee1",
                    "lastUpdated": "2019-08-13T19:55:21.916+00:00"
                },
                "issue": [
                    {
                        "severity": "warning",
                        "code": "not-supported",
                        "details": {
                            "coding": [
                                {
                                    "system": "http://hl7.org/fhir/dotnet-api-operation-outcome",
                                    "code": "5003"
                                }
                            ],
                            "text": "Parameter foo is not implemented"
                        }
                    }
                ]
            }
        }
```

This is in addition to the self and next links omitting the invalid parameter (as we do as well).

CaitlinV39:
AB#73598

@CaitlinV39 update description in DevOps

PR #1731, opened 2021-03-09
"Handling" parameter in search

Ivanidzo4ka:
## Description
Adds support for Prefer header with handling= parameter.
Throw exception if handling set to strict, adds operation outcome if there is unsupported parameters to search bundle.

For strict:
image
OperationOutcome in bundle:
image

## Related issues
Addresses [#1050 ]
Addresses [#612 ].

## Testing
Manual and unit tests

## FHIR Team Checklist

Patch

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/KnownQueryParameterNames.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/KnownQueryParameterNames.cs
@@ -46,6 +46,8 @@ namespace Microsoft.Health.Fhir.Core.Features
 
         public const string Elements = "_elements";
 
+        public const string Handling = "handling";
+
         /// <summary>
         /// The total query parameter.
         /// </summary>
--- a/src/Microsoft.Health.Fhir.Core/Features/KnownQueryParameterNames.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/KnownQueryParameterNames.cs
@@ -46,8 +46,6 @@ namespace Microsoft.Health.Fhir.Core.Features
 
         public const string Elements = "_elements";
 
-        public const string Handling = "handling";
-
         /// <summary>
         /// The total query parameter.
         /// </summary>
--- a/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/BasicSearchTests.cs
+++ b/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/BasicSearchTests.cs
@@ -811,15 +811,22 @@ namespace Microsoft.Health.Fhir.Tests.E2E.Rest.Search
         public async Task GivenASearchRequestWithInvalidParametersAndStrictHandling_WhenHandled_ReturnsBadRequest()
         {
             using FhirException ex = await Assert.ThrowsAsync<FhirException>(() =>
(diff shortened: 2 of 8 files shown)


src/Microsoft.Health.Fhir.Core/Features/KnownQueryParameterNames.cs
```diff
@@ -46,6 +46,8 @@ public static class KnownQueryParameterNames
 
         public const string Elements = "_elements";
 
+        public const string Handling = "handling";
```
Ivanidzo4ka, line 49: >Handling [](start = 28, length = 8)

this should be header -_- not query parameter

feordin (approved):
Just a question.  When we have an invalid search parameter, and it is set to Lenient handling, do we still remove the searchparameters from the self link that is returned for the Bundle? or do we leave them?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/BasicSearchTests.cs
```diff
@@ -775,6 +775,54 @@ public async Task GivenASearchRequestWithInvalidValues_WhenHandled_ReturnsABadRe
...
+            ValidateOperationOutcome(expectedDiagnostics, expectedIssueSeverities, expectedCodeTypes, outcome);
+        }
+
+        [Fact]
+        public async Task GivenASearchRequestWithInvalidParametersAndStrictHandling_WhenHandled_ReturnsBadRequest()
+        {
+            using FhirException ex = await Assert.ThrowsAsync<FhirException>(() =>
+            Client.SearchAsync("/Patient?Cookie=Chip&Ramen=Spicy", Tuple.Create(KnownHeaders.Prefer, "handling=strict")));
+            Assert.Equal(HttpStatusCode.BadRequest, ex.StatusCode);
+        }
```
LTA-Thinking, line 816: Can we add a test that sets the strict header and has valid search parameters?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/CanonicalSearchTests.cs
```diff
@@ -49,7 +49,7 @@ public async Task GivenAnObservationWithProfile_WhenSearchingByCanonicalUri_Then
         {
             Skip.IfNot(Fixture.TestFhirServer.Metadata.SupportsSearchParameter("Observation", "_profile"), _skipReason);
 
-            FhirResponse<Bundle> result = await Fixture.TestFhirClient.SearchAsync($"Observation?_profile={Fixture.ObservationProfileUri}&_order=_lastModified");
+            FhirResponse<Bundle> result = await Fixture.TestFhirClient.SearchAsync($"Observation?_profile={Fixture.ObservationProfileUri}");
```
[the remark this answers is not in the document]
Ivanidzo4ka: There is no such thing as `_order` in Observation resource and with my changes it start returning back OperationOutcome in bundle and that broke the test. I don't know what this `_order` suppose to mean, and it's not supported, so I just removing it from test. Maybe intent was to do sort, but I don't know this, and how well sort supported.

LTA-Thinking (approved):

Issue #1681, opened 2021-02-19
Doc generation throw error


Ivanidzo4ka:
**Describe the bug**
If I put <GenerateDocumentationFile>true</GenerateDocumentationFile> into https://github.com/microsoft/fhir-server/blob/main/Directory.Build.props build would start throw errors regarding absence of documentation.

**Expected behavior**
Build works fine.

CaitlinV39:
AB#79888

EXPEkesheth:
Closing the issue

PR #1732, opened 2021-03-09
Added comments to one of public class - CodableConceptInfo

gauravgupta98:
## Added comments to CodableConceptInfo class.
I think we should add comments to our public methods/classes etc.
For now, I've just added comments to single class. If we are good to go with this proposed change then I can go ahead and pick other public facing methods and improve the coding standards.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Models/CodableConceptInfo.cs
+++ b/src/Microsoft.Health.Fhir.Core/Models/CodableConceptInfo.cs
@@ -32,7 +32,7 @@ namespace Microsoft.Health.Fhir.Core.Models
         }
 
         /// <summary>
-        /// Gets the collection of Coding.
+        /// Gets the Coding collection.
         /// </summary>
         public ICollection<Coding> Coding { get; }
     }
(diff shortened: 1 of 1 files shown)


brendankowitz:
Hi @gauravgupta98, yes I think this makes sense. The more urgent task relating to docs is fixing the existing errors that are blocking generation: #1681

gauravgupta98:
Hi @brendankowitz, thanks for the review. I'll have a look at the issue you mentioned. Also, If we are good on this change, can we go ahead and complete the PR?

src/Microsoft.Health.Fhir.Core/Models/CodableConceptInfo.cs
```diff
@@ -12,18 +12,28 @@ namespace Microsoft.Health.Fhir.Core.Models
...
+        /// <param name="coding">The Coding collection.</param>
         public CodableConceptInfo(IEnumerable<Coding> coding)
         {
             EnsureArg.IsNotNull(coding);
 
             Coding = coding.ToList();
         }
 
+        /// <summary>
+        /// Gets the collection of Coding.
```
brendankowitz, line 35: nit: it think this reads better as `Gets the Coding collection` or `Gets the collection of Codings`

gauravgupta98: Shouldn't the property name be Codings? What are your thoughts?

brendankowitz (approved):

gauravgupta98:
Looks like some tests are failing. @brendankowitz can you please have a look. Thanks

Issue #1256, opened 2020-09-10
When using the _elements parameter mandatory fields aren't returned


LTA-Thinking:
**Describe the bug**
When a search is performed with the _elements parameter mandatory fields are not returned unless specified.

**To Reproduce**
1. Run the search ```/Observation?_elements=id```
2. Only ids are returned, but code and status are mandatory fields of Observation

**Expected behavior**
According to the spec "Clients SHOULD list all mandatory and modifier elements in a resource as part of the list of elements." [spec](https://www.hl7.org/fhir/search.html#elements)

**Actual behavior**
Only fields that are specified in the _elements parameter are returned.

brendankowitz:
And also `Servers SHOULD mark the resources with the tag SUBSETTED to ensure that the incomplete resource is not actually used to overwrite a complete resource.`

anyname2:
1. This is (probably) a bug in https://github.com/FirelyTeam/fhir-net-api
    FHIR Server only calls the API: [FhirJsonOutputFormatter.cs#L133](https://github.com/microsoft/fhir-server/blob/master/src/Microsoft.Health.Fhir.Shared.Api/Features/Formatters/FhirJsonOutputFormatter.cs#L133)
    ```csharp
    _fhirJsonSerializer.Serialize(resource, jsonWriter, context.HttpContext.GetSummaryType(_logger), elementsSearchParameter);
    ```
2. > And also `Servers SHOULD mark the resources with the tag SUBSETTED to ensure that the incomplete resource is not actually used to overwrite a complete resource.`

    This is already done when the meta is present, but you need to have `_elements=meta` so that it is not excluded from the result (i.e. should be resolved when 1 is fixed).

CaitlinV39:
I'm going to open a bug on the Firely side and then track from there.

Ivanidzo4ka:
https://github.com/FirelyTeam/firely-net-sdk/issues/1653
It doesn't count as a bug.

I've created PR to add required fields to resource (and meta field)

CaitlinV39:
AB#81256

PR #1741, opened 2021-03-12
Show mandatory fields with _elements

Ivanidzo4ka:
## Description
Currently _elements filter out all fields even required one, (and meta tag which should have information what this resource is truncated) So walk through resources to find what elements we need to show and add them to serializer.

## Related issues
Addresses [#1256].

## Testing
Manually and unit tests

## FHIR Team Checklist

Patch


src/Microsoft.Health.Fhir.Shared.Api/Features/Formatters/FhirJsonOutputFormatter.cs
```diff
@@ -74,7 +76,10 @@ public override async Task WriteResponseBodyAsync(OutputFormatterWriteContext co
             var elementsSearchParameter = context.HttpContext.GetElementsOrDefault();
             var summarySearchParameter = context.HttpContext.GetSummaryTypeOrDefault();
             var pretty = context.HttpContext.GetPrettyOrDefault();
+            var hasElements = elementsSearchParameter?.Any() == true;
             Resource resource = null;
+            IStructureDefinitionSummaryProvider summaryProvider = new PocoStructureDefinitionSummaryProvider();
```
brendankowitz, line 81: There is an instance of this on `IModelInfoProvider.StructureDefinitionSummaryProvider`

src/Microsoft.Health.Fhir.Shared.Api/Features/Formatters/FhirJsonOutputFormatter.cs
```diff
@@ -119,6 +136,12 @@ public override async Task WriteResponseBodyAsync(OutputFormatterWriteContext co
                 resource = (Resource)context.Object;
```
brendankowitz, line 136: This finding the required elements also happen here?

LTA-Thinking:
How should elements and profiles interact? Profiles can add more required fields, which I don't think will be picked up here. The profile will be included in the elements summary since it is part of the meta property, but not all of the profile's required fields will be guarantied to be present.

LTA-Thinking:
From [this](https://www.hl7.org/fhir/search.html#elements) line of the spec: 
>Servers SHOULD mark the resources with the tag SUBSETTED to ensure that the incomplete resource is not actually used to overwrite a complete resource.
Should we have a check as part of Create/Update to block resources with the SUBSETTED tag from being persisted? Probably out of scope for this story, but it came to mind.

src/Microsoft.Health.Fhir.Shared.Api/Features/Formatters/FhirJsonOutputFormatter.cs
```diff
@@ -91,6 +96,12 @@ public override async Task WriteResponseBodyAsync(OutputFormatterWriteContext co
                         if (rawBundleEntryComponent is RawBundleEntryComponent)
                         {
                             rawBundleEntryComponent.Resource = ((RawBundleEntryComponent)rawBundleEntryComponent).ResourceElement.ToPoco<Resource>(_deserializer);
+                            if (hasElements)
```
LTA-Thinking, line 99: Nit: Could the if statements on lines 89 and 116 use hasElements instead of checking elementsSearchParameter?

Ivanidzo4ka:
So there is default profile and supported profiles. Default profile for resource stated in our capability statement is "[http://hl7.org/fhir/StructureDefinition/{resourceType}](http://hl7.org/fhir/StructureDefinition/{resourceType})". Supported profiles are not stated yet, but I plan to add them. I think we should play with default profile and display only mandatory fields for it, it can be weird if you start support supported one, because one resource can conform to multiple profiles and they can require different field.  Since default profile is basically definition of resource, I think it's ok to do what I'm doing here.

---
In reply to: [797604400](https://github.com/microsoft/fhir-server/pull/1741#issuecomment-797604400) [](ancestors = 797604400)

Ivanidzo4ka:
I somewhat agree, but again, spec doesn't have anything in create/update section about this.
On one hand it sounds right, we know this resource is subsetted, but in same time, does it mean we shouldn't accept it? What if this is all customer has right now, and they can update that resource later? Yeah, it's not full information, but it's all required information and it can be still useful. So by introducing this, we can break peoples pipelines.

We can produce warning in output, but if you don't clean your meta tag, what's the chance you gonna read your warning?

More I read, more it sound for me what purpose of this tag is informational, rather than be some kind of policy.

Can be wrong, tho, FHIR spec is still mystery for me, maybe it's worth to follow up on Zulip.

---
In reply to: [797605910](https://github.com/microsoft/fhir-server/pull/1741#issuecomment-797605910) [](ancestors = 797605910)

brendankowitz (approved):

LTA-Thinking (approved):

LTA-Thinking:
Created Zulip [conversation](https://chat.fhir.org/#narrow/stream/179166-implementers/topic/SUBSETTED.20tags.20and.20create.2Fupdate/near/231336652) for the create/update and subsetted question.

Issue #1231, opened 2020-08-27
Support for FHIR Release #5: Preview #3  v4.5.0


uxbridgeroad:
User story
As a user I want my R4/R5 FHIR API to conform to the v4.5.0: Release 5 Preview #3 specification.

http://hl7.org/fhir/2020Sept/index.html  

Acceptance criteria

When I provision a FHIR API, then the CapabilityStatement should say FHIR version 4.5.0.

CaitlinV39:
@uxbridgeroad we will update this as soon as the FHIR-NET-API nuget packages are available.

uxbridgeroad:
@CaitlinV39 , just to confirm that this is one of the issue we were discussing earlier today with Marc and yourself.
We are planning for our next couple of sprints and it would be great to be able to align with an expected completion date for this.

CaitlinV39:
@uxbridgeroad thank you so much for mentioning this. We are working on implementing this now in this PR #1743 We should have this merged in in the next couple of weeks, if not sooner.

uxbridgeroad:
@CaitlinV39, I just want to clarify what happens once #1743  is merged.

1. Will FHIR Release #5: Preview #3 v4.5.0 be the default when R5 is chosen during the ARM deployment step?
2. The Microsoft.Health.Fhir.R5.* packages on main will only support R5 Preview #3 v4.5.0.?
3. Any further fixes to Microsoft.Health.Fhir.R5.* will not be available on R5  Preview #2?

Thanks,

PR #1743, opened 2021-03-12
Updates to fhir-net-api-3

brendankowitz:
## Description
Updates Fhir-net-api to v3.
Requires: https://github.com/microsoft/FHIR-Tools-for-Anonymization/pull/97

## Questions / issues:
- [x] [Date serialization issue](https://github.com/microsoft/fhir-server/pull/1743#issuecomment-797827925)
- [x] ["System.Boolean" being returned as FHIR Instance type](https://github.com/microsoft/fhir-server/pull/1743/files#r593466675)

## Related issues
Addresses [AB#79420](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/79420).

## Testing

## References:
https://github.com/FirelyTeam/firely-net-sdk/wiki/Breaking-changes-in-3.0
https://github.com/FirelyTeam/firely-net-sdk/wiki/Breaking-changes-in-2.0

## FHIR Team Checklist


Ivanidzo4ka:
Probably too soon for this PR, but we can get rid of `OperationOutcomeIssue` class since it should be part of common library.

src/Microsoft.Health.Fhir.Core/Features/Conformance/Serialization/CodingJsonConverter.cs
```diff
@@ -0,0 +1,66 @@
...
+using EnsureThat;
+using Hl7.Fhir.Model;
+using Newtonsoft.Json;
+
+namespace Microsoft.Health.Fhir.Core.Features.Conformance.Serialization
+{
+    /// <summary>
+    /// Converts <see cref="Coding"/> into a simple serializable model for <see cref="CapabilityStatementBuilder"/>
+    /// </summary>
+    internal class CodingJsonConverter : JsonConverter
```
Ivanidzo4ka, line 16: >CodingJsonConverter [](start = 19, length = 19)

just curious, what it show up in json if you don't use this converter?

src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterDefinitionBuilder.cs
```diff
@@ -113,22 +113,29 @@ private static SearchParameterInfo GetOrCreateSearchParameterInfo(SearchParamete
...
+                try
+                {
+                    return new SearchParameterWrapper(x);
+                }
+                catch (ArgumentException)
+                {
+                    AddIssue(Core.Resources.SearchParameterDefinitionInvalidResource, entryIndex);
+                    return null;
+                }
+            }).ToList();
```
Ivanidzo4ka, line 127: >ToList [](start = 15, length = 6)

I would filter .Where(x=>x!=null) to simplify code below.

brendankowitz: I think I also initially tried this, the parsing code below respecting the index of the entry in the Bundle. Null is only used when there is an entry in the Bundle that isn't a SearchParameter resource. I made this change to to avoid parsing into SearchParameterWrapper multiple times

src/Microsoft.Health.Fhir.Core/Features/Search/Converters/BooleanNodeToTokenSearchValueTypeConverter.cs
```diff
@@ -16,7 +16,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Converters
     public class BooleanNodeToTokenSearchValueTypeConverter : FhirNodeToSearchValueTypeConverter<TokenSearchValue>
     {
         public BooleanNodeToTokenSearchValueTypeConverter()
-            : base("boolean")
+            : base("boolean", "System.Boolean")
```
brendankowitz, line 19: TODO: Check other converter types

src/Microsoft.Health.Fhir.Core/Features/Conformance/Serialization/CodingJsonConverter.cs
brendankowitz: The `CodingElement` fields are serialized (https://github.com/FirelyTeam/firely-net-common/blob/develop/src/Hl7.Fhir.Support.Poco/Model/Generated/Coding.cs)

brendankowitz:
> Probably too soon for this PR, but we can get rid of OperationOutcomeIssue class since it should be part of common library.
@Ivanidzo4ka, yes, this should open up a lot of improvements, refactorings and simplifications we can make :)

src/Microsoft.Health.Fhir.Core/Features/Search/Converters/BooleanNodeToTokenSearchValueTypeConverter.cs
brendankowitz: `System.Boolean` was returned when evaluating `Patient.deceased.exists() and Patient.deceased != false`. This previously only returned `boolean` as the ITypedElement.InstanceType

anyname2:
I am not sure if stricter validation introduced in v2 will affect this PR or not, so just a heads up.

In v1.9.0 the below patient would be considered valid (even though it is not according to spec), but v2 will throw an error parsing this:
```json
{
  "resourceType": "Patient",
  "birthDate": "1991-02-03T11:22:33Z"
}
```

This could have an implication, that FHIR server running with v1 could have resources in the databse which it could fail to parse after this PR is merged (which would manifest as quite random errors during search).

CaitlinV39:
@brendankowitz I believe that this addresses issue #1231 Let me know if it doesn't fully address.

src/Microsoft.Health.Fhir.Core/Features/Conformance/Serialization/CodingJsonConverter.cs
Ivanidzo4ka: I feel silly. I've been calling metadata, but it's not been writing at all (probably because only coding in metadata is messaging.endpoint.protocol and we don't output it).

What spark it creation if never hit it? Some experemental branch?

brendankowitz: Do you have auth turned on?

Ivanidzo4ka: 🤦

src/Microsoft.Health.Fhir.Shared.Api/Modules/FhirModule.cs
```diff
@@ -37,13 +38,7 @@ namespace Microsoft.Health.Fhir.Api.Modules
...
     public class FhirModule : IStartupModule
     {
-        private readonly FeatureConfiguration _featureConfiguration;
-
-        public FhirModule(FhirServerConfiguration fhirServerConfiguration)
-        {
-            EnsureArg.IsNotNull(fhirServerConfiguration, nameof(fhirServerConfiguration));
-            _featureConfiguration = fhirServerConfiguration.Features;
-        }
+        private static readonly Regex MessageChecker = new Regex("Type checking the data: Literal '(.*)' cannot be parsed as a (.*). \\(at (.*)\\)", RegexOptions.Compiled);
```
brendankowitz, line 41: nit: Capture groups might look cleaner
```suggestion
        private static readonly Regex MessageChecker = new Regex("Type checking the data: Literal '(?<value>.*)' cannot be parsed as a (?<type>.*). \\(at (?<location>.*)\\)", RegexOptions.Compiled);
```

src/Microsoft.Health.Fhir.Shared.Tests/Deserializers.cs
```diff
@@ -16,11 +18,51 @@ namespace Microsoft.Health.Fhir.Tests.Common
...
+                    }
+                    catch (StructuralTypeException ex)
+                    {
+                        var match = MessageChecker.Match(ex.Message);
+                        if (match.Success && match.Groups.Count == 4 && match.Groups[2].Value == "date")
+                        {
+                            var valueToReplace = match.Groups[1].Value;
+                            var location = match.Groups[3].Value;
+                            var replace = valueToReplace.Substring(0, 10);
+                            var root = FhirJsonNode.Parse(str, "Resource");
```
brendankowitz, line 41: nit:
```suggestion
                            var root = FhirJsonNode.Parse(str, KnownResourceTypes.Resource);
```

src/Microsoft.Health.Fhir.Shared.Api/Modules/FhirModule.cs
```diff
@@ -79,9 +74,47 @@ ResourceElement SetMetadata(Resource resource, string versionId, DateTimeOffset
...
+
+                                        if (currentNode.Location == location)
+                                        {
+                                            break;
+                                        }
+                                    }
+
+                                    (currentNode as FhirJsonNode).JsonValue.Value = replace;
+                                    str = root.ToJson();
+                                    goto Parse;
```
brendankowitz, line 113: lol Goto. I think I'd prefer something like:
```
try{
Parse()
}
catch{
ApplyBackwardsCompatabilityFixes()
Parse()
// if parse fails again, fail.
}

```

src/Microsoft.Health.Fhir.Shared.Api/Modules/FhirModule.cs
```diff
@@ -79,9 +74,47 @@ ResourceElement SetMetadata(Resource resource, string versionId, DateTimeOffset
...
+                                        }
+
+                                        if (currentNode.Location == location)
+                                        {
+                                            break;
+                                        }
+                                    }
+
+                                    (currentNode as FhirJsonNode).JsonValue.Value = replace;
+                                    str = root.ToJson();
```
brendankowitz, line 112: From here could we use `root.ToTypedElement(_modelInfoProvider.StructureDefinitionSummaryProvider).ToPoco<Resource>()`?

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/SearchParameters/SearchParameterToTypeResolverTests.cs
```diff
@@ -18,22 +18,27 @@ public class SearchParameterToTypeResolverTests
...
 
-            if (ModelInfoProvider.Version == FhirSpecification.Stu3)
-            {
-                Assert.Equal(38, results?.Length);
-            }
-            else
-            {
-                Assert.Equal(49, results?.Length);
-            }
+            Assert.Equal("DataType", results.Single().FhirNodeType);
```
brendankowitz, line 29: @Ivanidzo4ka @feordin test how/if this changes the SearchParameter support/detection when posting in an extension like this

src/Microsoft.Health.Fhir.Shared.Api/Modules/FhirModule.cs
Ivanidzo4ka: Parse gives me one exception at time.
I can have multiple date fields in resource (see Evidence for example)

So I need to parse until I get rid of errors, or I encounter error which is not related to date parsing.

I can probably wrap in it do while loop...

Wonder how goto handles stacktrace...

src/Microsoft.Health.Fhir.Shared.Api/Modules/FhirModule.cs
Ivanidzo4ka: I don't know what it does.
What it achive?

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/SearchParameters/SearchParameterToTypeResolverTests.cs
Ivanidzo4ka: yeah, it's broken.

Looking into it...

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/SearchParameters/SearchParameterToTypeResolverTests.cs
```diff
@@ -18,22 +18,51 @@ public class SearchParameterToTypeResolverTests
...
             }
             else
             {
-                Assert.Equal(49, results?.Length);
+                Assert.Equal(62, types.Length);
             }
         }
+
+        [Fact]
+        public void BlaBla()
```
feordin, line 45: Test name?

Ivanidzo4ka: 🤦

src/Microsoft.Health.Fhir.Shared.Api/Modules/FhirModule.cs
```diff
@@ -80,8 +75,48 @@ ResourceElement SetMetadata(Resource resource, string versionId, DateTimeOffset
...
+                                                break;
+                                            }
+                                        }
+
+                                        (currentNode as FhirJsonNode).JsonValue.Value = replace;
+                                        str = root.ToJson();
+                                    }
+                                }
+                            }
+                            while (!parsed);
```
brendankowitz, line 119: Final checkpoint, is there a chance we'll get into this loop and our modifications don't fix all the errors and it gets stuck?

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/SearchParameters/SearchParameterToTypeResolverTests.cs
```diff
@@ -18,22 +18,51 @@ public class SearchParameterToTypeResolverTests
...
             SearchParameterTypeResult[] results = SearchParameterToTypeResolver.Resolve(
                 KnownResourceTypes.Patient,
                 (SearchParamType.Token, expression, new Uri("http://hl7.org/fhir/SearchParameter/Patient-race")),
                 null).ToArray();
+            var types = results.Select(x => x.FhirNodeType).ToArray();
 
             if (ModelInfoProvider.Version == FhirSpecification.Stu3)
             {
-                Assert.Equal(38, results?.Length);
+                Assert.Equal(51, types.Length);
```
brendankowitz, line 32: Should this be 38 again? I don't think Stu3 has changed

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/SearchParameters/SearchParameterToTypeResolverTests.cs
```diff
@@ -18,22 +18,51 @@ public class SearchParameterToTypeResolverTests
...
+            }
+            else if (ModelInfoProvider.Version == FhirSpecification.R4)
+            {
+                Assert.Equal(59, types.Length);
```
brendankowitz, line 36: Same with R4, should this be 49? What are these extra datatypes?

Ivanidzo4ka: before:
https://github.com/FirelyTeam/firely-net-sdk/blob/f028de2ea5d5b937f4a51f79c1dcbe9887615204/src/Hl7.Fhir.Core/Model/Extension.cs#L87

now:
https://github.com/FirelyTeam/firely-net-common/blob/127c771950e2a5f852607d0b3699cce894ea2126/src/Hl7.Fhir.Support.Poco/Model/Extension.cs#L82

If previously they stated which parameters are ok, now I need to improvise and derive them on a fly.

So I do this:
```
foreach (var pair in ModelInfo.FhirTypeToCsType)
                                {
                                    if ((ModelInfo.IsDataType(pair.Value) || ModelInfo.IsPrimitive(pair.Value))
                                        && !pair.Value.IsAbstract)
```

https://www.hl7.org/fhir/datatypes.html
In the past they also derived General-Purpose Data type from DataType, and rest (with exception of primitive) was derived from element.
Not the case anymore, everything is derived from DataType, so metadata types get leaked into it, special purpose types.
I don't see way to filter them out automatically, and manually doesn't feel right.

image

src/Microsoft.Health.Fhir.Shared.Api/Modules/FhirModule.cs
Ivanidzo4ka: Put count up to 100, and add logging, but I doubt I can log json itself.

Ivanidzo4ka (approved):

brendankowitz:
Added Breaking label for model changes in R5. Added warning for all other versions.

PR #1747, opened 2021-03-13
Code cleanup and comments addition to some files under Core -> Models.

gauravgupta98:
## Added comments to six files under Core -> Models.

Some observations which making the changes -
1. Why don't we have regions in the code? It makes code more structured and readable? (eg. region Properties, region Methods etc.)
2. Can we make sure that any new code must contain comments and add this in our standards?

## FHIR Team Checklist


Ivanidzo4ka:
Thank you for trying to help our repository.

Let me try to answer on questions in PR body first:
>Why don't we have regions in the code? It makes code more structured and readable? (eg. region Properties, region Methods etc.)

As quite often happens, most of the answers we have, is already answered on stackoverflow :) I would suggest to read most popular answer [here](https://softwareengineering.stackexchange.com/questions/53086/are-regions-an-antipattern-or-code-smell ). To add to it, I would say that purpose of #regions is to improve readability of the code, and that's not the best solution. Not every IDE supports them, I don't think viewing code from github supports them, and there is better way how you can solve that problem, for example via StyleCop rules which force you to put properties and methods in grouped manner.

>Can we make sure that any new code must contain comments and add this in our standards?

I think technically it's possible either via StyleCop rules like this [one](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1600.md) or by implementing your own analyzer. But why?

Main purpose of comments is to bring clarity and make sure code is maintainable and usable. In some cases: like complex operation, or object with state, which changes behavior depending on the state, it's crucial to have comments, otherwise, code can be misused, or you have to spend hours in order to figure out what that code does. Are comments must be present 100% of time? I personally think they don't have to. I would even argue that can be detrimental to code health.

I'm not trying to pick on you, but as example I'll take code out of this PR:
```        /// <summary>
        /// Initializes a new instance of the <see cref="SearchParameterInfo"/> class.
        /// </summary>
        /// <param name="name">The name.</param>
        /// <param name="code">The code.</param>
        /// <param name="searchParamType">The search parameter type.</param>
        /// <param name="url">The url./</param>
        /// <param name="components">The search parameter components info list.</param>
        /// <param name="expression">The expression.</param>
        /// <param name="targetResourceTypes">The list of target resource type.</param>
        /// <param name="baseResourceTypes">The list of base resource types.</param>
        /// <param name="description">The description.</param>
public SearchParameterInfo(
            string name,
            string code,
            SearchParamType searchParamType,
            Uri url = null,
            IReadOnlyList<SearchParameterComponentInfo> components = null,
            string expression = null,
            IReadOnlyList<string> targetResourceTypes = null,
            IReadOnlyList<string> baseResourceTypes = null,
            string description = null)
```
Does any of these comments bring clarity, improves readability and help use this code? 

Let's start with summary. This is a constructor and for everyone who has small experience with C# it should be clear what it's create object of same name. It can be helpful for unexperienced developer who doesn't know much about C# development, though.

Parameter descriptions. If I use constructor in code, I can see parameter names and their types. In many cases names and types quite self explanatory and I don't think there is much need to add comment what `name` is "The name."

In well written code you can deduct how code would behave from just names of methods, parameters and variables in code.

One can argue what adding these comments is bring no harm, and make code more user friendly towards unexperienced developers. 

On which I would argue what it makes code bloated and hard to read, it reduce amount of code people can fit on their screens, and therefore reduce span of attention. It increase entropy, and force developer to blur out parts of screen, which can lead to mistakes, and overall slowdown of working with code.  All that been said is related to people who are proficient in language. 

So in choice between user friendly towards beginner vs causing harm to experienced developers I would prefer to omit comments like that.
 
I'm not saying comments are not needed in general. They needed in case of ambiguity or peculiarity of code, to guide and warn user how to use and not to use code. To help express something we can't express in programming language in a easy way.

But comments for sake of comments are evil. 

If you interested in helping us, I would suggest to check issues marked with [Up-for-grabs](https://github.com/microsoft/fhir-server/labels/Up%20for%20grabs) label, like two on the top, they not require knowledge of FHIR spec.

gauravgupta98:
Hi @Ivanidzo4ka,
Thanks for your detailed review. All your points make sense and I agree with you on this.
I've seen some open source organisations follow that structure so was curious to know why we are not following it.
I'll check the up-for-grabs and will look into contributing in that area.
Thanks

PR #1752, opened 2021-03-17
Draft for Canonical reference.

Ivanidzo4ka:
This is draft, I can add tests and everything else, but I'm a bit confused with spec, and it looks like its a easy change, but if that would be case, I'm sure it would be part of canonical PR.

Is there are hidden stones I don't see? @brendankowitz


src/Microsoft.Health.Fhir.Core/Features/Search/Converters/CanonicalNodeToReferenceSearchValueTypeConverter.cs
```diff
@@ -0,0 +1,48 @@
...
+            }
+
+            // Contained resources will not be searchable.
+            if (reference.StartsWith("#", StringComparison.Ordinal)
+                || reference.StartsWith("urn:", StringComparison.Ordinal))
+            {
+                yield break;
+            }
+
+            yield return _referenceSearchValueParser.Parse(reference);
```
brendankowitz, line 45: Its all the 'implicit' behaviors that go with it. This affects both the indexing and the searching. Canonicals may have 
a version and fragment:
` <QuestionnaireResponse questionnaire="http://fhir.acme.com/Questionnaire/example|1.0#vs1"/>` (https://www.hl7.org/fhir/references.html#canonical-fragments)

But when searching _with_ a canonical, these _should_ be equivalent:
`/Questionnaire?url=http://fhir.acme.com/Questionnaire/example|1.0` 
and `/Questionnaire?url=http://fhir.acme.com/Questionnaire/example&version=1.0`, but only on resources that support the ["business version"](https://www.hl7.org/fhir/resource.html#versions) concept

In references:
`QuestionnaireResponse?questionnaire=http://fhir.acme.com/Questionnaire/example|1.0`
Should return questionnaire responses where `/Questionnaire?url=http://fhir.acme.com/Questionnaire/example&version=1.0`. If no version is present, then the latest version.

This could get more complicated in chained searches that contain canonical references:
`/QuestionnaireResponse?questionnaire.code=my-coding`
or including the original questionnaire with the responses: 
`/QuestionnaireResponse?_include=QuestionnaireResponse.questionnaire` which applies the above rules.

PR #1753, opened 2021-03-18
Adds support for new search parameters with abstract base type

rotodd:
## Description

This PR adds support for new search parameters that have a base type of "Resource" or "DomainResource".

It also improves the performance of reindexing operations that cover multiple resource types by only executing non-count queries when the resource-specific count is greater than 0.

## Related issues
Addresses [AB#79561](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Resolute/Stories/?workitem=79561).

## Testing
- Added a manual test scenario to a new `docs/rest` folder, along with instructions on how to use VS Code's REST Client extension
- Added a new integration test and made minor refactoring changes to existing reindexing tests

## FHIR Team Checklist

Minor (new functionality, performance improvements)


src/Microsoft.Health.Fhir.Core/Features/Operations/Export/Models/ExportFileInfo.cs
```diff
@@ -42,21 +42,21 @@ protected ExportFileInfo()
         [JsonProperty(JobRecordProperties.Sequence)]
         public int Sequence { get; private set; }
 
-        [JsonProperty(JobRecordProperties.Count)]
-        public int Count { get; private set; }
+        [JsonProperty(JobRecordProperties.TotalCount)]
+        public int TotalCount { get; private set; }
```
rotodd, line 46: @namadabu will changing these properties have any implications on the PaaS side of things?

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -105,21 +109,36 @@ public async Task ExecuteAsync(ReindexJobRecord reindexJobRecord, WeakETag weakE
...
-                        }
+                            if (baseResourceType == "Resource")
+                            {
+                                resourceList.UnionWith(_modelInfoProvider.GetResourceTypeNames().ToHashSet());
 
-                        if (param.BaseResourceTypes != null)
-                        {
-                            resourceList.UnionWith(param.BaseResourceTypes);
+                                // We added all possible resource types, so no need to continue
+                                break;
```
Ivanidzo4ka, line 119: >break [](start = 32, length = 5)

It feels brittle to me. Who controls order in `BaseResourceTypes`? We can stop populating `resourceList` if "Resource" show up as first type.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -105,21 +109,36 @@ public async Task ExecuteAsync(ReindexJobRecord reindexJobRecord, WeakETag weakE
...
+                                resourceList.UnionWith(_modelInfoProvider.GetResourceTypeNames().ToHashSet());
+
+                                // Remove types that don't inherit from domain resource
+                                resourceList.Remove("Binary");
+                                resourceList.Remove("Bundle");
+                                resourceList.Remove("Parameters");
+                            }
+                            else
+                            {
+                                resourceList.UnionWith(new HashSet<string> { baseResourceType });
```
Ivanidzo4ka, line 133: >new HashSet<string>  [](start = 55, length = 20)

I would go with `new[]{baseResourceType}` or maybe ``Enumerable.Repeat(baseResourceType,1)`;

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/Models/ReindexJobRecord.cs
```diff
@@ -58,6 +58,9 @@ protected ReindexJobRecord()
 
         public ConcurrentDictionary<ReindexJobQueryStatus, byte> QueryList { get; private set; } = new ConcurrentDictionary<ReindexJobQueryStatus, byte>();
 
+        [JsonProperty(JobRecordProperties.ResourceCount)]
+        public ConcurrentDictionary<string, int> ResourceCount { get; private set; } = new ConcurrentDictionary<string, int>();
```
Ivanidzo4ka, line 62: >ResourceCount [](start = 49, length = 13)

This is collection, something in this name should be plural :) `ResourceCounts`?

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -134,13 +153,16 @@ public async Task ExecuteAsync(ReindexJobRecord reindexJobRecord, WeakETag weakE
                     // Generate separate queries for each resource type and add them to query list.
                     foreach (string resourceType in _reindexJobRecord.Resources)
                     {
-                        var query = new ReindexJobQueryStatus(resourceType, continuationToken: null)
+                        if (_reindexJobRecord.ResourceCount[resourceType] > 0)
```
Ivanidzo4ka, line 156: >ResourceCount [](start = 46, length = 13)

Situation: We have lot of Patients, and no Organizations in DB. We start doing re-indexing and it takes day to go through all patients.
Somewhere in the middle of the day few organizations been posted.
Do we care about them?

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Features/Operations/Reindex/ReindexJobTests.cs
```diff
@@ -355,35 +438,34 @@ private async Task<CreateReindexResponse> SetUpForReindexing()
...
+        private async Task<SearchParameter> CreateSearchParam(string searchParamName, SearchParamType searchParamType, ResourceType baseType, string expression, string searchParamCode)
         {
-            var searchParam = new SearchParameter()
+            var searchParam = new SearchParameter
             {
-                Url = $"http://hl7.org/fhir/SearchParameter/Patient-{searchParamName}",
-                Type = SearchParamType.String,
-                Base = new List<ResourceType?>() { ResourceType.Patient },
-                Expression = "Patient.name",
+                Url = $"http://hl7.org/fhir/SearchParameter/{baseType.ToString()}-{searchParamName}",
```
Ivanidzo4ka, line 457: >ToString() [](start = 70, length = 10)

isn't ToString() applies automatically to object if you pass it this way?

rotodd: Ah yes you're right. Fixed in 7555909.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/Models/ReindexJobRecord.cs
rotodd: Fixed in 7555909.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
rotodd: Fixed in 7555909.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
rotodd: If the first value in `param.BaseResourceTypes` is "Resource", we add every type to `resourceList`, since a search parameter with this base type will affect all resources.

Not breaking out of this for-loop wouldn't cause any harm, but it would be inefficient since we would be attempting to add the same thing more than once.

Ivanidzo4ka: I've re-read code, and I agree with you.
There is still corner case like {"Bundle", "DomainResource"} vs {"DomainResource","Bundle"} I think it would produce different resourceList, but my guess no one will do that.

---
In reply to: [597055066](https://github.com/microsoft/fhir-server/pull/1753#discussion_r597055066) [](ancestors = 597055066)

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
rotodd: This is a really good question.

As discussed, any new resources created during a reindexing operation will be indexed with all search parameters marked as `supported` on creation, so it is by design that they aren't picked up here by the worker.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -463,8 +485,13 @@ private async Task CalculateTotalCount(CancellationToken cancellationToken)
                 SearchResult countOnlyResults = await ExecuteReindexQueryAsync(queryForCount, countOnly: true, cancellationToken);
                 if (countOnlyResults != null)
                 {
+                    _reindexJobRecord.ResourceCounts.TryAdd(resourceType, countOnlyResults.TotalCount.Value); // TODO: What action should be taken if this fails?
```
feordin, line 488: Since this is a performance improvement, if it happens to fail, we can simply go ahead with the query for this resource type.  But we will need to protect line 156 above, in case this resource type doesn't make it into the dictionary.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -134,13 +153,16 @@ public async Task ExecuteAsync(ReindexJobRecord reindexJobRecord, WeakETag weakE
                     // Generate separate queries for each resource type and add them to query list.
                     foreach (string resourceType in _reindexJobRecord.Resources)
                     {
-                        var query = new ReindexJobQueryStatus(resourceType, continuationToken: null)
+                        if (_reindexJobRecord.ResourceCounts[resourceType] > 0)
```
feordin, line 156: In relationship to the TODO comment below on line 488.  If the resource fails to be added to the dictionary, it is not terrible. as long as we don't rely upon it being added successfully.  Perhaps we can check if this key exists first?  If not, or if the count is not 0, then continue with adding the query.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
rotodd: Good catch - even if it's unlikely I think it makes sense to consider it. Fixed in acab74a.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
rotodd: Makes sense, thanks. Fixed in a8a26fb.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
rotodd: Thank you for catching these TODOs, I missed them when putting up my PR! Added a check to see if the resource key does not exist in a8a26fb.

docs/rest/ReindexFlowResourceBase.http
```diff
@@ -0,0 +1,170 @@
...
+# is enabled and the resources were successfully reindexed.
+# Note: this assumes that appsettings.json has Security.Enabled = false and Reindex.Enabled = true
+
+@baseUrl = https://localhost:44348
+@contentType = application/json
+
+###
+# Create a patient resource.
+# Copy the patient's ID in the response body to use later.
+# You can paste it here for now: e181cab2-9c1d-4025-916d-d63786ccbdcb
```
LTA-Thinking, line 15: Could we add variables to store the Patient id, Observation id, Content-Location header?

docs/rest/ReindexFlowResourceBase.http
```diff
@@ -0,0 +1,170 @@
...
+
+###
+# Check the status of the reindexing operation (scroll down in the response body to see the status).
+# Make sure the ID used here is the one pulled from the last request's response.
+GET {{baseUrl}}/_operations/reindex/34dbddb7-6b02-46d3-9627-095b8afeed25 HTTP/1.1
+
+###
+# Replace the ID with the value retrieved when creating the patient.
+# This should return a single patient.
+GET {{baseUrl}}/Patient?test-id-code=e181cab2-9c1d-4025-916d-d63786ccbdcb&_summary=count HTTP/1.1
```
LTA-Thinking, line 157: This works as intended, but I find it strange that _summary=count doesn't return a self link. For this test a self link would show more proof that the search parameter was honored. I also don't think it lines up with the spec's intention for _summary. I don't feel it is meant to summarize the bundle, just the resources in the bundle.

docs/rest/ReindexFlowResourceBase.http
rotodd: Good idea! Added in d7eb5eb.

docs/rest/ReindexFlowResourceBase.http
rotodd: I agree that having the self link would be helpful for testing. Replaced `_summary=true` with `_total=accurate` in 907bcfa.

I created an investigate story to examine the behaviour of `_summary` (please see [AB#80540](https://microsofthealth.visualstudio.com/Health/_workitems/edit/80540))

feordin (approved):

docs/rest/ReindexFlowResourceBase.http
Ivanidzo4ka: I would argue what `_summary` behavior is correct, at least from spec perspective.

LTA-Thinking (approved):

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -105,21 +109,38 @@ public async Task ExecuteAsync(ReindexJobRecord reindexJobRecord, WeakETag weakE
...
+                            }
+
+                            if (baseResourceType == "DomainResource")
+                            {
+                                var domainResourceChildResourceTypes = _modelInfoProvider.GetResourceTypeNames().ToHashSet();
+
+                                // Remove types that inherit from Resource directly
+                                domainResourceChildResourceTypes.Remove("Binary");
+                                domainResourceChildResourceTypes.Remove("Bundle");
+                                domainResourceChildResourceTypes.Remove("Parameters");
```
brendankowitz, line 129: nit: KnownResources.Parameters

docs/rest/ReindexFlowResourceBase.http
```diff
@@ -0,0 +1,167 @@
...
+@baseUrl = https://localhost:44348
+@contentType = application/json
+@patientId = 1407ee02-d505-49fe-9f6f-5fddcf9985f5
+@observationId = da1be095-0a16-481a-a821-e85d676c466e
+@reindexJobId = 438fc1ad-d022-4046-9a57-571488ac1426
+
+###
+# Create a patient resource.
+# Copy the patient's ID in the response body and paste it in the patientId variable.
+POST {{baseUrl}}/Patient HTTP/1.1
```
brendankowitz, line 18: These scripts look good. Do you think we need to consider FHIR versioning at all? i.e. like we do for test data and have /Normative/*.http, or mention in the filename or other?

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
rotodd: That's much nicer, thanks. Fixed in a80b86e.

docs/rest/ReindexFlowResourceBase.http
rotodd: As discussed, I added the compatible versions in a comment in 53cf946.

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Operations/Reindex/ReindexJobTaskTests.cs
```diff
@@ -224,10 +230,12 @@ public async Task GivenRunningJob_WhenExecuted_ThenQueuedQueryCompleted()
...
-            Assert.Equal(10, job.Count);
+            Assert.Equal(_mockedSearchCount * 2, job.Count);
             Assert.Equal(4, job.Progress);
-            Assert.Equal("Appointment,AppointmentResponse", job.ResourceList);
-            Assert.Equal("http://hl7.org/fhir/SearchParameter/AppointmentResponse-appointment", job.SearchParamList);
+            Assert.Contains("Appointment", job.ResourceList);
+            Assert.Contains("AppointmentResponse", job.ResourceList);
+            Assert.Contains("http://hl7.org/fhir/SearchParameter/AppointmentResponse-appointment", job.SearchParamList);
+            Assert.Contains("http://hl7.org/fhir/SearchParameter/Appointment-date", job.SearchParamList);
             Assert.Collection<ReindexJobQueryStatus>(
```
Ivanidzo4ka, line 239: >Collection [](start = 19, length = 10)

I think tests failing on order of ResourceTypes

rotodd: Thanks, I removed the ordering in c36800e.

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Features/Operations/Reindex/ReindexJobTests.cs
```diff
@@ -322,20 +330,97 @@ public async Task GivenNewSearchParamCreatedAfterResourcesToBeIndexed_WhenReinde
...
+            // Set the delay count limit to be higher here for debugging purposes
+            while (reindexJobWrapper.JobRecord.Status != operationStatus && delayCount < 100)
             {
                 await Task.Delay(1000);
                 delayCount++;
                 reindexJobWrapper = await _fhirOperationDataStore.GetReindexJobByIdAsync(response.Job.JobRecord.Id, cancellationTokenSource.Token);
             }
 
-            Assert.InRange(delayCount, 0, 9);
+            Assert.InRange(delayCount, 0, 39);
```
Ivanidzo4ka, line 423: >Assert.InRange(delayCount, 0, 39); [](start = 12, length = 34)

Considering how painful this part is, I would suggest to rewrite it to Assert.Equal(operationStatus,reindexJobWrapper.JobRecord.Status)

I think original intent of this check was to make sure we successfully finished operation and not timeout it.

rotodd: If I understand correctly, you are suggesting that we take out `delayCount` entirely.

In that case, would have `while (reindexJobWrapper.JobRecord.Status != operationStatus)`, so we could simply take out this check, right?

Ivanidzo4ka: Not exactly. Delay count can stay because something can go wrong and we have to wait 60 minutes in tests to fail or whatever is timeout on running container. I saying our while loop has two statements, and instead of checking delay count to make sure we left loop in timely manner, we can check other statement.

rotodd: That's much nicer, thanks 😊 fixed in ba197e5.

PR #1756, opened 2021-03-18
add cda converter

hongyh13:
## Description
Add new data type C-CDA in convert API.

Use the updated converter NuGet package version 3.5 that supports data type C-CDA.

Unify the parameters (inputDataType and default templateCollectionReference) of $convert-data to case-insensitive.

## Related issues
Addresses [issue #].

## Testing
Unit test on newly added functions.
E2E tests on local dev box and remote fhir service (set environment variable: TestEnvironmentUrl)

## FHIR Team Checklist

Feature


src/Microsoft.Health.Fhir.Core/Features/Operations/ConvertData/Models/InputDataTypeAndDefaultTemplateCollectionInconsistentException.cs
```diff
@@ -0,0 +1,25 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Diagnostics;
+using Microsoft.Health.Fhir.Core.Exceptions;
+using Microsoft.Health.Fhir.Core.Models;
+
+namespace Microsoft.Health.Fhir.Core.Features.Operations.ConvertData.Models
+{
+    public class InputDataTypeAndDefaultTemplateCollectionInconsistentException : FhirException
```
Ivanidzo4ka, line 12: >InputDataTypeAndDefaultTemplateCollectionInconsistentException [](start = 17, length = 62)

Is there are any benefit of creating this exception comparing to just use `RequestNotValidException`?

Ivanidzo4ka:
>     }

Are they gonna change in the future?
If not is it really worth to have 2 different calls doing same thing?

super nit: you can use `=>` (Expression bodied) syntax in tests as well. :) 

---
Refers to: src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Controllers/ConvertDataControllerTests.cs:349 in df17b95. [](commit_id = df17b95a75a5316b5d444e166c9d8f21b1b5ee92, deletion_comment = False)

Ivanidzo4ka:
>     }

one more nit: why it's not just 
`private const string DefaultConvertResponse = ""{ \"resourceType\": \"Bundle\" }"";`



---
In reply to: [802109408](https://github.com/microsoft/fhir-server/pull/1756#issuecomment-802109408) [](ancestors = 802109408)

---
Refers to: src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Controllers/ConvertDataControllerTests.cs:349 in df17b95. [](commit_id = df17b95a75a5316b5d444e166c9d8f21b1b5ee92, deletion_comment = False)

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Controllers/ConvertDataControllerTests.cs
```diff
@@ -147,7 +197,72 @@ private static Parameters GetParamsResourceWithTooManyParams()
...
+
+            return parametersResource;
+        }
+
+        private static Parameters GetParamsResourceWithInconsistentParamsWrongCcdaDefaultTemplates()
+        {
+            var parametersResource = new Parameters();
+            parametersResource.Parameter = new List<Parameters.ParameterComponent>();
+
+            parametersResource.Parameter.Add(new Parameters.ParameterComponent() { Name = ConvertDataProperties.InputData, Value = new FhirString(GetSampleHl7v2Message()) });
```
Ivanidzo4ka, line 244: >parametersResource.Parameter.Add(new Parameters.ParameterComponent() { Name = ConvertDataProperties.InputData, Value = new FhirString(GetSampleHl7v2Message()) }); [](start = 12, length = 162)

Considering how often this pattern repeats, I would suggest to add 
```
private static void AddParamComponent(Parameters resource, string name, string value)=>
resource.Parameter.Add(new Parameters.ParameterComponent() { Name =name, Value =new FhirString(value)});
```
and just do:
`AddParamComponent(parametersResource,ConvertDataProperties.InputDataType, "Hl7V2");`

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Operations/ConvertData/ConvertDataRequestHandlerTests.cs
```diff
@@ -76,12 +96,27 @@ private ConvertDataRequestHandler GetRequestHandler()
...
         }
+
```
Ivanidzo4ka, line 111: I can see this messages and method across multiple tests.
Can we refactor them into one common class which class tests can derived from?

src/Microsoft.Health.Fhir.Core/Features/Operations/ConvertData/Models/InputDataTypeAndDefaultTemplateCollectionInconsistentException.cs
hongyh13: The response message of RequestNotValidException is "Value of the following parameter {0} is invalid.". There is only one parameter in the message, while there are two parameters in the inconsistent message.

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Operations/ConvertData/ConvertDataRequestHandlerTests.cs
hongyh13: Thanks for the good suggestion. I will move them into a common class.

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Controllers/ConvertDataControllerTests.cs
hongyh13: Good idea. I will do it.

src/Microsoft.Health.Fhir.Core/Features/Operations/ConvertData/Models/InputDataTypeAndDefaultTemplateCollectionInconsistentException.cs
Ivanidzo4ka: Then I look on this exception and `RequestNotValidException` only difference I see is name of the class, they both accept strings, I understand what you want to display different string for case of inconsistency, but you can just pass different string to already established exception instead of creating copy of it.

---
In reply to: [597468434](https://github.com/microsoft/fhir-server/pull/1756#discussion_r597468434) [](ancestors = 597468434)

Ivanidzo4ka:
That conversion_templates.gz doesn't looks like it should be part of PR.

src/Microsoft.Health.Fhir.Shared.Tests/Samples.cs
```diff
@@ -94,6 +94,12 @@ public static Stream GetDefaultConversionTemplates()
...
+
+        public static string GetSampleConvertDataResponse() => "{ \"resourceType\": \"Bundle\" }";
+
```
Ivanidzo4ka, line 102: nitpicking and senile grumbling: Methods implies what some calculation would be performed inside. For example in methods above we read something. Here it's just return a string. Why we need methods when we can just have `const string`?

hongyh13:
> That conversion_templates.gz doesn't looks like it should be part of PR.

I modified it to pass the E2E tests, and will remove it after Yue's fix pr (https://github.com/microsoft/fhir-server/pull/1759) is merged,

src/Microsoft.Health.Fhir.Shared.Tests/Samples.cs
hongyh13: Thanks for the good suggestion. I will use const string instead :)

src/Microsoft.Health.Fhir.Core/Features/Operations/ConvertData/Models/InputDataTypeAndDefaultTemplateCollectionInconsistentException.cs
hongyh13: I removed `InputDataTypeAndDefaultTemplateCollectionInconsistentException`, and use `RequestNotValidException` now.

Ivanidzo4ka:
Code looks good, feel free to move out of Draft as soon as #1759 merged and conversion_templates.gz get removed from PR.

hongyh13:
> Code looks good, feel free to move out of Draft as soon as #1759 merged and conversion_templates.gz get removed from PR.

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Controllers/ConvertDataControllerTests.cs
```diff
@@ -60,40 +86,62 @@ public async Task GivenAConvertDataRequest_WhenInvalidBodySent_ThenRequestNotVal
             await Assert.ThrowsAsync<RequestNotValidException>(() => _convertDataeEnabledController.ConvertData(body));
         }
 
+        [Theory]
+        [MemberData(nameof(InconsistentBody), MemberType = typeof(ConvertDataControllerTests))]
+        public async Task GivenAConvertDataRequest_WhenInconsistentBodySent_ThenInconsistentThrown(Parameters body)
+        {
+            _convertDataeEnabledController.ControllerContext.HttpContext.Request.Method = HttpMethods.Post;
```
Ivanidzo4ka, line 93: >_convertDataeEnabledController.ControllerContext.HttpContext.Request.Method = HttpMethods.Post; [](start = 12, length = 95)

I don't understand this line. There is direct call to controller method, you don't have to set method type in  context to just call method. In e2e tests you do, because you invoke controller action via going through all the debris of ASP.NET, but to just call method on object...

I also don't see any checks inside controller (other than attribute) for method (and it's correct).

Can we remove this line from this test and other tests in this file.

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Operations/ConvertData/ConvertDataEngineTests.cs
```diff
@@ -68,37 +91,65 @@ public async Task GivenConvertDataRequest_WithDefaultTemplates_CorrectResultShou
...
+        [Theory]
+        [InlineData("       ")]
+        [InlineData("Abc")]
+        [InlineData("¶Š™œãý£¾")]
+        [InlineData("<?xml version=\"1.0\"?><?xml-stylesheet type='text/xsl' href=''?>")]
+        public async Task GivenCcdaConvertDataRequest_WithWrongInputData_ConvertDataFailedExceptionShouldBeThrown(string inputData)
+        {
+            var convertDataEngine = GetDefaultEngine();
+            var request = GetCcdaRequestWithInputData(inputData);
+            var exception = await Assert.ThrowsAsync<ConvertDataFailedException>(() => convertDataEngine.Process(request, CancellationToken.None));
```
Ivanidzo4ka, line 123: >ConvertDataFailedException [](start = 53, length = 26)

I don't see that exception in `OperationOutcomeExceptionFilterAttribute`.
What we return to customer in case of bad data? InternalServerError?

moria97: We will return 400 Bad Request as this exception captures user errors like invalid input. The ```ConvertDataUnhandledException ``` will capture unexpecture expection and return ```500 Internal Server Error```

Ivanidzo4ka: My bad, for some reason I thought default behavior for `OperationOutcomeExceptionFilterAttribute` to return 500. I would still suggest to expilcitly specify that exception in filter, but you right it returns 400 and there is test for that in E2E tests.

---
In reply to: [600156327](https://github.com/microsoft/fhir-server/pull/1756#discussion_r600156327) [](ancestors = 600156327)

Ivanidzo4ka (approved):
:shipit:

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Operations/ConvertData/ConvertDataEngineTests.cs
moria97: Agree. We can add exception handling to ```OperationOutcomeExceptionFilterAttribute``` explicitly to avoid issue in future updates.

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Controllers/ConvertDataControllerTests.cs
hongyh13: Thanks for pointing out it. I removed and tested it, it works! Now, all the controllerContext are removed in this file.

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Operations/ConvertData/ConvertDataEngineTests.cs
hongyh13: Added `ConvertDataUnhandledException `  to `OperationOutcomeExceptionFilterAttribute `.

Ivanidzo4ka:
@hongyh13 Code looks good to me, if you think you done with it, feel free to merge it in.

PR #1757, opened 2021-03-18
Attempt to issue selective cross-partition searches in parallel

johnstairs:
## Description

Searches require issuing cross-partition queries, which we currently perform sequentially. This can result in poor performance when a search result yields a small number of resources spread out over a large number of physical partitions. The Cosmos DB SDK allows you to specify that a query should be run with parallelism. When this option is selected, the SDK will query ALL physical partitions (with the degree of parallelism requested) even though it may already have retrieved more that enough results to satisfy the requested page size. I've opened https://github.com/Azure/azure-cosmos-dotnet-v3/issues/2290 to request that we have more control over the prefetching behavior.

There is therefore a significant downside to selecting parallelism when a search is not selective: we will be discarding results after having incurred the request charge to fetch them. So we want to select parallelism only when we are pretty confident that it is the right thing to do for the query.

With  this change, we're going to start keeping a running average of the number of physical partitions a search needs to hit and using this average to decide whether to issue a query with parallelism enabled or not the next time we see a similar query. This cache is keyed on a value-insensitive comparison of search expression trees. The assumption is that the selectivity of a search like `Patient?date=1990-09-09&name=Ferguson` would be similar to a query like `Patient?date=2001-02-03&name=Jordan`. This is course will not always be true: e.g. the selectivity of `Observation?date=2013-08-23T22:44:23` will probably be a lot higher than `Observation?date=2013`.

As an example, on a server that is connection to a database with 10 physical partitions that is on the other side of the Atlantic (so as to exaggerate latencies), the first time we run this query:

```
/Observation?date=2013-08-23T22:44:23
```

It takes around 5 seconds, but the next time, we execute it with parallelism and the request takes around 600ms.

This change is made up of:

- An LRU cache of selectivity of queries
- Value-insensitive hash code and equality members implemented for all expression types
- `CosmosDbCollectionPhysicalPartitionInfo`, which periodically obtains the number of physical partitions in the Cosmos DB collection. Unfortunately, the SDK does not expose this information so we have to use the REST API.
- Improved query logging. We now include physical partition info and DB-reported duration, and this will be logged for all interactions with the database, not just queries. 

## Related issues
Addresses [AB#80211](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/80211)

## Testing

Manual testing.

Patch


src/Microsoft.Health.Fhir.Core/Features/Context/IFhirRequestContext.cs
```diff
@@ -29,5 +29,10 @@ public interface IFhirRequestContext : IRequestContext
         /// Indicates whether this running as part of a background task instead of an HTTP request
         /// </summary>
         bool IsBackgroundTask { get; set; }
+
+        /// <summary>
+        /// A weakly-typed property bag
```
brendankowitz, line 34: Sounds similar to https://docs.microsoft.com/en-us/dotnet/api/system.web.httpcontext.items?view=netframework-4.8. Perhaps expand the comment the intended use?

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosDbCollectionPhysicalPartitionInfo.cs
```diff
@@ -0,0 +1,166 @@
...
+                }
+            }
+        }
+
+        private async Task<int> GetPhysicalPartitionCount(CancellationToken cancellationToken)
+        {
+            using HttpClient client = _httpClientFactory.CreateClient();
+
+            string host = _dataStoreConfiguration.Host;
+            string key = _dataStoreConfiguration.Key;
```
brendankowitz, line 87: Does this rotate or change in the managed service?

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosDbCollectionPhysicalPartitionInfo.cs
```diff
@@ -0,0 +1,166 @@
...
+                    : GenerateAuthToken(
+                        "get",
+                        "pkranges",
+                        $"dbs/{_dataStoreConfiguration.DatabaseId}/colls/{_collectionConfiguration.CollectionId}",
+                        date,
+                        key));
+
+            using var httpRequestMessage = new HttpRequestMessage(
+                HttpMethod.Get,
+                $"{host}/dbs/{_dataStoreConfiguration.DatabaseId}/colls/{_collectionConfiguration.CollectionId}/pkranges")
```
brendankowitz, line 111: Interesting, in the v2 sdk it was https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.documents.client.documentclient.readpartitionkeyrangefeedasync?view=azure-dotnet

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosDbCollectionPhysicalPartitionInfo.cs
```diff
@@ -0,0 +1,166 @@
...
+        {
+            _backgroundLoopCancellationTokenSource.Cancel();
+            await _backgroundLoopTask;
+            _backgroundLoopCancellationTokenSource.Dispose();
+            _backgroundLoopTask.Dispose();
+        }
+
+        private static string GenerateAuthToken(string verb, string resourceType, string resourceId, string date, string key)
+        {
+#pragma warning disable CA1308 // Normalize strings to uppercase. Required to be lowercase.
```
brendankowitz, line 151: Should we disable this rule globally?

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosDbCollectionPhysicalPartitionInfo.cs
```diff
@@ -0,0 +1,166 @@
...
+                throw new RequestRateExceededException(null);
+            }
+
+            _logger.LogError("Unable to get physical partition count. {ResponseBody}", await httpResponseMessage.Content.ReadAsStringAsync(cancellationToken));
+
+            httpResponseMessage.EnsureSuccessStatusCode();
+            return 0; // will not reach this
+        }
+
+        public static bool IsResourceToken(string key) => key.StartsWith("type=resource&", StringComparison.InvariantCulture);
```
brendankowitz, line 139: Does this need a nullcheck?

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
```diff
@@ -342,7 +447,7 @@ private async Task<Expression> RecurseChainedExpression(ChainedExpression expres
         {
             var feedOptions = new QueryRequestOptions
             {
-                MaxConcurrency = -1, // execute counts across all partitions
+                MaxConcurrency = MaxQueryConcurrency, // execute counts across all partitions
```
brendankowitz, line 450: Should this remain as -1, or should the comment be updated?

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
```diff
@@ -333,7 +348,97 @@ private async Task<Expression> RecurseChainedExpression(ChainedExpression expres
...
+        {
+            int? averagePartitionCount = queryPartitionStatistics.GetAveragePartitionCount();
+
+            if (averagePartitionCount.HasValue)
+            {
+                // this is not a new query
+
+                double fractionOfPartitionsHit = (double)averagePartitionCount.Value / _physicalPartitionInfo.PhysicalPartitionCount;
+
+                if (fractionOfPartitionsHit > 0.5)
```
brendankowitz, line 435: Just clarifying, does this mean that on average, the query was executed across more than 50% of the physical partitions?

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosDbCollectionPhysicalPartitionInfo.cs
johnstairs: Sure, I think that makes sense

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
johnstairs: Good catch, I'll update the comment

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
johnstairs: Yes, that's right.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosDbCollectionPhysicalPartitionInfo.cs
```diff
@@ -0,0 +1,166 @@
...
+using Microsoft.Health.Extensions.DependencyInjection;
+using Microsoft.Health.Fhir.CosmosDb.Configs;
+
+namespace Microsoft.Health.Fhir.CosmosDb.Features.Storage
+{
+    /// <summary>
+    /// Maintains the count of physical partitions in a Cosmos DB collection. We have to resort to the REST API to get this information.
+    /// The count can evolve over time, so we periodically query the collection to update the value.
+    /// </summary>
+    internal class CosmosDbCollectionPhysicalPartitionInfo : IRequireInitializationOnFirstRequest, IAsyncDisposable, ICosmosDbCollectionPhysicalPartitionInfo
```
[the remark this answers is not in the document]
johnstairs: In the managed service, the CosmosClient is created in an IHostedService. In order to avoid problems with dependencies between IHostedServices, I opted for IRequireInitializationOnFirstRequest.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosDbCollectionPhysicalPartitionInfo.cs
johnstairs: Yes, these change.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosDbCollectionPhysicalPartitionInfo.cs
johnstairs: Would have made things easier...

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
johnstairs: More precisely, on average, we needed more than 50% of the physical partitions to fill a page of results.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosDbCollectionPhysicalPartitionInfo.cs
johnstairs: Made this method private and added validation for key and host in the calling method.

brendankowitz (approved):

PR #1759, opened 2021-03-19
Remove old version templates in convert data test

moria97:
## Description
Remove old version templates in convert data test which may cause parse failures in E2E test. Now we directly upload embedded templates to remote ACR.

## Related issues
Addresses [issue #].

## Testing
E2E test

## FHIR Team Checklist

Skip


LTA-Thinking:
When do these tests get run? The comments says they are skipped locally and I see them being skipped in the PR.

moria97:
> When do these tests get run? The comments says they are skipped locally and I see them being skipped in the PR.

Actually these  tests do not run in the OSS pipeline because they need to inject the target ACR info into environment variables (We need the ACR secrets to upload a template, for PaaS testing we have made it work by saving the information in the Key Vault of test FHIR account). 
I just made a few changes to add ACR information to account key vault, does that make sense?

samples/templates/default-azuredeploy-docker.json
```diff
@@ -487,6 +487,19 @@
                 "[resourceId('Microsoft.Sql/servers', variables('serviceName'))]"
             ]
         },
+        {
```
LTA-Thinking, line 490: Could this value be calculated in the run-test file instead of being added to the deployment template? The templates are used by customers to deploy their systems and I don't think we should have test code in here.

moria97: Agree. Since the acr account names are generated with [uniqueString](https://github.com/microsoft/fhir-server/blob/1a0c0bffdc75f376e96c5e3a19b109ea23f97ad3/samples/templates/default-azuredeploy-docker.json#L235) method, we can only read the value from service configuration where we register the ACR login servers. I have removed this from deployment template

Ivanidzo4ka:
>       ScriptType: inlineScript

this script and script starting in line 159 looks identical for me. Can they be refactored into one script file with link to it?
Only difference I see is name of ACR container is suffixed with -sql for sql.
Can same ACR container be used for SQL and Cosmos backend? Is there are difference or ACR just can't handle request from different systems?


---
Refers to: build/jobs/run-tests.yml:62 in 734a0c8. [](commit_id = 734a0c845797049ced5bc15aaa8c5d7da2b40a32, deletion_comment = False)

moria97:
> into

Yes. They are two parts for setting up Cosmos and Sql E2E variables. I feel we can refactor the whole into a set-variable task, but that would take some time to test the pipeline. We can merge this bug fix first and create an item to track this? 

Same ACR can be shared by different server, but in the pipeline settings, all resources are deployed with ARM template where each ACR are binded to a single FHIR server. Just like the storage account in export scenario.

moria97:
> > ```
> >   ScriptType: inlineScript
> > ```
> 
> this script and script starting in line 159 looks identical for me. Can they be refactored into one script file with link to it?
> Only difference I see is name of ACR container is suffixed with -sql for sql.
> Can same ACR container be used for SQL and Cosmos backend? Is there are difference or ACR just can't handle request from different systems?
> 
> Refers to: build/jobs/run-tests.yml:62 in [734a0c8](https://github.com/microsoft/fhir-server/commit/734a0c845797049ced5bc15aaa8c5d7da2b40a32). [](commit_id = [734a0c8](https://github.com/microsoft/fhir-server/commit/734a0c845797049ced5bc15aaa8c5d7da2b40a32), deletion_comment = False)

Yes. They are two parts for setting up Cosmos and Sql E2E variables. I feel we can refactor the whole into a set-variable task, but that would take some time to test the pipeline. We can merge this bug fix first and create an item to track this?

Same ACR can be shared by different server, but in the pipeline settings, all resources are deployed with ARM template where each ACR are binded to a single FHIR server. Just like the storage account in export scenario.

Ivanidzo4ka (approved):
:shipit:

PR #1762, opened 2021-03-22
Update 7.sql and 7.diff.sql to be idempotent

YazanMSFT:
## Description
Updates the latest SQL script to be idempotent

## Related issues
Addresses [issue #].

## Testing
Added test that runs the full script and diff script twice.
Ran a local test to compare existing schema with the updated idempotent schema and verified that they're the same (aside from the comments on the stored proc scripts not showing up due to new statements that got inserted above it)

## FHIR Team Checklist


johnstairs:
Why does the full .sql file need to be idempotnent? Can't we run it in a transaction?

YazanMSFT:
> Why does the full .sql file need to be idempotnent? Can't we run it in a transaction?

Good point, I guess it doesn't. I was thinking that we were planning on removing the transaction in the schema management tool as we discussed before, but I guess that only applies on upgrades. I can revert the changes to the full .sql file.

johnstairs (approved):

PR #1768, opened 2021-03-23
Fix CI deploy pipeline

moria97:
## Description
Fix [CI deploy pipeline](https://microsofthealthoss.visualstudio.com/FhirServer/_build?definitionId=29&_a=summary) which is broken due to several issues:
1. Syntax error in referencing add-aad-test-environment.yml job.
2. Wrong config in setting up aad which needs windows platform agent.
3. Wrong schemaAutomaticUpdatesEnabled value in sql deployment tasks.
 
## Related issues
Addresses [issue #].

## Testing
Manual test.

## FHIR Team Checklist

Skip

### DIFF
--- a/build/ci-deploy.yml
+++ b/build/ci-deploy.yml
@@ -21,7 +21,7 @@ stages:
   displayName: Provision Environment
   dependsOn: []
   jobs:
-  - template: ./jobs/cleanup.yml
+  - template: ./jobs/cleanup-resourcegroup-aad.yml
   - job: provision
     dependsOn: DeleteResourceGroup 
     steps:
--- a/build/ci-deploy.yml
+++ b/build/ci-deploy.yml
@@ -81,7 +81,7 @@ stages:
       resourceGroup: $(DeploymentEnvironmentName)
       testEnvironmentUrl: $(TestEnvironmentUrl)
       imageTag: $(ImageTag)
-      schemaAutomaticUpdatesEnabled: 'tool'
+      schemaAutomaticUpdatesEnabled: 'auto'
 
 - stage: deployR4
   displayName: 'Deploy R4 Site'
@@ -118,7 +118,7 @@ stages:
       resourceGroup: $(DeploymentEnvironmentName)
       testEnvironmentUrl: $(TestEnvironmentUrl)
       imageTag: $(ImageTag)
-      schemaAutomaticUpdatesEnabled: 'tool'
+      schemaAutomaticUpdatesEnabled: 'auto'
 
 - stage: deployR5
   displayName: 'Deploy R5 Site'
@@ -155,4 +155,4 @@ stages:
       resourceGroup: $(DeploymentEnvironmentName)
       testEnvironmentUrl: $(TestEnvironmentUrl)
       imageTag: $(ImageTag)
-      schemaAutomaticUpdatesEnabled: 'tool'
+      schemaAutomaticUpdatesEnabled: 'auto'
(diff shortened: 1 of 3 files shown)


build/ci-deploy.yml
```diff
@@ -114,7 +118,7 @@ stages:
       resourceGroup: $(DeploymentEnvironmentName)
       testEnvironmentUrl: $(TestEnvironmentUrl)
       imageTag: $(ImageTag)
-      schemaAutomaticUpdatesEnabled: 'tool'
+      schemaAutomaticUpdatesEnabled: 'auto'
```
Ivanidzo4ka, line 121: >auto [](start = 38, length = 4)

Can I ask why? I was under impression we always push latest version of schema, and you need to update schema only if your service is living through rollout of new schemas, which shouldn't be the case for CI build.

build/ci-deploy.yml
```diff
@@ -40,7 +40,11 @@ stages:
   - provisionEnvironment
   - DockerBuild
   jobs:
-  - template: ./jobs/add-aad-test-environment.yml
+  - job: setup
+    pool:
+      vmImage: '$(WindowsVmImage)'
```
Ivanidzo4ka, line 45: >WindowsVmImage [](start = 18, length = 14)

Powershell not working on non-windows machines?

Ivanidzo4ka:
Any chance you can add link to example of broken CI pipeline in description?

build/ci-deploy.yml
moria97: It requires higher version to run in other platform. And in our case, it fails to install AzureAD module without this platform selection.

build/ci-deploy.yml
moria97: Yes, I'm also confused here. This pipeline is to replace the deployment in ci resource group. With the parameter set to false, all SQL sites cannot find the stored procedures (seems no initilization) and fail to start.

Ivanidzo4ka: [https://github.com/microsoft/healthcare-shared-components/issues/196](https://github.com/microsoft/healthcare-shared-components/issues/196)
Seems like a bug to me, I filed an issue.

---
In reply to: [599859457](https://github.com/microsoft/fhir-server/pull/1768#discussion_r599859457) [](ancestors = 599859457)

LTA-Thinking: For CI setting this to `auto` is correct. We don't want to have to manually update the databases for every change and the CI databases are persistent.

LTA-Thinking:
The script won't run in a PR environment currently. The cleanup script it is referencing had a step added to cleanup docker images, which shouldn't be done in this case as it will prevent the deployments latter in the build from happening. This script needs to copy the resource group deletion part of the cleanup script and run the aad-cleanup job, but not delete docker images.

build/ci-deploy.yml
brendankowitz: This looks ok, I believe the Connect-AAD call in this job required Windows atm

moria97:
> The script won't run in a PR environment currently. The cleanup script it is referencing had a step added to cleanup docker images, which shouldn't be done in this case as it will prevent the deployments latter in the build from happening. This script needs to copy the resource group deletion part of the cleanup script and run the aad-cleanup job, but not delete docker images.

Agree. I will setup another cleanup job

moria97:
> Any chance you can add link to example of broken CI pipeline in description?

Sure, I have updated the description. Thanks

LTA-Thinking (approved):

PR #1771, opened 2021-03-23
Guard against double-dispose of CosmosDbCollectionPhysicalPartitionInfo

johnstairs:
## Description
We get an ObjectDisposedException when shutting down the server because CosmosDbCollectionPhysicalPartitionInfo is registeres as an IRequireInitializationOnFirstRequest and an ICosmosDbCollectionPhysicalPartitionInfo, and so is disposed twice. The second time it throws.

Patch


Ivanidzo4ka:
>so is disposed twice

Does it mean it run `EnsureInitialized` twice as well and we have extra useless while loop running?

Ivanidzo4ka (approved):
:shipit:

PR #1772, opened 2021-03-23
Update Authentication.md

vladimir-golovchenko:
* replaced the old-version method with [AddDevelopmentAuthEnvironmentIfConfigured](https://github.com/microsoft/fhir-server/blob/main/src/Microsoft.Health.Fhir.Shared.Web/DevelopmentIdentityProviderRegistrationExtensions.cs#L126)
* extended the 'token response'.

### DIFF
--- a/docs/Authentication.md
+++ b/docs/Authentication.md
@@ -27,8 +27,10 @@ For the F5 experience and test environments, an in-process identity provider is 
 To enable the development identity provider, add the following code to your app configuration:
 
 ```csharp
-builder.AddDevelopmentAuthEnvironment("path/to/testAuthEnvironment.json");
+builder.AddDevelopmentAuthEnvironmentIfConfigured(configurationRoot);
 ```
+
+`configurationRoot` object should had key "TestAuthEnvironment:FilePath" pointing to [TestAuthEnvironment.json](https://github.com/microsoft/fhir-server/blob/main/testauthenvironment.json) location.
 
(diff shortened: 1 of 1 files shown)


docs/Authentication.md
```diff
@@ -27,7 +27,7 @@ For the F5 experience and test environments, an in-process identity provider is
 To enable the development identity provider, add the following code to your app configuration:
 
 ```csharp
-builder.AddDevelopmentAuthEnvironment("path/to/testAuthEnvironment.json");
+builder.AddDevelopmentAuthEnvironmentIfConfigured(configurationRoot);
 ```
 
 The code above, along with the `TestAuthEnvironment.json` file will configure and start the identity provider for use by the FHIR API.
```
Ivanidzo4ka, line 33: >The code above, along with the `TestAuthEnvironment.json [](start = 0, length = 56)

nit: I would add few words, something like : `configurationRoot` object should had key "TestAuthEnvironment:FilePath" pointing to `TestAuthEnvironment.json` location. In past it was direct string, so sentence was ok, but now it feels disconnected.

Ivanidzo4ka (approved):
:shipit:

docs/Authentication.md
vladimir-golovchenko: fixed, please check.

Ivanidzo4ka:
Thank you for your contribution @vladimir-golovchenko !

PR #1782, opened 2021-03-25
Read existing SearchParameter resources

feordin:
## Description
Reads the SearchParameter resources that are present in the FHIR data store and loads them into the SearchParameterDefinitionManager with the proper status.

## Related issues
Addresses [issue [AB#77543](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/77543)].

## Testing

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterResourceDataStore.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterResourceDataStore.cs
@@ -1,0 +1,78 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using EnsureThat;
+using Microsoft.Health.Extensions.DependencyInjection;
+using Microsoft.Health.Fhir.Core.Extensions;
+using Microsoft.Health.Fhir.Core.Features.Definition;
+using Microsoft.Health.Fhir.Core.Models;
+
+namespace Microsoft.Health.Fhir.Core.Features.Search.Registry
+{
+    public class SearchParameterResourceDataStore : IRequireInitializationOnFirstRequest
+    {
+        private readonly ISearchParameterDefinitionManager _searchParameterDefinitionManager;
+        private readonly SearchParameterStatusManager _searchParameterStatusManager;
+        private readonly Func<IScoped<ISearchService>> _searchServiceFactory;
+        private readonly IModelInfoProvider _modelInfoProvider;
+
+        public SearchParameterResourceDataStore(
+            ISearchParameterDefinitionManager searchParameterDefinitionManager,
+            SearchParameterStatusManager searchParameterStatusManager,
+            Func<IScoped<ISearchService>> searchServiceFactory,
+            IModelInfoProvider modelInfoProvider)
+        {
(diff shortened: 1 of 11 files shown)


src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterResourceDataStore.cs
```diff
@@ -0,0 +1,78 @@
...
+                    {
+                        continuationToken = null;
+                    }
+
+                    if (result != null && result.Results != null && result.Results.Count() > 0)
+                    {
+                        var searchParams = result.Results.Select(r => r.Resource.RawResource.ToITypedElement(_modelInfoProvider)).ToList();
+
+                        _searchParameterDefinitionManager.AddNewSearchParameters(searchParams);
+                        await _searchParameterStatusManager.AddSearchParameterStatusAsync(searchParams.Select(s => s.GetStringScalar("url")).ToList());
```
Ivanidzo4ka, line 71: Let's say I POST before hand "[http://hl7.org/fhir/SearchParameter/QuestionnaireResponse-questionnaire](http://hl7.org/fhir/SearchParameter/QuestionnaireResponse-questionnaire)" which is unsupported of our service.

What would be end result?

feordin: This was a great question!  It made me think about handling errors, duplicates, and invalid SearchParameters.  Code is updated, and test case updated as well.

src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterResourceDataStore.cs
```diff
@@ -0,0 +1,93 @@
...
+
+                        foreach (var searchParam in searchParams)
+                        {
+                            try
+                            {
+                                await _searchParameterOperations.AddSearchParameterAsync(searchParam);
+                            }
+                            catch (SearchParameterNotSupportedException ex)
+                            {
+                                _logger.LogError(ex, $"Error loading search parameter {searchParam.GetStringScalar("url")} from data store.");
```
Ivanidzo4ka, line 80: >LogError [](start = 40, length = 8)

Would this resolve into exception in geneva logs?
If yes, can we lower priority of this message?
Since there is no way to tell about this errors to customer we gonna get them on every service load, which can be overwhelming, and there is no action needed from our side.

I would rather have current messages as warning or informational, and had final `catch (Exception ex)` which would log error.
Generic unexpected error is something we need to worry about.

feordin: Agreed, lowered to warning.

Ivanidzo4ka (approved):
:shipit:

src/Microsoft.Health.Fhir.Shared.Api/Modules/SearchParametersModule.cs
```diff
@@ -0,0 +1,33 @@
...
+using Microsoft.Health.Fhir.Core.Registration;
+
+namespace Microsoft.Health.Fhir.Api.Modules
+{
+    /// <summary>
+    /// Registration of search components.
+    /// </summary>
+    public static class SearchParametersModule
+    {
+        public static IFhirServerBuilder AddSearchParametersModules(this IFhirServerBuilder fhirServerBuilder)
```
Ivanidzo4ka, line 18: >this IFhirServerBuilder [](start = 68, length = 23)

nit: I found only one extension on top of `IFhirServerBuilder` and it's in `FhirServerServiceCollectionExtensions`, not sure is it worth to create new class. Maybe we could add that code next to `AddBackgroundWorkers`

PR #1669, opened 2021-02-16
Adds reindex test

rotodd:
## Description

Refactors reindex integration tests and adds a new test to cover the scenario where a search parameter is created before the resources it impacts (more details can be found in [AB#79555](https://microsofthealth.visualstudio.com/Health/_workitems/edit/79555) which was fixed in #1650).

If this looks familiar: most of this refactoring already lives in the feature branch for SQL reindexing and was reviewed in #1577.

## Related issues
Addresses [AB#79650](https://microsofthealth.visualstudio.com/Health/_workitems/edit/79650).

## Testing
Manual testing:
- Ensured that the tests continue to pass for Cosmos DB
- Temporarily enabled the reindex tests for SQL to ensure that the tests run without fixture-related issues (although some tests fail, as expected)

## FHIR Team Checklist
- ✅ **Update the title** of the PR to be succinct and less than 50 characters
- ✅ **Add a milestone** to the PR for the sprint that it is merged (i.e. add S47)
- ✅ Tag the PR with the type of update: **Bug**, **Dependencies**, **Enhancement**, or **New-Feature**
- ✅ Tag the PR with **Azure API for FHIR** if this will release to the managed service
- Review [squash-merge requirements](https://github.com/microsoft/fhir-server/blob/master/SquashMergeRequirements.md)

PR #1785, opened 2021-03-25
Implements reindexing for SQL

rotodd:
## Description
This PR merges the SQL reindexing branch to `main`. A collection of PRs make up this PR, which:

- Add the SQL table and stored procedures for managing reindex jobs (#1570)
- Implement the data layer that works with the job management schema changes (#1573)
- Set up the test fixture so that existing reindex tests can run for SQL (#1577)
- Add the `SearchParamHash` column to the `dbo.Resource` table and update CRUD operations to support it (#1616)
- Implement the SQL query that is used when searching for resources to reindex (#1668)
- Add a proc to only update the resource's search indices instead of relying on calling `Upsert` (#1682)
- Add a proc to update a batch of resources' search indices in bulk (#1709)
- Make the new schema diff file idempotent (#1784)

## Related issues
Addresses [AB#78973](https://microsofthealth.visualstudio.com/Health/_workitems/edit/78973).

## Testing
- Tests were added to each incremental change in the PRs mentioned above
- All Cosmos reindexing tests have been enabled for SQL
- An additional integration test was added in #1669

## Next steps
- Investigate moving reindexing integration tests to E2E tests [AB#76384](https://microsofthealth.visualstudio.com/Health/_workitems/edit/76384)
- Add retry logic at the SQL data store level [AB#79925](https://microsofthealth.visualstudio.com/Health/_workitems/edit/79925)

## FHIR Team Checklist

Feature (new feature)


src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
```diff
@@ -195,9 +194,147 @@ public async Task<IReadOnlyCollection<ExportJobOutcome>> AcquireExportJobsAsync(
...
+                    if (rowVersion.NullIfEmpty() == null)
+                    {
+                        throw new OperationFailedException(string.Format(Core.Resources.OperationFailed, OperationsConstants.Reindex, "Failed to create reindex job because no row version was returned."), HttpStatusCode.InternalServerError);
+                    }
+
+                    return new ReindexJobWrapper(jobRecord, GetRowVersionAsEtag(rowVersion));
+                }
+                catch (SqlException e)
+                {
+                    if (e.Number == SqlErrorCodes.PreconditionFailed)
```
rotodd, line 271: TODO: instead of throwing, investigate adding a delay, getting the updated job and retrying

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -51,6 +52,7 @@ internal class SqlServerSearchService : SearchService
         private readonly SchemaInformation _schemaInformation;
         private readonly ISortingValidator _sortingValidator;
         private readonly IFhirRequestContextAccessor _requestContextAccessor;
+        private const int _resourceTableColumnCount = 11;
```
Ivanidzo4ka, line 55: >_resourceTableColumnCount [](start = 26, length = 25)

I count it twice and resource table has 10 fields!

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlSearchType.cs
```diff
@@ -0,0 +1,28 @@
...
+    /// Enum to specify the type of search to carry out.
+    /// </summary>
+    /// <remarks>
+    /// With the exception of None, the flags attribute requires enumeration constants to be powers of two, that is, 1, 2, 4, 8.
+    /// </remarks>
+    [Flags]
+    internal enum SqlSearchType
+    {
+        // Default value, set if we do not need to consider history or reindexing
+        None = 0,
```
Ivanidzo4ka, line 20: >None [](start = 8, length = 4)

nit: Maybe call it default, or regular? None somewhat imply what no search needed.

Ivanidzo4ka (approved):
:shipit:

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
rotodd: Investigated and created [AB#80852](https://microsofthealth.visualstudio.com/Health/_workitems/edit/80852), which will follow this PR.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
rotodd: 😵 interesting, good catch. @itye-msft, do you know why we compare against one greater than `reader.FieldCount`?

I set `_resourceTableColumnCount = 10` and used `_resourceTableColumnCount + 1` in 620e165.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlSearchType.cs
rotodd: Fixed in bdc594e.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
namadabu: I'll try answering the question. For non-sort queries, the final result set from the SQL query contains 10 fields (I think it's just a co-incidence that it's equal to the number of columns in the `dbo.ResourceTable` - since the 10 also includes IsMatch and IsPartial). For sort queries, the final result contains one more field since we also extract the `SortValue`. Hence the check.

namadabu: nit: I think it's a good idea to give a name to the "magic number" 10. But I feel that `_resourceTableColumnCount` might not be right since that is not what we are actually referring to here. Maybe something like `resultSetColumnCount`? Even if we do rename it, we should do it as part of a different PR and not this one.

PR #1788, opened 2021-03-26
Update Thirdpartynotices

namadabu:
## Description
Update versions of third party libraries used in the codebase.

## Related issues
Addresses [issue #].

## Testing
No functional change

## FHIR Team Checklist

### DIFF
--- a/THIRDPARTYNOTICES.md
+++ b/THIRDPARTYNOTICES.md
@@ -29,7 +29,7 @@ This file is based on or incorporates material from the projects listed below (T
   > OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
   > SOFTWARE.
 
-## Ensure.That 9.2.0
+## Ensure.That 10.0.0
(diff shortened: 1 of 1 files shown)


Ivanidzo4ka:
This feels like Sisyphean Task.  depenabot would make this file irrelevant in a blink of a eye. Are versions for libraries actually necessary?

namadabu:
> This feels like Sisyphean Task. depenabot would make this file irrelevant in a blink of a eye. Are versions for libraries actually necessary?

I don't know whether the version is a hard requirement. Would also prefer just referring to the library names if that is enough. @brendankowitz @CaitlinV39 thoughts?

brendankowitz:
There is a generated licenses file from [component governance](https://microsofthealth.visualstudio.com/Health/_componentGovernance/47646?_a=components&typeId=5886590&alerts-view-option=active) now, and also a Build Task. I'll check the backlog, ideally we could leverage these

brendankowitz:
Additionally, we should be copying this file into the root of our docker container too :)

brendankowitz (approved):

Issue #1792, opened 2021-03-29
Duplicate resources returned when paging through _sort results (SQL)


namadabu:
**Describe the bug**
When sorting resources (using parameters other than _lastUpdated) we can get duplicate resources in the response under certain situations.

**FHIR Version?**
Stu3/R4/R5

**Data provider?**
SQL Server

**To Reproduce**
1. Post 3 patients with the same birthdate
2. Run the following search query: GET <fhir-server>/Patient?_sort=birthdate&_count=2
3. Follow the next link

**Expected behavior**
When following the next link we should get only one patient in the response.

**Actual behavior**
When following the next link we get two patients in the response.

PR #1793, opened 2021-03-29
Fix pagination for SQL _sort queries

namadabu:
## Description
This PR handles two related issues with respect to paging through sorted results (when using SQL provider).

1. Let's assume we have multiple pages of patiens with the same birthdate. Let's execute the below query:
`GET <fhir-server>/Patient?_sort=birthdate`

When following then next links, we always return the last patient on the previous page as the first patient on the new page of results (duplicate results). This is because we execute the below query:
```
SELECT ResourceSurrogateId AS Sid1, StartDateTime as SortValue
FROM dbo.DateTimeSearchParam
WHERE IsHistory = 0
	AND SearchParamId = @p3
	AND ResourceTypeId = @p2
	AND ((StartDateTime = @p4 AND ResourceSurrogateId >= @p5) OR StartDateTime > @p4)
	AND ResourceSurrogateId IN (SELECT Sid1 FROM cte0)
```

Updated the code so that the ResourceSurrogateId filter is correct:
`AND ((StartDateTime = @p4 AND ResourceSurrogateId > @p5) OR StartDateTime > @p4)`

2. When we calculate the continuation token for queries containing _sort, we use the `ResourceSurrogateId` from the 10th row and the `SortValue` from the 11th row (assuming the requests asks for the default 10 rows). This can cause us to return incorrect results in certain situations (eg: when sort values for 10th and 11th rows are distinct and the `ResourceSurrogateId` for 11th row is less than the 10th row). Updated code to use both `SortValue` and `ResourceSurrogateId` from the 10th row to maintain consistency.

## Related issues
Addresses #1792 .

## Testing
Added more E2E tests to cover the above scenarios

## FHIR Team Checklist

Patch

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
@@ -153,7 +153,7 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Search
                             : Expression.LessThan(SqlFieldName.ResourceSurrogateId, null, continuationToken.ResourceSurrogateId);
 
                         var tokenExpression = Expression.SearchParameter(SqlSearchParameters.ResourceSurrogateIdParameter, lastUpdatedExpression);
-                        searchExpression = searchExpression == null ? tokenExpression : (Expression)Expression.And(tokenExpression, searchExpression);
+                        searchExpression = searchExpression == null ? tokenExpression : Expression.And(tokenExpression, searchExpression);
                     }
                 }
                 else
@@ -228,7 +228,17 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Search
 
                     while (await reader.ReadAsync(cancellationToken))
                     {
-                        (short resourceTypeId, string resourceId, int version, bool isDeleted, long resourceSurrogateId, string requestMethod, bool isMatch, bool isPartialEntry, bool isRawResourceMetaSet, Stream rawResourceStream) = reader.ReadRow(
+                        (short resourceTypeId,
+                        string resourceId,
+                        int version,
+                        bool isDeleted,
+                        long resourceSurrogateId,
+                        string requestMethod,
+                        bool isMatch,
+                        bool isPartialEntry,
+                        bool isRawResourceMetaSet,
+                        Stream rawResourceStream) =
+                            reader.ReadRow(
                             VLatest.Resource.ResourceTypeId,
                             VLatest.Resource.ResourceId,
                             VLatest.Resource.Version,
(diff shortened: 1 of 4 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -246,27 +256,28 @@ private async Task<SearchResult> SearchImpl(SearchOptions searchOptions, bool hi
...
-                            }
-
                             continue;
                         }
 
+                        string rawResource;
+                        using (rawResourceStream)
+                        {
+                            rawResource = await CompressedRawResourceConverter.ReadCompressedRawResource(rawResourceStream);
+                        }
```
Ivanidzo4ka, line 266: curious, why it's been moved?

Ivanidzo4ka (approved):
:shipit:

johnstairs (approved):

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
namadabu: Calling the `reader.GetValue` on the sort column prevented us from reading from the `rawResourceStream` (that is returned from the earlier `reader.ReadRow()` call) later on. Hence I switched the ordering to read the stream first and then call the reader again.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/SortTests.cs
```diff
@@ -45,22 +45,58 @@ public async Task GivenPatients_WhenSearchedWithSortParamsWithHyphen_ThenPatient
             await ExecuteAndValidateBundle($"Patient?_tag={tag}&_sort=-_lastUpdated", false, patients.Reverse().Cast<Resource>().ToArray());
         }
 
-        [Fact]
-        public async Task GivenMoreThanTenPatients_WhenSearchedWithSortParam_ThenPatientsAreReturnedInAscendingOrder()
+        [Theory]
```
brendankowitz, line 48: These should be [SkippableTheory] because birthdate might not be enabled on all dbs

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/SortTests.cs
```diff
@@ -45,22 +45,58 @@ public async Task GivenPatients_WhenSearchedWithSortParamsWithHyphen_ThenPatient
...
+        [InlineData("birthdate")]
+        [InlineData("-birthdate")]
+        [HttpIntegrationFixtureArgumentSets(dataStores: DataStore.SqlServer)]
+        public async Task GivenPatientsWithSameBirthdateAndMultiplePages_WhenSortedByBirthdate_ThenPatientsAreReturnedInCorrectOrder(string sortParameterName)
+        {
+            var tag = Guid.NewGuid().ToString();
+            var patients = await CreatePatientsWithSameBirthdate(tag);
+
+            await ExecuteAndValidateBundle($"Patient?_tag={tag}&_sort={sortParameterName}&_count=3", false, pageSize: 3, patients.Cast<Resource>().ToArray());
```
brendankowitz, line 79: If `-birthdate` isn't this not the correct order?

namadabu: For SQL, we break ties with `ResourceSurrogateId ASC`. Since this dataset contains all patients with same birthdate, even when we pass in `-birthdate` the result will be in "ascending order".

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/SortTests.cs
namadabu: Sent PR #1864  to address this

PR #1795, opened 2021-03-29
Add supported profiles to capability statement.

Ivanidzo4ka:
## Description
Adds supported profiles in capability statement. 

## Related issues
Addresses [issue #].

## Testing

## FHIR Team Checklist

Patch


src/Microsoft.Health.Fhir.Core/Features/Conformance/CapabilityStatementBuilder.cs
```diff
@@ -79,6 +86,19 @@ public ICapabilityStatementBuilder UpdateRestResourceComponent(string resourceTy
                         Reference = $"http://hl7.org/fhir/StructureDefinition/{resourceType}",
                     },
                 };
+                if (!_modelInfoProvider.Version.Equals(FhirSpecification.Stu3))
```
namadabu, line 89: Is there a specific reason we are excluding Stu3 version?

src/Microsoft.Health.Fhir.Core/Features/Conformance/SystemConformanceProvider.cs
```diff
@@ -53,7 +58,7 @@ public override async Task<ResourceElement> GetCapabilityStatementAsync(Cancella
                 {
                     if (_listedCapabilityStatement == null)
                     {
-                        ICapabilityStatementBuilder builder = CapabilityStatementBuilder.Create(_modelInfoProvider, _searchParameterDefinitionManager)
+                        ICapabilityStatementBuilder builder = CapabilityStatementBuilder.Create(_modelInfoProvider, _searchParameterDefinitionManager, _supportedProfiles)
```
namadabu, line 61: Since we cache the capability statement any new profiles posted will not show up unless we restart the server. Would it be useful to add a mechanism to regenerate the capability statement if a new profile is posted? For reference, we do something similar when reindex completes - we publish a `SearchParametersUpdated` notification and reset the capabilitystatement (down below) when we receive that notification.

src/Microsoft.Health.Fhir.Core/Features/Conformance/CapabilityStatementBuilder.cs
Ivanidzo4ka: Thanks, I look properly this time, and STU3 has different way of displaying supporting profiles which I'd missed before.

src/Microsoft.Health.Fhir.Core/Features/Conformance/ICapabilityStatementBuilder.cs
```diff
@@ -16,17 +15,17 @@ public interface ICapabilityStatementBuilder
...
 
         ICapabilityStatementBuilder AddDefaultResourceInteractions();
 
-        ICapabilityStatementBuilder AddDefaultSearchParameters();
+        ICapabilityStatementBuilder SyncSearchParameters();
+
+        ICapabilityStatementBuilder SyncProfiles();
 
-        ICapabilityStatementBuilder AddDefaultRestSearchParams();
+        ICapabilityStatementBuilder AddSharedSearchParameters();
```
brendankowitz, line 28: Should we rename/distinguish that these are `BuiltIn` or `Embedded` SearchParameters?

src/Microsoft.Health.Fhir.Core/Features/Conformance/SystemConformanceProvider.cs
```diff
@@ -14,36 +14,43 @@
...
-using Microsoft.Health.Fhir.Core.Messages.Search;
+using Microsoft.Health.Fhir.Core.Features.Validation;
+using Microsoft.Health.Fhir.Core.Messages.CapabilityStatement;
 using Microsoft.Health.Fhir.Core.Models;
 
 namespace Microsoft.Health.Fhir.Core.Features.Conformance
 {
     public sealed class SystemConformanceProvider
-        : ConformanceProviderBase, IConfiguredConformanceProvider, INotificationHandler<SearchParametersUpdated>, IDisposable
+        : ConformanceProviderBase, IConfiguredConformanceProvider, INotificationHandler<RebuildCapabilityStatement>, IDisposable
```
brendankowitz, line 24: I don't really agree that a Notification should be implied in this direction, for these points:

- If `SystemConformanceProvider` is listening/responding to notifications, then I think it should contain the responsibility of knowing when it should regenerate, based on the type of events it is interested in. i.e. `SearchParametersUpdated`
- If the notification is so specific (i.e. `RebuildCapabilityStatement`), and the generating class knew it needed to do this, it would make more sense for it to explicitly call  something like IConfiguredConformanceProvider.Regerenate()

src/Microsoft.Health.Fhir.Core/Features/Conformance/ICapabilityStatementBuilder.cs
Ivanidzo4ka: I'm not sure what does `BuiltIn` or `Embedded` means, but I know what that thing is doing!
Search parameter can be specific for an resource, but there is search parameters which are applicable for all resources, like LastUpdated or Id, or Meta. I call them Shared, I think that's are improvement over `DefaultRest`

If you talking about having difference between something set at server start and never can be changed and things which are mutable through life of server... Probably it need better name. So do some kind of proper refactoring.
At this point I'm just trying to see would this changes work in not my machine setting.

brendankowitz: Oh ok, then I think "Base" or "Global", but perhaps just a comment then :)

src/Microsoft.Health.Fhir.Core/Features/Conformance/SystemConformanceProvider.cs
Ivanidzo4ka: I agree and I prefer first approach. I was trying to get it pass tests first, and then refactor into normal state.
Didn't anticipate change of few line in json to be 37 files -_-'

Ivanidzo4ka:
>     public interface ICapabilityStatementBuilder

I'm adding comments, to this class and I don't understand why we have this interface.
I would understand if there is multiple implementations, but there is none. There is also nothing in tests to mock interface.

I'm questioning purpose of it.

---
Refers to: src/Microsoft.Health.Fhir.Core/Features/Conformance/ICapabilityStatementBuilder.cs:12 in 496c6d9. [](commit_id = 496c6d9dd9f691e6a5b7b27b6c2b01c5d10556aa, deletion_comment = False)

src/Microsoft.Health.Fhir.Core/Features/Conformance/ICapabilityStatementBuilder.cs
Ivanidzo4ka: I like `Global` more. I also sprinkle few comments around code.

---
In reply to: [606404241](https://github.com/microsoft/fhir-server/pull/1795#discussion_r606404241) [](ancestors = 606404241)

src/Microsoft.Health.Fhir.Core/Features/Conformance/IConformanceProvider.cs
```diff
@@ -12,7 +12,20 @@ namespace Microsoft.Health.Fhir.Core.Features.Conformance
 {
     public interface IConformanceProvider
     {
-        Task<ResourceElement> GetCapabilityStatementAsync(CancellationToken cancellationToken = default(CancellationToken));
+        /// <summary>
+        /// Gets capability statement builded during server start up.
```
namadabu, line 16: >builded [](start = 38, length = 7)

nit: built

src/Microsoft.Health.Fhir.Core/Features/Validation/IKnowSupportedProfiles.cs
```diff
@@ -0,0 +1,14 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+
+namespace Microsoft.Health.Fhir.Core.Features.Validation
+{
+    public interface IKnowSupportedProfiles
+    {
+        IEnumerable<string> GetSupportedProfiles(string resourceType, bool disablePull = false);
```
namadabu, line 12: >disablePull [](start = 75, length = 11)

nit: instead of `disablePull` do you think `disableCacheRefresh` might be a more descriptive name here?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/ProfileResourcesBehaviour.cs
```diff
@@ -52,12 +54,21 @@ public async Task<DeleteResourceResponse> Handle(DeleteResourceRequest request,
...
-                throw new UnauthorizedFhirActionException();
-            }
+                if (await _authorizationService.CheckAccess(DataActions.EditProfileDefinitions, cancellationToken) != DataActions.EditProfileDefinitions)
+                {
+                    throw new UnauthorizedFhirActionException();
+                }
 
-            return await next();
+                var result = await next();
+                _profilesResolver.Refresh();
```
namadabu, line 65: nit: just for my understanding, do we expect the profile refresh to happen quickly enough to not have any impact on the response times for this request?

src/Microsoft.Health.Fhir.Core/Features/Conformance/SystemConformanceProvider.cs
```diff
@@ -104,16 +112,59 @@ public void ConfigureOptionalCapabilities(Action<ListedCapabilityStatement> buil
...
-            _listedCapabilityStatement = null;
+            await _metadataSemaphore.WaitAsync(cancellationToken);
+            try
+            {
+                if (_metadata == null)
+                {
+                    _ = await GetCapabilityStatementOnStartup(cancellationToken);
 
-            return Task.CompletedTask;
+                    _metadata = _builder.Build().ToResourceElement();
```
namadabu, line 159: just for my understanding, the `_listedCapabilityStatement` now no longer contains the latest capability statement. All the updates are now made to capability statement "copy" inside the `CapabilityStatementBuilder`. Is there a reason we want to maintain two versions like that?

namadabu (approved):
:shipit:

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/ProfileResourcesBehaviour.cs
Ivanidzo4ka: It probably would add few hundred milliseconds to request, but this is for specific resources so I don't think it's critical to optimise it right now.

---
In reply to: [611309021](https://github.com/microsoft/fhir-server/pull/1795#discussion_r611309021) [](ancestors = 611309021)

src/Microsoft.Health.Fhir.Core/Features/Conformance/SystemConformanceProvider.cs
Ivanidzo4ka: I see capability statement like object which has dynamic and constant parts. Constant parts get populated once during start up and we do multiple checks against them via `IConformanceProvider.SatisfiesAsync`, dynamic parts can be changed at any time. 
Since constant part is constant we don't have to generate it all the time, so it's get initialized once via `_listedCapabilityStatement ` and just get queired all the time.
`_metadata` can be different so we reset it in `Handle` and build again in `GetMetadata` method.

We use same builder object as optimization, so we don't have to create constant part second time for metadata. 


---
In reply to: [611313190](https://github.com/microsoft/fhir-server/pull/1795#discussion_r611313190) [](ancestors = 611313190)

src/Microsoft.Health.Fhir.Core/Features/Conformance/CapabilityStatementBuilder.cs
Ivanidzo4ka: Not anymore!

---
In reply to: [605744661](https://github.com/microsoft/fhir-server/pull/1795#discussion_r605744661) [](ancestors = 605744661)

src/Microsoft.Health.Fhir.Core/Features/Conformance/SystemConformanceProvider.cs
namadabu: Thanks for the explanation!

PR #1787, opened 2021-03-26
Refactor some code in SqlQueryGenerator

namadabu:
## Description
Pushed all code under each case into separate methods to make the code more readable and easier to navigate.

## Related issues
Addresses [issue #].

## Testing
No functional changes.

## FHIR Team Checklist
- [ ] **Update the title** of the PR to be succinct and less than 50 characters
- [ ] **Add a milestone** to the PR for the sprint that it is merged (i.e. add S47)
- [ ] Tag the PR with the type of update: **Bug**, **Dependencies**, **Enhancement**, or **New-Feature**
- [ ] Tag the PR with **Azure API for FHIR** if this will release to the managed service
- Review [squash-merge requirements](https://github.com/microsoft/fhir-server/blob/master/SquashMergeRequirements.md)

### Semver Change ([docs](https://github.com/microsoft/fhir-server/blob/master/docs/Versioning.md))
Patch|Skip|Feature|Breaking (reason)

PR #1797, opened 2021-03-30
Update SQL sort query generation to support sort-with-filter scenario

namadabu:
## Description
This PR adds code to support sorting by parameters that are also present as a query parameter.
Eg: `GET <fhir-server>/Patient?family=R&_sort=family`

To handle this scenario I have added a new `SearchParamTableKind` (SortWithFilter). Updated the `SortRewriter` to walk through the expression tree to check whether we have a search parameter expression that contains the sort parameter. Updated code to create/read continuation token when we are sorting by parameter of type String (previously we supported only `DateTime`).

I have also updated the "Top" query to be optimized in case a sort expression is present. The addition of IsMin/IsMax filters as part of sort will ensure there are no repeat entries in the previous cte. Hence it is not necessary that we do a `SELECT DISTINCT TOP` in such scenarios. When there is no sort expression present, we will continue to do a `SELECT DISTINCT TOP`.

**Context:**
There are two "types" of sort queries. When we sort using a parameter that is not included in the query parameters, then the behvaior is different compared to when it is present in the query parameters.
Eg: GET <fhir-server>/Patient?_sort=family - for this type of query we will have to consider even patients that do not have family information when sorting (Kind - Sort)
Eg: GET <fhir-server>/Patient?family=R&_sort=family - for this type of query we just have to consider patients whose family name starts with R when sorting (Kind - SortWithFilter)

**Note:**

- The existing Sort (SearchParamTableKind.Sort) behaviour still needs to be updated and that will be part of a future PR.
- ~~This PR (targeting feature branch) also includes some changes that are part of other PRs (targeting main) (#1787 #1793 ).~~

## Related issues
Addresses [AB#80138](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/80138)

## Testing
Added new E2E tests.

## FHIR Team Checklist


src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -271,10 +269,18 @@ private async Task<SearchResult> SearchImpl(SearchOptions searchOptions, SqlSear
                             newContinuationId = resourceSurrogateId;
 
                             // Keep track of sort value if this is the last row.
-                            // if we have more than 10 columns, it means sort expressions were added.
+                            // if we have more than 11 columns, it means sort expressions were added.
```
Ivanidzo4ka, line 272: >11 [](start = 52, length = 2)

You just updating comment, right?
`_resourceTableColumnCount` is remaining the same, so it's a bit confusing, since that number is based on it.
Maybe it's worth change comment to base it on `_resourceTableColumnCount` instead of number?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/SortRewriter.cs
```diff
@@ -42,14 +42,50 @@ public override Expression VisitSqlRoot(SqlRootExpression expression, SearchOpti
...
             return new SqlRootExpression(newTableExpressions, expression.ResourceTableExpressions);
         }
+
+        public override Expression VisitSearchParameter(SearchParameterExpression expression, SearchOptions context)
+        {
+            if (context.Sort.Count > 0)
+            {
+                if (expression.Parameter.Equals(context.Sort[0].searchParameterInfo))
+                {
+                    return new SearchParameterExpression(expression.Parameter, expression.Expression);
```
johnstairs, line 79: This allocation is not used anywhere?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/SortRewriter.cs
```diff
@@ -42,14 +42,50 @@ public override Expression VisitSqlRoot(SqlRootExpression expression, SearchOpti
...
+                }
+            }
+
+            return expression;
+        }
+
+        public override Expression VisitMultiary(MultiaryExpression expression, SearchOptions context)
```
johnstairs, line 86: looks like this can be reverted

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -390,7 +394,8 @@ private void HandleTableKindTop(SearchOptions context)
             var sortExpression = (paramInfo == null || paramInfo.Code == KnownQueryParameterNames.LastUpdated) ? null : $"{tableExpressionName}.SortValue";
 
             // Everything in the top expression is considered a match
-            StringBuilder.Append("SELECT DISTINCT TOP (").Append(Parameters.AddParameter(context.MaxItemCount + 1)).Append(") Sid1, 1 AS IsMatch, 0 AS IsPartial ")
+            string selectTop = sortExpression == null ? "SELECT DISTINCT" : "SELECT";
```
johnstairs, line 397: nit: maybe "Top" in the name is misleading since the TOP comes in the next line :)

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -798,6 +803,74 @@ private void HandleTableKindSort(SearchParamTableExpression searchParamTableExpr
...
+                    }
+
+                    AppendIntersectionWithPredecessor(delimited, searchParamTableExpression);
+                }
+            }
+
+            _sortVisited = true;
+        }
+
+        private void HandleTableKindSortWithFilter(SearchParamTableExpression searchParamTableExpression, SearchOptions context)
```
johnstairs, line 832: This methd is like 90% the same as HandleTableKindSort. Can you consolidate a bit?

namadabu: Agreed. I did think of consolidating things in this PR. But since I am going to change the code in HandleTableKindSort (to support querying for resources with "null values for sort parameter") my plan was to do the refactoring in the next PR.

---
In reply to: [613565040](https://github.com/microsoft/fhir-server/pull/1797#discussion_r613565040) [](ancestors = 613565040)

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/SortRewriter.cs
namadabu: My idea was to use this just as a "signal" to the caller that we have a `SearchParameterExpression` with a search parameter that is the same as the sort parameter. 

---
In reply to: [613560818](https://github.com/microsoft/fhir-server/pull/1797#discussion_r613560818) [](ancestors = 613560818)

johnstairs: Right. You could use a sentinel value for that

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
johnstairs: ok

johnstairs (approved):

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
namadabu: Updated name to `selectStatement` instead of `selectTop`

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
namadabu: Updated comment. Also renamed the variable to `_defaultResourceTableFinalSelectColumnCount` from `_resourceTableColumnCount`

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/SortRewriter.cs
namadabu: Updated code to return null and avoid the extra allocation.

PR #1799, opened 2021-03-30
Remove SecretValueText

LTA-Thinking:
## Description
The property SecretValueText has been removed from the key vault secret object. This has caused secrets to not be passed to E2E tests.

## Related issues
Failed PR and CI builds

## Testing
E2E tests

## FHIR Team Checklist

Skip (deployment)

### DIFF
--- a/build/jobs/run-export-tests.yml
+++ b/build/jobs/run-export-tests.yml
@@ -28,7 +28,8 @@ jobs:
             $environmentVariableName = $secret.Name.Replace("--","_")
 
             $secretValue = Get-AzKeyVaultSecret -VaultName $keyVault -Name $secret.Name
-            Write-Host "##vso[task.setvariable variable=$($environmentVariableName)]$($secretValue.SecretValueText)"
+            $plainValue = ([System.Net.NetworkCredential]::new("", $secretValue.SecretValue).Password).ToString()
+            Write-Host "##vso[task.setvariable variable=$($environmentVariableName)]$($plainValue)"
         }
 
         $storageAccounts = Get-AzStorageAccount -ResourceGroupName $(ResourceGroupName)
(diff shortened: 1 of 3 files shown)


build/jobs/run-export-tests.yml
```diff
@@ -28,7 +28,8 @@ jobs:
             $environmentVariableName = $secret.Name.Replace("--","_")
 
             $secretValue = Get-AzKeyVaultSecret -VaultName $keyVault -Name $secret.Name
-            Write-Host "##vso[task.setvariable variable=$($environmentVariableName)]$($secretValue.SecretValueText)"
+            $plainValue = ([System.Net.NetworkCredential]::new("", $secretValue.SecretValue).Password).ToString()
```
Ivanidzo4ka, line 31: >$plainValue [](start = 12, length = 11)

I would add:
if ($plainValue) throw "Secret is empty" to catch this thing in future.

build/jobs/run-tests-aks.yml
```diff
@@ -32,7 +32,8 @@ jobs:
             $environmentVariableName = $secret.Name.Replace("--","_")
 
             $secretValue = Get-AzKeyVaultSecret -VaultName $(DeploymentEnvironmentName)-ts -Name $secret.Name
-            Write-Host "##vso[task.setvariable variable=$($environmentVariableName)]$($secretValue.SecretValueText)"
+            $plainValue = ([System.Net.NetworkCredential]::new("", $secretValue.SecretValue).Password).ToString()
```
Ivanidzo4ka, line 35: >$plainValue [](start = 12, length = 11)

Maybe it's worth to create second PR with `-AsPlainText` to see is it working or not. If it's not working we can close it, and if does, I would prefer to use offical way rather than this hack.

Ivanidzo4ka (approved):
:shipit:

build/jobs/run-tests-aks.yml
LTA-Thinking: Marius has confirmed that the Az module running right now is 5.0, and that variable isn't available until 5.3. 5.0 was released end of October and 5.3 end of December, so we could make the change in about 2 months if that is a consistent timeframe.

Ivanidzo4ka: I would add comment about this magic conversion and why we don't use `-AsPlainText`. It would probably stop working in two years, and people would scratch their heads about this line and what is going on here and why it's not using plaintext flag. 
Not insisting, tho.

Issue #1789, opened 2021-03-27
Unable to load DB schema


stephenstack:
Hi

I'm unable to load the DB schema. Set the Automatic Updates to 'True'.
![image](https://user-images.githubusercontent.com/7944609/112718288-ea0a4d80-8ee9-11eb-969b-d5fb85099b77.png)

The migration tool (the docs need a pre-req for dotnet sdk), does not work per the following. Am I not doing something correctly? I'm happy to run the migration sql files directly, but that's not intuitive either. Please help!

C:\Users\Dell>dotnet tool install -g Microsoft.Health.SchemaManager --version 1.0.0-microsoft-data-sqlclient-20200929-1
error NU1101: Unable to find package microsoft.health.schemamanager. No packages exist with this id in source(s): nuget.org
The tool package could not be restored.
Tool 'microsoft.health.schemamanager' failed to install. This failure may have been caused by:

C:\Users\Dell>dotnet tool install -g Microsoft.Health.SchemaManager
error NU1101: Unable to find package microsoft.health.schemamanager. No packages exist with this id in source(s): nuget.org
The tool package could not be restored.
Tool 'microsoft.health.schemamanager' failed to install. This failure may have been caused by:

CaitlinV39:
Thanks so much @stephenstack for the question.

@rbans96 - Can you review this?

rbans96:
Hi @stephenstack - 
If I understand correctly, you want to initialize schema on startup. And that's where you have selected 'auto' value for the field Sql schema automatic updates enabled. It is the right setting to enable auto upgrade of schema over startup. If you click on the little info icon, we can see the details for the option to select.
![image](https://user-images.githubusercontent.com/57157506/112890023-3771f080-908b-11eb-9f17-b568bc88c5b3.png)

So I am wondering, whether you proceeded with the Sql schema automatic updates enabled as 'auto' and did it work for you?

Ivanidzo4ka:
I think this issue is actually two issues. 
First is `auto` in template doesn't work as expected (still need more information what exactly is not working)

Second is package couldn't be found. I think it's problem in our documentation. We need to clearly specify what that command line need to be run from cloned repository (otherwise nuget.config which contains requisites wouldn't be picked by dotnet tool).

rbans96:
Thanks @Ivanidzo4ka! I will definitely take a look how can we make documentation better and intuitive.

[...]

PR #1800, opened 2021-03-30
Fix the schema manager tool documentation and turn app service setting 'Always On'

rbans96:
## Description
This PR address two issues
- Update the schema manager tool documentation steps to install from public feed
- When fhir service deployed with the non docker template, the default setting for the app service was off, As a result, service didn't start and waits for the first request to initialize schema. This PR updates the app service setting to 'Always on' so that the service starts up and schema is initialized itself.

## Related issues
Addresses [bug](https://github.com/microsoft/fhir-server/issues/1789)

## Testing
The updated documentation steps were tested by copying the nuget.config in the current folder and install/uninstall tool with the dotnet command.
The fhir service was deployed using the updated template and setting is turned to 'Always On'
image

## FHIR Team Checklist

### DIFF
--- a/docs/SchemaMigrationTool.md
+++ b/docs/SchemaMigrationTool.md
@@ -23,14 +23,7 @@ Note - The tool can't downgrade a schema version.
 
 - #### How to install the tool
 
-    - ##### If available as .NET Core global tool (not available yet)
-
-        It can be installed using below steps:
-
-        - Open a terminal/command prompt 
-        - Type 'dotnet tool install -g Microsoft.Health.SchemaManager'
-
-     - ##### If not available as .NET core global tool, then install it from public feed
+     - ##### Install the tool from public feed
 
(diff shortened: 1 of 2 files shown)


docs/SchemaMigrationTool.md
```diff
@@ -23,30 +23,26 @@ Note - The tool can't downgrade a schema version.
 
 - #### How to install the tool
 
-    - ##### If available as .NET Core global tool 
+    - ##### If available as .NET Core global tool (not available yet)
```
LTA-Thinking, line 26: Why do we document this yet if it isn't available?

LTA-Thinking (approved):

docs/SchemaMigrationTool.md
rbans96: Makes sense! Removed for now and I will add back when it will be available as global tool.

PR #1806, opened 2021-04-01
Hotfix for _include and _revinclude performance

johnstairs:
We have customer requests that are timing out on _revinclude searches when there are a large number of physical partitions.

~_include can be improved by performing point reads for each resource to include, rather than doing a cross-partition query.~
EDIT: not going to take this as part of the hotfix.

_revinclude can be improved by querying for the includes across physical partitions in parallel. Note that I will change the implementation for main based on the newer parallel querying approach that is not yet in production.


Ivanidzo4ka:
can you merge this PR https://github.com/microsoft/fhir-server/pull/1799 to your branch?

brendankowitz:
Moved to a hotfix branch `hotfix/cosmos-revinclude-perf`

PR #1809, opened 2021-04-02
Improve _include performance

johnstairs:
## Description
Updates the approach for retrieving included resources for `_include` searches. Instead of a query that needs to execute on all partitions, we perform point reads for each resource. 

## Testing
Manual and automated tests

Patch

### DIFF
--- a/src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
+++ b/src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
@@ -30,7 +30,6 @@ namespace Microsoft.Health.Fhir.CosmosDb.Features.Search
 {
     internal class FhirCosmosSearchService : SearchService
     {
-        private static readonly SearchParameterInfo _typeIdCompositeSearchParameter = new(SearchValueConstants.TypeIdCompositeSearchParameterName, SearchValueConstants.TypeIdCompositeSearchParameterName);
         private static readonly SearchParameterInfo _wildcardReferenceSearchParameter = new(SearchValueConstants.WildcardReferenceSearchParameterName, SearchValueConstants.WildcardReferenceSearchParameterName);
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
```diff
@@ -496,54 +494,28 @@ private SearchResult CreateSearchResult(SearchOptions searchOptions, IEnumerable
...
-                        Sort = Array.Empty<(SearchParameterInfo searchParameterInfo, SortOrder sortOrder)>(),
-                    };
-
-                    QueryDefinition includeQuery = _queryBuilder.BuildSqlQuerySpec(includeSearchOptions);
+                    // issue the requests in parallel
+                    var tasks = resourceTypeAndIds.Select(r => _fhirDataStore.GetAsync(new ResourceKey(r.ResourceTypeName, r.ResourceId), cancellationToken)).ToList();
 
-                    (IReadOnlyList<FhirCosmosResourceWrapper> results, string continuationToken) includeResponse = default;
-                    do
+                    foreach (Task<ResourceWrapper> task in tasks)
```
LTA-Thinking, line 502: Would responding to a task when it finishes, instead of waiting for the tasks to finish in order, help performance? Something like [this](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/async/start-multiple-async-tasks-and-process-them-as-they-complete).

brendankowitz: Good suggestion RB

brendankowitz: @LTA-Thinking, I think we'll take the original change first. After thinking about it, there may be an implicit order applied (i.e. lastUpdated) which may yield slightly more consistent results

brendankowitz (approved):

PR #1811, opened 2021-04-02
UpdateReindexJobAsync retries on precondition failed

rotodd:
## Description
Note: this PR won't build until [this change](https://github.com/microsoft/healthcare-shared-components/commit/1e0dc8532113a5f7fdd0692c922e91b8273c240e) from the [healthcare-shared-components repo](https://github.com/microsoft/healthcare-shared-components) is pulled in.

When we attempt to reindex a large number of resources in SQL, we run into a `JobConflictException`.

There are two things that should be done to fix this:

1) The `ReindexJobTask` should prevent multiple threads from updating the same reindex job record at the same time (there is a bug for this: [AB#80853](https://microsofthealth.visualstudio.com/Health/_workitems/edit/80853))
2) We should add retry logic to `UpdateReindexJobAsync()` that will get the updated job and retry when a `PreconditionFailedException` is thrown by the stored procedure (what this PR addresses)

## Related issues
Addresses [AB#80852](https://microsofthealth.visualstudio.com/Health/_workitems/edit/80852).

## Testing
Manual load tests following the steps outlined in `docs/ReindexFlowResourceBase.http` with a large database.

## FHIR Team Checklist

None (bug fix)

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
@@ -20,26 +20,39 @@ using Microsoft.Health.Fhir.Core.Features.Operations.Reindex.Models;
 using Microsoft.Health.Fhir.Core.Features.Persistence;
 using Microsoft.Health.Fhir.SqlServer.Features.Schema.Model;
 using Microsoft.Health.SqlServer;
+using Microsoft.Health.SqlServer.Configs;
+using Microsoft.Health.SqlServer.Extensions;
 using Microsoft.Health.SqlServer.Features.Client;
 using Microsoft.Health.SqlServer.Features.Storage;
 using Newtonsoft.Json;
+using Polly;
+using Polly.Contrib.WaitAndRetry;
+using Polly.Retry;
 
 namespace Microsoft.Health.Fhir.SqlServer.Features.Storage
 {
     internal class SqlServerFhirOperationDataStore : IFhirOperationDataStore
     {
         private readonly SqlConnectionWrapperFactory _sqlConnectionWrapperFactory;
+        private readonly IPollyRetryLoggerFactory _pollyRetryLoggerFactory;
         private readonly ILogger<SqlServerFhirOperationDataStore> _logger;
         private readonly JsonSerializerSettings _jsonSerializerSettings;
+        private readonly IEnumerable<TimeSpan> _sleepDurations;
+        private readonly PolicyBuilder _preconditionFailedPolicyBuilder;
 
         public SqlServerFhirOperationDataStore(
             SqlConnectionWrapperFactory sqlConnectionWrapperFactory,
+            SqlServerDataStoreConfiguration sqlServerDataStoreConfiguration,
+            IPollyRetryLoggerFactory pollyRetryLoggerFactory,
             ILogger<SqlServerFhirOperationDataStore> logger)
         {
             EnsureArg.IsNotNull(sqlConnectionWrapperFactory, nameof(sqlConnectionWrapperFactory));
+            EnsureArg.IsNotNull(sqlServerDataStoreConfiguration, nameof(sqlServerDataStoreConfiguration));
+            EnsureArg.IsNotNull(pollyRetryLoggerFactory, nameof(pollyRetryLoggerFactory));
             EnsureArg.IsNotNull(logger, nameof(logger));
 
             _sqlConnectionWrapperFactory = sqlConnectionWrapperFactory;
+            _pollyRetryLoggerFactory = pollyRetryLoggerFactory;
             _logger = logger;
 
             _jsonSerializerSettings = new JsonSerializerSettings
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
```diff
@@ -248,16 +292,32 @@ public async Task<ReindexJobWrapper> UpdateReindexJobAsync(ReindexJobRecord jobR
             using (SqlConnectionWrapper sqlConnectionWrapper = await _sqlConnectionWrapperFactory.ObtainSqlConnectionWrapperAsync(cancellationToken, true))
             using (SqlCommandWrapper sqlCommandWrapper = sqlConnectionWrapper.CreateSqlCommand())
             {
+                async void OnRetry(Exception exception, TimeSpan timeSpan, int i, Context context)
+                {
+                    _pollyRetryLoggerFactory.Create();
+
+                    if (exception is SqlException { Number: SqlErrorCodes.PreconditionFailed })
```
wsugarman, line 299: Ah, property patterns! I don't see this enough! Very neat

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
```diff
@@ -129,16 +157,32 @@ public async Task<ExportJobOutcome> UpdateExportJobAsync(ExportJobRecord jobReco
             using (SqlConnectionWrapper sqlConnectionWrapper = await _sqlConnectionWrapperFactory.ObtainSqlConnectionWrapperAsync(cancellationToken, true))
             using (SqlCommandWrapper sqlCommandWrapper = sqlConnectionWrapper.CreateSqlCommand())
             {
+                async void OnRetry(Exception exception, TimeSpan timeSpan, int i, Context context)
+                {
+                    _pollyRetryLoggerFactory.Create();
```
wsugarman, line 162: We still want to log, right? So we'll need to use the resulting delegate. Furthermore, can we resolve the delegate earlier in the ctor as opposed to every call?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
```diff
@@ -129,16 +157,32 @@ public async Task<ExportJobOutcome> UpdateExportJobAsync(ExportJobRecord jobReco
             using (SqlConnectionWrapper sqlConnectionWrapper = await _sqlConnectionWrapperFactory.ObtainSqlConnectionWrapperAsync(cancellationToken, true))
             using (SqlCommandWrapper sqlCommandWrapper = sqlConnectionWrapper.CreateSqlCommand())
             {
+                async void OnRetry(Exception exception, TimeSpan timeSpan, int i, Context context)
```
wsugarman, line 160: We may be able to refactor this such that the `OnRetry` object is re-used across all methods. I think the only thing we'd need to do is probably read the `Parameters` from the `Context` object as opposed to capturing it in the method as `sqlCommandWrapper.Parameters`

wsugarman: If we do that can have the resulting `AsyncRetryPolicy` as the class member as opposed to the policy builder

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlServerFhirStorageTestsFixture.cs
```diff
@@ -145,7 +145,7 @@ internal SqlServerFhirStorageTestsFixture(int maximumSupportedSchemaVersion, str
                 NullLogger<SqlServerFhirDataStore>.Instance,
                 schemaInformation);
 
-            _fhirOperationDataStore = new SqlServerFhirOperationDataStore(SqlConnectionWrapperFactory, NullLogger<SqlServerFhirOperationDataStore>.Instance);
+            _fhirOperationDataStore = new SqlServerFhirOperationDataStore(SqlConnectionWrapperFactory, config, Substitute.For<IPollyRetryLoggerFactory>(), NullLogger<SqlServerFhirOperationDataStore>.Instance);
```
wsugarman, line 148: If we're going to substitute for a `IPollyRetryLoggerFactory` in tests, we should add some assertions (like we called `Create()` n times). Otherwise we can keep it as the no-op version

wsugarman (changes requested):

rotodd:
After chatting with @namadabu and @feordin, we agreed that the SQL update methods for both reindexing and export operations should not be retrying on `PreconditionFailed` exceptions, and we should move forward with preventing these exceptions in `ReindexJobTask.cs` ([AB#80853](https://microsofthealth.visualstudio.com/Health/_workitems/edit/80853)). I created a story to revert the related changes in the healthcare-shared-components repo (please see [AB#81418](https://microsofthealth.visualstudio.com/Health/_workitems/edit/81418)).

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlServerFhirStorageTestsFixture.cs
rotodd: To summarize a bit of our discussion:

Here, we would be using a substitute of the `IPollyRetryLoggerFactory` as a dummy. Normally, substitutions involve assertions that check method invocation counts, etc. While using `NSubstitute` here is fine, it makes sense to investigate a way to distinguish scenarios where we are using `NSubstitute` to create dummy values from more traditional substitution use cases.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
rotodd: As a follow up to our discussion, I confirmed that `Create()` returns a logging Action:


```
public Action<Exception, TimeSpan, int, Context> Create()
{
    ILogger logger = _loggerFactory.CreateLogger("PollyRetryLogger");

    return (exception, sleepDuration, retryCount, context) =>
    {
        LogRetryDelegate(logger, sleepDuration, retryCount, exception);
    };
}
```

The `LogRetryDelegate` is defined as:

```
private static readonly Action<ILogger, TimeSpan, int, Exception> LogRetryDelegate =
  LoggerMessage.Define<TimeSpan, int>(
      LogLevel.Warning,
      default,
      "The operation failed. Will retry in '{SleepDuration}'. Retried {RetryCount} of time(s) so far.");
```

In `SqlServerTransientFaultRetryPolicyFactory` (the class that creates the default retry policy), we set the `OnRetry` action to `LogRetryDelegate`. If we were to move forward with this PR, we could avoid passing in `PollyRetryLoggerFactory` all together by adding a logging call to the custom `OnRetry` Actions.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
rotodd: As discussed, it makes sense to avoid setting up the `AsyncRetryPolicy` in the `OnRetry` Actions. Since the `OnRetry` Actions are operation-specific and can't be combined, we could create two custom policies that are set up in the constructor. This would prevent us from having to create policies at runtime.

Additionally, we could avoid using free variables in the `OnRetry` Actions to avoid the work that would need to be done in the background to capture the current state before `OnRetry` is executed (I used [this](https://stackoverflow.com/questions/595482/what-are-closures-in-c) to review a brief overview of closures).

Issue #1738, opened 2021-03-11
Remove legacy SearchIndexer classes


brendankowitz:
**User story**
We have moved to the [TypedElementSearchIndexer](https://github.com/microsoft/fhir-server/blob/main/src/Microsoft.Health.Fhir.Core/Features/Search/TypedElementSearchIndexer.cs), which is now used by default. It is time to remove the old [SearchIndexer](https://github.com/microsoft/fhir-server/blob/main/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchIndexer.cs), related [Converters](https://github.com/microsoft/fhir-server/tree/main/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Converters), Unit tests and [feature flag](https://github.com/microsoft/fhir-server/blob/main/src/Microsoft.Health.Fhir.Core/Configs/CoreFeatureConfiguration.cs#L28).

CaitlinV39:
@Ivanidzo4ka  - Can we close this issue?

Ivanidzo4ka:
Yes, we can!

PR #1826, opened 2021-04-06
Remove legacy SearchIndexer classes

cmelo05:
## Description
This PR removes the code related to legacy SearchIndexer.cs
- Removed SearchIndexer.cs
- Removed TypedSearchIndexer Feature Flag
- Changed the static method call GetSearchValueTypeForSearchParamType from class SearchIndexer.cs to TypedElementSearchIndexer.cs
- Removed **_all_** the Converters inside ``Features/Search/Converters`` folder, except the ones that are inside NodeConverters folders
- Removed Unit Tests from SearchIndexer and Converters

## Related issues
Addresses issue #1738 .

## Testing
I ran all the Tests and checked that the ones that are failing are due to Culture differences (e.g. expecting '.' instead of ','). E2E tests are also failing.

## FHIR Team Checklist

Patch


src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/Converters/FhirElementToSearchValueTypeConverterTests.cs
```diff
@@ -1,55 +0,0 @@
...
-using System.Linq;
-using Hl7.Fhir.Model;
-using Microsoft.Health.Fhir.Core.Features.Search.Converters;
-using Microsoft.Health.Fhir.Core.Features.Search.SearchValues;
-using Xunit;
-
-namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Search.Converters
-{
-    public abstract class FhirElementToSearchValueTypeConverterTests<TTypeConverter, TElement>
-        where TTypeConverter : IFhirElementToSearchValueTypeConverter, new()
```
Ivanidzo4ka, line 17: >IFhirElementToSearchValueTypeConverter [](start = 31, length = 38)

I can be wrong, but I don't think we have unit tests for `IFhirNodeToSearchValueTypeConverter`
Only one I can see is `ResourceReferenceNodeToReferenceSearchValueTypeConverter`

I would advice to keep tests, but move them to `IFhirNodeToSearchValueTypeConverter`

There is still E2E tests, but I'm not sure we cover all possible scenarios.

To rephase: It's ok to delete SearchIndexTests, everything else I think need to be just repointed to new converter type, instead of removing them.

Ivanidzo4ka (approved):
:shipit:

Ivanidzo4ka:
LGTM, I wouldn't mind more refactoring in term of names, like `IFhirNodeToSearchValueTypeConverter` to just `ITypedElementToSearchValueConverter` and moving content of `NodeConverterTests` from Shared.Core.UnitTests to upper `Converter` folder (no need to separate them anymore), but that can be done in separate PR.

cmelo05:
> LGTM, I wouldn't mind more refactoring in term of names, like `IFhirNodeToSearchValueTypeConverter` to just `ITypedElementToSearchValueConverter` and moving content of `NodeConverterTests` from Shared.Core.UnitTests to upper `Converter` folder (no need to separate them anymore), but that can be done in separate PR.

I updated the PR with your suggestions

Ivanidzo4ka:
Thank you for doing this!
Depends on how much time you want to spend on this, I wouldn't mind a bit more of refactoring.

But first a bit of history.
We had two different things doing same work. First we had SearchIndexer and `IFhirElementToSearchValueTypeConverter` and converters which look like: `CodeableConceptToTokenSearchValueTypeConverter`
At some point we wanted to switch to different approach, but to keep old code present, so we can flip value in config and make sure both approaches are working.
That's how we get to `IFhirNodeToSearchValueTypeConverter` and `CodeableConceptNodeToTokenSearchValueTypeConverter`
One is `Element` and another is `Node`, one is *Something*ToSearchValueTypeConverter, and another is *SomethingNode*ToSearchValueTypeConverter.

Since we removing old part of code, I don't think there is huge need to have Something*Node* because word `Node` was used mostly to differentiate between two types, but since first type is removed, it doesn't serve any purpose.

So it would be nice to take all current non-abstract converters and remove word `Node` from them, since it doesn't serve much purpose.
Tests can be renamed as well.
FhirNodeToSearchValueTypeConverter and FhirNodeToSearchValueTypeConverterManager file can be renamed to FhirTypeElementTo.... (to get rid of that Node notion)

It's not necessery, and up to your desire to mess with a code :)



---
In reply to: [814469138](https://github.com/microsoft/fhir-server/pull/1826#issuecomment-814469138) [](ancestors = 814469138)

PR #1827, opened 2021-04-07
Added logging for number of requests in flight

abhijeett:
## Description
Added log statements for tracking requests in flight. 

## Related issues
[AB#80629](https://microsofthealth.visualstudio.com/Health/_workitems/edit/80629)

## Testing
Tests pass

## FHIR Team Checklist
- [x] **Update the title** of the PR to be succinct and less than 50 characters
- [x] **Add a milestone** to the PR for the sprint that it is merged (i.e. add S47)
- [x] Tag the PR with the type of update: **Bug**, **Dependencies**, **Enhancement**, or **New-Feature**
- [x] Tag the PR with **Azure API for FHIR** if this will release to the managed service
- Review [squash-merge requirements](https://github.com/microsoft/fhir-server/blob/master/SquashMergeRequirements.md)

### Semver Change ([docs](https://github.com/microsoft/fhir-server/blob/master/docs/Versioning.md))
None(Minor logging change)

PR #1833, opened 2021-04-08
Remove string interpolation from calls to logger.

Ivanidzo4ka:
## Description
Introduce Fhir analyzer which holds one rule.
It's checking calls to ILogger.Log* methods and make sure we don't pass variable string via string interpolation.
Also clean up code to conform to that rule.

## Related issues
https://github.com/microsoft/fhir-server/pull/1827 From conversation in that PR

## Testing
Manually.

## FHIR Team Checklist

Patch


src/Microsoft.Health.Fhir.Api/Controllers/AadSmartOnFhirProxyController.cs
```diff
@@ -81,7 +81,7 @@ public AadSmartOnFhirProxyController(IOptions<SecurityConfiguration> securityCon
             {
                 if (ex is HttpRequestException || ex is OperationCanceledException)
                 {
-                    logger.LogWarning(ex, $"There was an exception while attempting to read the OpenId Configuration from \"{openIdConfigurationUrl}\".");
+                    logger.LogWarning(ex, "There was an exception while attempting to read the OpenId Configuration from \"{0}\".", openIdConfigurationUrl);
```
brendankowitz, line 84: Shouldn't this be as below? From what I've seen this would use the variable name `0`

```suggestion
                    logger.LogWarning(ex, "There was an exception while attempting to read the OpenId Configuration from \"{openIdConfigurationUrl}\".", openIdConfigurationUrl);
```

src/Microsoft.Health.Fhir.Azure/ContainerRegistry/AzureContainerRegistryAccessTokenProvider.cs
```diff
@@ -168,7 +168,7 @@ private async Task<string> GetAcrAccessToken(string registryServer, string refre
             }
             else if (!accessTokenResponse.IsSuccessStatusCode)
             {
-                _logger.LogError($"Failed to get ACR access token with ACR refresh token. Status code: {accessTokenResponse.StatusCode}.");
+                _logger.LogError("Failed to get ACR access token with ACR refresh token. Status code: {0}.", accessTokenResponse.StatusCode);
```
brendankowitz, line 171: Another example: 

```suggestion
                _logger.LogError("Failed to get ACR access token with ACR refresh token. Status code: {statusCode}.", accessTokenResponse.StatusCode);
```

feordin (approved):

src/Microsoft.Health.Fhir.Api/Controllers/AadSmartOnFhirProxyController.cs
Ivanidzo4ka: TIL.
I was under impression they just do string.Format, but that's not the case. 
Updated.

---
In reply to: [610157238](https://github.com/microsoft/fhir-server/pull/1833#discussion_r610157238) [](ancestors = 610157238)

tools/FhirAnalyzer/LogInterpolation.cs
```diff
@@ -0,0 +1,111 @@
...
+
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace FhirAnalyzer
```
johnstairs, line 14: Did you consider adding this to the shared components repo? Seems like this rule could be useful to dicom as well

tools/FhirAnalyzer/FhirAnalyzer.csproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
brendankowitz, line 4: Also if this gets packaged into a nuget we'll get flagged unless it starts with `Microsoft.`

tools/FhirAnalyzer/LogInterpolation.cs
Ivanidzo4ka: I did, and I can't figure out why it's not working, despite me doing same thing.
I agree what it's would be nice to share, but it's a bit of work to do.

I would like to commit this one in, and have issue to move it to shared component repo, than this PR just hanging here.

PR #1838, opened 2021-04-10
Refactor SqlServer SchemaUpgrade tests

namadabu:
## Description

Some of the SchemaUpgrade tests were not doing what we wanted them to do for a couple of reasons:

- The `CreateAndInitializeDatabase()` method was not respecting the `maximumSupportedSchemaVersion` if we did not pass in our own `SchemaInitializer`. It defaulted to using `SchemaVersionConstants.Max`
- The `_schemaRunner` was attached to the database that was created as part of the `FhirStorageTestsFixture` and hence was not applying the schema update to the database that we had created within the test.

It seems like we don't need the entire `SqlServerFhirStorageTestsFixture` for these tests since we are only concerned with the schema (and not any data in the server). Modified the tests to not use a fixture. Also moved `CompareDatabaseSchemas()` since it does not depend on the state of the `ISqlServerFhirStorageTestHelper` and hence seemed more appropriate as a standalone helper method.

## Related issues
Addresses [issue #].

## Testing
Verified by manually validating that the `dbo.SchemaVersion` table of the created databases contains expected values.

## FHIR Team Checklist


test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/ISqlServerFhirStorageTestHelper.cs
```diff
@@ -27,13 +27,5 @@ public interface ISqlServerFhirStorageTestHelper
...
         /// <param name="cancellationToken">The cancellation token.</param>
         Task DeleteDatabase(string databaseName, CancellationToken cancellationToken = default);
-
-        /// <summary>
-        /// Compares two database schemas.
-        /// </summary>
-        /// <param name="databaseName1">The name of the first database to compare.</param>
-        /// <param name="databaseName2">The name of the second database to compare.</param>
-        /// <returns>True if the schemas are equal, false otherwise.</returns>
-        bool CompareDatabaseSchemas(string databaseName1, string databaseName2);
```
rotodd, line 37: +1 I agree that this doesn't need to live in the storage test helper

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlServerFhirStorageTestHelper.cs
```diff
@@ -51,7 +51,7 @@ public class SqlServerFhirStorageTestHelper : IFhirStorageTestHelper, ISqlServer
         public async Task CreateAndInitializeDatabase(string databaseName, int maximumSupportedSchemaVersion, bool forceIncrementalSchemaUpgrade, SchemaInitializer schemaInitializer = null, CancellationToken cancellationToken = default)
         {
```
rotodd, line 52: Ah, so we _were_ passing in `maximumSupportedSchemaVersion` but only using it in the initialization of the `SqlServerFhirModel` 🤔

rotodd (approved):
Looks good! Thanks for cleaning this up :)

PR #1855, opened 2021-04-14
Handle cancel reindex job request in ReindexJobWorker

feordin:
## Description
Check for job cancellation inside ReindexJobTask

## Related issues
Addresses [issue [AB#81248](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/81248)].

## Testing
Integration tests updated, and manual testing performed.

## FHIR Team Checklist

Skip

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
@@ -357,7 +357,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Reindex
                 _logger.LogInformation($"Reindex throttle delay: {throttleDelayTime}");
                 await Task.Delay(_reindexJobRecord.QueryDelayIntervalInMilliseconds + throttleDelayTime, _cancellationToken);
 
-                // Remove all finished tasks from the collections of tasks_
+                // Remove all finished tasks from the collections of tasks
                 // and cancellationTokens
                 if (queryTasks.Count >= _reindexJobRecord.MaximumConcurrency)
                 {
@@ -371,25 +371,21 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Reindex
                     }
                 }
 
-                    // Get the latest version of the reindex job in case another thread has updated it
-                    // for most cases if another process updates the job (such as a DELETE request)
-                    // the _etag change will cause a JobConflict exception and this task will be aborted
-                    // but here we add one more check before attempting to mark the job as complete,
-                    // or starting another iteration of the loop
-                    await jobSemaphore.WaitAsync();
-                    try
+                // Get the latest version of the reindex job in case another thread has updated it
+                await _jobSemaphore.WaitAsync();
+                try
+                {
+                    using (IScoped<IFhirOperationDataStore> store = _fhirOperationDataStoreFactory())
                     {
-                    using (IScoped<IFhirOperationDataStore> store = _fhirOperationDataStoreFactory.Invoke())
-                        {
-                            var wrapper = await store.Value.GetReindexJobByIdAsync(_reindexJobRecord.Id, cancellationToken);
-                            _weakETag = wrapper.ETag;
-                            _reindexJobRecord = wrapper.JobRecord;
-                        }
+                        var wrapper = await store.Value.GetReindexJobByIdAsync(_reindexJobRecord.Id, _cancellationToken);
+                        _weakETag = wrapper.ETag;
+                        _reindexJobRecord = wrapper.JobRecord;
                     }
-                    finally
-                    {
-                        jobSemaphore.Release();
-                    }
+                }
+                finally
+                {
+                    _jobSemaphore.Release();
+                }
 
                 // if our received CancellationToken is cancelled, or the job has been marked canceled we should
                 // pass that cancellation request onto all the cancellationTokens
--- a/src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
@@ -371,11 +371,15 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Reindex
                     }
                 }
 
-                // Get the latest version of the reindex job in case another thread has updated it
+
+                // for most cases if another process updates the job (such as a DELETE request)
+                // the _etag change will cause a JobConflict exception and this task will be aborted
+                // but here we add one more check before attempting to mark the job as complete,
+                // or starting another iteration of the loop
                 await _jobSemaphore.WaitAsync();
                 try
                 {
-                    using (IScoped<IFhirOperationDataStore> store = _fhirOperationDataStoreFactory())
+                    using (IScoped<IFhirOperationDataStore> store = _fhirOperationDataStoreFactory.Invoke())
                     {
                         var wrapper = await store.Value.GetReindexJobByIdAsync(_reindexJobRecord.Id, _cancellationToken);
                         _weakETag = wrapper.ETag;
--- a/src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
@@ -357,7 +357,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Reindex
                 _logger.LogInformation($"Reindex throttle delay: {throttleDelayTime}");
                 await Task.Delay(_reindexJobRecord.QueryDelayIntervalInMilliseconds + throttleDelayTime, _cancellationToken);
 
-                // Remove all finished tasks from the collections of tasks
+                // Remove all finished tasks from the collections of tasks_
                 // and cancellationTokens
                 if (queryTasks.Count >= _reindexJobRecord.MaximumConcurrency)
                 {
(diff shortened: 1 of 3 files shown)


src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/CancelReindexRequestAuthHandler.cs
```diff
@@ -0,0 +1,38 @@
...
+using EnsureThat;
+using MediatR;
+using Microsoft.Health.Core.Features.Security.Authorization;
+using Microsoft.Health.Fhir.Core.Exceptions;
+using Microsoft.Health.Fhir.Core.Features.Security;
+using Microsoft.Health.Fhir.Core.Messages.Reindex;
+
+namespace Microsoft.Health.Fhir.Core.Features.Operations.Reindex
+{
+    public class CancelReindexRequestAuthHandler<TCancelReindexRequest, TCancelReindexResponse> : IPipelineBehavior<CancelReindexRequest, CancelReindexResponse>
```
Ivanidzo4ka, line 17: >IPipelineBehavior [](http://example.com/codeflow?start=98&length=17)

Why it become pipeline behaviour?

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobWorker.cs
```diff
@@ -65,8 +91,10 @@ public async Task ExecuteAsync(CancellationToken cancellationToken)
                             foreach (ReindexJobWrapper job in jobs)
                             {
                                 _logger.LogTrace("Picked up reindex job: {jobId}.", job.JobRecord.Id);
-
-                                runningTasks.Add(_reindexJobTaskFactory().ExecuteAsync(job.JobRecord, job.ETag, cancellationToken));
+                                var source = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
```
Ivanidzo4ka, line 94: >CancellationTokenSource.CreateLinkedTokenSource(cancellationToken) [](http://example.com/codeflow?start=45&length=66)

`ReindexJobWorker.ExecuteAsync` get `cancellationToken` from `ReindexJobWorkerBackgroundService` which is background service.

I assume if we stop service, that token become cancelled. By become cancelled it will trigger LinkedToken to be cancelled as well. 

Which will cancel current reindex job. Shouldn't we pause job instead of cancel it if we shutdown service?

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobWorker.cs
```diff
@@ -85,5 +113,38 @@ public async Task ExecuteAsync(CancellationToken cancellationToken)
...
+                    // If the job is already completed for any reason, return conflict status.
+                    if (outcome.JobRecord.Status.IsFinished())
+                    {
+                        throw new RequestNotValidException(
+                            string.Format(Resources.ReindexJobInCompletedState, outcome.JobRecord.Id, outcome.JobRecord.Status));
+                    }
+
+                    // Try to cancel the job.
+                    outcome.JobRecord.Status = OperationStatus.Canceled;
+                    outcome.JobRecord.CanceledTime = Clock.UtcNow;
```
Ivanidzo4ka, line 136: >outcome.JobRecord.CanceledTime = Clock.UtcNow; [](http://example.com/codeflow?start=20&length=46)

Can we have situation where handle of the job and request to cancel job executes on different machines?

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobWorker.cs
feordin: You are correct that we will not want to mark the job as cancelled.  We do want to shutdown any running tasks, which is the intent of linking the cancellation tokens.  I will double check that we not update the job record to Canceled unless that is the intent.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobWorker.cs
feordin: Excellent point!  I will add some additional checks in the task class.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/CancelReindexRequestAuthHandler.cs
feordin: I moved most of the logic to the ReindexJobWorker as it may contain a handle to the cancellationToken.  But the authorization piece needs to run first, so auth is checked here before the ReindexJobWorker handles the remainder of the logic.  I could not include the authorization code in the ReindexJobWorker due to initialization dependencies, but this class is only instantiated upon a DELETE request, so initialization will have completed before then.

Ivanidzo4ka: I think it's a bit of overkill to add behaviour.
It just looks strange to have auth seprated only for cancel request?
I would understand if that would serve all reindex requests, but I think it's better to go with abstract class derived from IPipelineHandler<TRequest,TResponse> where TRequest: ReindexRequest, TResponse: ReindexResponse and inherit all other reindex handlers from it.

Moving handler into job worker feels breaking general approach for the sake of optimization for case of one machine, and I'm not sure it's worth it.

feordin: Agreed, I have simplified the fix to simply check the value in the datastore.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -283,15 +283,34 @@ public async Task ExecuteAsync(ReindexJobRecord reindexJobRecord, WeakETag weakE
                         }
                     }
 
-                    // if our received CancellationToken is cancelled we should
+                    // Get the latest version of the reindex job in case another thread has updated it
```
Ivanidzo4ka, line 286: >Get the latest version of the reindex job in case another thread has updated it [](http://example.com/codeflow?start=23&length=79)

I would drop few words why we doing it here and only here considering there is plenty of places where we update job record.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -283,15 +283,34 @@ public async Task ExecuteAsync(ReindexJobRecord reindexJobRecord, WeakETag weakE
...
+                    await jobSemaphore.WaitAsync();
+                    try
+                    {
+                        using (IScoped<IFhirOperationDataStore> store = _fhirOperationDataStoreFactory())
+                        {
+                            var wrapper = await store.Value.GetReindexJobByIdAsync(_reindexJobRecord.Id, cancellationToken);
+                            _weakETag = wrapper.ETag;
```
Ivanidzo4ka, line 293: >_weakETag = wrapper.ETag; [](http://example.com/codeflow?start=28&length=25)

Why this is necessary? In case if no one put an eye on this job, etag would be the same, and in case job been cancelled tag doesn't matter anymore.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
feordin: I have added a comment.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
feordin: It is true.  This is more of a just in case type check and update.

Ivanidzo4ka (approved):
:shipit:

Ivanidzo4ka:
>             try

just a rant, but this try section starts here and ends in line 379 which is almost 300 lines long. Even if I turn my 32" monitor sideways, I think it wouldn't fit it. 

---
Refers to: src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs:92 in 15d600e. [](commit_id = 15d600e6a7e79c8a6cfe3b67c12bc87c27271579, deletion_comment = False)

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -357,7 +357,7 @@ private async Task ProcessJob()
                 _logger.LogInformation($"Reindex throttle delay: {throttleDelayTime}");
                 await Task.Delay(_reindexJobRecord.QueryDelayIntervalInMilliseconds + throttleDelayTime, _cancellationToken);
 
-                // Remove all finished tasks from the collections of tasks
+                // Remove all finished tasks from the collections of tasks_
```
Ivanidzo4ka, line 360: >_ [](http://example.com/codeflow?start=74&length=1)

nit: extra _

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -371,15 +371,38 @@ private async Task ProcessJob()
...
+                    using (IScoped<IFhirOperationDataStore> store = _fhirOperationDataStoreFactory.Invoke())
+                        {
+                            var wrapper = await store.Value.GetReindexJobByIdAsync(_reindexJobRecord.Id, cancellationToken);
+                            _weakETag = wrapper.ETag;
+                            _reindexJobRecord = wrapper.JobRecord;
+                        }
+                    }
+                    finally
+                    {
+                        jobSemaphore.Release();
```
Ivanidzo4ka, line 391: >jobSemaphore [](http://example.com/codeflow?start=24&length=12)

should be _jobSemaphore and _cancellationToken. Sorry for messing with that code

PR #1876, opened 2021-04-21
Introduces table partitioning

johnstairs:
## Description
This change brings in [table partitioning](https://docs.microsoft.com/en-us/sql/relational-databases/partitions/partitioned-tables-and-indexes?view=sql-server-ver15) by resource type to the Resource and search parameter tables.

The nonclustered indexes are also being tweaked to add the ResourceSurrogateId as the trailing index column. This makes searches that perform equality filters (like `?code=abc`) more efficient because the data is already sorted in the non-clustered index. Previously, these searches would either perform a clustered index table scan or would would use the non-clustered index but would need to sort the entire resultset.

This change required changes to the query generation to use the resource type id in more parts of the queries, and also required special work to make searches that span multiple resource types efficient.

The default sort order is now `_type,_lastUpdated` instead of `_lastUpdated`. In cases where `_lastUpdated` is specified, we can turn it into `_type,_lastUpdated` if the search is over one resource type. But queries that explicitly sort by `_lastUpdated` and are over multiple resource types with criteria over search parameter tables will not be efficient.

The schema migration (9.diff.sql) is "online" (does not require downtime) but is very slow. For a database with 54 million resources (the migration took around 10 hours on a General Purpose Azure SQL database). For that reason, we are separately working on an alternate script that will be faster but will require a maintenance window.

## Testing
Manual performance testing. ~Unit tests coming in a subsequent update to this PR.~ Unit tests added for new visitors.

Breaking (searches that specify _sort=_lastUpdated AND are over multiple resources types will not always run efficiently)


src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/9.diff.sql
```diff
@@ -0,0 +1,1875 @@
+/*************************************************************
+    This migration introduces table partitioning by ResourceTypeId on the Resource and all search parameter tables.
```
namadabu, line 2: I am not familiar with partitioned tables, but the [docs](https://docs.microsoft.com/en-us/sql/relational-databases/partitions/partitioned-tables-and-indexes?view=sql-server-ver15#benefits-of-partitioning) mention that setting `LOCK_ESCALATION = AUTO` can improve performance. Would it be helpful for us to do that?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSortingValidator.cs
```diff
@@ -47,8 +65,20 @@ public bool ValidateSorting(IReadOnlyList<(SearchParameterInfo searchParameter,
...
+                    {
+                        errorMessages = new[] { Resources.TypeAndLastUpdatedMustHaveSameSortDirection };
+                        return false;
+                    }
+
+                    errorMessages = Array.Empty<string>();
+                    return true;
                 default:
-                    errorMessages = new[] { Core.Resources.MultiSortParameterNotSupported };
+                    errorMessages = new[] { Resources.OnlyTypeAndLastUpdatedSupportedForCompountSort };
```
namadabu, line 81: >CompountSort [](start = 88, length = 12)

nit: typo, CompoundSort

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/PartitionEliminationRewriter.cs
```diff
@@ -0,0 +1,186 @@
...
+                if (hasTypeRestriction)
+                {
+                    // This is already constrained to be one or more resource types.
+                    return expression;
+                }
+
+                if (_schemaInformation.Current < SchemaVersionConstants.PartitionedTables)
+                {
+                    return expression;
+                }
```
namadabu, line 105: nit: Is there a reason we don't want this check at the beginning of the method?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -385,18 +414,62 @@ private void HandleTableKindNotExists(SearchParamTableExpression searchParamTabl
 
         private void HandleTableKindTop(SearchOptions context)
         {
-            var (paramInfo, sortOrder) = context.Sort.Count == 0 ? default : context.Sort[0];
             var tableExpressionName = TableExpressionName(_tableExpressionCounter - 1);
-            var sortExpression = (paramInfo == null || paramInfo.Code == KnownQueryParameterNames.LastUpdated) ? null : $"{tableExpressionName}.SortValue";
+            bool isSpecialCaseSort = IsPrimaryKeySort(context);
+            var sortExpression = isSpecialCaseSort ? null : "SortValue";
```
namadabu, line 419: >"SortValue" [](start = 60, length = 11)

Shoud this be `"{tableExpressionName}.SortValue"` instead of `"SortValue"` ?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -295,32 +314,40 @@ private void HandleTableKindNormal(SearchParamTableExpression searchParamTableEx
...
+                        delimited.BeginDelimitedElement().Append(VLatest.Resource.ResourceTypeId, null).Append(" = ").Append(cte).Append(".T1");
                         delimited.BeginDelimitedElement().Append(VLatest.Resource.ResourceSurrogateId, null).Append(" = ").Append(cte).Append(".Sid1");
                     }
                 }
             }
             else
             {
-                StringBuilder.Append("SELECT Sid1, ").Append(VLatest.Resource.ResourceSurrogateId, null).AppendLine(" AS Sid2")
+                StringBuilder.Append("SELECT T1, Sid1, ")
+                    .Append(VLatest.Resource.ResourceTypeId, null).AppendLine(" AS T2, ")
```
namadabu, line 343: Just for my understanding - now we are generating queries where we are selecting the ResourceTypeId in all cases. Will this have a performance impact on tables that are using schemas older than `SchemaVersionConstants.PartitionedTables`?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/9.diff.sql
johnstairs: Great catch! Thanks.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/PartitionEliminationRewriter.cs
johnstairs: Good point.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
johnstairs: The reason I took it out is that it's not permitted in the window function `ROW_NUMBER() OVER(ORDER BY SortValue ASC, Sid1 ASC)` and it shouldn't be ambiguous in this context (there is only one CTE).

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
johnstairs: `ResourceTypeId` was always an included column in nonclustered indexes, and we were already using in `WHERE` clauses, so these new queries should have very similar performance.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/9.diff.sql
johnstairs: Added in https://github.com/microsoft/fhir-server/pull/1876/commits/c0d33abae9b6358959fb104c8e7108ce01a8b4b8

src/Microsoft.Health.Fhir.Core/Features/Search/ContinuationToken.cs
```diff
@@ -36,6 +36,19 @@ public long ResourceSurrogateId
             }
         }
 
+        public short? ResourceTypeId
+        {
+            get
+            {
+                return _tokens.Length > 1 && _tokens[^2] is long s ? (short)s : null;
```
Ivanidzo4ka, line 43: >is long s [](http://example.com/codeflow?start=57&length=9)

Didn't checkout branch, but from changes in code it looks like we only initilaze token with:

```
continuationToken = new ContinuationToken(new object[]
                            {
                                newContinuationType ?? 0,
                                newContinuationId ?? 0,
                            }
```

where `short? newContinuationType`

Why we cast it to long first?

johnstairs: Hmm. So the JSON is deserialized as a long, which we then need to cast to a short. But you are right, the property won't work if it's an object that is created normally though the constructor.

src/Microsoft.Health.Fhir.Api/Features/Routing/UrlResolver.cs
```diff
@@ -168,10 +169,40 @@ public Uri ResolveRouteUrl(IEnumerable<Tuple<string, string>> unsupportedSearchP
...
+
+                                    if (sortOrder == SortOrder.Descending)
+                                    {
+                                        sb.Append('-');
+                                    }
+
+                                    sb.Append(searchParameterInfo.Name);
+                                }
+
+                                routeValues.Add(searchParam.Key, sb.ToString());
```
Ivanidzo4ka, line 204: >routeValues [](http://example.com/codeflow?start=32&length=11)

```
routeValues.Add(searchParam.Key, string.Join(',' resultSortOrder.Select(s => $"{(s.sortOrder == SortOrder.Ascending ? string.Empty : "-")}{s.searchParameterInfo.Code}")
```
to replace previous code, this one looks a bit chunky.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/PrimaryKeyValue.cs
```diff
@@ -0,0 +1,15 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Fhir.SqlServer.Features.Search
+{
+    /// <summary>
+    /// Represents a row in the Resource or search parameter tables by its primary key.
+    /// </summary>
+    internal record PrimaryKeyValue(short ResourceTypeId, long ResourceSurrogateId)
```
Ivanidzo4ka, line 11: >record [](http://example.com/codeflow?start=13&length=6)

C# 9! woohoo!

johnstairs: I think the Expression classes could benefit from becoming records one day. In many cases, we just need to update one property, so `with` expressions could be quite nice.

src/Microsoft.Health.Fhir.Api/Features/Routing/UrlResolver.cs
johnstairs: Yeah. True.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/SortRewriter.cs
```diff
@@ -35,9 +35,9 @@ public override Expression VisitSqlRoot(SqlRootExpression expression, SearchOpti
                 return expression;
             }
 
-            // _lastUpdated sort param is handled differently than others, because it can be
+            // _type and _lastUpdated sort param is handled differently than others, because it can be
```
Ivanidzo4ka, line 38: >it [](http://example.com/codeflow?start=93&length=2)

super nit: probaby `they`

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/TypeConstraintVisitor.cs
```diff
@@ -0,0 +1,138 @@
...
+
+namespace Microsoft.Health.Fhir.SqlServer.Features.Search.Expressions.Visitors
+{
+    /// <summary>
+    /// Looks for _type parameters to determine the set of ResourceTypeIds allowed in an expression.
+    /// Assumes that <see cref="ResourceColumnPredicatePushdownRewriter"/> has not run yet on the expression.
+    /// </summary>
+    internal class TypeConstraintVisitor : DefaultSqlExpressionVisitor<(BitArray allowedTypes, ISqlServerFhirModel model), short?>
+    {
+        internal const short NoTypes = -1;
```
Ivanidzo4ka, line 21: >NoTypes [](http://example.com/codeflow?start=29&length=7)

nit: 

It feels a bit strange what it's not used outside of this class (tests of this class is exception).

I mean al this work to indicate what there is no types, and we will ignore that and use allAllowTypes to find what is supported types.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -200,17 +204,32 @@ public override object VisitSqlRoot(SqlRootExpression expression, SearchOptions
...
+                    if (_rootExpression.SearchParamTableExpressions.Any(t => t.Kind == SearchParamTableExpressionKind.Include))
                     {
-                        StringBuilder
-                            .Append(VLatest.Resource.ResourceSurrogateId, resourceTableAlias).Append(" ")
-                            .AppendLine(sortOrder == SortOrder.Ascending ? "ASC" : "DESC");
+                        // ensure the matches appear before includes
+                        StringBuilder.Append("IsMatch DESC, ");
+                    }
+
+                    if (isSpecialCaseSort)
```
Ivanidzo4ka, line 214: >isSpecialCaseSort [](http://example.com/codeflow?start=24&length=17)

if previous if condition is true and this one as well
we will have double comma.

`IsMatch Desc,,`

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -853,10 +931,12 @@ private void AppendIntersectionWithPredecessor(IndentedStringBuilder.DelimitedSc
             {
                 delimited.BeginDelimitedElement();
 
-                string columnToSelect = (searchParamTableExpression.Kind == SearchParamTableExpressionKind.Chain ? searchParamTableExpression.ChainLevel - 1 : searchParamTableExpression.ChainLevel) == 0 ? "Sid1" : "Sid2";
+                bool intersectWithFirst = (searchParamTableExpression.Kind == SearchParamTableExpressionKind.Chain ? searchParamTableExpression.ChainLevel - 1 : searchParamTableExpression.ChainLevel) == 0;
 
-                StringBuilder.Append(VLatest.Resource.ResourceSurrogateId, tableAlias).Append(" IN (SELECT ").Append(columnToSelect)
-                    .Append(" FROM ").Append(TableExpressionName(predecessorIndex)).Append(")");
+                StringBuilder.Append("EXISTS(SELECT * FROM ").Append(TableExpressionName(predecessorIndex))
```
Ivanidzo4ka, line 936: >* [](http://example.com/codeflow?start=52&length=1)

I wonder, does SQL optimize this and check only where clause fields , or it's actually reads content of whole row just to perform exist on it....

johnstairs: It optimizes. It will use the best index, which does not have to include all columns.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/TypeConstraintVisitor.cs
johnstairs: By returning `NoTypes`, we are not returning null, where null means many types. When we return NoTypes, we don't touch `allowedTypes`.

Ivanidzo4ka: My bad, you right.
Still a bit weird to return type id and use it to only as indication of using BitArray or not.

Considering what BitArray is class we can just return null for bit array instead of returning null and bitarray.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -164,6 +187,9 @@ private async Task<SearchResult> SearchImpl(SearchOptions searchOptions, SqlSear
                 }
             }
 
+            var originalSearchOptions = searchOptions;
```
Ivanidzo4ka, line 190: >var originalSearchOptions = searchOptions; [](http://example.com/codeflow?start=12&length=42)

since it's uses only `originalSearchOptions.Sort` maybe it should be
`var originalSort = searchOption.Sort`?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -333,11 +353,121 @@ private async Task<SearchResult> SearchImpl(SearchOptions searchOptions, SqlSear
...
+            if (sqlSearchType == SqlSearchType.History)
+            {
+                // history is always sorted by _lastUpdated.
+                searchOptions = searchOptions.Clone();
+
+                ISearchParameterDefinitionManager searchParameterDefinitionManager = _searchParameterDefinitionManagerResolver.Invoke();
+
+                searchOptions.Sort = new (SearchParameterInfo searchParameterInfo, SortOrder sortOrder)[]
+                {
+                    (searchParameterDefinitionManager.GetSearchParameter(KnownResourceTypes.Resource, SearchParameterNames.LastUpdated), SortOrder.Ascending),
```
Ivanidzo4ka, line 381: >searchParameterDefinitionManager.GetSearchParameter(KnownResourceTypes.Resource, SearchParameterNames.LastUpdated [](http://example.com/codeflow?start=21&length=113)

is it worth to resolve this and `ResourceType` in constuctor as property and reuse it?

Ivanidzo4ka (approved):
:shipit:

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
johnstairs: It would be nice, but this is currently not registered as a singleton. (but it should be)

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/TypeConstraintVisitor.cs
johnstairs: Done.

PR #1877, opened 2021-04-22
Optimizes exact string search on SQL

johnstairs:
## Description
We can achieve better query efficiency for exact string queries, such as `Patient?address-state:exact=Massachusetts`, by adding an additional predicate to the WHERE clause:

image

Adding in the case/accent-insensitive predicate allows the optimizer to choose an index seek (and that now has the results pre-sorted by ResourceSurrogateId).

In the example, above, the query plan goes from this:
image

To this:
image

## Testing
Manual testing

Patch


LTA-Thinking:
I'm confused by the example. It looks like this would add a duplicate AND statement.

Ivanidzo4ka:
> I'm confused by the example. It looks like this would add a duplicate AND statement.

It does two comparison, first one is just compare strings based on default collation  for column (`TextOverflow nvarchar(max) COLLATE Latin1_General_100_CI_AI_SC` - Case insensitive, Accent insensitive ) and second compare string based on specified collation.

if you have text like
```
A
a
Á
```
and you looking for `a`
first condition would pick all three rows, while second will filter out all except `a`.

Ivanidzo4ka (approved):
:shipit:

namadabu (approved):
:shipit:

PR #1880, opened 2021-04-23
Improves efficiency of normal string searches

johnstairs:
## Description

SQL Server does not support arbitrarily large index widths. For that reason, we have a truncated `Text` column on the `StringSearchParam` table, and a `TextOverflow` column in cases where the text does not fit in the `Text` column.

With this change, we are tweaking a nonclustered index to not be filtered on `WHERE TextOverflow IS NULL` which allows for simpler and more efficient queries. Example (for `/Patient?family=Breitenberg`):

image

For the query  `/Patient?family=Breitenberg`, logical reads are reduced from 10 to 5 on the `StringSearchParam` table.

The difference becomes more important for a query like `/Patient?family=Breitenberg&gender=male` when the optimizer chooses a bad plan (it does an inner loop join with `males` on in the outer loop in my database). There logical reads are reduced from 32,821 to 16,429.


namadabu (approved):
:shipit:

PR #1882, opened 2021-04-25
Avoid recompiling every SQL query

johnstairs:
## Description

We currently add `OPTION(RECOMPILE)` to every query we generate for searches. The reason for this is that the optimizer chooses different query plans based on parameter values and sometimes there can be an enormous differences in performance between plans.

But there is a cost to recompiling queries:

|                      | Logical reads WITHOUT caching | Duration WITHOUT caching (ms) | Logical reads WITH cache hit| Duration WITH cache hit (ms) |
|----------------------|-------------------------------|-------------------------------|----------------------------|---------------------------|
| /Patient?gender=male | 191                           | 15                            | 39                         | 2                         |
| /Patient?family=Smith&birthdate=1982-11-05| 631                           | 57                            | 21                         | 3                         |
| /MedicationRequest?subject:Patient.identifier=999-60-8644 | 382                           | 32                            | 3                          | 4                         |

This PR makes a first change in avoiding recompilations for every query. We remove the `OPTION(RECOMPILE)` hint, and instead add a hash of a subset of the parameter values in a comment. This means that two identical searches will have the same query text and can use the same query plan. 

Excluded from the hashes are `TOP` values and parameters related to continuation tokens (so we don't require recompiling for every page of a query).

We'll want to to more here; this is just a first step. A promising option is 
[`OPTIMIZE FOR ( @variable_name { UNKNOWN | = <literal_constant> } [ , ...n ] ) ` ](https://docs.microsoft.com/en-us/sql/t-sql/queries/hints-transact-sql-query?view=sql-server-ver15).

## Testing
Manual and unit tests

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Search/HashingSqlQueryParameterManager.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Search/HashingSqlQueryParameterManager.cs
@@ -1,0 +1,201 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.Data;
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+using System.Security.Cryptography;
+using EnsureThat;
+using Microsoft.Data.SqlClient;
+using Microsoft.Health.SqlServer;
+using Microsoft.Health.SqlServer.Features.Schema.Model;
+using Microsoft.Health.SqlServer.Features.Storage;
+
+namespace Microsoft.Health.Fhir.SqlServer.Features.Search
+{
+    /// <summary>
+    /// Wraps a <see cref="SqlQueryParameterManager"/>, adding the ability to compute a hash of a subset of the parameters.
+    /// </summary>
+    public class HashingSqlQueryParameterManager
+    {
+        private readonly SqlQueryParameterManager _inner;
+        private readonly HashSet<SqlParameter> _setToHash = new();
+
+        public HashingSqlQueryParameterManager(SqlQueryParameterManager inner)
+        {
+            EnsureArg.IsNotNull(inner, nameof(inner));
+            _inner = inner;
+        }
+
+        public bool HasParametersToHash => _setToHash.Count > 0;
+
+        public SqlParameter AddParameter<T>(Column<T> column, T value, bool includeInHash = true)
+        {
+            return AddParameter((Column)column, value, includeInHash);
+        }
+
+        public SqlParameter AddParameter(Column column, object value, bool includeInHash = true)
+        {
+            SqlParameter parameter = _inner.AddParameter(column, value);
+            if (includeInHash)
+            {
+                _setToHash.Add(parameter);
+            }
+
+            return parameter;
+        }
+
+        public SqlParameter AddParameter(object value, bool includeInHash = true)
+        {
+            SqlParameter parameter = _inner.AddParameter(value);
+            if (includeInHash)
+            {
+                _setToHash.Add(parameter);
+            }
+
+            return parameter;
+        }
+
+        /// <summary>
+        /// Appends a Base64-encoded SHA-256 hash of the parameters currently added to this instance with includeInHash = true
+        /// </summary>
+        /// <param name="stringBuilder">A string builder to append the hash to.</param>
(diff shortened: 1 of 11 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Search/HashingSqlQueryParameterManager.cs
```diff
@@ -0,0 +1,201 @@
...
+        /// <param name="stringBuilder">A string builder to append the hash to.</param>
+        public void AppendHash(IndentedStringBuilder stringBuilder)
+        {
+            IncrementalHash incrementalHash = null;
+            Span<byte> buf = stackalloc byte[256];
+            int currentBufferIndex = 0;
+
+            foreach (SqlParameter sqlParameter in _setToHash)
+            {
+                if (!_setToHash.Contains(sqlParameter))
```
namadabu, line 76: >if (!_setToHash.Contains(sqlParameter)) [](start = 16, length = 39)

Is there a reason we have this check here since we looping through `_setToHash` ?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/HashingSqlQueryParameterManager.cs
```diff
@@ -0,0 +1,201 @@
...
+        }
+
+        public bool HasParametersToHash => _setToHash.Count > 0;
+
+        public SqlParameter AddParameter<T>(Column<T> column, T value, bool includeInHash = true)
+        {
+            return AddParameter((Column)column, value, includeInHash);
+        }
+
+        public SqlParameter AddParameter(Column column, object value, bool includeInHash = true)
```
namadabu, line 42: >bool includeInHash = true) [](start = 70, length = 26)

nit: would it be better to make this a non-optional parameter so that callers explicitly set `includeInHash`. I feel that having a default value in this situation makes it easy for someone in the future to introduce a bug by calling `AddParameter()` without realizing what the default behavior does.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/HashingSqlQueryParameterManager.cs
johnstairs: Ha! yes, good point

src/Microsoft.Health.Fhir.SqlServer/Features/Search/HashingSqlQueryParameterManager.cs
```diff
@@ -0,0 +1,201 @@
...
+            stringBuilder.Append(hashChars[..hashCharsLength]);
+        }
+
+        private static void WriteAndAdvance<T>(Span<byte> buffer, ref int currentIndex, ref IncrementalHash incrementalHash, T element)
+            where T : struct
+        {
+            int elementLength = Unsafe.SizeOf<T>();
+            if (currentIndex + elementLength > buffer.Length)
+            {
+                incrementalHash ??= IncrementalHash.CreateHash(HashAlgorithmName.SHA256);
```
namadabu, line 166: >HashAlgorithmName.SHA256 [](start = 63, length = 24)

nit: since security is not important here, would it be helpful to use a faster non-cryptographic hash algorithm? I guess the length of the data to hash is small enough here that it shouldn't really have an impact.

namadabu (changes requested):
:clock1: - edit: ignore this, clicked by mistake

namadabu (approved):
:shipit:

src/Microsoft.Health.Fhir.SqlServer/Features/Search/HashingSqlQueryParameterManager.cs
johnstairs: Yes, this is a good point. I tried out xxHash using [this ](https://github.com/uranium62/xxHash) implementation.

With parameters `@p0=690,@p1='1982-11-05 00:00:00',@p2='1982-11-05 23:59:59.9999999',@p3=103,@p4=1016,@p5=N'Smith%',@p6=11`, here was the difference:

| Method |       Mean |    Error |   StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|------- |-----------:|---------:|---------:|------:|------:|------:|----------:|
|      Sha256 | 1,155.3 ns | 22.65 ns | 35.92 ns |     - |     - |     - |         - |
|      xxHash |   268.6 ns |  5.30 ns |  4.96 ns |     - |     - |     - |         - |

But I'd like to hold off on making this change though, until we evaluate the package.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/HashingSqlQueryParameterManager.cs
johnstairs: Good idea. Updated.

PR #1884, opened 2021-04-26
Implement Patient $everything operation in Cosmos DB

BoyaWu10:
Implement Patient $everything operation in Cosmos DB.

## Description
Customers can make a [base]/Patient/[id]/$everything call against the FHIR Server and get the data related to a patient.
The design spec could be found [here](https://microsofthealth.visualstudio.com/Health/_git/health-paas-docs/pullrequest/15114).

## Related issues
[#81387](https://microsofthealth.visualstudio.com/Health/_workitems/edit/81387)

## Testing
Unit tests and E2E tests are added to cover the workflow.

## FHIR Team Checklist


src/Microsoft.Health.Fhir.Api/Features/Routing/KnownRoutes.cs
```diff
@@ -52,6 +52,9 @@ internal class KnownRoutes
         public const string ConvertData = "$convert-data";
         public const string ConvertDataOperationDefinition = OperationDefinition + "/" + OperationsConstants.ConvertData;
 
+        public const string Everything = "$everything";
+        public const string EverythingResourceTypeById = ResourceTypeById + "/" + Everything;
```
Ivanidzo4ka, line 56: >EverythingResourceTypeById [](http://example.com/codeflow?start=28&length=26)

I think it's better restict on route level, than catch wrong resource type in controller.

So I would suggest to have 
`"Pateint"+"/"+IdRouteSegment+"/"+Everithing`

src/Microsoft.Health.Fhir.Api/Features/Routing/RouteNames.cs
```diff
@@ -39,6 +39,8 @@ internal static class RouteNames
 
         internal const string PostBundle = "PostBundle";
 
+        internal const string EverythingOperationById = "EverythingOperationById";
```
Ivanidzo4ka, line 42: >EverythingOperationById [](http://example.com/codeflow?start=30&length=23)

it doesn't do much, other than provide name for attribute.
Since this is operation, we need to provide information about that operation as well.

See `ResolveOperationDefinitionUrl` in `UrlResolver` and 
`src\Microsoft.Health.Fhir.Core/Data/OperationDefinition` folder with json files.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/OperationController.cs
```diff
@@ -0,0 +1,64 @@
...
+using Microsoft.Health.Fhir.Core.Models;
+using Microsoft.Health.Fhir.ValueSets;
+
+namespace Microsoft.Health.Fhir.Api.Controllers
+{
+    [ServiceFilter(typeof(AuditLoggingFilterAttribute))]
+    [ServiceFilter(typeof(OperationOutcomeExceptionFilterAttribute))]
+    [ValidateResourceTypeFilter]
+    [ValidateModelState]
+    public class OperationController : Controller
```
Ivanidzo4ka, line 29: >OperationController [](http://example.com/codeflow?start=17&length=19)

I think right now we put every spec defined functionallity into FhirContoller, and everything not aligned with spec in separate controllers (Export, Reindex).

I would move it to FhirController.

src/Microsoft.Health.Fhir.Core/Messages/Operation/EverythingOperationRequest.cs
```diff
@@ -0,0 +1,51 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using EnsureThat;
+using MediatR;
+using Microsoft.Health.Fhir.Core.Models;
+
+namespace Microsoft.Health.Fhir.Core.Messages.Operation
```
Ivanidzo4ka, line 12: >Operation [](http://example.com/codeflow?start=46&length=9)

I would create it's own namespace and folder for `Everything`

There is two types of files right now in this namespace.
One to handle requests to get information about non-defined spec operation (Reindex, Convert) and validate operation requests, which I put here by mistake.

src/Microsoft.Health.Fhir.Core/Features/Search/SearchService.cs
```diff
@@ -22,7 +22,6 @@ namespace Microsoft.Health.Fhir.Core.Features.Search
     /// </summary>
     public abstract class SearchService : ISearchService
     {
-        private readonly ISearchOptionsFactory _searchOptionsFactory;
```
Ivanidzo4ka, line 25: >ISearchOptionsFactory [](http://example.com/codeflow?start=25&length=21)

why?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/EverythingOperationTests.cs
```diff
@@ -0,0 +1,152 @@
...
+
+            await ExecuteAndValidateBundle(searchUrl, true, 4, Fixture.Patient, Fixture.Organization, Fixture.Device, Fixture.Observation, Fixture.Appointment, Fixture.Encounter);
+        }
+
+        [Fact]
+        [Trait(Traits.Priority, Priority.One)]
+        [HttpIntegrationFixtureArgumentSets(DataStore.CosmosDb, Format.Json)]
+        public async Task GivenAPatientEverythingOperationWithNonExistentId_WhenSearched_ThenResourcesInScopeShouldBeReturned()
+        {
+            string searchUrl = $"Patient/{Fixture.NonExistentPatient.Id}/$everything";
```
Ivanidzo4ka, line 40: >Fixture.NonExistentPatient.Id [](http://example.com/codeflow?start=42&length=29)

[https://www.hl7.org/fhir/patient-operation-everything.html](https://www.hl7.org/fhir/patient-operation-everything.html)

>At a minimum, the patient resource(s) itself is returned
>The intended use for this operation is to provide a patient with access to their entire record

From what I'm reading in operation definition we should return `Not found` here.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/EverythingOperationTests.cs
```diff
@@ -0,0 +1,152 @@
...
+        [Fact]
+        [Trait(Traits.Priority, Priority.One)]
+        [HttpIntegrationFixtureArgumentSets(DataStore.CosmosDb, Format.Json)]
+        public async Task GivenAEverythingOperationWithUnsupportedType_WhenSearched_ThenBadRequestShouldBeReturned()
+        {
+            string searchUrl = "Observation/bar/$everything";
+
+            using FhirException ex = await Assert.ThrowsAsync<FhirException>(() => Client.SearchAsync(searchUrl));
+
+            Assert.Equal(HttpStatusCode.BadRequest, ex.StatusCode);
```
Ivanidzo4ka, line 54: >BadRequest [](http://example.com/codeflow?start=40&length=10)

it should be `not found` since `$everything` defined only for patient resource.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/EverythingOperationTests.cs
```diff
@@ -0,0 +1,152 @@
...
+
+            await ExecuteAndValidateBundle(searchUrl, true, 4, Fixture.Patient, Fixture.Organization, Fixture.Device, Fixture.Appointment, Fixture.Encounter);
+        }
+
+        [Fact]
+        [Trait(Traits.Priority, Priority.One)]
+        [HttpIntegrationFixtureArgumentSets(DataStore.CosmosDb, Format.Json)]
+        public async Task GivenSinceSpecified_WhenSearched_ThenResourcesOfSpecifiedRangeShouldBeReturned()
+        {
+            string searchUrl = $"Patient/{Fixture.Patient.Id}/$everything?_since=3000";
```
Ivanidzo4ka, line 115: >3000 [](http://example.com/codeflow?start=81&length=4)

Who ever gonna debug this test failure in 979 years, would be so mad :D

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/EverythingOperationTests.cs
```diff
@@ -0,0 +1,152 @@
...
+        [Fact]
+        [Trait(Traits.Priority, Priority.One)]
+        [HttpIntegrationFixtureArgumentSets(DataStore.SqlServer, Format.Json)]
+        public async Task GivenAEverythingOperationWithSqlServer_WhenSearched_ThenMethodNotAllowedShouldBeReturned()
+        {
+            string searchUrl = "Patient/bar/$everything";
+
+            using FhirException ex = await Assert.ThrowsAsync<FhirException>(() => Client.SearchAsync(searchUrl));
+
+            Assert.Equal(HttpStatusCode.MethodNotAllowed, ex.StatusCode);
```
Ivanidzo4ka, line 149: >MethodNotAllowed [](http://example.com/codeflow?start=40&length=16)

I think it should be `Not found` since there is no such patient.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/EverythingOperationTests.cs
```diff
@@ -0,0 +1,152 @@
...
+using Microsoft.Health.Fhir.Client;
+using Microsoft.Health.Fhir.Tests.Common.FixtureParameters;
+using Microsoft.Health.Fhir.Tests.E2E.Rest.Search;
+using Microsoft.Health.Test.Utilities;
+using Xunit;
+using Task = System.Threading.Tasks.Task;
+
+namespace Microsoft.Health.Fhir.Tests.E2E.Rest
+{
+    [HttpIntegrationFixtureArgumentSets(DataStore.All, Format.Json)]
```
Ivanidzo4ka, line 17: >All [](http://example.com/codeflow?start=50&length=3)

CosmosDB?

src/Microsoft.Health.Fhir.Core/Messages/Operation/EverythingOperationRequest.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        public string Type { get; }
+
+        public int? Count { get; }
+
+        public string ContinuationToken { get; }
+
+        public IReadOnlyList<string> Includes { get; } = new[] { "general-practitioner", "organization" };
+
+        public IReadOnlyList<Tuple<string, string>> Revincludes { get; } = new[] { Tuple.Create("Device", "patient") };
```
Ivanidzo4ka, line 49: Why these ones, how they been determined?

Ivanidzo4ka: Also consider what this is readonly lists, wouldn't it be simplier to define them in handler and pass to search, than define them here?

This is just message, it feels strange to hide implementation details in it.

src/Microsoft.Health.Fhir.Api/Features/Routing/KnownRoutes.cs
```diff
@@ -52,6 +52,9 @@ internal class KnownRoutes
         public const string ConvertData = "$convert-data";
         public const string ConvertDataOperationDefinition = OperationDefinition + "/" + OperationsConstants.ConvertData;
 
+        public const string Everything = "$everything";
```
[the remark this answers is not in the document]
BoyaWu10: `OperationsConstants` hosts operation names for operations, like "export", "reindex", "convert-data" without "$". Their routes are defined in `KnownRoutes` in Line 33, 46, 52 with "$". My understanding is that we can keep "$everything" here for consistency. If we are going to reference its operation name "everything", we add an operation name in `OperationConstants`.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/EverythingOperationTests.cs
CaitlinV39: Based on customer requirements, I'd like the data to still be returned. Cigna currently doesn't load the patient in the database but wants to get the data related to the patient (they have a placeholder id).

I think the question is more around what if the patient actually doesn't exist. How is that handled?

BoyaWu10: There are 3 kinds of "non-existent" cases: (1) The Patient resource is not created, and its ID is not referenced in other resources. In this case, currently a Bundle with empty entry is returned. (2) The Patient resource is not created, but its ID has been referenced in other resources. (I think Cigna belongs to this case.) In this case, currently the Device resources and compartment resources are searched. (3) The Patient resource has been created, but is now deleted. The behavior is the same as (2).

src/Microsoft.Health.Fhir.Core/Features/Search/SearchService.cs
Ivanidzo4ka: And that's what I don't like.
Purpose of searchOptionFactory is to build searchOptions object which has Expression and other things.

And later we treat Expression based on persistence implementation.

request->common representation as expressions-> Cosmos or Sql logic which tailor expressions to it's own needs.

If we add Or condition (cosmos or sql) on second step, it's gonna be nightmare to support.

Ivanidzo4ka:
I want to talk about design.

There is new additions of `SearchOptionFactory.Create` for everything case.
`SearchService` getting `SearchForEverythingOperationAsync`
`FhirCosmosSearchService` getting `SearchReferencesForEverythingOperation`.

Whole flow looks like -> Everything handler -> searchService.SearchForEverything -> SearchOptionsFactory-> SearchService.Search + SearchReferencesForEverythingOperation -> SearchOptionsFactory ->SearchService.Search

What I don't like is everyone who uses `SearchOptionFactory` would get access for access everything stuff.
Everyone who has access to `SearchService` would get access for everything stuff.
This classes gain functionality which no one other than patient everything consumer need.
It made to work for cosmos only, but it uses all the working in both cosmos and sql server building blocks.


What would I do.
Everything handler remain the same, just instead of calling searchService.SearchForEverything I would have my own PatientEverythingService with one method Search.

`PatientEveythingService` would basically execute `SearchForEverythingOperationInternalAsync` method from `FhirCosmosSearchService`. 
Push logic out of `SearchOptionFactory`  related to patientEverything into that service as well. 

You couldn't do it in first iteration, but now `SearchAsync(searchOptions, cancellationToken)` method you calling is public, so you can access it without any problems.

No need to add additional methods to `SearchService`, `SearchOptionsFactory`, everything is nice and tidy in one file, rather than be split around multiple places. And bonus for that, is there is high chance SQL would work out of box.

src/Microsoft.Health.Fhir.Api/Features/Routing/KnownRoutes.cs
BoyaWu10: Agree. I've moved the restrictions on route level in the update.

src/Microsoft.Health.Fhir.Core/Features/Search/SearchService.cs
BoyaWu10: Got the point. The OR condition in expressions are now removed. Instead, it records phases in continuation token and get the results by multiple searches.

src/Microsoft.Health.Fhir.Core/Messages/Everything/EverythingOperationRequest.cs
```diff
@@ -0,0 +1,45 @@
...
+            ResourceId = resourceId;
+            Start = start;
+            End = end;
+            Since = since;
+            Type = type;
+            Count = count;
+            ContinuationToken = continuationToken;
+        }
+
+        public string ResourceType { get; }
```
Ivanidzo4ka, line 29: >ResourceType [](http://example.com/codeflow?start=22&length=12)

it's a bit confusing, ResourceType and Type.

Maybe call this EverythingOperationType and rename Type to ResourceTypes?

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/EverythingOperationContinuationToken.cs
```diff
@@ -0,0 +1,44 @@
...
+{
+    public class EverythingOperationContinuationToken
+    {
+        public EverythingOperationContinuationToken(int phase, string internalContinuationToken)
+        {
+            Phase = phase;
+            InternalContinuationToken = internalContinuationToken;
+        }
+
+        public int Phase { get; internal set; }
```
Ivanidzo4ka, line 18: >internal [](http://example.com/codeflow?start=32&length=8)

would it work with `private set` instead of `internal set`?

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/IPatientEverythingService.cs
```diff
@@ -0,0 +1,26 @@
...
+using System.Threading.Tasks;
+using Microsoft.Health.Fhir.Core.Features.Search;
+using Microsoft.Health.Fhir.Core.Models;
+
+namespace Microsoft.Health.Fhir.Core.Features.Operations.Everything
+{
+    public interface IPatientEverythingService
+    {
+        Task<SearchResult> SearchAsync(
+            string resourceType,
```
Ivanidzo4ka, line 16: >string resourceType, [](http://example.com/codeflow?start=12&length=20)

I don't think you need it. This is I`Patient`EverythingService. You can use const string inside instead of passing parameter.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/FhirServerBuilderEverythingOperationRegistrationExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+
+using EnsureThat;
+using Microsoft.Health.Fhir.Core.Features.Operations.Everything;
+using Microsoft.Health.Fhir.Core.Registration;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    public static class FhirServerBuilderEverythingOperationRegistrationExtensions
+    {
+        public static IFhirServerBuilder AddEverythingOperation(this IFhirServerBuilder fhirServerBuilder)
```
Ivanidzo4ka, line 14: >AddEverythingOperation [](http://example.com/codeflow?start=41&length=22)

Considering you aiming to have multiple operations, maybe it's worth to call  AddEverythingOperation`s` ?

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/EverythingOperationHandler.cs
```diff
@@ -0,0 +1,53 @@
...
+        public async Task<EverythingOperationResponse> Handle(EverythingOperationRequest request, CancellationToken cancellationToken)
+        {
+            EnsureArg.IsNotNull(request, nameof(request));
+
+            if (await _authorizationService.CheckAccess(DataActions.Read, cancellationToken) != DataActions.Read)
+            {
+                throw new UnauthorizedFhirActionException();
+            }
+
+            SearchResult searchResult = await _patientEverythingService.SearchAsync(
```
Ivanidzo4ka, line 45: >searchResult [](http://example.com/codeflow?start=25&length=12)

Reading code I have feeling what EverythingOperationHandler would eventually support all 4 operations. I would add if check (or switch) to check request.ResourceType and call `_patientEverythingService` if it's `Patient` and throw exception with (`resourceType` not supported for $everything operation or something like that).

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
```diff
@@ -0,0 +1,369 @@
...
+        }
+
+        public async Task<SearchResult> SearchAsync(
+            string resourceType,
+            string resourceId,
+            PartialDateTime start,
+            PartialDateTime end,
+            PartialDateTime since,
+            string type,
+            int? count,
```
Ivanidzo4ka, line 55: >int? count, [](http://example.com/codeflow?start=12&length=11)

I don't think it's been validated and used anywhere.

Regarding validation, it probably should be some reasonalbe value, from 1. to some const in service or config value.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
```diff
@@ -0,0 +1,369 @@
...
+
+            searchParameters.AddRange(_includes.Select(include => Tuple.Create(SearchParameterNames.Include, $"{resourceType}:{include}")));
+
+            // Search
+            SearchOptions searchOptions = _searchOptionsFactory.Create(resourceType, searchParameters);
+            SearchResult searchResult = await search.Value.SearchAsync(searchOptions, cancellationToken);
+            searchResultEntries.AddRange(searchResult.Results.Select(x => new SearchResultEntry(x.Resource)));
+
+            // Filter results by IsDeleted
+            searchResultEntries = searchResultEntries.Where(s => !s.Resource.IsDeleted).ToList();
```
Ivanidzo4ka, line 162: Are we actually return deleted resources?
I don't this should be the case.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
```diff
@@ -0,0 +1,369 @@
...
+            // Filter results by _type
+            if (!string.IsNullOrEmpty(type))
+            {
+                IReadOnlyList<string> types = type.SplitByOrSeparator();
+                searchResultEntries = searchResultEntries.Where(s => types.Contains(s.Resource.ResourceTypeName)).ToList();
+            }
+
+            // Filter results by _since
+            if (since != null)
+            {
```
Ivanidzo4ka, line 173: You had code in previous iteration which constructed expressions, you can just reuse same code, and create manually SearchOptions object rather than call searchOptionFactory.

That way you filter out results on server rather than application level.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
```diff
@@ -0,0 +1,369 @@
...
+            SearchResult searchResult = await search.Value.SearchAsync(searchOptions, cancellationToken);
+            searchResultEntries.AddRange(searchResult.Results.Select(x => new SearchResultEntry(x.Resource)));
+
+            // Filter results by IsDeleted
+            searchResultEntries = searchResultEntries.Where(s => !s.Resource.IsDeleted).ToList();
+
+            // Filter results by _type
+            if (!string.IsNullOrEmpty(type))
+            {
+                IReadOnlyList<string> types = type.SplitByOrSeparator();
```
Ivanidzo4ka, line 167: >type.SplitByOrSeparator() [](http://example.com/codeflow?start=46&length=25)

It's been called 5 times in service. Can we call it once in beggining and pass `IReadOnlyList<string> types` ot methods?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/EverythingOperationTests.cs
```diff
@@ -0,0 +1,138 @@
...
+    public class EverythingOperationTests : SearchTestsBase<EverythingOperationTestFixture>
+    {
+        public EverythingOperationTests(EverythingOperationTestFixture fixture)
```
Ivanidzo4ka, line 19: >EverythingOperationTests [](http://example.com/codeflow?start=15&length=24)

Is any of these tests triggers continuation token and iterate with it over collection?

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Controllers/FhirControllerTests.cs
```diff
@@ -0,0 +1,99 @@
...
+
+        public FhirControllerTests()
+        {
+            _fhirController = GetController();
+            var controllerContext = new ControllerContext { HttpContext = _httpContext };
+            _fhirController.ControllerContext = controllerContext;
+        }
+
+        [Fact]
+        public async Task GivenAnEverythingOperationRequest_WhenValid_ThenProperResponseShouldBeReturned()
```
Ivanidzo4ka, line 48: >GivenAnEverythingOperationRequest_WhenValid_ThenProperResponseShouldBeReturned [](http://example.com/codeflow?start=26&length=78)

Can we have negative test(s)?
negative count, type contains garbage or non existing type, garbage continuation token?

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
```diff
@@ -0,0 +1,369 @@
...
+                        token.Phase = 3;
+                        goto case 3;
+                    }
+
+                    break;
+                case 3:
+                    searchResult = await SearchRevinclude(resourceId, since, type, encodedInternalContinuationToken, cancellationToken);
+                    break;
+                default:
+                    return new SearchResult(Enumerable.Empty<SearchResultEntry>(), null, null, Array.Empty<Tuple<string, string>>());
```
Ivanidzo4ka, line 122: >return new SearchResult(Enumerable.Empty<SearchResultEntry>(), null, null, Array.Empty<Tuple<string, string>>()); [](http://example.com/codeflow?start=20&length=113)

I would throw exception.
This shouldn't happen in theory, so it sounds like exceptional situation for me.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
```diff
@@ -0,0 +1,369 @@
...
+            if (string.Equals(search.Value.GetType().Name, "SqlServerSearchService", StringComparison.Ordinal))
+            {
+                throw new OperationNotImplementedException("$everything operation is not yet implemented in SQL Server.");
+            }
+
+            EverythingOperationContinuationToken token = string.IsNullOrEmpty(continuationToken)
+                ? new EverythingOperationContinuationToken(0, null)
+                : EverythingOperationContinuationToken.FromString(DecodeContinuationTokenFromBase64String(continuationToken));
+
+            if (token == null)
```
Ivanidzo4ka, line 71: >token == null [](http://example.com/codeflow?start=16&length=13)

|| token.Phase<0 || token.Phase>3

src/Microsoft.Health.Fhir.Core/Messages/Everything/EverythingOperationRequest.cs
BoyaWu10: Agree. The names are fixed in the update.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/EverythingOperationContinuationToken.cs
BoyaWu10: `Phase` needs `internal set` because it is set in `PatientEverythingService` as well. `InternalContinuationToken` could be `private set` and I've fixed it in the update. Thanks for pointing this out.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/IPatientEverythingService.cs
BoyaWu10: Agree. It is removed in the update.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/FhirServerBuilderEverythingOperationRegistrationExtensions.cs
BoyaWu10: Sure! It is fixed in the update.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/EverythingOperationHandler.cs
BoyaWu10: Yes, I have the same feeling that `EverythingOperationHandler` will eventually support all the required $everything operation. A check is added here. If `EverythingOperationType` is not Patient, `RequestNotValidException` is thrown.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
BoyaWu10: Yes, `count` is not used anywhere. In the [spec](https://www.hl7.org/fhir/patient-operation-everything.html) of Patient $everything, there is a `_count` input parameter. Currently it is not supported, as we use `_include` to pull the patient's direct references and resources pulled by `_include` are not limited by `count`. We may support it in the future so I just put a variable in the interface.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
BoyaWu10: `SearchAsync` in Cosmos DB does not return deleted resources. But we when pulling `_include`, we use `_fhirDataStore.GetAsync` and this does not limit `isDeleted=false`. So a deleted resource will be returned, but only contains some metadata. Here we just filter it out to keep consistency with other resources pulled by `SearchAsync`.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
BoyaWu10: Agree. It is refined in the update.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/EverythingOperationTests.cs
BoyaWu10: Yes, tests that use `ExecuteAndValidateBundle` trigger continuation token to get and validate the second page as well. I also refined the first test to iterate over all resources in response across three pages.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
BoyaWu10: Agree. It is refined in the update. If this happens, a `EverythingOperationException` would be thrown and return `InternalServerError`.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
BoyaWu10: Good point!

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Controllers/FhirControllerTests.cs
BoyaWu10: Sure! We have negative tests in E2E tests that include all valid/all invalid/some invalid type and garbage continuation token. As count is not supported, we do not verify the input `_count` parameter. I've looked into `FhirController.PatientEverythingById` and it does not do any checks, so I do not add it in unit tests and just keep those tests in E2E tests.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
BoyaWu10: I'm not sure whether I get the point correctly. Here we do some post-process to filter out the results because resources pulled by `_include` are "less controlled". For example, it is not limited by `_since`. So we just remove those resources in application level.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
Ivanidzo4ka: searchOptions has field `IncludeCount` which I believe should control amount of data returned in include.

>Alternatively, clients may choose to page through the result set (or servers may require this). Paging through the results is done the same as for Searching, using the _count parameter, and Bundle links.

I think spec is quite clear about count use. If we have 1000 results in database, I should be able to pass count=100 and iterate over that collection with continuation token. 

Note that count is upper limit to return results, it's ok to return 10 results if count set to be 20, as long as we fetch all data via continuation token, but in same time it's not ok to return 21 result. (Fine to underdeliver but never ok to overdeliver)

BoyaWu10: Yes, searchOptions has a field `IncludeCount` to limit the number of resources searched by `_include`. But in the scenario of Patient $everything I think there are two things that worth attention.

1). There's no continuation token provided for resources searched by `_include`. In Cosmos DB, the `_include` IDs are pulled together with the Patient resource. Then `_include` resources themselves are pulled by `_fhirDataStore.GetAsync()`. They are pulled from database at once without a continuation token, which is very different from other search queries and this would require recording extra information to support `_count`.

Taking following case as an example, if there are 10 `_include` resources, when the user specifies `_count` to 3, we will need to record extra information to pull the next 3 resources in subsequent pages (like record current returned number in continuation token, perform the same `_include` search again and return the next batch of resources).

2). I agree that from the implementation level, we could possibly figure out some way to achieve `_count`, but I also think that this requires more tests to make sure it works smoothly, because one different point between Patient $everything and other searches is that it could possibly issue more queries into the database to construct one page (searching Patient, `_include`, compartment with data, compartment without data, Device).

Given that we need to support Patient $everything for CMS requirement before July 1st and `_count` is considered low priority in our PM spec, we have come to an agreement to skip supporting `_count` in the initial release. We can try to support this in future releases after more detailed design and thorough tests.

Ivanidzo4ka: I can be wrong, but in my head `_count` and `continuationToken` is pair which goes together and only together. If `_count` is not supported we need to return all results in bundle. If `_count` supported we need to iterate over collection with support of `_continuationToken`.

If `_count` is low priority, I would suggest to just exclude it from all interfaces together with `_continuationToken`

BoyaWu10: In the [spec](https://www.hl7.org/fhir/http.html#paging) of paging, there's no requirement about supporting `_count` or making it together with the continuation token, so I think it would be OK to skip `_count` here. Besides, if we do not have a continuation token, we will need to return all resources (including compartment resources) in one response, which can too large to work.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/FhirController.cs
```diff
@@ -512,5 +513,30 @@ public async Task<IActionResult> BatchAndTransactions([FromBody] Resource bundle
...
+        /// <param name="idParameter">The resource ID</param>
+        /// <param name="start">The start date relates to care dates</param>
+        /// <param name="end">The end date relates to care dates</param>
+        /// <param name="since">Resources updated after this period will be included in the response</param>
+        /// <param name="type">Comma-delimited FHIR resource types to include in the return resources</param>
+        /// <param name="ct">The continuation token</param>
+        [HttpGet]
+        [Route(KnownRoutes.PatientEverythingById, Name = RouteNames.PatientEverythingById)]
+        [AuditEventType(AuditEventSubType.Everything)]
+        public async Task<IActionResult> PatientEverythingById(
```
Ivanidzo4ka, line 529: >PatientEverythingById [](http://example.com/codeflow?start=41&length=21)

[https://www.hl7.org/fhir/operations.html#executing](https://www.hl7.org/fhir/operations.html#executing)

[https://www.hl7.org/fhir/operations.html#request](https://www.hl7.org/fhir/operations.html#request)


I was always under impression what default behaviour for any operation is to have POST request with `Parameters` object inside body, and in some specific cases it's ok to have alternative of running GET request 

>If all the parameters for the operation are primitive types and the operation has affectsState = false, the operation may be invoked by performing an HTTP GET operation where all of the values of the parameters are appended to the URL in the search portion of the URL.


So I would add HttpPost action to controller as well.

src/Microsoft.Health.Fhir.Core/Features/KnownQueryParameterNames.cs
```diff
@@ -71,5 +71,12 @@ public static class KnownQueryParameterNames
         public const string OutputFormat = "_outputFormat";
 
         public const string TypeFilter = "_typeFilter";
+
+        /// <summary>
+        /// The date range parameters for $everything operation
+        /// </summary>
+        public const string Start = "start";
```
Ivanidzo4ka, line 78: >Start [](http://example.com/codeflow?start=28&length=5)

I would keep it separatly from this list.
With exception of Anonymization all that parameters are global or search specific.

maybe something like:
[https://github.com/microsoft/fhir-server/blob/main/src/Microsoft.Health.Fhir.Core/Features/Operations/ConvertData/Models/ConvertDataProperties.cs](https://github.com/microsoft/fhir-server/blob/main/src/Microsoft.Health.Fhir.Core/Features/Operations/ConvertData/Models/ConvertDataProperties.cs)

src/Microsoft.Health.Fhir.Core/Resources.resx
```diff
@@ -563,4 +563,11 @@
   <data name="FailedToAnonymizeResource" xml:space="preserve">
     <value>Failed to anonymize resource. The job will be marked as failed. {0}</value>
   </data>
+  <data name="UnsupportedResourceType" xml:space="preserve">
```
Ivanidzo4ka, line 566: >UnsupportedResourceType [](http://example.com/codeflow?start=14&length=23)

there is `ResourceNotSupported` maybe we can reuse it?

src/Microsoft.Health.Fhir.Api/Features/Routing/KnownRoutes.cs
```diff
@@ -55,6 +55,9 @@ internal class KnownRoutes
         public const string MemberMatch = "Patient/$member-match";
         public const string MemberMatchOperationDefinition = OperationDefinition + "/" + OperationsConstants.MemberMatch;
 
+        public const string Everything = "$everything";
+        public const string PatientEverythingById = "Patient/" + IdRouteSegment + "/" + Everything;
```
Ivanidzo4ka, line 59: >PatientEverythingById [](http://example.com/codeflow?start=28&length=21)

Just to double check.
[https://www.hl7.org/fhir/operation-patient-everything.html](https://www.hl7.org/fhir/operation-patient-everything.html)

define two endpoints:

URL: [base]/Patient/$everything

URL: [base]/Patient/[id]/$everything

We don't add first endpoint yet, because it's low priority?

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
```diff
@@ -0,0 +1,366 @@
...
+        }
+
+        private async Task<SearchResult> SearchRevinclude(
+            string resourceId,
+            PartialDateTime since,
+            IReadOnlyList<string> types,
+            string continuationToken,
+            CancellationToken cancellationToken)
+        {
+            if (types.Any() && !types.Contains(_revinclude.Item1))
```
Ivanidzo4ka, line 195: >_revinclude.Item1 [](http://example.com/codeflow?start=47&length=17)

nit: This is so hard to read.
I would break that tuple into two const strings and use them.

`!types.Contains(Device)`
or 
`Tuple.Create(Patient, resourceId)`
is more readable than

`!types.Contains(_revinclude.Item1)`
and
`Tuple.Create(_revinclude.Item2, resourceId)`

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/EverythingOperationContinuationToken.cs
Ivanidzo4ka: I've check code I don't see why can't you just do 
```
var phase =  token.Phase;
switch(phase){
...
}
```
I just personally prefer readonly fields because they save you from mutatable state mistakes, but it's up to you.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
```diff
@@ -0,0 +1,366 @@
...
+namespace Microsoft.Health.Fhir.Core.Features.Operations.Everything
+{
+    public class PatientEverythingService : IPatientEverythingService
+    {
+        private readonly Func<IScoped<ISearchService>> _searchServiceFactory;
+        private readonly ISearchOptionsFactory _searchOptionsFactory;
+        private readonly ISearchParameterDefinitionManager _searchParameterDefinitionManager;
+        private readonly ICompartmentDefinitionManager _compartmentDefinitionManager;
+
+        private IReadOnlyList<string> _includes = new[] { "general-practitioner", "organization" };
```
Ivanidzo4ka, line 30: >"general-practitioner", "organization" [](http://example.com/codeflow?start=58&length=38)

what about `link`?

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
Ivanidzo4ka: Or if you plan to have collection of revInclude can you use
`private readonly (string ResourceType, string SearchCode) _revinclude = new ("Device", "patient");
instead?

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
Ivanidzo4ka: That sounds like a bug, on our side...

BoyaWu10: Agree that it might be better to remove deleted resources in `FhirCosmosSearchService.PerformIncludeQueries()`, though those deleted resources only contain some metadata.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/EverythingOperationTests.cs
BoyaWu10: As discussed in previous comments, the constraint is now set at route level. If the resource type is not supported, `404 Not Found` is returned.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
BoyaWu10: We do not search `link` here, because including `link` might make the result Bundle containing duplicate resources. For example, if Patient A and B link to each other in `link`, when doing [base]\Patient\A\$everything, Patient B would be returned twice, one pulled by _include and the other pulled by Patient A's compartment search. By discussing with Caitlin, returning duplicate resources is a deal breaker, so we do not include `link` in `_include`.

src/Microsoft.Health.Fhir.Api/Features/Routing/KnownRoutes.cs
BoyaWu10: This is because there is a discussion about Patient level export. Patient level export accomplishes a similar functionality as the first endpoint, which pulls all patient data into a storage account. Now we will direct customers to continue to use patient export if they need all patients. In the future, we may want to update the Patient level export to use the same logic as $patient-everything. The details could be found in the PM spec.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/EverythingOperationContinuationToken.cs
BoyaWu10: Agree. I've set `Phase` to private in the update.

src/Microsoft.Health.Fhir.Core/Features/KnownQueryParameterNames.cs
BoyaWu10: Sure, `EverythingOperationParameterNames` is added to hold these parameters.

src/Microsoft.Health.Fhir.Core/Resources.resx
BoyaWu10: Good point! It is refined in the update.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
BoyaWu10: Agree! It is fixed in the update.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/FhirController.cs
BoyaWu10: Thanks for sharing this. I've read through the specs. My understanding is that this operation has some characteristics that make it more suitable for GET.

(1) The input parameters can be empty, which is different from other operations like $convert-data or $member-match. The [spec](https://www.hl7.org/fhir/operations.html#executing) mentioned that:

> Executing operations without any parameters is a special case. For an operation that doesn't cause any state change, the operation is invoked in a straight forward fashion:
GET [base]/Composition/example/$document
For operations that call state changes, they must be invoked by a POST. There is no parameters resource in this case because a parameters resource cannot be empty. So the operation is invoked with a POST with an empty body:
POST [base]/Claim/example/$submit
Content-Length: 0

I think Patient $everything belongs to the first case that doesn't cause any state change. Besides, in current FHIR Server, posting empty body will result in bad request.

(2) This operation supports paging. A continuation token would be returned and `link.next` provides as a GET call for following pages.

(3) There are several things we need to consider before supporting POST, e.g., handling conflicts between parameters specified in URL and body.

So I would suggest enabling only GET first.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
BoyaWu10: Since PR #1956 fixed this, the isDeleted check is removed in the post-process of $everything operation.

Ivanidzo4ka (approved):
:shipit:

BoyaWu10:
Hi @Ivanidzo4ka, one change is made since the last approval. The registration of `PatientEverythingService` is moved to `OperationsModule`, which is a good place to register dependencies for operations. Another benefit is that we don't need to update `FhirService.{R4}.Startup.cs` to register this service in PaaS.

PR #1894, opened 2021-04-28
Reference resolution on ingest for faster chained searches

johnstairs:
## Description

We currently defer the resolution of resource references to search time. This PR changes that to perform and maintain it on ingestion. We add a `ResolvedResourceSurrogateId` to the `ReferenceSearchParam` table and indexes so that the field can be read and updated efficiently. 

The result is a improvement in performance for chained searches. This PR does not address `_include` performance.

The plan was to use an indexed view but performance hit was a lot higher than maintaining this column ourselves.

Patch

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/10.diff.sql
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/10.diff.sql
@@ -391,7 +391,7 @@ AS
 
     /* Update inbound references. When we support referential integrity, this only needs to be done on update */
     UPDATE dbo.ReferenceSearchParam WITH (HOLDLOCK) -- make sure nobody can insert new entries in this range
-        SET ReferenceResourceSurrogateId = @resourceSurrogateId
+        SET ReferenceResourceSurrogateId = CASE @isDeleted WHEN 1 THEN NULL ELSE @resourceSurrogateId END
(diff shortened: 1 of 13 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/10.diff.sql
```diff
@@ -0,0 +1,585 @@
+SET NOCOUNT ON
```
Ivanidzo4ka, line 1: >SET [](http://example.com/codeflow?start=0&length=3)

I can probably do comparison between 9 and 10 version, and then between 10 and 11, and try to deduct why 2 versions needed, but I better ask directly.

Most of the time it's one time migration per PR, why it's double migration this time?

johnstairs: First one ensures that new data will have the new column set (the stored procedure writes is). The second one updates all the existing data. When the second is done, you can start reading from the column.

johnstairs: Actually, I guess you could combine them.

CaitlinV39:
We are going to close this for now based on priorities but will leave the code for reference for when we have time to revisit.

PR #1897, opened 2021-04-30
Large db query perf optimizations

brendankowitz:
## Description
Before if a query failed to find enough results it would switch to querying all results in parallel, which is increases the RU cost greatly with very large DBs. This scales that back to keep doubling the partitions queried in parallel until it finds the appropriate watermark.

Other changes:
- It looks like there is a bug where if the number of results found (in total) is less than the the MaxItemCount then it was switching to querying all results in parallel (I inverted the string.IsNullOrEmpty(continuationToken) logic
- Added a feature flag to turn off using query statistics
- Added the ability to use the average required number of partitions needed to get results as the degree of parallelism

## Related issues
Addresses [issue #].

## Testing

## FHIR Team Checklist

Patch

### DIFF
--- a/src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
+++ b/src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
@@ -419,7 +419,7 @@ namespace Microsoft.Health.Fhir.CosmosDb.Features.Search
                     if (results.Count < feedOptions.MaxItemCount * CosmosFhirDataStore.ExecuteDocumentQueryAsyncMinimumFillFactor && string.IsNullOrEmpty(continuationToken))
                     {
                         // ExecuteDocumentQueryAsync gave up on filling the pages. This suggests that we would have been better off querying in parallel.
-                        queryPartitionStatistics.Update(_physicalPartitionInfo.PhysicalPartitionCount);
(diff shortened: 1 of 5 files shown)


src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
```diff
@@ -416,15 +424,31 @@ private async Task<Expression> RecurseChainedExpression(ChainedExpression expres
...
-                        queryPartitionStatistics.Update(_physicalPartitionInfo.PhysicalPartitionCount);
+                        int partitions = queryPartitionStatistics.GetAveragePartitionCount() ??
+                            messagesList.Select(r => r.Headers[_documentDbPartitionkeyRangeIdHeaderName]).Distinct().Count();
+
+                        // At least 1
+                        int minNumberOfPartitions = Math.Max(partitions, 1);
+
+                        // Choose the number of partitions used * 2, or the max db partitions, whichever is less.
+                        int increasePartitionsOrMaxPartitions = Math.Min(minNumberOfPartitions * 2, _physicalPartitionInfo.PhysicalPartitionCount);
+                        queryPartitionStatistics.Update(increasePartitionsOrMaxPartitions);
```
brendankowitz, line 439: I'm not sure if there is something better here. The right number is somewhere between 1 and PhysicalPartitionCount :)

PR #1907, opened 2021-05-04
Custom search fixes for multi-instance scenario

namadabu:
## Description
Updating our deployment pipeline to deploy 2 instances of the fhir-server by default. This will allow us to test scenarios where multiple instances of the fhir-server are running.

Update:
Updated code to handle different scenarios that might happen when we have multiple instances of the fhir-server.
1. Updating a newly created search parameter. Lets say the `POST SearchParameter` request goes to instance 1. If the subsequent `PUT SearchParameter` request goes to instance 2 we will fail the request since instance 2 does not know about the SearchParameter yet. 
2. Triggering a reindex job. Similar to the above except that we send a `POST $reindex` which lands on instance 2. Since instance 2 might not have the latest search parameters (there is a 10 second time window for the refresh cycle), it creates a reindex job with the "old" hash values. So when reindex runs it doesn't find the resources that need to be re-indexed since it is searching using the old hash. We end up marking the search parameter as enabled even though the resources have not been re-indexed. 

To address both scenarios, I have added code to get search parameter updates from the data store while processing the respective requests.

## Related issues
Addresses [AB#81878](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/81878)

## Testing
Locally deployed multiple-instances and tested the above scenarios.
Manually deployed the branch and verified that the app service plan had instance count set to 2:

image

Also verified using application insights logs that requests were being sent to both instances:

image

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/CreateReindexRequestHandler.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/CreateReindexRequestHandler.cs
@@ -62,6 +62,11 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Reindex
                 throw new JobConflictException(string.Format(Resources.OnlyOneResourceJobAllowed, reindexJobId));
             }
 
+            // We need to pull in latest search parameter updates from the data store before creating a reindex job.
+            // There could be a potential delay of <see cref="ReindexJobConfiguration.JobPollingFrequency"/> before
+            // search parameter updates on one instance propagates to other instances. If we store the reindex
+            // job with the old hash value in _searchParameterDefinitionManager.SearchParameterHashMap, then we will
+            // not detect the resources that need to be reindexed.
             await _searchParameterOperations.GetAndApplySearchParameterUpdates(cancellationToken);
 
             var jobRecord = new ReindexJobRecord(
--- a/src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/CreateReindexRequestHandler.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/CreateReindexRequestHandler.cs
@@ -13,6 +13,7 @@ using Microsoft.Health.Fhir.Core.Configs;
 using Microsoft.Health.Fhir.Core.Exceptions;
 using Microsoft.Health.Fhir.Core.Features.Definition;
 using Microsoft.Health.Fhir.Core.Features.Operations.Reindex.Models;
+using Microsoft.Health.Fhir.Core.Features.Search.Parameters;
 using Microsoft.Health.Fhir.Core.Features.Security;
 using Microsoft.Health.Fhir.Core.Messages.Reindex;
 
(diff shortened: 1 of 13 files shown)


Ivanidzo4ka (approved):
:shipit:

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/CreateReindexRequestHandler.cs
```diff
@@ -57,6 +62,8 @@ public async Task<CreateReindexResponse> Handle(CreateReindexRequest request, Ca
                 throw new JobConflictException(string.Format(Resources.OnlyOneResourceJobAllowed, reindexJobId));
             }
 
+            await _searchParameterOperations.GetAndApplySearchParameterUpdates(cancellationToken);
```
namadabu, line 65: Add comments here on why this call is required

Ivanidzo4ka:
It looks like r4 cosmos tests had issue and didn't properly start asp.net core app, which is scary, any chance you can look on actual app in azure and see is there anything in logs?
It's basically lot of Not Found and Method not allowed exception from app

src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterStatusManager.cs
```diff
@@ -113,7 +113,7 @@ public async Task UpdateSearchParameterStatusAsync(IReadOnlyCollection<string> s
             {
                 var searchParamUri = new Uri(uri);
 
-                var paramInfo = _searchParameterDefinitionManager.GetSearchParameter(searchParamUri);
+                SearchParameterInfo paramInfo = _searchParameterDefinitionManager.GetSearchParameter(searchParamUri);
```
feordin, line 116: nit: was this change intended?

namadabu: Yes. I had initially edited that file with some changes that I later reverted. I left this one in since it seemed more readable.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/CreateReindexRequestHandler.cs
```diff
@@ -57,6 +62,13 @@ public async Task<CreateReindexResponse> Handle(CreateReindexRequest request, Ca
                 throw new JobConflictException(string.Format(Resources.OnlyOneResourceJobAllowed, reindexJobId));
             }
 
+            // We need to pull in latest search parameter updates from the data store before creating a reindex job.
+            // There could be a potential delay of <see cref="ReindexJobConfiguration.JobPollingFrequency"/> before
+            // search parameter updates on one instance propagates to other instances. If we store the reindex
+            // job with the old hash value in _searchParameterDefinitionManager.SearchParameterHashMap, then we will
+            // not detect the resources that need to be reindexed.
+            await _searchParameterOperations.GetAndApplySearchParameterUpdates(cancellationToken);
```
Ivanidzo4ka, line 70: >GetAndApplySearchParameterUpdates [](http://example.com/codeflow?start=45&length=33)

@feordin 
I've give short glance and it looks like there is no locking inside that function.

Which is a bit worriesome. Considering there is 3 places now which would call this method, one of them is on a loop, could it be an issue?

namadabu: That's a good point!

feordin: It is also true that SearchParameterStatusManager updates fields on the parameters, which would also change state.  Perhaps we can have a quick discussion on the best solution.

feordin (approved):

Issue #1898, opened 2021-04-30
INCOMPLETE DATA RETURNED DURING SEARCH


ellgre:
**Describe the bug**
When we are doing a search to fetch all organizations and use the next link to go through all pages we do not get all organizations.

**FHIR Version?**
R4

**Data provider?**
CosmosDB

**To Reproduce**
1. Make a search to get all organizations
2. Note total
3. Use next link and note number of organizations returned
4. Repeat step 3 until there are no more next links
5. Sum opp the actual organizations returned.
6. The actual number returned is not equal to total.

**Expected behavior**
We expect to get the same number of organizations as total

**Actual behavior**
The total parameter = 2175. With _count = 100 we get 2115 when we go through all pages and with _count=1000 we get 2058 organizations.

Until recently we used a version of FHIR server from autumn 2020, and it worked in that version.

CaitlinV39:
@ellgre - Can you send me the exact search syntax you are using for each one? I'm not able to reproduce so just wanted to see if I am missing something.

ellgre:
We are using the following: <baseurl>/Organization/?_format=application%2Fjson&_count=1000.

ebjornset:
Just want to mention that we are running the 1.3.8 release, with the appsetting FhirServer:CoreFeatures:IncludeTotalInBundle = Accurate

Ivanidzo4ka:
I'm trying to reproduce it (in 1.3.8 release branch), and I have trouble with reproducing it as well.

I've check code, and I don't see (so far) anything what could affect search from generating cosmos query.

Query we send right now is:
```
SELECT c.id,c.isSystem,c.partitionKey,c.lastModified,c.rawResource,c.request,c.isDeleted,c.resourceId,c.resourceTypeName,c.isHistory,c.version,c._self,c._etag, c.searchParameterHash
FROM c
WHERE c.isSystem = false
AND c.resourceTypeName = 'Organization'
AND c.isHistory = false
AND c.isDeleted = false
```
And we just specify how much is limit for max items in query settings.

If you know what release you had before (autumn 2020 is a bit vague) I can check what query used to be, but I doubt it was any different.

Main difference I can see between your output and mine is continuation token which cosmos return us back.

if I decode your continuation token it looks like this:
```
[{"token":null,"range":{"min":"05C1DFFFFFFFF8","max":"05C1E7FFFFFFFA"}}]
```
and 
```
[...]

PR #1911, opened 2021-05-05
Fix cosmos iterating over big resources.

Ivanidzo4ka:
## Description
Cosmos db has limit in 4mb of data be returned per request. If we didn't get enough data from cosmos we need to issue more requests to fetch more resources (up to half of requested). There is logic which discards results, and it's flawed, so I fixed it.

## Related issues
https://github.com/microsoft/fhir-server/issues/1898

## Testing
Added E2E tests

## FHIR Team Checklist

Patch

### DIFF
--- a/src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
+++ b/src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
@@ -441,7 +441,7 @@ namespace Microsoft.Health.Fhir.CosmosDb.Features.Storage
             var maxCount = executingWithMaxParallelism
                 ? totalDesiredCount * (mustNotExceedMaxItemCount ? 1 : ExecuteDocumentQueryAsyncMaximumFillFactor) // in this mode, the SDK likely has already fetched pages, so we might as well consume them
                 : totalDesiredCount * ExecuteDocumentQueryAsyncMinimumFillFactor;
-
+            _logger.LogInformation("Doing extra fetch thingy total {total} {maxCount}", totalDesiredCount, maxCount);
             while (cosmosQuery.HasMoreResults &&
                    (results.Count < maxCount)) // we still want to get more results
             {
@@ -458,11 +458,13 @@ namespace Microsoft.Health.Fhir.CosmosDb.Features.Storage
                     feedOptions.MaxItemCount = currentDesiredCount;
                     context = new CosmosQueryContext(sqlQuerySpec, feedOptions, page.ContinuationToken);
                     cosmosQuery = _cosmosQueryFactory.Create<T>(_containerScope.Value, context);
+                    _logger.LogInformation("Running new query looking for {maxItems}", feedOptions.MaxItemCount);
                 }
 
                 try
                 {
                     page = await cosmosQuery.ExecuteNextAsync(linkedTokenSource.Token);
+                    _logger.LogInformation("Get back {count} results", page.Count);
                     if (page.Count > 0)
                     {
                         results.AddRange(page);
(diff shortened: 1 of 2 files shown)


feordin (approved):

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
```diff
@@ -461,11 +461,11 @@ internal async Task<(IReadOnlyList<T> results, string continuationToken)> Execut
                     if (page.Count > 0)
                     {
                         results.AddRange(page);
-                        if (mustNotExceedMaxItemCount && results.Count > feedOptions.MaxItemCount)
+                        if (mustNotExceedMaxItemCount && page.Count > feedOptions.MaxItemCount)
```
Ivanidzo4ka, line 464: >feedOptions.MaxItemCount [](http://example.com/codeflow?start=70&length=24)

This thing get reassign in line 453.

In my test I have 20 results, and cosmos can return me only 9.
So I fetch first 9 and then go to all this fetch more loop.

maxCount is 10 (half of what we ask - 20)

currentDesiredCount is 11, and we set that value to be feedOptions.MaxItemCount.

We do another fetch from cosmos and it returns 9 more results.
So total 18 results.

In old code we would take this 18 results compare them to 
feedOptions.MaxItemCount (which is 11) so that means we have 7 extra results.

So we would flush them off.

In reality, 18 results is still smaller than 20, so we need to return them all.

It looks like bug where we don't consider change in feedOptions.MaxItemCount to happen, so I make change to incorporate that.

brendankowitz, line 464: I had to read this a few times. Can we add a comment, i.e. the page returned more items than we expected...

Ivanidzo4ka: I honestly don't understand this code, and why we even want to throw things away and in what condition it should happen.

I added comment to what is going on right now with test.

namadabu: I am also not familiar with this code, but I wonder whether this PR #1439 can help put things in context

brendankowitz: So take these two examples:
1. Query the DB sequentially, we receive the max or less items then we specify as we scan through the partitions.
2. We issue the query in parallel, we get back results that may be equal to or less than the number items specified, but because it has been issued in parallel, there could be many pages more then we need

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
```diff
@@ -461,14 +461,6 @@ internal async Task<(IReadOnlyList<T> results, string continuationToken)> Execut
                     if (page.Count > 0)
                     {
                         results.AddRange(page);
-                        if (mustNotExceedMaxItemCount && results.Count > feedOptions.MaxItemCount)
```
brendankowitz, line 464: if(executingWithMaxParallelism) probably should have been here. I still see an impact when QueryStatistics is on, but having this disabled in other cases looks right

brendankowitz: Also if `feedOptions.MaxItemCount` was being changed, it should have used the captured `totalDesiredCount`

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
Ivanidzo4ka: Sorry, I don't understand your comment.

Issue #158, opened 2018-10-30
For URL search parameter, the :above modifier is not implemented correctly.


jackliums:
The :above modifier is not implemented correctly. 

The STU3 spec was a little bit vague about what :above should do but the CI build of the R4 spec (http://build.fhir.org/search.html#uri) has a bit more context and example for how :above should behave. Based on that description, the current implementation is incorrect.

CaitlinV39:
AB#76065

PR #1919, opened 2021-05-06
Fix how :above modifier works for URLs

Ivanidzo4ka:
## Description
As it states in spec:
```
GET [base]/ValueSet?url:above=http://acme.org/fhir/ValueSet/123/_history/5

 search for any value set above a given specific URL. 

This will match on any value set with the specified URL, but also on http://acme.org/ValueSet/123. 
Note that there are not many use cases where :above is useful as compared to the :below search
```
Simply put if you looking for :above=ABC
and you have urls: A, AB, ABC they should be fetched, but everything else should be ignored.

## Related issues
Addresses [#158].

## Testing
Updated unit tests.

## FHIR Team Checklist

Patch

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/Parsers/SearchValueExpressionBuilderHelper.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/Parsers/SearchValueExpressionBuilderHelper.cs
@@ -302,7 +302,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Expressions
                     break;
                 case SearchModifierCode.Above:
                     _outputExpression = Expression.And(
-                        Expression.EndsWith(FieldName.Uri, _componentIndex, uri.Uri, false),
+                        Expression.LeftSideStartsWtih(FieldName.Uri, _componentIndex, uri.Uri, false),
                         Expression.NotStartsWith(FieldName.Uri, _componentIndex, KnownUriSchemes.Urn, false));
                     break;
                 case SearchModifierCode.Below:
--- a/src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/Parsers/SearchValueExpressionBuilderHelper.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/Parsers/SearchValueExpressionBuilderHelper.cs
@@ -302,7 +302,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Expressions
                     break;
                 case SearchModifierCode.Above:
                     _outputExpression = Expression.And(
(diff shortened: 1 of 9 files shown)


src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/Parsers/SearchValueExpressionBuilderHelper.cs
```diff
@@ -302,7 +302,7 @@ void ISearchValueVisitor.Visit(UriSearchValue uri)
                     break;
                 case SearchModifierCode.Above:
                     _outputExpression = Expression.And(
-                        Expression.EndsWith(FieldName.Uri, _componentIndex, uri.Uri, false),
+                        Expression.LeftSideStartsWtih(FieldName.Uri, _componentIndex, uri.Uri, false),
```
LTA-Thinking, line 305: Nit: typo

LTA-Thinking (approved):

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SearchParameterQueryGenerator.cs
```diff
@@ -176,7 +176,10 @@ protected static SearchParameterQueryGeneratorContext VisitSimpleBinary(BinaryOp
 
         protected static SearchParameterQueryGeneratorContext VisitSimpleString(StringExpression expression, SearchParameterQueryGeneratorContext context, StringColumn column, string value)
         {
-            AppendColumnName(context, column, expression);
+            if (expression.StringOperator != StringOperator.LeftSideStartsWith)
+            {
+                AppendColumnName(context, column, expression);
+            }
```
rotodd, line 182: To verify my understanding, in this method we are looking to append the following SQL code to the query:

```
AND @p1 LIKE Uri+'%'
AND Uri NOT  LIKE @p2
```
and, in the case of the :above modifier, we need to specify `@p1 LIKE ` before the column name `Uri` (which is different than the other cases). Is this correct?

rotodd (approved):

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SearchParameterQueryGenerator.cs
Ivanidzo4ka: yes that's the query we expect. I don't know much about second AND where we check it doesn't start with `urn:`, there probably a reason for that.

if `:below` is a check what string 'a' is a prefix of 'b', `:above` is switching `a` and `b` places. 

Another way how I could formulate that expression is by using `CHARINDEX(@p1, Uri)=1`, which would check what @p1 value is prefix of value in Uri column. But as I found you can use LIKE expression as well.

Issue #1628, opened 2021-02-01
Conditional Delete


CaitlinV39:
**User story**
As a user of the FHIR Server, I would like to be able to leverage Conditional Delete, where I can delete resources based on some passed condition and the item is deleted if that item is found. 

https://www.hl7.org/fhir/http.html#3.1.0.7.1 

The spec includes an option of:
> A server may choose to delete all the matching resources, or it may choose to return a 412 Precondition Failed error

In this case we should return a `412` and not delete multiple matches.

**Acceptance criteria**
I can specify criteria and delete based on that critera.

PR #1923, opened 2021-05-11
Adds conditional delete

brendankowitz:
## Description
Adds conditional delete, defaulting to Single row behavior and a flag to specify multiple.

Examples (also see included http file):

```
# Conditional delete:
DELETE https://{{hostname}}/Patient?identifier=1032704

# Conditional hardDelete:
DELETE https://{{hostname}}/Patient?identifier=1032704&hardDelete=true

# Conditional delete multiple:
DELETE https://{{hostname}}/Patient?identifier=1032704&_count=100

# Conditional hardDelete multiple:
DELETE https://{{hostname}}/Patient?identifier=1032704&_count=100&hardDelete=true
```

## Related issues
Addresses [#1628].

## Testing
Adds E2E tests

## FHIR Team Checklist

Patch


src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Delete/ConditionalDeleteResourceHandler.cs
```diff
@@ -0,0 +1,112 @@
...
+using Microsoft.Health.Fhir.Core.Extensions;
+using Microsoft.Health.Fhir.Core.Features.Conformance;
+using Microsoft.Health.Fhir.Core.Features.Persistence;
+using Microsoft.Health.Fhir.Core.Features.Search;
+using Microsoft.Health.Fhir.Core.Features.Security;
+using Microsoft.Health.Fhir.Core.Messages.Delete;
+
+namespace Microsoft.Health.Fhir.Core.Features.Resources.Delete
+{
+    public class ConditionalDeleteResourceHandler : BaseResourceHandler, IRequestHandler<ConditionalDeleteResourceRequest, DeleteResourceResponse>
```
brendankowitz, line 26: Add tests for this

src/Microsoft.Health.Fhir.Shared.Api/Controllers/FhirController.cs
```diff
@@ -368,6 +369,26 @@ public async Task<IActionResult> Delete(string typeParameter, string idParameter
...
+        [HttpDelete]
+        [Route(KnownRoutes.ResourceType)]
+        [AuditEventType(AuditEventSubType.Delete)]
+        public async Task<IActionResult> ConditionalDelete(string typeParameter, [FromQuery] bool hardDelete, [FromQuery] bool multiple)
+        {
+            IReadOnlyList<Tuple<string, string>> conditionalParameters = GetQueriesForSearch();
+
+            DeleteResourceResponse response = await _mediator.Send(new ConditionalDeleteResourceRequest(typeParameter, conditionalParameters, hardDelete, multiple), HttpContext.RequestAborted);
+
+            Response.Headers.Add(KnownHeaders.ItemsDeleted, (response?.ResourcesDeleted ?? 0).ToString(CultureInfo.InvariantCulture));
```
brendankowitz, line 387: Is this useful? Or should it not expose this value?

src/Microsoft.Health.Fhir.Core/Extensions/SearchServiceExtensions.cs
```diff
@@ -21,15 +21,20 @@ public static class SearchServiceExtensions
         /// "If-Exists" headers on the API. Additional logic is used to filter parameters that don't restrict
         /// results, and also ensure that the query meets criteria requirements
         /// </summary>
-        public static async Task<IReadOnlyCollection<SearchResultEntry>> ConditionalSearchAsync(this ISearchService searchService, string instanceType, IReadOnlyList<Tuple<string, string>> conditionalParameters, CancellationToken cancellationToken)
+        public static async Task<(IReadOnlyCollection<SearchResultEntry> results, string continuationToken)> ConditionalSearchAsync(
```
Ivanidzo4ka, line 24: >ConditionalSearchAsync [](http://example.com/codeflow?start=109&length=22)

It's getting overly complicated...

I would break it into two method:
`ConditionalSearchForSingleEntity`
and `ConditionalSearchForCollection`.

Only change this method is needed is to actually force _count=2 (no need to search for more than two entities)

For second method maybe it make sense to pass token and count as separate typed parameters... instead of List of tuples...
Or maybe don't have second method at all, and do all that code in handler itself.

I just don't like how for one specific case we have to alter logic in 3 different places unrelated to it.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/FhirController.cs
Ivanidzo4ka: Wrap it in `if (multiple)` ?

src/Microsoft.Health.Fhir.Core/Features/Conformance/CapabilityStatementBuilder.cs
```diff
@@ -248,6 +248,10 @@ public ICapabilityStatementBuilder PopulateDefaultResourceInteractions()
                     if (!string.Equals(resource, KnownResourceTypes.AuditEvent, StringComparison.Ordinal))
                     {
                         component.ConditionalUpdate = true;
+
+                        component.ConditionalDelete.Add(ConditionalDeleteStatus.NotSupported);
+                        component.ConditionalDelete.Add(ConditionalDeleteStatus.Single);
+                        component.ConditionalDelete.Add(ConditionalDeleteStatus.Multiple);
```
Ivanidzo4ka, line 254: [https://www.hl7.org/fhir/capabilitystatement-definitions.html#CapabilityStatement.rest.resource.conditionalDelete](https://www.hl7.org/fhir/capabilitystatement-definitions.html#CapabilityStatement.rest.resource.conditionalDelete)

Cardinality 0..1

We should put only one value here.

Also I would update comment in line 247 to reflect what delete is not supported for AuditEvent as well.

src/Microsoft.Health.Fhir.Core/Messages/Delete/DeleteResourceResponse.cs
```diff
@@ -10,16 +10,24 @@ namespace Microsoft.Health.Fhir.Core.Messages.Delete
 {
     public class DeleteResourceResponse
     {
-        public DeleteResourceResponse(ResourceKey resourceKey, WeakETag weakETag = null)
+        public DeleteResourceResponse(ResourceKey resourceKey, int resourcesDeleted, WeakETag weakETag = null)
```
Ivanidzo4ka, line 13: >resourcesDeleted [](http://example.com/codeflow?start=67&length=16)

is it worth to make it optional and set default ot 1?

src/Microsoft.Health.Fhir.Shared.Api/Controllers/FhirController.cs
```diff
@@ -368,6 +369,26 @@ public async Task<IActionResult> Delete(string typeParameter, string idParameter
...
+
+            return FhirResult.Accepted().SetETagHeader(response?.WeakETag);
```
Ivanidzo4ka, line 389: >Accepted [](http://example.com/codeflow?start=30&length=8)

Why `Accepted` and not `NoContent` as in normal delete?

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Delete/ConditionalDeleteResourceHandler.cs
```diff
@@ -0,0 +1,112 @@
...
+                        foreach (IEnumerable<SearchResultEntry> batch in matchedResults.TakeBatch(10))
+                        {
+                            DeleteResourceResponse[] results = await Task.WhenAll(batch.Select(result => _mediator.Send(new DeleteResourceRequest(request.ResourceType, result.Resource.ResourceId, request.HardDelete), cancellationToken)));
+
+                            itemsDeleted += results.Sum(x => x.ResourcesDeleted);
+
+                            if (itemsDeleted >= MaxDeletedItems)
+                            {
+                                break;
+                            }
```
Ivanidzo4ka, line 90: So if I ask delete 45 items it would delete 50 for me, because we operate on `TakeBatch(10)` size?

src/Microsoft.Health.Fhir.Core/Messages/Delete/ConditionalDeleteResourceRequest.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        public IReadOnlyList<Tuple<string, string>> ConditionalParameters { get; }
+
+        public bool HardDelete { get; }
+
+        public bool DeleteMultiple { get; }
+
+        public IEnumerable<CapabilityQuery> RequiredCapabilities()
+        {
+            yield return new CapabilityQuery($"CapabilityStatement.rest.resource.where(type = '{ResourceType}').interaction.where(code = 'delete').exists()");
```
brendankowitz, line 39: this needs updating for conditional delete check

src/Microsoft.Health.Fhir.Shared.Api/Controllers/FhirController.cs
brendankowitz: Makes sense

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Delete/ConditionalDeleteResourceHandler.cs
```diff
@@ -0,0 +1,112 @@
...
+            EnsureArg.IsNotNull(mediator, nameof(mediator));
+            EnsureArg.IsNotNull(searchService, nameof(searchService));
+
+            _searchService = searchService;
+            _mediator = mediator;
+        }
+
+        public async Task<DeleteResourceResponse> Handle(ConditionalDeleteResourceRequest request, CancellationToken cancellationToken)
+        {
+            if (await AuthorizationService.CheckAccess(DataActions.Read | DataActions.Write, cancellationToken) != (DataActions.Read | DataActions.Write))
```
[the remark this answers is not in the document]
Ivanidzo4ka: DeleteResourceHandler authorize differently and I can see how adding Read role can be right thing to do for conditional case, I think they should be aligned and conditional delete should be just extension of normal delete.

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Delete/ConditionalDeleteResourceHandler.cs
```diff
@@ -0,0 +1,112 @@
...
+using Microsoft.Health.Fhir.Core.Features.Persistence;
+using Microsoft.Health.Fhir.Core.Features.Search;
+using Microsoft.Health.Fhir.Core.Features.Security;
+using Microsoft.Health.Fhir.Core.Messages.Delete;
+
+namespace Microsoft.Health.Fhir.Core.Features.Resources.Delete
+{
+    public class ConditionalDeleteResourceHandler : BaseResourceHandler, IRequestHandler<ConditionalDeleteResourceRequest, DeleteResourceResponse>
+    {
+        private const int MaxDeletedItems = 500;
```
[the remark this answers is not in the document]
brendankowitz: There is a balance here, but yep, I'll move to config

src/Microsoft.Health.Fhir.Core/Messages/Delete/ConditionalDeleteResourceRequest.cs
```diff
@@ -0,0 +1,43 @@
...
+        public IReadOnlyList<Tuple<string, string>> ConditionalParameters { get; }
+
+        public bool HardDelete { get; }
+
+        public bool DeleteMultiple { get; }
+
+        public IEnumerable<CapabilityQuery> RequiredCapabilities()
+        {
+            yield return new CapabilityQuery($"CapabilityStatement.rest.resource.where(type = '{ResourceType}').conditionalDelete.exists()");
+            ////yield return new CapabilityQuery($"CapabilityStatement.rest.resource.where(type = '{ResourceType}').conditionalDelete != 'not-supported'");
```
Ivanidzo4ka, line 40: >//// [](http://example.com/codeflow?start=12&length=4)

obligatory comment about comment

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Delete/ConditionalDeleteResourceHandler.cs
```diff
@@ -0,0 +1,114 @@
...
+            {
+                return null;
+            }
+            else if (count == 1)
+            {
+                return await _mediator.Send(new DeleteResourceRequest(request.ResourceType, matchedResults.First().Resource.ResourceId, request.HardDelete), cancellationToken);
+            }
+            else
+            {
+                if (!request.DeleteMultiple || _conditionalDeleteMaxItems <= 0)
```
Ivanidzo4ka, line 76: Any chance I can convince you to rewrite this if logic?

I think it would be a bit cleaner if first condition we checked is DeleteMultiple on request, and have two different branches with two different calls to ConditionalSearchAsync methods.

first branch would be copy of delete,
second branch would have while loop.

It just strange to fetch 2 results and then fetch (10) default amount of count results in while loop.

brendankowitz: Will give it a shot. The loop actually does `_conditionalDeleteMaxItems - itemsDeleted` items on the second pass, so its aiming to fetch the remaining 98 items (if the max is 100) then limit to 10 delete requests in parallel.

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Delete/ConditionalDeleteResourceHandler.cs
```diff
@@ -0,0 +1,124 @@
...
+                        cancellationToken,
+                        ct);
+                }
+                else
+                {
+                    matchedResults = Array.Empty<SearchResultEntry>();
+                }
+            }
+
+            return new DeleteResourceResponse(itemsDeleted);
```
Ivanidzo4ka, line 121: >itemsDeleted [](http://example.com/codeflow?start=46&length=12)

If itemsDeleted ==0  I think it should return null instead of response.

Also FhirController should properly handle null as response and return Ok as we do in other conditional cases... Not sure about that one, tho

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ConditionalDeleteTests.cs
```diff
@@ -0,0 +1,128 @@
...
+        private readonly TestFhirClient _client;
+
+        public ConditionalDeleteTests(HttpIntegrationTestFixture<Startup> fixture)
+        {
+            _client = fixture.TestFhirClient;
+        }
+
+        [Fact]
+        [Trait(Traits.Priority, Priority.One)]
+        public async Task GivenNoExistingResources_WhenDeletingConditionally_TheServerShouldReturnAccepted()
```
Ivanidzo4ka, line 37: >GivenNoExistingResources_WhenDeletingConditionally_TheServerShouldReturnAccepted [](http://example.com/codeflow?start=26&length=80)

I would make it theory with multipleDelete = true,false

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ConditionalDeleteTests.cs
```diff
@@ -0,0 +1,128 @@
...
+        [InlineData(false)]
+        [Theory]
+        [Trait(Traits.Priority, Priority.One)]
+        public async Task GivenOneMatchingResource_WhenDeletingConditionally_TheServerShouldDeleteSuccessfully(bool hardDelete)
+        {
+            var identifier = Guid.NewGuid().ToString();
+            await CreateWithIdentifier(identifier);
+            await ValidateResults(identifier, 1);
+
+            FhirResponse response = await _client.DeleteAsync($"{KnownResourceTypes.Observation}?identifier={identifier}&hardDelete={hardDelete}", CancellationToken.None);
```
Ivanidzo4ka, line 56: >hardDelete [](http://example.com/codeflow?start=121&length=10)

is it worth add multipleDelete here as well? For single case it should behave the same, why not check that?

src/Microsoft.Health.Fhir.Core/Features/KnownHeaders.cs
```diff
@@ -13,6 +13,7 @@ internal static class KnownHeaders
         public const string RetryAfter = "Retry-After";
         public const string ProvenanceHeader = "X-Provenance";
         public const string Prefer = "Prefer";
+        public const string ItemsDeleted = "Items-Deleted";
```
Ivanidzo4ka, line 16: >Items-Deleted [](http://example.com/codeflow?start=44&length=13)

since it specific for our server maybe `x-ms-items-deleted`?

brendankowitz: I could do it for consistency, but I was conflicted by this: https://datatracker.ietf.org/doc/html/rfc6648

Ivanidzo4ka: TIL.

I think it's better to leave it as is.

Maybe it's worth to add comments and break this headers into grops like:

```
  //  FHIR Defined headers
  public const string IfNoneExist = "If-None-Exist";
  public const string ProvenanceHeader = "X-Provenance";
 // Generic headers
        public const string RetryAfter = "Retry-After";
        public const string Prefer = "Prefer";
        public const string RequestId = "X-Request-Id";
        public const string CorrelationId = "X-Correlation-Id";
// Microsoft defined headers
     public const string RetryAfterMilliseconds = "x-ms-retry-after-ms";
// Headers our service uses
// Please try to not use x-ms convention because it's obsolete: [https://datatracker.ietf.org/doc/html/rfc6648](https://datatracker.ietf.org/doc/html/rfc6648)
        public const string PartiallyIndexedParamsHeaderName = "x-ms-use-partial-indices";
          public const string ItemsDeleted = "Items-Deleted";
        public const string EnableChainSearch = "x-ms-enable-chained-search";
        public const string ProfileValidation = "x-ms-profile-validation";

```

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Delete/ConditionalDeleteResourceHandler.cs
brendankowitz: Changed the other return to have 0, I think I like having the outcome explicit

src/Microsoft.Health.Fhir.Core/Messages/Delete/ConditionalDeleteResourceRequest.cs
```diff
@@ -0,0 +1,41 @@
...
+            DeleteCount = deleteCount;
+        }
+
+        public string ResourceType { get; }
+
+        public IReadOnlyList<Tuple<string, string>> ConditionalParameters { get; }
+
+        public bool HardDelete { get; }
+
+        public int DeleteCount { get; }
```
Ivanidzo4ka, line 33: >DeleteCount [](http://example.com/codeflow?start=19&length=11)

MaxDeleteCount maybe?

src/Microsoft.Health.Fhir.Shared.Api/Controllers/FhirController.cs
```diff
@@ -368,6 +369,29 @@ public async Task<IActionResult> Delete(string typeParameter, string idParameter
...
+        /// <param name="hardDelete">A flag indicating whether to hard-delete the resource or not.</param>
+        /// <param name="maxDeleteCount">Specifies the maximum number of resources that can be deleted.</param>
+        [HttpDelete]
+        [Route(KnownRoutes.ResourceType)]
+        [AuditEventType(AuditEventSubType.Delete)]
+        public async Task<IActionResult> ConditionalDelete(string typeParameter, [FromQuery] bool hardDelete, [FromQuery(Name = KnownQueryParameterNames.Count)] int? maxDeleteCount)
+        {
+            IReadOnlyList<Tuple<string, string>> conditionalParameters = GetQueriesForSearch();
+
+            DeleteResourceResponse response = await _mediator.Send(new ConditionalDeleteResourceRequest(typeParameter, conditionalParameters, hardDelete, maxDeleteCount.GetValueOrDefault(1)), HttpContext.RequestAborted);
```
Ivanidzo4ka, line 385: >maxDeleteCount.GetValueOrDefault(1) [](http://example.com/codeflow?start=154&length=35)

Maybe it's worth to allow pass null here?

let's say I pass nothing, and I have two matching items, I would get Exception as result.

let's say I pass 2 and I have four items, now I don't get exception and 2 out 4 got deleted.

Now I pass 1 and I still get back to exception.

Feels weird...

Ivanidzo4ka (approved):
:shipit:

src/Microsoft.Health.Fhir.Shared.Api/Controllers/FhirController.cs
brendankowitz: `_count` is not technically used on delete requests. I was talking to @CaitlinV39 about this too. I think we should default to a `Single row` deletion behavior unless you specify its ok to delete more. I think we can just document what should happen

docs/rest/ConditionalDeleteRequests.http
```diff
@@ -0,0 +1,181 @@
...
+Authorization: Bearer {{bearer.response.body.access_token}}
+
+### Access last inserted Patient by history url (verify hard delete below)
+GET https://{{hostname}}/Patient/{{patient.response.body.id}}/_history/1
+content-type: application/json
+Authorization: Bearer {{bearer.response.body.access_token}}
+
+###
+# Hard Delete one resource conditionally 
+DELETE https://localhost:44348/Patient?identifier=1032704&hardDelete=true
```
brendankowitz, line 155:
```suggestion
DELETE https://{{hostname}}/Patient?identifier=1032704&hardDelete=true
```

docs/rest/ConditionalDeleteRequests.http
```diff
@@ -0,0 +1,181 @@
...
+content-type: application/json
+Authorization: Bearer {{bearer.response.body.access_token}}
+
+###
+# Delete multiple resources conditionally (no criteria = fails)
+DELETE https://localhost:44348/Patient?_count=100
```
brendankowitz, line 161:
```suggestion
DELETE https://{{hostname}}/Patient?_count=100
```

docs/rest/ConditionalDeleteRequests.http
```diff
@@ -0,0 +1,181 @@
...
+content-type: application/json
+Authorization: Bearer {{bearer.response.body.access_token}}
+
+###
+# Delete multiple resources conditionally 
+DELETE https://localhost:44348/Patient?identifier=1032704&_count=100
```
brendankowitz, line 167:
```suggestion
DELETE https://{{hostname}}/Patient?identifier=1032704&_count=100
```

docs/rest/ConditionalDeleteRequests.http
```diff
@@ -0,0 +1,181 @@
...
+content-type: application/json
+Authorization: Bearer {{bearer.response.body.access_token}}
+
+###
+# Delete multiple resources conditionally (out of range = fails)
+DELETE https://localhost:44348/Patient?identifier=1032704&_count=-1
```
brendankowitz, line 173:
```suggestion
DELETE https://{{hostname}}/Patient?identifier=1032704&_count=-1
```

docs/rest/ConditionalDeleteRequests.http
```diff
@@ -0,0 +1,181 @@
...
+content-type: application/json
+Authorization: Bearer {{bearer.response.body.access_token}}
+
+###
+# Hard Delete multiple resources conditionally 
+DELETE https://localhost:44348/Patient?identifier=1032704&hardDelete=true&_count=100
```
brendankowitz, line 179:
```suggestion
DELETE https://{{hostname}}/Patient?identifier=1032704&hardDelete=true&_count=100
```

docs/rest/ConditionalDeleteRequests.http
```diff
@@ -0,0 +1,181 @@
...
+###############################################################################
+### The Delete requests
+
+### Traditional DELETE
+DELETE https://{{hostname}}/Patient/{{patient.response.body.id}}
+content-type: application/json
+Authorization: Bearer {{bearer.response.body.access_token}}
+
+### Post One Patient above and Delete one resource conditionally 
+DELETE https://localhost:44348/Patient?identifier=1032704
```
brendankowitz, line 144:
```suggestion
DELETE https://{{hostname}}/Patient?identifier=1032704
```

brendankowitz (commented):
Update host names

PR #1937, opened 2021-05-13
Validates code value on search parameter creation

rotodd:
## Description
New search parameters can't have the same code value as other search parameters with the same base resource types. 

Before this PR, when a user would post a new search parameter with a duplicate code value, we would throw an exception in the `SearchParameterDefinitionBuilder`, after the search parameter information was added to the `SearchParameterDefinitionManager`'s `UrlLookup` dictionary. If the user posted the search parameter again after fixing the code value, they would see the error message: "A search parameter with the same definition URL already exists." 

This PR adds a check during search parameter validation to ensure the code value of a search parameter does not match the code value of any existing search parameters that have the same resource type. 

## Related issues
Addresses [AB#81596](https://microsofthealth.visualstudio.com/Health/_workitems/edit/81596).

## Testing
Adds new unit test cases and added an http file for manual testing.

## FHIR Team Checklist

None (bug fix)

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Parameters/SearchParameterValidator.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Parameters/SearchParameterValidator.cs
@@ -86,6 +86,10 @@ namespace Microsoft.Health.Fhir.Shared.Core.Features.Search.Parameters
                                     nameof(searchParam.Url),
                                     string.Format(Resources.SearchParameterDefinitionDuplicatedEntry, searchParam.Url)));
                         }
+                        else if (method.Equals(HttpPutName, StringComparison.OrdinalIgnoreCase))
+                        {
+                            CheckForConflictingCodeValue(searchParam, validationFailures);
+                        }
                     }
                     else
                     {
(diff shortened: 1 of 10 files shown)


src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Parameters/SearchParameterValidator.cs
```diff
@@ -75,18 +74,44 @@ public async Task ValidateSearchParamterInput(SearchParameter searchParam, strin
...
+                            validationFailures.Add(
+                                new ValidationFailure(
+                                    nameof(searchParam.Url),
+                                    string.Format(Resources.SearchParameterDefinitionNotFound, searchParam.Url)));
+                        }
+
+                        // Ensure the new search parameter's code value does not already exist for its base type(s)
+                        foreach (ResourceType? baseType in searchParam.Base)
+                        {
+                            if (_searchParameterDefinitionManager.TryGetSearchParameter(baseType.ToString(), searchParam.Code, out _))
```
namadabu, line 106: >baseType [](start = 88, length = 8)

Since `baseType` is nullable should we have a null check here?

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Parameters/SearchParameterValidator.cs
```diff
@@ -75,18 +74,44 @@ public async Task ValidateSearchParamterInput(SearchParameter searchParam, strin
...
+                        if (method.Equals(HttpPutName, StringComparison.OrdinalIgnoreCase) ||
+                            method.Equals(HttpDeleteName, StringComparison.OrdinalIgnoreCase))
+                        {
+                            validationFailures.Add(
+                                new ValidationFailure(
+                                    nameof(searchParam.Url),
+                                    string.Format(Resources.SearchParameterDefinitionNotFound, searchParam.Url)));
+                        }
+
+                        // Ensure the new search parameter's code value does not already exist for its base type(s)
```
namadabu, line 103: Is it possible for a PUT request to update an existing valid search parameter with wrong code information? If so, shouldn't we do this same validation for the following scenario - SearchParameter with same uri already exists and it is a PUT request (the if condition on line 78)?

rotodd: Good catch, thanks :). Fixed and added a new test case in 5258560.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Parameters/SearchParameterValidator.cs
rotodd: Good question. I first tried to test this by putting `null` as a base value in the search parameter json that is passed in, but it looks like it gets parsed away. 

So I then added this line right before the for-loop to see what would happen:

`searchParam.Base = new ResourceType?[] { ResourceType.Patient, null };`

and it looks like the null value gets converted to `""`. This fails gracefully since it is simply considered a code value that does not exist in the `TypeLookup` dictionary.

Let me know what you think - but I believe it should be alright to leave this as is.

namadabu (approved):
:shipit:

PR #1938, opened 2021-05-14
Add support for client id and secret in basic auth header

feordin:
## Description
Inferno tests uses Basic authorization header with client id and client secret in the header value.  This update moves one step closer to being able to run Inferno tests.

## Related issues
Addresses [issue #].

## Testing
Ran Inferno tests against our endpoint.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Api/Controllers/AadSmartOnFhirProxyController.cs
+++ b/src/Microsoft.Health.Fhir.Api/Controllers/AadSmartOnFhirProxyController.cs
@@ -352,6 +352,8 @@ namespace Microsoft.Health.Fhir.Api.Controllers
             {
                 EnsureArg.IsNotNull(compoundCode, nameof(compoundCode));
                 EnsureArg.IsNotNull(redirectUri, nameof(redirectUri));
+                EnsureArg.IsNotNull(clientId, nameof(clientId));
+                EnsureArg.IsNotNull(clientSecret, nameof(clientSecret));
             }
             catch (ArgumentNullException ex)
             {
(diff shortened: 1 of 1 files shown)


src/Microsoft.Health.Fhir.Api/Controllers/AadSmartOnFhirProxyController.cs
```diff
@@ -304,6 +303,23 @@ public AadSmartOnFhirProxyController(IOptions<SecurityConfiguration> securityCon
             var client = _httpClientFactory.CreateClient();
 #pragma warning restore CA2000
 
+            if (Request.Headers.TryGetValue("Authorization", out var authHeaderValues)
+                && string.IsNullOrEmpty(clientId))
```
LTA-Thinking, line 307: What should we do if the client passes an auth header and a client_id/client_secret form value?

src/Microsoft.Health.Fhir.Api/Controllers/AadSmartOnFhirProxyController.cs
```diff
@@ -356,7 +372,16 @@ public AadSmartOnFhirProxyController(IOptions<SecurityConfiguration> securityCon
 
             Uri callbackUrl = _urlResolver.ResolveRouteNameUrl(RouteNames.AadSmartOnFhirProxyCallback, new RouteValueDictionary { { "encodedRedirect", Base64UrlEncoder.Encode(redirectUri.ToString()) } });
 
-            // TODO: Deal with client secret in basic auth header
+            try
```
LTA-Thinking, line 375: Could this be combined with the try-catch on line 350?

src/Microsoft.Health.Fhir.Api/Controllers/AadSmartOnFhirProxyController.cs
feordin: As currently written, if the clientId is specified in the form value, then we will prioritize that value and ignore the auth header.   I suppose that is a valid approach, but am open to other suggestions.

src/Microsoft.Health.Fhir.Api/Controllers/AadSmartOnFhirProxyController.cs
feordin: Yes 👍

CaitlinV39:
@feordin - Do you think we will be able to merge this PR sometime in an upcoming sprint?

LTA-Thinking (approved):

PR #1946, opened 2021-05-17
Add health check for web sites

Ivanidzo4ka:
## Description
Our tests for Sql are failing down.
I've track issue to the root cause of two containers attempt to initialize schema for sql, but on of them fails, and basically put web site into permanent 500.

So with two web site I have 50% chance of getting 500 during request and that fails sql e2e tests.

I assume we don't encounter this issue in PaaS because we have configured health checks and dead container get restarted. So I'm adding health check monitoring to app service.

## Testing
E2E tests.

## FHIR Team Checklist

### DIFF
--- a/samples/templates/default-azuredeploy-docker.json
+++ b/samples/templates/default-azuredeploy-docker.json
@@ -324,7 +324,8 @@
                     "properties": {
                         "linuxFxVersion": "[concat('DOCKER|', variables('registryName'), '/', toLower(parameters('fhirVersion')), '_fhir-server',':', parameters('imageTag'))]",
                         "appCommandLine": "azure-fhir-api",
-                        "alwaysOn": true
+                        "alwaysOn": true,
+                        "healthCheckPath": "/health/check"
                     }
                 }
             ]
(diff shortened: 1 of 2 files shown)


namadabu (approved):
:shipit:

Ivanidzo4ka:
docker version of template tells me what health check is not configured:
image
but if I click on it:
image

I guess I'll merge PR and we will see if sql tests continue to fail.

samples/templates/default-azuredeploy-docker.json
```diff
@@ -324,7 +324,8 @@
                     "properties": {
                         "linuxFxVersion": "[concat('DOCKER|', variables('registryName'), '/', toLower(parameters('fhirVersion')), '_fhir-server',':', parameters('imageTag'))]",
                         "appCommandLine": "azure-fhir-api",
-                        "alwaysOn": true
+                        "alwaysOn": true,
+                        "healthCheckPath": "/health/check"
```
namadabu, line 328: @Ivanidzo4ka I am wondering whether this should be part of properties (line 275) instead of resources -> web -> properties.

Ivanidzo4ka: It doesn't work that way. I've run template with proposed changes.

PR #1951, opened 2021-05-19
Allow targeting reindex jobs to a specific resource type

feordin:
## Description
This adds a new parameter called targetResourceTypes for reindex jobs which will restrict the job to running only on specific resource types.

## Related issues
Addresses [issue [AB#81445](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/81445)].

## Testing
New E2E tests were created.

## FHIR Team Checklist

Feature


src/Microsoft.Health.Fhir.Core/Resources.resx
```diff
@@ -554,6 +554,10 @@
   <data name="MemberMatchNoMatchFound" xml:space="preserve">
     <value>No match found.</value>
   </data>
+  <data name="UnknownResourceType" xml:space="preserve">
```
Ivanidzo4ka, line 557: >UnknownResourceType [](http://example.com/codeflow?start=14&length=19)

there is `ResourceNotSupported`

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/Models/ReindexJobRecord.cs
```diff
@@ -20,12 +20,14 @@ public class ReindexJobRecord : JobRecord
     {
         public ReindexJobRecord(
             IReadOnlyDictionary<string, string> searchParametersHash,
+            IReadOnlyCollection<string> targetResourceTypes,
```
Ivanidzo4ka, line 23: >targetResourceTypes [](http://example.com/codeflow?start=40&length=19)

All other properties have [JsonProperty] attribute

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/Models/ReindexJobRecord.cs
```diff
@@ -104,6 +107,12 @@ protected ReindexJobRecord()
         [JsonProperty(JobRecordProperties.TargetDataStoreUsagePercentage)]
         public ushort? TargetDataStoreUsagePercentage { get; set; }
 
+        /// <summary>
+        /// A user can optionally limit the scope of the Reindex job to specific
+        /// resource types
+        /// </summary>
+        public IReadOnlyCollection<string> TargetResourceTypes { get; private set; } = new List<string>();
```
Ivanidzo4ka, line 114: >TargetResourceTypes [](http://example.com/codeflow?start=43&length=19)

All other properties have [JsonProperty] attribute

src/Microsoft.Health.Fhir.Shared.Core/Extensions/ReindexJobRecordExtensions.cs
```diff
@@ -55,6 +55,7 @@ public static ResourceElement ToParametersResourceElement(this ReindexJobWrapper
             parametersResource.Parameter.Add(new Parameters.ParameterComponent() { Name = JobRecordProperties.MaximumConcurrency, Value = new FhirDecimal(job.MaximumConcurrency) });
             parametersResource.Parameter.Add(new Parameters.ParameterComponent() { Name = JobRecordProperties.Resources, Value = new FhirString(job.ResourceList) });
             parametersResource.Parameter.Add(new Parameters.ParameterComponent() { Name = JobRecordProperties.SearchParams, Value = new FhirString(job.SearchParamList) });
+            parametersResource.Add(JobRecordProperties.TargetResourceTypes, new FhirString(job.TargetResourceTypeList));
```
Ivanidzo4ka, line 58: >parametersResource.Add [](http://example.com/codeflow?start=12&length=22)

It really stands out...
Maybe it's worth to rewrite other parametes as well?

src/Microsoft.Health.Fhir.Tests.Common/TestFiles/Normative/SearchParameter-Resource-id.json
```diff
@@ -0,0 +1,42 @@
+{
+    "resourceType": "SearchParameter",
+    "id": "Resource-idfoo",
+    "extension": [
+        {
+            "url": "http://hl7.org/fhir/StructureDefinition/structuredefinition-standards-status",
+            "valueCode": "trial-use"
+        }
+    ],
+    "url": "http://hl7.org/fhir/SearchParameter/Resource-idfoo",
```
Ivanidzo4ka, line 10: >idfoo [](http://example.com/codeflow?start=57&length=5)

I would change name of file to `idfoo` as well.

src/Microsoft.Health.Fhir.Tests.Common/TestFiles/R5/Immunization.json
```diff
@@ -0,0 +1,122 @@
+{
+    "resourceType": "Immunization",
```
Ivanidzo4ka, line 2: minor, but you can put R4 version into normative, and keep only stu3 version. That way you will add only two files instead of three.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -173,10 +178,42 @@ private async Task<bool> TryPopulateNewJobFields()
         {
             // Build query based on new search params
             // Find supported, but not yet searchable params
-            var notYetIndexedParams = _supportedSearchParameterDefinitionManager.GetSearchParametersRequiringReindexing();
+            var possibleNotYetIndexedParams = _supportedSearchParameterDefinitionManager.GetSearchParametersRequiringReindexing();
+            _notYetIndexedParams = new List<SearchParameterInfo>();
```
Ivanidzo4ka, line 182: >_notYetIndexedParams [](http://example.com/codeflow?start=12&length=20)

It looks like it's been used only inside `TryPopulateNewJobFields` function. Maybe it should be local variable?

feordin: Correct, changed to local.

src/Microsoft.Health.Fhir.Core/Resources.resx
feordin: Yep, that looks good, better to not add a new one unnecesarily.

src/Microsoft.Health.Fhir.Shared.Core/Extensions/ReindexJobRecordExtensions.cs
feordin: Yes, it does look cleaner.

src/Microsoft.Health.Fhir.Tests.Common/TestFiles/Normative/SearchParameter-Resource-id.json
feordin: Agreed, updated.

src/Microsoft.Health.Fhir.Tests.Common/TestFiles/R5/Immunization.json
feordin: I left it, just in case we need to customize.

rotodd (commented):
+1 looks good!

As we discussed, we'll wait to merge this until a fix for the existing reindexing bug is in so we can double check that the manual test scenario behaves as expected

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -173,10 +178,42 @@ private async Task<bool> TryPopulateNewJobFields()
...
+            var resourceList = new HashSet<string>();
+
+            // filter list of SearchParameters by the target resource types
+            if (_reindexJobRecord.TargetResourceTypes.Any())
+            {
+                foreach (var searchParam in possibleNotYetIndexedParams)
+                {
+                    var searchParamResourceTypes = GetDerivedResourceTypes(searchParam.BaseResourceTypes);
+                    var matchingResourceTypes = searchParamResourceTypes.Intersect(_reindexJobRecord.TargetResourceTypes);
+                    if (matchingResourceTypes.Any())
```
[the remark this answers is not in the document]
feordin: I am not sure if that is a warning, as it is the goal of specifying target types in the reindex job.  The user wants to avoid reindexing certain searchparameters, at least in some scenarios.  But more info is good, I will add an info log.

src/Microsoft.Health.Fhir.Shared.Core/Extensions/ReindexMediatorExtensions.cs
```diff
@@ -21,11 +23,26 @@ public static class ReindexMediatorExtensions
...
 
-            var request = new CreateReindexRequest(maximumConcurrency, maxResourcesPerQuery, queryDelay, targetDataStoreResourcePercentage);
+            var targetResourceTypes = new List<string>();
+
+            if (!string.IsNullOrEmpty(targetResourceTypesString))
+            {
+                targetResourceTypes.AddRange(targetResourceTypesString.Split(","));
+                foreach (var resourceType in targetResourceTypes)
+                {
+                    if (!Enum.TryParse<Hl7.Fhir.Model.ResourceType>(resourceType, out var result))
```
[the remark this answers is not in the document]
feordin: good catch, I will trim the whitespace.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -629,7 +638,21 @@ private async Task CalculateTotalAndResourceCounts()
 
         private async Task UpdateParametersAndCompleteJob()
         {
-            (bool success, string error) = await _reindexUtilities.UpdateSearchParameterStatus(_reindexJobRecord.SearchParams.ToList(), _cancellationToken);
+            // here we check if all the resource types which are base types of the search parameter
+            // were reindexed by this job.  If so, then we should mark the search parameters
```
Ivanidzo4ka, line 642: >were reindexed by this job [](http://example.com/codeflow?start=15&length=26)

It doesn't seems right.

I reindex one type out of two to make search parameter work for that type.

To fully enable searchparameter I need to run reindex job for both of them, and there is reason why I opt out of running it only for one type.

Which mean user would have to constantly use header based approach to run searches, which would be upsetting.

feordin: This is true, with this implementation the user would have to continually use the header.  We do not currently track status of whether a searchparameter is fully indexed for a given type, only for all types.  To support searching after indexing only one type, we would need to add more data to the tracking logic and track search parameter status per resource type.

CaitlinV39:
@rotodd

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -645,6 +665,36 @@ private async Task UpdateParametersAndCompleteJob()
             }
         }
 
+        private ICollection<string> GetDerivedResourceTypes(IReadOnlyCollection<string> resourceTypes)
+        {
+            var completeResourceList = new HashSet<string>(resourceTypes);
+
+            foreach (var possibleBaseType in resourceTypes)
```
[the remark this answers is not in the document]
rotodd: Confirmed that this was renamed in f9a7c19.

rotodd (approved):
Gave this a second pass. Looks good 😎

src/Microsoft.Health.Fhir.Core/Features/Operations/JobRecordProperties.cs
```diff
@@ -122,5 +122,7 @@ public static class JobRecordProperties
         public const string QueryDelayIntervalInMilliseconds = "queryDelayIntervalInMilliseconds";
 
         public const string TargetDataStoreUsagePercentage = "targetDataStoreUsagePercentage";
+
+        public const string TargetResourceTypes = "targetResourceTypes";
```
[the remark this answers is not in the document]
rotodd: On second look, I don't love these suggested names^. I'm thinking about this as "the target resource types at the reindex operation level", whereas the target resource types specified in a search parameter are "the target resource types at the search parameter level". Maybe we can think about distinguishing the two with naming and/or comments in the future, but I don't think it should hold up this PR.

Issue #1941, opened 2021-05-15
Unable to create Practitionerrole-network search parameter


CaitlinV39:
**Describe the bug**
When trying to post a Da Vinci search parameter, receive an error.

http://hl7.org/fhir/us/davinci-pdex-plan-net/SearchParameter/practitionerrole-network

**FHIR Version?**
Stu3/R4/R5

**Data provider?**
CosmosDB/SQL Server

**To Reproduce**
1. Post the search parameter http://hl7.org/fhir/us/davinci-pdex-plan-net/SearchParameter/practitionerrole-network

**Expected behavior**
Search parameter created

**Actual behavior**
Receive this outcome:
{
  "resourceType": "OperationOutcome",
  "id": "50828dda-50e3-4414-9f92-ca8fbfbeb935",
  "issue": [
    {
      "severity": "error",
      "code": "not-supported",
      "diagnostics": "The search parameter with definition URL 'http://hl7.org/fhir/us/davinci-pdex-plan-net/SearchParameter/practitionerrole-network' is not supported."
    },
    {
      "severity": "error",
      "code": "exception",
      "diagnostics": "An error occurred creating the custom search parameter.  The issue must be resolved and the parameter resubmitted to become functional."
    }
  ]
}

CaitlinV39:
@feordin @rotodd could one of you take a look?

CaitlinV39:
AB#82413

PR #1952, opened 2021-05-19
Updates search parameter error handling

rotodd:
## Description
Before, if we posted a search parameter that didn't have a complete expression in the search parameter definition, we returned an error that simply stated that the search parameter was not supported.

This PR updates the error message that is returned to be more specific. 

## Related issues
Addresses [AB#82165](https://microsofthealth.visualstudio.com/Health/_workitems/edit/82165) and #1941.

## Testing
Adds error message check to existing unit tests and adds new .http test file for manual testing.

## FHIR Team Checklist

Skip (minor error message tweaking)

### DIFF
--- a/docs/rest/ReindexFlowReferenceExtension.http
+++ b/docs/rest/ReindexFlowReferenceExtension.http
@@ -1,0 +1,219 @@
+# This test flow confirms that reindexing operations can handle search parameters
+# that target the network extension of PractitionerRole resources.
+# It also checks that the appropriate error message is returned when the
+# search parameter's expression is not specific enough.
+#
+# This test assumes the following local environment setup:
+# 1. appsettings.json has Security.Enabled = false
+# 2. The version is R4 or R5
+
+@baseUrl = https://localhost:44348
+@contentType = application/json
+
+###
+# Create a PractitionerRole resource with a network extension.
+POST {{baseUrl}}/PractitionerRole HTTP/1.1
+content-type: {{contentType}}
+
+{
+  "resourceType" : "PractitionerRole",
+  "id" : "CounselorRole1",
+  "meta" : {
+    "lastUpdated" : "2020-07-07T13:26:22.0314215+00:00",
+    "profile" : [
+      "http://hl7.org/fhir/us/davinci-pdex-plan-net/StructureDefinition/plannet-PractitionerRole"
+    ]
+  },
+  "language" : "en-US",
+  "text" : {
+    "status" : "extensions",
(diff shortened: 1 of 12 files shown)


docs/rest/ReindexFlowReferenceExtension.http
```diff
@@ -0,0 +1,219 @@
...
+        }
+      ]
+    }
+  ],
+  "code": "network-incomplete-expression",
+  "base": [
+    "PractitionerRole"
+  ],
+  "type": "reference",
+  "expression": "PractitionerRole.extension.where(url='http://hl7.org/fhir/us/davinci-pdex-plan-net/StructureDefinition/network-reference')",
```
rotodd, line 207: @Ivanidzo4ka pointed out that this is how the expression is specified in the spec [here](https://build.fhir.org/ig/HL7/davinci-pdex-plan-net/SearchParameter-practitionerrole-network.json.html). Currently looking to see if we can add support to parse expressions that don't specify the exact property in the extension.

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/ISearchParameterSupportResolver.cs
```diff
@@ -12,7 +12,10 @@ public interface ISearchParameterSupportResolver
         /// <summary>
         /// Determines if the given search parameter is able to be indexed
         /// </summary>
-        /// <param name="parameterInfo">Search Parameter info</param>
-        (bool Supported, bool IsPartiallySupported) IsSearchParameterSupported(SearchParameterInfo parameterInfo);
+        /// <param name="parameterInfo">The search parameter to check.</param>
+        /// <returns name="Supported">True if one or more matching converters is found.</returns>
+        /// <returns name="IsPartiallySupported">False if all matching converters are found or no matching converters are found, true otherwise.</returns>
+        /// <returns name="ErrorMessage">Provides more detail if a search parameter is not supported.</returns>
+        (bool Supported, bool IsPartiallySupported, string ErrorMessage) IsSearchParameterSupported(SearchParameterInfo parameterInfo);
```
feordin, line 19: Three return values for a method seem like a lot.  Not sure if we need to change it, but perhaps these should be contained in a struct? or other method.  Just a suggestion.

feordin (approved):

rotodd:
While the error handling here is useful, I think it makes sense to close this for now and investigate adding support to resolve types for custom search parameters that target extensions without using the `as(<fhirType>)` function.

PR #1957, opened 2021-05-20
Fixes a race condition when reading SearchParameter resources from datastore

feordin:
## Description
Reading existing SearchParameters from the database occurs simultaneously with setting the SearchParameters status values.  This code sets an explicit order between the two activities.

## Related issues
Addresses [issue [AB#82254](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/82254)].

## Testing
Existing E2E tests were run, unit tests update for this scenario.

## FHIR Team Checklist

Skip


src/Microsoft.Health.Fhir.Shared.Api/Modules/SearchModule.cs
```diff
@@ -79,12 +80,7 @@ public void Load(IServiceCollection services)
             services.Add<SearchParameterStatusManager>()
                 .Singleton()
                 .AsSelf()
-                .ReplaceService<INotificationHandler<StorageInitializedNotification>>();
```
Ivanidzo4ka, line 82: >StorageInitializedNotification [](http://example.com/codeflow?start=53&length=30)

You probably want to delete this class as well.

feordin: StorageInitializedNotification is still being published by the data layer, (cosmos or SQL) and handled by SearchParameterDefinitionManager

src/Microsoft.Health.Fhir.Core/Messages/Search/SearchParameterDefinitionManagerInitialized.cs
```diff
@@ -0,0 +1,16 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using MediatR;
+
+namespace Microsoft.Health.Fhir.Core.Messages.Search
+{
+    /// <summary>
+    /// A notification that is raised when the SearchParameters are initialized
+    /// </summary>
+    public class SearchParameterDefinitionManagerInitialized : INotification
```
rotodd, line 13: Nit: if we're renaming `SearchParametersUpdated` to be `SearchParametersUpdatedNotification`, does it make sense for this to be `SearchParameterDefinitionManagerInitializedNotification`?

Ivanidzo4ka (approved):
:shipit:

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/SearchParameterOperations.cs
```diff
@@ -107,6 +107,11 @@ public async Task DeleteSearchParameterAsync(RawResource searchParamResource)
         {
             try
             {
+                // We need to make sure we have the latest search parameters before trying to delete
+                // existing search parameter. This is to avoid trying to update a search parameter that
+                // was recently added and that hasn't propogated to all fhir-server instances.
+                await GetAndApplySearchParameterUpdates(CancellationToken.None);
```
rotodd, line 113: I noticed that we also call this in `UpdateSearchParameterAsync()`. 

For my understanding, why isn't this needed in `AddSearchParameterAsync()` as well? I'm wondering if there a scenario where we could get duplicate search parameters if a newly added param isn't synced to all instances

rotodd (approved):

src/Microsoft.Health.Fhir.Core/Messages/Search/SearchParameterDefinitionManagerInitialized.cs
rotodd: (we can defer this to another PR if you'd like to get this merged)

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/SearchParameterOperations.cs
Ivanidzo4ka: Not an author, but I spend some time thinking about it, and I would say, what delete and update prefer to have updated collection since you modify it in a certain way and you want to make sure you modify latest version.

With adding new search parameter only thing you care about is uniquness of search parameter, and I think it's get validated before we get to this place.

PR #1959, opened 2021-05-21
Adds caching for CosmosDbSearchParameterStatusDataStore

brendankowitz:
## Description
Checks the last updated date before fetching all SearchParameterStatuses.

This brings the RU usage from ~60 to ~2.8

## Related issues
Addresses [issue #].

## Testing

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Registry/CosmosDbSearchParameterStatusDataStore.cs
+++ b/src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Registry/CosmosDbSearchParameterStatusDataStore.cs
@@ -18,11 +18,13 @@ using Microsoft.Health.Fhir.CosmosDb.Configs;
 
 namespace Microsoft.Health.Fhir.CosmosDb.Features.Storage.Registry
 {
-    public class CosmosDbSearchParameterStatusDataStore : ISearchParameterStatusDataStore
+    public sealed class CosmosDbSearchParameterStatusDataStore : ISearchParameterStatusDataStore, IDisposable
     {
         private readonly Func<IScoped<Container>> _containerScopeFactory;
-        private readonly CosmosDataStoreConfiguration _cosmosDataStoreConfiguration;
         private readonly ICosmosQueryFactory _queryFactory;
+        private DateTimeOffset? _lastRefreshed = null;
+        private List<ResourceSearchParameterStatus> _statusList = new();
+        private readonly SemaphoreSlim _statusListSemaphore = new(1, 1);
 
         public CosmosDbSearchParameterStatusDataStore(
             Func<IScoped<Container>> containerScopeFactory,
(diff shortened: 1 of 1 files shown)


feordin:
I think the idea of querying only when the datetime indicates that we need to is a very good idea.  I wonder if we should put this logic in the SearchParameterStatusManager as it is already tracking a timestamp for the last time the searchparameter status was checked.  In the data layer perhaps we add a simple method to get the latest timestamp, and the status manager then compares that against the last time?  This way that same simple method can be implemented for both SQL and Cosmos and the status manager will contain the logic.

namadabu (approved):
:shipit:

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Registry/CosmosDbSearchParameterStatusDataStore.cs
```diff
@@ -34,43 +36,82 @@ public class CosmosDbSearchParameterStatusDataStore : ISearchParameterStatusData
...
+                    new QueryRequestOptions
+                    {
+                        PartitionKey = new PartitionKey(SearchParameterStatusWrapper.SearchParameterStatusPartitionKey),
+                        MaxItemCount = 1,
+                    }));
 
-            return parameterStatus;
+            var response = await query.ExecuteNextAsync(cancellationToken);
+
+            return response.FirstOrDefault();
```
LTA-Thinking, line 114: What happens if the query returns no results?

brendankowitz: It might be DateTimeOffset.MinValue, which should be fine, its only matters when there is a date that is more current

LTA-Thinking (approved):

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Registry/CosmosDbSearchParameterStatusDataStore.cs
```diff
@@ -34,43 +36,82 @@ public class CosmosDbSearchParameterStatusDataStore : ISearchParameterStatusData
...
+                _statusListSemaphore.Release();
+            }
+        }
+
+        private async Task<DateTimeOffset> GetSearchParameterStatusesLastUpdatedAsync(IScoped<Container> container, CancellationToken cancellationToken)
+        {
+            var query = _queryFactory.Create<DateTimeOffset>(
+                container.Value,
+                new CosmosQueryContext(
+                    new QueryDefinition("select value c.lastUpdated from c order by c.lastUpdated desc"),
```
namadabu, line 105: this query will miss detection of any deleted search parameter status documents. We need to check for the count as well. Working on an update that does that.

PR #1962, opened 2021-05-21
Adds files for testing C4BB

CaitlinV39:
## Description
Adds some rest files for testing C4BB and getting the system setup with search parameters and profiles

## Related issues
N/A

## Testing

## FHIR Team Checklist
[X] **Update the title** of the PR to be succinct and less than 50 characters
[X] **Add a milestone** to the PR for the sprint that it is merged (i.e. add S47)
[X] Tag the PR with the type of update: **Bug**, **Dependencies**, **Enhancement**, or **New-Feature**
[X] Tag the PR with **Azure API for FHIR** if this will release to the managed service

Skip

### DIFF
--- a/docs/rest/CB44/C4BB.http
+++ b/docs/rest/CB44/C4BB.http
@@ -1,0 +1,130 @@
+# This file includes all the search parameters, reindex operation, and profiles to load to get the necessary pieces for the Carin Blue Button IG
+### REST Client
+
+@fhirurl= <FHIR SERVER URL>
+@clientid= <CLIENT ID>
+@clientsecret= <CLIENT SECRET>
+@tenantid= <TENANT ID>
+@contentType = application/json
+
+@reindexJobId = <REINDEX JOB ID>
+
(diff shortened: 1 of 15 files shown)


CaitlinV39:
@Ivanidzo4ka LMK if this is easier to manage :)

Ivanidzo4ka:
Yeah, it's much easier to operate over C4BB.http file.

I think it's ok in it's current state, but if you willing to spend some more time I would do:

- rename `txt` files to `json` so they can be opened as json files.
- move all resources from C4BB_Sampl_Resources to files, it's a mix bag now of two first resource are in files and rest are in http file.

Ivanidzo4ka (approved):

CaitlinV39:
@Ivanidzo4ka - I'll fix the json files now but will have to close and recreate :)

On the other piece with being consistent, I'm trying to pass in IDs into the resources so maybe they have to be directly in the http file? Do you know a way around this?

Issue #1974, opened 2021-05-26
Sql server can't process resource for custom search parameter.


Ivanidzo4ka:
**Describe the bug**
If there is two or more instances of fhir-server adding search parameter to resource can lead to other instances of fhir-server would fail to process resources for that search parameter.

`SqlServerFhirModel` has a `private Dictionary<Uri, short> _searchParamUriToId;` which contains mapping of searchParameter to it's id in sql server.

Currently only place which update this mapping is `UpsertStatuses` in `SqlServerSearchParameterStatusDataStore` which we call during `UpdateSearchParameterStatusAsync` in `SearchParameterStatusManager`.
Which works for fhir-server instance which accepting Post request to add search parameter.

If we syncing search parameters via `GetAndApplySearchParameterUpdates` in `SearchParameterOperations` we don't call that operation.

In case if resource get posted and we already synced search parameters we will extract search indices and will try to save them to database. To save them to database we need mapping between url of search parameter and it's id in db. For fhir-server instances which didn't process original post request that dictionary wouldn't have mapping and fails with following exception:

The given key 'http://hl7.org/fhir/SearchParameter/Patient-917dd1e1bf838a9c5893706650d088c6f1ff' was not present in the dictionary.

/repo/src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/BulkSearchParameterRowGenerator.cs: 51

**FHIR Version?**
Stu3/R4/R5

**Data provider?**
SQL Server

**To Reproduce**
1. 
2. 
3. 

**Expected behavior**
We can post resource for new search parameter.

**Actual behavior**
We can't post resource for new search parameter.

CaitlinV39:
AB#82454

PR #1976, opened 2021-05-28
Synchronizes search parameters in SQL

rotodd:
## Description
When we update or create search parameters in SQL, we need to reflect the changes in the C# dictionary that maps a search parameter's URI to its ID.

Before this PR, we were only doing this in the `UpsertStatuses()` method in `SqlServerSearchParameterStatusDataStore.cs`, which successfully updated the C# dictionary for the calling instance but not any other instances. This was causing us to see the error: `The given key 'http://hl7.org/fhir/SearchParameter/Patient-917dd1e1bf838a9c5893706650d088c6f1ff' was not present in the dictionary.`.

This PR:
1. Adds a `SyncStatuses` method that is called from the `SearchParameterStatusManager`, which synchronizes `dbo.SearchParam` with the `SqlServerFhirModel` dictionary, `_searchParamUriToId`
2. Adds a new schema file that makes the `GetSearchParamStatuses` stored proc return search parameter ID in addition to the `UpsertSearchParamStatuses` stored proc
3. Creates a SQL-only `ResourceSearchParameterStatus` class to track the search parameter ID in SQL

## Related issues
Addresses [AB#82454](https://microsofthealth.visualstudio.com/Health/_workitems/edit/82454) and #1974.

## Testing
Ran two instances locally, created a new search parameter, and created a resource that the new search parameter targets.

## FHIR Team Checklist

Patch (new schema version)


src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterStatusManager.cs
```diff
@@ -191,6 +191,8 @@ internal async Task ApplySearchParameterStatus(IReadOnlyCollection<ResourceSearc
 
             _latestSearchParams = updatedSearchParameterStatus.Select(p => p.LastUpdated).Max();
 
+            _searchParameterStatusDataStore.SyncStatuses(updatedSearchParameterStatus);
```
rotodd, line 194: TODO: Currently investigating a potential timing issue that entails the C# dictionary being accessed before the search parameter info is synchronized.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/12.diff.sql
```diff
@@ -0,0 +1,56 @@
...
+--     UpsertSearchParams
+--
+-- DESCRIPTION
+--     Given a set of search parameters, creates or updates the parameters.
+--
+-- PARAMETERS
+--     @searchParams
+--         * The updated existing search parameters or the new search parameters
+--
+ALTER PROCEDURE dbo.UpsertSearchParams
```
Ivanidzo4ka, line 32: >UpsertSearchParams [](http://example.com/codeflow?start=20&length=18)

@summaryOfChanges is no longer needed?

src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterStatusManager.cs
rotodd: As discussed with @feordin, it makes sense to keep the single-instance dictionary update in the `UpsertStatuses()` method. This will ensure the dictionary for the calling instance is immediately updated, since there is a delay in propagating the notification to other instances. I added this back in in 101d872.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/12.diff.sql
rotodd: It looks like we won't need to make this fix now that we are keeping `dbo.UpsertSearchParams` as is. But good catch, thanks!

feordin (approved):

rotodd:
## Next Steps

Remove the C# dictionary that tracks the URI to ID mapping a replace it with a table in SQL. Storing this info at the data store layer will prevent us from having to worry about synchronizing across instances. I created a work item to track this: [AB#82593](https://microsofthealth.visualstudio.com/Health/_workitems/edit/82593)

FYI @feordin

Issue #1985, opened 2021-06-01
Paging through results for queries with multiple _type parameters returns 500


namadabu:
**Describe the bug**
For search requests that have multiple _type parameters (_type=resource1,resource2,...) we get 500 responses in certain situations. It happens when we are paging through results for the last resource type that is returned.

**FHIR Version?**
Stu3/R4/R5

**Data provider?**
SQL Server

**To Reproduce**
1. Assuming starting from a blank fhir-server, POST 5 patients and 5 observations (can be any resources) to the server.
2. Do the following search: GET <fhir-server>/_type=Patient,Observation&_count=6 and follow the continuation token.
3. When following the continuation token, we get a 500 response.

**Expected behavior**
200 response with resources.

**Actual behavior**
500 response.

CaitlinV39:
@namadabu - Can this be closed?

PR #1986, opened 2021-06-01
Fix SQL query gen for multiple _type params

namadabu:
## Description
As mentioned in the issue, in some situations we generate an incorrect query when paging through results for queries that contain multiple `_type` parameters. This happens when we are returning the results for the "last" resource type. An example of the query we generate:

```
SELECT DISTINCT TOP (@p0) r.ResourceTypeId, r.ResourceId, r.Version, r.IsDeleted, r.ResourceSurrogateId, r.RequestMethod, CAST(1 AS bit) AS IsMatch, CAST(0 AS bit) AS IsPartial, r.IsRawResourceMetaSet, r.SearchParamHash, r.RawResource
      FROM dbo.Resource r
      WHERE (
              ResourceTypeId = @p1 AND ResourceSurrogateId > @p2
              OR ResourceTypeId IN ()
          )
          AND IsHistory = 0
          AND IsDeleted = 0
      ORDER BY r.ResourceTypeId ASC, r.ResourceSurrogateId ASC
```

The extra `()` throws an exception since that is not a valid query. The `OR` clause is only required when there are other resource types to search for. Since we are only returning the resources of the "last" resource type (represented by p2) we don't need the `OR` clause. 

Updated the query generator logic to add the `OR` clause only when there are other resource types to search for.

## Related issues
Addresses #1985 

## Testing
Added new tests. Manual tests on local deployment.

## FHIR Team Checklist

### DIFF
--- a/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/BasicSearchTests.cs
+++ b/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/BasicSearchTests.cs
@@ -49,24 +49,11 @@ namespace Microsoft.Health.Fhir.Tests.E2E.Rest.Search
         {
             // Create various resources.
             Patient[] patients = await Client.CreateResourcesAsync<Patient>(
-                p => SetPatientInfo(p, "Seattle", "Robinson"),
-                p => SetPatientInfo(p, "Portland", "Williamas"),
-                p => SetPatientInfo(p, "Seattle", "Jones"));
+                p => SetPatientInfo(p, "Seattle", "Robinson", tag: null),
+                p => SetPatientInfo(p, "Portland", "Williamas", tag: null),
+                p => SetPatientInfo(p, "Seattle", "Jones", tag: null));
 
             await ExecuteAndValidateBundle("Patient?address-city=seattle&family=Jones", patients[2]);
-
-            void SetPatientInfo(Patient patient, string city, string family)
-            {
-                patient.Address = new List<Address>()
-                {
-                    new Address() { City = city },
-                };
-
-                patient.Name = new List<HumanName>()
-                {
-                    new HumanName() { Family = family },
-                };
-            }
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/BasicSearchTests.cs
```diff
@@ -192,6 +164,68 @@ public async Task GivenVariousTypesOfResources_WhenSearchingAcrossAllResourceTyp
...
+            {
+                expectedResources.Add(patient);
+            }
+
+            foreach (Observation observation in observations)
+            {
+                expectedResources.Add(observation);
+            }
+
+            await ExecuteAndValidateBundle($"?_type=Patient,Observation&_tag={tag}&_count=3", sort: false, pageSize: 3, expectedResources.ToArray());
```
Ivanidzo4ka, line 226: >ExecuteAndValidateBundle [](http://example.com/codeflow?start=18&length=24)

nit: can we sort by lastupdated and have one test?

namadabu: I had two separate tests because I wanted to highlight the difference in behavior between Cosmos and SQL for such queries.

Ivanidzo4ka (approved):
:shipit:

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/BasicSearchTests.cs
```diff
@@ -49,24 +49,11 @@ public async Task GivenResourceWithVariousValues_WhenSearchedWithMultipleParams_
         {
             // Create various resources.
             Patient[] patients = await Client.CreateResourcesAsync<Patient>(
-                p => SetPatientInfo(p, "Seattle", "Robinson"),
-                p => SetPatientInfo(p, "Portland", "Williamas"),
-                p => SetPatientInfo(p, "Seattle", "Jones"));
+                p => SetPatientInfo(p, "Seattle", "Robinson", tag: null),
+                p => SetPatientInfo(p, "Portland", "Williamas", tag: null),
+                p => SetPatientInfo(p, "Seattle", "Jones", tag: null));
```
LTA-Thinking, line 54: Should we add a tag to these tests? I know that not having tags has caused flaky tests.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/BasicSearchTests.cs
```diff
@@ -77,30 +64,15 @@ public async Task GivenResourceWithVariousValues_WhenSearchedWithCityParam_ThenO
...
-                p => SetPatientInfo(p, "Amsterdam", "Hill"));
+                p => SetPatientInfo(p, "seattle", "Robinson", tag), // match
+                p => SetPatientInfo(p, "Portland", "Williamas", tag),
+                p => SetPatientInfo(p, "SEATTLE", "Skouras", tag), // match
+                p => SetPatientInfo(p, "Sea", "Luecke", tag),
+                p => SetPatientInfo(p, "Seattle", "Jones", tag), // match
+                p => SetPatientInfo(p, "New York", "Cook", tag),
+                p => SetPatientInfo(p, "Amsterdam", "Hill", tag));
 
             await ExecuteAndValidateBundle("Patient?address-city=Seattle", patients[0], patients[2], patients[4]);
```
LTA-Thinking, line 75: Not part of this PR, but this test defines a tag but doesn't use it in its search.

LTA-Thinking (approved):

PR #1992, opened 2021-06-02
Add initial import operation

tongwu-msft:
## Description
Add import operation for customer to initial load data to FHIR service.

Sample request:
```
curl --location --request POST 'https://localhost:44348/$import' \
--header 'Accept: application/fhir+json' \
--header 'Prefer: respond-async' \
--header 'Content-Type: application/fhir+json' \
--data-raw '{
	"resourceType": "Parameters",
	"parameter": [
		{
			"name": "inputFormat",
			"valueString": "application/fhir+ndjson"
                },
                {
			"name": "mode",
			"valueString": "InitialLoad"
                },
		{
			"name": "input",
			"part": [
				{
					"name": "type",
					"valueString": "Patient"
				},
				{
					"name": "url",
					"valueUri": "http://127.0.0.1:10000/devstoreaccount1/0000/Patient.ndjson"
				}
			]
		},
		{
			"name": "input",
			"part": [
				{
					"name": "type",
					"valueString": "Patient"
				},
				{
					"name": "url",
					"valueUri": "http://127.0.0.1:10000/devstoreaccount1/0000/PatientErr.ndjson"
				}
			]
		},
		{
			"name": "storageDetail",
			"part": [
				{
					"name": "type",
					"valueString": "azure-blob"
				}
			]
		}
	]
}'
```
Sample result
```
{
    "transactionTime": "2021-05-17T06:23:47.0892657+00:00",
    "request": "https://localhost:44348/$import",
    "output": [
        {
            "type": "Patient",
            "count": 4,
            "inputUrl": "http://127.0.0.1:10000/devstoreaccount1/0000/Patient.ndjson"
        },
        {
            "type": "Patient",
            "count": 4,
            "inputUrl": "http://127.0.0.1:10000/devstoreaccount1/0000/PatientErr.ndjson"
        }
    ],
    "error": [
        {
            "type": "Patient",
            "count": 1,
            "inputUrl": "http://127.0.0.1:10000/devstoreaccount1/0000/PatientErr.ndjson",
            "url": "http://127.0.0.1:10000/devstoreaccount1/fhirlogs/Patient38d5d864efcc40ecbdf31ee63dea7f89.ndjson"
        }
    ]
}
```

## Related issues
Addresses [issue #1867 ].

## Testing
1. Unit tests for each component.
2. Integration test for SQL operation
3. E2E tests for import operation
4. E2E tests for imported FHIR resource with current search test cases.

## FHIR Team Checklist


LTA-Thinking:
Some thoughts:
- How should we test this in CI? We currently don't reset our database between runs.
- In PaaS we will need a control for users to manage whether import is enabled or not. I don't think customers would want import enabled after their service is in use.

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/GetBulkImportRequestHandler.cs
```diff
@@ -0,0 +1,84 @@
...
+
+            _taskManager = taskManager;
+            _authorizationService = authorizationService;
+        }
+
+        public async Task<GetImportResponse> Handle(GetImportRequest request, CancellationToken cancellationToken)
+        {
+            EnsureArg.IsNotNull(request, nameof(request));
+
+            if (await _authorizationService.CheckAccess(DataActions.Import, cancellationToken) != DataActions.Import)
```
LTA-Thinking, line 38: Should only import users be able to view import job status?

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/GetBulkImportRequestHandler.cs
```diff
@@ -0,0 +1,84 @@
...
+using Microsoft.Health.Core.Features.Security.Authorization;
+using Microsoft.Health.Fhir.Core.Exceptions;
+using Microsoft.Health.Fhir.Core.Features.Security;
+using Microsoft.Health.Fhir.Core.Messages.Import;
+using Microsoft.Health.TaskManagement;
+using Newtonsoft.Json;
+
+namespace Microsoft.Health.Fhir.Core.Features.Operations.Import
+{
+    public class GetBulkImportRequestHandler : IRequestHandler<GetImportRequest, GetImportResponse>
```
LTA-Thinking, line 20: Can this be renamed to GetImportRequestHandler to fit with the pattern that the Handler, Request, and Response all have the same prefix?

src/Microsoft.Health.Fhir.Core/Messages/BulkImport/GetImportRequest.cs
```diff
@@ -0,0 +1,33 @@
...
+            EnsureArg.IsNotNullOrWhiteSpace(taskId, nameof(taskId));
+
+            RequestUri = requestUri;
+            TaskId = taskId;
+        }
+
+        /// <summary>
+        /// Request uri for get request
+        /// </summary>
+        public Uri RequestUri { get; }
```
LTA-Thinking, line 26: Can the RequestUri be removed? It doesn't seem to be used.

src/Microsoft.Health.Fhir.Api/Features/Operations/Import/InitialImportLockMiddleware.cs
```diff
@@ -0,0 +1,80 @@
...
+            {
+                (HttpMethods.Get, ".*"), // Exclude all read operations
+                (HttpMethods.Post, "/\\$import"),
+                (HttpMethods.Delete, "/_operations/import/.+"),
+            };
+        }
+
+        public async Task Invoke(HttpContext context)
+        {
+            if (!_importTaskConfiguration.Enabled || !_importTaskConfiguration.InitalImportMode)
```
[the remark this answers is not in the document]
tongwu-msft: Yes, all non-excluded request would fail if customer enable import and explicit set initial import mode. We want to make sure customer do not have any write operation for the initial import mode.

samples/templates/default-azuredeploy-docker.json
```diff
@@ -228,11 +243,15 @@
             "SqlServer__Initialize": "[equals(parameters('solutionType'),'FhirServerSqlServer')]",
             "SqlServer__SchemaOptions__AutomaticUpdatesEnabled": "[if(equals(parameters('sqlSchemaAutomaticUpdatesEnabled'),'auto'), true(), false())]",
             "DataStore": "[if(equals(parameters('solutionType'),'FhirServerCosmosDB'), 'CosmosDb', 'SqlServer')]",
+            "TaskHosting__Enabled": "[if(equals(parameters('solutionType'),'FhirServerCosmosDB'), false(), parameters('enableImport'))]",
+            "TaskHosting__MaxRunningTaskCount": "[parameters('backgroundTaskCount')]",
+            "FhirServer__Operations__IntegrationDataStore__StorageAccountUri": "[if(parameters('enableImport'), concat('https://', variables('storageAccountName'), variables('blobStorageUri')), 'null')]",
```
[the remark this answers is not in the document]
tongwu-msft: It is not necessary, we have this config for further improvement on different runtime environment. Also remove from all templates.

src/Microsoft.Health.Fhir.R4.Web/Properties/launchSettings.json
```diff
@@ -27,6 +27,10 @@
                 "TestAuthEnvironment:FilePath": "..//..//testauthenvironment.json",
                 "SqlServer:ConnectionString": "server=(local);Initial Catalog=FHIR_R4;Integrated Security=true",
                 "DataStore": "SqlServer",
+                "TaskHosting:Enabled": "true",
+                "TaskHosting:MaxRunningTaskCount": "2",
```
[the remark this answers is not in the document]
tongwu-msft: Yes, to accelerate the test, we allow test server to run multiple tasks on single node.

src/Microsoft.Health.Fhir.Shared.Web/appsettings.json
```diff
@@ -83,6 +83,9 @@
             },
             "Validate": {
                 "CacheDurationInSeconds": 14400
+            },
+            "Import": {
+                "Enabled": false
```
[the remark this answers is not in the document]
tongwu-msft: Agree, added some setting placeholders to the appsettings.

src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/DataGenerator/ResourceWriteClaimTableBulkCopyDataGenerator.cs
```diff
@@ -0,0 +1,63 @@
...
+            _generator = generator;
+        }
+
+        internal override string TableName => VLatest.ResourceWriteClaim.TableName;
+
+        internal override void FillDataTable(DataTable table, SqlBulkCopyDataWrapper input)
+        {
+            IEnumerable<BulkResourceWriteClaimTableTypeV1Row> claims = _generator.GenerateRows(new ResourceWrapper[] { input.Resource });
+
+            if (claims != null)
```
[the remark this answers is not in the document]
tongwu-msft: Previous we reference ResourceWriteClaimV1RowGenerator it might return null. Yes, currently it is not needed.

src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/DataGenerator/CompartmentAssignmentTableBulkCopyDataGenerator.cs
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        public CompartmentAssignmentTableBulkCopyDataGenerator(ITableValuedParameterRowGenerator<IReadOnlyList<ResourceWrapper>, BulkCompartmentAssignmentTableTypeV1Row> generator)
+        {
+            EnsureArg.IsNotNull(generator, nameof(generator));
+
+            _generator = generator;
+        }
+
+        internal override string TableName => VLatest.CompartmentAssignment.TableName;
```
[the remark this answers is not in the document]
tongwu-msft: Yes, we can keep them in same convention. Thanks!

tongwu-msft:
> Some thoughts:
> 
> * How should we test this in CI? We currently don't reset our database between runs.
> * In PaaS we will need a control for users to manage whether import is enabled or not.

- Currently we import resource and clean it when import resource completed. And also check the id just generated.
- For PaaS, this change should not impact PaaS. And yes, we should have management operation for import mode change for PaaS solution.

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/GetBulkImportRequestHandler.cs
tongwu-msft: I design to allow only importer to check import status. Not sure you have any concern on that?

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/GetBulkImportRequestHandler.cs
tongwu-msft: Updated. Thanks.

src/Microsoft.Health.Fhir.Core/Messages/BulkImport/GetImportRequest.cs
tongwu-msft: yes, it is not used now.

src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/DataGenerator/TableBulkCopyDataGenerator.cs
```diff
@@ -0,0 +1,32 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Data;
+
+namespace Microsoft.Health.Fhir.SqlServer.Features.Operations.Import.DataGenerator
+{
+    internal abstract class TableBulkCopyDataGenerator<TDataType>
```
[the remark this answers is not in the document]
tongwu-msft: yes, now we do not need generic type here.

src/Microsoft.Health.Fhir.Core/Configs/ImportTaskConfiguration.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        /// <summary>
+        /// Determines whether bulk import is enabled or not.
+        /// </summary>
+        public bool Enabled { get; set; }
+
+        /// <summary>
+        /// Initial import mode
+        /// </summary>
+        public bool InitalImportMode { get; set; }
```
yankunhuang-pku: typo. There are 13 related codes.

tongwu-msft: Thanks, fixed.

src/Microsoft.Health.Fhir.Azure/IntegrationDataStore/AzureAccessTokenClientInitializerV2.cs
```diff
@@ -0,0 +1,75 @@
+﻿// -------------------------------------------------------------------------------------------------
```
LTA-Thinking, line 1: This file and the AzureConnectionStringClientInitializerV2 are identical to the versions used for Export except for the error types they throw. Would it be possible to make the existing versions compatible with both export and import?

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ImportController.cs
```diff
@@ -0,0 +1,205 @@
...
+                 HttpContext.RequestAborted);
+
+            var bulkImportResult = ImportResult.Accepted();
+            bulkImportResult.SetContentLocationHeader(_urlResolver, OperationsConstants.Import, response.TaskId);
+            return bulkImportResult;
+        }
+
+        [HttpDelete]
+        [Route(KnownRoutes.ImportJobLocation, Name = RouteNames.CancelImport)]
+        [ServiceFilter(typeof(ValidateImportRequestFilterAttribute))]
```
LTA-Thinking, line 121: The validator isn't needed on the Cancel or Get requests. None of the headers are relevant since they have no bodies and are synchronous.

src/Microsoft.Health.TaskManagement/IContextUpdater.cs
```diff
@@ -17,7 +18,7 @@ public interface IContextUpdater
         /// </summary>
         /// <param name="taskId">Id for the task</param>
```
LTA-Thinking, line 19: Parameter taskId no longer exists.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Import/StartupForImportBasicSearchTestProvider.cs
```diff
@@ -0,0 +1,26 @@
+﻿// -------------------------------------------------------------------------------------------------
```
LTA-Thinking, line 1: Is it possible for the test fixtures to use one common startup class instead of having 10 duplicate files?

src/Microsoft.Health.Fhir.Shared.Tests/Samples.cs
```diff
@@ -96,6 +97,11 @@ public static ResourceElement GetDefaultConvertDataParameter()
             return GetJsonSample("Parameter-Convert-Data");
         }
 
+        public static ImportRequest GetDefaultBulkImportRequest()
```
LTA-Thinking, line 100: It doesn't look like this sample is used in any tests. Can it be removed?

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportOrchestratorTask.cs
```diff
@@ -0,0 +1,505 @@
...
+    public class ImportOrchestratorTask : ITask
+    {
+        public const short ImportOrchestratorTaskId = 2;
+
+        private const int DefaultPollingFrequencyInSeconds = 3;
+        private const long DefaultResourceSizePerByte = 64;
+
+        private ImportOrchestratorTaskInputData _orchestratorInputData;
+        private RequestContextAccessor<IFhirRequestContext> _contextAccessor;
+        private ImportOrchestratorTaskContext _orchestratorTaskContext; // TODO : changed later
```
LTA-Thinking, line 32: What is this TODO for?

src/Microsoft.Health.Fhir.Azure/IntegrationDataStore/AzureAccessTokenClientInitializerV2.cs
tongwu-msft: The current AzureConnectionStringClientInitialize load settings from ExportJobConfiguration. To make sure not impact current features on PaaS, we would like to add new config and ready on PaaS first, then migrate to new configuration.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ImportController.cs
tongwu-msft: Thanks, removed.

src/Microsoft.Health.TaskManagement/IContextUpdater.cs
tongwu-msft: Thanks, removed.

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportOrchestratorTask.cs
tongwu-msft: It is not valid now. Updated.

src/Microsoft.Health.Fhir.Shared.Tests/Samples.cs
tongwu-msft: Yes, it is not used.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Import/StartupForImportBasicSearchTestProvider.cs
tongwu-msft: we do not need seperate db for each test, merged to single provider.

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportOrchestratorTask.cs
```diff
@@ -0,0 +1,505 @@
...
+                try
+                {
+                    await _fhirDataBulkImportOperation.DeleteDuplicatedResourcesAsync(CancellationToken.None);
+                    await _fhirDataBulkImportOperation.PostprocessAsync(CancellationToken.None);
+
+                    _logger.LogInformation("Postprocess Completed");
+                }
+                catch (Exception ex)
+                {
+                    _logger.LogInformation(ex, "Failed at postprocess step.");
```
LTA-Thinking, line 211: Should something be done with taskResultData in the event that one of the earlier steps hit an exception and then the post process steps hit an exception? Currently it looks like there is no way to tell if two exceptions were hit.

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportOrchestratorTask.cs
```diff
@@ -0,0 +1,505 @@
...
+            long beginSequenceId = _sequenceIdGenerator.GetCurrentSequenceId();
+
+            foreach (var input in _orchestratorInputData.Input)
+            {
+                string taskId = Guid.NewGuid().ToString("N");
+
+                Dictionary<string, object> properties = await _integrationDataStoreClient.GetPropertiesAsync(input.Url, cancellationToken);
+                long blobSizeInBytes = (long)properties[IntegrationDataStoreClientConstants.BlobPropertyLength];
+                long estimatedResourceNumber = CalculateResourceNumberByResourceSize(blobSizeInBytes, DefaultResourceSizePerByte);
+                long endSequenceId = beginSequenceId + estimatedResourceNumber;
```
LTA-Thinking, line 275: What are the begin and end sequence id used for? It is just to give the user an estimate of how many resources were ingested?

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportOrchestratorTask.cs
```diff
@@ -0,0 +1,505 @@
...
+                {
+                    _logger.LogWarning(ex, "failed to cancel task {0}", taskInfo.TaskId);
+                }
+            }
+
+            // Wait task cancel for WaitRunningTaskCancelTimeoutInSec
+            await WaitRunningTaskCompleteAsync(runningTaskIds);
+        }
+
+        private async Task WaitRunningTaskCompleteAsync(List<string> runningTaskIds)
```
LTA-Thinking, line 459: This method looks like it could get into an infinite loop if it can't get info for a task.

cunninghamjc:
> CREATE TABLE dbo.SchemaMigrationProgress

How is this table used?  Should it have a clustered index and primary key?  I don't see where LogSchemaMigrationProgress is called.

---
Refers to: src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.sql:17 in 8c01a24. [](commit_id = 8c01a24c0c951f51dcfc7d9b87c7ec276e1b38fb, deletion_comment = False)

cunninghamjc:
>     Timestamp datetime2(3) default CURRENT_TIMESTAMP,

Why only (3)?

---
Refers to: src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.sql:19 in 8c01a24. [](commit_id = 8c01a24c0c951f51dcfc7d9b87c7ec276e1b38fb, deletion_comment = False)

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportProcessingTask.cs
```diff
@@ -0,0 +1,274 @@
...
+                    throw;
+                }
+                catch (OperationCanceledException)
+                {
+                    throw;
+                }
+                catch (Exception ex)
+                {
+                    _logger.LogInformation(ex, "Failed to import data.");
+                    throw new RetriableTaskException("Failed to import data.", ex);
```
LTA-Thinking, line 168: I'm nervous about an unknown error being labeled retriable. This could put the system in a loop if it is retrying an import that can reproduce this exception.

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportProcessingTask.cs
```diff
@@ -0,0 +1,274 @@
...
+                // Update progress for checkpoints
+                await foreach (ImportProcessingProgress progress in progressChannel.Reader.ReadAllAsync())
+                {
+                    if (cancellationToken.IsCancellationRequested)
+                    {
+                        throw new OperationCanceledException("Import task is canceled by user.");
+                    }
+
+                    _importProgress.SucceedImportCount = progress.SucceedImportCount + succeedImportCount;
+                    _importProgress.FailedImportCount = progress.FailedImportCount + failedImportCount;
```
LTA-Thinking, line 133: This looks like it will not record all the success or failures if there are multiple progresses. The succeedImportCount and failedImportCount are not updated inside the foreach loop.

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportOrchestratorTask.cs
LTA-Thinking: It looks like this is the block of Ids assigned to this input file. So the expectation is that DefaultResourceSizePerByte is smaller than the average resource size in the file?

LTA-Thinking:
Review progress: I've now looked at all the non-Test and non-SQL specific files.

src/Microsoft.Health.Fhir.Azure.UnitTests/IntegrationDataStore/AzureBlobIntegrationDataStoreClientTests.cs
```diff
@@ -0,0 +1,334 @@
...
+using Microsoft.Health.Fhir.Azure.IntegrationDataStore;
+using Microsoft.Health.Fhir.Core.Configs;
+using Microsoft.Health.Fhir.Core.Features.Operations;
+using Xunit;
+
+namespace Microsoft.Health.Fhir.Azure.UnitTests.IntegrationDataStore
+{
+    public class AzureBlobIntegrationDataStoreClientTests
+    {
+        [Fact(Skip = "Local tests need emulator.")]
```
LTA-Thinking, line 24: What is the use case for these? It looks like they all need the emulator to run.

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Operations/Import/ImportRequestExtensionsTests.cs
```diff
@@ -0,0 +1,44 @@
...
+using Microsoft.Health.Fhir.Api.Features.Operations.Import;
+using Microsoft.Health.Fhir.Core.Features.Operations.Import.Models;
+using Xunit;
+
+namespace Microsoft.Health.Fhir.Shared.Api.UnitTests.Features.Operations.Import
+{
+    public class ImportRequestExtensionsTests
+    {
+        [Fact]
+        public void GivenImportReuqestInParamtersFormat_WhenConvert_ThenImportRequestShouldBeReturned()
```
LTA-Thinking, line 17: Nit: typo 'Reuqest'. It is also in the other test name.

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Operations/Import/ImportRequestExtensionsTests.cs
```diff
@@ -0,0 +1,44 @@
...
+        {
+            ImportRequest input = new ImportRequest();
+            input.InputFormat = "test";
+            input.InputSource = new Uri("http://dummy");
+            input.Input = new List<InputResource>() { new InputResource() { Etag = "etag", Type = "type", Url = new Uri("http://dummy/resource") } };
+            input.StorageDetail = new ImportRequestStorageDetail() { Type = "blob" };
+
```
LTA-Thinking, line 24: Could you add Mode and Force to check they are getting converted properly?

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Operations/Import/ImportErrorSerializerTests.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        private readonly FhirJsonSerializer _jsonSerializer = new FhirJsonSerializer();
+
+        [Fact]
+        public void GivenImportProcessError_WhenSerialize_ValidStringShouldBeReturned()
+        {
+            string errorMessage = "Test Error";
+            ImportErrorSerializer serializer = new ImportErrorSerializer(_jsonSerializer);
+
+            string outcome = serializer.Serialize(10, new Exception("Test Error"));
```
LTA-Thinking, line 24: Nit: can you use the errorMessage variable to initialize the Exception.

cunninghamjc:
> CREATE TABLE dbo.SchemaMigrationProgress

Ignore -- I see this is unrelated to this PR.

---
In reply to: [863540619](https://github.com/microsoft/fhir-server/pull/1992#issuecomment-863540619) [](http://example.com/codeflow?ancestors=863540619)

---
Refers to: src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.sql:17 in 8c01a24. [](commit_id = 8c01a24c0c951f51dcfc7d9b87c7ec276e1b38fb, deletion_comment = False)

cunninghamjc:
>     Timestamp datetime2(3) default CURRENT_TIMESTAMP,

Ignore -- I see this is unrelated to this PR.

---
In reply to: [863540889](https://github.com/microsoft/fhir-server/pull/1992#issuecomment-863540889) [](http://example.com/codeflow?ancestors=863540889)

---
Refers to: src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.sql:19 in 8c01a24. [](commit_id = 8c01a24c0c951f51dcfc7d9b87c7ec276e1b38fb, deletion_comment = False)

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.diff.sql
```diff
@@ -0,0 +1,424 @@
...
+    SET XACT_ABORT ON
+
+    SET TRANSACTION ISOLATION LEVEL SERIALIZABLE
+    BEGIN TRANSACTION
+
+    DECLARE @Sql NVARCHAR(MAX);
+    DECLARE @ParmDefinition NVARCHAR(512);
+
+    SET @sql = N'DELETE TOP(@BatchSizeParam) FROM ' + @tableName	+ N' WHERE ResourceTypeId = @ResourceTypeIdParam AND ResourceSurrogateId >= @StartResourceSurrogateIdParam AND ResourceSurrogateId < @EndResourceSurrogateIdParam'
+    SET @parmDefinition = N'@BatchSizeParam int, @ResourceTypeIdParam smallint, @StartResourceSurrogateIdParam bigint, @EndResourceSurrogateIdParam bigint';
```
cunninghamjc, line 123: This dynamic SQL is pretty risky -- is there an alternative that would avoid SQL injection risks?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.diff.sql
```diff
@@ -0,0 +1,424 @@
...
+CREATE OR ALTER PROCEDURE [dbo].[DisableIndexes]
+    @indexes dbo.IndexTableType_1 READONLY
+AS
+    SET NOCOUNT ON
+    SET XACT_ABORT ON
+
+    SET TRANSACTION ISOLATION LEVEL SERIALIZABLE
+    BEGIN TRANSACTION
+
+    declare commands cursor for
```
cunninghamjc, line 158: Consider declaring the cursor as FAST_FORWARD.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.diff.sql
```diff
@@ -0,0 +1,424 @@
...
+    SELECT N'ALTER INDEX [' + indexes.IndexName + '] ON ' + indexes.TableName + ' Disable;'
```
cunninghamjc, line 159: >[' + indexes.IndexName + '] [](http://example.com/codeflow?start=25&length=27)

Dynamic SQL based on user-provided input is risky.  Any alternatives?  Separately, consider using QUOTENAME() instead of hard-coding the brackets (what if indexes.IndexName contains ']'?).

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportOrchestratorTask.cs
tongwu-msft: Put inner error for multiple error scenario.

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportOrchestratorTask.cs
tongwu-msft: For our scenario of SQL, it is ResourceSurrogateId. 

For DefaultResourceSizePerByte, it is much small than resource private const long DefaultResourceSizePerByte = 64;

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportOrchestratorTask.cs
tongwu-msft: the task should be created first then added to running task queue, so it should not happen. Make it clear looks like also remove task id could be more readable.

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportProcessingTask.cs
tongwu-msft: So far we thought this should not happen, let me use log error for this to track if there's any unexpected error, max retry count would help to prevent it fail into infinite loop.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.diff.sql
```diff
@@ -0,0 +1,424 @@
...
+--         * The maximum number for retry operation
+--     @inputData
+--         * Input data payload for the task
+--     @isUniqueTaskByType
+--         * Only create task if there's no other active task with same task type id
+--
+CREATE OR ALTER PROCEDURE [dbo].[CreateTask_2]
+    @taskId varchar(64),
+    @queueId varchar(64),
+	@taskTypeId smallint,
```
cunninghamjc, line 311: Nit: this file has a mix of spaces and tabs -- could we please convert all tabs to spaces?

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportProcessingTask.cs
tongwu-msft: for multiple progress, we have initial import count at beginning from context. So the result would resume from last checkpoint.
 long succeedImportCount = _importProgress.SucceedImportCount;
 long failedImportCount = _importProgress.FailedImportCount;

src/Microsoft.Health.Fhir.Azure.UnitTests/IntegrationDataStore/AzureBlobIntegrationDataStoreClientTests.cs
tongwu-msft: Currently integration test do not support deploy storage. So I created these tests for local check.

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Operations/Import/ImportRequestExtensionsTests.cs
tongwu-msft: thanks, updated.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.diff.sql
tongwu-msft: Sure, update. Thanks.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.diff.sql
tongwu-msft: Table name and index name can not  be used as parameters in dynamic sql. For table name looks like we can use OBJECT_NAME(OBJECT_ID(@tableName)) to prevent SQL injection risks.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.diff.sql
tongwu-msft: Thanks QUOTENAME works here. And for dynamic index name here, it looks like can not verify as tablename...

Since it is only used by our code and indexes names are predefined in code, I think it should be safe for no direct customer use.

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Import/ImportProcessingTaskTests.cs
```diff
@@ -0,0 +1,398 @@
...
+                        finally
+                        {
+                            resourceChannel.Writer.Complete();
+                        }
+                    });
+
+                    return (resourceChannel, loadTask);
+                });
+
+            importer.Import(Arg.Any<Channel<ImportResource>>(), Arg.Any<IImportErrorStore>(), Arg.Any<CancellationToken>())
```
LTA-Thinking, line 159: Nit: the import and updateContextAsync method mocks aren't needed in this test since the exception is thrown before they are called.

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Controllers/ImportControllerTests.cs
```diff
@@ -0,0 +1,210 @@
...
+using Microsoft.Health.Fhir.Core.Messages.Import;
+using NSubstitute;
+using Xunit;
+using Task = System.Threading.Tasks.Task;
+
+namespace Microsoft.Health.Fhir.Api.UnitTests.Controllers
+{
+    public class ImportControllerTests
+    {
+        private ImportController _bulkImportEnabledController;
```
LTA-Thinking, line 30: Nit: this variable isn't used and can be removed.

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportProcessingTask.cs
LTA-Thinking: I think adding a max retry count would be a good idea. I don't think it needs to be part of the first pass due to the other issues we are working through right now, but having a work item to add it would be good.

samples/templates/default-azuredeploy.json
```diff
@@ -198,6 +198,20 @@
...
+            "defaultValue": false,
+            "metadata": {
+                "description": "Determines whether the $import operation will be enabled for this fhir instance."
+            }
+        },
+        "backgroundTaskCount": {
+            "type": "int",
+            "defaultValue": 1,
+            "metadata": {
+                "description": "Support parallel backgound task running"
```
yankunhuang-pku, line 213: typo:  should be "background"

samples/templates/default-azuredeploy-docker.json
```diff
@@ -198,6 +198,20 @@
...
+            "defaultValue": false,
+            "metadata": {
+                "description": "Determines whether the $import operation will be enabled for this fhir instance."
+            }
+        },
+        "backgroundTaskCount": {
+            "type": "int",
+            "defaultValue": 1,
+            "metadata": {
+                "description": "Support parallel backgound task running"
```
yankunhuang-pku, line 213: same typo:  should be "background"

src/Microsoft.Health.Fhir.Api/Features/BackgroundTaskService/TaskFactory.cs
```diff
@@ -0,0 +1,115 @@
...
+                    _taskmanager,
+                    _sequenceIdGenerator,
+                    contextUpdater,
+                    _contextAccessor,
+                    _fhirDataBulkImportOperation,
+                    _integrationDataStoreClient,
+                    _loggerFactory);
+            }
+
+            _logger.LogWarning("Can not create task for type {taskType}", taskInfo.TaskTypeId);
```
yankunhuang-pku, line 111: Redundant logger? Is it handled in line 107 of TaskHosting

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportProcessingTask.cs
tongwu-msft: Yes, currently we have max retry at the task level. For details: 

                TaskInfo processingTask = new TaskInfo()
                {
                    QueueId = _orchestratorInputData.ProcessingTaskQueueId,
                    TaskId = taskId,
                    TaskTypeId = ImportProcessingTask.ImportProcessingTaskId,
                    InputData = JsonConvert.SerializeObject(importTaskPayload),
                    MaxRetryCount = _orchestratorInputData.ProcessingTaskMaxRetryCount,
                };

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportOrchestratorTask.cs
```diff
@@ -0,0 +1,507 @@
...
+        {
+            List<string> runningTaskIds = new List<string>();
+
+            if ((_orchestratorTaskContext?.DataProcessingTasks?.Count ?? 0) == 0)
+            {
+                // No data processing task created.
+                return;
+            }
+
+            foreach (TaskInfo taskInfo in _orchestratorTaskContext.DataProcessingTasks.Values)
```
yankunhuang-pku, line 439: Is it safer to check if it is null? _orchestratorTaskContext.DataProcessingTasks.Values

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportOrchestratorTask.cs
```diff
@@ -0,0 +1,507 @@
...
+            {
+                taskResultData = new TaskResultData(TaskResult.Success, JsonConvert.SerializeObject(_orchestratorTaskContext.ImportResult));
+            }
+
+            return taskResultData;
+        }
+
+        public void Cancel()
+        {
+            if (!_cancellationTokenSource.IsCancellationRequested)
```
yankunhuang-pku, line 233: Is it safer to check if it is null? _cancellationTokenSource

src/Microsoft.Health.Fhir.Azure/IntegrationDataStore/AzureAccessTokenClientInitializerV2.cs
```diff
@@ -0,0 +1,75 @@
...
+            IAccessTokenProvider accessTokenProvider,
+            IOptions<IntegrationDataStoreConfiguration> integrationDataStoreConfiguration,
+            ILogger<AzureAccessTokenClientInitializerV2> logger)
+        {
+            EnsureArg.IsNotNull(accessTokenProvider, nameof(accessTokenProvider));
+            EnsureArg.IsNotNull(integrationDataStoreConfiguration?.Value, nameof(integrationDataStoreConfiguration));
+            EnsureArg.IsNotNull(logger, nameof(logger));
+
+            _accessTokenProvider = accessTokenProvider;
+            _integrationDataStoreConfiguration = integrationDataStoreConfiguration.Value;
```
yankunhuang-pku, line 36: Is it safer to check if it is null? integrationDataStoreConfiguration?.Value

src/Microsoft.Health.Fhir.Api/Features/BackgroundTaskService/TaskHostingBackgroundService.cs
```diff
@@ -0,0 +1,73 @@
...
+            Func<IScoped<TaskHosting>> taskHostingFactory,
+            IOptions<TaskHostingConfiguration> taskHostingConfiguration,
+            ILogger<TaskHostingBackgroundService> logger)
+        {
+            EnsureArg.IsNotNull(taskHostingFactory, nameof(taskHostingFactory));
+            EnsureArg.IsNotNull(taskHostingConfiguration?.Value, nameof(taskHostingConfiguration));
+            EnsureArg.IsNotNull(logger, nameof(logger));
+
+            _taskHostingFactory = taskHostingFactory;
+            _taskHostingConfiguration = taskHostingConfiguration.Value;
```
yankunhuang-pku, line 38: same~ Is it safer to check if it is null? taskHostingConfiguration?.Value

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ImportController.cs
```diff
@@ -0,0 +1,203 @@
...
+            EnsureArg.IsNotNull(operationsConfig, nameof(operationsConfig));
+            EnsureArg.IsNotNull(urlResolver, nameof(urlResolver));
+            EnsureArg.IsNotNull(features?.Value, nameof(features));
+            EnsureArg.IsNotNull(mediator, nameof(mediator));
+            EnsureArg.IsNotNull(logger, nameof(logger));
+
+            _fhirRequestContextAccessor = fhirRequestContextAccessor;
+            _importConfig = operationsConfig.Value.Import;
+            _urlResolver = urlResolver;
+            _features = features.Value;
```
yankunhuang-pku, line 80: same~ features?.Value

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ImportController.cs
```diff
@@ -0,0 +1,203 @@
...
+
+        [HttpDelete]
+        [Route(KnownRoutes.ImportJobLocation, Name = RouteNames.CancelImport)]
+        [AuditEventType(AuditEventSubType.Import)]
+        public async Task<IActionResult> CancelImport(string idParameter)
+        {
+            CancelImportResponse response = await _mediator.CancelImportAsync(idParameter, HttpContext.RequestAborted);
+
+            _logger.LogInformation($"CancelImport {response?.StatusCode}");
+            return new ImportResult(response.StatusCode);
```
yankunhuang-pku, line 127: same~ response?.StatusCode

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportProcessingTask.cs
```diff
@@ -0,0 +1,274 @@
...
+                if (!_cancellationTokenSource.IsCancellationRequested)
+                {
+                    _cancellationTokenSource.Cancel();
+                }
+            }
+        }
+
+        public void Cancel()
+        {
+            if (!_cancellationTokenSource.IsCancellationRequested)
```
yankunhuang-pku, line 222: same~ !_cancellationTokenSource?.IsCancellationRequested

src/Microsoft.Health.Fhir.Azure/IntegrationDataStore/AzureConnectionStringClientInitializerV2.cs
```diff
@@ -0,0 +1,66 @@
...
+    {
+        private readonly IntegrationDataStoreConfiguration _integrationDataStoreConfiguration;
+        private readonly ILogger<AzureConnectionStringClientInitializerV2> _logger;
+
+        public AzureConnectionStringClientInitializerV2(IOptions<IntegrationDataStoreConfiguration> integrationDataStoreConfiguration, ILogger<AzureConnectionStringClientInitializerV2> logger)
+        {
+            EnsureArg.IsNotNull(integrationDataStoreConfiguration?.Value, nameof(integrationDataStoreConfiguration));
+            EnsureArg.IsNotNull(logger, nameof(logger));
+
+            _integrationDataStoreConfiguration = integrationDataStoreConfiguration.Value;
```
yankunhuang-pku, line 30: same~ integrationDataStoreConfiguration?.Value;

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportOrchestratorTask.cs
tongwu-msft: We check it before using it.

            if ((_orchestratorTaskContext?.DataProcessingTasks?.Count ?? 0) == 0)
            {
                // No data processing task created.
                return;
            }

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportOrchestratorTask.cs
tongwu-msft: Yes, we need to check.

src/Microsoft.Health.Fhir.Azure/IntegrationDataStore/AzureAccessTokenClientInitializerV2.cs
tongwu-msft: We check before using it here:
EnsureArg.IsNotNull(integrationDataStoreConfiguration?.Value, nameof(integrationDataStoreConfiguration));

src/Microsoft.Health.Fhir.Api/Features/BackgroundTaskService/TaskHostingBackgroundService.cs
tongwu-msft: Same here: EnsureArg.IsNotNull(taskHostingConfiguration?.Value, nameof(taskHostingConfiguration));

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ImportController.cs
tongwu-msft: Yes, we should check here.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ImportController.cs
tongwu-msft: yes, we can add check here.

tongwu-msft: The response should not be null, otherwise looks like deserve 500.

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportProcessingTask.cs
tongwu-msft: yes, we should check this.

src/Microsoft.Health.Fhir.Azure/IntegrationDataStore/AzureConnectionStringClientInitializerV2.cs
tongwu-msft: We have precheck here:
EnsureArg.IsNotNull(integrationDataStoreConfiguration?.Value, nameof(integrationDataStoreConfiguration));

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportOrchestratorTask.cs
```diff
@@ -0,0 +1,507 @@
...
+                errorResult.HttpStatusCode = HttpStatusCode.BadRequest;
+                errorResult.ErrorMessage = processingEx.Message;
+
+                taskResultData = new TaskResultData(TaskResult.Fail, JsonConvert.SerializeObject(errorResult));
+            }
+            catch (Exception ex)
+            {
+                _logger.LogInformation(ex, "Failed to import data.");
+
+                errorResult = new ImportTaskErrorResult();
```
yankunhuang-pku, line 193: We can use new method in another concise way, but it's not necessary
```
errorResult = new ImportTaskErrorResult
{
    HttpStatusCode = HttpStatusCode.InternalServerError,
    ErrorMessage = ex.Message,
};
```

src/Microsoft.Health.Fhir.Azure/IntegrationDataStore/AzureAccessTokenClientInitializerV2.cs
```diff
@@ -0,0 +1,75 @@
...
+            {
+                throw new IntegrationDataStoreClientInitializerException(Resources.InvalidStorageUri, HttpStatusCode.BadRequest);
+            }
+
+            if (!Uri.TryCreate(integrationDataStoreConfiguration.StorageAccountUri, UriKind.Absolute, out Uri storageAccountUri))
+            {
+                throw new IntegrationDataStoreClientInitializerException(Resources.InvalidStorageUri, HttpStatusCode.BadRequest);
+            }
+
+            string accessToken = null;
```
yankunhuang-pku, line 57: Unnecessary assignment. We may use ```string accessToken```; because we assign a value to it after that.

src/Microsoft.Health.Fhir.Azure/IntegrationDataStore/AzureConnectionStringClientInitializerV2.cs
```diff
@@ -0,0 +1,66 @@
...
+            {
+                throw new IntegrationDataStoreClientInitializerException(Resources.InvalidConnectionSettings, HttpStatusCode.BadRequest);
+            }
+
+            if (!CloudStorageAccount.TryParse(integrationDataStoreConfiguration.StorageAccountConnection, out CloudStorageAccount cloudAccount))
+            {
+                throw new IntegrationDataStoreClientInitializerException(Resources.InvalidConnectionSettings, HttpStatusCode.BadRequest);
+            }
+
+            CloudBlobClient blobClient = null;
```
yankunhuang-pku, line 51: Same~unnecessary assignment.

src/Microsoft.Health.Fhir.Azure/IntegrationDataStore/AzureAccessTokenClientInitializerV2.cs
tongwu-msft: Thanks, updated.

src/Microsoft.Health.Fhir.Azure/IntegrationDataStore/AzureConnectionStringClientInitializerV2.cs
tongwu-msft: Thanks, updated.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
```diff
@@ -0,0 +1,424 @@
...
+    @indexes dbo.IndexTableType_1 READONLY
+AS
+    SET NOCOUNT ON
+    SET XACT_ABORT ON
+
+    SET TRANSACTION ISOLATION LEVEL SERIALIZABLE
+    BEGIN TRANSACTION
+
+    DECLARE commands CURSOR FAST_FORWARD
+    FOR SELECT N'ALTER INDEX ' + QUOTENAME(indexes.IndexName) + ' ON ' + OBJECT_NAME(OBJECT_ID(indexes.TableName)) + ' Disable;'
```
cunninghamjc, line 159: Small nit: OBJECT_NAME() doesn't return the schema, so if we were to have multiple schemas with tables with the same name, this might be ambiguous.  Not a blocker now but something to consider for future code maintainability.

tongwu-msft: Thanks for the information, we change to check tablename before use. Then we can directly use @table without object_name. Thanks!

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.diff.sql
tongwu-msft: Thanks, now we check index and table name before use to prevent inject attack.

samples/templates/default-azuredeploy.json
```diff
@@ -198,6 +198,20 @@
...
+            "defaultValue": false,
+            "metadata": {
+                "description": "Determines whether the $import operation will be enabled for this fhir instance."
+            }
+        },
+        "backgroundTaskCount": {
+            "type": "int",
+            "defaultValue": 1,
+            "metadata": {
+                "description": "Support parallel background task running"
```
moria97, line 213: nit: Supports parallel background task running.

src/Microsoft.Health.Fhir.Api/Features/Filters/ValidateImportRequestFilterAttribute.cs
```diff
@@ -0,0 +1,51 @@
...
+                preferHeaderValue.Count != 1 ||
+                !string.Equals(preferHeaderValue[0], PreferHeaderExpectedValue, StringComparison.OrdinalIgnoreCase))
+            {
+                throw new RequestNotValidException(string.Format(Resources.UnsupportedHeaderValue, PreferHeaderName));
+            }
+
+            if (string.Equals(context.HttpContext.Request.Method, "POST", StringComparison.OrdinalIgnoreCase))
+            {
+                if (!context.HttpContext.Request.Headers.TryGetValue(HeaderNames.ContentType, out var contentTypeHeaderValue) ||
+                    preferHeaderValue.Count != 1 ||
```
moria97, line 43: Should be contentTypeHeaderValue.Count != 1?

src/Microsoft.Health.Fhir.Azure/FhirServerBuilderAzureRegistrationExtensions.cs
```diff
@@ -77,5 +79,36 @@ public static IFhirServerBuilder AddContainerRegistryTokenProvider(this IFhirSer
 
             return fhirServerBuilder;
         }
+
+        public static IFhirServerBuilder AddAzureIntegrationDataStoreClient(this IFhirServerBuilder fhirServerBuilder, IConfiguration configuration)
```
brendankowitz, line 83: >public static IFhirServerBuilder AddAzureIntegrationDataStoreClient(this IFhirServerBuilder fhirServerBuilder, IConfiguration configuration) [](http://example.com/codeflow?start=8&length=140)

Can we add a comment for this to make it clearer. Looks like its adding Blob storage clients?

src/Microsoft.Health.Fhir.Azure/IntegrationDataStore/AzureBlobIntegrationDataStoreClient.cs
```diff
@@ -0,0 +1,256 @@
...
+        {
+            EnsureArg.IsNotNull(resourceUri, nameof(resourceUri));
+
+            try
+            {
+                return await Policy.Handle<StorageException>()
+                    .WaitAndRetryAsync(
+                        retryCount: 3,
+                        sleepDurationProvider: (retryCount) => TimeSpan.FromSeconds(5 * (retryCount - 1)))
+                    .ExecuteAsync(async () =>
```
brendankowitz, line 167: This looks repeated, perhaps move the policy setup to a function. (e.g. [https://github.com/microsoft/fhir-server/blob/main/src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/RetryExceptionPolicyFactory.cs](https://github.com/microsoft/fhir-server/blob/main/src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/RetryExceptionPolicyFactory.cs))

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/CancelImportRequestHandler.cs
```diff
@@ -0,0 +1,65 @@
...
+            _taskManager = taskManager;
+            _authorizationService = authorizationService;
+            _logger = logger;
+        }
+
+        public async Task<CancelImportResponse> Handle(CancelImportRequest request, CancellationToken cancellationToken)
+        {
+            EnsureArg.IsNotNull(request, nameof(request));
+
+            if (await _authorizationService.CheckAccess(DataActions.Import, cancellationToken) != DataActions.Import)
```
brendankowitz, line 41: >DataActions.Import [](http://example.com/codeflow?start=56&length=18)

New DataActions will require an update in managed service

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportFileEtagNotMatchException.cs
```diff
@@ -0,0 +1,26 @@
...
+using EnsureThat;
+
+namespace Microsoft.Health.Fhir.Core.Features.Operations.Import
+{
+    public class ImportFileEtagNotMatchException : Exception
+    {
+        public ImportFileEtagNotMatchException(string message)
+            : base(message)
+        {
+            EnsureArg.IsNotNull(message, nameof(message));
```
brendankowitz, line 16: >EnsureArg.IsNotNull [](http://example.com/codeflow?start=12&length=19)

Exceptions should use Debug.Assert. Else we risk throwing another exception

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportResourceLoader.cs
```diff
@@ -0,0 +1,180 @@
...
+                    long id = idGenerator(index);
+
+                    try
+                    {
+                        ImportResource importResource = _importResourceParser.Parse(id, index, content);
+
+                        if (!string.IsNullOrEmpty(resourceType) && !resourceType.Equals(importResource.Resource?.ResourceTypeName, StringComparison.Ordinal))
+                        {
+                            throw new FormatException("Resource type not match.");
+                        }
```
brendankowitz, line 165: For my context, why do we need to do this? Are there optimizations applied based on resourceType?

tongwu-msft: We want to do batch import for same partition which now is for resource type. Customer provide same type resources can achieve better performance.

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportFileEtagNotMatchException.cs
tongwu-msft: Thanks, updated.

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/CancelImportRequestHandler.cs
tongwu-msft: Thanks for reminder, for paas integration we need to add this for managed service.

src/Microsoft.Health.Fhir.Azure/IntegrationDataStore/AzureBlobIntegrationDataStoreClient.cs
tongwu-msft: Thanks, updated.

src/Microsoft.Health.Fhir.Azure/FhirServerBuilderAzureRegistrationExtensions.cs
tongwu-msft: Sure, added. For more details, we would like to reuse storage as export, but now it is at export operation level. To keep backward, we added a new blob store at common level, after we enable this on Paas, we would like to unify the usage for integration store.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Import/ImportResourceParser.cs
```diff
@@ -0,0 +1,86 @@
...
+                }
+
+                if (reference.Reference.Contains("?", StringComparison.Ordinal))
+                {
+                    throw new NotSupportedException("Conditional reference not supported for initial import.");
+                }
+            }
+        }
+
+        private Stream GenerateCompressedRawResource(string rawResource)
```
brendankowitz, line 76: Can we use/refactor the existing CompressedRawResourceConverter to meet these needs? I think having this compression logic exist in two places will be harder to keep in sync if it ever changes

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Import/ImportResourceParser.cs
```diff
@@ -0,0 +1,86 @@
...
+                if (string.IsNullOrWhiteSpace(reference.Reference))
+                {
+                    continue;
+                }
+
+                if (reference.Reference.Contains("?", StringComparison.Ordinal))
+                {
+                    throw new NotSupportedException("Conditional reference not supported for initial import.");
+                }
+            }
```
brendankowitz, line 73: This seems like a fairly expensive check to do on every resource, assuming its following similar logic to Batch bundles? Are we ever planning to support resolving these conditional references?

src/Microsoft.Health.TaskManagement/IContextUpdaterFactory.cs
```diff
@@ -0,0 +1,12 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.TaskManagement
+{
+    public interface IContextUpdaterFactory
```
brendankowitz, line 8: >IContextUpdaterFactory [](http://example.com/codeflow?start=21&length=22)

Document interfaces

src/Microsoft.Health.Fhir.SqlServer.UnitTests/Features/Import/SqlServerBulkImportOperationTests.cs
```diff
@@ -0,0 +1,68 @@
...
+using Microsoft.Health.Fhir.SqlServer.Features.Storage;
+using Microsoft.Health.SqlServer.Features.Schema.Model;
+using Xunit;
+
+namespace Microsoft.Health.Fhir.SqlServer.UnitTests.Features.Import
+{
+    public class SqlServerBulkImportOperationTests
+    {
+        [Fact]
+        public void GivenResourceRelatedTables_WhenNewIndexesAdded_BulkImportOperationShouldSupportNewIndexes()
```
brendankowitz, line 18: >BulkImportOperationShouldSupportNewIndexes [](http://example.com/codeflow?start=67&length=42)

What about indexes added via Azure's Database Advisor

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerTaskContextUpdater.cs
```diff
@@ -0,0 +1,67 @@
...
+            EnsureArg.IsNotNull(sqlConnectionWrapperFactory, nameof(sqlConnectionWrapperFactory));
+            EnsureArg.IsNotNull(logger, nameof(logger));
+
+            _taskId = taskId;
+            _runId = runId;
+            _sqlConnectionWrapperFactory = sqlConnectionWrapperFactory;
+            _logger = logger;
+        }
+
+        public async Task UpdateContextAsync(string context, CancellationToken cancellationToken)
```
brendankowitz, line 43: >context [](http://example.com/codeflow?start=52&length=7)

Can context be Null?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerTaskContextUpdaterFactory.cs
```diff
@@ -0,0 +1,34 @@
...
+            ILoggerFactory loggerFactory)
+        {
+            EnsureArg.IsNotNull(sqlConnectionWrapperFactory, nameof(sqlConnectionWrapperFactory));
+            EnsureArg.IsNotNull(loggerFactory, nameof(loggerFactory));
+
+            _sqlConnectionWrapperFactory = sqlConnectionWrapperFactory;
+            _loggerFactory = loggerFactory;
+        }
+
+        public IContextUpdater CreateContextUpdater(string taskId, string runId)
```
brendankowitz, line 29: >string taskId, string runId [](http://example.com/codeflow?start=52&length=27)

Can these be null?

src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/DataGenerator/UriSearchParamsTableBulkCopyDataGenerator.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        internal override string TableName
+        {
+            get
+            {
+                return VLatest.UriSearchParam.TableName;
+            }
+        }
+
+        internal override void FillDataTable(DataTable table, SqlBulkCopyDataWrapper input)
```
brendankowitz, line 38: >table [](http://example.com/codeflow?start=55&length=5)

Can DataTable be null?

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Import/ImportResourceParser.cs
tongwu-msft: Update CompressedRawResourceConverter for resue between components.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Import/ImportResourceParser.cs
tongwu-msft: We can not simply resue similar logic for batch bundles to support conditional reference. The sequence for import resource is random, so we plan to do update reference after all import complete. It looks like is an expensive operation, and we do not plan to support it at first version.

src/Microsoft.Health.TaskManagement/IContextUpdaterFactory.cs
tongwu-msft: Sure, updated.

src/Microsoft.Health.Fhir.SqlServer.UnitTests/Features/Import/SqlServerBulkImportOperationTests.cs
tongwu-msft: Thanks for this comment, we do not aware this before. I think we can leave it for now since disable indexes now is not necessary, and for this test we want to block change for indexes and go through if we need to add the new indexes to this list. We would test this and maybe we can add document for this in advanced section of bulk import operation.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerTaskContextUpdater.cs
tongwu-msft: This is possible null.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerTaskContextUpdaterFactory.cs
tongwu-msft: Thanks, add null check here.

src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/DataGenerator/UriSearchParamsTableBulkCopyDataGenerator.cs
tongwu-msft: This should not be null, add check here for protect. Thanks.

src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/SqlImportOperation.cs
```diff
@@ -0,0 +1,388 @@
...
+                catch (SqlException sqlEx)
+                {
+                    _logger.LogError(sqlEx, "Failed to merge resources. " + sqlEx.Message);
+
+                    throw;
+                }
+            }
+        }
+
+        public async Task CleanBatchResourceAsync(string resourceType, long beginSequenceId, long endSequenceId, CancellationToken cancellationToken)
```
brendankowitz, line 184: I've done some successful imports, but also some problems if the task is killed before the import is complete. For example:

```
fail: Microsoft.Health.Fhir.SqlServer.Features.Storage.SqlServerFhirModel[0]
      The SqlServerFhirModel instance has not run the initialization required for the current schema version
info: Microsoft.Health.Fhir.SqlServer.Features.Operations.Import.SqlResourceBulkImporter[0]
      Failed to clean batch resource.
      Microsoft.Health.Abstractions.Exceptions.ServiceUnavailableException: The operation could not be completed, because the service was unable to accept new requests. It is safe to retry the operation. If the issue persists, please contact support.
         at Microsoft.Health.Fhir.SqlServer.Features.Storage.SqlServerFhirModel.ThrowIfNotInitialized() in C:\src\fhir-server\src\Microsoft.Health.Fhir.SqlServer\Features\Storage\SqlServerFhirModel.cs:line 462
         at Microsoft.Health.Fhir.SqlServer.Features.Storage.SqlServerFhirModel.GetResourceTypeId(String resourceTypeName) in C:\src\fhir-server\src\Microsoft.Health.Fhir.SqlServer\Features\Storage\SqlServerFhirModel.cs:line 98
         at Microsoft.Health.Fhir.SqlServer.Features.Storage.SqlImportOperation.CleanBatchResourceAsync(String resourceType, Int64 beginSequenceId, Int64 endSequenceId, CancellationToken cancellationToken) in C:\src\fhir-server\src\Microsoft.Health.Fhir.SqlServer
\Features\Operations\Import\SqlImportOperation.cs:line 186
         at Microsoft.Health.Fhir.SqlServer.Features.Operations.Import.SqlResourceBulkImporter.CleanResourceAsync(ImportProcessingTaskInputData inputData, ImportProcessingProgress progress, CancellationToken cancellationToken) in C:\src\fhir-server\src\Microsoft.
Health.Fhir.SqlServer\Features\Operations\Import\SqlResourceBulkImporter.cs:line 155.
```

tongwu-msft: Thanks for catching this, update with fix.

src/Microsoft.Health.Fhir.Azure/IntegrationDataStore/AzureAccessTokenClientInitializerV2.cs
```diff
@@ -0,0 +1,75 @@
...
+using Microsoft.Azure.Storage.Auth;
+using Microsoft.Azure.Storage.Blob;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Microsoft.Health.Fhir.Core.Configs;
+using Microsoft.Health.Fhir.Core.Features.Operations;
+
+namespace Microsoft.Health.Fhir.Azure.IntegrationDataStore
+{
+    public class AzureAccessTokenClientInitializerV2 : IIntegrationDataStoreClientInitilizer<CloudBlobClient>
```
brendankowitz, line 20: >AzureAccessTokenClientInitializerV2 [](http://example.com/codeflow?start=17&length=35)

What is the difference between AzureAccessTokenClientInitializer and AzureAccessTokenClientInitializerV2 these look pretty similar?

tongwu-msft: The only difference is AzureAccessTokenClientInitializer requires export configuration as input, we create v2 is a migration from export only to both export and import. To avoid breaking change, first we create V2 version, after new features ready in OSS and PaaS, then we can migrate to AzureAccessTokenClientInitializerV2.

brendankowitz (approved):

PR #1993, opened 2021-06-02
Create files for Payer Data Exchange

CaitlinV39:
## Description
This adds files for testing the Payer Data Exchange

## Related issues
N/A

## Testing
Touchstone

## FHIR Team Checklist

Skip

### DIFF
--- a/docs/rest/PayerDataExchange/PDex_Sample_Data.http
+++ b/docs/rest/PayerDataExchange/PDex_Sample_Data.http
@@ -1,4 +1,4 @@
-# This file includes all the search parameters, reindex operation, and profiles to load to get the necessary pieces for the Carin Blue Button IG
+# This file includes all the sample data to load for the PDex IG
 ### REST Client
 @fhirurl=<FHIR SERVER URL>
 @clientid=<Client Application ID>
--- a/docs/rest/PayerDataExchange/PDex_Sample_Data.http
+++ b/docs/rest/PayerDataExchange/PDex_Sample_Data.http
@@ -1,0 +1,284 @@
+# This file includes all the search parameters, reindex operation, and profiles to load to get the necessary pieces for the Carin Blue Button IG
+### REST Client
+@fhirurl=<FHIR SERVER URL>
+@clientid=<Client Application ID>
+@clientsecret=<Client Secret>
+@tenantid=<Tenant ID>
+@contentType = application/json
+
+### Get access token
+# @name getAADToken
+POST https://login.microsoftonline.com/{{tenantid}}/oauth2/token
+Content-Type: application/x-www-form-urlencoded
+
+grant_type=client_credentials
+&resource={{fhirurl}}
+&client_id={{clientid}}
+&client_secret={{clientsecret}}
+
+### Capture access token from getToken request
+@token={{getAADToken.response.body.access_token}}
+
+### Create a Patient
+# @name Patient1
+POST {{fhirurl}}/Patient
+Authorization: Bearer {{token}}
+content-type: {{contentType}}
+
+{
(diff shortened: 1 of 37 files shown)


LTA-Thinking:
Would it be better to fetch the US core definitions instead of saving copies of them to our repo? That way we don't need to worry if they update the definitions. Something like this could do it:
```
### Get the definition
# @name allergybody
GET http://hl7.org/fhir/us/core/StructureDefinition-us-core-allergyintolerance.json

### Save the definition
POST {{baseUrl}}/StructureDefinition
content-type: {{contentType}}

{{allergybody.response.body.*}}
```

docs/rest/PayerDataExchange/PDex_Sample_Data.http
```diff
@@ -0,0 +1,284 @@
+# This file includes all the search parameters, reindex operation, and profiles to load to get the necessary pieces for the Carin Blue Button IG
```
moiradillon12, line 1: Update to DaVinci Payer Data Exchange

docs/rest/PayerDataExchange/USCore.http
```diff
@@ -0,0 +1,265 @@
...
+### REST Client
+@fhirurl= <FHIR URL>
+@clientid= <CLIENT ID>
+@clientsecret= <CLIENT SECRET>
+@tenantid= <SUBSCRIPTION ID>
+@contentType = application/json
+
+### Get Metadata - Allows you to get the current capability statement
+# this will update as you add profiles to the database
+GET https://caitlinjune.azurewebsites.net/Patient/57dfd262-a0b7-4901-84d2-4909195d9022/$everything
```
moiradillon12, line 11: Should this be updated to not include your URL?

moiradillon12 (commented):
I tested all of the http files and was able to load required profiles, search parameters and data. Didn't test via Touchstone (limited number of tests available)

moiradillon12:
> Would it be better to fetch the US core definitions instead of saving copies of them to our repo? That way we don't need to worry if they update the definitions. Something like this could do it:
> 
> ```
> ### Get the definition
> # @name allergybody
> GET http://hl7.org/fhir/us/core/StructureDefinition-us-core-allergyintolerance.json
> 
> ### Save the definition
> POST {{baseUrl}}/StructureDefinition
> content-type: {{contentType}}
> 
> {{allergybody.response.body.*}}
> ```

CaitlinV39:
> Would it be better to fetch the US core definitions instead of saving copies of them to our repo? That way we don't need to worry if they update the definitions. Something like this could do it:
> 
> ```
> ### Get the definition
> # @name allergybody
> GET http://hl7.org/fhir/us/core/StructureDefinition-us-core-allergyintolerance.json
> 
> ### Save the definition
> POST {{baseUrl}}/StructureDefinition
> content-type: {{contentType}}
> 
> {{allergybody.response.body.*}}
> ```

@LTA-Thinking I think this would be amazing but it doesn't consistently work because sometimes there are issues in the profiles in the text section where the json isn't valid. I can spend more time opening JIRA issues to get those resolved and then we could move the whole file into this way.

LTA-Thinking:
> @LTA-Thinking I think this would be amazing but it doesn't consistently work because sometimes there are issues in the profiles in the text section where the json isn't valid. I can spend more time opening JIRA issues to get those resolved and then we could move the whole file into this way.

:(
Ah, ok. That is a shame. Best to keep our own records then if the official ones aren't reliable.

CaitlinV39:
@LTA-Thinking Moira can't approve this yet since shes not in the FHIR Server team. If you get a chance could you sign off?

LTA-Thinking (approved):

Issue #1989, opened 2021-06-01
Capability statement missing support for Patient $Everything


CaitlinV39:
**Describe the bug**
Patient/$everything is now supported on cosmos DB but the capability statement doesn't state that it is supported. 

**FHIR Version?**
Stu3/R4/R5

**Data provider?**
CosmosDB

**To Reproduce**
1. GET {FHIR SERVER URL}/metadata
2. Look at CapabilityStatement.rest.operation and patient-everything should be listed

**Expected behavior**
Capability Statement lists patient-everything as supported

**Actual behavior**
Patient-everything isn't listed

CaitlinV39:
AB#82601

PR #2000, opened 2021-06-03
Add capability statement for patient-everything

BoyaWu10:
## Description
List `patient-everything` in `CapabilityStatement.rest.operation` when the data store is Cosmos DB.

## Related issues
Addresses #1989.

## Testing
E2E tests are added to cover changes.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Api/Features/Operations/OperationsCapabilityProvider.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Api/Features/Operations/OperationsCapabilityProvider.cs
@@ -15,7 +15,6 @@ using Microsoft.Health.Fhir.Core.Features.Operations;
 using Microsoft.Health.Fhir.Core.Features.Routing;
 using Microsoft.Health.Fhir.Core.Features.Search;
 using Microsoft.Health.Fhir.Core.Models;
-using Microsoft.Health.Fhir.ValueSets;
 
 namespace Microsoft.Health.Fhir.Api.Features.Operations
 {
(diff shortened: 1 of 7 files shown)


src/Microsoft.Health.Fhir.Shared.Api/Features/Operations/OperationsCapabilityProvider.cs
```diff
@@ -75,5 +84,36 @@ public void AddMemberMatchDetails(ListedCapabilityStatement capabilityStatement)
         {
             GetAndAddOperationDefinitionUriToCapabilityStatement(capabilityStatement, OperationsConstants.MemberMatch);
         }
+
+        public void AddPatientEverythingDetails(ListedCapabilityStatement capabilityStatement)
```
LTA-Thinking, line 88: We currently have methods like this in the OperationsCapabilityProvider files in the version specific packages so that we don't need to check for the FHIR version here. Could you move this method to those files?

LTA-Thinking:
In src/Microsoft.Health.Fhir.ValueSets/OperationTypes.cs we define the link to the validate operation's definition. The url for that definition and for patient-everything are in different styles. I like the one for patient-everything better as it is more complete and uses https. Could you update the link for validate to be this: https://www.hl7.org/fhir/operation-resource-validate.html

src/Microsoft.Health.Fhir.Shared.Api/Features/Operations/OperationsCapabilityProvider.cs
BoyaWu10: Sure, the methods are moved to version specific OperationsCapabilityProvider files in the update.

BoyaWu10:
> In src/Microsoft.Health.Fhir.ValueSets/OperationTypes.cs we define the link to the validate operation's definition. The url for that definition and for patient-everything are in different styles. I like the one for patient-everything better as it is more complete and uses https. Could you update the link for validate to be this: https://www.hl7.org/fhir/operation-resource-validate.html

Sure, it is updated to the new link.

LTA-Thinking (approved):

PR #2005, opened 2021-06-03
Fix case where all detected changes are from another instance starting

feordin:
## Description
Fixes a bug where all the SearchParameters status updates from another instance starting up would be detected but not correctly processed.

## Related issues
Addresses [issue [AB#82428](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/82428)].

## Testing
Ran all E2E tests, manually tested multiple instances.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/SearchParameterOperations.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/SearchParameterOperations.cs
@@ -224,7 +224,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Parameters
 
                 if (searchParamResource == null)
                 {
-                    _logger.LogWarning(
+                    _logger.LogInformation(
                         "Updated SearchParameter status found for SearchParameter: {0}, but did not find any SearchParameter resources when querying for this url.",
                         searchParam.Uri);
                     continue;
@@ -243,12 +243,12 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Parameters
             if (paramsToAdd.Any())
             {
                 _searchParameterDefinitionManager.AddNewSearchParameters(paramsToAdd);
-
-                // Once added to the definition manager we can update their status
-                await _searchParameterStatusManager.ApplySearchParameterStatus(
-                    updatedSearchParameterStatus.Where(p => p.Status != SearchParameterStatus.Deleted).ToList(),
-                    cancellationToken);
             }
+
+            // Once added to the definition manager we can update their status
+            await _searchParameterStatusManager.ApplySearchParameterStatus(
+                updatedSearchParameterStatus.Where(p => p.Status != SearchParameterStatus.Deleted).ToList(),
+                cancellationToken);
         }
 
         private void DeleteSearchParameter(string url)
--- a/src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterStatusManager.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterStatusManager.cs
@@ -173,30 +173,32 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Registry
 
         internal async Task ApplySearchParameterStatus(IReadOnlyCollection<ResourceSearchParameterStatus> updatedSearchParameterStatus, CancellationToken cancellationToken)
         {
-            if (updatedSearchParameterStatus.Any())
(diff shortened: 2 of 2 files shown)


src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/SearchParameterOperations.cs
```diff
@@ -243,12 +243,12 @@ public async Task GetAndApplySearchParameterUpdates(CancellationToken cancellati
...
                 _searchParameterDefinitionManager.AddNewSearchParameters(paramsToAdd);
-
-                // Once added to the definition manager we can update their status
-                await _searchParameterStatusManager.ApplySearchParameterStatus(
-                    updatedSearchParameterStatus.Where(p => p.Status != SearchParameterStatus.Deleted).ToList(),
-                    cancellationToken);
             }
+
+            // Once added to the definition manager we can update their status
+            await _searchParameterStatusManager.ApplySearchParameterStatus(
```
Ivanidzo4ka, line 249: >ApplySearchParameterStatus [](http://example.com/codeflow?start=48&length=26)

From reading code it seems like only non deleted status search parameters would be added to `paramsToAdd`.

But it looks like that's not the case?

Can you explain a bit why it's not inside `if (paramsToAdd.Any())`

src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterStatusManager.cs
```diff
@@ -189,7 +189,10 @@ internal async Task ApplySearchParameterStatus(IReadOnlyCollection<ResourceSearc
                 updated.Add(param);
             }
 
-            _latestSearchParams = updatedSearchParameterStatus.Select(p => p.LastUpdated).Max();
+            if (updatedSearchParameterStatus.Any())
```
Ivanidzo4ka, line 192: >updatedSearchParameterStatus [](http://example.com/codeflow?start=16&length=28)

can we just assert it's not empty in first line of method and if it is, just leave?

Or it's necessary to run sync and process notification?

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/SearchParameterOperations.cs
feordin: I moved the call to ApplySearchParameterStatus outside the if (paramsToAdd.Any()) because it contains the call to update the _latestSearchParams in SearchParameterStatusManager.  For the case where we detect changes on startup, the paramsToAdd collection will be empty, as those spec defined parameters will not have any SearchParameter resource in the DB.  But, we still need to call ApplySearchParameterStatus.

src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterStatusManager.cs
feordin: You are correct, we can simply end if the collection is empty.

src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterStatusManager.cs
```diff
@@ -173,27 +173,30 @@ internal async Task<IReadOnlyCollection<ResourceSearchParameterStatus>> GetSearc
 
         internal async Task ApplySearchParameterStatus(IReadOnlyCollection<ResourceSearchParameterStatus> updatedSearchParameterStatus, CancellationToken cancellationToken)
         {
-            var updated = new List<SearchParameterInfo>();
-
-            foreach (var paramStatus in updatedSearchParameterStatus)
+            if (updatedSearchParameterStatus.Any())
```
Ivanidzo4ka, line 176: >if (updatedSearchParameterStatus.Any() [](http://example.com/codeflow?start=12&length=38)

super nit:
I personally find 
```
if (!updatedSearchParameterStatus.Any())
[
 return Task.CompletedTask;
}
```
more clean than wrapping whole method in if statement.

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/SearchParameterOperations.cs
Ivanidzo4ka: Are we talking about predefined search parameters? 
I just not sure how you can have status and not be in DB...

Ivanidzo4ka (approved):

PR #2008, opened 2021-06-04
Add OperationOutcome for unsupported parameters in patient-everything

BoyaWu10:
## Description
Currently _count is not supported in patient-everything.
Add an OperationOutcome in the response Bundle if there are unsupported input parameters.

## Related issues
[#82636](https://microsofthealth.visualstudio.com/Health/_workitems/edit/82636)

## Testing
Unit tests and E2E tests are updated to cover changes.

## FHIR Team Checklist


src/Microsoft.Health.Fhir.Shared.Api/Controllers/EverythingController.cs
```diff
@@ -0,0 +1,126 @@
...
+using Microsoft.Health.Fhir.ValueSets;
+
+namespace Microsoft.Health.Fhir.Api.Controllers
+{
+    [ServiceFilter(typeof(AuditLoggingFilterAttribute))]
+    [ServiceFilter(typeof(OperationOutcomeExceptionFilterAttribute))]
+    [ServiceFilter(typeof(ValidateFormatParametersAttribute))]
+    [ValidateResourceTypeFilter]
+    [ValidateModelState]
+    public class EverythingController : Controller
```
BoyaWu10, line 34: Move this to a separate file to be consistent with $validate operation.

LTA-Thinking:
This is a separate issue, but for other operations when there is an unsupported parameter we remove it from the self link. The self link retuned from $everything still has the unsupported parameters.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/EverythingController.cs
brendankowitz: @Ivanidzo4ka thoughts on creating a /Controllers/Operations folder?

src/Microsoft.Health.Fhir.Shared.Api/Controllers/EverythingController.cs
```diff
@@ -0,0 +1,126 @@
...
+
+            start = ReadPartialDateTimeParameter(inputParameters, EverythingOperationParameterNames.Start);
+            end = ReadPartialDateTimeParameter(inputParameters, EverythingOperationParameterNames.End);
+            since = ReadPartialDateTimeParameter(inputParameters, KnownQueryParameterNames.Since);
+            type = ReadStringParameter(inputParameters, KnownQueryParameterNames.Type);
+            ct = ReadStringParameter(inputParameters, KnownQueryParameterNames.ContinuationToken);
+            ProcessUnsupportedInputParameters(inputParameters);
+        }
+
+        private static string ReadStringParameter(IReadOnlyList<Tuple<string, string>> inputParameters, string parameterName)
```
LTA-Thinking, line 93: Why do we need these parameter reading methods instead of the FromQuery method used in the old version?

BoyaWu10:
> This is a separate issue, but for other operations when there is an unsupported parameter we remove it from the self link. The self link retuned from $everything still has the unsupported parameters.

Thanks for sharing this. The unsupported parameters are removed from `Bundle.link.url` in the update.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/EverythingController.cs
BoyaWu10: Right, the valid ones are read from the FromQuery (as in the old version) in the update.

LTA-Thinking (approved):

PR #2014, opened 2021-06-05
Added cancellation token to SearchParameterStatusDataStore operations

### LINKED ISSUE #1282: Add cancellation token to SQL search parameter status data layer
**User story**
Give users the ability to pass in cancellation tokens in the SQL search parameter status data layer. See PR #1191 (discussion_r472542412)

**Acceptance criteria**
User can pass in cancellation token

### LINKED ISSUE #1869: Add cancellation logic to SearchParameterStatusDataStore operations
**User story**
The search parameter status operations GetSearchParameterStatuses() and UpsertStatuses() should take in a cancellation token as input.

**Acceptance criteria**
Cancellation logic is added to search parameter status operations.

PTaladay:
## Description
 It extended into other areas in order to ensure we were passing the token all the way through. New cancellation token parameters that were added were set with default cancellation token if not provided in order to support backwards compatibility. SetupContainerAsync in the CollectionUpgradeManager does not pass a cancellation token and instead instantiates one so I left that for now.

## Related issues
This fixes #1869 , fixes #1282

## Testing
Ran existing test suite provided by solution. I don't have a local setup to run some of the integration tests, but I validated tests that explicitly call the updated methods.

## FHIR Team Checklist


PTaladay:
Based on the failing tests it looks like an authentication issue, unrelated to changes in this PR. Might be transient, not sure.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Registry/CosmosDbSearchParameterStatusDataStore.cs
```diff
@@ -39,13 +39,12 @@ public sealed class CosmosDbSearchParameterStatusDataStore : ISearchParameterSta
...
 
-        public async Task<IReadOnlyCollection<ResourceSearchParameterStatus>> GetSearchParameterStatuses()
+        public async Task<IReadOnlyCollection<ResourceSearchParameterStatus>> GetSearchParameterStatuses(CancellationToken cancellationToken = default)
         {
-            using var cancellationSource = new CancellationTokenSource(TimeSpan.FromMinutes(1));
             using IScoped<Container> clientScope = _containerScopeFactory.Invoke();
             DateTimeOffset startedCheck = Clock.UtcNow;
 
-            await _statusListSemaphore.WaitAsync(cancellationSource.Token);
+            await _statusListSemaphore.WaitAsync(cancellationToken);
```
namadabu, line 47: Having a default value for the CancellationToken here changes existing behavior (which is to have a CancellationToken that gets cancelled after 1 minute). Is this something we want to do for this method? If not, maybe we should add logic to check for the default value and handle appropriately?

PTaladay: That is a fair point. In a few of the locations CancellationToken.None was being passed, so I took this as something that was put in place as a cancellation token wasn't being passed here. If there was an intent here to keep it short then we could look into additional logic. I don't have enough context to know if that is the case, so if @CaitlinV39 maybe knows, that would be helpful.

CaitlinV39: @rotodd originally found this so I'll see if she has additional context.

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/ISearchParameterOperations.cs
```diff
@@ -12,11 +12,11 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Parameters
 {
     public interface ISearchParameterOperations
     {
-        Task AddSearchParameterAsync(ITypedElement searchParam);
+        Task AddSearchParameterAsync(ITypedElement searchParam, CancellationToken cancellationToken = default);
```
brendankowitz, line 15: In general we avoid using default values with ContinuationToken, this helps ensure that the calling code is doing the right thing and not accidentally missing it

PTaladay: @brendankowitz  If the preference is to break when it isn't passed in then that is fine. Would you like me to switch to throwing an argument null exception if that value isn't passed instead of going for a default value? 

There are other portions of the code that also utilize a default cancellation token and that is typically what I do to prevent the system from breaking when a new parameter is introduced. So for consistency would those need to be removed?

rotodd: To summarize our discussion: we will remove the default cancellation token as an optional argument so that it can be controlled from the top of the call stack (the FHIR controller).

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Registry/CosmosDbSearchParameterStatusDataStore.cs
rotodd: Changes in 038d041 look good to me. We will have two means of cancellation:

1. We will keep the existing behaviour to timeout after 1 minute (a thought: should we also have this for SQL?)
2. We will pass the cancellation token created at the top of the call stack into the database calls

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Registry/CosmosDbSearchParameterStatusDataStore.cs
```diff
@@ -39,11 +39,11 @@ public sealed class CosmosDbSearchParameterStatusDataStore : ISearchParameterSta
             _queryFactory = queryFactory;
         }
 
-        public async Task<IReadOnlyCollection<ResourceSearchParameterStatus>> GetSearchParameterStatuses()
+        public async Task<IReadOnlyCollection<ResourceSearchParameterStatus>> GetSearchParameterStatuses(CancellationToken cancellationToken)
```
rotodd, line 42: Super nit: do you think it would be useful to add a comment here to explain why we have two means of cancellation? Or maybe we could rename `cancellationSource` to add clarity?

rotodd (approved):

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Registry/CosmosDbSearchParameterStatusDataStore.cs
PTaladay: Yep, makes sense. I changed the name to retryDelayToken as that seems to be it's purpose.

Issue #1967, opened 2021-05-24
Export doesn't recognize the seconds of the instance


CaitlinV39:
**Describe the bug**
A customer reported that the Azure API for FHIR doesn't respect instance on export when using _since. It only looks at date. 

**FHIR Version?**
Stu3/R4/R5

**Data provider?**
CosmosDB/SQL Server

**To Reproduce**
1. Run an export with _since set to an instance
2. 
3. 

**Expected behavior**
Data would be exported since that instant

**Actual behavior**
Data is exported since the full date, not looking at instance.

Ivanidzo4ka:
There is two layers in this bug.
- `PartialDateTime` class we use is not working well with `instant` time and fail to parse correct `instant` string.
- There is no model validation in Export controller so we ignore fact what `_since` parameter throw exception during parsing.

For `PartialDateTime` there is new classes in 3.0 firely library like `DateTime` which I think we can utilize.

CaitlinV39:
AB#82455

LTA-Thinking:
A manifestation of this bug. If _since is the last parameter in a url and there is a space included after the datetime, the datetime will fail to parse and will be ignored entirely.

PR #2023, opened 2021-06-14
Don't ignore parse error in partial date time binder

Ivanidzo4ka:
## Description
Make sure we fail request for export in date is incorrect.

## Related issues
Addresses [#1967].

## Testing
Unit tests

## FHIR Team Checklist

Patch

### DIFF
--- a/src/Microsoft.Health.Fhir.Api/Features/Binders/PartialDateTimeBinder.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/Binders/PartialDateTimeBinder.cs
@@ -3,6 +3,7 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
+using System;
 using System.Linq;
 using System.Threading.Tasks;
 using EnsureThat;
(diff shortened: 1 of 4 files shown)


src/Microsoft.Health.Fhir.Api/Features/Binders/PartialDateTimeBinder.cs
```diff
@@ -27,8 +28,9 @@ public Task BindModelAsync(ModelBindingContext bindingContext)
 
                     bindingContext.Result = ModelBindingResult.Success(result);
                 }
-                catch
+                catch (FormatException ex)
```
LTA-Thinking, line 31: Should we still have a universal exception catch just in case, or let it bubble up?

LTA-Thinking (approved):

src/Microsoft.Health.Fhir.Api/Features/Binders/PartialDateTimeBinder.cs
Ivanidzo4ka: I'm not big fan of blank policies. They tend to hide mistakes from you.
Right now only exception we intent to throw during parse is FormatException, and if something would go wrong and I put catch for Exception, we will never see it in logs, therefore we would be blind to something going on.

As an option I can probably plumb logger here and log critical exception, but I'm too lazy for that.

PR #2027, opened 2021-06-15
Supports extension search parameter expressions

rotodd:
## Description
This PR adds support to resolve the FHIR type for custom search parameters that target choice type values in resource extensions. Previously, users would need to write more specific search parameter expressions:

`Organization.extension.where(url = 'https://mycompany.com/health-dictionary/DirectoryType').value.coding.code`

Now, they should be able to simply put:

`Organization.extension.where(url = 'https://mycompany.com/health-dictionary/DirectoryType')`

Note: this was a team effort! @Ivanidzo4ka gets creds for writing the extension converter.

## Related issues
Addresses [AB#82415](https://microsofthealth.visualstudio.com/Health/_workitems/edit/82415).

## Testing
Adds new unit tests and new HTTP file for manual test scenario.

## FHIR Team Checklist

Skip (bug fix)

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Search/Converters/FhirTypedElementToSearchValueConverterManager.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Search/Converters/FhirTypedElementToSearchValueConverterManager.cs
@@ -42,7 +42,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Converters
             return _converterDictionary.TryGetValue((fhirType, searchValueType), out converter);
         }
 
-        private class ExtensionConverter<T> : ITypedElementToSearchValueConverter
+        internal class ExtensionConverter<T> : ITypedElementToSearchValueConverter
         {
             private readonly IEnumerable<ITypedElementToSearchValueConverter> _underlying;
             private readonly Type _type;
--- a/src/Microsoft.Health.Fhir.Core/Features/Search/Converters/FhirTypedElementToSearchValueConverterManager.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Search/Converters/FhirTypedElementToSearchValueConverterManager.cs
@@ -64,11 +64,19 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Converters
             {
                 if (value == null)
                 {
-                    return new List<ISearchValue>();
+                    return Enumerable.Empty<ISearchValue>();
                 }
 
                 var typed = value.Select("value").FirstOrDefault();
-                var converter = _searchValueTypeConverters.First(x => x.FhirTypes.Contains(typed.InstanceType));
+                var converter = _searchValueTypeConverters.FirstOrDefault(x => x.FhirTypes.Contains(typed?.InstanceType));
+
+                // If the resource's extension type can't be converted to the user-specified search parameter type, we won't return any results.
+                // This means reindexing will succeed but the search parameter will not pick up resources with extensions that have an incompatible type.
+                if (converter == null)
+                {
+                    return Enumerable.Empty<ISearchValue>();
+                }
+
                 return converter.ConvertTo(typed);
             }
         }
--- a/src/Microsoft.Health.Fhir.Core/Features/Search/Converters/FhirTypedElementToSearchValueConverterManager.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Search/Converters/FhirTypedElementToSearchValueConverterManager.cs
@@ -59,6 +59,11 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Converters
 
             public IEnumerable<ISearchValue> ConvertTo(ITypedElement value)
             {
+                if (value == null)
+                {
+                    return new List<ISearchValue>();
+                }
+
                 var typed = value.Select("value").FirstOrDefault();
                 var converter = _underlying.Where(x => x.FhirTypes.Contains(typed.InstanceType)).First();
                 return converter.ConvertTo(typed);
--- a/src/Microsoft.Health.Fhir.Core/Features/Search/Converters/FhirTypedElementToSearchValueConverterManager.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Search/Converters/FhirTypedElementToSearchValueConverterManager.cs
@@ -7,6 +7,9 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using EnsureThat;
+using Hl7.Fhir.ElementModel;
+using Hl7.FhirPath;
+using Microsoft.Health.Fhir.Core.Features.Search.SearchValues;
 
 namespace Microsoft.Health.Fhir.Core.Features.Search.Converters
 {
@@ -21,8 +24,10 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Converters
         {
             EnsureArg.IsNotNull(converters, nameof(converters));
 
-            _converterDictionary = converters
-                .SelectMany(converter => converter.FhirTypes.Select(type => new { FhirType=type, converter.SearchValueType, Converter=converter }))
+            var extensions = converters.GroupBy(x => x.SearchValueType).Select(group => new ExtensionConverter<int>(group.ToList())).ToArray();
+
+            _converterDictionary = converters.Concat(extensions)
+                .SelectMany(converter => converter.FhirTypes.Select(type => new { FhirType = type, converter.SearchValueType, Converter = converter }))
                 .ToDictionary(
                     converter => (converter.FhirType, converter.SearchValueType),
                     converter => converter.Converter);
@@ -35,6 +40,29 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Converters
             EnsureArg.IsNotNull(searchValueType, nameof(searchValueType));
 
             return _converterDictionary.TryGetValue((fhirType, searchValueType), out converter);
+        }
+
+        private class ExtensionConverter<T> : ITypedElementToSearchValueConverter
+        {
+            private readonly IEnumerable<ITypedElementToSearchValueConverter> _underlying;
+            private readonly Type _type;
+
+            public ExtensionConverter(IEnumerable<ITypedElementToSearchValueConverter> underlying)
+            {
+                _underlying = underlying;
+                _type = underlying.First().SearchValueType;
+            }
+
+            public IReadOnlyList<string> FhirTypes => new List<string> { "Extension" };
+
+            public Type SearchValueType => _type;
+
+            public IEnumerable<ISearchValue> ConvertTo(ITypedElement value)
+            {
+                var typed = value.Select("value").FirstOrDefault();
+                var converter = _underlying.Where(x => x.FhirTypes.Contains(typed.InstanceType)).First();
+                return converter.ConvertTo(typed);
+            }
         }
     }
 }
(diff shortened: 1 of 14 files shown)


src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/SearchParameters/SearchParameterFixtureData.cs
```diff
@@ -66,8 +68,17 @@ private static async Task<FhirTypedElementToSearchValueConverterManager> CreateF
...
             await codeSystemResolver.StartAsync(CancellationToken.None);
 
-            var fhirElementToSearchValueConverters =
-                types.Select(x => (ITypedElementToSearchValueConverter)Mock.TypeWithArguments(x, referenceSearchValueParser, codeSystemResolver));
+            var fhirElementToSearchValueConverters = new List<ITypedElementToSearchValueConverter>();
+
+            foreach (Type type in types)
+            {
+                // The extension converters will be added to the converter dictionary by the converter manager
+                if (!type.Name.Contains("ExtensionConverter"))
```
Ivanidzo4ka, line 76: >type.Name.Contains("ExtensionConverter") [](http://example.com/codeflow?start=21&length=40)

I would try to use `nameof`
another option is ExtensionsConverter is a generic type, while all other converters are not. 
`if (type.ContainsGenericParameters)`

src/Microsoft.Health.Fhir.Core/Features/Search/Converters/FhirTypedElementToSearchValueConverterManager.cs
```diff
@@ -36,5 +41,33 @@ public bool TryGetConverter(string fhirType, Type searchValueType, out ITypedEle
 
             return _converterDictionary.TryGetValue((fhirType, searchValueType), out converter);
         }
+
+        internal class ExtensionConverter<T> : ITypedElementToSearchValueConverter
```
Ivanidzo4ka, line 45: ><T> [](http://example.com/codeflow?start=41&length=3)

Can we add comment why it's generic?

Reason why it's generic is to not add this converter to DI.

[https://github.com/microsoft/healthcare-shared-components/blob/1e0dc8532113a5f7fdd0692c922e91b8273c240e/src/Microsoft.Health.Extensions.DependencyInjection/TypeRegistrationExtensions.cs#L23](https://github.com/microsoft/healthcare-shared-components/blob/1e0dc8532113a5f7fdd0692c922e91b8273c240e/src/Microsoft.Health.Extensions.DependencyInjection/TypeRegistrationExtensions.cs#L23)

we use this method to inject all interface implemenations into DI, and one of the checks to add it - is this class generic or not.

Ivanidzo4ka (approved):
:shipit:

src/Microsoft.Health.Fhir.Core/Features/Search/Converters/FhirTypedElementToSearchValueConverterManager.cs
```diff
@@ -36,5 +41,33 @@ public bool TryGetConverter(string fhirType, Type searchValueType, out ITypedEle
...
+        {
+            private readonly IEnumerable<ITypedElementToSearchValueConverter> _underlying;
+            private readonly Type _type;
+
+            public ExtensionConverter(IEnumerable<ITypedElementToSearchValueConverter> underlying)
+            {
+                _underlying = underlying;
+                _type = underlying.First().SearchValueType;
```
Ivanidzo4ka, line 53: >underlying.First() [](http://example.com/codeflow?start=24&length=18)

Can we assert Enumerable to have an element?

src/Microsoft.Health.Fhir.Core/Features/Search/Converters/FhirTypedElementToSearchValueConverterManager.cs
```diff
@@ -36,5 +41,33 @@ public bool TryGetConverter(string fhirType, Type searchValueType, out ITypedEle
...
+
+            public IEnumerable<ISearchValue> ConvertTo(ITypedElement value)
+            {
+                if (value == null)
+                {
+                    return new List<ISearchValue>();
+                }
+
+                var typed = value.Select("value").FirstOrDefault();
+                var converter = _underlying.Where(x => x.FhirTypes.Contains(typed.InstanceType)).First();
```
Ivanidzo4ka, line 68: >First() [](http://example.com/codeflow?start=97&length=7)

This has a chance to blow out.

src/Microsoft.Health.Fhir.Core/Features/Search/Converters/FhirTypedElementToSearchValueConverterManager.cs
brendankowitz: This doesn't sound like a good reason to use a Generic... This problem also exists in other places too where there is a class that shouldn't be included in the scanning convention.

The simple solution would just be at add this to a `.Where` clause to exclude it:
https://github.com/microsoft/fhir-server/blob/1c5cd55a505a1f98cd82d3532fa7e198a713af15/src/Microsoft.Health.Fhir.Shared.Api/Modules/SearchModule.cs#L92-L95

A broader solution might be to somehow use an attribute or tag for `[NoScanningRegistration]` (I don't like the name), but thoughts?

src/Microsoft.Health.Fhir.Core/Features/Search/Converters/FhirTypedElementToSearchValueConverterManager.cs
brendankowitz: Can we add a comment for whats getting passed in here :)? I'm also not totally following why its a collection but using the first?

Ivanidzo4ka: and do 
```
new ExtensionConverter<int>(group.Key, group.ToList()
```

src/Microsoft.Health.Fhir.Core/Features/Search/Converters/FhirTypedElementToSearchValueConverterManager.cs
Ivanidzo4ka, line 45: >internal [](http://example.com/codeflow?start=8&length=8)

can it be private?

rotodd: I ended up making this internal so that I could access it from the test files!

src/Microsoft.Health.Fhir.Core/Features/Search/Converters/FhirTypedElementToSearchValueConverterManager.cs
rotodd: Good idea, implemented this in 51adeb6.

src/Microsoft.Health.Fhir.Core/Features/Search/Converters/FhirTypedElementToSearchValueConverterManager.cs
```diff
@@ -36,5 +41,36 @@ public bool TryGetConverter(string fhirType, Type searchValueType, out ITypedEle
...
+
+            public IReadOnlyList<string> FhirTypes => new List<string> { "Extension" };
+
+            public Type SearchValueType { get; }
+
+            public IEnumerable<ISearchValue> ConvertTo(ITypedElement value)
+            {
+                if (value == null)
+                {
+                    return new List<ISearchValue>();
```
Ivanidzo4ka, line 67: >List<ISearchValue>() [](http://example.com/codeflow?start=31&length=20)

nit: [https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.empty?view=net-5.0](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.empty?view=net-5.0)

src/Microsoft.Health.Fhir.Core/Features/Search/Converters/FhirTypedElementToSearchValueConverterManager.cs
rotodd: Made `ExtensionConverter` non-generic and added a `.Where` clause in 51adeb6.

Using an attribute is an interesting idea... I created a story to investigate ([AB#82937](https://microsofthealth.visualstudio.com/Health/_workitems/edit/82937)).

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/SearchParameters/SearchParameterFixtureData.cs
rotodd: Changed to use `nameof` in 51adeb6.

src/Microsoft.Health.Fhir.Core/Features/Search/Converters/FhirTypedElementToSearchValueConverterManager.cs
Ivanidzo4ka: ExtensionConverter groups together multiple underlying converters.

There is chance what we trying to convert type A to type B, and we have converter for A to C.

We will have ExtensionConverter which takes (A) type (because we have A->C), but if somehow we looking to extract type B it will explode with null exception.

So it should be FirstOrDefault() and null check.

rotodd: Still working on how to best handle the case where we don't find the right converter. Ideally, we should throw an error during search parameter validation; however, right now, this will allow users to post the search parameter and reindexing will later fail.

Ivanidzo4ka: Or maybe that's not the case based on nature of relationship between Type and FhirType.

src/Microsoft.Health.Fhir.Core/Features/Search/Converters/FhirTypedElementToSearchValueConverterManager.cs
rotodd: Ah yes that is nicer, thanks. Fixed in a7c8f99.

src/Microsoft.Health.Fhir.Core/Features/Search/Converters/FhirTypedElementToSearchValueConverterManager.cs
rotodd: A summary of my discussion with @Ivanidzo4ka:

If no matching converter is found, it means that the resource's extension type can't be converted to the type specified in the search parameter definition.

For example, no matching converter would be found if we posted a resource with the following extension:

```
{
    "resourceType": "Condition",
   ...
    "extension": [
        {
            "extension": [
                {
                    ...
                }
            ],
            "url": ...
        },
        {
            "extension": [
                {
                    "url": "serviceProviderId",
                    "valueId": "123456"
                },
                {
                    "url": "journeyId",
                    "valueId": "678910"
                },
            ],
            "url": "http://example.org/fhir/StructureDefinition/RobinsPRExample"
        }
    ],
...
}
```

and a search parameter with the following type:

```
{
   "resourceType":"SearchParameter",
   ...
   "base":[
      "Condition",
      "Patient",
      "Observation"
   ],
   "type":"string",
   "expression":"DomainResource.extension.where(url='http://example.org/fhir/StructureDefinition/RobinsPRExample').extension.where(url='journeyId')"
}
```

since we can't convert `FhirType` `Id` to `Type` `StringSearchValue`.

We can't throw an error during search parameter validation, since we do not know the extension type at this point. Another thing to consider: we could have multiple resources with different type-variations of this extension.

If we decided to throw an error in the `ExtensionConverter`, we would fail to reindex all resources if one or more resources had an extension with an incompatible type. We would also be unable to post any new resources with this extension.

For now, we have (tentatively) decided to simply return nothing if no compatible type is found. This will allow resources with the correct type to be properly indexed, and it won't block us from creating new resources with this extension. It will not, however, give information to the user as to why resources with an incompatible extension type aren't being picked up.

@Ivanidzo4ka is there anything I missed? @brendankowitz thoughts?

As Ivan recommended, I am planning to reach out to the FHIR community to see how others have handled this edge case. I'll post the link to the post once I create it.

PR #2028, opened 2021-06-15
Configurable max query concurrency

rbans96:
## Description
Adds configurable max number of parallel queries to execute selective queries across physical partitions

## Related issues
Addresses [issue #].

## Testing

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Web/appsettings.json
+++ b/src/Microsoft.Health.Fhir.Shared.Web/appsettings.json
@@ -130,7 +130,7 @@
             "http://hl7.org/fhir/SearchParameter/clinical-date"
         ],
         "ParallelQueryOptions": {
-            "MaxQueryConcurrency": 1000
+            "MaxQueryConcurrency": 500
         }
     },
     "DataStore": "CosmosDb",
--- a/src/Microsoft.Health.Fhir.Shared.Web/appsettings.json
+++ b/src/Microsoft.Health.Fhir.Shared.Web/appsettings.json
@@ -128,7 +128,10 @@
             "http://hl7.org/fhir/SearchParameter/individual-given",
             "http://hl7.org/fhir/SearchParameter/individual-birthdate",
             "http://hl7.org/fhir/SearchParameter/clinical-date"
-        ]
+        ],
+        "ParallelQueryOptions": {
+            "MaxQueryConcurrency": 1000
+        }
     },
     "DataStore": "CosmosDb",
     "KeyVault": {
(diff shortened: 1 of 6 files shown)


src/Microsoft.Health.Fhir.Shared.Web/appsettings.json
```diff
@@ -128,7 +128,10 @@
             "http://hl7.org/fhir/SearchParameter/individual-given",
             "http://hl7.org/fhir/SearchParameter/individual-birthdate",
             "http://hl7.org/fhir/SearchParameter/clinical-date"
-        ]
+        ],
+        "ParallelQueryOptions": {
+            "MaxQueryConcurrency": 1000
```
deepak-ms, line 133: Should we have default value as 500?

rbans96, line 133: @namadabu If I am not wrong, as per your queries testing using portal, 1000 is the max number for which cosmosdb sdk has returned results successfully so I am thinking if we could use max number that is known to us so far. Thoughts?

brendankowitz: Agree, for now, lets use 500. 
I think its also fine for now for this to be the hardcoded default because this is the best known current value. This gives us the option to change at a later point without recompile.

rbans96: Thanks, I have defaulted it to 500. I understand for 1000 it is taking much longer, but what about setting it to 750 by default. 
Just another thought - As per the test numbers shared by Narasi, I see that one of the occurence of 500 took 27secs and 750 took 28secs, so I think 750 is still better.

image

namadabu: @rbans96 for 1000 we got a much slower response compared to when using 500. So it would be better to go with 500 and then modify according to the behavior we see.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/CustomSearchParamTests.cs
```diff
@@ -232,7 +232,7 @@ public async Task GivenAnInvalidSearchParam_WhenCreatingParam_ThenMeaningfulErro
             }
         }
 
-        [SkippableFact]
+        [Fact(Skip = "To re-enable when the bug(82891) is fixed")]
```
rbans96, line 235: As discussed with @feordin I am skipping couple of tests which is failing due to bug. Once that gets resolved, we will re-enable the tests. Workitem created for this one - https://microsofthealth.visualstudio.com/Health/_boards/board/t/Resolute/Stories/?workitem=82891

src/Microsoft.Health.Fhir.Shared.Web/appsettings.json
brendankowitz: The ticket @namadabu opened confirmed that this is a CPU/Task limitation (https://github.com/Azure/azure-cosmos-dotnet-v3/issues/2556) I think we should stick with a lower number. It is configurable so we can alter later in higher perf environments

rbans96: Makes sense, thanks!

brendankowitz (approved):
Ship it!

PR #2032, opened 2021-06-16
SQL Schema changes for resource changes

bbatjargal:
## Description
When the resource change capture feature is enabled, the CUD information of Fhir resources will be stored in a ResourceChangeData table on the SQL database. 

This PR:
- Adds a ResourceChangeData table that stores resource change information.
- Adds a ResourceChangeType table which is a reference table to store change types.
- Adds a CaptureResourceChanges procedure which inserts resource change data and called from the UpsertResource proc.
- Adds a FetchResourceChanges procedure which returns the number of resource change records from startId.
- Adds a new version of the UpsertResource procedure (UpsertResource_4).
- Adds a core feature config called SupportsResourceChangeCapture.
- Adds a new schema version (version 14). 

## Related issues
Addresses  [#82760](https://microsofthealth.visualstudio.com/Health/_workitems/edit/82760).

## Testing
Ran a Fhir instance locally, CUD a resource, and validated resource changes on the database. Integration and e2e test would be added when it integrated with Fhir service.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Configs/CoreFeatureConfiguration.cs
+++ b/src/Microsoft.Health.Fhir.Core/Configs/CoreFeatureConfiguration.cs
@@ -56,5 +56,10 @@ namespace Microsoft.Health.Fhir.Core.Configs
         /// Maximum items allowed to be deleted when using Conditional Delete.
         /// </summary>
         public int ConditionalDeleteMaxItems { get; set; } = 100;
+
+        /// <summary>
+        /// Gets or sets a value whether capturing resource change data is enabled or not.
+        /// </summary>
+        public bool SupportsResourceChangeCapture { get; set; } = false;
     }
 }
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.diff.sql
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.diff.sql
@@ -22,62 +22,62 @@ AS
     SET NOCOUNT ON
 
     SET XACT_ABORT ON
-    BEGIN TRANSACTION
+BEGIN TRANSACTION
 
     DECLARE @resourceSurrogateIds TABLE(ResourceSurrogateId bigint NOT NULL)
 
-            DELETE FROM dbo.Resource
+    DELETE FROM dbo.Resource
         OUTPUT deleted.ResourceSurrogateId
     INTO @resourceSurrogateIds
     WHERE ResourceTypeId = @resourceTypeId AND ResourceId = @resourceId
       AND NOT(@keepCurrentVersion=1 and IsHistory=0)
 
-            DELETE FROM dbo.ResourceWriteClaim
+    DELETE FROM dbo.ResourceWriteClaim
     WHERE ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @resourceSurrogateIds)
 
-            DELETE FROM dbo.CompartmentAssignment
+    DELETE FROM dbo.CompartmentAssignment
     WHERE ResourceTypeId = @resourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @resourceSurrogateIds)
 
-            DELETE FROM dbo.ReferenceSearchParam
+    DELETE FROM dbo.ReferenceSearchParam
     WHERE ResourceTypeId = @resourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @resourceSurrogateIds)
 
-            DELETE FROM dbo.TokenSearchParam
+    DELETE FROM dbo.TokenSearchParam
     WHERE ResourceTypeId = @resourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @resourceSurrogateIds)
 
-            DELETE FROM dbo.TokenText
+    DELETE FROM dbo.TokenText
     WHERE ResourceTypeId = @resourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @resourceSurrogateIds)
 
-            DELETE FROM dbo.StringSearchParam
+    DELETE FROM dbo.StringSearchParam
     WHERE ResourceTypeId = @resourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @resourceSurrogateIds)
 
-            DELETE FROM dbo.UriSearchParam
+    DELETE FROM dbo.UriSearchParam
     WHERE ResourceTypeId = @resourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @resourceSurrogateIds)
 
-            DELETE FROM dbo.NumberSearchParam
+    DELETE FROM dbo.NumberSearchParam
     WHERE ResourceTypeId = @resourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @resourceSurrogateIds)
 
-            DELETE FROM dbo.QuantitySearchParam
+    DELETE FROM dbo.QuantitySearchParam
     WHERE ResourceTypeId = @resourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @resourceSurrogateIds)
 
-            DELETE FROM dbo.DateTimeSearchParam
+    DELETE FROM dbo.DateTimeSearchParam
     WHERE ResourceTypeId = @resourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @resourceSurrogateIds)
 
-            DELETE FROM dbo.ReferenceTokenCompositeSearchParam
+    DELETE FROM dbo.ReferenceTokenCompositeSearchParam
     WHERE ResourceTypeId = @resourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @resourceSurrogateIds)
 
-            DELETE FROM dbo.TokenTokenCompositeSearchParam
+    DELETE FROM dbo.TokenTokenCompositeSearchParam
     WHERE ResourceTypeId = @resourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @resourceSurrogateIds)
 
-            DELETE FROM dbo.TokenDateTimeCompositeSearchParam
+    DELETE FROM dbo.TokenDateTimeCompositeSearchParam
     WHERE ResourceTypeId = @resourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @resourceSurrogateIds)
 
-            DELETE FROM dbo.TokenQuantityCompositeSearchParam
+    DELETE FROM dbo.TokenQuantityCompositeSearchParam
     WHERE ResourceTypeId = @resourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @resourceSurrogateIds)
 
-            DELETE FROM dbo.TokenStringCompositeSearchParam
+    DELETE FROM dbo.TokenStringCompositeSearchParam
     WHERE ResourceTypeId = @resourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @resourceSurrogateIds)
 
-            DELETE FROM dbo.TokenNumberNumberCompositeSearchParam
+    DELETE FROM dbo.TokenNumberNumberCompositeSearchParam
     WHERE ResourceTypeId = @resourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @resourceSurrogateIds)
 
     COMMIT TRANSACTION
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.diff.sql
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.diff.sql
@@ -22,62 +22,62 @@ AS
     SET NOCOUNT ON
 
     SET XACT_ABORT ON
-BEGIN TRANSACTION
+    BEGIN TRANSACTION
 
     DECLARE @resourceSurrogateIds TABLE(ResourceSurrogateId bigint NOT NULL)
 
-    DELETE FROM dbo.Resource
+            DELETE FROM dbo.Resource
         OUTPUT deleted.ResourceSurrogateId
     INTO @resourceSurrogateIds
     WHERE ResourceTypeId = @resourceTypeId AND ResourceId = @resourceId
       AND NOT(@keepCurrentVersion=1 and IsHistory=0)
 
-    DELETE FROM dbo.ResourceWriteClaim
+            DELETE FROM dbo.ResourceWriteClaim
     WHERE ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @resourceSurrogateIds)
 
-    DELETE FROM dbo.CompartmentAssignment
+            DELETE FROM dbo.CompartmentAssignment
     WHERE ResourceTypeId = @resourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @resourceSurrogateIds)
 
-    DELETE FROM dbo.ReferenceSearchParam
+            DELETE FROM dbo.ReferenceSearchParam
     WHERE ResourceTypeId = @resourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @resourceSurrogateIds)
 
-    DELETE FROM dbo.TokenSearchParam
+            DELETE FROM dbo.TokenSearchParam
     WHERE ResourceTypeId = @resourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @resourceSurrogateIds)
 
-    DELETE FROM dbo.TokenText
+            DELETE FROM dbo.TokenText
     WHERE ResourceTypeId = @resourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @resourceSurrogateIds)
 
-    DELETE FROM dbo.StringSearchParam
+            DELETE FROM dbo.StringSearchParam
     WHERE ResourceTypeId = @resourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @resourceSurrogateIds)
 
-    DELETE FROM dbo.UriSearchParam
+            DELETE FROM dbo.UriSearchParam
     WHERE ResourceTypeId = @resourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @resourceSurrogateIds)
 
-    DELETE FROM dbo.NumberSearchParam
+            DELETE FROM dbo.NumberSearchParam
     WHERE ResourceTypeId = @resourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @resourceSurrogateIds)
 
-    DELETE FROM dbo.QuantitySearchParam
+            DELETE FROM dbo.QuantitySearchParam
     WHERE ResourceTypeId = @resourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @resourceSurrogateIds)
 
-    DELETE FROM dbo.DateTimeSearchParam
+            DELETE FROM dbo.DateTimeSearchParam
     WHERE ResourceTypeId = @resourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @resourceSurrogateIds)
 
-    DELETE FROM dbo.ReferenceTokenCompositeSearchParam
+            DELETE FROM dbo.ReferenceTokenCompositeSearchParam
     WHERE ResourceTypeId = @resourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @resourceSurrogateIds)
 
-    DELETE FROM dbo.TokenTokenCompositeSearchParam
+            DELETE FROM dbo.TokenTokenCompositeSearchParam
     WHERE ResourceTypeId = @resourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @resourceSurrogateIds)
 
-    DELETE FROM dbo.TokenDateTimeCompositeSearchParam
+            DELETE FROM dbo.TokenDateTimeCompositeSearchParam
     WHERE ResourceTypeId = @resourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @resourceSurrogateIds)
 
-    DELETE FROM dbo.TokenQuantityCompositeSearchParam
+            DELETE FROM dbo.TokenQuantityCompositeSearchParam
     WHERE ResourceTypeId = @resourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @resourceSurrogateIds)
 
-    DELETE FROM dbo.TokenStringCompositeSearchParam
+            DELETE FROM dbo.TokenStringCompositeSearchParam
     WHERE ResourceTypeId = @resourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @resourceSurrogateIds)
 
-    DELETE FROM dbo.TokenNumberNumberCompositeSearchParam
+            DELETE FROM dbo.TokenNumberNumberCompositeSearchParam
     WHERE ResourceTypeId = @resourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @resourceSurrogateIds)
 
     COMMIT TRANSACTION
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.sql
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.sql
@@ -1748,7 +1748,7 @@ AS
     DECLARE @resourceSurrogateIds TABLE(ResourceSurrogateId bigint NOT NULL)
 
     DELETE FROM dbo.Resource
-        OUTPUT deleted.ResourceSurrogateId
+    OUTPUT deleted.ResourceSurrogateId
     INTO @resourceSurrogateIds
     WHERE ResourceTypeId = @resourceTypeId AND ResourceId = @resourceId
       AND NOT(@keepCurrentVersion=1 and IsHistory=0)
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.sql
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.sql
@@ -1748,7 +1748,7 @@ AS
     DECLARE @resourceSurrogateIds TABLE(ResourceSurrogateId bigint NOT NULL)
 
     DELETE FROM dbo.Resource
-    OUTPUT deleted.ResourceSurrogateId
+        OUTPUT deleted.ResourceSurrogateId
     INTO @resourceSurrogateIds
     WHERE ResourceTypeId = @resourceTypeId AND ResourceId = @resourceId
       AND NOT(@keepCurrentVersion=1 and IsHistory=0)
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
@@ -1,0 +1,542 @@
+/*************************************************************
+    Resource change capture feature
+**************************************************************/
+
+/*************************************************************
+    Resource change data table
+**************************************************************/
+
+IF NOT EXISTS (
+    SELECT 1 
+    FROM sys.tables
+    WHERE name = 'ResourceChangeData')
+BEGIN
+    CREATE TABLE dbo.ResourceChangeData
+    (
+        Id bigint IDENTITY(1,1) NOT NULL,
+        Timestamp datetime2(7) NOT NULL,
+        ResourceId varchar(64) NOT NULL,
+        ResourceTypeId smallint NOT NULL,
+        ResourceVersion int NOT NULL,
+        ResourceChangeTypeId smallint NOT NULL,
+     CONSTRAINT PK_ResourceChangeData PRIMARY KEY CLUSTERED 
+    (
+        [Id] ASC
+    )WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, IGNORE_DUP_KEY = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON, OPTIMIZE_FOR_SEQUENTIAL_KEY = OFF) ON [PRIMARY]
+    ) ON [PRIMARY]
+END
+GO
+
+IF EXISTS (
+  SELECT 1 FROM information_schema.columns WHERE table_name = 'ResourceChangeData'
+and column_name = 'Timestamp'
+and Table_schema = 'dbo'
+and column_default is NULL)
+BEGIN
+     ALTER TABLE dbo.ResourceChangeData ADD  CONSTRAINT DF_ResourceChangeData_Timestamp  DEFAULT (sysutcdatetime()) FOR Timestamp
+END
+GO
+
+
+/*************************************************************
+    Resource change type table
+**************************************************************/
+
+IF NOT EXISTS (
+    SELECT 1 
+    FROM sys.tables
+    WHERE name = 'ResourceChangeType')
+BEGIN
+
+     CREATE TABLE dbo.ResourceChangeType(
+         ResourceChangeTypeId smallint IDENTITY(0,1) NOT NULL,
+         Name nvarchar(50) NOT NULL,
+      CONSTRAINT PK_ResourceChangeType PRIMARY KEY CLUSTERED 
+     (
+         ResourceChangeTypeId ASC
+     )WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, IGNORE_DUP_KEY = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON, OPTIMIZE_FOR_SEQUENTIAL_KEY = OFF) ON [PRIMARY],
+      CONSTRAINT UQ_ResourceChangeType_Name UNIQUE NONCLUSTERED 
+     (
+         Name ASC
+     )WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, IGNORE_DUP_KEY = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON, OPTIMIZE_FOR_SEQUENTIAL_KEY = OFF) ON [PRIMARY]
+     ) ON [PRIMARY]
+
+     SET IDENTITY_INSERT [dbo].[ResourceChangeType] ON 
+
+     INSERT dbo.ResourceChangeType (ResourceChangeTypeId, Name) VALUES (0, N'Creation')
+     INSERT dbo.ResourceChangeType (ResourceChangeTypeId, Name) VALUES (1, N'Update')
+     INSERT dbo.ResourceChangeType (ResourceChangeTypeId, Name) VALUES (2, N'Deletion')
+
+     SET IDENTITY_INSERT dbo.ResourceChangeType OFF
+END
+GO
+
+/*************************************************************
+    Stored procedures for capturing and fetching resource changes
+**************************************************************/
+--
+-- STORED PROCEDURE
+--     CaptureResourceChanges
+--
+-- DESCRIPTION
+--     Inserts resource change data
+--
+-- PARAMETERS
+--     @isDeleted
+--         * Whether this resource marks the resource as deleted.
+--     @version
+--         * The version of the resource being written
+--     @resourceId
+--         * The resource ID
+--     @resourceTypeId
+--         * The ID of the resource type
+--
+-- RETURN VALUE
+--     It does not return a value.
+--
+CREATE OR ALTER PROCEDURE dbo.CaptureResourceChanges
+    @isDeleted bit, 
+    @version int, 
+    @resourceId varchar(64),
+    @resourceTypeId smallint
+AS
+BEGIN
+    DECLARE @changeType SMALLINT
+    IF (@isDeleted = 1) BEGIN
+        SET @changeType = 2    /* DELETION */
+    END
+    ELSE BEGIN
+        IF (@version = 1) BEGIN
+            SET @changeType = 0 /* CREATION */
+        END
+        ELSE BEGIN
+            SET @changeType = 1 /* UPDATE */
+        END
+    END
+
+    INSERT INTO dbo.ResourceChangeData
+        (ResourceId, ResourceTypeId, ResourceVersion, ResourceChangeTypeId)
+    VALUES
+        (@resourceId, @resourceTypeId, @version, @changeType)
+END
+GO
+
+--
+-- STORED PROCEDURE
+--     FetchResourceChanges
+--
+-- DESCRIPTION
+--     Returns the number of resource change records from startId. The start id is inclusive.
+--
+-- PARAMETERS
+--     @startId
+--         * The start id of resource change records to fetch.
+--     @pageSize
+--         * The page size for fetching resource change records.
+--
+-- RETURN VALUE
+--     Resource change data rows.
+--
+CREATE OR ALTER PROCEDURE dbo.FetchResourceChanges
+    @startId bigint, 
+    @pageSize int
+AS
+BEGIN
+
+    SET NOCOUNT ON;
+
+    SET TRANSACTION ISOLATION LEVEL REPEATABLE READ;
+
+    SELECT TOP(@pageSize) c.Id
+      ,c.Timestamp
+      ,c.ResourceId
+      ,c.ResourceTypeId
+      ,t.Name AS ResourceTypeName
+      ,c.ResourceVersion
+      ,c.ResourceChangeTypeId
+      FROM dbo.ResourceChangeData c
+      LEFT JOIN dbo.ResourceType t ON c.ResourceTypeId = t.ResourceTypeId
+    WHERE c.Id >= @startId ORDER BY c.Id ASC
+END
+GO
+
+/*************************************************************
+    Stored procedures for creating and deleting
+**************************************************************/
+
+--
+-- STORED PROCEDURE
+--     UpsertResource_4
+--
+-- DESCRIPTION
+--     Creates or updates (including marking deleted) a FHIR resource
+--
+-- PARAMETERS
+--     @baseResourceSurrogateId
+--         * A bigint to which a value between [0, 80000) is added, forming a unique ResourceSurrogateId.
+--         * This value should be the current UTC datetime, truncated to millisecond precision, with its 100ns ticks component bitshifted left by 3.
+--     @resourceTypeId
+--         * The ID of the resource type (See ResourceType table)
+--     @resourceId
+--         * The resource ID (must be the same as the in the resource itself)
+--     @etag
+--         * If specified, the version of the resource to update
+--     @allowCreate
+--         * If false, an error is thrown if the resource does not already exist
+--     @isDeleted
+--         * Whether this resource marks the resource as deleted
+--     @keepHistory
+--         * Whether the existing version of the resource should be preserved
+--     @requestMethod
+--         * The HTTP method/verb used for the request
+--     @searchParamHash
+--          * A hash of the resource's latest indexed search parameters
+--     @rawResource
+--         * A compressed UTF16-encoded JSON document
+--     @resourceWriteClaims
+--         * Claims on the principal that performed the write
+--     @compartmentAssignments
+--         * Compartments that the resource is part of
+--     @referenceSearchParams
+--         * Extracted reference search params
+--     @tokenSearchParams
+--         * Extracted token search params
+--     @tokenTextSearchParams
+--         * The text representation of extracted token search params
+--     @stringSearchParams
+--         * Extracted string search params
+--     @numberSearchParams
+--         * Extracted number search params
+--     @quantitySearchParams
+--         * Extracted quantity search params
+--     @uriSearchParams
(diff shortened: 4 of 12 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.diff.sql
```diff
@@ -0,0 +1,540 @@
...
+-- PARAMETERS
+--     @startId
+--         * The start id of resource change records to fetch.
+--     @pageSize
+--         * The page size for fetching resource change records.
+--
+-- RETURN VALUE
+--     Resource change data rows.
+--
+CREATE OR ALTER PROCEDURE dbo.FetchResourceChangeData
```
ms-teli, line 138: Nit: can we rename to FetchResourceChanges to be in similar convention as CaptureResourceChanges

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.sql
```diff
@@ -0,0 +1,3443 @@
...
+    SELECT TOP(@pageSize) c.Id
+      ,c.Timestamp
+      ,c.ResourceId
+      ,c.ResourceTypeId
+      ,t.Name AS ResourceTypeName
+      ,c.ResourceVersion
+      ,c.ResourceChangeTypeId
+      FROM dbo.ResourceChangeData c
+      LEFT JOIN dbo.ResourceType t ON c.ResourceTypeId = t.ResourceTypeId
+    WHERE c.Id > @startId ORDER BY c.Id ASC
```
ms-teli, line 3440: If we wanted to use ">" instead of ">=" for better improvement, can we have 

```
SET @startId = @startId - 1
```

In most cases, startId should be included.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.diff.sql
```diff
@@ -0,0 +1,540 @@
...
+        [Id] ASC
+    )WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, IGNORE_DUP_KEY = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON, OPTIMIZE_FOR_SEQUENTIAL_KEY = OFF) ON [PRIMARY]
+    ) ON [PRIMARY]
+
+    ALTER TABLE dbo.ResourceChangeData ADD  CONSTRAINT DF_ResourceChangeData_Timestamp  DEFAULT (sysutcdatetime()) FOR Timestamp
+END
+GO
+
+IF EXISTS (
+  select 1 from information_schema.columns where table_name = 'ResourceChangeData'
```
ms-teli, line 33: Can we keep the letter case consistent?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.diff.sql
```diff
@@ -0,0 +1,540 @@
...
+
+IF NOT EXISTS (
+    SELECT 1 
+    FROM sys.tables
+    WHERE name = 'ResourceChangeType')
+BEGIN
+
+	 CREATE TABLE dbo.ResourceChangeType(
+		 ResourceChangeTypeId smallint IDENTITY(0,1) NOT NULL,
+		 Name nvarchar(50) NOT NULL,
```
ms-teli, line 55: Also unique?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.diff.sql
```diff
@@ -0,0 +1,540 @@
...
+--         * The start id of resource change records to fetch.
+--     @pageSize
+--         * The page size for fetching resource change records.
+--
+-- RETURN VALUE
+--     Resource change data rows.
+--
+CREATE OR ALTER PROCEDURE dbo.FetchResourceChangeData
+    @startId BIGINT, 
+    @pageSize BIGINT
```
ms-teli, line 140: nit: INT should be enough for the pageSize.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.diff.sql
```diff
@@ -0,0 +1,540 @@
...
+    SELECT TOP(@pageSize) c.Id
+      ,c.Timestamp
+      ,c.ResourceId
+      ,c.ResourceTypeId
+      ,t.Name AS ResourceTypeName
+      ,c.ResourceVersion
+      ,c.ResourceChangeTypeId
+      FROM dbo.ResourceChangeData c
+      LEFT JOIN dbo.ResourceType t ON c.ResourceTypeId = t.ResourceTypeId
+    WHERE c.Id > @startId ORDER BY c.Id ASC
```
ms-teli, line 157: Same here, I suggest to include @startId

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.diff.sql
```diff
@@ -0,0 +1,540 @@
...
+    INSERT INTO dbo.TokenNumberNumberCompositeSearchParam
+        (ResourceTypeId, ResourceSurrogateId, SearchParamId, SystemId1, Code1, SingleValue2, LowValue2, HighValue2, SingleValue3, LowValue3, HighValue3, HasRange, IsHistory)
+    SELECT DISTINCT @resourceTypeId, @resourceSurrogateId, SearchParamId, SystemId1, Code1, SingleValue2, LowValue2, HighValue2, SingleValue3, LowValue3, HighValue3, HasRange, 0
+    FROM @tokenNumberNumberCompositeSearchParams
+
+    SELECT @version
+
+	IF (@isResourceChangeCaptureEnabled) = 1 BEGIN
+        --If the capture resource changes feature is enabled, to execute a stored procedure called CaptureResourceChanges to insert resource change data.
+		EXEC dbo.CaptureResourceChanges @isDeleted=@isDeleted, @version=@version, @resourceId=@resourceId, @resourceTypeId=@resourceTypeId
```
ms-teli, line 536: Nit: should we reduce indention to align up with "--if"

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.sql
```diff
@@ -0,0 +1,3443 @@
...
+GO
+
+ALTER TABLE dbo.ResourceChangeData ADD  CONSTRAINT DF_ResourceChangeData_Timestamp  DEFAULT (sysutcdatetime()) FOR Timestamp
+GO
+
+/*************************************************************
+    Resource change type table
+**************************************************************/
+CREATE TABLE dbo.ResourceChangeType(
+	ResourceChangeTypeId smallint IDENTITY(0,1) NOT NULL,
```
ms-teli, line 3336: Nit: what's the convention for indention?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.sql
```diff
@@ -0,0 +1,3443 @@
...
+) ON [PRIMARY]
+GO
+
+SET IDENTITY_INSERT [dbo].[ResourceChangeType] ON 
+
+INSERT dbo.ResourceChangeType (ResourceChangeTypeId, Name) VALUES (0, N'Creation')
+INSERT dbo.ResourceChangeType (ResourceChangeTypeId, Name) VALUES (1, N'Update')
+INSERT dbo.ResourceChangeType (ResourceChangeTypeId, Name) VALUES (2, N'Deletion')
+
+SET IDENTITY_INSERT dbo.ResourceChangeType OFF
```
ms-teli, line 3351: This is great!

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.sql
```diff
@@ -0,0 +1,3443 @@
...
+
+    SELECT TOP(@pageSize) c.Id
+      ,c.Timestamp
+      ,c.ResourceId
+      ,c.ResourceTypeId
+      ,t.Name AS ResourceTypeName
+      ,c.ResourceVersion
+      ,c.ResourceChangeTypeId
+      FROM dbo.ResourceChangeData c
+      LEFT JOIN dbo.ResourceType t ON c.ResourceTypeId = t.ResourceTypeId
```
ms-teli, line 3439: I found a similar workflow pattern to handle the ResourceType, given that example, maybe we can do similar thing? https://github.com/microsoft/fhir-server/blob/ecf39a2888cb8449b1085867288153f3f8a8b607/src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs#L279

ms-teli (commented):
Overall looks good. Please see the questions inline.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.sql
bbatjargal: Replaced ">" with ">=". StartId is now inclusive.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.diff.sql
bbatjargal: Renamed as FetchResourceChanges.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.diff.sql
dustinburson: The start Id is the watermark correct?  If the event agent is always passing in the last saved watermark you will always return a record once bootstrapped and send the same recorded over and over.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.diff.sql
bbatjargal: Made keywords uppercase.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.diff.sql
bbatjargal: ResourceChangeTypeId is a primary key that guarantees uniqueness.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.diff.sql
bbatjargal: Yes, you're right. Changed it to int.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.diff.sql
bbatjargal: Fixed indentation.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.sql
bbatjargal: It is 4 spaces.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.sql
bbatjargal: Technically, yes we can use a similar approach. As we discussed early, calling two separate queries from PulibshEventsService, one for ResourceType select, and the second one for ResourceChangeData select. We can store resource types in the memory using a dictionary. Then, we can dynamically set a resource type name in the C# code.  

To go with this approach, we have to conduct a performance test. So, we can have enough information to compare both approaches. That would be more related to later tasks conducting performance tests and performance improvement.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.diff.sql
bbatjargal: As discussed in the chat, The start id is included like other pagination APIs. The event agent will keep tracking a watermark without duplication. A worker should pass @startId as lastProcessedId + 1.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.diff.sql
ms-teli: Sorry, I should be more clear, the comment was pointing to the "Name".

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.sql
ms-teli: I see. I'm fine with doing it at a later PR. I think the bottle neck overall will be the select query cause it's a recurring query. The query to get resource type into memory should be one time thing in the cold start, I would worry less comparing to left-join.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.diff.sql
bbatjargal: I see. No, the name is not unique.

ms-teli: I thought we could make it unique to make it less error prone.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.sql
```diff
@@ -0,0 +1,3443 @@
...
+--
+CREATE PROCEDURE dbo.FetchResourceChanges
+    @startId bigint, 
+    @pageSize int
+AS
+BEGIN
+
+    SET NOCOUNT ON;
+
+    SET TRANSACTION ISOLATION LEVEL REPEATABLE READ;
```
ms-teli, line 3429: Hi @brendankowitz, please be aware of this change here. 

we have this ResourceChangeData table that the FHIR server will write into and the Event Agent will read from. With this change, when this select query is performed, it will block other writes, currently the upserting sproc. We have a few backup plans to make it better for synchronization and concurrency handling, but those will need more cycles to prove and implement. 

Currently, the impact of this select should be so small, we project 200 items limit per query and 1 query per second, this literally should be pretty minimum impact, except some severe cases maybe like SQL connection issues. 

Also, the EventAgent should be the only consumer of this at this point, and EventAgent will only be turned on for customers who opt-in the event Private Preview. Before Public Preview, we will re-evaluate this part and review with FHIR team again.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.sql
```diff
@@ -0,0 +1,3443 @@
...
+        (@resourceId, @resourceTypeId, @version, @changeType)
+END
+GO
+
+--
+-- STORED PROCEDURE
+--     FetchResourceChanges
+--
+-- DESCRIPTION
+--     Returns the number of resource change records from startId
```
ms-teli, line 3410: Nit: can we explicitly state that the startId will be included? :)

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.diff.sql
bbatjargal: Added a unique constraint.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.sql
bbatjargal: Updated.

src/Microsoft.Health.Fhir.Core/Configs/CoreFeatureConfiguration.cs
```diff
@@ -56,5 +56,10 @@ public class CoreFeatureConfiguration
         /// Maximum items allowed to be deleted when using Conditional Delete.
         /// </summary>
         public int ConditionalDeleteMaxItems { get; set; } = 100;
+
+        /// <summary>
+        /// Gets or sets a value whether capturing resource change data is enabled or not.
+        /// </summary>
+        public bool SupportsResourceChangeCapture { get; set; } = false;
```
brendankowitz, line 63: I actually think this should be tied to the DataStoreConfig since its is only planned as a SQL feature. Currently we use it straight from https://github.com/microsoft/healthcare-shared-components/blob/master/src/Microsoft.Health.SqlServer/Configs/SqlServerDataStoreConfiguration.cs

bbatjargal: From the SQL feature perspective, the SqlServerDataStoreConfiguration class might be a suitable place to have the property. But, this config is general and indicates whether an event feature is enabled in the Fhir Application. Also, it will be used to toggle the Cosmos DB's Change Feed feature. We might rename it to make it more generic. What you do think @ms-teli?

Ivanidzo4ka:
Hi.
I've read spec and watch the video from Wednesday, and I have question.

Only mention of transactional nature of events was "because we want it". Is it actual requirement? Main reason why I'm asking about this, because this change would introduce another load on our database, and in current architecture database is basically our bottleneck. And introducing writes and reads on a database would only make it slower.

If there is requirement 100% changes would have events I understand. But if you willing to have 99.999, I would rather had this functionality in separate database and had an interface to which I can save that, and it's up to the event agent to figure out what to do with it. Basically same what we do with logging. There is interface, you can define who would subscribe to that interface in DI, and if it's an event agent, it's up to event agent how to preserve and push notifications (you can use disk, you can use sqllite, you can use outside db, you can use redis). That would also make it database independent feature (while current one is sql specific)

But if we want to go this route, why not [CDC](https://docs.microsoft.com/en-us/sql/relational-databases/track-changes/about-change-data-capture-sql-server?view=sql-server-ver15)? There is already implementation (it's in preview for azure sql, but it exist), and I think reading from log, rather than creating extra tables is better option performance wise.

ms-teli:
> Hi.
> I've read spec and watch the video from Wednesday, and I have question.
> 
> Only mention of transactional nature of events was "because we want it". Is it actual requirement? Main reason why I'm asking about this, because this change would introduce another load on our database, and in current architecture database is basically our bottleneck. And introducing writes and reads on a database would only make it slower.
> 
> If there is requirement 100% changes would have events I understand. But if you willing to have 99.999, I would rather had this functionality in separate database and had an interface to which I can save that, and it's up to the event agent to figure out what to do with it. Basically same what we do with logging. There is interface, you can define who would subscribe to that interface in DI, and if it's an event agent, it's up to event agent how to preserve and push notifications (you can use disk, you can use sqllite, you can use outside db, you can use redis). That would also make it database independent feature (while current one is sql specific)
> 
> But if we want to go this route, why not [CDC](https://docs.microsoft.com/en-us/sql/relational-databases/track-changes/about-change-data-capture-sql-server?view=sql-server-ver15)? There is already implementation (it's in preview for azure sql, but it exist), and I think reading from log, rather than creating extra tables is better option performance wise.

Thanks for

src/Microsoft.Health.Fhir.Core/Configs/CoreFeatureConfiguration.cs
ms-teli: @brendankowitz  There are some parallel discussion going on to leverage the event source pattern for more business requirements, like support change feed. The PM can give more info. Thus why we decide to make it more generic.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
```diff
@@ -0,0 +1,542 @@
...
+     (
+         ResourceChangeTypeId ASC
+     )WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, IGNORE_DUP_KEY = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON, OPTIMIZE_FOR_SEQUENTIAL_KEY = OFF) ON [PRIMARY],
+      CONSTRAINT UQ_ResourceChangeType_Name UNIQUE NONCLUSTERED 
+     (
+         Name ASC
+     )WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, IGNORE_DUP_KEY = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON, OPTIMIZE_FOR_SEQUENTIAL_KEY = OFF) ON [PRIMARY]
+     ) ON [PRIMARY]
+
+     SET IDENTITY_INSERT [dbo].[ResourceChangeType] ON
```
cunninghamjc, line 64: What's the reason for making ResourceChangeTypeId an identity column but then inserting ids directly anyway?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
```diff
@@ -0,0 +1,542 @@
...
+
+    SELECT TOP(@pageSize) c.Id
+      ,c.Timestamp
+      ,c.ResourceId
+      ,c.ResourceTypeId
+      ,t.Name AS ResourceTypeName
+      ,c.ResourceVersion
+      ,c.ResourceChangeTypeId
+      FROM dbo.ResourceChangeData c
+      LEFT JOIN dbo.ResourceType t ON c.ResourceTypeId = t.ResourceTypeId
```
cunninghamjc, line 158: What will we do if ResourceTypeId is missing from dbo.ResourceType?  Is that an expected scenario?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
```diff
@@ -0,0 +1,542 @@
...
+    WHERE c.Id >= @startId ORDER BY c.Id ASC
```
cunninghamjc, line 159: Is it possible for some events to fail to be delivered but others to succeed?  What happens then?  Do we block all processing when we hit a failure ?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
bbatjargal: ResourceChangeTypeId is the identity and primary key for preventing duplications. Wanted to specify explicitly that 0 for creation, 1 for the update, and 2 for deletion.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
```diff
@@ -0,0 +1,542 @@
...
+    WHERE name = 'ResourceChangeData')
+BEGIN
+    CREATE TABLE dbo.ResourceChangeData
+    (
+        Id bigint IDENTITY(1,1) NOT NULL,
+        Timestamp datetime2(7) NOT NULL,
+        ResourceId varchar(64) NOT NULL,
+        ResourceTypeId smallint NOT NULL,
+        ResourceVersion int NOT NULL,
+        ResourceChangeTypeId smallint NOT NULL,
```
cunninghamjc, line 21: Why smallint for the change type id?  Do we expect more than 256 change types in the future?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
bbatjargal: The dbo.ResourceType is a reference table. Unlikely ResourceTypeId will be missing. But, if it is, the resource type name will be empty on the event payload.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
bbatjargal: We submit the change feed to the Event Grid API endpoint and ensure that a set of changes are reached in the Event Grid. A particular event is delivered to the subscriber or not is out of the scope. But, Event Grid ensures reliable delivery of events to the destination Endpoints.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
cunninghamjc: But why does it need to be identity?  Just being a PK will prevent duplicates.  Are we planning to insert new rows in the future without specifying the id?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
cunninghamjc: Are we going to document that behavior and what customers should do in that case?  The left join implies that it's possible for the resource type name to be missing.  If we don't intend for that to happen, then perhaps something should prevent/handle it.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
cunninghamjc: Can the call to Event Grid succeed for some events but not others?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
```diff
@@ -0,0 +1,542 @@
...
+--
+CREATE OR ALTER PROCEDURE dbo.FetchResourceChanges
+    @startId bigint, 
+    @pageSize int
+AS
+BEGIN
+
+    SET NOCOUNT ON;
+
+    SET TRANSACTION ISOLATION LEVEL REPEATABLE READ;
```
Ivanidzo4ka, line 148: >SET TRANSACTION ISOLATION LEVEL REPEATABLE READ; [](http://example.com/codeflow?start=4&length=48)

Please add comment about this level of isolation and reason for it.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
bbatjargal: No, that can not happen.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
bbatjargal: Let me take back my last statement. Given the fact that the resource type is one of the important factors in the Fhir application and the resource change data is inserted during upsertResource transactions, resource type id will be a valid value. And, to ensure to have a valid resource type name on even payload, adding logic in the publish event service to check if resource name and resource Id are not empty. If they are empty, log exceptions to investigate a problem, and skip that record to send an event since the record is invalid.

cunninghamjc: Makes sense, thanks.  Since the ResourceType table rarely changes, another option is to read the table once into memory in the eventing processor, map the id to the name there, and skip the join here in the sproc.  Thoughts?  We would need to make sure the table is reloaded if the FHIR service adds new resource types.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
bbatjargal: No, not expecting that. Changed to tinyint.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
bbatjargal: Sure, added comments.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
```diff
@@ -0,0 +1,550 @@
...
+    -- due to interleaved transactions on the resource change data table.
+    SET TRANSACTION ISOLATION LEVEL REPEATABLE READ;
+    
+    -- In Repeatable Read, the select query execution will be blocked until other open transactions are completed
+    -- for rows that match the search condition of the select statement. 
+    -- A write transaction (update/delete) on the rows that match 
+    -- the search condition of the select statement will wait until the read transaction is completed. 
+    -- But, other transactions can insert new rows.
+    SELECT TOP(@pageSize) c.Id
+      ,c.Timestamp
```
namadabu, line 159: nit: is there a reason the commas are at the beginning of the line (instead of at the end which is the usual style we follow)?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
```diff
@@ -0,0 +1,550 @@
...
+        Timestamp datetime2(7) NOT NULL,
+        ResourceId varchar(64) NOT NULL,
+        ResourceTypeId smallint NOT NULL,
+        ResourceVersion int NOT NULL,
+        ResourceChangeTypeId tinyint NOT NULL,
+     CONSTRAINT PK_ResourceChangeData PRIMARY KEY CLUSTERED 
+    (
+        [Id] ASC
+    )WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, IGNORE_DUP_KEY = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON, OPTIMIZE_FOR_SEQUENTIAL_KEY = OFF) ON [PRIMARY]
+    ) ON [PRIMARY]
```
namadabu, line 26: nit: I feel that having consistent indentation and style would make it easier to read the code (this file and 14.sql file as well). We also have a style guide (that is linked to at the top of 14.sql) and it might be helpful to follow that.

A couple of examples: this block ends with two ON [PRIMARY] sentences and it's not very intuitive to what block they apply to. In other places, sometimes `(` starts on the same line, in other cases it starts on the next line.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.sql
```diff
@@ -0,0 +1,3458 @@
...
+--     It does not return a value.
+--
+CREATE PROCEDURE dbo.CaptureResourceChanges
+    @isDeleted bit, 
+    @version int, 
+    @resourceId varchar(64),
+    @resourceTypeId smallint
+AS
+BEGIN
+    DECLARE @changeType SMALLINT
```
rbans96, line 3393: Do we need to begin transaction here, since while fetching resourceChanges, we are setting transaction isolation to REPEATABLE READ?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -176,7 +180,24 @@ private void PopulateUpsertResourceCommand(SqlCommandWrapper sqlCommandWrapper,
                     requestMethod: resource.Request.Method,
                     searchParamHash: resource.SearchParameterHash,
                     rawResource: stream,
-                    tableValuedParameters: _upsertResourceTvpGeneratorVLatest.Generate(new List<ResourceWrapper> { resource }));
+                    tableValuedParameters: _upsertResourceTvpGeneratorVLatest.Generate(new List<ResourceWrapper> { resource }),
+                    isResourceChangeCaptureEnabled: isResourceChangeCaptureEnabled);
+            }
+            else if (_schemaInformation.Current >= SchemaVersionConstants.SearchParameterHashSchemaVersion)
```
rbans96, line 186: nit: Should it just be equality operator (schemaInformation.Current == SchemaVersionConstants.SearchParameterHashSchemaVersion)
since it seems this block will only be executed at schema version: 13.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
bbatjargal: Good point. No, not planning to insert new rows without specifying the id. Removed the IDENTITY property.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
bbatjargal: Updated the commas.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
bbatjargal: Updated the style.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
```diff
@@ -0,0 +1,550 @@
...
+END
+GO
+
+IF EXISTS (
+  SELECT 1 FROM information_schema.columns WHERE table_name = 'ResourceChangeData'
+and column_name = 'Timestamp'
+and Table_schema = 'dbo'
+and column_default is NULL)
+BEGIN
+     ALTER TABLE dbo.ResourceChangeData ADD  CONSTRAINT DF_ResourceChangeData_Timestamp  DEFAULT (sysutcdatetime()) FOR Timestamp
```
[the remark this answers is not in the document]
bbatjargal: Included the constraint in the table creation.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
```diff
@@ -0,0 +1,550 @@
...
+     )WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, IGNORE_DUP_KEY = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON, OPTIMIZE_FOR_SEQUENTIAL_KEY = OFF) ON [PRIMARY],
+      CONSTRAINT UQ_ResourceChangeType_Name UNIQUE NONCLUSTERED 
+     (
+         Name ASC
+     )WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, IGNORE_DUP_KEY = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON, OPTIMIZE_FOR_SEQUENTIAL_KEY = OFF) ON [PRIMARY]
+     ) ON [PRIMARY]
+
+     SET IDENTITY_INSERT [dbo].[ResourceChangeType] ON 
+
+     INSERT dbo.ResourceChangeType (ResourceChangeTypeId, Name) VALUES (0, N'Creation')
```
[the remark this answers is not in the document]
bbatjargal: Made each insert idempotent.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
```diff
@@ -0,0 +1,550 @@
...
+    WHERE name = 'ResourceChangeType')
+BEGIN
+
+     CREATE TABLE dbo.ResourceChangeType(
+         ResourceChangeTypeId tinyint IDENTITY(0,1) NOT NULL,
+         Name nvarchar(50) NOT NULL,
+      CONSTRAINT PK_ResourceChangeType PRIMARY KEY CLUSTERED 
+     (
+         ResourceChangeTypeId ASC
+     )WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, IGNORE_DUP_KEY = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON, OPTIMIZE_FOR_SEQUENTIAL_KEY = OFF) ON [PRIMARY],
```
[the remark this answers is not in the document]
bbatjargal: Good point. No, there is no explicit reason, and using default configs. Removed, and simplified the constraint creation.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.sql
```diff
@@ -0,0 +1,3458 @@
...
+/*************************************************************
+    Resource change type table
+**************************************************************/
+CREATE TABLE dbo.ResourceChangeType(
+    ResourceChangeTypeId tinyint IDENTITY(0,1) NOT NULL,
+    Name nvarchar(50) NOT NULL,
+ CONSTRAINT PK_ResourceChangeType PRIMARY KEY CLUSTERED 
+(
+    ResourceChangeTypeId ASC
+)WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, IGNORE_DUP_KEY = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON, OPTIMIZE_FOR_SEQUENTIAL_KEY = OFF) ON [PRIMARY],
```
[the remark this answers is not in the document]
bbatjargal: Removed and simplified the constraint creation.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.sql
bbatjargal: It is not required. It is going to be a part of the upsertResource transaction in RCSI. Repeatable Read isolation level does not block other insert transactions.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
bbatjargal: No, it doesn't. the SchemaVersionConstants.SearchParameterHashSchemaVersion indicates SchemaVersion.V8.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
bbatjargal: We had a discussion about caching resource types at the startup of an event agent service. Based on the fact that resource types are not changed during runtime, removing the left join of the resource change data with a resource type table from the select statement to simplify the query execution plan.

ms-teli (approved):

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
```diff
@@ -0,0 +1,537 @@
...
+-- PARAMETERS
+--     @startId
+--         * The start id of resource change records to fetch.
+--     @pageSize
+--         * The page size for fetching resource change records.
+--
+-- RETURN VALUE
+--     Resource change data rows.
+--
+CREATE OR ALTER PROCEDURE dbo.FetchResourceChanges
```
rotodd, line 129: For my understanding, where will this procedure be used?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
```diff
@@ -0,0 +1,537 @@
...
+-- RETURN VALUE
+--     It does not return a value.
+--
+CREATE OR ALTER PROCEDURE dbo.CaptureResourceChanges
+    @isDeleted bit, 
+    @version int, 
+    @resourceId varchar(64),
+    @resourceTypeId smallint
+AS
+BEGIN
```
rotodd, line 92: Nit: it might be nice to add a comment explaining that this procedure is intended to be called from the `UpsertResource_4` stored procedure (which is why it doesn't begin a new transaction)

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
```diff
@@ -0,0 +1,537 @@
...
+    @pageSize int
+AS
+BEGIN
+
+    SET NOCOUNT ON;
+
+    -- Given the fact that Read Committed Snapshot isolation level is enabled on the FHIR database, 
+    -- using the Repeatable Read isolation level to avoid skipping resource changes 
+    -- due to interleaved transactions on the resource change data table.
+    SET TRANSACTION ISOLATION LEVEL REPEATABLE READ;
```
rotodd, line 140: So that I understand, would using the `UPDLOCK` hint also work in this scenario? If so, why is it preferable to instead change the isolation level?

rotodd, line 140: Will the isolation level return to read committed after the current transaction is committed, or do we need to manually set it back?

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlServerSchemaUpgradeTests.cs
```diff
@@ -181,6 +181,7 @@ private bool CompareDatabaseSchemas(string databaseName1, string databaseName2)
             {
                 ("Procedure", "[dbo].[UpsertResource]"),
                 ("Procedure", "[dbo].[UpsertResource_2]"),
+                ("Procedure", "[dbo].[UpsertResource_3]"),
```
rotodd, line 184: Nice catch :)

src/Microsoft.Health.Fhir.Core/Configs/CoreFeatureConfiguration.cs
Ivanidzo4ka: I would move it to sql configuration as well.

If there is need to support this on all persistence levels we can always move it here, but that's in future.

Ivanidzo4ka (approved):
:shipit:

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
```diff
@@ -0,0 +1,537 @@
...
+--         * The start id of resource change records to fetch.
+--     @pageSize
+--         * The page size for fetching resource change records.
+--
+-- RETURN VALUE
+--     Resource change data rows.
+--
+CREATE OR ALTER PROCEDURE dbo.FetchResourceChanges
+    @startId bigint, 
+    @pageSize int
```
Ivanidzo4ka, line 131: >int [](http://example.com/codeflow?start=14&length=3)

nit: I would go with small int.

I doubt you need to read more than 30k records once.

Ivanidzo4ka, line 131: I would add `@DataCount int OUTPUT` parameter and set it to @@ROWCOUNT value, so you always know how much data you getting back.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
bbatjargal: It will be called from the Event Agent service.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
bbatjargal: Good point. But, there is no business logic dependent on row counts, and don't see many benefits at this moment. For simplicity, not adding the output parameter. We can use the size of the list in C# if needed.

Ivanidzo4ka: It let you know how big of collection you expect back, so you don't have to reallocate memory for list resize during reading of collection. It's not business driven, but it's for performance, 

I'm talking instead of:
```
var list = new List<Class>();
while (reader.Read()){
list.Add (new Class())
}
```
you can do

```
var list = new List<Class>((int)command.Parameter["DataCount"]);
while (reader.Read()){
list.Add (new Class())
}
```

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
bbatjargal: Added a comment.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
bbatjargal: Good point. Changed to smallint.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/14.diff.sql
bbatjargal: Good point. Thanks. Changed it to a Repeatable Read table hint instead of setting a Repeatable Read isolation level. Based on my research, using a repeatable read table hint is more suitable for our scenario due to the Updlock table hint acquires an [update (U) lock](https://docs.microsoft.com/en-us/previous-versions/sql/sql-server-2008-r2/ms175519(v=sql.105)#update-locks) on the resource which avoids a potential deadlock problem but blocks other read transactions for the same resource until the transaction completes. On the other hand, the Repeatable Read hint acquires a [shared (S) lock](https://docs.microsoft.com/en-us/previous-versions/sql/sql-server-2008-r2/ms175519(v=sql.105)#shared-locks) which allows other transactions to read the same resource in concurrent transactions. In our case, we insert and read often, don't update the resource change data, and don't delete the resource at the same as reading the resource. So,  I think that shared (S) locks on a resource are appropriate for this scenario.

src/Microsoft.Health.Fhir.Core/Configs/CoreFeatureConfiguration.cs
bbatjargal: I think that the SupportsResourceChangeCapture config is an application-level configuration. If we move the config to the SqlServerDataStoreConfiguration class based on the SQL feature perspective like CDC, we need another config value that indicates whether an event feature is enabled in the FhirApplication in the appsetting.json. It leads to an issue managing two config values. For simplicity, we are using one application-level config which also toggles a SQL level feature.

Ivanidzo4ka: If it's a application level feature, why it works only for SQL and doesn't do anything for Cosmos?

If you move to CDC I expect complete rewrite of current functionality and drop all sql changes made in this PR.

If you want to have grace period, and switch back and forth, yes, it need second feature flag which would indicate implmentation, or you can always change type of this feauture from boolen to enum: `None,CDC,CustomTable`

PR #2035, opened 2021-06-17
Enable concurrent query execution if sequential execution timed out

rbans96:
## Description
Since sequential query execution is taking longer for selective queries, we intend to enable parallel query execution as soon as the sequential query execution time exceeds the time limit(specified time limit as 5 secs).
And then we start querying concurrently across 500(or whatever value is set to MaxQueryConcurrency) physical partitions.

## Related issues
Addresses [issue #].

## Testing

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.CosmosDb/Configs/CosmosDataStoreParallelQueryOptions.cs
+++ b/src/Microsoft.Health.Fhir.CosmosDb/Configs/CosmosDataStoreParallelQueryOptions.cs
@@ -3,6 +3,8 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
+using System;
+
 namespace Microsoft.Health.Fhir.CosmosDb.Configs
 {
     public class CosmosDataStoreParallelQueryOptions
@@ -10,7 +12,7 @@ namespace Microsoft.Health.Fhir.CosmosDb.Configs
         /// <summary>
         /// Gets the maximum degree of parallelism for the SDK to use when querying physical partitions in parallel.
         /// </summary>
-        public int MaxQueryConcurrency { get; set; }
+        public int MaxQueryConcurrency { get; set; } = Environment.ProcessorCount * 10;
 
         /// <summary>
         /// Enables the parallelism if sequential query execution takes more time
--- a/src/Microsoft.Health.Fhir.CosmosDb/Configs/CosmosDataStoreParallelQueryOptions.cs
+++ b/src/Microsoft.Health.Fhir.CosmosDb/Configs/CosmosDataStoreParallelQueryOptions.cs
@@ -11,5 +11,10 @@ namespace Microsoft.Health.Fhir.CosmosDb.Configs
         /// Gets the maximum degree of parallelism for the SDK to use when querying physical partitions in parallel.
         /// </summary>
         public int MaxQueryConcurrency { get; set; }
+
+        /// <summary>
+        /// Enables the parallelism if sequential query execution takes more time
+        /// </summary>
+        public bool EnableConcurrencyIfQueryExceedsTimeLimit { get; set; }
     }
 }
--- a/src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
+++ b/src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
@@ -429,13 +429,34 @@ namespace Microsoft.Health.Fhir.CosmosDb.Features.Search
 
             try
             {
-                (IReadOnlyList<T> results, string nextContinuationToken) = await _fhirDataStore.ExecuteDocumentQueryAsync<T>(sqlQuerySpec, feedOptions, continuationToken, searchOptions.MaxItemCountSpecifiedByClient, feedOptions.MaxConcurrency == null ? searchEnumerationTimeoutOverrideIfSequential : null, cancellationToken);
+                (IReadOnlyList<T> results, string nextContinuationToken) = (null, null);
+                var desiredItemCount = feedOptions.MaxItemCount * CosmosFhirDataStore.ExecuteDocumentQueryAsyncMinimumFillFactor;
+
+                // Set timeout for sequential query execution
+                if (feedOptions.MaxConcurrency == null &&
+                    _cosmosConfig.ParallelQueryOptions.EnableConcurrencyIfQueryExceedsTimeLimit == true &&
+                    string.IsNullOrEmpty(searchOptions.ContinuationToken))
+                {
+                    searchEnumerationTimeoutOverrideIfSequential = TimeSpan.FromSeconds(5);
+
+                    // Executing query sequentially until the timeout
+                    (results, nextContinuationToken) = await _fhirDataStore.ExecuteDocumentQueryAsync<T>(sqlQuerySpec, feedOptions, continuationToken, searchOptions.MaxItemCountSpecifiedByClient, searchEnumerationTimeoutOverrideIfSequential, cancellationToken);
+
+                    // check if we need to restart the query in parallel
+                    if (results.Count < desiredItemCount && !string.IsNullOrEmpty(nextContinuationToken))
+                    {
+                        feedOptions.MaxConcurrency = _cosmosConfig.ParallelQueryOptions.MaxQueryConcurrency;
+                        (results, nextContinuationToken) = await _fhirDataStore.ExecuteDocumentQueryAsync<T>(sqlQuerySpec, feedOptions, continuationToken, searchOptions.MaxItemCountSpecifiedByClient, null, cancellationToken);
+                    }
+                }
+                else
+                {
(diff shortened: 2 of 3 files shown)


src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
```diff
@@ -429,13 +429,34 @@ private async Task<Expression> RecurseChainedExpression(ChainedExpression expres
...
+                    searchEnumerationTimeoutOverrideIfSequential = TimeSpan.FromSeconds(5);
+
+                    // Executing query sequentially until the timeout
+                    (results, nextContinuationToken) = await _fhirDataStore.ExecuteDocumentQueryAsync<T>(sqlQuerySpec, feedOptions, continuationToken, searchOptions.MaxItemCountSpecifiedByClient, searchEnumerationTimeoutOverrideIfSequential, cancellationToken);
 
+                    // check if we need to restart the query in parallel
                     if (results.Count < desiredItemCount && !string.IsNullOrEmpty(nextContinuationToken))
+                    {
+                        feedOptions.MaxConcurrency = _cosmosConfig.ParallelQueryOptions.MaxQueryConcurrency;
+                        (results, nextContinuationToken) = await _fhirDataStore.ExecuteDocumentQueryAsync<T>(sqlQuerySpec, feedOptions, continuationToken, searchOptions.MaxItemCountSpecifiedByClient, null, cancellationToken);
```
brendankowitz, line 449: Observations:

1. We try sequentially for 5 seconds and give up
2. We reissue the query with MaxParallelism, resetting the initial resultset

In the case that (2) doesn't return the desired number of results, what happens in the cases of:

- The execution with MaxParallelism returns less than the desired number of results 
- If the first statement had `searchOptions.MaxItemCountSpecifiedByClient = true` and used the previous page's continuation token, would we execute in parallel unnecessary?

rbans96: I think in this case searchOptions.ContinuationToken would not be null, so it won't be re-executed.

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
```diff
@@ -429,13 +429,34 @@ private async Task<Expression> RecurseChainedExpression(ChainedExpression expres
...
             {
-                (IReadOnlyList<T> results, string nextContinuationToken) = await _fhirDataStore.ExecuteDocumentQueryAsync<T>(sqlQuerySpec, feedOptions, continuationToken, searchOptions.MaxItemCountSpecifiedByClient, feedOptions.MaxConcurrency == null ? searchEnumerationTimeoutOverrideIfSequential : null, cancellationToken);
+                (IReadOnlyList<T> results, string nextContinuationToken) = (null, null);
+                var desiredItemCount = feedOptions.MaxItemCount * CosmosFhirDataStore.ExecuteDocumentQueryAsyncMinimumFillFactor;
 
-                if (queryPartitionStatistics != null && messagesList != null)
+                // Set timeout for sequential query execution
+                if (feedOptions.MaxConcurrency == null &&
+                    _cosmosConfig.ParallelQueryOptions.EnableConcurrencyIfQueryExceedsTimeLimit == true &&
+                    string.IsNullOrEmpty(searchOptions.ContinuationToken))
```
brendankowitz, line 438: Do we need to check searchEnumerationTimeoutOverrideIfSequential also?
```suggestion
                if (feedOptions.MaxConcurrency == null &&
                    _cosmosConfig.ParallelQueryOptions.EnableConcurrencyIfQueryExceedsTimeLimit == true &&
                    string.IsNullOrEmpty(searchOptions.ContinuationToken) &&
                    searchEnumerationTimeoutOverrideIfSequential == null)
```

rbans96:
\azp run

src/Microsoft.Health.Fhir.CosmosDb/Configs/CosmosDataStoreParallelQueryOptions.cs
```diff
@@ -11,5 +11,10 @@ public class CosmosDataStoreParallelQueryOptions
         /// Gets the maximum degree of parallelism for the SDK to use when querying physical partitions in parallel.
         /// </summary>
         public int MaxQueryConcurrency { get; set; }
```
brendankowitz, line 13: After seeing this in the Cosmos code, thoughts on using a default based on the processor count? 

```suggestion
        public int MaxQueryConcurrency { get; set; } = Environment.ProcessorCount * 10;
```

/cc @Ivanidzo4ka also your thoughts?

Ivanidzo4ka: Sure, why not. TBH I'm not sure why fanning out queries happening on our side and not cosmos side, but if that's the case, why not.

rbans96: ok thanks, I am setting this to default. But just to mention we are overriding this with value 500 :)

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
```diff
@@ -429,13 +429,35 @@ private async Task<Expression> RecurseChainedExpression(ChainedExpression expres
...
-
-                if (queryPartitionStatistics != null && messagesList != null)
+                (IReadOnlyList<T> results, string nextContinuationToken) = (null, null);
+                var desiredItemCount = feedOptions.MaxItemCount * CosmosFhirDataStore.ExecuteDocumentQueryAsyncMinimumFillFactor;
+
+                // Set timeout for sequential query execution
+                if (feedOptions.MaxConcurrency == null &&
+                    _cosmosConfig.ParallelQueryOptions.EnableConcurrencyIfQueryExceedsTimeLimit == true &&
+                    string.IsNullOrEmpty(searchOptions.ContinuationToken) &&
+                    searchEnumerationTimeoutOverrideIfSequential == null)
```
brendankowitz, line 439: While testing I also realized that `continuationToken` is being passed into this function, this option is also only available when this is null.

```suggestion
                if (feedOptions.MaxConcurrency == null &&
                    _cosmosConfig.ParallelQueryOptions.EnableConcurrencyIfQueryExceedsTimeLimit == true &&
                    string.IsNullOrEmpty(searchOptions.ContinuationToken) &&
                    string.IsNullOrEmpty(continuationToken) &&
                    searchEnumerationTimeoutOverrideIfSequential == null)
```

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
```diff
@@ -429,13 +429,35 @@ private async Task<Expression> RecurseChainedExpression(ChainedExpression expres
...
                 {
-                    var desiredItemCount = feedOptions.MaxItemCount * CosmosFhirDataStore.ExecuteDocumentQueryAsyncMinimumFillFactor;
+                    searchEnumerationTimeoutOverrideIfSequential = TimeSpan.FromSeconds(5);
+
+                    // Executing query sequentially until the timeout
+                    (results, nextContinuationToken) = await _fhirDataStore.ExecuteDocumentQueryAsync<T>(sqlQuerySpec, feedOptions, continuationToken, searchOptions.MaxItemCountSpecifiedByClient, searchEnumerationTimeoutOverrideIfSequential, cancellationToken);
```
brendankowitz, line 444:
```suggestion
                    (results, nextContinuationToken) = await _fhirDataStore.ExecuteDocumentQueryAsync<T>(sqlQuerySpec, feedOptions, null, searchOptions.MaxItemCountSpecifiedByClient, searchEnumerationTimeoutOverrideIfSequential, cancellationToken);
```

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
brendankowitz:
```suggestion
                        (results, nextContinuationToken) = await _fhirDataStore.ExecuteDocumentQueryAsync<T>(sqlQuerySpec, feedOptions, null, searchOptions.MaxItemCountSpecifiedByClient, null, cancellationToken);
```

src/Microsoft.Health.Fhir.CosmosDb/Configs/CosmosDataStoreParallelQueryOptions.cs
brendankowitz: @Ivanidzo4ka this setting is the number of partitions they read from which == the number of Parallel tasks they create in the SDK :)

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
rbans96: Thanks! Done

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
rbans96: Thanks! Done

Ivanidzo4ka (approved):
:shipit:

Issue #2026, opened 2021-06-15
Bundle.response missing required field "status"


mosiermt:
**Describe the bug**
According to the FHIR R4 documentation for a Bundle ( https://www.hl7.org/fhir/bundle.html ), if a response object is provided, it Must have the "status" string representing the status code of the response. When a bundle is returned in the paas, it does not return this field. 

**FHIR Version?**
R4

**Data provider?**
CosmosDB

**To Reproduce**
1. Send a search-type, history-type, or history-instance request to the PaaS with or without search parameters
2. Inspect items in Bundle.entry.*.response
3. Status field is seen as not present

**Expected behavior**
This field should represent the status code returned by that particular response with the text of the status code optionally provided (i.e. "401 Unauthorized" or "200")

**Actual behavior**
The actual result only includes an "etag" and "lastModified" field in the response object.

CaitlinV39:
AB#82981

PR #2045, opened 2021-06-22
Added status to bundle response

apurvabhaleMS:
## Description
Added status for bundle.Resource.Response while returning history bundle. This status depends on  bundle.Resource.Request method httpverb.

## Related issues
Addresses [issue #2026]. 

## Testing
Added Unit tests and also did some manually testing

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/BundleFactory.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/BundleFactory.cs
@@ -72,14 +72,14 @@ namespace Microsoft.Health.Fhir.Core.Features.Search
                         statusString = ((int)HttpStatusCode.Created).ToString() + " " + HttpStatusCode.Created;
                         break;
                     case Bundle.HTTPVerb.PUT:
+                    case Bundle.HTTPVerb.GET:
                         statusString = ((int)HttpStatusCode.OK).ToString() + " " + HttpStatusCode.OK;
                         break;
                     case Bundle.HTTPVerb.DELETE:
                         statusString = ((int)HttpStatusCode.NoContent).ToString() + " " + HttpStatusCode.NoContent;
                         break;
                     default:
-                        statusString = ((int)HttpStatusCode.OK).ToString() + " " + HttpStatusCode.OK;
-                        break;
+                        throw new NotImplementedException();
                 }
 
                 resource.Response = new Bundle.ResponseComponent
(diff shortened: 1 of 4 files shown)


test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/HistoryTests.cs
```diff
@@ -83,6 +83,34 @@ public async Task GivenNoType_WhenGettingSystemHistory_TheServerShouldReturnTheA
             Assert.NotEmpty(readResponse.Resource.Entry);
         }
 
+        [Fact]
+        public async Task GivenAType_WhenGettingResourceHistory_TheServerShouldReturnTheAppropriateBundleSuccessfullyWithResponseStatus()
```
Ivanidzo4ka, line 87: >GivenAType_WhenGettingResourceHistory_TheServerShouldReturnTheAppropriateBundleSuccessfullyWithResponseStatus [](http://example.com/codeflow?start=26&length=109)

>[http://hl7.org/fhir/bundle.html](http://hl7.org/fhir/bundle.html)

response: Results of execution (transaction/batch/history)

It's good to have history test, but can we also have test for transaction and batch actions?

You can find them by `BatchTest` and `TransactionTest`

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/BundleFactory.cs
```diff
@@ -63,8 +64,27 @@ public ResourceElement CreateHistoryBundle(SearchResult result)
...
+                        statusString = ((int)HttpStatusCode.Created).ToString() + " " + HttpStatusCode.Created;
+                        break;
+                    case Bundle.HTTPVerb.PUT:
+                        statusString = ((int)HttpStatusCode.OK).ToString() + " " + HttpStatusCode.OK;
+                        break;
+                    case Bundle.HTTPVerb.DELETE:
+                        statusString = ((int)HttpStatusCode.NoContent).ToString() + " " + HttpStatusCode.NoContent;
+                        break;
+                    default:
+                        statusString = ((int)HttpStatusCode.OK).ToString() + " " + HttpStatusCode.OK;
```
Ivanidzo4ka, line 81: >statusString = ((int)HttpStatusCode.OK).ToString() + " " + HttpStatusCode.OK; [](http://example.com/codeflow?start=24&length=77)

I would throw exception here.

If we get something we don't expect, it's dangerous to hide it, and better to fail execution.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/HistoryTests.cs
apurvabhaleMS: Looks like we are already covered for Batch and Transaction

Ivanidzo4ka (approved):
:shipit:

PR #2050, opened 2021-06-24
Bumps healthcare-shared-package-version and add transaction in initialization SQL Script

rbans96:
## Description
Bumps healthcare-shared-package major version which removes the wrapping up of the schema script under a transaction.

So also including the transaction in the latest schema initialization script(13.sql) in itself.

## Related issues
Addresses [#AB82137](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Resolute/Stories/?workitem=82137)

## Testing
Manually tested and existing tests continue to pass

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.sql
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.sql
@@ -6,3304 +6,3305 @@ SET XACT_ABORT ON
 
 BEGIN TRANSACTION
 
-/*************************************************************
-Schema Version
-**************************************************************/
-
-INSERT INTO dbo.SchemaVersion
-VALUES
-(13, 'started')
-
-GO
-
-/*************************************************************
-Migration progress
-**************************************************************/
-
-CREATE TABLE dbo.SchemaMigrationProgress
-(
-Timestamp datetime2(3) default CURRENT_TIMESTAMP,
-Message nvarchar(max)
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/13.sql
```diff
@@ -1,5 +1,11 @@
 -- Style guide: please see: https://github.com/ktaranov/sqlserver-kit/blob/master/SQL%20Server%20Name%20Convention%20and%20T-SQL%20Programming%20Style.md
 
+-- The wrapping up of the initialization script under a transaction is removed from the schema-manager tool, so adding transaction in the script itself.
+
+SET XACT_ABORT ON
```
Ivanidzo4ka, line 5: Do we need to do this retroactivily to all scripts?

Would it be bad if someone with new schema manager tool decide to update their script from let's say 8.sql to 9.sql?

rbans96: No, we don't need to do this for older scripts. Since when we initialize the schema we directly execute the x.sql where x is the latest.
I am not sure if I got the second part - but if it is about migration i.e. going from schema version 8 to 9 then this change won't impact because for the schema migration from we consider x.diff.sql.

Ivanidzo4ka (approved):
:shipit:

PR #2054, opened 2021-06-25
Resource change data store (SQL)

bbatjargal:
## Description

Adding model and data store classes for a resource change data table for the SQL database.

This PR:
 - Adds a model class.
 - Adds a data store called SqlServerFhirResourceChangeDataStore.
 - Adds unit tests and integration tests for the resource change capture.

## Related issues
Addresses [#82790](https://microsofthealth.visualstudio.com/Health/_workitems/edit/82790).

## Testing
Added unit tests and integration tests.

## FHIR Team Checklist


src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
```diff
@@ -0,0 +1,78 @@
...
+                                    ResourceChangeTypeId = (byte)sqlDataReader["ResourceChangeTypeId"],
+                                });
+                            }
+                        }
+
+                        return listResourceChangeData;
+                    }
+                }
+            }
+            catch (Exception ex)
```
ms-teli, line 71: Questions: are there any different types of exceptions we can pivot here? For example, we can differentiate these scenarios:

1. Intermittent connection error => we can retry
2. Un-processable data => we should not retry, as there could be a bug.
3. Other unknown exception => just log them.

src/Microsoft.Health.Fhir.SqlServer.UnitTests/Features/ChangeFeed/SqlServerFhirResourceChangeDataStoreTests.cs
```diff
@@ -0,0 +1,45 @@
...
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.Health.Fhir.SqlServer.Features.ChangeFeed;
+using Microsoft.Health.SqlServer;
+using Microsoft.Health.SqlServer.Configs;
+using Xunit;
+
+namespace Microsoft.Health.Fhir.SqlServer.UnitTests.Features.ChangeFeed
+{
+    public class SqlServerFhirResourceChangeDataStoreTests
```
ms-teli, line 17: question: is there a way that we can mock and have an end-to-end test? For example, mocking the results from the sqlfactory and sqlcommand, then validating the data processing and returning as well.

ms-teli:
LGTM overall, left a few questions. Please also have FHIR team review it as well.

src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
bbatjargal: Are you saying that "retry logic" should be in this function? or we need to define custom exception classes. I think "retry logic", and handling exceptions should be on the caller side (publish event service).

kannanMSFT: One thing I would do is use this to document the potential exception types that could be thrown ( like SqlException, TimeoutException etc) https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/xmldoc/exception

src/Microsoft.Health.Fhir.SqlServer.UnitTests/Features/ChangeFeed/SqlServerFhirResourceChangeDataStoreTests.cs
bbatjargal: The scenario is going to covered by an integration test which I am going to add after PR 2032.

src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
bbatjargal: Added comments and exceptions.

ms-teli (approved):

src/Microsoft.Health.Fhir.Core/Models/ResourceChangeData.cs
```diff
@@ -0,0 +1,24 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+
+namespace Microsoft.Health.Fhir.Core.Models
+{
+    public class ResourceChangeData : IResourceChangeData
```
Ivanidzo4ka, line 10: >IResourceChangeData [](http://example.com/codeflow?start=38&length=19)

What do you get out of having this interface?
What purpose does it serves?

Ivanidzo4ka, line 10: >ResourceChangeData [](http://example.com/codeflow?start=17&length=18)

I would make this class as `record`, i.e. immutable with only place where you can set values is constuctor.

It's generally dangerous to keep setters public for something which repressent DTO.

src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
```diff
@@ -0,0 +1,111 @@
...
+        /// <exception cref="System.Exception">Thrown when errors occur during execution.</exception>
+        public async Task<IReadOnlyCollection<IResourceChangeData>> GetRecordsAsync(long startId, int pageSize, CancellationToken cancellationToken)
+        {
+            EnsureArg.IsGte(startId, 0, nameof(startId));
+            EnsureArg.IsGte(pageSize, 0, nameof(pageSize));
+
+            var listResourceChangeData = new List<ResourceChangeData>();
+            try
+            {
+                using (SqlConnection sqlConnection = await _sqlConnectionFactory.GetSqlConnectionAsync(cancellationToken: cancellationToken))
```
Ivanidzo4ka, line 65: >GetSqlConnectionAsync [](http://example.com/codeflow?start=81&length=21)

How often this GetRecordsAsync would be called?

I don't know implementation details, but I'm conscern what we can create too many connections to sql and just exhaust sockets sql side.

src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
```diff
@@ -0,0 +1,111 @@
...
+    /// <summary>
+    /// Data store for resource changes.
+    /// </summary>
+    public class SqlServerFhirResourceChangeDataStore : IChangeFeedSource<IResourceChangeData>
+    {
+        private readonly ISqlConnectionFactory _sqlConnectionFactory;
+        private readonly ILogger<SqlServerFhirResourceChangeDataStore> _logger;
+
+        // dbnetlib error value for timeout expired
+        public const short TIMEOUTEXPIRED = -2;
```
Ivanidzo4ka, line 29: >public [](http://example.com/codeflow?start=8&length=6)

can be private

src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
```diff
@@ -0,0 +1,111 @@
...
+using Microsoft.Data.SqlClient;
+using Microsoft.Extensions.Logging;
+using Microsoft.Health.Abstractions.Data;
+using Microsoft.Health.Fhir.Core.Models;
+using Microsoft.Health.SqlServer;
+
+namespace Microsoft.Health.Fhir.SqlServer.Features.ChangeFeed
+{
+    /// <summary>
+    /// Data store for resource changes.
```
Ivanidzo4ka, line 21: >Data store for resource changes. [](http://example.com/codeflow?start=8&length=32)

It's better to add comment `why` this class needed than summary which just rephrase name of the class.

src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
```diff
@@ -0,0 +1,111 @@
...
+                }
+            }
+            catch (SqlException ex)
+            {
+                switch (ex.Number)
+                {
+                    case TIMEOUTEXPIRED:
+                        throw new TimeoutException(ex.Message, ex);
+                    default:
+                        _logger.LogError(ex, Resources.SqlExceptionOccurredWhenFetchingResourceChanges);
```
Ivanidzo4ka, line 100: >LogError [](http://example.com/codeflow?start=32&length=8)

would it log `ex.Number`? if not, I would suggest to add it, just make it easier on person who would need to expand this switch.

src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
```diff
@@ -0,0 +1,111 @@
...
+        /// <returns>Resource change data rows.</returns>
+        /// <exception cref="System.ArgumentOutOfRangeException">Thrown if startId or pageSize is less than zero.</exception>
+        /// <exception cref="System.InvalidOperationException">Thrown when a method call is invalid for the object's current state.</exception>
+        /// <exception cref="Microsoft.Data.SqlClient.SqlException">Thrown when SQL Server returns a warning or error.</exception>
+        /// <exception cref="System.TimeoutException">Thrown when the time allotted for a process or operation has expired.</exception>
+        /// <exception cref="System.Exception">Thrown when errors occur during execution.</exception>
+        public async Task<IReadOnlyCollection<IResourceChangeData>> GetRecordsAsync(long startId, int pageSize, CancellationToken cancellationToken)
+        {
+            EnsureArg.IsGte(startId, 0, nameof(startId));
+            EnsureArg.IsGte(pageSize, 0, nameof(pageSize));
```
deepak-ms, line 60: Is "pageSize==0" a valid scenario?

src/Microsoft.Health.Fhir.Core/Models/ResourceChangeData.cs
bbatjargal: It is general practice to use abstraction to enable various implementations for different purposes such as testing.

bbatjargal: You're right. Thank you for pointing it out. At this moment, I would like to keep it as a POCO and add this item using the "record" to our backlog due to there is no usage of "record" in our code base, and as a new feature, it might require a version upgrade for the healthcare shared components.

Ivanidzo4ka: In general, yes, interfaces are very powerful tool for decoupling.
But in this particular case, does it bring anything?

src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
bbatjargal: It would be called every second by one agent. Currently, there are no concurrent agents. The using block closes a connection automatically when it exits. Based on our test, there will only one active connection, and there will be some suspending connections that would be removed at any given moment for the publish event agent. At this time, it would be a good option that opens and closes a connection for each call, and we don't have many options to manage DB connections. We will evaluate database connection pooling options when supporting concurrent agents.

src/Microsoft.Health.Fhir.Core/Models/ResourceChangeData.cs
Ivanidzo4ka: It doesn't have to be `record`, but I think it would be good if this class had immutable properties. I would assume this information is important and to give anyone in the code ability to change it's values sounds a bit dangerous.

src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
```diff
@@ -0,0 +1,111 @@
...
+            try
+            {
+                using (SqlConnection sqlConnection = await _sqlConnectionFactory.GetSqlConnectionAsync(cancellationToken: cancellationToken))
+                {
+                    await sqlConnection.OpenAsync(cancellationToken);
+                    using (SqlCommand sqlCommand = new SqlCommand("dbo.FetchResourceChanges", sqlConnection))
+                    {
+                        sqlCommand.CommandType = CommandType.StoredProcedure;
+                        sqlCommand.Parameters.AddWithValue("@startId", SqlDbType.BigInt).Value = startId;
+                        sqlCommand.Parameters.AddWithValue("@pageSize", SqlDbType.Int).Value = pageSize;
```
Ivanidzo4ka, line 72: >SqlDbType.Int [](http://example.com/codeflow?start=72&length=13)

I think you changed it to smallint in other PR.

src/Microsoft.Health.Fhir.Core/Models/ResourceChangeData.cs
bbatjargal: Basically defining a contract for the resource change data. So, the published event worker can define an input contract using the interface. Also, it can be extended later.

bbatjargal: Made the class immutable.

src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
bbatjargal: Added a summary.

src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
bbatjargal: Added "ex.Number" for logging.

src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
bbatjargal: You're right. Changed it as greater than equal to 1 for the pageSize.

src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
bbatjargal: Changed the data type of page size to smallint.

src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
bbatjargal: Changed it to private.

src/Microsoft.Health.Fhir.Core/Models/ResourceChangeData.cs
Ivanidzo4ka: This PR has class `SqlServerFhirResourceChangeDataStore` which derived from `IChangeFeedSource<T>`

`IChangeFeedSource` is your main contract of getting data. Why `T` has to be interface. What do you get out of this second degree of freedom? 

I could understand some logic if interface contains methods, but that interface is just plain POCO class and maps to a POCO class. 

Can you provide example how you plan to use it in future as interface, because in current PR there is no need for that interface.

src/Microsoft.Health.Fhir.SqlServer.UnitTests/Features/ChangeFeed/SqlServerFhirResourceChangeDataStoreTests.cs
bbatjargal: Integration tests are added to the PR.

src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
```diff
@@ -0,0 +1,159 @@
...
+        /// <param name="cancellationToken">Cancellation token.</param>
+        /// <returns>Resource change data rows.</returns>
+        /// <exception cref="System.ArgumentOutOfRangeException">Thrown if startId or pageSize is less than zero.</exception>
+        /// <exception cref="System.InvalidOperationException">Thrown when a method call is invalid for the object's current state.</exception>
+        /// <exception cref="Microsoft.Data.SqlClient.SqlException">Thrown when SQL Server returns a warning or error.</exception>
+        /// <exception cref="System.TimeoutException">Thrown when the time allotted for a process or operation has expired.</exception>
+        /// <exception cref="System.Exception">Thrown when errors occur during execution.</exception>
+        public async Task<IReadOnlyCollection<IResourceChangeData>> GetRecordsAsync(long startId, int pageSize, CancellationToken cancellationToken)
+        {
+            EnsureArg.IsGte(startId, 0, nameof(startId));
```
Ivanidzo4ka, line 59: >startId [](http://example.com/codeflow?start=28&length=7)

supernit: 
`Id bigint IDENTITY(1,1)`
it starts from 1, so having start id equal to 0 is same as letting it be negative (doesn't break anything, but doesn't do anything as well.)

src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
```diff
@@ -0,0 +1,159 @@
...
+        /// </summary>
+        /// <param name="cancellationToken">Cancellation Token.</param>
+        /// <returns>Returns a resource type map.</returns>
+        /// <exception cref="System.InvalidOperationException">Thrown when a method call is invalid for the object's current state.</exception>
+        /// <exception cref="Microsoft.Data.SqlClient.SqlException">Thrown when SQL Server returns a warning or error.</exception>
+        /// <exception cref="System.TimeoutException">Thrown when the time allotted for a process or operation has expired.</exception>
+        /// <exception cref="System.Exception">Thrown when errors occur during execution.</exception>
+        public async Task<IDictionary<short, string>> GetResourceTypeMapAsync(CancellationToken cancellationToken)
+        {
+            var resourceTypeIdToNameMapping = new Dictionary<short, string>();
```
Ivanidzo4ka, line 120: >resourceTypeIdToNameMapping [](http://example.com/codeflow?start=16&length=27)

Cache it please.

That collection is constant for service life duration.

src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
bbatjargal: Nice catch. Changed the start id validation as greater than or equal to 1.

src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
bbatjargal: Sure, resource types will be cached on a caller side in the publish event worker.

Ivanidzo4ka: What if they don't?  
Why rely on upstream when we can cache it here?

bbatjargal: I think that a GetResourceTypeMapAsync is a DAO function for the SQL server which gets resource types from the database. Caching should be on the caller's side.

Ivanidzo4ka: Yes, it's a DAO function and it has specific behaviour of returning same results again and again. Pushing responsibility to account to this specific behaviour on caller is not a good thing to do, because caller can ignore to do it, and just bombard us with requests.

It's better to incapsulate this logic inside method. 

As alternative we can always make this one time call function and throw exception on second call to force caller to call it only once. But that's not a good solution comparing to caching results inside function.

kannanMSFT: That doesn't sound right to me Ivan. Let's say you have a Web App and SQL server. Would you cache from your middle tier or your DOA layer.

src/Microsoft.Health.Fhir.Core/Models/ResourceChangeData.cs
bbatjargal: Removed the IResourceChangeData interface due to no functionality and it is only used for a POCO class. Will add later if needed.

src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
Ivanidzo4ka: There should be a difference between handling generic cases and specific cases, right?

If I know what collection is same all the time, I'll cache in DAO, because it's I know it's same all the time, and I'm not sure person who works in middle tier know about it.


[https://github.com/microsoft/fhir-server/blob/ecf39a2888cb8449b1085867288153f3f8a8b607/src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs#L356](https://github.com/microsoft/fhir-server/blob/ecf39a2888cb8449b1085867288153f3f8a8b607/src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs#L356)

We already cache this collection by reading it once.


Which also brings another issue, before first request to database that collection would be empty.

So if event agent is hosted service, it wouldn't be right to cache it.

Any chance you can event agent can use ISqlServerFhirModel to get dictionary? (I know there is no dictionary in interface, but we can extend interface) to protect from been called in wrong time?

src/Microsoft.Health.Fhir.Core/Models/ResourceChangeData.cs
```diff
@@ -0,0 +1,34 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+
+namespace Microsoft.Health.Fhir.Core.Models
+{
+    public class ResourceChangeData
+    {
+        public ResourceChangeData(long id, DateTime timestamp, string resourceId, short resourceTypeId, int resourceVersion, byte resourceChangeTypeId)
```
ms-teli, line 12: Long parameter list is code smell, let's make sure we have a backlog for improvement. Options could be the new csharp feature record or by adding a model builder pattern or something else.

src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
```diff
@@ -0,0 +1,159 @@
...
+using Microsoft.Health.Abstractions.Data;
+using Microsoft.Health.Fhir.Core.Models;
+using Microsoft.Health.SqlServer;
+
+namespace Microsoft.Health.Fhir.SqlServer.Features.ChangeFeed
+{
+    /// <summary>
+    /// A data store that provides functionalities to get resource changes and resource types for change feed sources.
+    /// </summary>
+    public class SqlServerFhirResourceChangeDataStore : IChangeFeedSource<ResourceChangeData>
```
ms-teli, line 23: What's the plan for ResourceChangeType?

src/Microsoft.Health.Fhir.Core/Models/ResourceChangeData.cs
bbatjargal: Created a user store [83431](https://microsofthealth.visualstudio.com/Health/_workitems/edit/83431) as a backlog.

src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
bbatjargal: Mapping from ResourceChangeTypeId to ResourceChangeTypeName is unnecessary. Because, ResourceChangeTypeId is mapped to Event Grid type such as Microsoft.HealthcareApis.FhirResourceUpdated using an enum.

src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
bbatjargal: Added caching.

test/Microsoft.Health.Fhir.R4.Tests.Integration/Features/ChangeFeed/SqlServerFhirResourceChangeCaptureTests.cs
```diff
@@ -0,0 +1,313 @@
...
+        /// insert and read resource changes after inserting a resource.
+        /// </summary>
+        [Fact]
+        public async Task GivenADatabaseSupportsResourceChangeCapture_WhenInsertingAResource_ThenResourceChangesShouldBeReturned()
+        {
+            FhirStorageTestsFixture fhirStorageTestsFixture = null;
+
+            try
+            {
+                for (int i = SchemaVersionConstants.SupportsResourceChangeCaptureSchemaVersion; i <= SchemaVersionConstants.Max; i++)
```
rotodd, line 52: Just so I understand, is there a reason why we need to test future schema versions?

test/Microsoft.Health.Fhir.R4.Tests.Integration/Features/ChangeFeed/SqlServerFhirResourceChangeCaptureTests.cs
```diff
@@ -0,0 +1,313 @@
...
+                {
+                    string databaseName = $"FHIRRESOURCECHANGECAPTUREINSERTIONTEST_V{i}_{DateTimeOffset.UtcNow.ToUnixTimeSeconds()}";
+                    try
+                    {
+                        // this will either create the database or upgrade the schema.
+                        fhirStorageTestsFixture = new FhirStorageTestsFixture(new SqlServerFhirStorageTestsFixture(i, databaseName, _coreFeatureConfigOptions));
```
rotodd, line 58: Is it necessary to create a new test fixture for each of these tests? Could we instead keep the same database and check that the `resourceChanges.Count` increased by 1?

src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
```diff
@@ -0,0 +1,134 @@
...
+    /// A data store that provides functionalities to get resource changes for change feed sources.
+    /// </summary>
+    public class SqlServerFhirResourceChangeDataStore : IChangeFeedSource<ResourceChangeData>
+    {
+        private readonly ISqlConnectionFactory _sqlConnectionFactory;
+        private readonly ILogger<SqlServerFhirResourceChangeDataStore> _logger;
+        private ConcurrentDictionary<short, string> _resourceTypeIdToTypeName;
+
+        // dbnetlib error value for timeout expired
+        private const short TIMEOUTEXPIRED = -2;
```
rotodd, line 31: I realize this isn't a custom error code, but does it make sense to add this to `SqlErrorCodes.cs`?

src/Microsoft.Health.Fhir.Core/Models/ResourceChangeData.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        public long Id { get; }
+
+        public string ResourceId { get; }
+
+        public short ResourceTypeId { get; }
+
+        public int ResourceVersion { get; }
+
+        public byte ResourceChangeTypeId { get; }
```
rotodd, line 31: I understand that this is efficiently stored as a `tinyint` in the `dbo.ResourceChangeData` table. Have you considered using an enum instead? It would be less space efficient but more readable.

We made this trade off in past with the `OperationStatus` enum used for the `JobRecord.Status` property that is stored as `varchar(10)` in the `dbo.ExportJob` table.

rotodd (commented):
+1 Great work on this 😎

src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
bbatjargal: Added a comment.

src/Microsoft.Health.Fhir.SqlServer.UnitTests/Features/ChangeFeed/SqlServerFhirResourceChangeDataStoreTests.cs
```diff
@@ -0,0 +1,74 @@
...
+using Xunit;
+
+namespace Microsoft.Health.Fhir.SqlServer.UnitTests.Features.ChangeFeed
+{
+    /// <summary>
+    /// Unit tests to validates input parameters of a GetRecordsAsync function.
+    /// </summary>
+    public class SqlServerFhirResourceChangeDataStoreTests
+    {
+        private SqlServerFhirResourceChangeDataStore resourceChangeDataStore;
```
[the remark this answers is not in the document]
bbatjargal: Added readonly.

src/Microsoft.Health.Fhir.SqlServer.UnitTests/Features/ChangeFeed/SqlServerFhirResourceChangeDataStoreTests.cs
```diff
@@ -0,0 +1,74 @@
...
+        [Fact]
+        public async Task GivenTheStartIdLessThanZero_ExceptionShouldBeThrown()
+        {
+            var expectedStartString = "Value '-1' is not greater than or equal to limit '1'. (Parameter 'startId')";
+            var exception = await Assert.ThrowsAsync<ArgumentOutOfRangeException>(async () => await resourceChangeDataStore.GetRecordsAsync(-1, 200, CancellationToken.None));
+            Assert.StartsWith(expectedStartString, exception.Message);
+        }
+
+        [Fact]
+        public async Task GivenThePageSizeLessThanZero_ExceptionShouldBeThrown()
```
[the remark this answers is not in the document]
bbatjargal: Renamed functions.

test/Microsoft.Health.Fhir.R4.Tests.Integration/Features/ChangeFeed/SqlServerFhirResourceChangeCaptureTests.cs
bbatjargal: Wanted to validate all schema versions which support a resource change capture feature.

test/Microsoft.Health.Fhir.R4.Tests.Integration/Features/ChangeFeed/SqlServerFhirResourceChangeCaptureTests.cs
bbatjargal: Technically, it is possible if validating it on the same database for each test case. Currently, validating it on the multiple databases for schema versions that support a resource change capture feature.

src/Microsoft.Health.Fhir.Core/Models/ResourceChangeData.cs
bbatjargal: There is a separate table called ResourceChangeType for resource change types due to support ad-hoc queries.

test/Microsoft.Health.Fhir.R4.Tests.Integration/Features/ChangeFeed/SqlServerFhirResourceChangeCaptureTests.cs
```diff
@@ -0,0 +1,313 @@
...
+using Microsoft.Health.Fhir.Core.Messages.Delete;
+using Microsoft.Health.Fhir.SqlServer.Features.ChangeFeed;
+using Microsoft.Health.Fhir.SqlServer.Features.Schema;
+using Microsoft.Health.Fhir.Tests.Common;
+using Microsoft.Health.Fhir.Tests.Integration.Persistence;
+using Microsoft.Health.SqlServer;
+using Microsoft.Health.SqlServer.Configs;
+using Xunit;
+
+namespace Microsoft.Health.Fhir.Tests.Integration.Features.ChangeFeed
```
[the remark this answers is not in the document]
bbatjargal: Moved integration tests in the Microsoft.Health.Fhir.Shared.Tests.Integration.

test/Microsoft.Health.Fhir.R4.Tests.Integration/Features/ChangeFeed/SqlServerFhirResourceChangeCaptureTests.cs
rotodd: I'm sorry to be a pain, but I am still not convinced this is necessary! So I can understand, may you outline a scenario where this would catch a backwards compatibility error?

I don't think the `FhirStorageTestsFixture` was intended to be instantiated multiple times, and I believe there could be a cost to testing with multiple schemas instead of using the test fixture that is already created for other integration tests.

test/Microsoft.Health.Fhir.R4.Tests.Integration/Features/ChangeFeed/SqlServerFhirResourceChangeCaptureTests.cs
rotodd: Unresolving this for now as it is related to [this discussion](https://github.com/microsoft/fhir-server/pull/2054#discussion_r668075412).

test/Microsoft.Health.Fhir.R4.Tests.Integration/Features/ChangeFeed/SqlServerFhirResourceChangeCaptureTests.cs
bbatjargal: No worries. There are the following reasons why using the different instances of FhirStorageTestsFixture instead. 
- Validate the functionality for old schema versions.
- Need to initialize the SqlServerFhirDataStore with different settings. 
- Avoid breaking the existing test cases, and minimize the changes.

src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
```diff
@@ -0,0 +1,136 @@
...
+            try
+            {
+                // The GetRecordsAsync function would be called every second by one agent.
+                // So, it would be a good option that opens and closes a connection for each call,
+                // and there is no database connection pooling in the Application at this time.
+                using (SqlConnection sqlConnection = await _sqlConnectionFactory.GetSqlConnectionAsync(cancellationToken: cancellationToken))
+                {
+                    await sqlConnection.OpenAsync(cancellationToken);
+
+                    if (resourceTypeIdToTypeNameMap.IsEmpty)
```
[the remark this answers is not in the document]
bbatjargal: [ConcurrentDictionary](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2?redirectedfrom=MSDN&view=net-5.0#thread-safety) guaranties thread-safe.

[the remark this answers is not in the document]
ms-teli: can you elaborate? the ConcurrentDictionary is not the object we're synchronizing on in this code. ideally, we want line 74 - line 77 be atomic (another way is we can get this if with another concurrent flag), but I don't understand how that happens in here. For example, we have two thread 1 and 2. 1 get to this if clause. before 1 start putting stuff to the concurrentMap after reading from SQL, 2 gets to this if as well, I don't think 2 will be blocked, so # will proceed as well, then more reads to the SQL and other redundancy occurred.

test/Microsoft.Health.Fhir.R4.Tests.Integration/Features/ChangeFeed/SqlServerFhirResourceChangeCaptureTests.cs
rotodd: I still have a few questions about this - would be great to chat and clear things up before you merge

src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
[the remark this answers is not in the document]
bbatjargal: Added lock.

src/Microsoft.Health.Fhir.Core/Models/ResourceChangeData.cs
rotodd: As discussed, there will be lot of rows in the `dbo.ResourceChangeData` table, so small space efficiencies will go a long way here. Good to keep as is!

src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
rotodd: This change was merged into the healthcare shared components repo in [this PR](https://github.com/microsoft/healthcare-shared-components/pull/269). As discussed, since it isn't crucial that we pull this in now, we can just create a user story so we don't forget to update it later.

bbatjargal: Created a [user store](https://microsofthealth.visualstudio.com/Health/_workitems/edit/83513).

test/Microsoft.Health.Fhir.R4.Tests.Integration/Features/ChangeFeed/SqlServerFhirResourceChangeCaptureTests.cs
bbatjargal: As discussed, added a fixture class to share a single database instance among all resource change capture tests.

test/Microsoft.Health.Fhir.R4.Tests.Integration/Features/ChangeFeed/SqlServerFhirResourceChangeCaptureTests.cs
bbatjargal: Added a fixture class, and validated a resource change by the resource id.

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Features/ChangeFeed/SqlServerFhirResourceChangeCaptureTests.cs
```diff
@@ -96,56 +74,34 @@ public async Task GivenADatabaseSupportsResourceChangeCapture_WhenInsertingAReso
...
+                var deserialized = updateResult.RawResourceElement.ToResourceElement(Deserializers.ResourceDeserializer);
+
+                // get resource changes
+                var resourceChangeDataStore = new SqlServerFhirResourceChangeDataStore(GetSqlConnectionFactory(_fixture.DatabaseName), NullLogger<SqlServerFhirResourceChangeDataStore>.Instance);
+                var resourceChanges = await resourceChangeDataStore.GetRecordsAsync(1, 200, CancellationToken.None);
+
+                Assert.NotNull(resourceChanges);
+                Assert.Single(resourceChanges.Where(x => x.ResourceVersion == 2 && x.ResourceId == deserialized.Id));
+
+                var resourceChangeData = resourceChanges.Where(x => x.ResourceVersion == 2 && x.ResourceId == deserialized.Id).FirstOrDefault();
```
rotodd, line 97: Should the integer here be 1? Might be nice to store this value in a variable (`var resourceUpdateId = 1`).

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Features/ChangeFeed/SqlServerFhirResourceChangeCaptureTests.cs
```diff
@@ -96,56 +74,34 @@ public async Task GivenADatabaseSupportsResourceChangeCapture_WhenInsertingAReso
...
+
+                Assert.NotNull(resourceChangeData);
+                Assert.Equal(1, resourceChangeData.ResourceChangeTypeId); // 1 is for resource udpate.
             }
             catch (Exception e)
             {
-                throw new Exception($"An error occurred on a resource change capture integration test", e);
+                throw new Exception($"An error occurred on a resource change capture integration test. Failure using schema version {SchemaVersionConstants.Max}.", e);
             }
```
rotodd: Now that we aren't running this test for multiple schema versions, I don't think we need the try-catch block!

bbatjargal: Removed the try-catch blocks.

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Features/ChangeFeed/SqlServerFhirResourceChangeCaptureTests.cs
bbatjargal: No, it is a test for updating a resource, and it is checking if there is a resource change record for the update. Used the version id from the object.

rotodd: Ah, I see, sorry! 2c7a40b looks good

rotodd (approved):

src/Microsoft.Health.Fhir.Core/Models/ResourceChangeData.cs
```diff
@@ -0,0 +1,37 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+
+namespace Microsoft.Health.Fhir.Core.Models
+{
+    public class ResourceChangeData
+    {
+        public ResourceChangeData(long id, DateTime timestamp, string resourceId, short resourceTypeId, int resourceVersion, byte resourceChangeTypeId, string resourceTypeName)
```
brendankowitz, line 12: If anything shouldn't be null, use EnsureArg.IsNotNullOrEmpty

src/Microsoft.Health.Fhir.SqlServer.UnitTests/Features/ChangeFeed/SqlServerFhirResourceChangeDataStoreTests.cs
```diff
@@ -0,0 +1,74 @@
...
+        public SqlServerFhirResourceChangeDataStoreTests()
+        {
+            var config = new SqlServerDataStoreConfiguration { ConnectionString = string.Empty };
+            var connectionStringProvider = new DefaultSqlConnectionStringProvider(config);
+            var connectionFactory = new DefaultSqlConnectionFactory(connectionStringProvider);
+            resourceChangeDataStore = new SqlServerFhirResourceChangeDataStore(connectionFactory, NullLogger<SqlServerFhirResourceChangeDataStore>.Instance);
+        }
+
+        [Fact]
+        public async Task GivenTheStartIdLessThanOne_ExceptionShouldBeThrown()
```
brendankowitz, line 33: Test names are usually "Arrage_Act_Assert" which we use as "Given_When_Then"

src/Microsoft.Health.Fhir.Core/Models/ResourceChangeData.cs
bbatjargal: Added null checks.

src/Microsoft.Health.Fhir.SqlServer.UnitTests/Features/ChangeFeed/SqlServerFhirResourceChangeDataStoreTests.cs
bbatjargal: Renamed the tests.

PR #2061, opened 2021-06-29
Logs when search parameter statuses change

rotodd:
## Description
Adds a trace message to track search parameter status changes. This was something raised during a recent investigation, and it will be helpful for any future custom search investigations.

## FHIR Team Checklist

None (minor logging change)

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterStatusManager.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterStatusManager.cs
@@ -109,6 +109,13 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Registry
 
         public async Task UpdateSearchParameterStatusAsync(IReadOnlyCollection<string> searchParameterUris, SearchParameterStatus status)
         {
+            EnsureArg.IsNotNull(searchParameterUris);
+
+            if (searchParameterUris.Count == 0)
+            {
+                return;
+            }
+
             var searchParameterStatusList = new List<ResourceSearchParameterStatus>();
             var updated = new List<SearchParameterInfo>();
             var parameters = (await _searchParameterStatusDataStore.GetSearchParameterStatuses())
@@ -116,6 +123,8 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Registry
 
             foreach (string uri in searchParameterUris)
             {
+                _logger.LogTrace("Setting the search parameter status of '{0}' to '{1}'", uri, status.ToString());
+
                 var searchParamUri = new Uri(uri);
 
                 SearchParameterInfo paramInfo = _searchParameterDefinitionManager.GetSearchParameter(searchParamUri);
--- a/src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterStatusManager.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterStatusManager.cs
@@ -123,7 +123,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Registry
 
             foreach (string uri in searchParameterUris)
             {
-                _logger.LogTrace("Setting the search parameter status of '{0}' to '{1}'", uri, status.ToString());
+                _logger.LogTrace("Setting the search parameter status of '{uri}' to '{newStatus}'", uri, status.ToString());
 
                 var searchParamUri = new Uri(uri);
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterStatusManager.cs
```diff
@@ -109,13 +109,22 @@ public async Task Handle(SearchParameterDefinitionManagerInitialized notificatio
...
+            }
+
             var searchParameterStatusList = new List<ResourceSearchParameterStatus>();
             var updated = new List<SearchParameterInfo>();
             var parameters = (await _searchParameterStatusDataStore.GetSearchParameterStatuses())
                 .ToDictionary(x => x.Uri);
 
             foreach (string uri in searchParameterUris)
             {
+                _logger.LogTrace("Setting the search parameter status of '{0}' to '{1}'", uri, status.ToString());
```
Ivanidzo4ka, line 126: >to [](http://example.com/codeflow?start=79&length=2)

Can we do: `from {1} to {2}` ?

Ivanidzo4ka, line 126: >Setting the search parameter status of '{0}' to '{1} [](http://example.com/codeflow?start=34&length=52)

super nit.
I think you can write it that way:

`Setting the search parameter status of '{uri}' to '{newstatus}`

it's a bit more readable.

feordin (approved):

src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterStatusManager.cs
rotodd: Good call, that is also more consistent with our other trace messages. Updated in fe49a0a.

Ivanidzo4ka (approved):
:shipit:

src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterStatusManager.cs
```diff
@@ -109,13 +109,22 @@ public async Task Handle(SearchParameterDefinitionManagerInitialized notificatio
...
+            }
+
             var searchParameterStatusList = new List<ResourceSearchParameterStatus>();
             var updated = new List<SearchParameterInfo>();
             var parameters = (await _searchParameterStatusDataStore.GetSearchParameterStatuses())
                 .ToDictionary(x => x.Uri);
 
             foreach (string uri in searchParameterUris)
             {
+                _logger.LogTrace("Setting the search parameter status of '{uri}' to '{newStatus}'", uri, status.ToString());
```
Ivanidzo4ka, line 126: >LogTrace [](http://example.com/codeflow?start=24&length=8)

I would do one extra step, and log stack trace as well.

That way you not only now what it changed, but also why it changed!

rotodd: I agree it would be nice to know a bit more about why it is changing - but at the same time it feels weird to add the stack trace to a trace log when there is no exception (this could just be me though).

`UpdateSearchParameterStatusAsync()` can be called from either `AddSearchParameterStatusAsync()`, `DeleteSearchParameterStatusAsync()` or `ReindexUtilities.UpdateSearchParameterStatus()` . Could it make sense to instead add logging to those methods as well?

Ivanidzo4ka: I understand feeling of weirdness. Why would we pollute out logs with stack traces? It's ok to do that for exception because it's... exception or exceptional case.

Reason because we add this logging is also has nature of exception to some extent. We trying to figure out why search parameters change it's state out of thin air, and we can't do it via deductive reasoning. So this log is to catch time where change of state occurs.

And after catching time of change we will have to investigate why it's actually changed. Which would involve going through logs and hope we have operation id, and if not it's still not exactly clear why it's changing and what triggers it.

So simple and dirty solution to add stack trace and just see exact culprit.

Adding more logs is fine and I would actually encourage that, but there is chance (haven't check all paths in code so not 100% guarantee) it wouldn't give you whole picture, and maybe would involve few more PRs which would add more logging in different parts of system. 

Which is fine by me.

rotodd: I checked all the paths in the code and confirmed that a search parameter status will be updated when:
1) A new search parameter is posted
2) A search parameter is deleted
3) A search parameter is updated (in this case, we delete and then add a new search parameter status record)
4) A reindex operation is completing and we set the fully indexed search parameters to `Enabled`

I added trace logs that will indicate which of the four scenarios above is why a given search parameter status is being updated. Please let me know if ffeeea7 looks alright.

PR #2075, opened 2021-07-07
Event agent checkpointing with SQL

wi-y:
## Description
There is a separate [resource change data pull request](https://github.com/microsoft/fhir-server/pull/2032) which enables the FHIR server to capture resources that have changed. 

There will also be an event agent that reads these changes. One feature that we would like to add for the event agent is to be able to occasionally set checkpoints for what changes have been read. This is needed because in the case that the event agent stops or restarts, we would like to be able to resume from the latest checkpoint rather than reading from the beginning of the change feed. 

This PR Creates a table (dbo.BackgroundJobCheckpoint) and stored procedures (dbo.FetchBackgroundJobCheckpoint, dbo.UpdateBackgroundJobCheckpoint) to get and set checkpoints for the event agent. The checkpoints contain information about the latest data that was successfully processed. The checkpoints can be used on startup or restart to resume from a given position or datetime.

## Related issues
Related to the [resource change data pull request](https://github.com/microsoft/fhir-server/pull/2032) and future event agent support.

## Testing
Compile .sln and build docker container. Start up application and db and verify that table and stored procedures are created.

Feature


src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/15.diff.sql
```diff
@@ -0,0 +1,98 @@
...
+
+/*************************************************************
+    Background job checkpoint table
+**************************************************************/
+
+IF NOT EXISTS (SELECT 1 FROM sys.tables WHERE name = 'BackgroundJobCheckpoint')
+BEGIN
+     CREATE TABLE dbo.BackgroundJobCheckpoint
+     (
+         Id bigint IDENTITY(1,1) NOT NULL,
```
Ivanidzo4ka, line 13: >Id bigint IDENTITY(1,1) NOT NULL, [](http://example.com/codeflow?start=9&length=33)

Is this column actually needed?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/15.diff.sql
```diff
@@ -0,0 +1,98 @@
...
+         UpdatedOn datetime2(7) NOT NULL DEFAULT sysutcdatetime(),
+         CheckpointId varchar(64) NOT NULL,
+         Partition varchar(64) NOT NULL,
+         LastProcessedDateTime datetimeoffset(7),
+         LastProcessedIdentifier varchar(64),
```
Ivanidzo4ka, line 18: Is there are document which describes how this table would be used?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/15.diff.sql
```diff
@@ -0,0 +1,98 @@
...
+    Background job checkpoint feature
+**************************************************************/
+
+/*************************************************************
+    Background job checkpoint table
+**************************************************************/
+
+IF NOT EXISTS (SELECT 1 FROM sys.tables WHERE name = 'BackgroundJobCheckpoint')
+BEGIN
+     CREATE TABLE dbo.BackgroundJobCheckpoint
```
Ivanidzo4ka, line 11: >Background [](http://example.com/codeflow?start=22&length=10)

I prefer to have more meaningful name. `Background` sound like implementation details which  can be omited.

ms-teli (commented):
Nit: since Batjargal has merged the change, we can rebase this PR to reflect the deltas in scope.

wi-y:
> Nit: since Batjargal has merged the change, we can rebase this PR to reflect the deltas in scope.

yep, I will be doing this today

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/15.diff.sql
wi-y: Thanks for the review. It is not needed. Removed

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/15.diff.sql
wi-y: Updated to EventAgentCheckpoint since that will be the application writing to this table. I was initially trying to keep the name generic enough so that other background or long running jobs could use it, but am also okay with naming it something more specific.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/15.diff.sql
wi-y: Created an internal spec

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/15.diff.sql
```diff
@@ -0,0 +1,97 @@
...
+      CheckpointId,
+      Partition,
+      LastProcessedDateTime,
+      LastProcessedIdentifier
+      FROM dbo.EventAgentCheckpoint
+    WHERE CheckpointId = @CheckpointId AND Partition = @Partition
+END
+GO
+    COMMIT TRANSACTION
+GO
```
bbatjargal, line 97: This commit transaction might be additional.

wi-y: thanks, removed

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/15.diff.sql
```diff
@@ -0,0 +1,88 @@
...
+
+IF NOT EXISTS (SELECT 1 FROM sys.tables WHERE name = 'EventAgentCheckpoint')
+BEGIN
+     CREATE TABLE dbo.EventAgentCheckpoint
+     (
+         CheckpointId varchar(64) NOT NULL,
+         LastProcessedDateTime datetimeoffset(7),
+         LastProcessedIdentifier varchar(64),
+         UpdatedOn datetime2(7) NOT NULL DEFAULT sysutcdatetime(),
+        CONSTRAINT PK_EventAgentCheckpoint PRIMARY KEY CLUSTERED (CheckpointId)
```
ms-teli, line 17: Nit: indention misaligned.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/15.diff.sql
```diff
@@ -0,0 +1,88 @@
...
+
+/*************************************************************
+    Event Agent checkpoint table
+**************************************************************/
+
+IF NOT EXISTS (SELECT 1 FROM sys.tables WHERE name = 'EventAgentCheckpoint')
+BEGIN
+     CREATE TABLE dbo.EventAgentCheckpoint
+     (
+         CheckpointId varchar(64) NOT NULL,
```
ms-teli, line 13: what's the strategy of assigning the checkpointId? This schema instance should be per database instance, and database instance should  be per FHIR account, if we currently only has one agent for the FHIR service at a time, how this column is used?

wi-y: I was thinking the CheckpointId could be a constant identifier set by the Event Agent in our first iteration. It is not totally needed if there will only be one Event agent. However the interfaces I had defined in healthcare-shared-components require an identifier currently because I was thinking of the future cases where there could be multiple Event Agents.

ms-teli: Got it, thanks for explaining.

We would make sure that the checkpointId should be consistent if the EventAgent is down and restarted, otherwise, the checkpoint will not be picked up.

ms-teli (approved):
I dont have other questions, thanks for updating.

Ivanidzo4ka (approved):
:shipit:

Ivanidzo4ka:
@wi-y Let me know do you want me merge this PR, or it still need more changes?

wi-y:
> @wi-y Let me know do you want me merge this PR, or it still need more changes?

@Ivanidzo4ka yes it's done - can you merge it? I haven't been able to get the SQL R4 E2E tests to succeed but I believe the failures are not related to this PR

PR #2099, opened 2021-07-17
Adds final reindex count check

rotodd:
## Description
At the end of a reindexing operation, the count of resources that are not reindexed should always be zero. This PR adds an additional check for this for patient safety reasons. 

## Related issues
Addresses [AB#80638](https://microsofthealth.visualstudio.com/Health/_workitems/edit/80638).

## Testing
To test this, I manually updated the value that `CalculateTotalCount()` returned and verified that:
1) the reindex status document indicates that the operation failed
2) the reindex status document has the appropriate error message(s)
3) the error was logged.

## FHIR Team Checklist

Skip (added safety measure)


src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -543,7 +550,17 @@ private async Task CheckJobCompletionStatus()
                 // all queries marked as complete, reindex job is done, check success or failure
                 if (_reindexJobRecord.QueryList.Keys.All(q => q.Status == OperationStatus.Completed))
                 {
-                    await UpdateParametersAndCompleteJob();
+                    // Perform a final check to make sure there are no resources left to reindex
+                    int totalCount = await CalculateTotalCount();
+                    if (totalCount != 0)
+                    {
+                        string message = $"{totalCount} resources failed to be reindexed.";
+                        await HandleRetryableError(message, OperationOutcomeConstants.IssueType.Incomplete);
```
feordin, line 558: Is the intent here to retry this total count again? Or is this error not something we expect to be able to recover from on a retry?

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -543,7 +550,17 @@ private async Task CheckJobCompletionStatus()
                 // all queries marked as complete, reindex job is done, check success or failure
                 if (_reindexJobRecord.QueryList.Keys.All(q => q.Status == OperationStatus.Completed))
                 {
-                    await UpdateParametersAndCompleteJob();
+                    // Perform a final check to make sure there are no resources left to reindex
+                    int totalCount = await CalculateTotalCount();
+                    if (totalCount != 0)
+                    {
+                        string message = $"{totalCount} resources failed to be reindexed.";
```
feordin, line 557: Do we want to explain to the user that they should resubmit the same reindex job again in order to finish indexing the remaining resources?

feordin (commented):
I have a couple questions about how to handle the case when we do find some resources needing to be reindexed.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -543,7 +550,17 @@ private async Task CheckJobCompletionStatus()
                 // all queries marked as complete, reindex job is done, check success or failure
                 if (_reindexJobRecord.QueryList.Keys.All(q => q.Status == OperationStatus.Completed))
                 {
-                    await UpdateParametersAndCompleteJob();
+                    // Perform a final check to make sure there are no resources left to reindex
```
Ivanidzo4ka, line 553: >Perform a final check to make sure there are no resources left to reindex [](http://example.com/codeflow?start=23&length=73)

Imagine frustration of the user after 3 days of reindexing your job failed with:
"1 resources failed to be reindexed" 

Do we communicate clearly what rest of resources are actually reindexed? 

What would happen if user submit same job, would it be 3 days job agian?

Is there are way for user to find out what resources giving trouble?

Should user have right to pick threshold for this?

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
rotodd: Good idea - updated the error message returned to the user in f1bce27.

2021-07-19 15_03_59-Response(47ms) - Visual Studio Code

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
rotodd: As discussed, it is difficult to know if we could recover from this on a retry, since this failure will ultimately never be reached!

@Ivanidzo4ka made a good point below that automatically retrying a process that could take many days might not be the best user experience. It would instead be favourable to give control to the user to decide if/when to retry the reindexing operation.

I removed the retry logic in f1bce27 and updated [AB#80638](https://microsofthealth.visualstudio.com/Health/_workitems/edit/80638).

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
rotodd: Removed the retry logic and updated the information returned to the user in f1bce27.

feordin (approved):

PR #2085, opened 2021-07-14
Disable auth tests to unblock build pipeline

namadabu:
## Description
Currently build pipelines are blocked since we are running into an AAD issue. To unblock ourselves, I have disabled the auth tests for now. This will give us more time to investigate and fix the underlying issue.

I created a separate copy of the run-tests.yml for PR build pipeline so that our CI build pipeline is not affected. The new run-pr-tests.yml file contains an extra test filter - Category!=Authorization. We will revert back these changes once we have a proper fix in place to address the AAD issue.

## Related issues
Addresses [issue #].

## Testing
Describe how this change was tested.

## FHIR Team Checklist
- [ ] **Update the title** of the PR to be succinct and less than 50 characters
- [ ] **Add a milestone** to the PR for the sprint that it is merged (i.e. add S47)
- [ ] Tag the PR with the type of update: **Bug**, **Dependencies**, **Enhancement**, or **New-Feature**
- [ ] Tag the PR with **Azure API for FHIR** if this will release to the managed service
- Review [squash-merge requirements](https://github.com/microsoft/fhir-server/blob/master/SquashMergeRequirements.md)

### Semver Change ([docs](https://github.com/microsoft/fhir-server/blob/master/docs/Versioning.md))
Patch|Skip|Feature|Breaking (reason)

PR #2105, opened 2021-07-21
Build: Update client app admin consent flow

namadabu:
## Description
This code changes the way we set up permissions for the client application that we use for tests. Instead of using admin consent flow, we are granting specific delegated permissions to the client app.

As part of the PR, I am also re-enabling auth tests (which were disabled in #2085 ) for the PR pipeline.

## Related issues
Addresses [AB#83101](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/83101)

## Testing

## FHIR Team Checklist


LTA-Thinking (approved):

release/scripts/PowerShell/FhirServerRelease/Private/Grant-ClientAppDelegatedPermissions.ps1
```diff
@@ -0,0 +1,108 @@
...
+    $permissionGrantUrl = "https://graph.microsoft.com/v1.0/oauth2PermissionGrants"
+
+    $bodyForReadPermisions = @{
+        clientId = $clientObjectId
+        consentType = "AllPrincipals"
+        resourceId = $windowsAadObjectId
+        scope = "User.Read"
+    }
+
+    # TODO: Handle scenario where application might already have these permissions.
```
LTA-Thinking, line 79: What would happen currently in this situation? Would the step fail, or just not be needed?

namadabu: The api call returns a 400 :(. It shouldn't have an impact on our pipeline since we clear everything and setup a fresh test env every run. I have also added code that checks for existing permissions in case we run into an error.

LTA-Thinking (approved):

PR #2109, opened 2021-07-22
Reindex parameter boundaries updates + Unit test

apurvabhaleMS:
## Description
Updates to Unit tests
Updated display error message

## Related issues
Addresses [[81246](https://microsofthealth.visualstudio.com/Health/_workitems/edit/81246)].

## Testing
Manual Test
Unit Test

## FHIR Team Checklist


feordin (approved):
I think it looks good

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/Models/ReindexJobRecord.cs
```diff
@@ -52,7 +52,7 @@ public class ReindexJobRecord : JobRecord
             // check for MaximumConcurrency boundary
             if (maxiumumConcurrency > MaxMaximumConcurrency)
             {
-                throw new BadRequestException(string.Format(Fhir.Core.Resources.InvalidReIndexParameterValue, nameof(MaximumConcurrency), MinMaximumConcurrency.ToString(), MaxMaximumConcurrency.ToString()));
+                throw new ArgumentOutOfRangeFhirException(string.Format(Fhir.Core.Resources.InvalidReIndexParameterValue, nameof(MaximumConcurrency), MinMaximumConcurrency.ToString(), MaxMaximumConcurrency.ToString()));
```
Ivanidzo4ka, line 55: >ArgumentOutOfRangeFhirException [](http://example.com/codeflow?start=26&length=31)

Can I ask what value it adds?

From what I see we already can achive what we want with combination of existing exception and text message inside it.

What's the point to add new exception?

apurvabhaleMS: I see what your point is and I agree but the idea behind this is to club all the Argument value constraints together for the better distinction

Ivanidzo4ka: >to club all the Argument value constraints 

Aren't they already club together inside this class?

Adding something new usually justified by fact what it allows to achive something new, for the price of increasing entropy of the code.

From changes in this PR, I don't think there is anything new in behavior, only difference would be type of exception been thrown, and I'm not sure we care is it BadRequest or ArgumentOut exception.

apurvabhaleMS: hmm I see your concern. if we are fine to stick with BadException type then I am good with it ☺

Ivanidzo4ka (approved):
:shipit:

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Features/Operations/Reindex/ReindexJobTests.cs
```diff
@@ -177,7 +177,7 @@ public async Task GivenOutOfRangeReindexParameter_WhenCreatingAReindexJob_ThenEx
             catch (FhirException fhirExp)
             {
                 Assert.NotNull(fhirExp);
-                Assert.Contains($"Argument '{jobRecordProperty}' was out of the range of valid values. Please specify different value within a range".ToLower(), fhirExp.Message.ToLower());
+                Assert.Contains($"Reindex operation parameter '{jobRecordProperty}' was out of the range of valid values. Please specify different value within a range".ToLower(), fhirExp.Message.ToLower());
```
Ivanidzo4ka, line 180: >Reindex operation parameter [](http://example.com/codeflow?start=34&length=27)

Just curious, why can't we use string from Resource file here?

PR #2118, opened 2021-07-26
Reindexes rest of batch if one resource fails

rotodd:
## Description

Prior to this PR, when one resource failed to reindex due to a versioning conflict, the entire batch would fail.

Now, the remaining resources in the batch are reindexed. The query will be retried, fetching the most recent version of the failed resource from the database. In the rare scenario where all consecutive retry attempts fail, the reindexing operation will fail with the following error message:

2021-07-26 09_56_20-Response(241ms) - Visual Studio Code

## Related issues
Addresses [AB#79925](https://microsofthealth.visualstudio.com/Health/_workitems/edit/79925)

## Testing
- Manually edited database to create version conflicts and confirming appropriate failure behaviour
- Performed schema upgrades to ensure compatibility

## Next steps
When one or more resources in a batch initially fail to reindex but succeed on later attempts, the reindex job progress does not pick up the resources that succeeded on the first attempt. I created [AB#83909](https://microsofthealth.visualstudio.com/Health/_workitems/edit/83909) to track this.

## FHIR Team Checklist

Patch (new schema version)


src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/16.diff.sql
```diff
@@ -0,0 +1,285 @@
...
+    END
+
+    DECLARE @versionDiff int
+    SET @versionDiff = (SELECT COUNT(*) FROM @computedValues WHERE VersionProvided IS NOT NULL AND VersionProvided <> VersionInDatabase)
+
+    IF (@versionDiff > 0) BEGIN
+        -- Don't reindex resources that have outdated versions
+        DELETE FROM @computedValues
+        WHERE  VersionProvided IS NOT NULL AND VersionProvided <> VersionInDatabase
+    END
```
rotodd, line 111: Instead of throwing exceptions here, we simply remove the resources from `@computedValues`, the table of resources to be reindexed.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -336,22 +336,55 @@ public async Task BulkUpdateSearchParameterIndicesAsync(IReadOnlyCollection<Reso
...
+                    {
+                        if (!sqlDataReader.Read())
+                        {
+                            return;
+                        }
+
+                        int numberOfVersionConflicts = sqlDataReader.GetInt32(0);
+
+                        // TODO: Return total number of successfully reindexed resources.
+                        int numberOfDeletedResources = sqlDataReader.GetInt32(1);
```
rotodd, line 352: TBD in [AB#83909](https://microsofthealth.visualstudio.com/Health/_workitems/edit/83909).

src/Microsoft.Health.Fhir.Core/Resources.resx
```diff
@@ -584,4 +584,16 @@
...
+  </data>
+  <data name="ReindexingResourceVersionConflict" xml:space="preserve">
+    <value>One or more resources failed to reindex due to a non-matching version.</value>
+  </data>
+  <data name="ReindexingServiceUnavailable" xml:space="preserve">
+    <value>Failed to reindex resource because the Cosmos service was unavailable.</value>
+  </data>
+  <data name="ReindexingSqlError" xml:space="preserve">
+    <value>Error from SQL database on reindex.</value>
+  </data>
```
Ivanidzo4ka, line 598: I have feeling what it's worth to put strings to resource files if they are user facing one.

These two are for emitting message to a log, and I don't think it's worth to put them in resources.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -511,6 +511,12 @@ private async Task<ReindexJobQueryStatus> ProcessQueryAsync(ReindexJobQueryStatu
 
                     if (query.FailureCount >= _reindexJobConfiguration.ConsecutiveFailuresThreshold)
                     {
+                        var issue = new OperationOutcomeIssue(
+                            OperationOutcomeConstants.IssueSeverity.Error,
+                            OperationOutcomeConstants.IssueType.Exception,
+                            ex.Message);
```
Ivanidzo4ka, line 517: >ex [](http://example.com/codeflow?start=28&length=2)

If it was FhirException I think it would be ok to push it to user, but this is generic exception. 

What user suppose to do with it?

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
```diff
@@ -352,16 +352,11 @@ public async Task<ResourceWrapper> UpdateSearchParameterIndicesAsync(ResourceWra
...
 
-                    case HttpStatusCode.NotFound:
-                        throw new ResourceNotFoundException(string.Format(
-                            Core.Resources.ResourceNotFoundByIdAndVersion,
-                            resourceWrapper.ResourceTypeName,
-                            resourceWrapper.ResourceId,
-                            weakETag));
-
                     case HttpStatusCode.ServiceUnavailable:
+                        _logger.LogError(string.Format(Core.Resources.ReindexingServiceUnavailable));
```
Ivanidzo4ka, line 359: >string.Format(Core.Resources.ReindexingServiceUnavailable) [](http://example.com/codeflow?start=41&length=58)

string.Format(Core.Resources.ReindexingServiceUnavailable)

this looks strange. 
Why not just `Core.Resources.ReindexingServiceUnavailable`?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -336,22 +336,42 @@ public async Task BulkUpdateSearchParameterIndicesAsync(IReadOnlyCollection<Reso
                     sqlCommandWrapper,
                     _bulkReindexResourcesTvpGeneratorVLatest.Generate(resources.ToList()));
 
-                try
+                if (_schemaInformation.Current >= SchemaVersionConstants.BulkReindexReturnsFailuresVersion)
                 {
-                    await sqlCommandWrapper.ExecuteNonQueryAsync(cancellationToken);
+                    // We will reindex the rest of the batch if one resource has a versioning conflict
+                    var failedResourceCount = (int?)await sqlCommandWrapper.ExecuteScalarAsync(cancellationToken);
```
Ivanidzo4ka, line 342: Can we still have exception during execution?

In else block we log it, and here we don't

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -336,22 +336,42 @@ public async Task BulkUpdateSearchParameterIndicesAsync(IReadOnlyCollection<Reso
...
+                        {
+                            case SqlErrorCodes.PreconditionFailed:
+                                string message = string.Format(Core.Resources.ReindexingResourceVersionConflict);
+                                string userAction = string.Format(Core.Resources.ReindexingUserAction);
+
+                                _logger.LogError(message);
+                                throw new PreconditionFailedException(message + " " + userAction);
+
+                            default:
+                                _logger.LogError(e, string.Format(Core.Resources.ReindexingSqlError));
```
Ivanidzo4ka, line 372: I don't get reason behind `string.Format(ResourceString)`

`string.Format(str) == str`

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
```diff
@@ -352,16 +352,11 @@ public async Task<ResourceWrapper> UpdateSearchParameterIndicesAsync(ResourceWra
                 switch (exception.GetSubStatusCode())
                 {
                     case HttpStatusCode.PreconditionFailed:
+                        _logger.LogError(string.Format(Core.Resources.ResourceVersionConflict, weakETag));
                         throw new PreconditionFailedException(string.Format(Core.Resources.ResourceVersionConflict, weakETag));
 
-                    case HttpStatusCode.NotFound:
```
feordin, line 357: Why was the .NotFound case removed? Is this not a possible outcome any longer?

rotodd: Good question! I discovered that this scenario is not possible while attempting (and failing 😄) to write a test that would trigger the `NotFound` case.

The only situation that could lead to a resource not being found when updating its indices would be if the resource was deleted since the operation was kicked off. In this situation, a new deleted version of the resource would be created, and the `PreconditionFailed` case would catch this.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
rotodd: 🤦🏻‍♀️ thanks, removed redundant calls to `string.Format()` in 57a3a31

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
rotodd: Ah thank you, I was focused on the exceptions and overlooked the logging. Put this in in 57a3a31.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
rotodd: Same as below! Fixed in 57a3a31.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
rotodd: Good point. I am currently using this to pass an action item to the user on `PreconditionFailed` exceptions, but I can see how this might lead to us sending info to the user that wouldn't be helpful. Working on finding a nice fix now.

src/Microsoft.Health.Fhir.Core/Resources.resx
rotodd: +1 I agree. Removed the string resources in 57a3a31

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
rotodd: Added an extra catch block in 52ea697.

Ivanidzo4ka (approved):
:shipit:

Issue #2020, opened 2021-06-11
Observation-code-value-date search not work


tongwu-msft:
**Describe the bug**
When create a Observation resource with code and valueDateTime, search for Observation-code-value-date not return correct result.

I believe the root cause is the definition for SearchParameter is incorrect. 

"expression" : "value.as(DateTime) | value.as(Period)" should be "expression" : "value.as(dateTime) | value.as(period)"

**FHIR Version?**
Stu3/R4/R5

**Data provider?**
CosmosDB/SQL Server

**To Reproduce**
1. Create resource with following payload.
```
{
	"resourceType": "Observation",
	"id": "f205",
	"text": {
		"status": "generated",
		"div": "<div xmlns=\"http://www.w3.org/1999/xhtml\"><p><b>Generated Narrative</b></p><p><b>status</b>: entered-in-error</p><p><b>category</b>: <span>Vital Signs</span></p><p><b>code</b>: <span>Temperature</span></p><p><b>subject</b>: <a>Roel. Generated Summary: BSN: 123456789 (OFFICIAL), BSN: 123456789 (OFFICIAL); active; Roel(OFFICIAL); Phone: +31612345678, Phone: +31201234567; gender: male; birthDate: 1960-03-13; <span>Legally married</span>; </a></p><p><b>issued</b>: Apr 4, 2013 12:27:00 PM</p><p><b>performer</b>: <a>Generated Summary: UZI-nummer: 12345678901 (OFFICIAL); active; Dokter Bronsig(OFFICIAL); Phone: +31715269111; gender: male; birthDate: 1956-12-24</a></p><p><b>value</b>: 39 degrees C</p><p><b>interpretation</b>: <span>High</span></p><p><b>bodySite</b>: <span>Oral cavity</span></p><p><b>method</b>: <span>Oral temperature taking</span></p><h3>ReferenceRanges</h3><table><tr><td>-</td><td><b>High</b></td></tr><tr><td>*</td><td>38.2 degrees C</td></tr></table></div>"
	},
	"status": "entered-in-error",
	"category": [
		{
			"coding": [
				{
					"system": "http://terminology.hl7.org/CodeSystem/observation-category",
					"code": "vital-signs",
					"display": "Vital Signs"
				}
			]
		}
	],
	"code": {
		"coding": [
			{
				"system": "http://acme.lab",
				"code": "BT",
				"display": "Body temperature"
			},
			{
				"system": "http://loinc.org",
				"code": "8310-5",
[...]

Issue #901, opened 2020-02-27
Issues with search-parameters.json files


brendankowitz:
Adding a thread here to track currently manual fixes to the search-parameters.json files.

brendankowitz:
In R4, the component definition Urls for DocumentReference-Relationship are the wrong way around.
![image](https://user-images.githubusercontent.com/197221/75460811-c36c9280-5936-11ea-8898-3954aaaa2b30.png).

brendankowitz:
In Stu3, it seems like `clinical-code` should be `Observation-code`. 
![image](https://user-images.githubusercontent.com/197221/75461404-9bc9fa00-5937-11ea-998a-f50eed87e78b.png)

`Observation-code` only seems be referenced by a few places in Observation-code-value-*.

PR #2135, opened 2021-08-02
Remove case insensitivity in data types mapping 

Ivanidzo4ka:
## Description
https://github.com/microsoft/fhir-server/blob/2790a877ec4397f9c61c129a99dbf9e6145ad84a/src/Microsoft.Health.Fhir.Core/Features/Search/TypedElementSearchIndexer.cs#L190

`.Select(fhirPathExpression)` is case sensitive so there is different results in doing `value.as(DateTime)` and `value.as(dateTime)`. First yield nothing, and second produce dateTime node.

This change normalizes case in search-parameter json files and put restriction in place in order to avoid this situation in future.

On a side note: all these search parameters currently produce nothing and can't be searched, but they are in `searchable` state and I'm not sure what we can do to reindex them, or put them in `supported` state.

## Related issues
Addresses [#2020].

## Testing

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Data/R4/search-parameters.json
+++ b/src/Microsoft.Health.Fhir.Core/Data/R4/search-parameters.json
@@ -45840,7 +45840,7 @@
       },
       {
         "definition" : "http://hl7.org/fhir/SearchParameter/Observation-value-date",
-        "expression" : "value.as(DateTime) | value.as(Period)"
+        "expression" : "value.as(dateTime) | value.as(Period)"
       }]
     }
   },
--- a/src/Microsoft.Health.Fhir.Core/Data/R5/search-parameters.json
+++ b/src/Microsoft.Health.Fhir.Core/Data/R5/search-parameters.json
@@ -58554,7 +58554,7 @@
                     },
                     {
                         "definition": "http://hl7.org/fhir/SearchParameter/Observation-value-date",
-                        "expression": "value.as(DateTime) | value.as(Period)"
+                        "expression": "value.as(dateTime) | value.as(Period)"
                     }
                 ]
             }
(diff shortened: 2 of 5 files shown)


rotodd:
To have it handy, here are the search parameters that appear to be affected:

R4:

`http://hl7.org/fhir/SearchParameter/Observation-code-value-date`

R5:

`http://hl7.org/fhir/SearchParameter/Observation-code-value-date`

Stu3:
```
http://hl7.org/fhir/SearchParameter/clinical-date
http://hl7.org/fhir/SearchParameter/CommunicationRequest-occurrence
http://hl7.org/fhir/SearchParameter/ConceptMap-source-uri
http://hl7.org/fhir/SearchParameter/ConceptMap-target-uri
http://hl7.org/fhir/SearchParameter/DeviceRequest-event-date
http://hl7.org/fhir/SearchParameter/Goal-start-date
http://hl7.org/fhir/SearchParameter/Goal-target-date
http://hl7.org/fhir/SearchParameter/Observation-code-value-date
http://hl7.org/fhir/SearchParameter/Observation-value-date
http://hl7.org/fhir/SearchParameter/Observation-value-string
http://hl7.org/fhir/SearchParameter/Patient-death-date
```

brendankowitz:
Adding a reference to #901

src/Microsoft.Health.Fhir.Core/Data/R4/search-parameters.json
```diff
@@ -45840,7 +45840,7 @@
       },
       {
         "definition" : "http://hl7.org/fhir/SearchParameter/Observation-value-date",
-        "expression" : "value.as(DateTime) | value.as(Period)"
+        "expression" : "value.as(dateTime) | value.as(Period)"
```
brendankowitz, line 45843: This sounds like a parser bug too. `If the identifier cannot be resolved to a valid type identifier, the evaluator will throw an error.` http://hl7.org/fhirpath/N1/#as-type-specifier

brendankowitz (dismissed):
I don't think we should edit the search-parameters.json file.

brendankowitz:
@feordin @namadabu If these expressions change, will that give us a different hash and allow reindexing? Would there be any other side-effects here?

feordin:
@brendankowitz Yes, the expression value is used when computing the hash, so it would change the hash value.  However, as these parameters are part of the "out of the box" parameters, the startup code by default marks all of those parameters as fully indexed, a status of "enabled".  The Reindex logic will not detect them as needing to be indexed.  We would need an additional code changes to detect changes in the search-parameters.json file.

rbans96:
We should redo this for new parts of the spec(R4B, R5). @brendankowitz @feordin will create a user story for this.

PR #2143, opened 2021-08-05
Adds Binary encoded JSON Patch in Bundles.

brendankowitz:
## Description
In current moment when you submit Bundle there is no good way to pass PATCH operation.
Main reason is Bundle can work only with FHIR resources, and our PATCH implementation is based on JsonPatch  (where we pass json string of operations to be applied to resource) and not FhirPathPatch (where you pass Parameters object with operations to be applied, which is FHIR resource)

This is workaround, where instead of implementing FhirPatchPatch we treat Binary resource as base64 encoding of json string and pass it to the server.

## Related issues
Addresses [issue #].

## Testing

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
@@ -5,9 +5,11 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Net;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Transactions;
@@ -32,6 +34,7 @@ using Microsoft.Health.Fhir.Api.Configs;
 using Microsoft.Health.Fhir.Api.Features.Bundle;
 using Microsoft.Health.Fhir.Api.Features.ContentTypes;
 using Microsoft.Health.Fhir.Api.Features.Exceptions;
+using Microsoft.Health.Fhir.Api.Features.Formatters;
 using Microsoft.Health.Fhir.Api.Features.Routing;
 using Microsoft.Health.Fhir.Core.Exceptions;
 using Microsoft.Health.Fhir.Core.Extensions;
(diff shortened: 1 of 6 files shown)


src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -313,15 +316,28 @@ private async Task GenerateRequest(EntryComponent entry, int order, Cancellation
...
                 var memoryStream = new MemoryStream(_fhirJsonSerializer.SerializeToBytes(entry.Resource));
                 memoryStream.Seek(0, SeekOrigin.Begin);
                 httpContext.Request.Body = memoryStream;
             }
+#if !STU3
+            // Allow JSON Patch to be an encoded Binary in a Bundle (See: https://chat.fhir.org/#narrow/stream/179166-implementers/topic/Transaction.20with.20PATCH.20request)
+            else if (
+                requestMethod == HTTPVerb.PATCH &&
+                string.Equals(KnownResourceTypes.Binary, entry.Resource?.TypeName, StringComparison.Ordinal) &&
+                entry.Resource is Binary binaryResource && string.Equals(KnownMediaTypeHeaderValues.ApplicationJsonPatch.ToString(), binaryResource.ContentType, StringComparison.OrdinalIgnoreCase))
```
Ivanidzo4ka, line 333: >entry.Resource is Binary binaryResource  [](http://example.com/codeflow?start=16&length=40)

[https://thumbs.gfycat.com/CloseGoodAmericanriverotter-size_restricted.gif](https://thumbs.gfycat.com/CloseGoodAmericanriverotter-size_restricted.gif)

Do we need to do same thing with binary for other methods as well, or it's special hack for PATCH?

Is this wild spec interpretation, or FHIR spells this somewhere?

brendankowitz: Yes on both accounts.
See https://chat.fhir.org/#narrow/stream/179166-implementers/topic/Transaction.20with.20PATCH.20request and https://smilecdr.com/docs/fhir_repository/updating_data.html

Its not great or pretty, but in some cases where a conversion routine is going from hl7v2 => FHIR, PATCH operations are required inside bundles (I believe most importantly in transaction bundles), this "hack/workaround" unblocks this. It seems like the only near-term solution we can offer until we implement FHIR Path Patch which uses the Parameters object.

Ivanidzo4ka (approved):

PR #2147, opened 2021-08-05
Update SchemaUpgradeTests

namadabu:
## Description
Updated the GivenASchemaVersion_WhenApplyingDiffTwice_ShouldSucceed test to run for all applicable schema versions so that we don't miss running the tests for older schema versions.

## Related issues
Addresses [issue #].

## Testing

## FHIR Team Checklist

### DIFF
--- a/test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlServerSchemaUpgradeTests.cs
+++ b/test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlServerSchemaUpgradeTests.cs
@@ -4,6 +4,7 @@
 // -------------------------------------------------------------------------------------------------
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Numerics;
 using System.Threading;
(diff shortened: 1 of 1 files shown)


test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlServerSchemaUpgradeTests.cs
```diff
@@ -97,6 +97,17 @@ public async Task GivenASchemaVersion_WhenApplyingDiffTwice_ShouldSucceed(int sc
             }
         }
 
+        public static IEnumerable<object[]> GetSchemaVersions()
+        {
+            foreach (object item in Enum.GetValues(typeof(SchemaVersion)))
+            {
+                if ((int)item >= 7)
```
brendankowitz, line 104: I also like 7, where is this coming from? Is there a system const already defined somewhere?

namadabu: :) I just re-wrote the existing tests which started from version 7. The reason is that our schema upgrade scripts before v7 aren't idempotent. So we only run this test for v7 onwards.

namadabu: Added comment to the test to explain this

LTA-Thinking (approved):

PR #2156, opened 2021-08-10
Allows for search history bundles with PATCH requests

brendankowitz:
## Description
Allows for search history bundles with PATCH requests

## Related issues
Addresses [issue #].

## Testing

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/BundleFactory.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/BundleFactory.cs
@@ -9,6 +9,7 @@ using System.Linq;
 using System.Net;
 using EnsureThat;
 using Hl7.Fhir.Model;
+using Microsoft.Extensions.Logging;
 using Microsoft.Health.Core;
 using Microsoft.Health.Core.Features.Context;
 using Microsoft.Health.Fhir.Core.Extensions;
@@ -25,14 +26,17 @@ namespace Microsoft.Health.Fhir.Core.Features.Search
     {
         private readonly IUrlResolver _urlResolver;
         private readonly RequestContextAccessor<IFhirRequestContext> _fhirRequestContextAccessor;
+        private readonly ILogger<BundleFactory> _logger;
 
-        public BundleFactory(IUrlResolver urlResolver, RequestContextAccessor<IFhirRequestContext> fhirRequestContextAccessor)
+        public BundleFactory(IUrlResolver urlResolver, RequestContextAccessor<IFhirRequestContext> fhirRequestContextAccessor, ILogger<BundleFactory> logger)
         {
             EnsureArg.IsNotNull(urlResolver, nameof(urlResolver));
             EnsureArg.IsNotNull(fhirRequestContextAccessor, nameof(fhirRequestContextAccessor));
+            EnsureArg.IsNotNull(logger, nameof(logger));
 
             _urlResolver = urlResolver;
             _fhirRequestContextAccessor = fhirRequestContextAccessor;
+            _logger = logger;
         }
 
         public ResourceElement CreateSearchBundle(SearchResult result)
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Fhir.Shared.Core/Features/Search/BundleFactory.cs
```diff
@@ -65,21 +69,41 @@ public ResourceElement CreateHistoryBundle(SearchResult result)
                     Url = hasVerb ? $"{r.Resource.ResourceTypeName}/{(httpVerb == Bundle.HTTPVerb.POST ? null : r.Resource.ResourceId)}" : null,
                 };
 
-                string statusString;
+                string statusString = null;
+
+                string ConvertStatusString(HttpStatusCode statusCode)
```
Ivanidzo4ka, line 74: >ConvertStatusString [](http://example.com/codeflow?start=23&length=19)

supernit: Convert`To`StatusString ?

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/BundleFactory.cs
```diff
@@ -65,21 +69,41 @@ public ResourceElement CreateHistoryBundle(SearchResult result)
...
+#endif
+                        statusString = ConvertStatusString(HttpStatusCode.OK);
                         break;
                     case Bundle.HTTPVerb.DELETE:
-                        statusString = ((int)HttpStatusCode.NoContent).ToString() + " " + HttpStatusCode.NoContent;
+                        statusString = ConvertStatusString(HttpStatusCode.NoContent);
                         break;
                     default:
-                        throw new NotImplementedException();
+                        _logger.LogWarning(nameof(CreateHistoryBundle) + " tried to resolve a status string for {httpVerb} but a mapping wasn't defined.", httpVerb);
```
Ivanidzo4ka, line 105: >_logger.LogWarning(nameof(CreateHistoryBundle) + " tried to resolve a status string for {httpVerb} but a mapping wasn't defined.", httpVerb); [](http://example.com/codeflow?start=24&length=141)

is there are reason why logging preferred to throwing exception?

brendankowitz: Throwing an exception here will prevent data being returned. Should one unexpected value invalidate the entire payload? I'm open to being convinced otherwise.

Ivanidzo4ka: At least we monitor exceptions, no one looks into warning logs. 
Exception is a pain, but without pain you will never notice system is not working as intended. 

I guess it's a question of how much we care about history not working fully (in a rare case we start support HEAD verb).

Ivanidzo4ka (approved):

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/BundleFactory.cs
brendankowitz: ...makes sense, I'll refactor the tests to catch if new enums are added. I also don't want to randomly find this in prod again

Ivanidzo4ka:
It's actually still broken for STU3.

I'm not sure why 
```
#if !Stu3
                    case Bundle.HTTPVerb.PATCH:
                    case Bundle.HTTPVerb.HEAD:
#endif
```
was added, but in STU3 patch is possible, and if I execute it I would get cryptic error:
```
 System.NullReferenceException: Object reference not set to an instance of an object.
         at System.Object.GetType()
         at Hl7.Fhir.Utility.EnumUtility.GetLiteral(Enum e)
         at Microsoft.Health.Fhir.Api.Features.Resources.Bundle.BundleSerializer.<>c__DisplayClass3_0.<<Serialize>g__SerializeEntries|2>d.MoveNext() in D:\src\fhir-server\src\Microsoft.Health.Fhir.Shared.Api\Features\Resources\Bundle\BundleSerializer.cs:line 138
```
and bundle would return only partial information.

Ivanidzo4ka:
Let me correct myself. 
https://hl7.org/fhir/STU3/Bundle.html has only following verbs:
`GET | POST | PUT | DELETE`

It's perfectly fine to do patch in STU3 and we store PATCH in our DB, but we can't map it.
In same time, request.method is required field and we fill it up with null value which leads to exception.

brendankowitz:
Looks like there will be other problems in STU3 too, the Method/Url will be null too for Patch:

https://github.com/microsoft/fhir-server/blob/84957b0c924902ac9afc1480c4013eaeedee9196/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/BundleFactory.cs#L64-L71

Ivanidzo4ka:
Yep, I'm leaning towards wrapping resource.Request assignment in `if (has)` to not fill it at all for patch...
Not sure it's a right thing, but it's better than returning half of the bundle.

As another option we can map PATCH to PUT for STU3, but it's a bit meh

brendankowitz:
I think its less broken if PATCH is returned as PUT. Having the URL missing doesn't seem right.

We could consider removing PATCH support from STU3, but that would be considered breaking now

Ivanidzo4ka:
>We could consider removing PATCH support from STU3, but that would be considered breaking now

Let's not -> https://hl7.org/fhir/STU3/http.html#patch  it's a valid operation, it just happened what BundleVerb doesn't have patch verb for STU3.

brendankowitz:
So just a spec oversight then :)

Issue #1973, opened 2021-05-26
[Regression] Sorting DocumentReference by date is no longer supported


giacomodeliberali:
**Describe the bug**
Sorting _DocumentReference_ by `date` was supported (tested in `1.0.800`) but no longer work in the last version (`2.0.41`).

**FHIR Version?**
R4

**Data provider?**
SQL Server

**To Reproduce**
1. Run query `DocumentReference?_sort=date` 

**Expected behavior**
Should sort resources by date asc

**Actual behavior**
Does not sort resources and include the outcome

```json
        {
            "resource": {
                "resourceType": "OperationOutcome",
                "id": "2c773592-90a1-4393-b327-ddfcc5c753ac",
                "issue": [
                    {
                        "severity": "warning",
                        "code": "not-supported",
                        "diagnostics": "Sorting by the 'date' parameter is not supported."
                    }
                ]
            },
            "search": {
                "mode": "outcome"
            }
        }
```

CaitlinV39:
@giacomodeliberali thank you so much for sharing this. While this may have worked in the past, we didn't technically support for date on DocumentReference and so it was not included in our testing. However, we are working on a full implementation of _sort on all strings and dates in SQL which will be available by the end of June and that should resolve this issue.

Thanks!

CaitlinV39:
AB#82462

giacomodeliberali:
Hi @CaitlinV39 , any news on this? In which version a "full implementation of _sort on all strings and dates in SQL" should be available?

CaitlinV39:
Our goal is to have this merged by end of next week, but it may get pushed one week.

CaitlinV39:
We are still merging the feature branch. Goal is middle of August now.

Alyyc85:
Hello,
any news about this issue and its relative milestone?
Thank you

CaitlinV39:
Hi @Alyyc85 I just updated the milestone, we have the PR up and are planning to merge in the next couple of days.

CaitlinV39:
> Hi @Alyyc85 I just updated the milestone, we have the PR up and are planning to merge in the next couple of days.

Just an update that we are still working on merging. We've had some competing priorities so will continue to update the linked PR.

CaitlinV39:
@Alyyc85 @giacomodeliberali This just merged and is main now.

PR #2169, opened 2021-08-16
Support _sort for date and string (SQL)

namadabu:
## Description
Adding support for search queries to _sort using date and string type parameters (for SQL based servers).

## Related issues
Addresses [issue #].

## Testing
Added new UTs and E2E tests.

## FHIR Team Checklist


src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SqlServerSearchParameterStatusDataStore.cs
```diff
@@ -140,7 +141,16 @@ public async Task<IReadOnlyCollection<ResourceSearchParameterStatus>> GetSearchP
...
+                        SearchParameterInfo paramInfo = null;
+                        try
+                        {
+                            paramInfo = _searchParameterDefinitionManager.GetSearchParameter(resourceSearchParameterStatus.Uri);
+                        }
+                        catch (SearchParameterNotSupportedException)
+                        {
+                        }
+
+                        if (paramInfo != null && SqlServerSortingValidator.SupportedSortParamTypes.Contains(paramInfo.Type))
```
namadabu, line 153: Add conditional check based on schema version

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/SortRewriter.cs
```diff
@@ -42,14 +43,121 @@ public override Expression VisitSqlRoot(SqlRootExpression expression, SearchOpti
...
+            }
 
-            var newTableExpressions = new List<SearchParamTableExpression>(expression.SearchParamTableExpressions.Count + 1);
+            var newTableExpressions = new List<SearchParamTableExpression>();
             newTableExpressions.AddRange(expression.SearchParamTableExpressions);
+            var continuationToken = ContinuationToken.FromString(context.ContinuationToken);
+
+            if (!matchFound)
+            {
+                // We are running a sort query where the parameter by which we are sorting
```
namadabu, line 63: Add better comments

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -171,9 +218,28 @@ private async Task<SearchResult> SearchImpl(SearchOptions searchOptions, SqlSear
...
+                            {
+                                lastUpdatedExpression = Expression.GreaterThan(fieldName, null, keyValue);
+                            }
+                            else
+                            {
+                                lastUpdatedExpression = Expression.LessThan(fieldName, null, keyValue);
+                            }
+                        }
+
+                        /*
```
namadabu, line 238: remove

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -93,13 +95,55 @@ internal class SqlServerSearchService : SearchService
 
         public override async Task<SearchResult> SearchAsync(SearchOptions searchOptions, CancellationToken cancellationToken)
         {
-            SearchResult searchResult;
+            SearchResult searchResult = await SearchImpl(searchOptions, SqlSearchType.Default, null, cancellationToken);
+            int resultCount = searchResult.Results.Count();
+            if (!searchOptions.IsSortWithFilter &&
```
namadabu, line 100: Conditional check can be simplified. Add comments

src/Microsoft.Health.Fhir.Core/Features/Search/SearchOptions.cs
```diff
@@ -97,6 +97,16 @@ internal set
...
+        /// <summary>
+        /// Marks whether we need to execute the second set of queries for (certain types of) sort.
+        /// </summary>
+        public bool SortQuerySecondPhase { get; internal set; } = false;
+
+        /// <summary>
+        /// Sets whether this search query is of type sort with filter.
+        /// </summary>
+        public bool IsSortWithFilter { get; internal set; } = false;
+
```
Ivanidzo4ka, line 109: I would argue what both of this properties is SQL specific and shouldn't be part of this object. 

As option we can have SqlSearchOptions derived form SearchOptions and use it inside sql library only

src/Microsoft.Health.Fhir.Tests.Common/TestFiles/R4/DocumentReference-example-relatesTo-code-appends.indexes.json
```diff
@@ -296,4 +296,4 @@
       "isValidAsCompositeComponent": true
     }
   }
-]
```
Ivanidzo4ka, line 299: does changes in this files necessary

namadabu: These files were originally updated as part of this feature branch but then got merged earlier into main. Removed the extra whitespace

src/Microsoft.Health.Fhir.Core/Features/Search/SearchOptions.cs
```diff
@@ -26,6 +26,28 @@ internal SearchOptions()
         {
         }
 
+        internal SearchOptions(SearchOptions other)
```
brendankowitz, line 29: Where is this used?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -1036,6 +1099,74 @@ private static bool IsPrimaryKeySort(SearchOptions searchOptions)
...
+
+            return false;
+        }
+
+        private static void GetSortRelatedDetails(
+            SearchOptions context,
+            out SortOrder sortOrder,
+            out ContinuationToken continuationToken,
+            out object sortValue,
+            out Column sortColumnName)
```
brendankowitz, line 1131: This looks like it should just return an internal Class i.e. SortContext?

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlServerSchemaUpgradeTests.cs
```diff
@@ -229,7 +234,51 @@ private bool CompareDatabaseSchemas(string databaseName1, string databaseName2)
                         (d.TargetObject?.ObjectType.Name == i.type && d.TargetObject?.Name?.ToString() == i.name)))
                 .ToList();
 
-            return remainingDifferences.Count == 0;
+            // Some of the schema changes we are making to tables include addition of columns. In order to support
+            // upgrading older schemas to the newer ones we have to add default constraints to the upgrade script.
+            // These constraints will not be present in databases that were directly initialized with the latest schema.
+            // We need to exclude these constraints from the schema difference comparison.
```
brendankowitz, line 240: Is there a downside to just always having this as a default? Having schemas diverge or behave differently depending on how you got there seems a little scary

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/SortTests.cs
```diff
@@ -480,6 +480,136 @@ public async Task GivenQueryWithObservation_WhenSearchedForItemsWithNoSubjectAnd
...
+        public async Task GivenPatientsWithMultipleNamesAndPaginated_WhenFilteringAndSortingByFamilyName_ThenResourcesAreReturnedInAscendingOrder(int count)
+        {
+            var tag = Guid.NewGuid().ToString();
+            var patients = await CreatePatientsWithMultipleFamilyNames(tag);
+
+            await ExecuteAndValidateBundle($"Patient?_tag={tag}&family=R&_sort=family&_count={count}", sort: false, pageSize: count, patients[0..5]);
+        }
+
+        [Fact]
+        [HttpIntegrationFixtureArgumentSets(dataStores: DataStore.SqlServer)]
```
brendankowitz, line 505: Family should be supported in Cosmos by default, whats the diff for this?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/SortTests.cs
```diff
@@ -480,6 +480,136 @@ public async Task GivenQueryWithObservation_WhenSearchedForItemsWithNoSubjectAnd
...
+
+            await ExecuteAndValidateBundle($"Patient?_tag={tag}&_sort=-family&_count={count}", sort: false, pageSize: count, expectedPatients);
+        }
+
+        [SkippableTheory]
+        [InlineData(2)]
+        [InlineData(3)]
+        [InlineData(4)]
+        [HttpIntegrationFixtureArgumentSets(dataStores: DataStore.CosmosDb)]
+        public async Task GivenPatientsWithFamilyNameMissingAndPaginatedForCosmos_WhenSortingByFamilyNameWithHyphen_ThenThosePatientsAreIncludedInResult(int count)
```
brendankowitz, line 581: We can document why this differs to SQL?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -56,7 +57,8 @@ internal class SqlServerSearchService : SearchService
         private readonly SchemaInformation _schemaInformation;
         private readonly ICompressedRawResourceConverter _compressedRawResourceConverter;
         private readonly RequestContextAccessor<IFhirRequestContext> _requestContextAccessor;
-        private const int _resourceTableColumnCount = 10;
+        private const int _defaultResourceTableFinalSelectColumnCount = 11;
+        private bool? _didWeSearchForSortValue;
```
Ivanidzo4ka, line 61: >_didWeSearchForSortValue [](http://example.com/codeflow?start=22&length=24)

Would it make sense to push it inside SqlSearchOptions class?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -56,7 +57,8 @@ internal class SqlServerSearchService : SearchService
         private readonly SchemaInformation _schemaInformation;
         private readonly ICompressedRawResourceConverter _compressedRawResourceConverter;
         private readonly RequestContextAccessor<IFhirRequestContext> _requestContextAccessor;
-        private const int _resourceTableColumnCount = 10;
+        private const int _defaultResourceTableFinalSelectColumnCount = 11;
```
Ivanidzo4ka, line 60: >Final [](http://example.com/codeflow?start=47&length=5)

What `Final` means?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/BulkDateTimeSearchParameterV2RowGenerator.cs
```diff
@@ -0,0 +1,48 @@
...
+                (searchValue.Start - searchValue.End).Ticks > TimeSpan.TicksPerDay,
+                searchValue.IsMin,
+                searchValue.IsMax);
+
+            return true;
+        }
+
+        protected override void Initialize()
+        {
+            _lastUpdatedSearchParamId = Model.GetSearchParamId(SearchParameterNames.LastUpdatedUri);
```
Ivanidzo4ka, line 45: >Model.GetSearchParamId(SearchParameterNames.LastUpdatedUri) [](http://example.com/codeflow?start=40&length=59)

I can understand usage of Initialize() for something deffered, but we using Model object which we get in constructor.

Can this be part of constructor?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SqlServerSearchParameterStatusDataStore.cs
```diff
@@ -31,30 +32,34 @@ internal class SqlServerSearchParameterStatusDataStore : ISearchParameterStatusD
         private readonly VLatest.UpsertSearchParamsTvpGenerator<List<ResourceSearchParameterStatus>> _updateSearchParamsTvpGenerator;
         private readonly ISearchParameterStatusDataStore _filebasedSearchParameterStatusDataStore;
         private readonly SchemaInformation _schemaInformation;
+        private readonly SqlServerFhirModel _fhirModel;
```
Ivanidzo4ka, line 35: >SqlServerFhirModel [](http://example.com/codeflow?start=25&length=18)

nit: is there are reason to bump from interface to class?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/SearchTestsBase.cs
```diff
@@ -74,18 +64,33 @@ protected async Task<Bundle> ExecuteAndValidateBundle(string searchUrl, string s
...
             Bundle firstBundle = await Client.SearchAsync(searchUrl, customHeader);
 
-            var expectedFirstBundle = expectedResources.Length > pageSize ? expectedResources.ToList().GetRange(0, pageSize).ToArray() : expectedResources;
+            var expectedFirstBundle = expectedResources.Length > pageSize ? expectedResources[0..pageSize] : expectedResources;
 
             ValidateBundle(firstBundle, selfLink, sort, expectedFirstBundle);
 
             var nextLink = firstBundle.NextLink?.ToString();
-            if (nextLink != null)
+            int requestCount = 1;
```
Ivanidzo4ka, line 72: >requestCount [](http://example.com/codeflow?start=16&length=12)

super nit: `pageNumber`? or just `page`

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/SortRewriter.cs
```diff
@@ -42,14 +43,121 @@ public override Expression VisitSqlRoot(SqlRootExpression expression, SearchOpti
...
+                    return new SqlRootExpression(newTableExpressions, expression.ResourceTableExpressions);
+                }
+                else if (continuationToken == null)
+                {
+                    // This means we are in the first "phase" of searching for resources for the _sort query.
+                    // For ascending order, we will search for resources that do not have a value for the
+                    // corresponding _sort parameter.
+                    if (context.Sort[0].sortOrder == SortOrder.Ascending)
+                    {
+                        // This scenarip re
```
Ivanidzo4ka, line 117: >// This scenarip re [](http://example.com/codeflow?start=24&length=19)

nit: something is missing here.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/SortTests.cs
namadabu: The diff here is in the way we break ties. For Cosmos, we always choose the last updated resource (irrespective of the overall sort order). For SQL, we choose the resource based on overall sort order. Hence we see a difference when sorting by Descending order. Have added another test for Cosmos.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/SortTests.cs
namadabu: Added comments

src/Microsoft.Health.Fhir.Core/Features/Search/SearchOptions.cs
namadabu: Added this to use when creating a new `SqlSearchOptions` object using an existing `SearchOptions` object.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/SearchTestsBase.cs
namadabu: renamed to `pageNumber`

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
namadabu: I meant to refer to the "last" SELECT we do on the Resource table before returning the results. Maybe `Final` is not the right word to use...

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/SortTests.cs
brendankowitz: Just for my understanding, why is this different :D? Isn't the secondary sort ResourceSurrogateId, which is in a sense LastUpdated?

src/Microsoft.Health.Fhir.Core/Features/Search/SearchOptions.cs
namadabu: Updated code to have new SqlSearchOptions class.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
namadabu: Updated name to `_defaultNumberOfColumnsReadFromResult`

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/SortRewriter.cs
namadabu: removed

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/SortTests.cs
namadabu: Sorry, I messed up the explanation. I meant to say "oldest" last updated resource (instead of last updated resource). And I had the behavior reversed.

Cosmos always chooses the resource based on last updated time ordered by overall sort order. For SQL, we always choose the "oldest" resource based on last updated time i.e. ResourceSurrogateId ASC

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlServerSchemaUpgradeTests.cs
namadabu: I don't think there should be any downside to having the constraint added to the "fresh" schema (except that it is not actually required). In this situation, I don't think the schema difference would be an issue since the difference is just an additional default rule (and not, say, a different data type). 
I do like the idea of minimizing schema differences as much as possible to help keep things cleaner (especially when we are at v99 :)). I have updated the schema to add these constraints to 18.sql as well.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/BulkDateTimeSearchParameterV2RowGenerator.cs
namadabu: Doesn't seem like it is possible to push this to constructor. The `GetSearchParamId` depends on the Model object being intialized and throws if moved to the ctor.

CaitlinV39:
@namadabu can you confirm that this will address #1973? I just added it but we can remove if it won't.

namadabu:
> @namadabu can you confirm that this will address #1973? I just added it but we can remove if it won't.

Yes, this change will address that issue.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
namadabu: Added an internal class `SortContext` to keep track of these properties

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
namadabu: Moved this into `SqlSearchOptions` class

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -994,6 +1034,19 @@ private void AppendHistoryClause(in IndentedStringBuilder.DelimitedScope delimit
             }
         }
 
+        private void AppendMinOrMax(in IndentedStringBuilder.DelimitedScope delimited, SearchOptions context)
+        {
+            delimited.BeginDelimitedElement();
+            if (context.Sort[0].sortOrder == SortOrder.Ascending)
+            {
+                StringBuilder.Append(VLatest.StringSearchParam.IsMin, tableAlias: null).Append(" = 1");
```
brendankowitz, line 1042: How did we address the issue where records with no value get returned when a sort is applied? i.e. A Patient that didn't have a name entry?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -1036,6 +1089,69 @@ private static bool IsPrimaryKeySort(SearchOptions searchOptions)
...
+            sortContext.ContinuationToken = ContinuationToken.FromString(context.ContinuationToken);
+
+            if (searchParamInfo.Type == ValueSets.SearchParamType.Date)
+            {
+                sortContext.SortColumnName = VLatest.DateTimeSearchParam.StartDateTime;
+            }
+            else if (searchParamInfo.Type == ValueSets.SearchParamType.String)
+            {
+                sortContext.SortColumnName = VLatest.StringSearchParam.Text;
+            }
```
brendankowitz, line 1134: switch statement? Assuming this will keep growing

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -97,15 +98,58 @@ internal class SqlServerSearchService : SearchService
...
+                // We seem to have run a sort which has returned less results than what max we can return.
+                // Let's determine whether we need to execute another query or not.
+                if ((sqlSearchOptions.Sort[0].sortOrder == SortOrder.Ascending && sqlSearchOptions.DidWeSearchForSortValue.HasValue && !sqlSearchOptions.DidWeSearchForSortValue.Value) ||
+                    (sqlSearchOptions.Sort[0].sortOrder == SortOrder.Descending && sqlSearchOptions.DidWeSearchForSortValue.HasValue && sqlSearchOptions.DidWeSearchForSortValue.Value))
+                {
+                    if (sqlSearchOptions.MaxItemCount - resultCount == 0)
+                    {
+                        // Since we are already returning MaxItemCount number of resources we don't want
+                        // to execute another search right now just to drop all the resources. We will return
+                        // a "special" ct so that we the subsequent request will be handled correctly.
```
namadabu, line 120: It's also possible here that there could be no more resources to actually search for. In this scenario a user would end up following the continuation token only to get 0 results.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
LTA-Thinking: In the SortRewriter class it adds searches for resources that are missing the value. If the sort is in Ascending order then resources missing the value are listed first, if the sort is in Descending order then they are listed last. This is to match SQLs default behavior.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
LTA-Thinking: @brendankowitz Which situation would we prefer in this scenario?

- Have the user follow a continuation token and get 0 results
- Run an extra search to make sure there are more results

My preference would be to run the extra search. It is a little extra work on the server, but it saves the user a request/response cycle.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
LTA-Thinking: Refactored with a switch statement.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/SortRewriter.cs
LTA-Thinking: Added another comment, probably could still use more.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -1036,6 +1089,68 @@ private static bool IsPrimaryKeySort(SearchOptions searchOptions)
...
+                    searchParamTableExpression.Kind == SearchParamTableExpressionKind.SortWithFilter)
+                {
+                    return true;
+                }
+            }
+
+            return false;
+        }
+
+        private static SortContext GetSortRelatedDetails(SearchOptions context)
```
brendankowitz, line 1116: nit, naming suggestion
```suggestion
        private static SortContext CreateSortContext(SearchOptions context)
```

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -1036,6 +1089,68 @@ private static bool IsPrimaryKeySort(SearchOptions searchOptions)
...
+            {
+                case ValueSets.SearchParamType.Date:
+                    sortContext.SortColumnName = VLatest.DateTimeSearchParam.StartDateTime;
+                    if (sortContext.ContinuationToken != null)
+                    {
+                        DateTime dateSortValue;
+                        if (DateTime.TryParseExact(sortContext.ContinuationToken.SortValue, "o", null, DateTimeStyles.None, out dateSortValue))
+                        {
+                            sortContext.SortValue = dateSortValue;
+                        }
```
brendankowitz, line 1137: It looks like `DateTimeStyles.None` converts to DateTime (Local). I think we should use DateTimeOffset if possible, or use `DateTimeStyles.RoundtripKind` that converts to DateTime (Utc). 

https://docs.microsoft.com/en-us/dotnet/api/system.datetime.tryparseexact?view=net-5.0#System_DateTime_TryParseExact_System_String_System_String_System_IFormatProvider_System_Globalization_DateTimeStyles_System_DateTime__

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlServerSchemaUpgradeTests.cs
```diff
@@ -229,7 +234,34 @@ private bool CompareDatabaseSchemas(string databaseName1, string databaseName2)
...
+                        }
+                    }
+                }
+                else
+                {
+                    unexpectedDifference = true;
+                }
+            }
+
+            return !unexpectedDifference;
```
brendankowitz, line 264: nit: I think we should move this logic into another function, so its clear that this is special. i.e. `unexpectedDifference = IgnorePartitionColumn(remainingDifferences);`

LTA-Thinking: What about adding a comment before the first foreach? I don't like the feel of having a method that is only used in one place for a non-reusable reason.

PR #2177, opened 2021-08-20
Fix history with PATCH in STU3

Ivanidzo4ka:
## Description
In STU3 PATCH doesn't exist in BundleVerb which leads to null ref because request is empty.
So substitute it to PUT.

## Testing
Manually and updated unit tests.

## FHIR Team Checklist

Patch

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/BundleFactory.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/BundleFactory.cs
@@ -5,7 +5,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Linq;
 using System.Net;
 using EnsureThat;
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Fhir.Shared.Core/Features/Search/BundleFactory.cs
```diff
@@ -61,14 +60,25 @@ public ResourceElement CreateHistoryBundle(SearchResult result)
...
+                // STU3 doesn't have PATCH verb, so let's map it PUT.
+                if (!hasVerb && string.Equals("PATCH", r.Resource.Request?.Method, StringComparison.OrdinalIgnoreCase))
+                {
+                    hasVerb = true;
+                    httpVerb = Bundle.HTTPVerb.PUT;
+                }
+#endif
                 resource.FullUrlElement = new FhirUri(_urlResolver.ResolveResourceWrapperUrl(r.Resource, true));
-                resource.Request = new Bundle.RequestComponent
+                if (hasVerb)
```
LTA-Thinking, line 74: Should we throw an error if hasVerb is false, or will be it caught in the switch case below every time?

LTA-Thinking (approved):

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/BundleFactory.cs
Ivanidzo4ka: Switch statement wouldn't throw exception because it's out parameter and it must be filled with something and for enums it's probably first thing in enum, which is Http.Get.

I'm reluctant to throw, since I'm not sure in which cases we don't have method, so I'll just put creation of response and request under if statement.

PR #2188, opened 2021-08-24
Handle exceptions when schema does not exist

apurvabhaleMS:
## Description
When Schema is not initialized (SchemaOptions:AutomaticUpdatesEnabled : false), SchemaJob, ExportJob, Task hosting background services throw errors when they try to access stored procedures which do not exists as schema is not initialized. With this PR we are trying to avoid multiple exception loggings. Instead we are logging warnings for the same issues.

## Related issues
Addresses [[81886](https://microsofthealth.visualstudio.com/Health/_workitems/edit/81886)].

## Testing
Manual Testing -
1. Set "SqlServer:SchemaOptions:AutomaticUpdatesEnabled" as false in launchSettings.cs.  
2. Run Fhir server
3. Notice there are no errors are logged in console
4. You should see warning loggings for Could not find stored procedure...

## FHIR Team Checklist


src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
```diff
@@ -168,30 +168,39 @@ public async Task<ExportJobOutcome> UpdateExportJobAsync(ExportJobRecord jobReco
 
         public async Task<IReadOnlyCollection<ExportJobOutcome>> AcquireExportJobsAsync(ushort maximumNumberOfConcurrentJobsAllowed, TimeSpan jobHeartbeatTimeoutThreshold, CancellationToken cancellationToken)
```
Ivanidzo4ka, line 169: >AcquireExportJobsAsync [](http://example.com/codeflow?start=65&length=22)

Generic comment.
I think you approach would solve problem, but leave same problem open for future cases.

In cosmosDB we have approach where we have one object   
`Container` to interact with Cosmos and we have `CosmosContainerProvider` which not return `Container` unless it's initialized. 

I would look into option to do something similar approach.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
```diff
@@ -168,30 +168,39 @@ public async Task<ExportJobOutcome> UpdateExportJobAsync(ExportJobRecord jobReco
...
 
-                        acquiredJobs.Add(CreateExportJobOutcome(rawJobRecord, rowVersion));
+                            acquiredJobs.Add(CreateExportJobOutcome(rawJobRecord, rowVersion));
+                        }
                     }
                 }
-
-                return acquiredJobs;
             }
+            catch (Exception ex)
```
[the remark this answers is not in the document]
apurvabhaleMS: Reverted this change

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerTaskConsumer.cs
```diff
@@ -101,47 +101,58 @@ public async Task<TaskInfo> CompleteAsync(string taskId, TaskResultData taskResu
...
+                            InputData = inputData,
+                            Context = taskContext,
+                            Result = result,
+                        };
+
+                        output.Add(taskInfo);
+                    }
+                }
+            }
+            catch (Exception ex)
```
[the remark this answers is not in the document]
apurvabhaleMS: Updated this.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerTaskConsumer.cs
```diff
@@ -101,49 +101,57 @@ public async Task<TaskInfo> CompleteAsync(string taskId, TaskResultData taskResu
         public async Task<IReadOnlyCollection<TaskInfo>> GetNextMessagesAsync(short count, int taskHeartbeatTimeoutThresholdInSeconds, CancellationToken cancellationToken)
         {
             List<TaskInfo> output = new List<TaskInfo>();
-
-            using (SqlConnectionWrapper sqlConnectionWrapper = await _sqlConnectionWrapperFactory.ObtainSqlConnectionWrapperAsync(cancellationToken, true))
-            using (SqlCommandWrapper sqlCommandWrapper = sqlConnectionWrapper.CreateSqlCommand())
+            string storedProcedureName = string.Empty;
```
rbans96, line 104: Are we using this anywhere?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerTaskConsumer.cs
```diff
@@ -101,49 +101,57 @@ public async Task<TaskInfo> CompleteAsync(string taskId, TaskResultData taskResu
...
+                            Context = taskContext,
+                            Result = result,
+                        };
 
-                    output.Add(taskInfo);
+                        output.Add(taskInfo);
+                    }
                 }
             }
+            catch (SqlException e) when (e.Message.Contains("Could not find stored procedure", StringComparison.OrdinalIgnoreCase))
```
rbans96, line 151: I am wondering if we have sqlErrorCode for this exception and can we use that instead?

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Export/ExportJobWorkerTests.cs
```diff
@@ -53,6 +53,7 @@ public ExportJobWorkerTests()
                 _exportJobTaskFactory,
                 NullLogger<ExportJobWorker>.Instance);
 
+            _exportJobWorker.Handle(new Messages.Storage.StorageInitializedNotification(), CancellationToken.None);
```
rbans96, line 56: nit: using Messages.Storage

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobWorker.cs
```diff
@@ -50,30 +53,33 @@ public async Task ExecuteAsync(CancellationToken cancellationToken)
             {
                 try
                 {
-                    // Remove all completed tasks.
-                    runningTasks.RemoveAll(task => task.IsCompleted);
-
-                    // Get list of available jobs.
-                    if (runningTasks.Count < _exportJobConfiguration.MaximumNumberOfConcurrentJobsAllowed)
+                    if (_storageReady)
```
rbans96, line 56: nit: wondering if we can use this check in while statement itself instead of adding one more if block for better code readability?
probably while (!cancellationToken.IsCancellationRequested && _storageReady)

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerTaskConsumer.cs
apurvabhaleMS: It makes sense to check against sqlErrorCode, currently checking with a hardcoded value. Will get the changes in with a separate PR.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobWorker.cs
apurvabhaleMS: There is a delayed wait time logic inside the while loop, but pulled out an "if" statement from the try block

rbans96 (approved):

PR #2192, opened 2021-08-27
Wider catch for exceptions during applying patch.

Ivanidzo4ka:
## Description
JsonPatchDocument can be formatted in a way it would be accepted by server, but during patch it would throw error.
One of the known cases: absence of `path` in operation.
I'm making wide catch statement to catch exceptions and return bad request rather than 500.

## Related issues
Addresses [#84831].

## Testing
Unit test added.

## FHIR Team Checklist

Patch

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Patch/JsonPatchService.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Patch/JsonPatchService.cs
@@ -81,11 +81,6 @@ namespace Microsoft.Health.Fhir.Core.Features.Resources.Patch
                 {
                     throw new BadRequestException($"{operation.op} is invalid.");
                 }
-
-                if (operation.path == null)
-                {
-                    throw new BadRequestException($"{operation.op} has empty path");
-                }
             }
         }
 
@@ -96,6 +91,10 @@ namespace Microsoft.Health.Fhir.Core.Features.Resources.Patch
                 operations.ApplyTo(node.JsonObject);
             }
             catch (JsonPatchException e)
+            {
+                throw new RequestNotValidException(e.Message, OperationOutcomeConstants.IssueType.Processing);
+            }
+            catch (ArgumentNullException e)
             {
                 throw new RequestNotValidException(e.Message, OperationOutcomeConstants.IssueType.Processing);
             }
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Patch/JsonPatchService.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Patch/JsonPatchService.cs
@@ -82,7 +82,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Resources.Patch
             {
                 operations.ApplyTo(node.JsonObject);
             }
-            catch (JsonPatchException e)
+            catch (Exception e)
             {
                 throw new RequestNotValidException(e.Message, OperationOutcomeConstants.IssueType.Processing);
             }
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Patch/JsonPatchService.cs
```diff
@@ -82,7 +82,7 @@ private Resource GetPatchedJsonResource(FhirJsonNode node, JsonPatchDocument ope
             {
                 operations.ApplyTo(node.JsonObject);
             }
-            catch (JsonPatchException e)
+            catch (Exception e)
```
LTA-Thinking, line 85: What is the exception case that the general catch will cover that the JsonPatchException wont? I'm concerned about this catching and passing over an exception we would want logged.

Ivanidzo4ka: ArgmentNullException is one of them.
I also concern with widening catch net, but there is no way to validate JsonPatchDocument and we wrap only one string of code which we have not much control over.

I can add logging if you think it would be useful.

LTA-Thinking: My thought would be to have the specific exceptions (JsonPatch, ArgumentNull, etc) just throw the not valid exception, and have the catch all that still throws the request not valid exception but also logs the exception. That way we are aware of what is causing failures and we can either add it to the list of exceptions that don't log or fix it if it is a real bug.

Ivanidzo4ka: I revert widening, and add validation instead.

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Patch/JsonPatchService.cs
```diff
@@ -74,6 +74,19 @@ private static void Validate(ResourceWrapper currentDoc, WeakETag eTag)
...
             }
+
+            foreach (var operation in patchDocument.Operations)
+            {
+                if (operation.OperationType == AspNetCore.JsonPatch.Operations.OperationType.Invalid)
+                {
+                    throw new BadRequestException($"{operation.op} is invalid.");
+                }
+
+                if (operation.path == null)
```
brendankowitz, line 85: Should this be?
```suggestion
                if (string.IsNullOrWhiteSpace(operation.path))
```

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/PatchTests.cs
```diff
@@ -47,6 +47,28 @@ public async Task GivenAServerThatSupportsIt_WhenSubmittingALevel1PropertyPatch_
...
+            Assert.NotNull(response.Resource.Address);
+
+            string patchDocument =
+                "[{\"op\":\"replace\",\"value\":\"female\"}, {\"op\":\"remove\",\"path\":\"/address\"}]";
+
+            var exception = await Assert.ThrowsAsync<FhirException>(() => _client.PatchAsync(response.Resource, patchDocument));
+            Assert.Equal(HttpStatusCode.BadRequest, exception.Response.StatusCode);
+
+            patchDocument =
+               "[{\"op\":\"coo\",\"path\":\"/gender\",\"value\":\"female\"}, {\"op\":\"remove\",\"path\":\"/address\"}]";
```
brendankowitz, line 66: Convert to Theory?

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Patch/JsonPatchService.cs
Ivanidzo4ka: I think empty path is apply to whole resource, so I think it's valid thing.

Also I don't want to write my own validation for operations, it should come from library, so I'll just try to catch ANE during patching instead of this.

LTA-Thinking (approved):

PR #2217, opened 2021-09-09
Run Patient $everything on "seealso" Links

rotodd:
## Description
As outlined in [the spec](http://hl7.org/fhir/patient-definitions.html#Patient.link), Patient resources can contain links to other Patient or RelatedPerson resources that concern the same actual patient. This PR adds functionality to process these links when the [patient $everything](http://hl7.org/fhir/patient-operation-everything.html) operation is run. Functionality differs depending on the [link type](http://hl7.org/fhir/valueset-link-type.html):

* `seealso`:  run the patient $everything operation on these links as well (note: this will only go one level deep, so we do not run patient $everything on a `seealso` link's `seealso` links)
* `replaced-by`: return an operation outcome that indicates that the patient is no longer a relevant patient, and provide the id for the correct patient
* `replaces`/`refer`: ignore these links

## Related issues
Addresses [AB#82651](https://microsofthealth.visualstudio.com/Health/_workitems/edit/82651).

There is an upcoming task ([AB#85142](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Resolute/Stories/?workitem=85142)) to return RelatedPerson resources that are included in a patient's `seealso` links (currently, they are ignored). 

## Testing
Added e2e tests and an `http` file for manual testing/demos.

## FHIR Team Checklist

Feature (new functionality)

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Exceptions/EverythingOperationException.cs
+++ b/src/Microsoft.Health.Fhir.Core/Exceptions/EverythingOperationException.cs
@@ -11,12 +11,13 @@ namespace Microsoft.Health.Fhir.Core.Exceptions
 {
     public class EverythingOperationException : FhirException
     {
-        public EverythingOperationException(string message, HttpStatusCode httpStatusCode)
+        public EverythingOperationException(string message, HttpStatusCode httpStatusCode, string contentLocationHeaderValue = null)
             : base(message)
         {
             Debug.Assert(!string.IsNullOrEmpty(message), "Exception message should not be empty.");
 
             ResponseStatusCode = httpStatusCode;
+            ContentLocationHeaderValue = contentLocationHeaderValue;
 
             Issues.Add(new OperationOutcomeIssue(
                 OperationOutcomeConstants.IssueSeverity.Error,
@@ -25,5 +26,7 @@ namespace Microsoft.Health.Fhir.Core.Exceptions
         }
 
         public HttpStatusCode ResponseStatusCode { get; }
+
+        public string ContentLocationHeaderValue { get; }
     }
 }
--- a/src/Microsoft.Health.Fhir.Core/Exceptions/EverythingOperationException.cs
+++ b/src/Microsoft.Health.Fhir.Core/Exceptions/EverythingOperationException.cs
@@ -4,21 +4,26 @@
 // -------------------------------------------------------------------------------------------------
 
 using System.Diagnostics;
+using System.Net;
 using Microsoft.Health.Fhir.Core.Models;
 
 namespace Microsoft.Health.Fhir.Core.Exceptions
 {
     public class EverythingOperationException : FhirException
     {
-        public EverythingOperationException(string message)
+        public EverythingOperationException(string message, HttpStatusCode httpStatusCode)
             : base(message)
         {
             Debug.Assert(!string.IsNullOrEmpty(message), "Exception message should not be empty.");
+
+            ResponseStatusCode = httpStatusCode;
 
             Issues.Add(new OperationOutcomeIssue(
                 OperationOutcomeConstants.IssueSeverity.Error,
                 OperationOutcomeConstants.IssueType.Invalid,
                 message));
         }
+
+        public HttpStatusCode ResponseStatusCode { get; }
     }
 }
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/EverythingOperationContinuationToken.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/EverythingOperationContinuationToken.cs
@@ -41,6 +41,12 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Everything
         {
             get
             {
+                // If we are not processing a "seealso" link
+                if (CurrentSeeAlsoLinkIndex < 0)
+                {
+                    return null;
+                }
+
                 return SeeAlsoLinks[CurrentSeeAlsoLinkIndex];
             }
         }
@@ -84,7 +90,16 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Everything
 
         internal void ProcessNextSeeAlsoLink()
         {
-            CurrentSeeAlsoLinkIndex++;
+            // If there are more links to process
+            if (CurrentSeeAlsoLinkIndex < SeeAlsoLinks.Count - 1)
+            {
+                CurrentSeeAlsoLinkIndex++;
+            }
+            else
+            {
+                // This will set the flag that indicates if we are processing "seealso" links to false
+                CurrentSeeAlsoLinkIndex = -1;
+            }
         }
     }
 }
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/EverythingOperationContinuationToken.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/EverythingOperationContinuationToken.cs
@@ -3,25 +3,24 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
-using System.Collections.Generic;
 using Newtonsoft.Json;
 
 namespace Microsoft.Health.Fhir.Core.Features.Operations.Everything
 {
     internal class EverythingOperationContinuationToken
     {
+        // The $everything operation continuation token is used to retrieve the next phase of results when running the
+        // Patient $everything operation. The information stored in this class is serialized, encoded and returned as
+        // part of the "next" URL in the result set bundle.
+        // The length of the serialized token must not exceed the limit of the URL (2083 characters), so we need to
+        // avoid using the continuation token to store large amounts of data. This is why we only store the current
+        // "seealso" link id in the token (opposed to all "seealso" links).
         [JsonConstructor]
         internal EverythingOperationContinuationToken()
         {
             Phase = 0;
             InternalContinuationToken = null;
-
-            SeeAlsoLinks = new List<string>();
-            CurrentSeeAlsoLinkIndex = -1;
         }
-
-        [JsonProperty]
-        internal List<string> SeeAlsoLinks { get; }
 
         [JsonProperty]
         internal int Phase { get; set; }
@@ -29,38 +28,16 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Everything
         [JsonProperty]
         internal string InternalContinuationToken { get; set; }
 
-        internal bool MoreSeeAlsoLinksToProcess
-        {
-            get
-            {
-                return CurrentSeeAlsoLinkIndex < SeeAlsoLinks.Count - 1;
-            }
-        }
-
-        internal string CurrentSeeAlsoLinkId
-        {
-            get
-            {
-                // If we are not processing a "seealso" link
-                if (CurrentSeeAlsoLinkIndex < 0)
-                {
-                    return null;
-                }
-
-                return SeeAlsoLinks[CurrentSeeAlsoLinkIndex];
-            }
-        }
+        [JsonProperty]
+        internal string CurrentSeeAlsoLinkId { get; set; }
 
         internal bool IsProcessingSeeAlsoLink
         {
             get
             {
-                return CurrentSeeAlsoLinkIndex > -1;
+                return CurrentSeeAlsoLinkId != null;
             }
         }
-
-        [JsonProperty]
-        private int CurrentSeeAlsoLinkIndex { get; set; }
 
         public override string ToString()
         {
@@ -86,20 +63,6 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Everything
             }
 
             return token;
-        }
-
-        internal void ProcessNextSeeAlsoLink()
-        {
-            // If there are more links to process
-            if (CurrentSeeAlsoLinkIndex < SeeAlsoLinks.Count - 1)
-            {
-                CurrentSeeAlsoLinkIndex++;
-            }
-            else
-            {
-                // This will set the flag that indicates if we are processing "seealso" links to false
-                CurrentSeeAlsoLinkIndex = -1;
-            }
         }
     }
 }
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/EverythingOperationContinuationToken.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/EverythingOperationContinuationToken.cs
@@ -11,10 +11,10 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Everything
     internal class EverythingOperationContinuationToken
     {
         [JsonConstructor]
-        internal EverythingOperationContinuationToken(int phase, string internalContinuationToken)
+        internal EverythingOperationContinuationToken()
         {
-            Phase = phase;
-            InternalContinuationToken = internalContinuationToken;
+            Phase = 0;
+            InternalContinuationToken = null;
 
             SeeAlsoLinks = new List<string>();
             CurrentSeeAlsoLinkIndex = -1;
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/EverythingOperationContinuationToken.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/EverythingOperationContinuationToken.cs
@@ -39,12 +39,12 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Everything
             }
         }
 
-        public override string ToString()
+        public string ToJson()
         {
             return JsonConvert.SerializeObject(this);
         }
 
-        internal static EverythingOperationContinuationToken FromString(string json)
+        internal static EverythingOperationContinuationToken FromJson(string json)
         {
             if (string.IsNullOrEmpty(json))
             {
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/EverythingOperationContinuationToken.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/EverythingOperationContinuationToken.cs
@@ -3,7 +3,8 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
-using System.Text.Json;
+using System.Collections.Generic;
+using Newtonsoft.Json;
 
 namespace Microsoft.Health.Fhir.Core.Features.Operations.Everything
 {
@@ -13,15 +14,44 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Everything
         {
             Phase = phase;
             InternalContinuationToken = internalContinuationToken;
+
+            SeeAlsoLinks = new List<string>();
+            CurrentSeeAlsoLinkIndex = -1;
+            CurrentSeeAlsoLinkId = null;
         }
 
-        public int Phase { get; private set; }
+        [JsonProperty]
+        private List<string> SeeAlsoLinks { get; }
 
-        public string InternalContinuationToken { get; private set; }
+        [JsonProperty]
+        private int CurrentSeeAlsoLinkIndex { get; set; }
 
-        public static string ToString(int phase, string internalContinuationToken)
+        [JsonProperty]
+        public string CurrentSeeAlsoLinkId { get; private set; }
+
+        public int Phase { get; set; }
+
+        public string InternalContinuationToken { get; set; }
+
+        public bool ProcessingSeeAlsoLink
         {
-            return JsonSerializer.Serialize(new EverythingOperationContinuationToken(phase, internalContinuationToken));
+            get
+            {
+                return !string.IsNullOrEmpty(CurrentSeeAlsoLinkId);
+            }
+        }
+
+        public bool MoreSeeAlsoLinksToProcess
+        {
+            get
+            {
+                return CurrentSeeAlsoLinkIndex < SeeAlsoLinks.Count - 1;
+            }
+        }
+
+        public override string ToString()
+        {
+            return JsonConvert.SerializeObject(this);
         }
 
         public static EverythingOperationContinuationToken FromString(string json)
@@ -31,14 +61,18 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Everything
                 return null;
             }
 
-            try
-            {
-                return JsonSerializer.Deserialize<EverythingOperationContinuationToken>(json);
-            }
-            catch (JsonException)
-            {
-                return null;
-            }
+            return JsonConvert.DeserializeObject<EverythingOperationContinuationToken>(json);
+        }
+
+        public void AddSeeAlsoLink(string link)
+        {
+            SeeAlsoLinks.Add(link);
+        }
+
+        public void ProcessNextSeeAlsoLink()
+        {
+            CurrentSeeAlsoLinkIndex++;
+            CurrentSeeAlsoLinkId = SeeAlsoLinks[CurrentSeeAlsoLinkIndex];
         }
     }
 }
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/EverythingOperationContinuationToken.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/EverythingOperationContinuationToken.cs
@@ -66,7 +66,11 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Everything
 
         public void AddSeeAlsoLink(string link)
         {
-            SeeAlsoLinks.Add(link);
+            // Safe to linear search on addition as it isn't common for a patient to have many "seealso" links
+            if (!SeeAlsoLinks.Contains(link))
+            {
+                SeeAlsoLinks.Add(link);
+            }
         }
 
         public void ProcessNextSeeAlsoLink()
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/EverythingOperationContinuationToken.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/EverythingOperationContinuationToken.cs
@@ -8,40 +8,28 @@ using Newtonsoft.Json;
 
 namespace Microsoft.Health.Fhir.Core.Features.Operations.Everything
 {
-    public class EverythingOperationContinuationToken
+    internal class EverythingOperationContinuationToken
     {
-        public EverythingOperationContinuationToken(int phase, string internalContinuationToken)
+        [JsonConstructor]
+        internal EverythingOperationContinuationToken(int phase, string internalContinuationToken)
         {
             Phase = phase;
             InternalContinuationToken = internalContinuationToken;
 
             SeeAlsoLinks = new List<string>();
             CurrentSeeAlsoLinkIndex = -1;
-            CurrentSeeAlsoLinkId = null;
         }
 
         [JsonProperty]
-        private List<string> SeeAlsoLinks { get; }
+        internal List<string> SeeAlsoLinks { get; }
 
         [JsonProperty]
-        private int CurrentSeeAlsoLinkIndex { get; set; }
+        internal int Phase { get; set; }
 
         [JsonProperty]
-        public string CurrentSeeAlsoLinkId { get; private set; }
+        internal string InternalContinuationToken { get; set; }
 
-        public int Phase { get; set; }
-
-        public string InternalContinuationToken { get; set; }
-
-        public bool ProcessingSeeAlsoLink
-        {
-            get
-            {
-                return !string.IsNullOrEmpty(CurrentSeeAlsoLinkId);
-            }
-        }
-
-        public bool MoreSeeAlsoLinksToProcess
+        internal bool MoreSeeAlsoLinksToProcess
         {
             get
             {
@@ -49,12 +37,31 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Everything
             }
         }
 
+        internal string CurrentSeeAlsoLinkId
+        {
+            get
+            {
+                return SeeAlsoLinks[CurrentSeeAlsoLinkIndex];
+            }
+        }
+
+        internal bool IsProcessingSeeAlsoLink
+        {
+            get
+            {
+                return CurrentSeeAlsoLinkIndex > -1;
+            }
+        }
+
+        [JsonProperty]
+        private int CurrentSeeAlsoLinkIndex { get; set; }
+
         public override string ToString()
         {
             return JsonConvert.SerializeObject(this);
         }
 
-        public static EverythingOperationContinuationToken FromString(string json)
+        internal static EverythingOperationContinuationToken FromString(string json)
         {
             if (string.IsNullOrEmpty(json))
             {
@@ -65,7 +72,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Everything
 
             try
             {
-                 token = JsonConvert.DeserializeObject<EverythingOperationContinuationToken>(json);
+                token = JsonConvert.DeserializeObject<EverythingOperationContinuationToken>(json);
             }
             catch (JsonException)
             {
@@ -75,19 +82,9 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Everything
             return token;
         }
 
-        public void AddSeeAlsoLink(string link)
-        {
-            // Safe to linear search on addition as it isn't common for a patient to have many "seealso" links
-            if (!SeeAlsoLinks.Contains(link))
-            {
-                SeeAlsoLinks.Add(link);
-            }
-        }
-
-        public void ProcessNextSeeAlsoLink()
+        internal void ProcessNextSeeAlsoLink()
         {
             CurrentSeeAlsoLinkIndex++;
-            CurrentSeeAlsoLinkId = SeeAlsoLinks[CurrentSeeAlsoLinkIndex];
         }
     }
 }
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/EverythingOperationContinuationToken.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/EverythingOperationContinuationToken.cs
@@ -61,7 +61,18 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Everything
                 return null;
             }
 
-            return JsonConvert.DeserializeObject<EverythingOperationContinuationToken>(json);
+            EverythingOperationContinuationToken token;
+
+            try
+            {
+                 token = JsonConvert.DeserializeObject<EverythingOperationContinuationToken>(json);
+            }
+            catch (JsonException)
+            {
+                return null;
+            }
+
+            return token;
         }
 
         public void AddSeeAlsoLink(string link)
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
@@ -16,10 +16,10 @@ using Microsoft.Health.Fhir.Core.Exceptions;
 using Microsoft.Health.Fhir.Core.Extensions;
 using Microsoft.Health.Fhir.Core.Features.Definition;
 using Microsoft.Health.Fhir.Core.Features.Persistence;
+using Microsoft.Health.Fhir.Core.Features.Routing;
 using Microsoft.Health.Fhir.Core.Features.Search;
 using Microsoft.Health.Fhir.Core.Features.Search.SearchValues;
 using Microsoft.Health.Fhir.Core.Models;
-using Newtonsoft.Json;
 using CompartmentType = Microsoft.Health.Fhir.ValueSets.CompartmentType;
 
 namespace Microsoft.Health.Fhir.Core.Features.Operations.Everything
@@ -32,6 +32,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Everything
         private readonly ICompartmentDefinitionManager _compartmentDefinitionManager;
         private readonly IReferenceSearchValueParser _referenceSearchValueParser;
         private readonly IResourceDeserializer _resourceDeserializer;
+        private readonly IUrlResolver _urlResolver;
 
         private IReadOnlyList<string> _includes = new[] { "general-practitioner", "organization" };
         private readonly (string resourceType, string searchParameterName) _revinclude = new("Device", "patient");
@@ -42,7 +43,8 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Everything
             ISearchParameterDefinitionManager searchParameterDefinitionManager,
             ICompartmentDefinitionManager compartmentDefinitionManager,
             IReferenceSearchValueParser referenceSearchValueParser,
-            IResourceDeserializer resourceDeserializer)
+            IResourceDeserializer resourceDeserializer,
+            IUrlResolver urlResolver)
         {
             EnsureArg.IsNotNull(searchServiceFactory, nameof(searchServiceFactory));
             EnsureArg.IsNotNull(searchOptionsFactory, nameof(searchOptionsFactory));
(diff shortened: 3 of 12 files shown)


src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
```diff
@@ -177,6 +223,45 @@ public class PatientEverythingService : IPatientEverythingService
...
+                    foreach (Patient.LinkComponent link in links)
+                    {
+                        ReferenceSearchValue referenceSearchValue = _referenceSearchValueParser.Parse(link.Other.Reference);
+
+                        // If the link does not point back to the current patient
+                        if (!string.Equals(referenceSearchValue.ResourceId, resourceId, StringComparison.Ordinal))
+                        {
+                            if (link.Type == Patient.LinkType.ReplacedBy)
+                            {
+                                throw new EverythingOperationException(string.Format(Core.Resources.EverythingOperationResourceIrrelevant, resourceId, referenceSearchValue.ResourceId), HttpStatusCode.BadRequest);
```
rotodd, line 248: Is `BadRequest` the most appropriate status code for this scenario?

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
```diff
@@ -177,6 +223,45 @@ public class PatientEverythingService : IPatientEverythingService
...
+                            }
+
+                            if (link.Type == Patient.LinkType.Seealso)
+                            {
+                                // Links can be of type Patient or RelatedPerson - only attempt to run the $everything operation on Patient resources
```
rotodd, line 253: This will change in [AB#85142](https://microsofthealth.visualstudio.com/Health/_workitems/edit/85142)

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
Ivanidzo4ka: This coming from encrypted token, right?

What I'm as user suppose to do with this message?

Ivanidzo4ka: Maybe we can change it from BadRequest to Moved Permanently (301) and provide updated url in Location header?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/EverythingOperationTestFixture.cs
```diff
@@ -90,6 +101,35 @@ protected override async Task OnInitializedAsync()
...
 
             await TestFhirClient.DeleteAsync(NonExistentPatient);
+
+            // Test case #3
+            // Create patients with different types of links
+            patientToCreate = Samples.GetJsonSample<Patient>("PatientWithMinimalData");
+            patientToCreate.Id = "patient-reference";
+            PatientReferencedByLink = await TestFhirClient.CreateAsync(patientToCreate);
+
+            PatientWithSeeAlsoLink = await CreatePatientWithLink(Patient.LinkType.Seealso, PatientReferencedByLink);
```
Ivanidzo4ka, line 111: >PatientWithSeeAlsoLink [](http://example.com/codeflow?start=12&length=22)

What would happened if I have loop?
A pointing to B, B pointing to A?

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
rotodd: > What I'm as user suppose to do with this message?

The user would know that the patient they are currently running the $everything operation for has been replaced. They would run patient $everything on the id provided in the operation outcome

rotodd: > This coming from encrypted token, right?

I'm not sure what you mean by this exactly!

If it's helpful, here is a bit more context about this line of code: this will be reached in phase 0, after the patient has been retrieved. This method processes the links in the patient resource, and if a `replaced-by` link is found, an operation outcome is returned.

Ivanidzo4ka: resourceId = token.ProcessingSeeAlsoLink ? token.CurrentSeeAlsoLinkId : resourceId;

there is case where resourceId taken from token, and that token produced by us, so it's wrapped in base64, right?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/EverythingOperationTestFixture.cs
rotodd: We only process `seealso` links one layer deep. So, if patient A has a `seealso` link to patient B, we will:
1. run $everything on patient A, storing its `seealso` links
2. Check if there are any `seealso` links left to process --> we find that patient B needs to be processed
2. run $everything on patient B , but do not store its `seealso` links

rotodd: A few more loop scenarios that I covered:

- If patient A has a link to itself, we do not add the link as something that needs to be processed
- If patient A has two links to patient B, we only add one of them to the list of links to be processed

Please let me know if there is anything I missed!

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
```diff
@@ -177,6 +223,45 @@ public class PatientEverythingService : IPatientEverythingService
...
+                                if (string.Equals(referenceSearchValue.ResourceType, ResourceType.Patient.ToString(), StringComparison.Ordinal))
+                                {
+                                    token.AddSeeAlsoLink(referenceSearchValue.ResourceId);
```
Ivanidzo4ka, line 256: >AddSeeAlsoLink [](http://example.com/codeflow?start=42&length=14)

This has potential to blow up (we put that into url, and url are limited resource, while there is no limitation on having lot of seealso links for patient).

Can we have LastSeeAlsoId and just pick patient greater than it?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/EverythingOperationTestFixture.cs
Ivanidzo4ka: I know other people are smarter than me, but I couldn't figure that out from glancing over the code. Would it worth to put some comments about how patient everything works with these links somewhere in a code?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/EverythingOperationTests.cs
```diff
@@ -188,5 +188,73 @@ public async Task GivenMultipleInputParametersSpecified_WhenSomeInvalid_ThenInva
...
+            nextLink = secondBundle.Resource.NextLink.ToString();
+            FhirResponse<Bundle> thirdBundle = await Client.SearchAsync(nextLink);
+            ValidateBundle(thirdBundle, Fixture.PatientReferencedByLink);
+
+            nextLink = thirdBundle.Resource.NextLink.ToString();
+            FhirResponse<Bundle> fourthBundle = await Client.SearchAsync(nextLink);
+
+            // All the patients that have links to the "patient-reference" patient will be in its patient compartment
+            ValidateBundle(fourthBundle, Fixture.PatientWithSeeAlsoLink, Fixture.PatientWithReplacedByLink, Fixture.PatientWithReferLink, Fixture.PatientWithReplacesLink);
+        }
```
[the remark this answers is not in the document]
BoyaWu10: Another case is that when patient A points to patient B and a resource C points to both patient A and B, C will be returned twice in the response as well, one in the compartment search of A and the other in the compartment search of B.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
```diff
@@ -56,9 +68,18 @@ public class PatientEverythingService : IPatientEverythingService
...
-                : EverythingOperationContinuationToken.FromString(ContinuationTokenConverter.Decode(continuationToken));
+            EverythingOperationContinuationToken token;
+
+            try
+            {
+                token = string.IsNullOrEmpty(continuationToken)
+                    ? new EverythingOperationContinuationToken(0, null)
+                    : EverythingOperationContinuationToken.FromString(ContinuationTokenConverter.Decode(continuationToken));
+            }
+            catch (JsonReaderException)
```
BoyaWu10, line 79: JsonSerializationException might be thrown here as well. If the request is "$everything?ct=ew==", the decoded token will be "{" and JsonSerializationException will be thrown when deserializing it.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/EverythingOperationContinuationToken.cs
```diff
@@ -31,14 +61,22 @@ public static EverythingOperationContinuationToken FromString(string json)
...
+            return JsonConvert.DeserializeObject<EverythingOperationContinuationToken>(json);
+        }
+
+        public void AddSeeAlsoLink(string link)
+        {
+            // Safe to linear search on addition as it isn't common for a patient to have many "seealso" links
+            if (!SeeAlsoLinks.Contains(link))
             {
-                return null;
+                SeeAlsoLinks.Add(link);
```
BoyaWu10, line 72: If there are many "seealso" links for one patient, there are chances that the length of the serialized continuation token exceeds the limit of the url. We may need to handle that case.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
BoyaWu10: Maybe we can catch `JsonException` instead.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
```diff
@@ -177,6 +223,45 @@ public class PatientEverythingService : IPatientEverythingService
...
+            if (patientResource.Resource != null)
+            {
+                ResourceElement element = _resourceDeserializer.Deserialize(patientResource.Resource);
+                Patient patient = element.ToPoco<Patient>();
+
+                List<Patient.LinkComponent> links = patient.Link;
+
+                if (links != null)
+                {
+                    foreach (Patient.LinkComponent link in links)
```
BoyaWu10, line 239: What is the expected behavior if the patient has multiple links (like having a "seealso" link and a "replaceby" link at the same time)?

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
rotodd: The token is decoded at the beginning of the `SearchAsync()` method:

```
EverythingOperationContinuationToken token = string.IsNullOrEmpty(continuationToken)
    ? new EverythingOperationContinuationToken(0, null)
    : EverythingOperationContinuationToken.FromString(ContinuationTokenConverter.Decode(continuationToken));
```

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/EverythingOperationTests.cs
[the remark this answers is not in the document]
rotodd: Really good question. In the case where Patient A points to Patient B, I had it designed to consider the patient $everything operation on Patient A separately from the patient $everything operation on Patient B. I'll reach out to @CaitlinV39!

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
rotodd: Ah, thanks for bringing this up. I made the assumption that an error should be returned if one or more `replaced-by` links are present regardless of the other links. I will confirm with @CaitlinV39!

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/EverythingOperationTests.cs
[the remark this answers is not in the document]
CaitlinV39: As Boya mentioned this is why we didn't do it in the first place lol. It would be best if when we do the second everything call if the patient A isn't returned twice. Is that possible. On the second example from Boya, I agree that it would be great to ensure that no duplicates are returned but since this isn't all returned in one query for the post-processing, I can just document that there could be duplicates. 

This is making me wonder if having a flag to not follow the links would be valuable though. So you could do Patient/A/$everything where you could say you want to follow the first level links or you could pass a parameter to not follow the first level links. The first option would ensure no duplicates and the second would give more data, with a small chance of duplicates. If we do that, then I'm probably okay with duplicate Patient A as well. In SQL we can do some refactoring and we might be able to optimize this behavior (not for now, just a thought fo the future).

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
CaitlinV39: Agree with Robin!

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
Ivanidzo4ka: so there is two id's one for Patient/id/everything and there is do for seelso patient.

Patient-id is provided by user, it's part of api call.
seealso-id is provided by us and it's encoded in continuation token.

If this link talks about patient-id, it's ok to send that message to user because user can change id to appropriate one.

But if this is seealso id, user can't do much, because it's outside of user control.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
rotodd: Good catch, thanks! Fixed in 4cf5fbd (and moved the logic to the `EverythingOperationContinuationToken` class).

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/EverythingOperationContinuationToken.cs
```diff
@@ -13,15 +14,44 @@ public EverythingOperationContinuationToken(int phase, string internalContinuati
...
 
-        public int Phase { get; private set; }
+        [JsonProperty]
+        private List<string> SeeAlsoLinks { get; }
+
+        [JsonProperty]
+        private int CurrentSeeAlsoLinkIndex { get; set; }
+
+        [JsonProperty]
+        public string CurrentSeeAlsoLinkId { get; private set; }
```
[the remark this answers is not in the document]
rotodd: Refactored to reduce the number of serialized properties in 45267d2.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
rotodd: I was waiting for someone to comment on this... 😊. I agree. Used a hashset to avoid dups in 45267d2.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
BoyaWu10: Thanks! If so I think we can add some comments in relative codes to describe this behavior.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
BoyaWu10: > I was waiting for someone to comment on this... 😊. I agree. Used a hashset to avoid dups in [45267d2](https://github.com/microsoft/fhir-server/commit/45267d2aa3a2ae940fc786a1f73a8bf8ab1fbf18).

To check duplicates, can we use `token.SeeAlsoLinks.Contains(...)` directly? It looks we don't need to create an extra HashSet.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
```diff
@@ -6,16 +6,20 @@
...
 using Hl7.Fhir.Model;
 using Microsoft.Health.Extensions.DependencyInjection;
 using Microsoft.Health.Fhir.Core.Exceptions;
+using Microsoft.Health.Fhir.Core.Extensions;
 using Microsoft.Health.Fhir.Core.Features.Definition;
 using Microsoft.Health.Fhir.Core.Features.Persistence;
 using Microsoft.Health.Fhir.Core.Features.Search;
+using Microsoft.Health.Fhir.Core.Features.Search.SearchValues;
 using Microsoft.Health.Fhir.Core.Models;
+using Newtonsoft.Json;
```
BoyaWu10, line 22: Can we remove the unused namespaces here?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/EverythingOperationTests.cs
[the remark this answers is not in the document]
BoyaWu10: If we just want to make sure that the patient itself is not returned multiple times, it can be achieved by adding a post-process. This is because if the patient itself is in scope for response, it will always be returned in first page, so we can just filter it out in later pages. But this does not apply for other resources. For example, if Patient A and B "seealso" link to each other, when "Patient/A/$everything" is called, though A is no longer duplicate after adding the post-process, B is still duplicate, one in A's compartment and the other in the "seealso" link.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/EverythingOperationContinuationToken.cs
BoyaWu10: We could consider adding a limit of the maximum number of the "seealso" links. By doing this we can ensure that the continuation token won't be too long to exceed the limit of the url.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/EverythingOperationContinuationToken.cs
```diff
@@ -3,42 +3,88 @@
...
+using Newtonsoft.Json;
 
 namespace Microsoft.Health.Fhir.Core.Features.Operations.Everything
 {
-    public class EverythingOperationContinuationToken
+    internal class EverythingOperationContinuationToken
     {
-        public EverythingOperationContinuationToken(int phase, string internalContinuationToken)
+        [JsonConstructor]
+        internal EverythingOperationContinuationToken(int phase, string internalContinuationToken)
```
BoyaWu10, line 14: This constructor looks a bit confusing since some members are set by input variables while some others are set by default values. We can update this to `EverythingOperationContinuationToken()` and set `Phase` to 0 and `InternalContinuationToken` to null in the constructor. This is the only scenario in our code that it is called.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/EverythingOperationContinuationToken.cs
```diff
@@ -3,42 +3,88 @@
...
+            }
+        }
 
-        public static string ToString(int phase, string internalContinuationToken)
+        internal string CurrentSeeAlsoLinkId
         {
-            return JsonSerializer.Serialize(new EverythingOperationContinuationToken(phase, internalContinuationToken));
+            get
+            {
+                return SeeAlsoLinks[CurrentSeeAlsoLinkIndex];
```
BoyaWu10, line 44: We should check the index before accessing. Otherwise exceptions will be thrown when `CurrentSeeAlsoLinkIndex` is the default value -1.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
rotodd: As discussed, this code is only reached in Phase 0 of the first patient $everything call. I'll add some comments to make this more clear.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
rotodd: Yes, good point. I think the linear check will make more sense once we limit the number of links. Working on this now!

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/EverythingOperationTests.cs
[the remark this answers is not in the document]
rotodd: Thanks for all the thought on this!

RE:
> Another case is that when patient A points to patient B and a resource C points to both patient A and B, C will be returned twice in the response as well, one in the compartment search of A and the other in the compartment search of B.

I had a discussion with @CaitlinV39 and @brendankowitz and confirmed that we will allow duplicates in this scenario

RE:
> In this case where patient A points to patient B, if "Patient/A/$everything" is called, A will be returned twice in the response. It first appears when searching itself. It then appears when searching the compartment of B in the "seealso" link. In previous discussion of patient $everything, we are not going to return duplicate resources in the response. Are we expecting to change this in the update?

I removed the parent link from the `seealso` patient's compartment in c690d1d.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/EverythingOperationContinuationToken.cs
rotodd: Thanks for bringing this up - I agree it would be good to limit the number of `seealso` links. For now, I plan to return an error if the limit is reached, keeping in mind that we are assuming that this isn't a very common scenario (@CaitlinV39 ran this assumption by the Smoke Jumpers).

Also note: @Ivanidzo4ka recommended adding logic to retrieve the parent patient and to extract the remaining links once the limit is reached. I created a story for that: [AB#85289](https://microsofthealth.visualstudio.com/Health/_workitems/edit/85289/)

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
rotodd: Oops yep. Removed in 7a1557b.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/EverythingOperationContinuationToken.cs
rotodd: Consistency is always nice 😊. Fixed in 16ec002.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/EverythingOperationContinuationToken.cs
rotodd: Added some safety checks in a9f916c.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
```diff
@@ -299,16 +414,25 @@ public class PatientEverythingService : IPatientEverythingService
                 searchParameters.Add(Tuple.Create(KnownQueryParameterNames.ContinuationToken, continuationToken));
             }
 
+            // If we are processing a "seealso" link, the parent patient that links to it will be in its patient compartment
+            if (token.IsProcessingSeeAlsoLink)
+            {
+                // Add a parameter to prevent this, since we already returned the parent patient resource in a previous call
+                searchParameters.Add(Tuple.Create(KnownQueryParameterNames.Id, $"ne{parentPatientId}"));
```
BoyaWu10, line 421: Does this prefix `ne` work for string? If the id is "abc", will it search for an id equal to `neabc`?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/EverythingOperationTests.cs
```diff
@@ -188,5 +192,80 @@ public async Task GivenMultipleInputParametersSpecified_WhenSomeInvalid_ThenInva
...
+            FhirResponse<Bundle> secondBundle = await Client.SearchAsync(nextLink);
+            Assert.Empty(secondBundle.Resource.Entry);
+
+            nextLink = secondBundle.Resource.NextLink.ToString();
+            FhirResponse<Bundle> thirdBundle = await Client.SearchAsync(nextLink);
+            ValidateBundle(thirdBundle, Fixture.PatientReferencedByLink);
+
+            nextLink = thirdBundle.Resource.NextLink.ToString();
+            FhirResponse<Bundle> fourthBundle = await Client.SearchAsync(nextLink);
+            Assert.Empty(fourthBundle.Resource.Entry);
```
BoyaWu10, line 215: Can we add one compartment resource here for the "seealso" patient to make sure they are returned as expected?

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
rotodd: Ah yes! This should be using a modifier instead of a prefix. Fixed in f8bd4af.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/EverythingOperationTests.cs
rotodd: Good idea, added in 1d86e74.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
rotodd: Please let me know how the comments look in a8f5b52.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/EverythingOperationContinuationToken.cs
rotodd: I wasn't convinced we had much storage space left in the token, so I ended up fetching the parent patient for each `seealso` link in a8f5b52.

A bit more info: the internal continuation token can be up to 2kB and resource ids can be up to 64 characters (64B). From my research, a URL has a max of 2083 characters, which doesn't give us much extra space (please let me know if I made any false assumptions here).

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/EverythingOperationTests.cs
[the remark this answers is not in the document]
rotodd: Added a comment about this in a8f5b52.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
rotodd: As written below, I ended up only storing the current link id in the token in a8f5b52.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/EverythingOperationTestFixture.cs
rotodd: Added comments in a8f5b52.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
rotodd: Added comments in a8f5b52.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
```diff
@@ -177,6 +238,157 @@ public class PatientEverythingService : IPatientEverythingService
...
+                Tuple.Create(SearchParameterNames.Id, parentPatientId),
+            };
+
+            // And then execute the search.
+            SearchOptions searchOptions = _searchOptionsFactory.Create(ResourceType.Patient.ToString(), searchParameters);
+            SearchResult searchResult = await search.Value.SearchAsync(searchOptions, cancellationToken);
+            searchResultEntries.AddRange(searchResult.Results.Select(x => new SearchResultEntry(x.Resource)));
+
+            // The search result entries should contain one patient resource - the parent patient.
+            SearchResultEntry parentPatientResource = searchResultEntries.FirstOrDefault(s => string.Equals(s.Resource.ResourceTypeName, ResourceType.Patient.ToString(), StringComparison.Ordinal));
```
BoyaWu10, line 308: Do we have a special reason to create `searchResultEntries` here? It is created in `SearchIncludes` to change the search mode from "include" to "match", but it looks not needed in this function. We can consider using `searchResult.Results.FirstOrDefault()` directly.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
```diff
@@ -177,6 +238,157 @@ public class PatientEverythingService : IPatientEverythingService
...
+                // Then the next "seealso" link we process will be the first in the list.
+                token.CurrentSeeAlsoLinkId = seeAlsoLinkIdsSorted.First();
+            }
+            else
+            {
+                // Otherwise, the next "seealso" link we process will be the one that follows the current one in the sorted list.
+                int indexOfCurrent = seeAlsoLinkIdsSorted.FindIndex(id => id == token.CurrentSeeAlsoLinkId);
+                int indexOfNext = indexOfCurrent + 1;
+
+                if (indexOfNext < seeAlsoLinkIdsSorted.Count - 1)
```
BoyaWu10, line 373: My understanding is that the condition should be `indexOfNext < seeAlsoLinkIdsSorted.Count`. Otherwise we will lose the last "seealso" link.

rotodd: Thanks for catching this! Fixed in 6b9eba8 and added a test for it.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
rotodd: Fixed in 3188b31.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
```diff
@@ -177,6 +238,155 @@ public class PatientEverythingService : IPatientEverythingService
...
+                    // Specify the url that must be used instead.
+                    Uri url = _urlResolver.ResolveOperationResultUrl(OperationsConstants.PatientEverything, referenceSearchValue.ResourceId);
+
+                    throw new EverythingOperationException(
+                        string.Format(
+                            Core.Resources.EverythingOperationResourceIrrelevant,
+                            parentPatientId,
+                            referenceSearchValue.ResourceId),
+                        HttpStatusCode.MovedPermanently,
+                        url.ToString());
```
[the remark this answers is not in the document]
rotodd: Good question. I assumed that we could simply return an error as soon as we encounter a `replaced-by` link, since it doesn't make much sense for a patient to have multiple. However, from reading through the spec, it doesn't seem like there is anything that would stop someone from doing so.

One thought: we could return a separate error message if a patient has more than one `replaced-by` link. @CaitlinV39 thoughts?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/EverythingOperationTests.cs
```diff
@@ -8,9 +8,12 @@
 using System.Net;
 using Hl7.Fhir.Model;
 using Microsoft.Health.Fhir.Client;
+using Microsoft.Health.Fhir.Core.Features.Routing;
 using Microsoft.Health.Fhir.Tests.Common.FixtureParameters;
 using Microsoft.Health.Fhir.Tests.E2E.Rest.Search;
 using Microsoft.Health.Test.Utilities;
+using Microsoft.Net.Http.Headers;
+using NSubstitute;
```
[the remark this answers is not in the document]
rotodd: Fixed in 4fac0e4.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
```diff
@@ -177,6 +238,155 @@ public class PatientEverythingService : IPatientEverythingService
...
+            {
+                return;
+            }
+
+            // If a "replaced-by" link is present, it indicates that the patient resource containing this link must no
+            // longer be used. The link points forward to another patient resource that must be used in lieu of the
+            // resource that contains the "replaced-by" link.
+            foreach (Patient.LinkComponent link in links)
+            {
+                ReferenceSearchValue referenceSearchValue = _referenceSearchValueParser.Parse(link.Other.Reference);
```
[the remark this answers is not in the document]
rotodd: Fixed in 4fac0e4.

src/Microsoft.Health.Fhir.Core/Exceptions/EverythingOperationException.cs
```diff
@@ -4,21 +4,29 @@
...
+            ResponseStatusCode = httpStatusCode;
+            ContentLocationHeaderValue = contentLocationHeaderValue;
+
             Issues.Add(new OperationOutcomeIssue(
                 OperationOutcomeConstants.IssueSeverity.Error,
                 OperationOutcomeConstants.IssueType.Invalid,
                 message));
         }
+
+        public HttpStatusCode ResponseStatusCode { get; }
```
brendankowitz, line 28: I don't think we've coupled HttpStatusCode inside Exceptions yet, our codebase currently maps these. I'm fine if we've thought about it, is EverythingOperationException becoming too generic?

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
```diff
@@ -177,6 +238,155 @@ public class PatientEverythingService : IPatientEverythingService
             return new SearchResult(searchResultEntries, searchResult.ContinuationToken, searchResult.SortOrder, searchResult.UnsupportedSearchParameters);
         }
 
+        private void CheckForReplacedByLinks(string parentPatientId, SearchResultEntry parentPatientResource)
```
brendankowitz, line 241: Should these new methods also have EnsureArgs?

rotodd: Do we need EnsureArgs for private methods? Happy to add them in, but I noticed we don't have them for other private methods in the solution.

src/Microsoft.Health.Fhir.Core/Exceptions/EverythingOperationException.cs
rotodd: I noticed that `OperationFailedException` and `FhirTransactionFailedException` have a status code property - however I can definitely make changes if we think `EverythingOperationException` is becoming too generic. Thoughts on the best way to break it down? We could either rely on an existing exception type or create another new exception class (open to either).

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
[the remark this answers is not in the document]
CaitlinV39: I'm fine with whatever is simplest. I think it would be hard to know which would be the "right one". The **best** option is probably to say there are multiple, but I think passing the first one and then if they get another error thats okay. It sounds like a business process inefficiency if a single patient has multiple replaced-by links.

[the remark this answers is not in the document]
rotodd: Sounds good, will keep as-is for now. I created a backlog item that we can prioritize later: [AB#85435](https://microsofthealth.visualstudio.com/Health/_workitems/edit/85435).

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/EverythingOperationContinuationToken.cs
```diff
@@ -3,42 +3,66 @@
...
+        }
+
+        public override string ToString()
         {
-            return JsonSerializer.Serialize(new EverythingOperationContinuationToken(phase, internalContinuationToken));
+            return JsonConvert.SerializeObject(this);
         }
 
-        public static EverythingOperationContinuationToken FromString(string json)
+        internal static EverythingOperationContinuationToken FromString(string json)
```
brendankowitz, line 47: nit: ToJson(), FromJson()?

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
brendankowitz: I think it depends on input, if the value has been previously checked by another method then its fine

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
```diff
@@ -177,6 +238,155 @@ public class PatientEverythingService : IPatientEverythingService
...
+            ResourceElement element = _resourceDeserializer.Deserialize(parentPatientResource.Resource);
+            Patient parentPatient = element.ToPoco<Patient>();
+
+            return parentPatient.Link;
+        }
+
+        private async Task<string> CheckForNextSeeAlsoLinkAndSetToken(string parentPatientId, EverythingOperationContinuationToken token, CancellationToken cancellationToken)
+        {
+            // Retrieve the parent patient so that we can extract its links and process the next "seealso" link.
+            using IScoped<ISearchService> search = _searchServiceFactory();
```
brendankowitz, line 293: nit, prefer the following which more clearly indicates its a delegate:
```suggestion
            using IScoped<ISearchService> search = _searchServiceFactory.Invoke();
```

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
```diff
@@ -177,6 +238,155 @@ public class PatientEverythingService : IPatientEverythingService
...
+
+            // To do so, first create the search parameter to add to the query
+            var searchParameters = new List<Tuple<string, string>>
+            {
+                Tuple.Create(SearchParameterNames.Id, parentPatientId),
+            };
+
+            // And then execute the search.
+            SearchOptions searchOptions = _searchOptionsFactory.Create(ResourceType.Patient.ToString(), searchParameters);
```
brendankowitz, line 302: Nit, can use const:
```suggestion
            SearchOptions searchOptions = _searchOptionsFactory.Create(KnownResourceTypes.Patient, searchParameters);
```

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/EverythingOperationContinuationToken.cs
rotodd: Renamed in 6c96ab8.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
rotodd: Thanks, used const in 902fa60.

rotodd:
Closing to instead merge into a feature branch, as a new work item has come up to add a parameter to control this behaviour (please see [AB#85562](https://microsofthealth.visualstudio.com/Health/_workitems/edit/85562)).

PR #2237, opened 2021-09-21
Adds App Settings to Toggle Keep History

rotodd:
## Description
Creating a draft PR to get some feedback on how to best use the `KeepHistory` app settings value in the capability statement builder to turn history off.

## Related issues
Currently working on [AB#81153](https://microsofthealth.visualstudio.com/Health/_workitems/edit/81153).

## Testing
Tests are still TBD.

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Configs/VersioningConfiguration.cs
+++ b/src/Microsoft.Health.Fhir.Core/Configs/VersioningConfiguration.cs
@@ -1,0 +1,14 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using Microsoft.Health.Fhir.ValueSets;
+
+namespace Microsoft.Health.Fhir.Core.Configs
+{
+    public class VersioningConfiguration
+    {
+        public string Default { get; set; } = ResourceVersionPolicy.Versioned;
+    }
+}
--- a/src/Microsoft.Health.Fhir.Core/Features/Conformance/Models/ListedResourceComponent.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Conformance/Models/ListedResourceComponent.cs
@@ -18,12 +18,10 @@ namespace Microsoft.Health.Fhir.Core.Features.Conformance.Models
             Interaction = new HashSet<ResourceInteractionComponent>(new PropertyEqualityComparer<ResourceInteractionComponent>(x => x.Code));
             SearchParam = new HashSet<SearchParamComponent>(new PropertyEqualityComparer<SearchParamComponent>(x => x.Name, x => x.Type.ToString()));
 
-            // TODO: Not the right place for this? Is there a way to select the other options in the hash set?
-            Versioning = configuration != null && configuration.KeepHistory ? new DefaultOptionHashSet<string>(ResourceVersionPolicy.Versioned, StringComparer.Ordinal) : new DefaultOptionHashSet<string>(ResourceVersionPolicy.NoVersion, StringComparer.Ordinal);
-
             SearchRevInclude = new HashSet<string>(StringComparer.Ordinal);
             SearchInclude = new HashSet<string>(StringComparer.Ordinal);
             ReferencePolicy = new HashSet<string>(StringComparer.Ordinal);
+            Versioning = new DefaultOptionHashSet<string>(configuration.Versioning, StringComparer.Ordinal);
             SupportedProfile = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
             Operation = new HashSet<OperationComponent>(new PropertyEqualityComparer<OperationComponent>(x => x.Name, x => x.Definition.ToString()));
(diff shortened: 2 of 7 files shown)


src/Microsoft.Health.Fhir.Core/Features/Conformance/Models/ListedResourceComponent.cs
```diff
@@ -5,18 +5,22 @@
...
     {
-        public ListedResourceComponent()
+        public ListedResourceComponent(CoreFeatureConfiguration configuration)
         {
             Interaction = new HashSet<ResourceInteractionComponent>(new PropertyEqualityComparer<ResourceInteractionComponent>(x => x.Code));
             SearchParam = new HashSet<SearchParamComponent>(new PropertyEqualityComparer<SearchParamComponent>(x => x.Name, x => x.Type.ToString()));
-            Versioning = new DefaultOptionHashSet<string>("versioned", StringComparer.Ordinal);
+
+            // TODO: Not the right place for this? Is there a way to select the other options in the hash set?
+            Versioning = configuration != null && configuration.KeepHistory ? new DefaultOptionHashSet<string>(ResourceVersionPolicy.Versioned, StringComparer.Ordinal) : new DefaultOptionHashSet<string>(ResourceVersionPolicy.NoVersion, StringComparer.Ordinal);
```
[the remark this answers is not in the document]
feordin: Perhaps we should change the setting in appsettings to a string which could be one of the three values.  Then that string could be validated against the enumeration to ensure it is a valid option?  Then you could simply pass the string in to the constructor?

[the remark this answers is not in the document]
rotodd: Good idea - @brendankowitz also recommended storing the enum value in the constructor, along with making it an object so we can add more info easily later (like resource specific overrides). Working on this now.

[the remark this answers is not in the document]
rotodd: Used the enum value in 6c42f38 and made it an object in 144365b.

src/Microsoft.Health.Fhir.Core/Features/Conformance/Models/ListedResourceComponent.cs
```diff
@@ -5,21 +5,23 @@
...
+using Microsoft.Health.Fhir.Core.Configs;
 using Microsoft.Health.Fhir.Core.Models;
 using Microsoft.Health.Fhir.ValueSets;
 
 namespace Microsoft.Health.Fhir.Core.Features.Conformance.Models
 {
     public class ListedResourceComponent
     {
-        public ListedResourceComponent()
+        public ListedResourceComponent(CoreFeatureConfiguration configuration)
```
rotodd, line 16: @brendankowitz I ended up keeping the config info in this class and passing it into the `DefaultHashSet` constructor. This was a little cleaner than adding a setter to the `DefaultHashSet` class.

rotodd:
@CaitlinV39 @brendankowitz thoughts on how we should handle `_history` searches when we have history disabled? With this implementation, `_history` searches will only return the most recent version (the only version that is stored in the database). Should these searches return an error or have some sort of warning?

src/Microsoft.Health.Fhir.Core/Configs/VersioningConfiguration.cs
```diff
@@ -0,0 +1,14 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using Microsoft.Health.Fhir.ValueSets;
+
+namespace Microsoft.Health.Fhir.Core.Configs
+{
+    public class VersioningConfiguration
+    {
+        public string Default { get; set; } = ResourceVersionPolicy.Versioned;
```
feordin, line 12: Do we want to verify that this string is a valid value from the enumeration?  I am not sure if we check values in the appsettings.json in general, or if we want to do so in this case.  Perhaps if it is an invalid value, we default to "versioned"?

rotodd: Good point, I was thinking about this as I started working on some tests for this... It doesn't look like we validate the values of any other app settings, but I am happy to put in validation if this scenario calls for it. Thoughts @brendankowitz?

rotodd: As discussed, I created a user story for validating all app settings (please see [AB#85362](https://microsofthealth.visualstudio.com/Health/_workitems/edit/85362)).

rotodd:
> @CaitlinV39 @brendankowitz thoughts on how we should handle `_history` searches when we have history disabled? With this implementation, `_history` searches will only return the most recent version (the only version that is stored in the database). Should these searches return an error or have some sort of warning?

[The spec](https://www.hl7.org/fhir/valueset-versioning-policy.html) indicates that we should remove support for versioning all together when history is disabled. As discussed, we will postpone this until later. I created a work item for this here: [AB#85363](https://microsofthealth.visualstudio.com/Health/_workitems/edit/85363).

rotodd:
Closing to create non-draft PR.

PR #2257, opened 2021-10-05
Checks version when fetching "seealso" links

rotodd:
## Description
Previously, if a patient was updated during a patient $everything operation, the latest version of patient would be used to fetch `seealso` links. This could lead to unexpected results if `seealso` links were updated, added or removed. 

This PR adds a version check which ensures that asynchronous $everything operation results always reference the version of the patient captured at the time of the very first API call.

## Related issues
Addresses [AB#85686](https://microsofthealth.visualstudio.com/Health/_workitems/edit/85686).

## Testing
Updates unit tests and adds E2E test.

## FHIR Team Checklist
- ✅ **Update the title** of the PR to be succinct and less than 50 characters
- ✅ **Add a milestone** to the PR for the sprint that it is merged (i.e. add S47)
- ✅ Tag the PR with the type of update: **Bug**, **Dependencies**, **Enhancement**, or **New-Feature**
- ✅ Tag the PR with Azure API for FHIR if this will release to the Azure API for FHIR managed service (CosmosDB or common code related to service)
- ✅ Tag the PR with Azure Healthcare APIs if this will release to the Azure Healthcare APIs managed service (Sql server or common code related to service)
- Review [squash-merge requirements](https://github.com/microsoft/fhir-server/blob/master/SquashMergeRequirements.md)

### Semver Change ([docs](https://github.com/microsoft/fhir-server/blob/master/docs/Versioning.md))
Skip (minor bug fix)

PR #2261, opened 2021-10-06
Adds option to avoid processing $everything links 

rotodd:
## Description
Adds the ability to include an `excludeLinks` parameter to avoid processing a patient's links during a patient $everything operation.

Note: all link types will be ignored, including `replaced-by` links.

## Related issues
Addresses [AB#85562](https://microsofthealth.visualstudio.com/Health/_workitems/edit/85562).

## Testing
Adds e2e tests and a manual test case in the `.http` file.

## FHIR Team Checklist

Feature (new functionality)

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/KnownQueryParameterNames.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/KnownQueryParameterNames.cs
@@ -71,5 +71,7 @@ namespace Microsoft.Health.Fhir.Core.Features
         public const string OutputFormat = "_outputFormat";
 
         public const string TypeFilter = "_typeFilter";
+
+        public const string ExcludeLinks = "excludeLinks";
     }
 }
--- a/src/Microsoft.Health.Fhir.Core/Features/KnownQueryParameterNames.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/KnownQueryParameterNames.cs
@@ -71,7 +71,5 @@ namespace Microsoft.Health.Fhir.Core.Features
         public const string OutputFormat = "_outputFormat";
 
         public const string TypeFilter = "_typeFilter";
-
-        public const string ExcludeLinks = "excludeLinks";
     }
 }
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
@@ -44,7 +44,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Everything
         private readonly IReferenceSearchValueParser _referenceSearchValueParser;
         private readonly IResourceDeserializer _resourceDeserializer;
         private readonly IUrlResolver _urlResolver;
-        private readonly IFhirDataStore _fhirDataStore;
+        private readonly Func<IScoped<IFhirDataStore>> _fhirDataStoreFactory;
 
         private IReadOnlyList<string> _includes = new[] { "general-practitioner", "organization" };
         private readonly (string resourceType, string searchParameterName) _revinclude = new("Device", "patient");
@@ -57,7 +57,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Everything
             IReferenceSearchValueParser referenceSearchValueParser,
             IResourceDeserializer resourceDeserializer,
             IUrlResolver urlResolver,
-            IFhirDataStore fhirDataStore)
+            Func<IScoped<IFhirDataStore>> fhirDataStoreFactory)
         {
             EnsureArg.IsNotNull(searchServiceFactory, nameof(searchServiceFactory));
             EnsureArg.IsNotNull(searchOptionsFactory, nameof(searchOptionsFactory));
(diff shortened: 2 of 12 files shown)


src/Microsoft.Health.Fhir.Core/Features/KnownQueryParameterNames.cs
```diff
@@ -71,5 +71,7 @@ public static class KnownQueryParameterNames
         public const string OutputFormat = "_outputFormat";
 
         public const string TypeFilter = "_typeFilter";
+
+        public const string ExcludeLinks = "excludeLinks";
```
BoyaWu10, line 75: Nit: How about put it in [EverythingParameterNames](https://github.com/microsoft/fhir-server/blob/main/src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/EverythingOperationParameterNames.cs)? This class maintains the unique parameters for $everything operation.

BoyaWu10: And do we need to update the operation definition for patient-everything? Currently it points to FHIR spec in [OperationsCapabilityProvider](https://github.com/microsoft/fhir-server/blob/main/src/Microsoft.Health.Fhir.Shared.Api/Features/Operations/OperationsCapabilityProvider.cs), but the spec does not contain this input parameter.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
```diff
@@ -108,7 +110,7 @@ public class PatientEverythingService : IPatientEverythingService
             {
                 // Phase 0 gets the patient and any resources it references directly.
                 case 0:
-                    searchResult = await SearchIncludes(resourceId, parentPatientId, since, types, token, cancellationToken);
+                    searchResult = await SearchIncludes(resourceId, parentPatientId, since, types, token, !excludeLinks, cancellationToken);
```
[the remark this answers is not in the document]
rotodd: Good point. I took this approach because I thought it made the logic in the `SearchIncludes()` method more readable. I recognize it is a bit of a tradeoff between consistency and readability, but I think it is nice to favour readability in this file since the logic is decently complex. Thoughts?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/EverythingOperationTests.cs
```diff
@@ -298,5 +298,37 @@ public async Task GivenPatientWithReferLink_WhenRunningPatientEverything_ThenLin
...
+        {
+            string searchUrl = $"Patient/{Fixture.PatientWithSeeAlsoLink.Id}/$everything?excludeLinks=true";
+
+            FhirResponse<Bundle> firstBundle = await Client.SearchAsync(searchUrl);
+            ValidateBundle(firstBundle, Fixture.PatientWithSeeAlsoLink);
+
+            var nextLink = firstBundle.Resource.NextLink.ToString();
+            FhirResponse<Bundle> secondBundle = await Client.SearchAsync(nextLink);
+
+            Assert.Empty(secondBundle.Resource.Entry);
```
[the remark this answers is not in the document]
rotodd: The empty results bundle that we see here is a symptom of the asynchronous setup of the $everything operation.

A few more details: the first patient $everything API call picks up the patient resource and any resources it references directly (in this scenario, there is only the patient resource to return). When we are crafting the continuation token at the end of the first API call, we don't know if there are any patient compartment resources or devices. This check will happen in the second API call, and, if there are none, then we return an empty bundle.

However, your comment made me question if we could avoid returning empty bundles mid-operation. I tested this out locally and found that we can avoid this in some cases. I created a bug for this ([AB#85777](https://microsofthealth.visualstudio.com/Health/_workitems/edit/85777)) and will put a separate PR for those changes once I get this PR and PR #2257 merged. Thanks!

src/Microsoft.Health.Fhir.Core/Features/KnownQueryParameterNames.cs
rotodd: > Nit: How about put it in EverythingParameterNames? This class maintains the unique parameters for $everything operation.

Yes, good idea! Moved the const in 6a96e2a.

rotodd: > And do we need to update the operation definition for patient-everything? Currently it points to FHIR spec in [OperationsCapabilityProvider](https://github.com/microsoft/fhir-server/blob/main/src/Microsoft.Health.Fhir.Shared.Api/Features/Operations/OperationsCapabilityProvider.cs), but the spec does not contain this input parameter.

Really good question... I'm not sure. We didn't update the operation definition for export when we added the `_container` parameter (but we did document it [here](https://docs.microsoft.com/en-us/azure/healthcare-apis/data-transformation/export-data#query-parameters)). @CaitlinV39 thoughts?

rotodd: Created a user story to investigate updating the operation definitions for $export and patient $everything: [AB#85805](https://microsofthealth.visualstudio.com/Health/_workitems/edit/85805).

Issue #2148, opened 2021-08-05
Posting a Bundle with incorrect Media Type returns a 500


brendankowitz:
**Describe the bug**
Posting a Bundle with incorrect Media Type returns a 500.

**FHIR Version?**
Stu3/R4/R5

**Data provider?**
CosmosDB

**To Reproduce**
 
```
Request starting HTTP/1.1 POST https://***.azurehealthcareapis.com/ application/x-www-form-urlencoded 14

Route matched with {action = "BatchAndTransactions", controller = "Fhir"}. Executing controller action with signature System.Threading.Tasks.Task`1[Microsoft.AspNetCore.Mvc.IActionResult] BatchAndTransactions(Hl7.Fhir.Model.Resource) on controller Microsoft.Health.Fhir.Api.Controllers.FhirController (Microsoft.Health.Fhir.R4.Api).

System.ArgumentException: The key '  "entry": [ {' is invalid JQuery syntax because it is missing a closing bracket. (Parameter 'key')
   at Microsoft.AspNetCore.Mvc.ModelBinding.JQueryKeyValuePairNormalizer.NormalizeJQueryToMvc(StringBuilder builder, String key)
   at Microsoft.AspNetCore.Mvc.ModelBinding.JQueryKeyValuePairNormalizer.GetValues(IEnumerable`1 originalValues, Int32 valueCount)
   at Microsoft.AspNetCore.Mvc.ModelBinding.JQueryFormValueProviderFactory.AddValueProviderAsync(ValueProviderFactoryContext context)
   at Microsoft.AspNetCore.Mvc.ModelBinding.CompositeValueProvider.CreateAsync(ActionContext actionContext, IList`1 factories)
   at Microsoft.AspNetCore.Mvc.ModelBinding.CompositeValueProvider.TryCreateAsync(ActionContext actionContext, IList`1 factories)
   at Microsoft.AspNetCore.Mvc.Controllers.ControllerBinderDelegateProvider.<>c__DisplayClass0_0.<<CreateBinderDelegate>g__Bind|0>d.MoveNext()
--- End of stack trace from previous location ---
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.<InvokeInnerFilterAsync>g__Awaited|13_0(ControllerActionInvoker invoker, Task lastTask, State next, Scope scope, Object state, Boolean isCompleted)
[...]

PR #2264, opened 2021-10-06
Fix Post search bug

LTA-Thinking:
## Description
When posting a search with a key that contains certain characters a 500 is returned.

## Related issues
Addresses #2148 

## Testing
Manual testing

## FHIR Team Checklist

Patch

### DIFF
--- a/src/Microsoft.Health.Fhir.Api/Features/Routing/SearchPostReroutingMiddleware.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/Routing/SearchPostReroutingMiddleware.cs
@@ -42,6 +42,8 @@ namespace Microsoft.Health.Fhir.Api.Features.Routing
                         request.Query = new QueryCollection(dic);
                     }
 
+                    request.ContentType = null;
+                    request.Form = null;
                     request.Path = request.Path.Value.Substring(0, request.Path.Value.Length - KnownRoutes.Search.Length);
                     request.Method = "GET";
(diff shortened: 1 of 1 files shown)


src/Microsoft.Health.Fhir.Api/Features/Routing/SearchPostReroutingMiddleware.cs
```diff
@@ -42,6 +42,8 @@ public async Task Invoke(HttpContext context)
                         request.Query = new QueryCollection(dic);
                     }
 
+                    request.ContentType = null;
```
PTaladay, line 45: I'm not quite clear on how this resolves the issue with invalid search parameters. This just looks like it clears out the ContentType and Form if one is sent. Is the form considered a search param in this case?

LTA-Thinking: The issue here was that when doing a POST search request with form encoded data the form was being put through a converter by the AspNetCore router after the middleware but before the controller. The converter tried to turn the form keys into JQuery objects, which was triggering the 500. Since we no longer needed the form I removed it here so that the converter wasn't used.

PTaladay: Ok, that makes sense. Thanks for the explanation.

PR #2270, opened 2021-10-07
Fix datetime 500 error

LTA-Thinking:
## Description
Fixes a 500 error when a date with a time zone is passed into a datetime field.

## Related issues
Addresses [85702](https://microsofthealth.visualstudio.com/Health/_workitems/edit/85702)

## Testing
Manual and new unit/e2e tests

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Web/appsettings.json
+++ b/src/Microsoft.Health.Fhir.Shared.Web/appsettings.json
@@ -1,7 +1,7 @@
 ﻿{
     "FhirServer": {
         "Security": {
-            "Enabled": false,
+            "Enabled": true,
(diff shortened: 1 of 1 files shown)


brendankowitz:
> What are people's thoughts on allowing time zones with dates (2021-10-7+02:00)? The reason this was found is because the FHIR validator considers them valid datetimes. I added them as valid datetimes to align with this, but that change isn't needed if it isn't desired.

I don't think it should be valid (https://www.hl7.org/fhir/datatypes.html#date) `There SHALL be no time zone. Dates SHALL be valid dates`

PTaladay (approved):

Issue #2250, opened 2021-10-01
Pagination broken - Next URL more than 2Kb


shahedaAnsari:
**Describe the bug**
The "next" url in the link section of a Bundle resource results in a 404 response. I believe this is related to the URL specified as "next" and its length is greater than 4kb. 
We are using the Azure API for FHIR.

**FHIR Version?**
R4

**Data provider?**
CosmosDB

**To Reproduce**
1. This occurs when we get all resource example- suppose we get list of patient then in next link we get very long string for parameter "ct"

**Expected behavior**
Able to navigate to other pages using next link

**Actual behavior**
Not able to navigate to next link we get following error

The resource you are looking for has been removed, had its name changed, or is temporarily unavailable.

Screenshot for the same

![CosmosDBNextLinkTooLong](https://user-images.githubusercontent.com/40784469/135620736-85ca25fc-fef8-4a59-b388-c952ff0f33bc.PNG)

CaitlinV39:
@shahedaAnsari A couple quick questions. Are you using a proxy in front of the FHIR service or hitting it directly? In addition, are you on the open-source version of the FHIr service or using the Azure API for FHIR? If the open-source version, can you tell me which release you are running?

shahedaAnsari:
> @shahedaAnsari A couple quick questions. Are you using a proxy in front of the FHIR service or hitting it directly? In addition, are you on the open-source version of the FHIr service or using the Azure API for FHIR? If the open-source version, can you tell me which release you are running?

I am directly hitting the API and I am using the open source which has the version 2.0.172

shahedaAnsari:
> @shahedaAnsari A couple quick questions. Are you using a proxy in front of the FHIR service or hitting it directly? In addition, are you on the open-source version of the FHIr service or using the Azure API for FHIR? If the open-source version, can you tell me which release you are running?

I am directly hitting the API, and I am using the open source which has the version 2.0.172

CaitlinV39:
AB#86024

Sameer2810:
Still facing same issue with new build 2.1.22

PR #2279, opened 2021-10-14
Compress continuation tokens

brendankowitz:
## Description
To negate some of the impacts in using Base64 encoding, this PR uses a fast Deflate to reduce the overall size.

Findings for a selection of cases resulted in having the CT near or under the originally requested limit:
```
Original 2044, Continuation token Base64: 2728, Compressed+Base64: 1792
Original 2021, Continuation token Base64: 2696, Compressed+Base64: 1788
Original 621, Continuation token Base64: 828, Compressed+Base64: 464
Original 1929, Continuation token Base64: 2572, Compressed+Base64: 1040
```

## Related issues
Addresses #2250.

## Testing
* Should pass existing tests.
* Adds tests

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Search/ContinuationTokenConverter.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Search/ContinuationTokenConverter.cs
@@ -6,6 +6,7 @@
 using System;
 using System.IO;
 using System.IO.Compression;
+using System.Text;
 using EnsureThat;
 using Microsoft.Health.Fhir.Core.Features.Persistence;
 using Microsoft.IO;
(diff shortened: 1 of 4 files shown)


feordin:
This would only impact results where we generate continuation tokens, but would there be any noticeable impact on performance?

src/Microsoft.Health.Fhir.Core/Features/Search/ContinuationTokenConverter.cs
```diff
@@ -26,7 +44,15 @@ public static string Decode(string encodedContinuationToken)
         public static string Encode(string continuationToken)
         {
             EnsureArg.IsNotEmptyOrWhiteSpace(continuationToken);
-            return Convert.ToBase64String(System.Text.Encoding.UTF8.GetBytes(continuationToken));
+
+            var buffer = System.Text.Encoding.UTF8.GetBytes(continuationToken);
```
feordin, line 48: Do we need any exception handling here?  I am not sure if creating or writing to the DeflateStream can possibly throw an exception

feordin: Seems fine to let those bubble up, which as you saw would be unlikely.

PR #2292, opened 2021-10-19
Fixed AcquireExportJobs when job count>=job limit

PTaladay:
## Description
Fixed issue with the AcquireExportJobs stored procedure where having 0 or less for the calculated limit caused an exception to be thrown. This now works similar to the ReIndexJobs stored procedure where we check if the limit is a positive value and only do work if it is, otherwise we just commit the transtions with no work done. 

## Related issues
Addresses [AB#78024](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/78024).

## Testing
I wasn't able to create a good working test for this stored procedure. I was able to manually validate that this change works by adding in export jobs into the ExportJob table that were still value and then call the stored procedure with a limit of 1 which would result in a negative value, a limit that matches the number of current running jobs which would result in 0 as the limit, and a limit higher than the number of running jobs which would result in a positive value. All 3 are now successful.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Microsoft.Health.Fhir.SqlServer.csproj
+++ b/src/Microsoft.Health.Fhir.SqlServer/Microsoft.Health.Fhir.SqlServer.csproj
@@ -53,11 +53,7 @@
       <MutableClassVersion>17</MutableClassVersion>
     </EmbeddedResource>
     <EmbeddedResource Include="Features\Schema\Migrations\18.diff.sql" />
-    <EmbeddedResource Include="Features\Schema\Migrations\18.sql">
-      <InputToImmutableSqlGenerator>true</InputToImmutableSqlGenerator>
-      <InputToMutableSqlGenerator>true</InputToMutableSqlGenerator>
-      <MutableClassVersion>18</MutableClassVersion>
-    </EmbeddedResource>
+    <EmbeddedResource Include="Features\Schema\Migrations\18.sql" />
     <EmbeddedResource Include="Features\Schema\Migrations\19.diff.sql" />
     <EmbeddedResource Include="Features\Schema\Migrations\19.sql">
       <InputToImmutableSqlGenerator>true</InputToImmutableSqlGenerator>


src/Microsoft.Health.Fhir.SqlServer/Microsoft.Health.Fhir.SqlServer.csproj
```diff
@@ -3,6 +3,12 @@
     <TargetFramework>net5.0</TargetFramework>
   </PropertyGroup>
   <ItemGroup>
+    <EmbeddedResource Include="Features\Schema\Migrations\19.sql">
```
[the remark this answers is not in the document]
PTaladay: Done

src/Microsoft.Health.Fhir.SqlServer/Microsoft.Health.Fhir.SqlServer.csproj
```diff
@@ -56,8 +56,14 @@
...
       <InputToImmutableSqlGenerator>true</InputToImmutableSqlGenerator>
       <InputToMutableSqlGenerator>true</InputToMutableSqlGenerator>
+      <MutableClassVersion>18</MutableClassVersion>
+    </EmbeddedResource>
+    <EmbeddedResource Include="Features\Schema\Migrations\19.sql">
+      <InputToImmutableSqlGenerator>true</InputToImmutableSqlGenerator>
       <MutableClassVersion>Latest</MutableClassVersion>
+      <InputToMutableSqlGenerator>true</InputToMutableSqlGenerator>
     </EmbeddedResource>
+    <EmbeddedResource Include="Features\Schema\Migrations\19.diff.sql" />
```
rotodd, line 66: Super nit: can we make the ordering here consistent with the rest of the file?

PTaladay: The order has been updated.

PR #2314, opened 2021-11-02
Add exception logging and timestamps

LTA-Thinking:
## Description
Adds logging to the operation outcome attribute to ensure that exceptions are logged. 
Timestamps have also been added to operation outcomes so that when developers are given customer issues there is a timestamp built into the error.

## FHIR Team Checklist


PTaladay (approved):

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/OperationOutcomeExceptionFilterAttribute.cs
```diff
@@ -60,6 +64,10 @@ public override void OnActionExecuted(ActionExecutedContext context)
                     {
                         Id = _fhirRequestContextAccessor.RequestContext.CorrelationId,
                         Issue = fhirException.Issues.Select(x => x.ToPoco()).ToList(),
+                        Meta = new Meta()
+                        {
+                            LastUpdated = DateTimeOffset.UtcNow,
```
brendankowitz, line 69: nit: Prefer:
```suggestion
                            LastUpdated = Clock.UtcNow,
```

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/OperationOutcomeExceptionFilterAttribute.cs
```diff
@@ -235,6 +247,31 @@ public override void OnActionExecuted(ActionExecutedContext context)
...
+                    HttpStatusCode.InternalServerError);
+                context.ExceptionHandled = true;
+            }
+
+            if (context.ExceptionHandled)
+            {
+                HttpStatusCode? statusCode = (context.Result as OperationOutcomeResult).StatusCode;
+                if (statusCode != null && statusCode >= HttpStatusCode.InternalServerError)
+                {
+                    _logger.LogError(_fhirRequestContextAccessor.RequestContext.CorrelationId, context.Exception);
```
brendankowitz, line 273: What overload is this? Normally the exception is logged first? i.e. https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loggerextensions.logerror?view=dotnet-plat-ext-5.0#Microsoft_Extensions_Logging_LoggerExtensions_LogError_Microsoft_Extensions_Logging_ILogger_System_Exception_System_String_System_Object___

LTA-Thinking: It is the first one. I'm using the correlation id as an event id.

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/OperationOutcomeExceptionFilterAttribute.cs
```diff
@@ -235,6 +248,31 @@ public override void OnActionExecuted(ActionExecutedContext context)
...
+                            LastUpdated = Clock.UtcNow,
+                        },
+                    },
+                    HttpStatusCode.InternalServerError);
+                context.ExceptionHandled = true;
+            }
+
+            if (context.ExceptionHandled)
+            {
+                HttpStatusCode? statusCode = (context.Result as OperationOutcomeResult).StatusCode;
```
brendankowitz, line 271: `as` can return null if the cast fails

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/OperationOutcomeExceptionFilterAttribute.cs
```diff
@@ -202,12 +211,7 @@ public override void OnActionExecuted(ActionExecutedContext context)
                         healthExceptionResult = CreateOperationOutcomeResult(microsoftHealthException.Message, OperationOutcome.IssueSeverity.Error, OperationOutcome.IssueType.Invalid, HttpStatusCode.RequestHeaderFieldsTooLarge);
                         break;
                     default:
-                        healthExceptionResult = new OperationOutcomeResult(
-                            new OperationOutcome
-                            {
-                                Id = _fhirRequestContextAccessor.RequestContext.CorrelationId,
-                            },
-                            HttpStatusCode.InternalServerError);
+                        healthExceptionResult = CreateOperationOutcomeResult(microsoftHealthException.Message, OperationOutcome.IssueSeverity.Error, OperationOutcome.IssueType.Unknown, HttpStatusCode.InternalServerError);
```
Ivanidzo4ka, line 214: >(microsoftHealthException.Message [](http://example.com/codeflow?start=76&length=33)

I tend to think there was a reason to hide internal guts and not show them to an user.  Any reason why we want to show message nowadays?

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/OperationOutcomeExceptionFilterAttribute.cs
```diff
@@ -235,6 +239,22 @@ public override void OnActionExecuted(ActionExecutedContext context)
...
+                context.Result = CreateOperationOutcomeResult(context.Exception.Message, OperationOutcome.IssueSeverity.Error, OperationOutcome.IssueType.Unknown, HttpStatusCode.InternalServerError);
+                context.ExceptionHandled = true;
+            }
+
+            if (context.ExceptionHandled)
+            {
+                HttpStatusCode? statusCode = (context.Result as OperationOutcomeResult)?.StatusCode;
+                if (statusCode != null && statusCode >= HttpStatusCode.InternalServerError)
+                {
+                    _logger.LogError(context.Exception, "5xx error returned");
```
Ivanidzo4ka, line 256: >5xx [](http://example.com/codeflow?start=57&length=3)

Would it worth to make a parameter rather than just text?
I can see world where you would like to filter by error codes.

src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/OperationOutcomeExceptionFilterAttribute.cs
LTA-Thinking: That is a fair point. I made this change for internal consistency, but I can alter the CreateOperationOutcomeResult method to just make message optional.

LTA-Thinking: Used string.empty instead, less refactoring.

Ivanidzo4ka (approved):
:shipit:

PR #2324, opened 2021-11-05
Search/sort when "Prefer handling=strict"

### LINKED ISSUE #2208: Prefer handling doesn't applied to sort
**Describe the bug**
Query /Patient?_sort=family with header
`Prefer` `handling=strict`

I expect BadRequest but I get 200 with operation outcome.

**FHIR Version?**
Stu3/R4/R5

**Data provider?**
CosmosDB/SQL Server

**Expected behavior**
400 error.

**Actual behavior**
200 with operation outcome.
```
"resource": {
                "resourceType": "OperationOutcome",
                "id": "cf173395-e168-40a7-bf19-25cae93edc73",
                "issue": [
                    {
                        "severity": "warning",
                        "code": "not-supported",
                        "diagnostics": "Sorting by the 'family' parameter is not supported."

```

Tom-Markoch:
## Description
Changes sort so it returns 400 with error message if
Prefer handling=strict. Adds unit tests to check both cases
Prefer handling=strict and Prefer handling=lenient return proper error
code and OperationOutcome, with deep compare of returned and expected
OperationOutcome.

## Related issues
Fixes #2208

## Testing
Unit tests Microsoft.Health.Fhir.R4.Tests.E2E were ran.

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Persistence/BadRequestException.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Persistence/BadRequestException.cs
@@ -3,6 +3,7 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
+using System.Collections.Generic;
 using Microsoft.Health.Fhir.Core.Exceptions;
 using Microsoft.Health.Fhir.Core.Models;
 
@@ -17,6 +18,17 @@ namespace Microsoft.Health.Fhir.Core.Features.Persistence
                     OperationOutcomeConstants.IssueSeverity.Error,
                     OperationOutcomeConstants.IssueType.Invalid,
                     errorMessage));
+        }
+
+        public BadRequestException(IEnumerable<string> errorMessages)
+        {
+            foreach (string errorMessage in errorMessages)
+            {
+                Issues.Add(new OperationOutcomeIssue(
+                        OperationOutcomeConstants.IssueSeverity.Error,
+                        OperationOutcomeConstants.IssueType.Invalid,
+                        errorMessage));
+            }
         }
     }
 }
--- a/src/Microsoft.Health.Fhir.Core/Features/Persistence/BadRequestException.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Persistence/BadRequestException.cs
@@ -3,7 +3,9 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
+using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using Microsoft.Health.Fhir.Core.Exceptions;
 using Microsoft.Health.Fhir.Core.Models;
@@ -27,7 +29,12 @@ namespace Microsoft.Health.Fhir.Core.Features.Persistence
         /// <param name="errorMessages">Collection of error strings that are used to initialize OperationOutcomeIssue collection.
         /// Must be non-null and non-empty</param>
         public BadRequestException(ICollection<string> errorMessages)
-            : base((errorMessages == null || errorMessages.Count < 2) ? errorMessages.First() : "Multiple bad request errors.")
+            : base(((Func<string>)(() =>
+            {
+                Debug.Assert(errorMessages != null, "Parameter errorMessages must not be null.");
+                Debug.Assert(errorMessages == null || errorMessages.Any(), "Parameter errorMessages must not be empty.");
+                return (errorMessages == null || errorMessages.Count < 2) ? errorMessages.First() : "Multiple bad request errors."; // "Multiple bad request errors." message is used only internally, no need to add it to the Resources.resx file.
+            }))())
         {
             foreach (string errorMessage in errorMessages)
             {
--- a/src/Microsoft.Health.Fhir.Core/Features/Persistence/BadRequestException.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Persistence/BadRequestException.cs
@@ -27,7 +27,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Persistence
         /// <param name="errorMessages">Collection of error strings that are used to initialize OperationOutcomeIssue collection.
         /// Must be non-null and non-empty</param>
         public BadRequestException(ICollection<string> errorMessages)
-            : base((errorMessages.Count < 2) ? errorMessages.First() : "Multiple bad request errors.")
+            : base((errorMessages == null || errorMessages.Count < 2) ? errorMessages.First() : "Multiple bad request errors.")
         {
             foreach (string errorMessage in errorMessages)
             {
--- a/src/Microsoft.Health.Fhir.Core/Features/Persistence/BadRequestException.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Persistence/BadRequestException.cs
@@ -4,6 +4,7 @@
 // -------------------------------------------------------------------------------------------------
 
 using System.Collections.Generic;
+using System.Linq;
 using Microsoft.Health.Fhir.Core.Exceptions;
 using Microsoft.Health.Fhir.Core.Models;
 
@@ -20,7 +21,13 @@ namespace Microsoft.Health.Fhir.Core.Features.Persistence
                     errorMessage));
         }
 
-        public BadRequestException(IEnumerable<string> errorMessages)
+        /// <summary>
+        /// Constructor creates BadRequestException from collection of error strings.
+        /// </summary>
+        /// <param name="errorMessages">Collection of error strings that are used to initialize OperationOutcomeIssue collection.
+        /// Must be non-null and non-empty</param>
+        public BadRequestException(ICollection<string> errorMessages)
+            : base((errorMessages.Count < 2) ? errorMessages.First() : "Multiple bad request errors.")
         {
             foreach (string errorMessage in errorMessages)
             {
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
@@ -311,7 +311,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Search
                 var sortings = new List<(SearchParameterInfo, SortOrder)>(searchParams.Sort.Count);
                 bool sortingsValid = true;
 
-                // Check if sort parameters are valid search parameters.
+                // Only parameters that are valid for searching can also be used for sorting. Therefore first check if sort parameters are valid as search parameters.
                 foreach ((string, Hl7.Fhir.Rest.SortOrder) sorting in searchParams.Sort)
                 {
                     try
@@ -326,7 +326,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Search
                     }
                 }
 
-                // Check if sort parameters are valid and can be used for sorting.
+                // Sort parameters are valid search parameters. Now verify that sort parameters are also valid for sorting.
                 if (sortingsValid)
                 {
                     if (!_sortingValidator.ValidateSorting(sortings, out IReadOnlyList<string> errorMessages))
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
@@ -386,6 +386,19 @@ namespace Microsoft.Health.Fhir.Core.Features.Search
                             string.Format(CultureInfo.InvariantCulture, Core.Resources.SearchParameterNotSupported, unsupported.Item1, string.Join(",", resourceTypesString))));
                 }
 
+                /*if (unsupportedSearchParameters.Any())
+                {
+                    string error = string.Format(
+                            Core.Resources.SearchParameterNotSupported,
+                            string.Join(",", unsupportedSearchParameters.Select(x => x.Item1)),
+                            string.Join(",", resourceTypesString));
+
+                    _contextAccessor.RequestContext?.BundleIssues.Add(new OperationOutcomeIssue(
+                            OperationOutcomeConstants.IssueSeverity.Warning,
+                            OperationOutcomeConstants.IssueType.NotSupported,
+                            error));
+                }*/
+
                 foreach (string unsupported in searchSortErrors)
                 {
                     _contextAccessor.RequestContext?.BundleIssues.Add(new OperationOutcomeIssue(
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
@@ -326,7 +326,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Search
                     }
                 }
 
-                // Sort parameter values are valid search parameters. Now verify that sort paremeter values are also valid for sorting.
+                // Sort parameter values are valid search parameters. Now verify that sort parameter values are also valid for sorting.
                 if (sortingsValid)
                 {
                     if (!_sortingValidator.ValidateSorting(sortings, out IReadOnlyList<string> errorMessages))
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
@@ -361,15 +361,13 @@ namespace Microsoft.Health.Fhir.Core.Features.Search
             // depending on the state of the "Prefer" header.
             if (unsupportedSearchParameters.Any() || searchSortErrors.Any())
             {
-                // Put all errors into a set to avoid duplicate error messages, because search code and sort code may generate same error string.
-                // For example if non-searchable parameter is used both as a search and as a sort parameter then two same "non-searchable error" strings will be generated.
-                var allErrors = new SortedSet<string>();
+                var allErrors = new List<string>();
                 foreach (Tuple<string, string> unsupported in unsupportedSearchParameters)
                 {
                     allErrors.Add(string.Format(CultureInfo.InvariantCulture, Core.Resources.SearchParameterNotSupported, unsupported.Item1, string.Join(",", resourceTypesString)));
                 }
 
-                allErrors.UnionWith(searchSortErrors);
+                allErrors.AddRange(searchSortErrors);
 
                 if (_contextAccessor.GetIsStrictHandlingEnabled())
                 {
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
@@ -357,54 +357,32 @@ namespace Microsoft.Health.Fhir.Core.Features.Search
                 searchOptions.Sort = Array.Empty<(SearchParameterInfo searchParameterInfo, SortOrder sortOrder)>();
             }
 
-            // Processing of parameters is finished. If any of the parameters are unsupported warning is put into the bundle or error is returned,
+            // Processing of parameters is finished. If any of the parameters are unsupported warning is put into the bundle or exception is thrown,
             // depending on the state of the "Prefer" header.
             if (unsupportedSearchParameters.Any() || searchSortErrors.Any())
             {
+                // Put all errors into a set to avoid duplicate error messages, because search code and sort code may generate same error string.
+                // For example if non-searchable parameter is used both as a search and as a sort parameter then two same "non-searchable error" strings will be generated.
+                var allErrors = new SortedSet<string>();
+                foreach (Tuple<string, string> unsupported in unsupportedSearchParameters)
+                {
+                    allErrors.Add(string.Format(CultureInfo.InvariantCulture, Core.Resources.SearchParameterNotSupported, unsupported.Item1, string.Join(",", resourceTypesString)));
+                }
+
+                allErrors.UnionWith(searchSortErrors);
+
                 if (_contextAccessor.GetIsStrictHandlingEnabled())
                 {
-                    var errors = new List<string>();
-                    if (unsupportedSearchParameters.Any())
-                    {
-                        errors.Add(string.Format(
-                                Core.Resources.SearchParameterNotSupported,
-                                string.Join(",", unsupportedSearchParameters.Select(x => x.Item1)),
-                                string.Join(",", resourceTypesString)));
-                    }
-
-                    errors.AddRange(searchSortErrors);
-
-                    throw new BadRequestException(errors);
+                    throw new BadRequestException(allErrors);
                 }
 
                 // There is no "Prefer" header with handling value, or handling value is valid and not set to "Prefer: handling=strict".
-                foreach (Tuple<string, string> unsupported in unsupportedSearchParameters)
+                foreach (string error in allErrors)
                 {
-                    _contextAccessor.RequestContext?.BundleIssues.Add(new OperationOutcomeIssue(
-                            OperationOutcomeConstants.IssueSeverity.Warning,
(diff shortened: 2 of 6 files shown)


src/Microsoft.Health.Fhir.Core/Features/Persistence/BadRequestException.cs
```diff
@@ -18,5 +19,16 @@ public BadRequestException(string errorMessage)
                     OperationOutcomeConstants.IssueType.Invalid,
                     errorMessage));
         }
+
+        public BadRequestException(IEnumerable<string> errorMessages)
+        {
```
LTA-Thinking, line 24: I feel that some string should be passed to the base constructor because that is what will populate the exception message. Maybe something generic like "Multiple bad request errors"?

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
```diff
@@ -341,17 +322,16 @@ public SearchOptions Create(string compartmentType, string compartmentId, string
                     catch (SearchParameterNotSupportedException)
                     {
                         sortingsValid = false;
-                        _contextAccessor.RequestContext?.BundleIssues.Add(new OperationOutcomeIssue(
-                            OperationOutcomeConstants.IssueSeverity.Warning,
-                            OperationOutcomeConstants.IssueType.NotSupported,
-                            string.Format(CultureInfo.InvariantCulture, Core.Resources.SearchParameterNotSupported, sorting.Item1, string.Join(", ", resourceTypesString))));
+                        searchSortErrors.Add(string.Format(CultureInfo.InvariantCulture, Core.Resources.SearchParameterNotSupported, sorting.Item1, string.Join(", ", resourceTypesString)));
```
LTA-Thinking, line 325: Should we be handing multiple bad sort parameters the same way we handle multiple bad search parameters? Instead of one error per bad parameters we have one error that concatenates all the bad parameters into one message.

src/Microsoft.Health.Fhir.Core/Features/Persistence/BadRequestException.cs
Tom-Markoch: What about constructor parameter validation (null, non empty list, etc)? I see the BadRequestException class is public but it is used only internally so I guess it is not necessary, I'll just add a comment.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
Tom-Markoch: https://github.com/microsoft/fhir-server/blob/410600f34903bd7a9ff9a9ef525d4d5b1327fbdb/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs#L306-L325

Current code behaves differently for errors and warnings, shouldn't it behave the same for all the cases? For errors it concatenates and for warnings it creates separate OperationOutcomeIssue for each incorrect parameter. DO you really think we should just copy the current pattern or fix the existing code as well?

Tom-Markoch: My opinion is it should behave same for errors and warnings.

LTA-Thinking: I agree with that. I don't know what the original thoughts were on handling them differently. My preference would be for one operation outcome, instead of multiple. @brendankowitz What are your thoughts?

brendankowitz: We may end up with multiple issues anyway (can someone confirm?) i.e. unsupported modifiers, search parameter types, truncated includes, etc. 
We will only have one OperationOutcome object, but it can contain multiple issues, I think one issue per problem seems ok

src/Microsoft.Health.Fhir.Core/Features/Persistence/BadRequestException.cs
Tom-Markoch: Fixed, constructor will set exception message "Multiple bad request errors." in case of string collection containing multiple errors. If collection contains only one error, exception message will be set to the error string in the collection. No direct param validation since exception object is used only internally.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
Tom-Markoch: Fixed preexisting code. Now all errors/warnings/sort/search are handled in the same way. No concatenation, each issue gets it's own OperationOutcomeIssue instance. Reason is because multiple errors strings of same type could be created in very different places like search code and then again sort code. Concatenating would mean that already generated error strings would have to be parsed so they can be grouped by error type, then parameters in the strings would have to be extracted and then concatenated and then put into a new error string of same error type,

brendankowitz:
Related to #2344

LTA-Thinking (approved):

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/SearchTestsBase.cs
```diff
@@ -96,12 +120,25 @@ protected async Task<Bundle> ExecuteAndValidateBundle(string searchUrl, bool sor
...
+            string searchUrl,
+            Tuple<string, string> customHeader,
+            HttpStatusCode expectedStatusCode,
+            OperationOutcome expectedOperationOutcome)
+        {
+            FhirException ex = await Assert.ThrowsAsync<FhirException>(() => Client.SearchAsync(searchUrl, customHeader));
+            Assert.Equal(expectedStatusCode, ex.StatusCode);
+            ex.OperationOutcome.Id = null;
+            Assert.True(expectedOperationOutcome.IsExactly(ex.OperationOutcome), "Deep compare detected expected and actual OperationOutcome mismatch.");
+            return ex.OperationOutcome;
```
rotodd, line 133: Nit: since this method is only called once, could we keep this logic inside the unit test? Probably a personal preference, but I think it might also make the test a bit easier to read (happy to change my mind if you disagree)

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
```diff
@@ -341,17 +322,16 @@ public SearchOptions Create(string compartmentType, string compartmentId, string
...
                     }
                 }
 
+                // Check if sort parameters are valid and can be used for sorting.
                 if (sortingsValid)
                 {
                     if (!_sortingValidator.ValidateSorting(sortings, out IReadOnlyList<string> errorMessages))
                     {
+                        // Sanity check, ValidateSorting must output errors if it returns false.
```
rotodd, line 334: +1 thanks for adding comments to this method - I found it makes it easier to digest

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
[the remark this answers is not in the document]
Tom-Markoch: This code specifically checks if sort parameter is searchable. Logic is to first check if param is searchable, if not then it's not sortable either. This is the preexisting code. The functions that do the test throw Search exception so message is correct. Also, these functions have 90 references so I can't just change the behavior. Here's what is called to do the test, only search test is performed at this point, not sort.
https://github.com/microsoft/fhir-server/blob/287c33128e61bd733798d66af9f11f9914ac7f8c/src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterDefinitionManager.cs#L106-L112
https://github.com/microsoft/fhir-server/blob/287c33128e61bd733798d66af9f11f9914ac7f8c/src/Microsoft.Health.Fhir.Core/Features/Definition/SearchableSearchParameterDefinitionManager.cs#L69-L79

src/Microsoft.Health.Fhir.Core/Features/Persistence/BadRequestException.cs
```diff
@@ -18,5 +20,22 @@ public BadRequestException(string errorMessage)
...
                     errorMessage));
         }
+
+        /// <summary>
+        /// Constructor creates BadRequestException from collection of error strings.
+        /// </summary>
+        /// <param name="errorMessages">Collection of error strings that are used to initialize OperationOutcomeIssue collection.
+        /// Must be non-null and non-empty</param>
+        public BadRequestException(ICollection<string> errorMessages)
+            : base((errorMessages.Count < 2) ? errorMessages.First() : "Multiple bad request errors.")
```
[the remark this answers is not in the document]
Tom-Markoch: I'm afraid I don't understand. Message itself is not for internal use, message will be returned to the client. I meant Exception itself is used internally so there's no need for a call param validation (at least not in the RELEASE code) since we control the caller code.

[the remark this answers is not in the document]
Tom-Markoch: If you really want, I can remove. But I really like it this way, makes it perfectly clear without worrying about operator precedence.

src/Microsoft.Health.Fhir.Core/Features/Persistence/BadRequestException.cs
```diff
@@ -18,5 +20,22 @@ public BadRequestException(string errorMessage)
                     OperationOutcomeConstants.IssueType.Invalid,
                     errorMessage));
         }
+
+        /// <summary>
+        /// Constructor creates BadRequestException from collection of error strings.
+        /// </summary>
+        /// <param name="errorMessages">Collection of error strings that are used to initialize OperationOutcomeIssue collection.
+        /// Must be non-null and non-empty</param>
```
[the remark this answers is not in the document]
Tom-Markoch:
```C#
            : base(((Func<string>)(() =>
            {
                Debug.Assert(errorMessages != null, "Parameter errorMessages must not be null.");
                Debug.Assert(errorMessages == null || errorMessages.Any(), "Parameter errorMessages must not be empty.");
                return (errorMessages == null || errorMessages.Count < 2) ? errorMessages.First() : "Multiple bad request errors.";
            }))())
            /*: base((errorMessages == null || errorMessages.Count < 2) ? errorMessages.First() : "Multiple bad request errors.")*/
```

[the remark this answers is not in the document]
Tom-Markoch: I almost had this done but then something crossed my mind, Debug.Assert will only show in VS debug or output window. In the code I was debugging yesterday I saw "_output" used and that wrote into the pipelines logs, and I know that for ASP.NET projects using iLogger is proper way. So do we know the right way to output debug messages for on FHIR-server project?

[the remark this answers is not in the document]
Tom-Markoch: Also, just a note...
```C#
: base((errorMessages == null || errorMessages.Count < 2) ? errorMessages.First() : "Multiple bad request errors.")
```
this code does handle null and empty. In both cases First() will be executed and First() will check errorMessages and throw appropriate Exception for both cases (it will check for null as well because First() is static extension).

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/SearchTestsBase.cs
Tom-Markoch: I'm just following convention. All unit tests in this file use "ExecuteAndValidate" functions. I'm probably first one in this file to validate proper error OperationOutcome messages. I would rather add few more unit tests and call this function than put this inside the unit test. I can see that same function could be used for search test failures, too many sort parameters, mix of search and sort failures. I can do that instead.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
```diff
@@ -341,17 +322,16 @@ public SearchOptions Create(string compartmentType, string compartmentId, string
...
                         sortingsValid = false;
-                        _contextAccessor.RequestContext?.BundleIssues.Add(new OperationOutcomeIssue(
-                            OperationOutcomeConstants.IssueSeverity.Warning,
-                            OperationOutcomeConstants.IssueType.NotSupported,
-                            string.Format(CultureInfo.InvariantCulture, Core.Resources.SearchParameterNotSupported, sorting.Item1, string.Join(", ", resourceTypesString))));
+                        searchSortErrors.Add(string.Format(CultureInfo.InvariantCulture, Core.Resources.SearchParameterNotSupported, sorting.Item1, string.Join(", ", resourceTypesString)));
                     }
                 }
 
+                // Check if sort parameters are valid and can be used for sorting.
```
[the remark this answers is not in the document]
Tom-Markoch: Done

src/Microsoft.Health.Fhir.Core/Features/Persistence/BadRequestException.cs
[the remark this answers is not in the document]
rotodd: > If you really want, I can remove the parentheses. But I really like it this way, makes it perfectly clear without worrying about operator precedence. It's just code style.

Totally fair! Happy to keep the extra parentheses if it improves readability.

src/Microsoft.Health.Fhir.Core/Features/Persistence/BadRequestException.cs
rotodd: > I'm afraid I don't understand. Message itself is not for internal use, message will be returned to the client. I meant Exception itself is used internally so there's no need for a call param validation (at least not in the RELEASE code) since we control the caller code.

Sorry for the lack of clarity! From my understanding, the message, "Multiple bad request errors.", is only for internal use. If that's correct - could we add a comment for it here?

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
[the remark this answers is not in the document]
Tom-Markoch: Adding new error string that would clarify that for parameter to be a valid sort param, it must first be valid search param involves adding new string into src\Microsoft.Health.Fhir.Core\Resources.resx and src\Microsoft.Health.Fhir.Core\Resources.Designer.cs . Need clarity on this, are there also localized versions of Resources.resx in other languages that also need to be modified? This is preexisting behavior that is really not "handling strict" issue. Should be opened as a new work item, if we want to do it at all. (or we can just change message from search exception into sort exception message???)

src/Microsoft.Health.Fhir.Core/Features/Persistence/BadRequestException.cs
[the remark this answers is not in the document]
Tom-Markoch: I made expression bit more complex so decide to keep parentheses for clarity.
Decided not to add Debug.Assert, because this would only show in VS debug window. Proper way to log ASP.NET servers is by using ILogger interface https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/ because this can be setup to show messages both when debugging in VS and when deployed on Azure. But I don't see that in the files and should really be a separate item.
Also, any messaging at this point would be redundant because proper exception is thrown by First() for either null or empty.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/SearchTestsBase.cs
Tom-Markoch: Added unit tests for case when sort is not valid search param. Now method is reused.

src/Microsoft.Health.Fhir.Core/Features/Persistence/BadRequestException.cs
```diff
@@ -27,7 +27,7 @@ public BadRequestException(string errorMessage)
         /// <param name="errorMessages">Collection of error strings that are used to initialize OperationOutcomeIssue collection.
         /// Must be non-null and non-empty</param>
         public BadRequestException(ICollection<string> errorMessages)
-            : base((errorMessages.Count < 2) ? errorMessages.First() : "Multiple bad request errors.")
+            : base((errorMessages == null || errorMessages.Count < 2) ? errorMessages.First() : "Multiple bad request errors.")
```
LTA-Thinking, line 30: If there is going to be a null check here I think it needs to be handled better. Right now if it is null it will try to get the First element and get a null pointer exception. Without the null check in my mind it was implied that null wasn't an expected value, but with the null check I think we need to handle the possibility of it being null.

Tom-Markoch: First() is extension, it can and will check for null and throw ArgumentNullException. https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.first

Tom-Markoch: I added null check on purpose, even if constructor is by mistake called with null, it will not crash with null ptr but will instead execute First() extension which will handle null ptr properly.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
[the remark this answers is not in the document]
rotodd: Thanks for talking through this with me! As discussed, we can update the error message returned to indicate that the `_sort` parameter value is not a valid search parameter

src/Microsoft.Health.Fhir.Core/Features/Persistence/BadRequestException.cs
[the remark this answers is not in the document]
rotodd: > I almost had this done but then something crossed my mind, Debug.Assert will only show in VS debug or output window. In the code I was debugging yesterday I saw "_output" used in the unit tests and that wrote into the pipelines logs, and I know that for ASP.NET projects using iLogger is proper way. So, do we know the right way to output debug messages for on FHIR-server project?

This is a good question - we use `iLogger` in other parts of the solution, but I'm not sure why we don't use it in the exception classes that inherit from `FhirException`. As discussed, we can post in the fhir server channel to get a bit more info on why `Debug.Assert()` or `EnsureArg.IsNotNull()` is used. Once we decide on a good approach, it seems like we might need a user story to make our null checking across exceptions more uniform.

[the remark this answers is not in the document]
brendankowitz: > Also, any messaging at this point would be redundant because proper exception is thrown by First() for either null or empty.

The issue is that throwing an Exception creating an Exception hides the actual cause. This is why the Assert was added as a softer notification in the Exception class constructors. In other places EnsureArgs should be used as it clearly defines which parameter was null.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
```diff
@@ -322,11 +322,11 @@ public SearchOptions Create(string compartmentType, string compartmentId, string
                     catch (SearchParameterNotSupportedException)
                     {
                         sortingsValid = false;
-                        searchSortErrors.Add(string.Format(CultureInfo.InvariantCulture, Core.Resources.SearchParameterNotSupported, sorting.Item1, string.Join(", ", resourceTypesString)));
+                        searchSortErrors.Add(string.Format(CultureInfo.InvariantCulture, Core.Resources.SortParameterValueIsNotValidSearchParameter, sorting.Item1, string.Join(", ", resourceTypesString)));
                     }
                 }
 
-                // Sort parameters are valid search parameters. Now verify that sort parameters are also valid for sorting.
+                // Sort parameter values are valid search parameters. Now verify that sort paremeter values are also valid for sorting.
```
rotodd, line 329: Mini typo: "paremeter" --> "parameter"

src/Microsoft.Health.Fhir.Core/Features/Persistence/BadRequestException.cs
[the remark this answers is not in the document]
Tom-Markoch: Added Debug.Assert

src/Microsoft.Health.Fhir.Core/Features/Persistence/BadRequestException.cs
```diff
@@ -27,7 +29,12 @@ public BadRequestException(string errorMessage)
         /// <param name="errorMessages">Collection of error strings that are used to initialize OperationOutcomeIssue collection.
         /// Must be non-null and non-empty</param>
         public BadRequestException(ICollection<string> errorMessages)
-            : base((errorMessages == null || errorMessages.Count < 2) ? errorMessages.First() : "Multiple bad request errors.")
+            : base(((Func<string>)(() =>
+            {
+                Debug.Assert(errorMessages != null, "Parameter errorMessages must not be null.");
+                Debug.Assert(errorMessages == null || errorMessages.Any(), "Parameter errorMessages must not be empty.");
+                return (errorMessages == null || errorMessages.Count < 2) ? errorMessages.First() : "Multiple bad request errors."; // "Multiple bad request errors." message is used only internally, no need to add it to the Resources.resx file.
```
rotodd, line 36: Is there a reason why we have the check for `errorMessages == null` on lines 35 and 36? It looks like we cover this case on line 34.

Tom-Markoch: It's not the same, one is != other is ==. According to documentation Debug.Assert will put out debug message and then continue with execution. If errorMessages is null, then first DebugAssert will put the message "must not be null", and then execution will continue to the second Debug.Assert. If I didn't put that second errorMessages == null, then it would try Any() on null reference and throw exception.

So it's all because Debug.Assert continues to execute after the test, depending on how it's setup. https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.debug.assert

rotodd: Ah, it is my understanding that `Debug.Assert()` will halt execution and return the call stack.

I am reading this as:
Line 34 - if `errorMessages` is null, halt execution. Otherwise proceed to line 35.

Tom-Markoch: Doc says if app has user interface it will display message box with stack and options ignore, retry, abort. Clicking ignore will continue execution.

If app has no user interface it only says it will send the message to the listeners. Remarks: "If the result is false, it sends the specified diagnostic message to the Listeners collection." That's all it says.

It's here https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.debug.assert under "Remarks" for any Assert method

Tom-Markoch: System.Diagnostics.Debug.Assert(false, "helloA");
            System.Diagnostics.Debug.Assert(false, "helloB");

Doc says A will only output message.
Console test app executes B.
FHIR-server when driven by unit tests throws at A and does not reach B
FHIR-server when started standalone and invoked through rest executes both A and B

So, second test stays.

rotodd (approved):

src/Microsoft.Health.Fhir.Core/Features/Persistence/BadRequestException.cs
rotodd: Thanks - I didn't know `Debug.Assert()` will only halt execution when running the tests and not when running the server standalone. I created a user story to investigate how we might be able to make this consistent [AB#86744](https://microsofthealth.visualstudio.com/Health/_workitems/edit/86744).

Tom-Markoch:
When sort parameter value is invalid, response contains
OperationOutcomeIssues with appropriate warning/error messages,
depending on the state of the "Prefer handling" header. Also,
sort/search/errors/warnings are handled same way.

Tom-Markoch:
When sort parameter value is invalid, response contains
OperationOutcomeIssues with appropriate warning/error messages,
depending on the state of the "Prefer handling" header. Also,
sort/search/errors/warnings are handled same way.

rotodd (approved):

Tom-Markoch:
When sort parameter value is invalid, response contains
OperationOutcomeIssues with appropriate warning/error messages,
depending on the state of the "Prefer handling" header. Also,
sort/search/errors/warnings are handled same way.

Refs #2208

PR #2348, opened 2021-11-10
Adds parameters on a remove partition sproc

bbatjargal:
## Description
- Adds parameters on a RemovePartitionFromResourceChanges stored procedure.
- Adds a new schema version 21.

## Related issues
Addresses [issue #].

## Testing

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/21.diff.sql
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/21.diff.sql
@@ -1,5 +1,5 @@
 -- STORED PROCEDURE
---     RemovePartitionFromResourceChanges
+--     RemovePartitionFromResourceChanges_2
 --
 -- DESCRIPTION
 --     Switches a partition from a resource change data table to a staging table.
@@ -11,7 +11,7 @@
 --     @partitionBoundaryToMerge
 --         * The partition boundary value to merge.
 --
-CREATE OR ALTER PROCEDURE dbo.RemovePartitionFromResourceChanges
+CREATE OR ALTER PROCEDURE dbo.RemovePartitionFromResourceChanges_2
     @partitionNumberToSwitchOut int,
     @partitionBoundaryToMerge datetime2(7)
 AS
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/21.diff.sql
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/21.diff.sql
@@ -32,7 +32,7 @@ AS
         /* Switches a partition to the staging table. */
         ALTER TABLE dbo.ResourceChangeData SWITCH PARTITION @partitionNumberToSwitchOut TO dbo.ResourceChangeDataStaging;
         
-        /* Merges range to move lower boundary one partition ahead. */
+        /* Merges range to move boundary one partition ahead. */
         ALTER PARTITION FUNCTION PartitionFunction_ResourceChangeData_Timestamp() MERGE RANGE(@partitionBoundaryToMerge);
         
         /* Cleans up the staging table to purge resource changes. */
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/21.diff.sql
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/21.diff.sql
@@ -20,24 +20,16 @@ CREATE OR ALTER PROCEDURE dbo.RemovePartitionFromResourceChanges_2
     @partitionBoundaryToMerge datetime2(7)
 AS
   BEGIN
-    
-    /* using XACT_ABORT to force a rollback on any error. */
-    SET XACT_ABORT ON;
-    
-    BEGIN TRANSACTION
-
-        /* Cleans up a staging table if there are existing rows. */
-        TRUNCATE TABLE dbo.ResourceChangeDataStaging;
+    /* Cleans up a staging table if there are existing rows. */
+    TRUNCATE TABLE dbo.ResourceChangeDataStaging;
         
-        /* Switches a partition to the staging table. */
-        ALTER TABLE dbo.ResourceChangeData SWITCH PARTITION @partitionNumberToSwitchOut TO dbo.ResourceChangeDataStaging;
+    /* Switches a partition to the staging table. */
+    ALTER TABLE dbo.ResourceChangeData SWITCH PARTITION @partitionNumberToSwitchOut TO dbo.ResourceChangeDataStaging;
         
-        /* Merges range to move boundary one partition ahead. */
-        ALTER PARTITION FUNCTION PartitionFunction_ResourceChangeData_Timestamp() MERGE RANGE(@partitionBoundaryToMerge);
+    /* Merges range to move boundary one partition ahead. */
+    ALTER PARTITION FUNCTION PartitionFunction_ResourceChangeData_Timestamp() MERGE RANGE(@partitionBoundaryToMerge);
         
-        /* Cleans up the staging table to purge resource changes. */
-        TRUNCATE TABLE dbo.ResourceChangeDataStaging;
-
-    COMMIT TRANSACTION;
+    /* Cleans up the staging table to purge resource changes. */
+    TRUNCATE TABLE dbo.ResourceChangeDataStaging;
 END;
 GO 
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/21.diff.sql
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/21.diff.sql
@@ -1,0 +1,39 @@
+-- STORED PROCEDURE
+--     RemovePartitionFromResourceChanges
+--
+-- DESCRIPTION
+--     Switches a partition from a resource change data table to a staging table.
+--     Then merges two partitions based on a given partition boundary value.
+--     After that, truncates the staging table to purge the old resource change data.
+--
+--     @partitionNumberToSwtichOut
+--         * The partition number to switch out.
+--     @partitionBoundaryToMerge
+--         * The partition boundary value to merge.
+--
+CREATE OR ALTER PROCEDURE dbo.RemovePartitionFromResourceChanges
+    @partitionNumberToSwitchOut int,
+    @partitionBoundaryToMerge datetime2(7)
+AS
+  BEGIN
+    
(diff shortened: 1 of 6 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/21.sql
```diff
@@ -0,0 +1,4104 @@
...
+--     ConfigurePartitionOnResourceChanges
+--
+-- DESCRIPTION
+--     Creates initial partitions for future datetimes on the resource change data table if they do not already exist.
+--
+-- PARAMETERS
+--     @numberOfFuturePartitionsToAdd
+--         * The number of partitions to add for future datetimes.
+--
+CREATE OR ALTER PROCEDURE dbo.ConfigurePartitionOnResourceChanges
```
rbans96, line 3580: Since our codebase is backward compatible, the approach that we take to update the existing stored proc from previous version, is by adding incrementing version to it(for ex- [upsertResource ](https://github.com/microsoft/fhir-server/blob/main/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/9.diff.sql#L49)Stored proc has been updated in few migrations) and then add logic in the code to use the relevant store proc based on the schema version.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/21.diff.sql
```diff
@@ -0,0 +1,39 @@
+-- STORED PROCEDURE
+--     RemovePartitionFromResourceChanges
```
rbans96, line 2: Since our codebase is backward compatible, the approach that we take to update the existing stored proc from previous version, is by adding incrementing version to it(for ex- upsertResource Stored proc has been updated in few migrations) and then add logic in the code to use the relevant store proc based on the schema version.

bbatjargal: Renamed the stored procedure to RemovePartitionFromResourceChanges_2.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/21.sql
```diff
@@ -0,0 +1,4104 @@
...
+--     Switches a partition from a resource change data table to a staging table.
+--     Then merges two partitions based on a given partition boundary value.
+--     After that, truncates the staging table to purge the old resource change data.
+--
+--     @partitionNumberToSwtichOut
+--         * The partition number to switch out.
+--     @partitionBoundaryToMerge
+--         * The partition boundary value to merge.
+--
+CREATE PROCEDURE dbo.RemovePartitionFromResourceChanges_2
```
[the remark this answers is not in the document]
bbatjargal: Since it is the main schema file, that is not required.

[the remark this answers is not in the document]
bbatjargal: Added alter keyword.

wi-y (approved):

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/SchemaVersionConstants.cs
```diff
@@ -18,6 +18,6 @@ public static class SchemaVersionConstants
         public const int SupportsResourceChangeCaptureSchemaVersion = (int)SchemaVersion.V14;
         public const int BulkReindexReturnsFailuresVersion = (int)SchemaVersion.V16;
         public const int AddMinMaxForDateAndStringSearchParamVersion = (int)SchemaVersion.V18;
-        public const int SupportsPartitionedResourceChangeDataVersion = (int)SchemaVersion.V20;
+        public const int SupportsPartitionedResourceChangeDataVersion = (int)SchemaVersion.V21;
```
rbans96, line 21: I am wondering if this change would break the existing check in [SqlServerFhirResourceChangeDataStore](https://github.com/microsoft/fhir-server/blob/0d6e90c7d17dd1273147eb84fcfd24ccc5857e7b/src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs#L165)?

bbatjargal: Thank you for pointing it out. Added new schema version constant.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/SchemaVersionConstants.cs
```diff
@@ -18,6 +18,7 @@ public static class SchemaVersionConstants
         public const int SupportsResourceChangeCaptureSchemaVersion = (int)SchemaVersion.V14;
         public const int BulkReindexReturnsFailuresVersion = (int)SchemaVersion.V16;
         public const int AddMinMaxForDateAndStringSearchParamVersion = (int)SchemaVersion.V18;
-        public const int SupportsPartitionedResourceChangeDataVersion = (int)SchemaVersion.V21;
+        public const int SupportsPartitionedResourceChangeDataVersion = (int)SchemaVersion.V20;
+        public const int ParameterizedRemovePartitionFromResourceChangesVersion = (int)SchemaVersion.V21;
```
rbans96, line 22: Am I missing something? I don't see this new constant being used anywhere

bbatjargal: Yes, it is not used on the OSS side. It is going to be used on Health PaaS.

rbans96: I am wondering if we can define this constant in health-paas instead as someone might still get confused that its not being used any where?

bbatjargal: I think we should keep constants in one place since it is related to the DB schema version. In the future, we might release Event Agent for the OSS.

wi-y: I would prefer to keep the constant here for future event support in OSS, just to keep everything in one place, even though it would not be used in OSS at this time.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/21.diff.sql
```diff
@@ -0,0 +1,39 @@
...
+    BEGIN TRANSACTION
+
+        /* Cleans up a staging table if there are existing rows. */
+        TRUNCATE TABLE dbo.ResourceChangeDataStaging;
+        
+        /* Switches a partition to the staging table. */
+        ALTER TABLE dbo.ResourceChangeData SWITCH PARTITION @partitionNumberToSwitchOut TO dbo.ResourceChangeDataStaging;
+        
+        /* Merges range to move lower boundary one partition ahead. */
+        ALTER PARTITION FUNCTION PartitionFunction_ResourceChangeData_Timestamp() MERGE RANGE(@partitionBoundaryToMerge);
```
rotodd, line 32: Just for my understanding, is there a use case where `@partitionBoundaryToMerge`  will not equal the lowest boundary value, or are we passing in `@partitionBoundaryToMerge` because it makes more sense to determine what it is elsewhere?

bbatjargal: In our case, no, it should be always the lowest boundary value. The reason for adding parameters is to make the contract explicit to ensure only what passed gets deleted.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/SchemaVersionConstants.cs
bbatjargal: Adding a comment looks odd and might be confusing for OSS-only users. If possible, I would like to leave it as it is since it makes sense to each schema version has a corresponding descriptive constant.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/21.sql
```diff
@@ -0,0 +1,4104 @@
...
+--     Switches a partition from a resource change data table to a staging table.
+--     Then merges two partitions based on a given partition boundary value.
+--     After that, truncates the staging table to purge the old resource change data.
+--
+--     @partitionNumberToSwtichOut
+--         * The partition number to switch out.
+--     @partitionBoundaryToMerge
+--         * The partition boundary value to merge.
+--
+CREATE OR ALTER PROCEDURE dbo.RemovePartitionFromResourceChanges_2
```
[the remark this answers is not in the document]
bbatjargal: Removed ALTER keyword.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/21.diff.sql
rotodd: Okay thanks, that's good to know - does it make sense to add some sort of validation and/or documentation (comments) for this?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/SchemaVersionConstants.cs
bbatjargal: Adding a comment looks odd and might be confusing for the OSS-only users. I think it is safe to leave it as it is since it makes sense to each schema version has a corresponding descriptive constant.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/21.diff.sql
bbatjargal: Since the stored procedure is parameterized, the procedure can be used for the general purpose. It is safe. For the health PaaS side, we only pass the lowest boundary value.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/SchemaVersionConstants.cs
bbatjargal: Added a comment.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/21.diff.sql
bbatjargal: Added comments on the stored procedure.

rbans96 (approved):

PR #2353, opened 2021-11-11
Add primary keys to the fhir tables

rbans96:
## Description
This PR adds the primary keys to the existing fhir tables.
- dbo.SearchParam
- dbo.ResourceType
- dbo.System
- dbo.QuantityCode
- dbo.ClaimType
- dbo.CompartmentType
- dbo.CompartmentAssignment
- dbo.Resource
- dbo.ExportJob
- dbo.ReindexJob
- dbo.TaskInfo
- dbo.UriSearchParam

There are few approaches followed based on the nature of the columns in the tables.
Scenario 1: Table with IDENTITY column
Approach - 
1. Add non clustered PK on identity column

Scenario 2: Table with unique clustered index
Approach -
1. Create temporary non clustered index for the clustered index columns
2. Drop clustered index
3. Create PKC
4. Drop temporary non clustered index
e.g. dbo.Resource 

Scenario 3: Tables with  non unique clustered index
Approach - 
1. Identify set of non-nullable columns that makes it unique
2. Create non clustered PK on identified unique combination
3. Drop existing non-clustered index on the identified unique combination(if exits)

Scenario 4: TABLE with no unique set of columns
Not included in this PR

## Related issues
Addresses [#85866](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Olympus/Stories/?workitem=85866)

## Testing
Manually tested by running the migration script on 400 GB fragmented database.

## FHIR Team Checklist


src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Tables/StringSearchParam.sql
```diff
@@ -4,6 +4,9 @@
     ResourceSurrogateId         bigint              NOT NULL,
     SearchParamId smallint NOT NULL,
     Text nvarchar(256) COLLATE Latin1_General_100_CI_AI_SC NOT NULL,
+    CONSTRAINT PK_StringSearchParam PRIMARY KEY NONCLUSTERED (ResourceTypeId, SearchParamId, Text, ResourceSurrogateId)
```
cunninghamjc, line 7: Is this guaranteed to be unique?  What about TextOverflow?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/23.diff.sql
```diff
@@ -0,0 +1,559 @@
...
+-- We will create non-clustered index(IX) first, drop IXC, create PKC and finally drop IX
+-- so that the table would never be left without indexes.
+-- Creating temporary non-clustered index
+EXEC dbo.LogSchemaMigrationProgress 'Creating IX_ReindexJob'
+IF NOT EXISTS (
+    SELECT * 
+	FROM sys.indexes 
+	WHERE name='IX_ReindexJob' AND object_id = OBJECT_ID('dbo.ReindexJob'))
+BEGIN
+	CREATE NONCLUSTERED INDEX IX_ReindexJob ON dbo.ReindexJob
```
cunninghamjc, line 244: Make unique?

cunninghamjc:
> CREATE NONCLUSTERED INDEX IX_Resource_ResourceSurrogateId ON dbo.Resource

Should this be unique?

---
Refers to: src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Tables/Resource.sql:48 in a70a998. [](commit_id = a70a99802cc326ddc50a1e2fe79e6831e3a60361, deletion_comment = False)

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Tables/ResourceType.sql
```diff
@@ -1,6 +1,7 @@
 ﻿CREATE TABLE dbo.ResourceType
 (
     ResourceTypeId          smallint IDENTITY(1,1)          NOT NULL,
+    CONSTRAINT PK_ResourceType PRIMARY KEY NONCLUSTERED (ResourceTypeId),
     Name                    nvarchar(50)                    COLLATE Latin1_General_100_CS_AS  NOT NULL
```
cunninghamjc, line 5: This means two copies of the table.  Would it be better to just make the PK on Name?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Tables/ReferenceSearchParam.sql
```diff
@@ -6,6 +6,9 @@
     BaseUri                             varchar(128)            COLLATE Latin1_General_100_CS_AS NULL,
     ReferenceResourceTypeId             smallint                NULL,
     ReferenceResourceId                 varchar(64)             COLLATE Latin1_General_100_CS_AS NOT NULL,
+    CONSTRAINT PK_ReferenceSearchParam PRIMARY KEY NONCLUSTERED (ResourceTypeId, ResourceSurrogateId, SearchParamId, ReferenceResourceId)
```
cunninghamjc, line 9: Is this guaranteed to be unique?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Tables/ReferenceTokenCompositeSearchParam.sql
```diff
@@ -9,6 +9,9 @@
     ReferenceResourceVersion1 int NULL,
     SystemId2 int NULL,
     Code2 varchar(128) COLLATE Latin1_General_100_CS_AS NOT NULL,
+    CONSTRAINT PK_ReferenceTokenCompositeSearchParam PRIMARY KEY NONCLUSTERED (ResourceTypeId, SearchParamId, ReferenceResourceId1, Code2, ResourceSurrogateId)
```
cunninghamjc, line 12: This overlaps with IX_ReferenceTokenCompositeSearchParam_ReferenceResourceId1_Code2.  Should the non-clustered index be unique too?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/23.diff.sql
```diff
@@ -0,0 +1,559 @@
...
+
+EXEC dbo.LogSchemaMigrationProgress 'Beginning schema migration to version 23.';
+GO
+
+-- SearchParam table
+-- Adding nonclustered primary key on identity column
+EXEC dbo.LogSchemaMigrationProgress 'Adding PK_SearchParam'
+IF NOT EXISTS (
+    SELECT * 
+	FROM sys.key_constraints
```
cunninghamjc, line 13: Nit: this file has both spaces and tabs -- pick one?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/23.diff.sql
```diff
@@ -0,0 +1,559 @@
...
+END
+
+-- Deleting temporary non-clustered index created.
+EXEC dbo.LogSchemaMigrationProgress 'Dropping IX_Resource'
+IF EXISTS (
+    SELECT * 
+	FROM sys.indexes 
+	WHERE name='IX_Resource' AND object_id = OBJECT_ID('dbo.Resource'))
+BEGIN
+	DROP INDEX IX_Resource ON dbo.Resource
```
cunninghamjc, line 177: Do we need "ONLINE=ON" here?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/23.diff.sql
```diff
@@ -0,0 +1,559 @@
...
+-- We will create non-clustered index first, drop IXC, create PKC and finally drop IX
+-- so that the table would never be left without indexes.
+-- Creating temporary non-clustered index
+EXEC dbo.LogSchemaMigrationProgress 'Creating IX_Resource'
+IF NOT EXISTS (
+    SELECT * 
+	FROM sys.indexes 
+	WHERE name='IX_Resource' AND object_id = OBJECT_ID('dbo.Resource'))
+BEGIN
+	CREATE NONCLUSTERED INDEX IX_Resource ON dbo.Resource
```
cunninghamjc, line 137: Should this be unique since the clustered index is unique?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Tables/ResourceType.sql
rbans96: That makes sense! I also talked to Katherine on this and she too agreed to have primary key on the unique column rather than identity column since we don't use identity column in any of our query. She had a good suggestion to add UNIQUE constraint for the identity column though so incorporate that change as well.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/23.diff.sql
rbans96: Only a clustered index can be dropped online.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Tables/StringSearchParam.sql
rbans96: This is a good point. Somehow I missed to test the case where Text would only differ in last few bytes and thus multiple rows could have same Text but TextOverflow would be different. Since TextOverflow datatype is varchar(max), ideally this should not considered to be part of primary key so more analysis is required to find a primary key for this table. 
Created [user story](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Olympus/Stories/?workitem=87237) to investigate if we can add isTextOverflowDiff as bit which will only be set to 1 if TextOverflow is not null and different for two rows without impacting the current fhir functionality.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Tables/ReferenceSearchParam.sql
rbans96: Thanks Joyce to confirm. I found some gap in my analysis for this table and few other SearchParam tables as back in the time, I relied on the data in 1TB database and couple of resources on fhir spec. But I have now re-analyzed these tables and found some more work is required to be done on those since column that  need to be part of pk are NULLable and will be addressed in the next PR.

I have cleaned up this PR and it adds PK for 13 FHIR tables.

feordin (approved):

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/25.diff.sql
```diff
@@ -0,0 +1,548 @@
...
+)
+DELETE FROM cte WHERE row_num > 1
+GO
+
+-- Create nonclustered primary key on the set of non-nullable columns that makes it unique
+EXEC dbo.LogSchemaMigrationProgress 'Adding PK_UriSearchParam'
+IF NOT EXISTS (
+    SELECT * 
+	FROM sys.key_constraints 
+	WHERE name='PK_UriSearchParam' AND type='PK')
```
tongwu-msft, line 541: There's still clustered index "IXC_UriSearchParam" on this table? should we remove it?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/25.diff.sql
```diff
@@ -0,0 +1,548 @@
...
+BEGIN
+	ALTER TABLE dbo.UriSearchParam
+	ADD CONSTRAINT PK_UriSearchParam PRIMARY KEY NONCLUSTERED(ResourceTypeId, SearchParamId, Uri, ResourceSurrogateId)
```
tongwu-msft, line 544: Could we use clustered here?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/25.diff.sql
rbans96: We still need this clustered index since the primary key has additional column whereas this clustered index does not.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/25.diff.sql
rbans96: Thanks for checking. As the best practices and primary key design discussion, we don't want large primary key to be clustered as it would take more space than the current clustered indexes in place.

PR #2361, opened 2021-11-15
Log OperationCanceled as an informational message

wi-y:
## Description
Currently when a task or operation is canceled we are logging an exception in `SqlServerFhirResourceChangeDataStore.GetRecordsAsync()`. However, in many cases, cancelation is initiated by some known cause such as a code upgrade or a reprovision, so it is not appropriate to log an error. This PR updates the code so that we log an informational message instead of an error when we receive a request for cancelation.

## Related issues
86631

## Testing
Added unit test

## FHIR Team Checklist
- [ x] **Update the title** of the PR to be succinct and less than 50 characters
- [x ] **Add a milestone** to the PR for the sprint that it is merged (i.e. add S47)
- [x ] Tag the PR with the type of update: **Bug**, **Dependencies**, **Enhancement**, or **New-Feature**
- [ x] Tag the PR with Azure API for FHIR if this will release to the Azure API for FHIR managed service (CosmosDB or common code related to service)
- [ x] Tag the PR with Azure Healthcare APIs if this will release to the Azure Healthcare APIs managed service (Sql server or common code related to service)

Patch


src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
```diff
@@ -139,6 +139,11 @@ public async Task<IReadOnlyCollection<ResourceChangeData>> GetRecordsAsync(long
                     }
                 }
             }
+            catch (Exception ex) when ((ex is OperationCanceledException || ex is TaskCanceledException) && cancellationToken.IsCancellationRequested)
+            {
+                _logger.LogInformation(ex.Message);
+                return listResourceChangeData;
```
wi-y, line 145: If other prefer, we could instead opt to throw here without calling _logger.LogError and let the calling method deal with the exception.

In our use case, we are just going to call _logger.LogInformation anyway which is why I chose to call _logger.LogInformation here.

src/Microsoft.Health.Fhir.SqlServer.UnitTests/Features/ChangeFeed/SqlServerFhirResourceChangeDataStoreTests.cs
```diff
@@ -75,5 +76,29 @@ public async Task GivenEmptyConnectionString_WhenGetResourceChanges_ThenInvalidO
...
+                DataSource = "Test",
+                InitialCatalog = "Test",
+            }.ToString();
+
+            var config = Options.Create(new SqlServerDataStoreConfiguration { ConnectionString = mockConnection });
+            var connectionStringProvider = new DefaultSqlConnectionStringProvider(config);
+            var connectionFactory = new DefaultSqlConnectionFactory(connectionStringProvider);
+            var schemaInformation = new SchemaInformation(SchemaVersionConstants.Min, SchemaVersionConstants.Max);
+            schemaInformation.Current = SchemaVersionConstants.Max;
+            var resourceChangeDataStore = new SqlServerFhirResourceChangeDataStore(connectionFactory, NullLogger<SqlServerFhirResourceChangeDataStore>.Instance, schemaInformation);
```
bbatjargal, line 94: I think we need to re-use the constructor of the test class instead of redefining it from line 83 to line 94. If you need an actual test database call, you can add the test case to the integration test. Otherwise, you can adjust the constructor of the class and mock the resourceChangeDataStore.GetRecordsAsync function without making DB call.

src/Microsoft.Health.Fhir.SqlServer.UnitTests/Features/ChangeFeed/SqlServerFhirResourceChangeDataStoreTests.cs
```diff
@@ -75,5 +76,29 @@ public async Task GivenEmptyConnectionString_WhenGetResourceChanges_ThenInvalidO
...
+
+            CancellationTokenSource source = new CancellationTokenSource();
+            var token = source.Token;
+            source.Cancel();
+
+            var records = await resourceChangeDataStore.GetRecordsAsync(1, 1, token);
+            Assert.Equal(0, records.Count);
+        }
```
bbatjargal, line 102: I think we should check it with the actual exception type name like nameof(OperationCanceledException) or nameof(TaskCanceledException).

src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
```diff
@@ -139,6 +139,11 @@ public async Task<IReadOnlyCollection<ResourceChangeData>> GetRecordsAsync(long
                     }
                 }
             }
+            catch (Exception ex) when ((ex is OperationCanceledException || ex is TaskCanceledException) && cancellationToken.IsCancellationRequested)
```
bbatjargal, line 142: Need to update the function exception information on the top at lines 67 and 85.

src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
wi-y: updated to log an information and then throw an exception. added log message to string resx file

src/Microsoft.Health.Fhir.SqlServer.UnitTests/Features/ChangeFeed/SqlServerFhirResourceChangeDataStoreTests.cs
wi-y: updated. had to make some small changes to the ctor since it was previously doing some string.Empty stuff for one of the tests which meant the code would hit an always exception there. but now it is configurable

src/Microsoft.Health.Fhir.SqlServer.UnitTests/Features/ChangeFeed/SqlServerFhirResourceChangeDataStoreTests.cs
wi-y: updated to check for an exception now that we are throwing one

src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
wi-y: updated the help text with the new exceptions being thrown

src/Microsoft.Health.Fhir.SqlServer/Resources.resx
```diff
@@ -158,4 +158,7 @@
   <data name="ExecutionTimeoutExpired" xml:space="preserve">
     <value>The exection timeout expired from SQL Server.</value>
   </data>
+  <data name="GetRecordsAsyncOperationIsCanceled" xml:space="preserve">
+    <value>The operation to get resource changes has been canceled</value>
```
bbatjargal, line 162: nit: The period (.) is missing at the end of the sentence.

src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
```diff
@@ -139,6 +141,11 @@ public async Task<IReadOnlyCollection<ResourceChangeData>> GetRecordsAsync(long
...
+            {
+                _logger.LogInformation(string.Format(Resources.GetRecordsAsyncOperationIsCanceled));
```
bbatjargal, line 146: nit: I think we need to include an ex object in the log information like the following.

_logger.LogInformation(ex, Resources.GetRecordsAsyncOperationIsCanceled);

bbatjargal, line 146: nit: I think we don't need string.Format since there is no object argument.

src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
bbatjargal: nit: Need to add the same help text at line 67. There is an overload of the GetRecordsAsync function.

src/Microsoft.Health.Fhir.SqlServer.UnitTests/Features/ChangeFeed/SqlServerFhirResourceChangeDataStoreTests.cs
```diff
@@ -75,5 +77,24 @@ public async Task GivenEmptyConnectionString_WhenGetResourceChanges_ThenInvalidO
                 Assert.Equal(nameof(InvalidOperationException), ex.GetType().Name);
             }
         }
+
+        [Fact]
+        public async Task GivenOperationCanceled_WhenGetResourceChanges_ThenTaskCanceledExceptionShouldBeThrown()
+        {
+            var source = new CancellationTokenSource();
+            var token = source.Token;
+            source.Cancel();
```
bbatjargal, line 86: For clarification, is it safe to cancel directly here? Do we need to move this cancellation logic to line 91 after calling the GetSqlConnectionAsync function? 

I am not sure. But I'm a little concerning about that it might behave inconsistently (sometimes passes or sometimes fails) during unit test validation at PR or CI build stages. Is it fine?

src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
wi-y: updated. thanks

src/Microsoft.Health.Fhir.SqlServer/Resources.resx
wi-y: added

src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
wi-y: added

wi-y: removed

bbatjargal (approved):

rotodd (approved):

PR #2368, opened 2021-11-20
Adds resource specific versioning configuration

rotodd:
## Description
This PR adds the ability to specify resource-specific versioning in `appsettings.json`. @PTaladay will kindly be taking over this PR as I will be OOF next week.

## Related issues
- Addresses [AB#86649](https://microsofthealth.visualstudio.com/Health/_workitems/edit/86649).

## Up next

- We need to add the ability to deserialize the versioning configuration object so that we can pull this configuration info into PaaS ([AB#86832](https://microsofthealth.visualstudio.com/Health/_workitems/edit/86832))
- Check to see if there are any automated tests that we can add here ([AB#86833](https://microsofthealth.visualstudio.com/Health/_workitems/edit/86833))

## Testing
Added a new `.http` test file that outlines a manual test flow.

## FHIR Team Checklist

Feature (new disable history functionality)

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Conformance/CapabilityStatementBuilder.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Conformance/CapabilityStatementBuilder.cs
@@ -99,7 +99,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Conformance
 
             if (resourceComponent == null)
             {
-                resourceComponent = new ListedResourceComponent(_configuration)
+                resourceComponent = new ListedResourceComponent(_configuration, resourceType)
                 {
                     Type = resourceType,
                     Profile = new ReferenceComponent
--- a/src/Microsoft.Health.Fhir.Core/Features/Conformance/CapabilityStatementBuilder.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Conformance/CapabilityStatementBuilder.cs
@@ -99,7 +99,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Conformance
 
             if (resourceComponent == null)
             {
-                resourceComponent = new ListedResourceComponent(_configuration)
+                resourceComponent = new ListedResourceComponent()
                 {
                     Type = resourceType,
                     Profile = new ReferenceComponent
--- a/src/Microsoft.Health.Fhir.Core/Features/Conformance/CapabilityStatementBuilder.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Conformance/CapabilityStatementBuilder.cs
@@ -99,7 +99,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Conformance
 
             if (resourceComponent == null)
             {
-                resourceComponent = new ListedResourceComponent(_configuration, resourceType)
+                resourceComponent = new ListedResourceComponent(_configuration)
                 {
                     Type = resourceType,
                     Profile = new ReferenceComponent
@@ -107,7 +107,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Conformance
                         Reference = $"http://hl7.org/fhir/StructureDefinition/{resourceType}",
                     },
                 };
-
+                ((DefaultOptionHashSet<string>)resourceComponent.Versioning).DefaultOption = _configuration.Versioning.ResourceTypeOverrides.ContainsKey(resourceType) ? _configuration.Versioning.ResourceTypeOverrides[resourceType] : _configuration.Versioning.Default;
                 listedRestComponent.Resource.Add(resourceComponent);
             }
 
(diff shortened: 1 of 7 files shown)


src/Microsoft.Health.Fhir.Core/Features/Conformance/CapabilityStatementBuilder.cs
```diff
@@ -99,7 +99,7 @@ public ICapabilityStatementBuilder ApplyToResource(string resourceType, Action<L
 
             if (resourceComponent == null)
             {
-                resourceComponent = new ListedResourceComponent(_configuration)
+                resourceComponent = new ListedResourceComponent(_configuration, resourceType)
```
brendankowitz, line 102: I'm wondering if it is now easier to just make `DefaultOptionHashSet.Default` have a setter so this logic can be done below in the builder

PTaladay: I can take a look and see what can be done.

PTaladay: @brendankowitz I made the change. Let me know if the changes correlate to what you were asking.

src/Microsoft.Health.Fhir.Core/Features/Conformance/Models/DefaultOptionHashSet.cs
```diff
@@ -21,7 +21,7 @@ public DefaultOptionHashSet(T defaultOption, IEqualityComparer<T> comparer)
             DefaultOption = defaultOption;
         }
 
-        public T DefaultOption { get; }
+        public T DefaultOption { get; set; }
```
brendankowitz, line 24: I think add this to `IDefaultOption` also, it will make the usage code nicer in this case

PTaladay: From what I can tell that won't make a difference in the usage as you still need to implement the DefaultOption property in the DefaultoptioHashSet. Plus in the interface, DefaultOption is an object, while in the DefaultOptionHashSet it is of type T. Unless I'm missing something, you would still need to cast to "DefaultOptionHashset<string>" since the Versioning in the ListedResourceComponent is of type ICollection<string>.

brendankowitz: Got it, sounds ok for now 👍

src/Microsoft.Health.Fhir.Core/Features/Conformance/Models/ListedResourceComponent.cs
```diff
@@ -21,7 +21,10 @@ public ListedResourceComponent(CoreFeatureConfiguration configuration)
             SearchRevInclude = new HashSet<string>(StringComparer.Ordinal);
             SearchInclude = new HashSet<string>(StringComparer.Ordinal);
             ReferencePolicy = new HashSet<string>(StringComparer.Ordinal);
-            Versioning = new DefaultOptionHashSet<string>(configuration.Versioning.Default, StringComparer.Ordinal);
+
+            var versioningPolicy = configuration.Versioning.Default;
+
+            Versioning = new DefaultOptionHashSet<string>(versioningPolicy, StringComparer.Ordinal);
```
brendankowitz, line 27: With your new changes in the builder, can this now be simplified, Configuration probably doesn't need to be passed in anymore

```suggestion
            Versioning = new DefaultOptionHashSet<string>(ResourceVersionPolicy.Versioned, StringComparer.Ordinal);
```

PTaladay: This has now been simplified.

LTA-Thinking:
If a resource already has historical versions before versioning is turned off then the results from a call to _history show the old versions and the latest version, but have a gap from when versioning was turned off. This makes sense knowing how we store the data, but it seems weird from a user perspective. I don't have a better idea for what should be done, I just wanted to check that this was intended behavior.

LTA-Thinking:
Is it possible to add E2E tests for this feature?

LTA-Thinking (approved):

PTaladay:
> If a resource already has historical versions before versioning is turned off then the results from a call to _history show the old versions and the latest version, but have a gap from when versioning was turned off. This makes sense knowing how we store the data, but it seems weird from a user perspective. I don't have a better idea for what should be done, I just wanted to check that this was intended behavior.

Yes, this was a known issue. The main reason for doing this in the first place is to keep new historical data. In the specific case of Cigna, this would get wiped out eventually so it would only be temporary  for this old history data with a gap to show up. I don't believe there is any specific plan to fix this.

PTaladay:
> 

Yes, I will be taking a look into what is currently covered and see what we need to add.

PR #2376, opened 2021-11-23
Updates to add primary key constraint for QuantitySearchParam and Num…

rbans96:
…berSearchParam

## Description
This PR updates the column LowValue and HighValue as NOT NULL and populate SingleValue in both the columns if not range for quantity and number search parameters.
And thus, we are adding primary key constraint for QuantitySearchParam and NumberSearchParam tables with the combination of NOT NULL columns that uniquely identifies row in the table.

## Related issues
Addresses [#86362].](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Olympus/Stories/?workitem=86362)

## Testing
Existing tests continue to pass.
Tested manually by upgrading the schema and then 
- POSTing Observation resource with value-quantity and valueRange to validate the changes in dbo.QuantitySearchParamt table
- POSTing RiskAssessment resource with probabilityDecimal and probabililtyRange to validate the changes in dbo.NumberSearchParam table

## FHIR Team Checklist


LTA-Thinking:
Are any test changes needed for this update?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/23.diff.sql
```diff
@@ -0,0 +1,203 @@
...
+	WITH (DATA_COMPRESSION = PAGE, ONLINE=ON) 
+	ON PartitionScheme_ResourceTypeId(ResourceTypeId)
+END;
+GO
+
+/*************************************************************
+ Insert singleValue values into the low and high values for
+ dbo.NumberSearchParam table
+**************************************************************/
+EXEC dbo.LogSchemaMigrationProgress 'Populating LowValue and HighValue in NumberSearchParam if null'
```
feordin, line 115: Just a question for my curiosity, why do we set the values in the table after setting the column the NOT NULL.  Is the result when we set the column not null a default value is populated, which is then used in the index.  After that we set the value and the index is rebuilt?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/BulkNumberSearchParameterV1RowGenerator.cs
```diff
@@ -17,14 +17,14 @@ public BulkNumberSearchParameterV1RowGenerator(SqlServerFhirModel model, SearchP
 
         internal override bool TryGenerateRow(int offset, short searchParamId, NumberSearchValue searchValue, out BulkNumberSearchParamTableTypeV1Row row)
         {
-            bool isSingleValue = searchValue.Low == searchValue.High;
+            var singleValue = searchValue.Low == searchValue.High ? searchValue.Low : null;
```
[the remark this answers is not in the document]
rbans96: Good question! Earlier this logic used to populate singleValue field if quantity-value is present on the resource and range(lowValue and highValue fields) when quantity-range is present on the resource. Since main reason of the PR to make lowValue and highValue field as NOT NULL(to include these in primary key constraint) so this is making a small change where we are populating the fields(lowValue and highValue) even if quantity-value is present.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/23.diff.sql
rbans96: Sure!
So the script is following the below sequence for NumberSearchParam or QuantitySearchParam
1. Backfill lowValue and highValue if its null
2. Drop non-clustered indices which has lowValue and highValue in it since we can't alter columns if they exist
3. Alter columns lowValue and highValue as NOT NULL
4. Create dropped indices and primary keys.

rbans96:
> Are any test changes needed for this update?

Since the functionality is intact, and its a matter of the change in database columns (lowValue and highValue) to make it non-nullable, existing tests should still hold true.

Ivanidzo4ka:
>     INCLUDE(SystemId) WHERE IsHistory = 0

I don't understand these indicies.

Why ResourceSurrogateId is part of index, and not part of include, and why systemId is part of Include and not part of index?



---
Refers to: src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/23.sql:388 in ec0d2fd. [](commit_id = ec0d2fdf7d6f6d18f3e885af08fd24568734a373, deletion_comment = False)

rbans96:
> > ```
> > INCLUDE(SystemId) WHERE IsHistory = 0
> > ```
> 
> I don't understand these indicies.
> 
> Why ResourceSurrogateId is part of index, and not part of include, and why systemId is part of Include and not part of index?
> 
> Refers to: src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/23.sql:388 in [ec0d2fd](https://github.com/microsoft/fhir-server/commit/ec0d2fdf7d6f6d18f3e885af08fd24568734a373). [](commit_id = [ec0d2fd](https://github.com/microsoft/fhir-server/commit/ec0d2fdf7d6f6d18f3e885af08fd24568734a373), deletion_comment = False)

This is the existing design of fhir tables. ResourceSurrogateId is the timestamp of when the record is created and is part of almost all the indices to narrow down the search or join related searchParam tables with Resource table. While SystemId is included in index only for non-historical records.

feordin (approved):

Ivanidzo4ka:
It's a good PR, and it does what it say it does in a right way.
But I don't understand why we doing this. Is this yield better results? Is this in preparation for more changes?

To be honest, I don't know why schema looks like it looks right now, but I have hopes what whoever design it had benevolent intent, and put some effort (they even add word SPARSE for low,high values, so looks like person did that deliberately).

It's quite scary to accept schema changes, especially with GA around corner, so can some justification for changes can be produced?

rbans96:
> It's a good PR, and it does what it say it does in a right way. But I don't understand why we doing this. Is this yield better results? Is this in preparation for more changes?
> 
> To be honest, I don't know why schema looks like it looks right now, but I have hopes what whoever design it had benevolent intent, and put some effort (they even add word SPARSE for low,high values, so looks like person did that deliberately).
> 
> It's quite scary to accept schema changes, especially with GA around corner, so can some justification for changes can be produced?

Thanks Ivan for your thoughts and questions. 
This is in preparation to be able to add primary key for all fhir tables and be able to uniquely identify rows. Adding primary keys  is not necessarily expected to yield better results but its more towards aligning with best practices to have schema designed. Though this change is not tested with performance test plan but since query plan hasn't changed much, we would atleast get the same performance as of today.
And we might achieve better results with foreign keys implementation in future.
The motivation behind adding primary keys before GA as to do it sooner rather than later. It would be faster to do index changes, rebuild and backfill rows with smaller sizes of databases in production.
However, I am happy to discuss if you see any red flag in these changes and discuss through.

PR #1272, opened 2020-09-17
Added support for Patch interactions

anyname2:
## Description
Implemented Patch Operaion

## Related issues
Addresses #1268.

## Testing
Added Unit and E2E tests

Issue #1268, opened 2020-09-16
FHIR Patch


uxbridgeroad:
**User story**
As an MS FHIR Server user, I want use FHIR Patch so that as an alternative to updating an entire resource, clients can perform a patch operation. This can be useful when a client is seeking to minimize its bandwidth utilization, or in scenarios where a client has only partial access or support for a resource. T

http://hl7.org/fhir/http.html#patch

**Acceptance criteria**
1. When I do submit selected attributes on a resource, MS FHIR Server creates a new version of the resource which differs from the previous version only by the submitted attributes.

CaitlinV39:
@uxbridgeroad thank you so much for sharing this. We have $Patch on our list of enhancements but don't have a timeline yet for when we would start development. We will keep you updated as we progress.

uxbridgeroad:
@CaitlinV39 , this is great to hear. Is it possible to view the list of enhancements? As mentioned in (#695) Update Roadmap.md, it would be really useful to get a view on what is under consideration so we align our project features.

anyname2:
I have a draft PR for Patch ready #1272. But need this: https://github.com/FirelyTeam/fhir-net-api/pull/1375 to be merged and back-ported to a version FHIR Server is using (which I've done too: [R4](https://github.com/anyname2/fhir-net-api/tree/feature/fhir-patch-1.7.0-r4) [R5](https://github.com/anyname2/fhir-net-api/tree/feature/fhir-patch-1.7.0-r5) [STU3](https://github.com/anyname2/fhir-net-api/tree/feature/fhir-patch-1.7.0-stu3)).

CaitlinV39:
@uxbridgeroad as Michael mentioned in #695 we use open issues and PR to track features and give visibility into what we are releasing. We are thinking about how we can give more clarity into what we are releasing in upcoming months as well. If you'd like to jump on a call to share more ideas around this I'd love to hear them. Feel free to email me at cavoeg@microsoft.com

appetere:
Would you update https://docs.microsoft.com/en-us/azure/healthcare-apis/fhir/fhir-features-supported to say that PATCH is not supported, please?  I spent several hours trying to get it working, whereas it doesn't appear to be implemented yet?

![image](https://user-images.githubusercontent.com/1777396/128353651-d1d1a2e6-415e-4ba1-b381-9e9b87eea04d.png)

CaitlinV39:
[...]

PR #2383, opened 2021-11-29
FHIRPath Patch

mikaelweave:
## Description
Implements FHIRPath Patch as defined on [the HL7 website](http://hl7.org/fhir/fhirpatch.html). Co-coded with @mattpasco-microsoft and @stgricci.

##  Done
- [x] Implemented FHIRPatch
- [x] Check edge cases with FHIR chat group. Document behavior on undefined areas.
- [x] Re-use extensions and helper methods in `fhir-server` repo. (no `ElementNode` extensions exist elsewhere)
- [x] Complete unit testing.
- [x] Add E2E testing.
- [x] Fix FHIRPatch in bundles.
- [x] Add example `.http` file for FHIRPatch.

## Related issues
Addresses #1268. Related to unmerged PR #1272 by @valeneiko.

## FHIR Team Checklist


src/Microsoft.Health.Fhir.Core/Messages/Patch/ConditionalFhirPatchResourceRequest.cs
```diff
@@ -0,0 +1,39 @@
...
+using System.Collections.Generic;
+using EnsureThat;
+using Hl7.Fhir.Model;
+using Microsoft.AspNetCore.JsonPatch;
+using Microsoft.Health.Fhir.Core.Features.Persistence;
+using Microsoft.Health.Fhir.Core.Messages.Upsert;
+
+namespace Microsoft.Health.Fhir.Core.Messages.Patch
+{
+    public sealed class ConditionalFhirPatchResourceRequest : ConditionalResourceRequest<UpsertResourceResponse>
```
Ivanidzo4ka, line 16: >ConditionalFhirPatchResourceRequest [](http://example.com/codeflow?start=24&length=35)

This and FhirPatchResourceRequest looks excessive for me.
What are they doing what ConditionalPatchResourceRequest can't do?

src/Microsoft.Health.Fhir.Shared.Api/Controllers/FhirController.cs
```diff
@@ -445,6 +447,46 @@ public async Task<IActionResult> ConditionalPatch(string typeParameter, [FromBod
...
+        [HttpPatch]
+        [Route(KnownRoutes.ResourceType)]
+        [AuditEventType(AuditEventSubType.ConditionalPatch)]
+        [Consumes("application/fhir+json")]
+        public async Task<IActionResult> ConditionalPatchFhir(string typeParameter, [FromBody] Parameters paramsResource, [ModelBinder(typeof(WeakETagBinder))] WeakETag ifMatchHeader)
+        {
+            IReadOnlyList<Tuple<string, string>> conditionalParameters = GetQueriesForSearch();
+
+            UpsertResourceResponse response = await _mediator.Send<UpsertResourceResponse>(
+                new ConditionalFhirPatchResourceRequest(typeParameter, paramsResource, conditionalParameters, ifMatchHeader),
```
Ivanidzo4ka, line 482: >ConditionalFhirPatchResourceRequest [](http://example.com/codeflow?start=20&length=35)

This seems excessive for me. That whole class does same thing as ConditionalPatchRequest, with only difference is what is payload it carries.

I would do following:


Either make `PatchResourceRequest` generic by payload (either Parameters or JsonPatchDocument or just have two properites inside and two different constructors. 

Or make `PatchResourceRequest` an abstract class and derive two classes out of it FhirJson and Json

Rename JsonPatchService into PatchService and let it have either two public patch methods, or one public patch which would accept ConditionalPatchRequest and two private methods to handle patching. 

It feels like too much same code right now.

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Patch/FhirPathPatch/FhirPathPatchBuilder.cs
```diff
@@ -0,0 +1,155 @@
...
+        public FhirPathPatchBuilder(Resource resource, Parameters parameters)
+        : this(resource)
+        {
+            this.Build(parameters);
+        }
+
+        /// <summary>
+        /// Applies the list of pending operations to the resource.
+        /// </summary>
+        public Resource Apply()
```
Ivanidzo4ka, line 48: >Apply [](http://example.com/codeflow?start=24&length=5)

Since we use Firely maybe it make sense to check out their patch service implementation: [https://github.com/FirelyTeam/spark/blob/r4/master/src/Spark.Engine/Service/FhirServiceExtensions/PatchService.cs](https://github.com/FirelyTeam/spark/blob/r4/master/src/Spark.Engine/Service/FhirServiceExtensions/PatchService.cs)

src/Microsoft.Health.Fhir.Core/Messages/Patch/ConditionalFhirPatchResourceRequest.cs
mikaelweave: @mattpasco-microsoft can you comment on this one?

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Patch/FhirPathPatch/FhirPathPatchBuilder.cs
Ivanidzo4ka: My intent in comment was to make sure we don't reinvent bike, and it's good to know what you check other implementation.

mikaelweave: @Ivanidzo4ka - did some digging. Spark's implementation doesn't pass any tests involving [Anonymous Types](https://github.com/FHIR/fhir-test-cases/blob/752b01313ecbc1e13a942e1b3e25c96b3f7f3449/r5/patch/fhir-path-tests.xml#L450) or [Nested Primitives](https://github.com/FHIR/fhir-test-cases/blob/752b01313ecbc1e13a942e1b3e25c96b3f7f3449/r5/patch/fhir-path-tests.xml#L297). I remember trying to fix Spark's PatchService code, but failing since the code is complex. I'm going to do a simple performance comparison later just to make sure the two approaches aren't wildly different. As of right now, the approach in this PR is superior because it meets all of the FHIRPatch tests defined by HL7.

mikaelweave: @Ivanidzo4ka - performance difference is negligible testing with 100 sequential FHIRPatch `add` operations and 100 sequential FHIRPatch `replace` operations.

src/Microsoft.Health.Fhir.Core/Messages/Patch/ConditionalFhirPatchResourceRequest.cs
mattpasco-microsoft: Agreed, pushed some changes to reduce complexity so the only duplication is in the patch service itself

src/Microsoft.Health.Fhir.Shared.Api/Controllers/FhirController.cs
mattpasco-microsoft: I templated this and reduced complexity

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Patch/ConditionalPatchResourceHandler.cs
```diff
@@ -39,19 +42,27 @@ public sealed class ConditionalPatchResourceHandler : ConditionalResourceHandler
             EnsureArg.IsNotNull(modelInfoProvider, nameof(modelInfoProvider));
             EnsureArg.IsNotNull(mediator, nameof(mediator));
 
-            _patchService = new JsonPatchService(modelInfoProvider);
+            dynamic service;
```
mattpasco-microsoft, line 45: @Ivanidzo4ka @mikaelweave I templated away both the ResourceRequest and ConditionalPatchResourceRequest. Ideally we could guard the TData type to be either Parameters or JsonPatchDocument but I foresee defering type check logic somewhere else. I really hate this code for picking the service but im not sure a full factory pattern is necessary. My other idea would be to lazy an instance of the JsonPatchService and FhirPatchService.

Would love some more input here

mikaelweave: @mattpasco-microsoft this works but the resource handlers need to be registered explicitly with the DI container due to MediatR and possibly even .NET. [See here for details](https://github.com/jbogard/MediatR/issues/674).

I added the registration for the Patch handlers and used the container for`JsonPatchService` and `FhirParametersPatchService`. I added them where it made sense to me but should be reviewed by someone for familiar with the repo (@Ivanidzo4ka and/or @brendankowitz).

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Patch/BasePatchService.cs
```diff
@@ -0,0 +1,30 @@
...
+                "Resource.id",
+                "Resource.meta.lastUpdated",
+                "Resource.meta.versionId",
+                "Resource.text.div",
+                "Resource.text.status",
+            };
+
+        public abstract ResourceElement Patch(ResourceWrapper resourceToPatch, T paramsResource, WeakETag weakETag);
+
+        protected abstract Resource GetPatchedJsonResource(FhirJsonNode node, T operations);
```
Ivanidzo4ka, line 28: >protected abstract Resource GetPatchedJsonResource(FhirJsonNode node, T operations); [](http://example.com/codeflow?start=8&length=84)

nit: I don't think this been used outside of internal implementation of each patch service, so I would remove it from base implementation and keep each method as private inside specific implementations.

mattpasco-microsoft: Agreed, maybe there is some further refactoring we could do to move shared code.

mikaelweave:
@Ivanidzo4ka are there any docs or E2E test that we need to add for this PR? Besides those two items, this is about ready.

Ivanidzo4ka:
@mikaelweave `docs` can mean two things in my mind. One is for developers who work with this repository. Second is for our azure customers who use this server as a product, if you mean second one I would suggest you to talk to @CaitlinV39 since I don't have much expertise, and it probably would involve commit to different repository. If you thinking about first, I think doc/rest is bare minimum, and you can add more content into doc folder regarding how patch works in general, but something tells me you pursue second type of docs...

For E2E tests feel free to add more tests to this file:
https://github.com/microsoft/fhir-server/blob/main/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/PatchTests.cs 
Note that it uses client class https://github.com/microsoft/fhir-server/blob/6dfd829796954236bec67c8c71e98bbdba5bf6c3/src/Microsoft.Health.Fhir.Shared.Client/FhirClient.cs#L237 which currently hardcodes type of the content, so you will need to touch it as well (maybe by having two methods JsonPatch* and ParametersPatch* instead of Patch*)

CaitlinV39:
> @mikaelweave `docs` can mean two things in my mind. One is for developers who work with this repository. Second is for our azure customers who use this server as a product, if you mean second one I would suggest you to talk to @CaitlinV39 since I don't have much expertise, and it probably would involve commit to different repository. If you thinking about first, I think doc/rest is bare minimum, and you can add more content into doc folder regarding how patch works in general, but something tells me you pursue second type of docs...
> 
> For E2E tests feel free to add more tests to this file: https://github.com/microsoft/fhir-server/blob/main/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/PatchTests.cs Note that it uses client class
> 
> https://github.com/microsoft/fhir-server/blob/6dfd829796954236bec67c8c71e98bbdba5bf6c3/src/Microsoft.Health.Fhir.Shared.Client/FhirClient.cs#L237
> 
> which currently hardcodes type of the content, so you will need to touch it as well (maybe by having two methods JsonPatch* and ParametersPatch* instead of Patch*)

@mikaelweave - We would love if you could add a doc/rest that gives users examples of how to use the FHIRPath Patch. Once this is merged, I'll work to get Azure Docs updated with the details and ask for your review/input on things I don't know :) Typically the rest file helps me do this without having to hassle you as much!

mikaelweave:
> > @mikaelweave `docs` can mean two things in my mind. One is for developers who work with this repository. Second is for our azure customers who use this server as a product, if you mean second one I would suggest you to talk to @CaitlinV39 since I don't have much expertise, and it probably would involve commit to different repository. If you thinking about first, I think doc/rest is bare minimum, and you can add more content into doc folder regarding how patch works in general, but something tells me you pursue second type of docs...
> > For E2E tests feel free to add more tests to this file: https://github.com/microsoft/fhir-server/blob/main/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/PatchTests.cs Note that it uses client class
> > https://github.com/microsoft/fhir-server/blob/6dfd829796954236bec67c8c71e98bbdba5bf6c3/src/Microsoft.Health.Fhir.Shared.Client/FhirClient.cs#L237
> > 
> > which currently hardcodes type of the content, so you will need to touch it as well (maybe by having two methods JsonPatch* and ParametersPatch* instead of Patch*)
> 
> @mikaelweave - We would love if you could add a doc/rest that gives users examples of how to use the FHIRPath Patch. Once this is merged, I'll work to get Azure Docs updated with the details and ask for your review/input on things I don't know :) Typically the rest file helps me do this without having to hassle you as much!

Thanks for the info! We have a `.http` file in the docs folder! While re-testing everything, I did see that FHIRPatch in bundles errors because the correct patch type cannot be resolved. I'm going to move this back to draft until this and E2E testing gets resolved. I think E2E testing is needed since the patch type selector code is non-trivial.

I'm not working for the next 2 weeks, but can look once I'm back. @mattpasco-microsoft, if you have time could you look?

mikaelweave:
@Ivanidzo4ka E2E tests added, code is done. Ready for final review/testing.

CC @mattpasco-microsoft

mikaelweave:
@LTA-Thinking I can look today! Thanks for running the pipeline. 👍🏽

Ivanidzo4ka:
Failure are caused by making Handler (ConditionalPatch and just Patch) generic, and code which we use in tests, doesn't know how to create object for generic types (which makes sense.).

Sorry for diverting you on this path, and not thinking through all of it.

I'm ok with reverting code back to previous state without generics. As another option, instead of having TData as parameter we can introduce abstract class PatchPayload and combine it with `BasePatchService` by having `JsonPatchPayload` which knows how to patch  application/json and `ParametersPatchPayload` which knows how to patch application/parameters.

mikaelweave:
@Ivanidzo4ka no problem, good learning experience. I wasn't in love with this approach (especially changes to [MediationModule.cs](src/Microsoft.Health.Fhir.Shared.Api/Modules/MediationModule.cs)). I'll look at the abstract approach to reduce duplicated code. If it gets too complex, I'll just duplicate the code for easier understandability in the future.

mikaelweave:
@Ivanidzo4ka sorry for the delay, dealt with computer issues this week. Question about your approach - would `JsonPatchPayload` and `ParametersPatchPayload` be sent by the `PatchResourceRequest` function? Or would they be compiled in the handler?

After looking at this, I could easily make `PatchResourceRequest` and `PatchResourceHandler` abstract and then send different requests to different handlers with minimal code duplication. This would continue the MediatR pattern.

Ivanidzo4ka:
How I saw it in my head:
You have controller and controller creates handler request with either `PatchResourcePayload` or `ParameterResourcePayload`.

We have one handler which handles request by fetching document from DB, and asking Payload class to patch it, get result back and validate it.

mikaelweave:
I see what you are saying now, @Ivanidzo4ka. I changed the code but have one more question before final testing. I placed the `PatchPayload` files in `Microsoft.Health.Fhir.Shared.Core`. I cannot reference them from `Microsoft.Health.Fhir.Core` because adding a reference to Shared from that project creates a dependency loop. Have you seen this issue before in the fhir-server codebase?

Some possible solutions that come to mind:
- Define an interface in another assembly that is implemented in `Microsoft.Health.Fhir.Shared.Core`
- Define the classes in the `Microsoft.Health.Fhir.Core` and use extensions for the `PatchPayload` logic that should be in the Shared project
- Assemble the payload in Shared and use separate Requests to know which to build

mikaelweave:
Thanks for the help @LTA-Thinking, this is the largest project I've worked on and want to make sure I stick with the established patterns.

Ivanidzo4ka:
Let's talk first about difference between `Shared.Core` and `Core` and why we even have two projects.

Idea of having `Core` project should be transparent, I hope, bunch of basic classes we can use in different parts of solution. Very basic things applicable for anything in project.

Why we have `Shared.Core`? Well, we work with FHIR spec, and sometimes we need to work with specific entities defined in FHIR spec. As usual, let's pick Patient entity. There is 3 different variations of FHIR spec we work with: STU3, R4, R5. In all three of them Patient entity exist. But in some of them it has certain properties, and in some it doesn't. Spec evolves and changes over time, and STU3, R4,R5 is just snapshots of the spec in certain time.

So I want to have class which has reference on Patient class, how I do it? We do it by having 3 different projects (STU3, R4, and R5) and having identical classes with reference to Patient class in them. Copying code around 3 different projects, is cumbersome and error prone task.

To make our life easier, we have `Shared.Core` project which has code for all 3 projects, and each project just reuses code but for each project meaning of Patient class is different. Basically `Shared.Core` is collection of templates or generics which we resolve during compilation into specific classes.

Wherefore `Shared.Core` is imbued with domain knowledge, while classes in `Core` are basic things which have no knowledge of FHIR spec.

You put `*Payload` classes in `Shared.Core` because they have references on FHIR Spec classes I assume (Parameters`).
And you define Handler request in core, because it's basic class, untouched by FHIR spec specifics.
But you also have problem of putting `*Payload` into request. This is typical inversion of dependency problem.

I personally would go with abstract class `PatchPayload` in `Core` project which would have ` ResourceElement Patch(ResourceWrapper resourceToPatch, WeakETag weakETag)` method. 

You can have concreate implementations of this class (Parameters, Json) in `Shared.Core` project.
`FhirController` I think would be the place where you have to create concreate implementations, and `FhirController` already knows about specific FHIR spec, so shouldn't be a problem. (`FhirController` lives in another `Shared` -> `Shared.Web` project)

I personally would choose abstract class over interface, because I don't expect someone else have functionality of payload. But if you find it's easier for you to implement that way, go for it.

mikaelweave:
Thanks for the detailed explanation @Ivanidzo4ka. This clears up a LOT on the repository. The code reuse between STU3, R4, and R5 was obvious. Beyond that, I didn't understand the difference between `Core` and `Shared`.  I saw the separation but didn't understand the *why* behind the separation. I distinctly remember being unsure about where to put the Patch library initially and think I got lucky.

I'm also not super strong in design patterns, so keep that in mind while forming your persona of how a developer in my position understands the repo. I know what they are, but until you get to use them a lot you tend to miss things. 😄  

One piece of feedback (I can create an issue after the PR) is I looked for Developer docs and couldn't find any in the root of the repo or the `docs` folder. The sequence diagrams were helpful, but some architectural guidance in a `DEVELOPERS.md` file would help outside contributors.

❓  Final question, I put `PatchPayload in Core.Features.Operations.Resources` to match the other projects and keep everything in the same namespace. Any issues here? Not sure if it's super important but it made sense to me. This required the updating of some string definitions from starting with `Resources` to `Core.Resources`.

Seriously, THANK YOU for the long comment explaining your reasoning. I have become a better developer thanks to the time you've put into that comment and this PR.

LTA-Thinking:
@CaitlinV39 Adding developer focused documentation would be helpful for our open source contributors, and may encourage more contributions like this.

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Patch/FhirPathPatch/FhirPathPatchBuilder.cs
```diff
@@ -0,0 +1,136 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using FhirPathPatch.Operations;
+using Hl7.Fhir.Model;
+using static Hl7.Fhir.Model.Parameters;
+
+namespace FhirPathPatch
```
brendankowitz, line 12: Just a small one, namespaces don't look consistent with the rest of the project

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Patch/FhirPathPatch/Operations/EOperationType.cs
```diff
@@ -0,0 +1,38 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace FhirPathPatch.Operations
+{
+    /// <summary>
+    /// An enumeration of the supported types of FHIR Path Patch operations.
+    /// </summary>
+    public enum EOperationType
```
brendankowitz, line 11: Perhaps `PatchOperationType` for consistency?

src/Microsoft.Health.Fhir.Core/Messages/Patch/ConditionalPatchResourceRequest.cs
```diff
@@ -18,18 +18,18 @@ public sealed class ConditionalPatchResourceRequest : ConditionalResourceRequest
...
-            EnsureArg.IsNotNull(patchDocument, nameof(patchDocument));
+            EnsureArg.IsNotNull(payload, nameof(payload));
 
-            PatchDocument = patchDocument;
+            Payload = payload;
             WeakETag = weakETag;
         }
 
-        public JsonPatchDocument PatchDocument { get; }
+        public PatchPayload Payload { get; }
```
brendankowitz, line 32: I think this is good, always felt a little strange having the asp.net JsonPatchDocument here directly :)

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Patch/FhirPathPatch/FhirPathPatchBuilder.cs
mikaelweave: Good catch, thanks. We wrote this as a standalone library and never updated this. I've combed the logic piece of FHIRPatch and cleanup up a bunch of things.

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Patch/FhirPathPatch/Operations/EOperationType.cs
mikaelweave: Done, makes more sense. I didn't prefix `FhirPathPatch` on any of the Operation classes because all the code in `FhirPathPatch` is internal except for the builder and a helper used for testing.

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Patch/FhirPathPatch/Operations/IOperation.cs
```diff
@@ -0,0 +1,21 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using Hl7.Fhir.Model;
+
+namespace FhirPathPatch.Operations
+{
+    /// <summary>
+    /// Patch operations Interface
+    /// </summary>
+    public interface IOperation
```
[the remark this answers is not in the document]
mikaelweave: Removed the interface, it wasn't giving any benefit. No code outside `FhirPathPatch' needs to know the operation details.

src/Microsoft.Health.Fhir.Core/Messages/Patch/ConditionalPatchResourceRequest.cs
mikaelweave: Agreed! Big thanks to Ivan for the help and contributions to the PR by pushing us in the right direction.

docs/rest/FhirPatchRequests.http
```diff
@@ -0,0 +1,281 @@
...
+                    "name": "path",
+                    "valueString": "Patient.telecom.where(use = 'home')"
+                }
+            ]
+        }
+    ]
+}
+
+###
+## Add a phone number
```
LTA-Thinking, line 99: Super nitpick: I think this example should be above the example that removes a phone number to allow the remove example to demonstrate the functionality of the condition in the valueString. Right now there is only one phone number so it is unclear if the conditional statement is working.

mikaelweave: Good callout. I'm going to run all the rest requests again in a couple of hours for a sanity check.

mikaelweave: All looks good on these @LTA-Thinking. Tested and still working as intended.

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Resources/Patch/FhirPatchDeleteTests.cs
```diff
@@ -0,0 +1,259 @@
...
+                },
+                patchParam).Apply();
+
+            Assert.True(patchedPatientResource.Matches(new Patient()));
+        }
+
+        // Implements test case at:
+        // https://github.com/FHIR/fhir-test-cases/blob/752b01313ecbc1e13a942e1b3e25c96b3f7f3449/r5/patch/fhir-path-tests.xml#L494
+        [Fact]
+        public void GivenAFhirPatchDeleteRequest_WhenDeletingCDeleteAnonymousObject_ThenAnonymousShouldBeRemoved()
```
LTA-Thinking, line 124: Nit: Typo in the test name

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Resources/Patch/FhirPatchInsertTests.cs
```diff
@@ -0,0 +1,155 @@
...
+        public void GivenAFhirPatchInsertRequest_WhenInsertingToUninitializedList_ThenInvalidOperationExceptionShouldBeThrown()
+        {
+            var patchParam = new Parameters().AddInsertPatchParameter("Patient.identifier", new Identifier { System = "http://example.org", Value = "value 3" }, 2);
+
+            var builder = new FhirPathPatchBuilder(new Patient(), patchParam);
+            Assert.Throws<InvalidOperationException>(builder.Apply);
+        }
+
+        [Fact]
+        public void GivenAFhirPatchInsertRequest_WhenInsertingUninitializedList_ThenInvalidOperationExceptionShouldBeThrown()
```
LTA-Thinking, line 124: Nit: I think the test name is wrong. It looks like this test is for an index that is too high, not an uninitialized list.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ConditionalPatchTests.cs
```diff
@@ -63,16 +74,23 @@ public async Task GivenAResource_WhenPatchingConditionallyWithOneMatch_TheServer
...
-            Assert.Equal(AdministrativeGender.Female, patchResponse.Resource.Gender);
-            Assert.Empty(patchResponse.Resource.Address);
-            Assert.Equal(HttpStatusCode.OK, patchResponse.StatusCode);
+                _fhirPatchRequest);
+
+            Assert.Equal(AdministrativeGender.Female, jsonPatchResponse.Resource.Gender);
+            Assert.Empty(jsonPatchResponse.Resource.Address);
+            Assert.Equal(HttpStatusCode.OK, jsonPatchResponse.StatusCode);
+
+            Assert.Equal(AdministrativeGender.Female, fhirPatchResponse.Resource.Gender);
```
LTA-Thinking, line 91: For this test the JSON and FHIR patches should be in separate tests. Since they are working on the same patient if the JSON one succeeds then there are no unique changes made by the FHIR patch to check that it worked properly.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ConditionalPatchTests.cs
```diff
@@ -117,17 +137,24 @@ public async Task GivenSecondVersionOfResource_WhenPatchingConditionallyWithOneM
...
+                "Patient",
+                $"identifier={identifier}",
+                _fhirPatchRequest,
+                "3");
+
+            Assert.Equal(AdministrativeGender.Female, jsonPatchResponse.Resource.Gender);
+            Assert.Empty(jsonPatchResponse.Resource.Address);
+            Assert.Equal(HttpStatusCode.OK, jsonPatchResponse.StatusCode);
+
+            Assert.Equal(AdministrativeGender.Female, fhirPatchResponse.Resource.Gender);
```
LTA-Thinking, line 155: Same comment as above. When changes are being made there should be separate tests.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/FhirPathPatchTests.cs
```diff
@@ -0,0 +1,277 @@
...
+                .AddDeletePatchParameter("Patient.address");
+
+            FhirResponse<Patient> patch = await _client.FhirPatchAsync(response.Resource, patchRequest);
+
+            Assert.Equal(HttpStatusCode.OK, patch.Response.StatusCode);
+            Assert.Equal(AdministrativeGender.Female, patch.Resource.Gender);
+            Assert.Empty(patch.Resource.Address);
+        }
+
+        public static IEnumerable<object[]> GeIncorrectRequestTestData()
```
LTA-Thinking, line 55: Nit: Typo in method name.

LTA-Thinking (commented):
Finished with the test code. These were the last files for me to look at. We should be good to go after these changes are in. Thanks for working with us on this change, it is a big one.

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Resources/Patch/FhirPatchInsertTests.cs
mikaelweave: Test name was right, content misleading/wrong. Changed insert index to 0 to actually ensure the error we are getting matches the intended test case.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ConditionalPatchTests.cs
mikaelweave: Agreed. Was trying to be cute to remove duplication of code but it didn't really get us much. I separated these out.

mikaelweave:
@LTA-Thinking Thanks for the careful eye, I definitely left off some items in my final check. I'll go through the whole PR again just to be sure.

Happy to work with you all on this to produce high quality code I can be proud of. I hope to commit more in the future if I see other opportunities to enable use cases. 👍🏽

src/Microsoft.Health.Fhir.Core/Features/Operations/Resources/Patch/PatchPayload.cs
```diff
@@ -0,0 +1,54 @@
...
+            };
+
+        internal abstract ResourceElement GetPatchedResourceElement(ResourceWrapper resourceToPatch);
+
+        public ResourceElement Patch(ResourceWrapper resourceToPatch)
+        {
+            EnsureArg.IsNotNull(resourceToPatch, nameof(resourceToPatch));
+
+            // Capture the state of properties that are immutable
+            ITypedElement resource = resourceToPatch.RawResource.ToITypedElement(ModelInfoProvider.Instance);
```
mikaelweave, line 38: @LTA-Thinking - here I went from an `IModelInfoProvider` provided by the DI container to this statically provided version. Any issues with this approach?

I changed to this approach as `IModelInfoProvider` isn't injected in `FhirController`.

LTA-Thinking: This is the right way to do it. We use this pattern in other parts of our code.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ConditionalPatchTests.cs
LTA-Thinking: This one still needs to be split into two tests.

mikaelweave: I'm going to blame this on yesterday being Monday and point at the cleanup commit I did yesterday to distract you. 😆 

But seriously, missed this one, it's now split into multiple tests. Thanks.

LTA-Thinking (approved):

LTA-Thinking:
Thanks @mikaelweave and @mattpasco-microsoft!

PR #2384, opened 2021-11-30
Adds Sync to Devops workflow

brendankowitz:
## Description
Adds Sync to Devops workflow

## FHIR Team Checklist

Skip

### DIFF
--- a/.github/workflows/sync-to-devops.yml
+++ b/.github/workflows/sync-to-devops.yml
@@ -1,0 +1,26 @@
+name: Sync issue to Azure DevOps work item
+
+on:
+  issues:
+    types:
+      [opened, edited, deleted, closed, reopened, labeled, unlabeled, assigned]
+
+jobs:
+  alert:
+    runs-on: ubuntu-latest
+    name: Sync to DevOps workflow
+    steps:
+      - uses: danhellem/github-actions-issue-to-work-item@master
+        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/sync-to-devops.yml
```diff
@@ -0,0 +1,26 @@
...
+    steps:
+      - uses: danhellem/github-actions-issue-to-work-item@master
+        env:
+          ado_token: "${{ secrets.AZURE_DEVOPS_TOKEN }}"
+          github_token: "${{ secrets.GITHUB_TOKEN }}"
+          ado_organization: "${{ secrets.DEVOPS_ORGANIZATION }}"
+          ado_project: "${{ secrets.DEVOPS_PROJECT }}"
+          ado_area_path: "${{ secrets.DEVOPS_AREA_PATH }}"
+          ado_iteration_path: "${{ secrets.DEVOPS_ITERATION_PATH }}"
+          ado_wit: "User Story"
```
LTA-Thinking, line 21: Why make them user stories instead of bugs?

brendankowitz: I don't know which is the right way to go here, though we have 234 issues with the Bug label out of 685 total.

LTA-Thinking: Ok, makes sense. I would like some way of tagging these stories or putting them under a triage feature, but it doesn't look like this action allows that.

LTA-Thinking (approved):

PR #2390, opened 2021-12-02
add predicate for use fhir service

peizhou298:
## Description
This PR is add a predicate arguments in UseFhirServer function to enable the service filter the health check services it want to call, without it every registered services will be called in each health check.

## Related issues
Addresses [issue #].

## Testing

## FHIR Team Checklist


src/Microsoft.Health.Fhir.Api/Registration/FhirServerApplicationBuilderExtensions.cs
```diff
@@ -23,10 +24,21 @@ public static class FhirServerApplicationBuilderExtensions
         /// <param name="app">The application builder instance.</param>
         /// <returns>THe application builder instance.</returns>
         public static IApplicationBuilder UseFhirServer(this IApplicationBuilder app)
+        {
+            return app.UseFhirServer(null);
```
LTA-Thinking, line 28: Nit: Should this method have an EnsureArg for app like the other UseFhirServer method?

LTA-Thinking (approved):

LTA-Thinking:
This looks good. Where is this planned to be used?

src/Microsoft.Health.Fhir.Api/Registration/FhirServerApplicationBuilderExtensions.cs
```diff
@@ -23,10 +24,21 @@ public static class FhirServerApplicationBuilderExtensions
...
+        }
+
+        /// <summary>
+        /// Adds FHIR server functionality to the pipeline with health check filter.
+        /// </summary>
+        /// <param name="app">The application builder instance.</param>
+        /// <param name="predicate">The predicate used to filter health check services.</param>
+        /// <returns>THe application builder instance.</returns>
+        public static IApplicationBuilder UseFhirServer(this IApplicationBuilder app, Func<HealthCheckRegistration, bool> predicate)
```
brendankowitz, line 37: Could these two methods just be collapsed with a default value?
```suggestion
        public static IApplicationBuilder UseFhirServer(this IApplicationBuilder app, Func<HealthCheckRegistration, bool> predicate = null)
```

peizhou298:
> 



> This looks good. Where is this planned to be used?

This is used in Paas, the scenario is we need a to add a new health check service there, and we don't want default health check endpoint(/health/check) contains this service and added endpoint(health/check) only contain this, so we need a predicate to achieve it.

src/Microsoft.Health.Fhir.Api/Registration/FhirServerApplicationBuilderExtensions.cs
peizhou298: yes, it could. Already changed to that format.

brendankowitz:
CI was not green before merge

PR #2391, opened 2021-12-02
Adds clustered Id index for resource changes

bbatjargal:
## Description
This PR:
- Removes the existing primary key clustered index and adds a clustered index on the Id column in the ResourceChangeData table.
- Adds schema version 24.

Under the heavy load, we identified the following two issues during a validation test local. But we haven't seen any incidents related to the issues on the PROD.

1. Identity and timestamp columns are out of order (50+ RPS on local).
2. After the publish worker has fetched the list of resource changes, a record is inserted with an identity value that is greater than the min value and less than the max value of the batch (200/400+ RPS on local). 

- For the first issue, we see instances like the following in the resourceChangeData table, identity and timestamp columns are out of order. Since we get resource changes ordering by Timestamp and Id, it leads that a publish worker skips some resource changes without publishing. Therefore, we are removing the existing PK clustered index with Timestamp and Id, and adding a new clustered index on the Id column. Then, resource changes will be ordered only by the Id column for fetching.

Publish worker would fetch resource changes like the following when ordering by Timestamp and Id in ascending.
**Id | Timestamp**
1045 | 2021-11-22 17:22:12.2114800
1046 | 2021-11-22 17:22:12.2114800
**1049 | 2021-11-22 17:22:12.2114800**
**1047 | 2021-11-22 17:22:12.2124778**
1048 | 2021-11-22 17:22:12.2124778
1050 | 2021-11-22 17:22:12.2124778

- To address the second issue, we are adding TABLOCK and HOLDLOCK table hints on the fetch query to acquire and hold a table lock to prevent new resource changes from being created during the select query execution.

## Related issues
Addresses [issue #].

## Testing
Validated the changes locally. Run unit tests and integration tests.

## FHIR Team Checklist


src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/24.diff.sql
```diff
@@ -0,0 +1,146 @@
...
+    @startId bigint,
+    @lastProcessedUtcDateTime datetime2(7),
+    @pageSize smallint
+AS
+BEGIN
+   
+    SET NOCOUNT ON;
+
+    /* Finds the prior partition to the current partition where the last processed watermark lies. It is a normal scenario when a prior watermark exists. */
+    DECLARE @precedingPartitionBoundary AS datetime2(7) = (SELECT TOP(1) CAST(prv.value as datetime2(7)) AS value FROM sys.partition_range_values AS prv WITH (NOLOCK)
```
dustinburson, line 42: I think with the changes we no longer need to check the prior watermark partition.  Only the current partitions and the next.  Was there specific scenario the prior one was needed for in the latest testing?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/24.diff.sql
```diff
@@ -0,0 +1,146 @@
...
+        WHERE pf.name = N'PartitionFunction_ResourceChangeData_Timestamp'
+                AND SQL_VARIANT_PROPERTY(prv.Value, 'BaseType') = 'datetime2'
+                AND (
+                    /* Normal logic when prior watermark exists, Grab prior partition to ensure we do not miss data that was written across a partition boundary,
+                       and includes partitions until current one to ensure we keep moving to the next partition when some partitions do not have any resource changes. */
+                    CAST(prv.value AS datetime2(7)) BETWEEN @precedingPartitionBoundary AND @endDateTimeToFilter
+                    OR 
+                    /* It happens when no prior watermark exists or the last processed datetime of prior watermark is older than the last retention datetime.
+                       Uses the partition anchor datetime as the last processed DateTime. */
+                    @precedingPartitionBoundary IS NULL AND CAST(prv.value AS datetime2(7)) BETWEEN CONVERT(datetime2(7), N'1970-01-01T00:00:00.0000000') AND @endDateTimeToFilter
```
dustinburson, line 65: Suggestion: It would be cleaner to have an ISNULL check on the precedingPartitionBoundary outside the query.  I.E. if precedingPartitionBoundary is null just set it to the the default.  This will remove the OR and make the query a little simpler and I think the procedure a little more readable.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/24.diff.sql
```diff
@@ -0,0 +1,146 @@
...
+        /* Given the fact that Read Committed Snapshot isolation level is enabled on the FHIR database,
+           using TABLOCK and HOLDLOCK table hints to avoid skipping resource changes 
+           due to interleaved transactions on the resource change data table. */
+        SELECT TOP(@pageSize) Id,
+            Timestamp,
+            ResourceId,
+            ResourceTypeId,
+            ResourceVersion,
+            ResourceChangeTypeId
+        /* Acquires and holds the table lock to prevent new resource changes from being created during the select query execution. */
```
dustinburson, line 85: Suggestion: Extend comment to indicate that without the TABLOCK & HOLDLOCK key and row locks will be used which can result in missed reads if an insert occurs in range that was previously read and the locks were released on the read.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Tables/ResourceChangeData.sql
```diff
@@ -6,6 +6,8 @@ CREATE TABLE dbo.ResourceChangeData
     ResourceId varchar(64) NOT NULL,
     ResourceTypeId smallint NOT NULL,
     ResourceVersion int NOT NULL,
-    ResourceChangeTypeId tinyint NOT NULL,
-    CONSTRAINT PK_ResourceChangeData_TimestampId PRIMARY KEY (Timestamp, Id)
```
dustinburson, line 10: Suggestion:  Might be good to explain why we are making this change:
* This allows us to remove the partition column from the PK which leads to better query performance.
*  Uniqueness was only per partition before, i.e. same key could occur in different partitions.
*  We are using identity which will guarantee uniqueness unless identity insert is used on the table which shouldn't happen.
* To enforce global uniqueness requires a non clustered index without a partition which prevents partition swaps.

dustinburson (approved):
Overall looks good.  Have some minor suggestions.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/24.diff.sql
bbatjargal: Sure, updated and checked the NULL logic outside the select query.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/24.diff.sql
bbatjargal: Added an additional comment.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/24.diff.sql
bbatjargal: I believe that there is a scenario of a record would be inserted into the prior partition, with an identity value that is greater than the current watermark id.  For example:

Let's assume the last processed id is 49. Now, the current watermark will be 50.

Id | Timestamp | Partition number
49 2021-12-10 12:00:00.2000008             1
51 2021-12-10 11:59:59.9999988             1
50 2021-12-10 12:00:00.2222222             2

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Tables/ResourceChangeData.sql
```diff
@@ -6,6 +6,8 @@ CREATE TABLE dbo.ResourceChangeData
...
-) ON PartitionScheme_ResourceChangeData_Timestamp(Timestamp)
+    ResourceChangeTypeId tinyint NOT NULL
+) ON PartitionScheme_ResourceChangeData_Timestamp(Timestamp);
+
+CREATE CLUSTERED INDEX IXC_ResourceChangeData ON dbo.ResourceChangeData
+    (Id ASC) WITH(ONLINE = ON) ON PartitionScheme_ResourceChangeData_Timestamp(Timestamp);
```
rbans96, line 13: Is there a reason, we are not creating primary key on Id column?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/24.diff.sql
```diff
@@ -0,0 +1,149 @@
...
+        FROM dbo.ResourceChangeData WITH (TABLOCK, HOLDLOCK)
+            WHERE Id >= @startId
+                AND $PARTITION.PartitionFunction_ResourceChangeData_Timestamp(Timestamp) = $PARTITION.PartitionFunction_ResourceChangeData_Timestamp(p.PartitionBoundary)
+        ORDER BY Id ASC
+        ) AS rcd
+    ORDER BY rcd.Id ASC;
+END;
+GO
+
+EXEC dbo.LogSchemaMigrationProgress 'Deleting PK_ResourceChangeData_TimestampId index from ResourceChangeData table.';
```
rbans96, line 98: nit: Would it be more meaningful to add this after
IF EXISTS (SELECT 1 FROM sys.indexes WHERE name = 'PK_ResourceChangeData_TimestampId')
BEGIN

as it will only delete if already exists.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/24.diff.sql
```diff
@@ -0,0 +1,149 @@
...
+IF EXISTS (SELECT 1 FROM sys.indexes WHERE name = 'PK_ResourceChangeData_TimestampId')
+BEGIN
+    /* Drops PK_ResourceChangeData_TimestampId index. "ONLINE = ON" indicates long-term table locks aren't held for the duration of the index operation. 
+       During the main phase of the index operation, only an Intent Share (IS) lock is held on the source table. 
+       This behavior enables queries or updates to the underlying table and indexes to continue. */
+    ALTER TABLE dbo.ResourceChangeData DROP CONSTRAINT PK_ResourceChangeData_TimestampId WITH(ONLINE = ON);
+END;
+GO
+
+EXEC dbo.LogSchemaMigrationProgress 'Deleting PK_ResourceChangeDataStaging_TimestampId index from ResourceChangeDataStaging table.';
```
rbans96, line 110: Same comment as above

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/24.diff.sql
```diff
@@ -0,0 +1,149 @@
...
+IF EXISTS (SELECT 1 FROM sys.indexes WHERE name = 'PK_ResourceChangeDataStaging_TimestampId')
+BEGIN
+    /* Drops PK_ResourceChangeDataStaging_TimestampId index. "ONLINE = ON" indicates long-term table locks aren't held for the duration of the index operation. 
+       During the main phase of the index operation, only an Intent Share (IS) lock is held on the source table. 
+       This behavior enables queries or updates to the underlying table and indexes to continue. */
+    ALTER TABLE dbo.ResourceChangeDataStaging DROP CONSTRAINT PK_ResourceChangeDataStaging_TimestampId WITH(ONLINE = ON);
+END;
+GO
+
+EXEC dbo.LogSchemaMigrationProgress 'Creating IXC_ResourceChangeData index on ResourceChangeData table.';
```
rbans96, line 122: Same comment as above

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/24.diff.sql
```diff
@@ -0,0 +1,149 @@
...
+BEGIN
+    /* Adds a clustered index on ResourceChangeData table. "ONLINE = ON" indicates long-term table locks aren't held for the duration of the index operation. 
+       During the main phase of the index operation, only an Intent Share (IS) lock is held on the source table. 
+       This behavior enables queries or updates to the underlying table and indexes to continue. */
+    CREATE CLUSTERED INDEX IXC_ResourceChangeData ON dbo.ResourceChangeData
+        (Id ASC) WITH(ONLINE = ON) ON PartitionScheme_ResourceChangeData_Timestamp(Timestamp);
+END;
+GO
+
+EXEC dbo.LogSchemaMigrationProgress 'Creating IXC_ResourceChangeDataStaging index on ResourceChangeDataStaging table.';
```
rbans96, line 135: Same comment as above

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Tables/ResourceChangeData.sql
bbatjargal: Yes, the reason is to have a better performance on the fetch query. Also, since a resourceChangeData table is partitioned on timestamp, we can not create the primary key only on the Id column due to a SQL constraint, "partition columns for a unique index must be a subset of the index key".

Previously, the uniqueness was combined with the timestamp column and it was only per partition. To enforce global uniqueness requires a non clustered index without a partition but which prevents partition swaps.

We are using identity which will guarantee uniqueness unless an identity insert is used or reseed identity value on the table which shouldn't happen.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/24.diff.sql
bbatjargal: Moved the log inside the IF block.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/24.diff.sql
bbatjargal: Moved the log inside the IF block.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/24.diff.sql
bbatjargal: Moved the log inside the IF block.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/24.diff.sql
bbatjargal: Moved the log inside the IF block.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Tables/ResourceChangeData.sql
bbatjargal: Added a comment.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Tables/ResourceChangeData.sql
bbatjargal: Added a comment on the index creation in a ResourceChangeData table's SQL file, and diff file.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/24.diff.sql
```diff
@@ -0,0 +1,152 @@
...
+GO
+
+IF EXISTS (SELECT 1 FROM sys.indexes WHERE name = 'PK_ResourceChangeData_TimestampId')
+BEGIN
+    EXEC dbo.LogSchemaMigrationProgress 'Deleting PK_ResourceChangeData_TimestampId index from ResourceChangeData table.';
+
+    /* Drops PK_ResourceChangeData_TimestampId index. "ONLINE = ON" indicates long-term table locks aren't held for the duration of the index operation. 
+       During the main phase of the index operation, only an Intent Share (IS) lock is held on the source table. 
+       This behavior enables queries or updates to the underlying table and indexes to continue. */
+    ALTER TABLE dbo.ResourceChangeData DROP CONSTRAINT PK_ResourceChangeData_TimestampId WITH(ONLINE = ON);
```
rbans96, line 105: Is the primary key which we are dropping here or below clustered or nonclustered. Since if it is nonclustered, would it make sense to create clustered index first and then drop primary key for both the tables.

bbatjargal: The existing primary key clustered is clustered primary key index. So, we have to drop first before creating another clustered index.

rbans96: Thanks for confirming it!

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/24.diff.sql
```diff
@@ -0,0 +1,152 @@
...
+    /* Drops PK_ResourceChangeDataStaging_TimestampId index. "ONLINE = ON" indicates long-term table locks aren't held for the duration of the index operation. 
+       During the main phase of the index operation, only an Intent Share (IS) lock is held on the source table. 
+       This behavior enables queries or updates to the underlying table and indexes to continue. */
+    ALTER TABLE dbo.ResourceChangeDataStaging DROP CONSTRAINT PK_ResourceChangeDataStaging_TimestampId WITH(ONLINE = ON);
+END;
+GO
+
+IF NOT EXISTS (SELECT 1 FROM sys.indexes WHERE name = 'IXC_ResourceChangeData')
+BEGIN
+    EXEC dbo.LogSchemaMigrationProgress 'Creating IXC_ResourceChangeData index on ResourceChangeData table.';
```
rbans96, line 122: Might be this can move upward just after dropping the primary key to ensure that table is not left without index for longer time

bbatjargal: A reason we put the order in a current way is that:
- online index operation is used.
- partition switch-out operation requires that the main table and staging table have exactly the same structure and indexes. 
Considering these, we would like to keep the two table structures and indexes the same if migration is failed right after dropping the index from a staging table or right before creating an index on the main table.
  
I would like to highlight that dropping an index from a staging table will be very fast due staging table is always empty, except during partition switch-out operation.

Ivanidzo4ka (approved):
:shipit:

PR #2394, opened 2021-12-08
Updates to failing history test

feordin:
## Description
Attempts to improve the stability of the history test by changing thread.sleep to Task.Delay and also adding a retry loop.

## Related issues
Addresses [issue [AB#86231](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/86231)].

## Testing
Test was rerun many times

## FHIR Team Checklist

### DIFF
--- a/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/HistoryTests.cs
+++ b/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/HistoryTests.cs
@@ -194,52 +194,78 @@ namespace Microsoft.Health.Fhir.Tests.E2E.Rest
         [Fact]
         public async Task GivenAValueForSinceAndBeforeCloseToLastModifiedTime_WhenGettingSystemHistory_TheServerShouldNotMissRecords()
         {
-            var tag = Guid.NewGuid().ToString();
-
-            var since = await GetStartTimeForHistoryTest(tag);
-
-            var newResources = new List<Resource>();
-
-            // First make a few edits
-            _createdResource.Resource.Text = new Narrative { Div = "<div>Changed by E2E test</div>" };
-            await CreateResourceWithTag(_createdResource.Resource, tag);
-
-            newResources.Add(await CreateResourceWithTag(Samples.GetDefaultPatient().ToPoco(), tag));
-            newResources.Add(await CreateResourceWithTag(Samples.GetDefaultOrganization().ToPoco(), tag));
-            Thread.Sleep(1000);
-            newResources.Add(await CreateResourceWithTag(Samples.GetJsonSample("BloodGlucose").ToPoco(), tag));
-            newResources.Add(await CreateResourceWithTag(Samples.GetJsonSample("BloodPressure").ToPoco(), tag));
-            newResources.Add(await CreateResourceWithTag(Samples.GetJsonSample("Patient-f001").ToPoco(), tag));
-            newResources.Add(await CreateResourceWithTag(Samples.GetJsonSample("Condition-For-Patient-f001").ToPoco(), tag));
-
-            var sinceUriString = HttpUtility.UrlEncode(since.ToString("o"));
-
-            // Query all the recent changes
-            using FhirResponse<Bundle> allChanges = await _client.SearchAsync($"_history?_tag={tag}&_since=" + sinceUriString);
-
-            AssertCount(7, allChanges.Resource.Entry);
-
-            // now choose a value of before that is as close as possible to one of the last updated times
-            var lastUpdatedTimes = allChanges.Resource.Entry.Select(e => e.Resource.Meta.LastUpdated).OrderBy(d => d.Value);
-
-            var before = lastUpdatedTimes.ToList()[4];
-            var beforeUriString = HttpUtility.UrlEncode(before.Value.ToString("o"));
-            Thread.Sleep(500);
-            var firstSet = await _client.SearchAsync($"_history?_tag={tag}&_since={sinceUriString}&_before={beforeUriString}");
-
-            AssertCount(4, firstSet.Resource.Entry);
-
-            sinceUriString = beforeUriString;
-            before = DateTime.UtcNow;
-            beforeUriString = HttpUtility.UrlEncode(before.Value.ToString("o"));
-            Thread.Sleep(500); // wait 500 milliseconds to make sure that the value passed to the server for _before is not a time in the future
-            var secondSet = await _client.SearchAsync($"_history?_tag={tag}&_since={sinceUriString}&_before={beforeUriString}");
-
-            AssertCount(3, secondSet.Resource.Entry);
-
-            foreach (var r in newResources)
+            // as this test is timing dependent, we will try it three times
+            // to see if it can pass
+            for (int i = 0; i < 3; i++)
             {
-                await _client.DeleteAsync(r);
+                bool success = false;
+                var newResources = new List<Resource>();
+
+                try
+                {
+                    var tag = Guid.NewGuid().ToString();
+
+                    var since = await GetStartTimeForHistoryTest(tag);
+
+                    // First make a few edits
+                    _createdResource.Resource.Text = new Narrative { Div = "<div>Changed by E2E test</div>" };
+                    await CreateResourceWithTag(_createdResource.Resource, tag);
+
+                    newResources.Add(await CreateResourceWithTag(Samples.GetDefaultPatient().ToPoco(), tag));
+                    newResources.Add(await CreateResourceWithTag(Samples.GetDefaultOrganization().ToPoco(), tag));
+                    await Task.Delay(1000);
+                    newResources.Add(await CreateResourceWithTag(Samples.GetJsonSample("BloodGlucose").ToPoco(), tag));
+                    newResources.Add(await CreateResourceWithTag(Samples.GetJsonSample("BloodPressure").ToPoco(), tag));
+                    newResources.Add(await CreateResourceWithTag(Samples.GetJsonSample("Patient-f001").ToPoco(), tag));
+                    newResources.Add(await CreateResourceWithTag(Samples.GetJsonSample("Condition-For-Patient-f001").ToPoco(), tag));
+
+                    var sinceUriString = HttpUtility.UrlEncode(since.ToString("o"));
+
+                    // Query all the recent changes
+                    using FhirResponse<Bundle> allChanges = await _client.SearchAsync($"_history?_tag={tag}&_since=" + sinceUriString);
+
+                    AssertCount(7, allChanges.Resource.Entry);
+
+                    // now choose a value of before that is as close as possible to one of the last updated times
+                    var lastUpdatedTimes = allChanges.Resource.Entry.Select(e => e.Resource.Meta.LastUpdated).OrderBy(d => d.Value);
+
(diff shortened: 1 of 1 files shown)


test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/HistoryTests.cs
```diff
@@ -194,52 +194,78 @@ public async Task GivenAValueForSinceAndBeforeWithModifications_WhenGettingSyste
         [Fact]
         public async Task GivenAValueForSinceAndBeforeCloseToLastModifiedTime_WhenGettingSystemHistory_TheServerShouldNotMissRecords()
         {
-            var tag = Guid.NewGuid().ToString();
+            // as this test is timing dependent, we will try it three times
+            // to see if it can pass
+            for (int i = 0; i < 3; i++)
```
LTA-Thinking, line 199: Could we change 3 to be a variable and use it in the catch statement below? That way if we need to change it there isn't a chance it will be missed in the catch.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/HistoryTests.cs
```diff
@@ -194,52 +194,78 @@ public async Task GivenAValueForSinceAndBeforeWithModifications_WhenGettingSyste
...
-            sinceUriString = beforeUriString;
-            before = DateTime.UtcNow;
-            beforeUriString = HttpUtility.UrlEncode(before.Value.ToString("o"));
-            Thread.Sleep(500); // wait 500 milliseconds to make sure that the value passed to the server for _before is not a time in the future
-            var secondSet = await _client.SearchAsync($"_history?_tag={tag}&_since={sinceUriString}&_before={beforeUriString}");
+                    sinceUriString = beforeUriString;
+                    before = DateTime.UtcNow;
+                    beforeUriString = HttpUtility.UrlEncode(before.Value.ToString("o"));
+                    Thread.Sleep(500); // wait 500 milliseconds to make sure that the value passed to the server for _before is not a time in the future
+                    var secondSet = await _client.SearchAsync($"_history?_tag={tag}&_since={sinceUriString}&_before={beforeUriString}");
```
LTA-Thinking, line 244: There is another Thread.Sleep on line 242.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/HistoryTests.cs
```diff
@@ -194,52 +194,78 @@ public async Task GivenAValueForSinceAndBeforeWithModifications_WhenGettingSyste
...
+                bool success = false;
+                var newResources = new List<Resource>();
 
-            var since = await GetStartTimeForHistoryTest(tag);
+                try
+                {
+                    var tag = Guid.NewGuid().ToString();
 
-            var newResources = new List<Resource>();
+                    var since = await GetStartTimeForHistoryTest(tag);
```
Ivanidzo4ka, line 208: >GetStartTimeForHistoryTest [](http://example.com/codeflow?start=38&length=26)

can you do this instead?
(not in this line but after all resource created)
newResources.Select(x=>x.Meta.LastUpdated.Value).Min()

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/HistoryTests.cs
```diff
@@ -194,52 +194,78 @@ public async Task GivenAValueForSinceAndBeforeWithModifications_WhenGettingSyste
...
-            newResources.Add(await CreateResourceWithTag(Samples.GetDefaultPatient().ToPoco(), tag));
-            newResources.Add(await CreateResourceWithTag(Samples.GetDefaultOrganization().ToPoco(), tag));
-            Thread.Sleep(1000);
-            newResources.Add(await CreateResourceWithTag(Samples.GetJsonSample("BloodGlucose").ToPoco(), tag));
-            newResources.Add(await CreateResourceWithTag(Samples.GetJsonSample("BloodPressure").ToPoco(), tag));
-            newResources.Add(await CreateResourceWithTag(Samples.GetJsonSample("Patient-f001").ToPoco(), tag));
-            newResources.Add(await CreateResourceWithTag(Samples.GetJsonSample("Condition-For-Patient-f001").ToPoco(), tag));
+                    newResources.Add(await CreateResourceWithTag(Samples.GetDefaultPatient().ToPoco(), tag));
+                    newResources.Add(await CreateResourceWithTag(Samples.GetDefaultOrganization().ToPoco(), tag));
+                    await Task.Delay(1000);
```
Ivanidzo4ka, line 216: >await Task.Delay(1000); [](http://example.com/codeflow?start=20&length=23)

I don't understand why this delay is needed

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/HistoryTests.cs
```diff
@@ -194,52 +194,78 @@ public async Task GivenAValueForSinceAndBeforeWithModifications_WhenGettingSyste
...
+                    // now choose a value of before that is as close as possible to one of the last updated times
+                    var lastUpdatedTimes = allChanges.Resource.Entry.Select(e => e.Resource.Meta.LastUpdated).OrderBy(d => d.Value);
 
-            var before = lastUpdatedTimes.ToList()[4];
-            var beforeUriString = HttpUtility.UrlEncode(before.Value.ToString("o"));
-            Thread.Sleep(500);
-            var firstSet = await _client.SearchAsync($"_history?_tag={tag}&_since={sinceUriString}&_before={beforeUriString}");
+                    var before = lastUpdatedTimes.ToList()[4];
+                    var beforeUriString = HttpUtility.UrlEncode(before.Value.ToString("o"));
+                    await Task.Delay(500);
```
Ivanidzo4ka, line 234: >await Task.Delay(500); [](http://example.com/codeflow?start=20&length=22)

I don't understand why this delay is needed?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/HistoryTests.cs
```diff
@@ -194,52 +194,78 @@ public async Task GivenAValueForSinceAndBeforeWithModifications_WhenGettingSyste
...
 
-            sinceUriString = beforeUriString;
-            before = DateTime.UtcNow;
-            beforeUriString = HttpUtility.UrlEncode(before.Value.ToString("o"));
-            Thread.Sleep(500); // wait 500 milliseconds to make sure that the value passed to the server for _before is not a time in the future
-            var secondSet = await _client.SearchAsync($"_history?_tag={tag}&_since={sinceUriString}&_before={beforeUriString}");
+                    sinceUriString = beforeUriString;
+                    before = DateTime.UtcNow;
+                    beforeUriString = HttpUtility.UrlEncode(before.Value.ToString("o"));
+                    Thread.Sleep(500); // wait 500 milliseconds to make sure that the value passed to the server for _before is not a time in the future
```
Ivanidzo4ka, line 242: > Thread.Sleep(500); // wait 500 milliseconds to make sure that the value passed to the server for _before is not a time in the future [](http://example.com/codeflow?start=19&length=133)

I think this is a place where GetStartTimeForHistoryTest should be used, you can have more than 500ms time difference between local time and server time.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/HistoryTests.cs
feordin: This delay is to make sure we have a gap between the edits so we can query just the last 4 edits.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/HistoryTests.cs
feordin: Due to the possible time difference, without the delay, the server would complain that the time was in the future.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/HistoryTests.cs
feordin: GetStartTime makes some edits, and then gets a time after the last edit.  Here we want a timestamp which is very close to the current clock time, but after the edits made in the test case.  It is possible that we will have more than 500ms in difference, but since adding the 500ms delay, we have not seen the error that "before" is invalid due to being in the future.

Ivanidzo4ka: I'm still don't understand.

Here is my understanding.
We post resources to a server, one of the machine set lastupdated value to resource itself and it get converted to 
either ResourceSurrugateId or _lastUpdated field.
And it get stored.
As soon as it posted in persisted layer, it doesn't change.

You pick since and before timestamps. Lock them on testing machine, and send to a server, which convert them into parameters for search and filtering.

Neither values in persistent layer would be changed in this 500 ms, neither values you plan to send.

During history search, timeclock on machine which does search shouldn't matter.
There is no additional changes to resources after you create them, so there gonna be no new records.

What this delay try to accomplish? I truly don't understand.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/HistoryTests.cs
feordin: When just running the test locally, I have seen the exact same time stamp in milliseconds on all 7 resources, which is why the gap was added.  But, 1 second may be a bit too long.  I will reduce it to 100ms.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/HistoryTests.cs
feordin: When we send the "before" string value to the server the server will compare the value that is received with the current clock on the server.  If the value received is in the future (from the servers point of view) then it will throw an error.  This 500ms delay tries to avoid that.  There is not guarantee that it will completely avoid the problem, but it seems to be good enough.  That is the purpose of the delay.

Ivanidzo4ka: Ok, I see, there is validation on server side based on parameters.
Can we not have it?
Why we should disallow user to search for data in future, considering how finnicky time works.

And on a side note, I don't see `_before` parameter been part of history search not in STU3([http://hl7.org/fhir/STU3/http.html#history](http://hl7.org/fhir/STU3/http.html#history))
not in R4 [https://www.hl7.org/fhir/http.html#history](https://www.hl7.org/fhir/http.html#history)
not in R5 [http://hl7.org/fhir/2020Sep/http.html#history](http://hl7.org/fhir/2020Sep/http.html#history)

Another side note, fact what we use web app time rather than database time is quite scary, I hope place where we host them is take care of time drift, otherwise you can have web app putting records in yesterday or even further in the past if it's time clock is in not in sync.

feordin: Both good points.  I think you are correct, we could remove the check for the _before value, I don't think it would hurt anything, and we also should check on where we are getting our timestamps.  I added a user story to track them: 87440.

PR #2397, opened 2021-12-13
Adds hash column and pk for stringSearchParam table

rbans96:
## Description
In order to add primary key to StringSearchParam table, added a TextHash column that is computed based on if TextOverflow is null, then hash is calculated for Text and else hash is calculated for TextOverflow column.
We needed to add this column, since TextOverflow is nullable and of max size so can't be part of primary key.

## Related issues
Addresses [#87237](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Olympus/Stories/?workitem=87237)

## Testing
Existing schema upgrade tests continue to pass and tested manually schema upgrade

## FHIR Team Checklist


Ivanidzo4ka:
>     TextHash            NVARCHAR (32)  COLLATE Latin1_General_100_CI_AI_SC NOT NULL,

I was under impression what size of hash is always constant. 
I think NVARCHAR is redundant and CHAR(32) should suffice. 

---
Refers to: src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/24.sql:588 in d718dca. [](commit_id = d718dcaded47affc4f5d148fea44650814ff5811, deletion_comment = False)

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/24.diff.sql
```diff
@@ -0,0 +1,1176 @@
...
+Add Primary key for dbo.StringSearchParam
+**************************************************************/
+IF NOT EXISTS (
+    SELECT * 
+	FROM sys.key_constraints 
+	WHERE name='PK_StringSearchParam' AND type='PK')
+BEGIN
+    EXEC dbo.LogSchemaMigrationProgress 'Adding PK_StringSearchParam'
+	ALTER TABLE dbo.StringSearchParam 
+	ADD CONSTRAINT PK_StringSearchParam PRIMARY KEY NONCLUSTERED (ResourceTypeId, ResourceSurrogateId, SearchParamId, TextHash)
```
Ivanidzo4ka, line 78: >PK_StringSearchParam [](http://example.com/codeflow?start=16&length=20)

let consider following example.

I have Patient record and that patient have two duplicate names in it.
```
"name": [
    {
      "use": "official",
      "family": "Chalmers",
      "given": [
        "Peter",
        "James"
      ]
    },
     {
      "use": "official",
      "family": "Chalmers",
      "given": [
        "Peter",
        "James"
      ]
    },

```

Is it valid FHIR resource? I think it is.

This is what I would have in our db if I insert it :

ResourceTypeId ResourceSurrogateId SearchParamId Text
103 5102022608830401891 1016 Chalmers
103 5102022608830401891 1016 Chalmers


A duplicate!

This records gonna have same hash, and therefore, can be inserted.

So first of all, I don't think script handles such case during conversion, and would just fail.

Second, even if script would work, I think it app throw me an error during insertion. On a valid resource, btw.
No one would like it.

rbans96:
> > ```
> > TextHash            NVARCHAR (32)  COLLATE Latin1_General_100_CI_AI_SC NOT NULL,
> > ```
> 
> I was under impression what size of hash is always constant. I think NVARCHAR is redundant and CHAR(32) should suffice.
> 
> Refers to: src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/24.sql:588 in [d718dca](https://github.com/microsoft/fhir-server/commit/d718dcaded47affc4f5d148fea44650814ff5811). [](commit_id = [d718dca](https://github.com/microsoft/fhir-server/commit/d718dcaded47affc4f5d148fea44650814ff5811), deletion_comment = False)



> Details

Good point. Created a [PR](https://github.com/microsoft/healthcare-shared-components/pull/381) in healthcare-shared-components to add CharColumn so that we can use it in fhir

Ivanidzo4ka:
>     ON dbo.StringSearchParam(ResourceTypeId, SearchParamId, Text, ResourceSurrogateId)

Since TextHash uniquely represent text/TextOverflow column, there is opportunity to use it in our searches, and therefore move text into include part and replace it with texthash

---
Refers to: src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/24.sql:602 in d718dca. [](commit_id = d718dcaded47affc4f5d148fea44650814ff5811, deletion_comment = False)

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/24.diff.sql
rbans96: Thanks for the catch!
So I look into this and figure out that isMin(or isMax) values are set differently though the primary data is similar. 
image

There are two ways to approach this-
1. Include isMin and isMax in the primary key combination
        But it seems we are consider one of the two rows in our searches so the other row would just be eating space
2. Update application code to have same or consistent values for isMin and isMax when we get duplicate objects for search indices and then the respective rowGenerator class, would only insert distinct row into the table.
        This would need additional work and we would need to consider import operation too.

Hybrid approach-
Have isMin and isMax in primary key since they are just bits and this would ensure no errors while insert but we can work as a separate item to update our application code to have consistent values for isMin and isMax.
Thoughts?

rbans96:
> > ```
> > ON dbo.StringSearchParam(ResourceTypeId, SearchParamId, Text, ResourceSurrogateId)
> > ```
> 
> Since TextHash uniquely represent text/TextOverflow column, there is opportunity to use it in our searches, and therefore move text into include part and replace it with texthash
> 
> Refers to: src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/24.sql:602 in [d718dca](https://github.com/microsoft/fhir-server/commit/d718dcaded47affc4f5d148fea44650814ff5811). [](commit_id = [d718dca](https://github.com/microsoft/fhir-server/commit/d718dcaded47affc4f5d148fea44650814ff5811), deletion_comment = False)

I gave it a thought on this. But we can't do 'like' searches using TextHash since it is a hash :) We can only use TextHash for equals operator

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/24.diff.sql
Ivanidzo4ka: 1. IsMIn and IsMax. You can have only 4 combinations of them, FHIR spec doesn't stop me from having as many duplicate name records as I want to.

2. Maybe it would work.

rbans96:
Closing this PR, as a new PR is created to add primary key to StringSearchParam table to merge to primary key feature branch.

PR #2400, opened 2021-12-14
Add 429 retry and logging in BundleHandler 

feordin:
## Description
We sometimes encounter 429 errors when processing a bundle.  If we receive a 429 at the BundleHandler layer, we abort processing of the bundle and skip the remaining resources.  Here we add an additional retry (in addition to the retry present in the data store layer) that will execute one time per resource which encounters a 429.  Also some logging is added to help us identify this occurrence.

## Related issues
Addresses [issue [AB#87196](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/87196)].

## Testing
Unit was updated.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Resources/Bundle/BundleHandlerTests.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Resources/Bundle/BundleHandlerTests.cs
@@ -322,6 +322,53 @@ namespace Microsoft.Health.Fhir.Api.UnitTests.Features.Resources.Bundle
         }
 
         [Fact]
+        public async Task GivenABundle_WhenOneRequestProducesA429_429IsRetriedThenSucceeds()
+        {
+            var bundle = new Hl7.Fhir.Model.Bundle
+            {
+                Type = BundleType.Batch,
+                Entry = new List<EntryComponent>
+                {
+                    new EntryComponent { Request = new RequestComponent { Method = HTTPVerb.GET, Url = "/Patient" } },
+                    new EntryComponent { Request = new RequestComponent { Method = HTTPVerb.GET, Url = "/Patient" } },
+                    new EntryComponent { Request = new RequestComponent { Method = HTTPVerb.GET, Url = "/Patient" } },
+                },
+            };
+
+            int callCount = 0;
+
+            _router.When(r => r.RouteAsync(Arg.Any<RouteContext>()))
+                .Do(info =>
+                {
+                    info.Arg<RouteContext>().Handler = context =>
+                    {
+                        callCount++;
+                        if (callCount == 2)
+                        {
+                            context.Response.StatusCode = StatusCodes.Status429TooManyRequests;
+                        }
+                        else
+                        {
+                            context.Response.StatusCode = StatusCodes.Status200OK;
+                        }
+
+                        return Task.CompletedTask;
+                    };
+                });
+
+            var bundleRequest = new BundleRequest(bundle.ToResourceElement());
+            BundleResponse bundleResponse = await _bundleHandler.Handle(bundleRequest, default);
+
+            Assert.Equal(4, callCount);
+            var bundleResource = bundleResponse.Bundle.ToPoco<Hl7.Fhir.Model.Bundle>();
+            Assert.Equal(3, bundleResource.Entry.Count);
+            foreach (var entry in bundleResource.Entry)
+            {
+                Assert.Equal("200", entry.Response.Status);
+            }
+        }
+
+        [Fact]
         public async Task GivenAConfigurationEntryLimit_WhenExceeded_ThenBundleEntryLimitExceededExceptionShouldBeThrown()
         {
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -391,6 +391,25 @@ private async Task<EntryComponent> ExecuteRequests(Hl7.Fhir.Model.Bundle respons
 
                         await request.Handler.Invoke(httpContext);
 
+                        // we will retry a 429 one time per request in the bundle
+                        if (httpContext.Response.StatusCode == (int)HttpStatusCode.TooManyRequests
+                            && _bundleType == BundleType.Batch)
```
LTA-Thinking, line 396: Why do we care if it is a Batch request? Couldn't we do the same thing for a Transaction request?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -391,6 +391,25 @@ private async Task<EntryComponent> ExecuteRequests(Hl7.Fhir.Model.Bundle respons
...
+                        {
+                            _logger.LogTrace("BundleHandler received 429 message, attempting retry.  HttpVerb:{httpVerb} BundleSize: {_requestCount} entryIndex:{entryIndex}", httpVerb, _requestCount, entryIndex);
+                            int retryDelay = 2;
+                            var retryAfterValues = httpContext.Response.Headers.GetCommaSeparatedValues("Retry-After");
+                            if (retryAfterValues != StringValues.Empty)
+                            {
+                                if (int.TryParse(retryAfterValues[0], out var retryHeaderValue))
```
LTA-Thinking, line 403: Could this be collapsed into the if statement above it so we only need one?

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Resources/Bundle/BundleHandlerTests.cs
```diff
@@ -286,7 +286,7 @@ public async Task GivenABundle_WhenProcessed_CertainResponseHeadersArePropagated
         }
 
         [Fact]
-        public async Task GivenABundle_WhenOneRequestProducesA429_SubsequentRequestAreSkipped()
+        public async Task GivenABundle_WhenOneRequestProducesA429_429IsRetriedThenSubsequentRequestAreSkipped()
```
LTA-Thinking, line 289: Is it worth adding a test that has a retry succeed to show the happy retry path?

LTA-Thinking (approved):

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -391,6 +391,21 @@ private async Task<EntryComponent> ExecuteRequests(Hl7.Fhir.Model.Bundle respons
...
+                        if (httpContext.Response.StatusCode == (int)HttpStatusCode.TooManyRequests)
+                        {
+                            _logger.LogTrace("BundleHandler received 429 message, attempting retry.  HttpVerb:{httpVerb} BundleSize: {_requestCount} entryIndex:{entryIndex}", httpVerb, _requestCount, entryIndex);
+                            int retryDelay = 2;
+                            var retryAfterValues = httpContext.Response.Headers.GetCommaSeparatedValues("Retry-After");
+                            if (retryAfterValues != StringValues.Empty && int.TryParse(retryAfterValues[0], out var retryHeaderValue))
+                            {
+                                retryDelay = retryHeaderValue;
+                            }
+
```
Ivanidzo4ka, line 404: Is it worth to retry in absence of "Retry-After" header?

Feels that delay 2 milliseconds wouldn't do much to solve problem...

feordin: I was worried that 2 seconds might be too long :).  I used 2 seconds as a fall back, because we have similar logic in the CosmosFhirStore code.  I would still like to try, given that we did receive a 429.

Ivanidzo4ka: [https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.delay?view=net-6.0](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.delay?view=net-6.0)

reates a cancellable task that completes after a specified number of *milliseconds*.

It's not a seconds...

feordin: Thank for you for catching it Ivan!  Sorry I missed the "milliseconds" in your previous comment.  I merged this too quickly!  I will put up a new PR to fix it.

PR #2405, opened 2021-12-16
Adds code owners file (draft PR)

rotodd:
## Description

This PR adds a file that outlines the code owners in OSS. Among other things, this aims to direct the current on-call person to the best person to help with a given area of the repo. Please provide feedback on areas I may have missed, or areas where you think (or do not think) you are a feature owner.

## Related issues
Addresses [AB#87087](https://microsofthealth.visualstudio.com/Health/_workitems/edit/87087)

## Testing
N/A (documentation)

## FHIR Team Checklist
N/A (draft PR)

N/A (draft PR)

### DIFF
--- a/docs/CodeOwners.md
+++ b/docs/CodeOwners.md
@@ -1,0 +1,111 @@
+# Code Owners
+
+This file outlines the owners of the FHIR Server features in OSS.
+
+### Portal
+
+rojo
+
+patalada
+
+### Custom Search & reindexing
+
+jaerwin
+
+rotodd
+
+### Patient $everything
+
+rotodd
+
+ivmatan
+
+### Sort
+
+namadabu
+
+### Schema management
+
+ribans
+
+apurvabhale
+
+### Disable History
+
+rotodd
+
+patalada
+
+### SQL perfomance
+
+ivmatan
+
+ribans
+
+jaerwin
+
+### SQL CRUD operations
+
+rotodd
+
+### SQL search
+
+namadabu
+
+### Cosmos DB performance
+
+bkowitz
+
+jaerwin
+
+ivmatan
+
+### Cosmos CRUD operations
+
+bkowitz
+
+jaerwin
+
+### Cosmos search
+
+bkowitz
+
+jaerwin
+
+### Profiles & validation
+
+ivmatan
+
+### Bulk export
+
+namadabu
+rojo
+
+### Chained search
+
+bkowitz
+
+### Conditional delete
+
+bkowitz
+
+### Purge history
+
+bkowitz
+
+### Patch
+
+ivmatan
+
+bkowitz
+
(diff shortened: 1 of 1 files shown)


docs/CodeOwners.md
```diff
@@ -0,0 +1,112 @@
+# Code Owners
+
+This file outlines the owners of the FHIR Server features in OSS.
+
+### Portal
```
LTA-Thinking, line 5: Should we have Portal as a section in OSS?

docs/CodeOwners.md
```diff
@@ -0,0 +1,112 @@
...
+
+### Patient $everything
+
+rotodd (Robin Todd)
+
+ivmatan (Ivan Matantsev)
+
+### Sort
+
+namadabu (Narasimhan Madabusi)
```
LTA-Thinking, line 25: Probably shouldn't be on the list anymore.

docs/CodeOwners.md
```diff
@@ -0,0 +1,112 @@
...
+### Bulk import
+
+tongwu (Tong Wu)
+
+peizhou (Peiyuan Zhou)
+
+### Resource change capture
+
+teli (Teng Li)
+
```
LTA-Thinking, line 112: Should we have sections for the non-FHIR server parts of our Github repo?
Like for the build, release, samples, and tools folders?

docs/CodeOwners.md
rotodd: Good point - removed in 4eac110.

docs/CodeOwners.md
rotodd: 😢 removed in 4eac110.

CaitlinV39 (commented):
I had a couple of quick questions :)

1. Is Sort owner the same regardless of SQL vs. Cosmos DB?
2. If the search issue applies to both SQL and Cosmos DB is there one that trumps the other or does it just mean any of the search people can do it?
3. Member-match?
4. We called out conditional delete specifically but not the other conditional operations. Should conditional delete just be part of CRUD?
5. Same question as search, if the CRUD bug happens in both, is there a persistence provider that trumps it or is it everyone? 

Final thought is that I found the ordering a little random. Maybe it makes sense on how the code is stored, but otherwise might make sense to group SQL CRUD/Performance/Search and Cosmos CRUD/Performance/Search in one area and then maybe have all the operations together?

rotodd:
@CaitlinV39 thank you for taking a look at this! (in hindsight, I think you may have been the best author for this file)

Since we are moving this to our internal wiki, I started an email thread to discuss further.

rotodd:
Link to wiki [here](https://microsofthealth.visualstudio.com/Health/_wiki/wikis/Health.wiki/783/Code-Owners)

PR #2411, opened 2022-01-04
Remove duplicate search parameters

rbans96:
## Description
Remove duplicate search indices and added hashcode implementation for search values in parity with equals() implementation 
so that searchIndices.Distinct() method returns appropriate results.

## Related issues
Addresses [#87453](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Olympus/Stories/?workitem=87453)

## Testing
Verified manually and existing tests should continue to pass

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Persistence/ResourceWrapperFactory.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Persistence/ResourceWrapperFactory.cs
@@ -5,6 +5,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using EnsureThat;
 using Microsoft.Health.Core.Features.Context;
 using Microsoft.Health.Core.Features.Security;
@@ -71,6 +72,8 @@ namespace Microsoft.Health.Fhir.Core.Features.Persistence
         {
             RawResource rawResource = _rawResourceFactory.Create(resource, keepMeta);
             IReadOnlyCollection<SearchIndexEntry> searchIndices = _searchIndexer.Extract(resource);
+            searchIndices = searchIndices.Distinct().ToArray();
+
             string searchParamHash = _searchParameterDefinitionManager.GetSearchParameterHashForResourceType(resource.InstanceType);
 
             ExtractMinAndMaxValues(searchIndices);
--- a/src/Microsoft.Health.Fhir.Core/Features/Persistence/ResourceWrapperFactory.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Persistence/ResourceWrapperFactory.cs
@@ -72,7 +72,6 @@ namespace Microsoft.Health.Fhir.Core.Features.Persistence
         {
             RawResource rawResource = _rawResourceFactory.Create(resource, keepMeta);
             IReadOnlyCollection<SearchIndexEntry> searchIndices = _searchIndexer.Extract(resource);
-            searchIndices = searchIndices.Distinct().ToArray();
 
             string searchParamHash = _searchParameterDefinitionManager.GetSearchParameterHashForResourceType(resource.InstanceType);
 
--- a/src/Microsoft.Health.Fhir.Core/Features/Persistence/ResourceWrapperFactory.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Persistence/ResourceWrapperFactory.cs
@@ -5,7 +5,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
 using EnsureThat;
 using Microsoft.Health.Core.Features.Context;
 using Microsoft.Health.Core.Features.Security;
--- a/src/Microsoft.Health.Fhir.Core/Features/Search/SearchIndexEntry.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Search/SearchIndexEntry.cs
@@ -62,14 +62,14 @@ namespace Microsoft.Health.Fhir.Core.Features.Search
 
         public bool Equals(SearchIndexEntry other)
         {
-            if (other == null)
+            if ((object)other == null)
             {
                 return false;
             }
 
             if (other.SearchParameter.Url == SearchParameter.Url &&
                 other.SearchParameter.Code.Equals(SearchParameter.Code, StringComparison.OrdinalIgnoreCase) &&
-                other.Value == Value)
+                other.Value.Equals(Value))
             {
                 return true;
             }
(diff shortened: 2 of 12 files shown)


src/Microsoft.Health.Fhir.Core/Features/Search/SearchIndexEntry.cs
```diff
@@ -62,14 +62,14 @@ public SearchIndexEntry(SearchParameterInfo searchParameter, ISearchValue value)
 
         public bool Equals(SearchIndexEntry other)
         {
-            if (other == null)
+            if ((object)other == null)
```
Ivanidzo4ka, line 65: >object [](http://example.com/codeflow?start=17&length=6)

why cast to object is necessary?

src/Microsoft.Health.Fhir.Core/Features/Persistence/ResourceWrapperFactory.cs
```diff
@@ -71,6 +72,8 @@ public ResourceWrapper Create(ResourceElement resource, bool deleted, bool keepM
         {
             RawResource rawResource = _rawResourceFactory.Create(resource, keepMeta);
             IReadOnlyCollection<SearchIndexEntry> searchIndices = _searchIndexer.Extract(resource);
```
Ivanidzo4ka, line 74: >_searchIndexer.Extract(resource) [](http://example.com/codeflow?start=66&length=32)

would it be worth to put distinct part inside Extract method rather than call it here?

I would assume uniqueness of search indices would be useful in other places as well.

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/CompositeSearchValue.cs
```diff
@@ -65,5 +65,11 @@ public override string ToString()
         {
             return string.Join(" $ ", Components.Select(component => string.Join(", ", component.Select(v => $"({v})"))));
         }
+
+        /// <inheritdoc />
+        public override int GetHashCode()
+        {
+            return Components.GetHashCode();
```
Ivanidzo4ka, line 72: >Components [](http://example.com/codeflow?start=19&length=10)

I'm afraid it will count hashcode not by combination of collection entities, but by memory address of collection.

Ivanidzo4ka: should be HashCode.Combine(Components.Select(x=>x.GetHashCode())

src/Microsoft.Health.Fhir.Core/Features/Persistence/ResourceWrapperFactory.cs
rbans96: Good point, moved distinct to Extract method!

src/Microsoft.Health.Fhir.Core/Features/Search/SearchIndexEntry.cs
rbans96: Yes, not necessary, removed :)

src/Microsoft.Health.Fhir.Core/Features/Persistence/ResourceWrapperFactory.cs
```diff
@@ -5,6 +5,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
```
Ivanidzo4ka, line 8: super nit: you no longer need it ;)

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/StringSearchValue.cs
```diff
@@ -105,5 +105,10 @@ public override string ToString()
         {
             return String.EscapeSearchParameterValue();
         }
+
+        public override int GetHashCode()
+        {
+            return String != null ? String.GetHashCode(StringComparison.OrdinalIgnoreCase) : 0;
```
Ivanidzo4ka, line 111: >String != null ? [](http://example.com/codeflow?start=19&length=16)

nit:
I don't think you need this check.
Constructor of class make sure it's not null, and it's get only property, so it should never be null.

Ivanidzo4ka: Same I believe stands for DateTimeSearchValue.

Ivanidzo4ka (approved):
:shipit:

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/UriSearchValue.cs
```diff
@@ -143,5 +143,10 @@ public override string ToString()
 
             return asString.ToString();
         }
+
+        public override int GetHashCode()
+        {
+            return Uri != null ? Uri.ToString().GetHashCode(StringComparison.Ordinal) : 0;
```
rbans96, line 149: @LTA-Thinking If I understand correctly, I should also use OriginialString here instead ToString()?

LTA-Thinking: Yes, I think so.

rbans96: Ah never mind, this Uri is already of type string so I don't need ToString or OriginalString :) Thanks!

src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/ReferenceSearchValue.cs
```diff
@@ -85,8 +86,8 @@ public bool Equals([AllowNull] ISearchValue other)
 
             return Kind == referenceSearchValueOther.Kind &&
                    BaseUri == referenceSearchValueOther.BaseUri &&
-                   ResourceType.Equals(referenceSearchValueOther.ResourceType, StringComparison.OrdinalIgnoreCase) &&
-                   ResourceId.Equals(referenceSearchValueOther.ResourceId, StringComparison.OrdinalIgnoreCase);
+                   string.Equals(ResourceType, referenceSearchValueOther.ResourceType, StringComparison.OrdinalIgnoreCase) &&
```
feordin, line 89: Is it possible for these values to be null? Should we then check for the null case as you have done in other types?

rbans96: If I understand your ask correctly, string.Equals will handle null values for e..g if ResourceType is null, then null will be compared with  referenceSearchValueOther.ResourceType for equality and return true/false accordingly

PR #2412, opened 2022-01-05
Add fix for URL search

LTA-Thinking:
## Description
In the Reindex Worker if a search parameter's url is different escaped vs unescaped it can't be fetched when looking for new search parameters.

## Related issues
Addresses Bug [AB#87538](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/87538): Unable to update search parameters

## Testing
Manual testing

## FHIR Team Checklist

Patch

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/SearchParameterOperations.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/SearchParameterOperations.cs
@@ -219,7 +219,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Parameters
             // parameter will not cause conflicts
             foreach (var searchParam in updatedSearchParameterStatus.Where(p => p.Status == SearchParameterStatus.Deleted))
             {
-                DeleteSearchParameter(searchParam.Uri.ToString());
+                DeleteSearchParameter(searchParam.Uri.OriginalString);
             }
 
             var paramsToAdd = new List<ITypedElement>();
--- a/src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/SearchParameterOperations.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/SearchParameterOperations.cs
@@ -225,7 +225,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Parameters
             var paramsToAdd = new List<ITypedElement>();
             foreach (var searchParam in updatedSearchParameterStatus.Where(p => p.Status != SearchParameterStatus.Deleted))
             {
-                var searchParamResource = await GetSearchParameterByUrl(searchParam.Uri.AbsoluteUri, cancellationToken);
+                var searchParamResource = await GetSearchParameterByUrl(searchParam.Uri.OriginalString, cancellationToken);
 
                 if (searchParamResource == null)
                 {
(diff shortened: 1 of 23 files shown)


feordin (approved):

Ivanidzo4ka:
Any chance you can put small clarification why this works?
We probably have some place where we cast url to absoluteUri during search indices extraction or maybe somewhere else. I'll be curious to find where we doing that.

Main reason why I'm asking, this bug is very easy to repeat in other places, so I would rather have some refactoring for current code to avoid it (like accepting URI instead of string to avoid situation like we have right now).

brendankowitz:
Any way to add a regression test for this?

LTA-Thinking:
@Ivanidzo4ka 
Looking through our other code what we have been doing is using the string directly from the ITypeElement resource, which treats uris as strings. So the reason nothing is escaped is because we never tell our code to treat it like a uri (for which the escape patters are specific for). It is only when we fetch it from the database that we tell our code that this is a uri and should follow the escape patters for one. I've changed it to use the OriginalString field instead of AbsoluteUrl because OriginalString is what would be stored in the database and used elsewhere, while AbsoluteUri only got us halfway there.

Ivanidzo4ka:
>                 DeleteSearchParameter(searchParam.Uri.ToString());

I wonder why this one is untouched...

---
Refers to: src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/SearchParameterOperations.cs:222 in e525919. [](commit_id = e52591987a5325cdf911a13e220091de5681023e, deletion_comment = False)

Ivanidzo4ka:
>                 _searchParameterDefinitionManager.DeleteSearchParameter(url, false);

This thing internally cast url to Uri immediately. I would update signature to Uri instead of string....

---
Refers to: src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/SearchParameterOperations.cs:263 in e525919. [](commit_id = e52591987a5325cdf911a13e220091de5681023e, deletion_comment = False)

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/SearchParameterOperations.cs
```diff
@@ -238,7 +238,7 @@ public async Task GetAndApplySearchParameterUpdates(CancellationToken cancellati
                 if (_searchParameterDefinitionManager.TryGetSearchParameter(searchParam.Uri, out var existingSearchParam))
                 {
                     // if the search parameter exists we should delete the old information currently stored
-                    DeleteSearchParameter(searchParam.Uri.ToString());
```
Ivanidzo4ka, line 241: >Uri.ToString() [](http://example.com/codeflow?start=54&length=14)

I've count 4 more non-tests places where we cast it to string.
Wonder what should we do in that places...

Ivanidzo4ka: SearchParameterStatusV1RowGenerator
SqlServerSearchParameterStatusDataStore
SearchParameterStatusCollection
CosmosDbSearchParameterStatusInitializer

LTA-Thinking: This is an unfortunate case where the programmer's default (ToString) produces unexpected behavior. The places you've called out should use OriginalString as well since that is what happens when we use ITypeElement. I'll update them.

LTA-Thinking:
> Any way to add a regression test for this?

I added a space to the url used in the test search parameter, but there isn't a great way to regression test this. Because this relies on a multi-instance deployment and having the test run on the instance that the search parameter wasn't posted to. By my knowledge we don't have a great way to test instance specific issues.

PR #2430, opened 2022-01-12
Rewrite CompartmentSearch to main search index in Cosmos

brendankowitz:
## Description
Rewrite CompartmentSearch to main search index in Cosmos.

Example of expanded rewritten expressions:
image

Rewriter uses type filtering information to only include the relevant search params.

## Related issues
Addresses [AB#87577](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/87577).

## Testing
- Existing Unit and E2E tests
- Manual test cases

## FHIR Team Checklist

Patch (bug)

### DIFF
--- a/src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
+++ b/src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
@@ -40,6 +40,7 @@ namespace Microsoft.Health.Fhir.CosmosDb.Features.Search
         private readonly CosmosDataStoreConfiguration _cosmosConfig;
         private readonly ICosmosDbCollectionPhysicalPartitionInfo _physicalPartitionInfo;
         private readonly QueryPartitionStatisticsCache _queryPartitionStatisticsCache;
+        private readonly Lazy<IReadOnlyCollection<IExpressionVisitorWithInitialContext<object, Expression>>> _expressionRewriters;
         private readonly ILogger<FhirCosmosSearchService> _logger;
         private readonly SearchParameterInfo _resourceTypeSearchParameter;
         private readonly SearchParameterInfo _resourceIdSearchParameter;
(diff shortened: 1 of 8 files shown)


src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
```diff
@@ -73,6 +76,12 @@ internal class FhirCosmosSearchService : SearchService
             _logger = logger;
             _resourceTypeSearchParameter = SearchParameterInfo.ResourceTypeSearchParameter;
             _resourceIdSearchParameter = new SearchParameterInfo(SearchParameterNames.Id, SearchParameterNames.Id);
+
+            _expressionRewriters = new Lazy<IReadOnlyCollection<IExpressionVisitorWithInitialContext<object, Expression>>>(() => expressionRewriters
+                .OrderBy(x => x.Order)
+                .Cast<IExpressionVisitorWithInitialContext<object, Expression>>()
+                .Append(DateTimeEqualityRewriter.Instance)
```
feordin, line 83: Do we have tests which combine compartment search with DateTime comparison? Is that even possible?

rotodd:
I noticed we are still storing the compartment indices in the Cosmos DB resources - is this intentional?

brendankowitz:
> I noticed we are still storing the compartment indices in the Cosmos DB resources - is this intentional?

Good question, happy to null them out first if it feels like the right thing to do. Since this doesn't apply to SQL yet, it might make sense to remove all this code with that as a second part. Also, if we were forced to roll-back for any reason, I'd rather still have them there :)

rotodd (approved):

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
brendankowitz: There shouldn't be any restriction on multiple rewriters being used. I thought I there was a case in the Patient Everything tests, but I wasn't able to get both rewriters to fire

PR #2436, opened 2022-01-17
Case insensitive distinct for token text import

tongwu-msft:
## Description
Case insensitive distinct for token text import.
We are applying distinct while extracting SearchIndexes in TypedElementIndexer.cs in this [PR](https://github.com/microsoft/fhir-server/pull/2411/files#:~:text=return%20entries.Distinct().ToList()%3B). For the following resource we will still get duplicate rows for TokenText table as the codes are different 

`{
  "status": "final",
    "code": {
    "coding": [
        {
            "code": "29463-7",
            "display": "Body Weight"
        },
        {
            "code": "29463-8",
            "display": "BODY Weight"
        },
        {
            "code": "29463-9",
            "display": "BODY WEIGHT"
        }
    ]
}
}`

While making an API call, after the extract logic we will go through the UpserResource SP, which has the DISTINCT before inserting rows in TokenText table resulting in the unique entry. (For SQL Text column is a case insensitive) Since Inport doesn't go through the SP we get duplicate rows. The RemoveDuplicate added in this [PR ](https://github.com/microsoft/fhir-server/pull/2408/files) is not enough to get the distinct records as it is case sensitive.

## Related issues
Addresses [issue #].

## Testing
Add unit tests for case insensitive distinct.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/DataGenerator/TokenTextSearchParamsTableBulkCopyDataGenerator.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/DataGenerator/TokenTextSearchParamsTableBulkCopyDataGenerator.cs
@@ -3,8 +3,10 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
+using System;
 using System.Collections.Generic;
 using System.Data;
+using System.Linq;
 using EnsureThat;
 using Microsoft.Health.Fhir.Core.Features.Persistence;
 using Microsoft.Health.Fhir.SqlServer.Features.Schema.Model;
@@ -27,6 +29,8 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Operations.Import.DataGenerat
             _searchParamGenerator = searchParamGenerator;
         }
 
+        internal static BulkTokenTextTableTypeV1RowComparer Comparer { get; } = new BulkTokenTextTableTypeV1RowComparer();
+
         internal override string TableName
         {
             get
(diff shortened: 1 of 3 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/DataGenerator/TokenTextSearchParamsTableBulkCopyDataGenerator.cs
```diff
@@ -42,7 +46,7 @@ internal override void FillDataTable(DataTable table, SqlBulkCopyDataWrapper inp
 
             IEnumerable<BulkTokenTextTableTypeV1Row> searchParams = _searchParamGenerator.GenerateRows(new ResourceWrapper[] { input.Resource });
 
-            foreach (BulkTokenTextTableTypeV1Row searchParam in searchParams)
+            foreach (BulkTokenTextTableTypeV1Row searchParam in Distinct(searchParams))
```
Ivanidzo4ka, line 49: >Distinct [](http://example.com/codeflow?start=64&length=8)

I doubt this problem unique for text only.

Can we change `BulkSearchParameterRowGenerator` to Genereate unique rows and do distinct on ISearchValue objects instead of doing it here?

Ivanidzo4ka: I also don't like how responsibility of knowing what is distinct means leaks into here. We already define GetHashCode and Equals to `ISearchValues` we shouldn't define it again here.

Ivanidzo4ka: One more thing...
Richa recently merge code which make sure `ISearchIndexer` extracts unique values for resource (before it could return duplicates). Does problem still exist with new changes? If yes, can you point to place where we actually get duplicate records? 
Uniqueness of data should be solved once and on bottom level rather than solving it on higher levels in multiple places.

tongwu-msft: @Ivanidzo4ka good idea, agree that we should distinct at central place. @apurvabhaleMS do you have any idea on this?

tongwu-msft: For ISearchIndexer distinct the problem is tokentext contains code and text, and only text in tokentext table. That's possible text is same but code is different for search params.

apurvabhaleMS (approved):

PR #2446, opened 2022-01-20
Added Primary Keys to FHIR tables

apurvabhaleMS:
## Description
This PR serves the following purposes -
A.  Adds TrustServerCertificate=True to local DB connection string as it si required by the latest helath_shared package versions. This should fix the failing unit tests locally. Files - [SqlServerFhirStorageTestsFixture.cs](test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlServerFhirStorageTestsFixture.cs) and SqlServerSchemaUpgradeTests.cs (test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlServerSchemaUpgradeTests.cs)

B. Adds null value entry to the following tables with index 0-
We have SystemId and QuantityCodeId referenced in a few other tables and they need to be a part of PK. But it's a nullable column so we have to make it Not NULL to include in PK. To solve for this we decided to insert a row in the below tables to map to null values in the respective tables.
- dbo.System
- dbo.QuantityCode

C. Includes QuantityId to the PK of dbo.QuanittySearchParam table to solve PK violation error
In the previous iteration of[ PK changes](https://github.com/microsoft/fhir-server/pull/2376/files), we added PK to this table.
We are getting PK violation error when there are multiple valueQuantities with the same "value" and different QuanittyCode.

To fix this we have to -
           1. Make changes stated in pt. B above
           2. Remove duplicates
           3. Back-fill the column
           4. Drop PK and indexes
           5. Make QuanityCodeId column NOT NULL
           6. Add default constraint for QuantityCodeId
           7. Recreate Indexes and PK
           8. Update the row generator to ensure for future inserts instead of Null value 0 is inserted for SystemId and QuantityCodeId

Json used to duplicate the issue - 
`{
  "status": "final",
      "code": {
        "coding": [
            {
                "code": "29463-7",
                "display": "Body Weight"
            }
        ]
    },
  "component": [
    {
      "code": {
        "coding": [
          {
            "system": "http://loinc.org",
            "code": "20564-1",
            "display": "Oxygen saturation in blood"
          }
        ]
      },
      "valueQuantity": {
        "value": 20,
        "unit": "%",
        "system": "http://unitsofmeasure.org",
        "code": "%"
      }
    },
    {
      "code": {
        "coding": [
          {
            "system": "http://loinc.org",
            "code": "8310-5",
            "display": "Body temperature"
          }
        ]
      },
      "valueQuantity": {
        "value": 20,
        "unit": "C",
        "system": "http://unitsofmeasure.org",
        "code": "Cel"
      }
    }
  ]
}`

 The below image shows how the previous PK creates a problem as the rows only differ by QuantityCodeId
image

D. Adds the primary keys to the existing FHIR tables. Also, remove duplicate records.
- dbo.TokenText
- dbo.DateTimeSearchParam

## Related issues
Addresses [issue #].

## Testing

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/26.diff.sql
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/26.diff.sql
@@ -8,19 +8,19 @@ GO
 /******************************************************************
   Table - dbo.TokenText table
 *******************************************************************/
+
 -- Deleting duplicate rows based on all columns
 EXEC dbo.LogSchemaMigrationProgress 'Deleting redundant rows from dbo.TokenText'
 GO
-
-WITH cte AS (
+;WITH CTE AS (
     SELECT ResourceTypeId, ResourceSurrogateId, SearchParamId, Text, IsHistory, ROW_NUMBER() 
     OVER (
 		PARTITION BY ResourceTypeId, ResourceSurrogateId, SearchParamId, Text, IsHistory
-		ORDER BY ResourceTypeId, ResourceSurrogateId, SearchParamId, Text, IsHistory
+		ORDER BY ResourceTypeId, ResourceSurrogateId, SearchParamId, Text
 	) row_num
 	FROM dbo.TokenText
 )
-DELETE FROM cte WHERE row_num > 1
+DELETE FROM CTE WHERE row_num > 1
 GO
 
 -- Create nonclustered primary key on the set of non-nullable columns that makes it unique
@@ -40,6 +40,21 @@ GO
 /******************************************************************
   Table - dbo.DateTimeSearchParam table
 *******************************************************************/
+
+-- Deleting duplicate rows based on all columns
+EXEC dbo.LogSchemaMigrationProgress 'Deleting redundant rows from dbo.DateTimeSearchParam'
+GO
+;WITH CTE AS (
+    SELECT ResourceTypeId, ResourceSurrogateId, SearchParamId, StartDateTime, EndDateTime, IsLongerThanADay, IsHistory, IsMin, IsMax, ROW_NUMBER() 
+    OVER (
+		PARTITION BY ResourceTypeId, ResourceSurrogateId, SearchParamId, StartDateTime, EndDateTime, IsLongerThanADay, IsHistory, IsMin, IsMax
+		ORDER BY ResourceTypeId, ResourceSurrogateId, SearchParamId
+	) row_num
+	FROM dbo.DateTimeSearchParam
+)
+DELETE FROM CTE WHERE row_num > 1
+GO
+
 -- Create nonclustered primary key on the set of non-nullable columns that makes it unique
 IF NOT EXISTS (
     SELECT * 
@@ -87,6 +102,21 @@ GO
 /******************************************************************
   Table - dbo.QuantitySearchParam table
 *******************************************************************/
+
+-- Deleting duplicate rows based on all columns
+EXEC dbo.LogSchemaMigrationProgress 'Deleting redundant rows from dbo.QuantitySearchParam'
+GO
+;WITH CTE AS (
+    SELECT ResourceTypeId, ResourceSurrogateId, SearchParamId, SystemId, QuantityCodeId, SingleValue, LowValue, HighValue, IsHistory, ROW_NUMBER() 
+    OVER (
+		PARTITION BY ResourceTypeId, ResourceSurrogateId, SearchParamId, SystemId, QuantityCodeId, SingleValue, LowValue, HighValue, IsHistory
+		ORDER BY ResourceTypeId, ResourceSurrogateId, SearchParamId, SystemId, QuantityCodeId, SingleValue, LowValue, HighValue
+	) row_num
+	FROM dbo.QuantitySearchParam
+)
+DELETE FROM CTE WHERE row_num > 1
+GO
+
 -- Backfill table dbo.QuantitySearchParam with non-null QuantityCodeId value
 EXEC dbo.LogSchemaMigrationProgress 'Back-fill table dbo.QuantitySearchParam'
 UPDATE dbo.QuantitySearchParam
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/26.diff.sql
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/26.diff.sql
@@ -10,7 +10,7 @@ GO
 *******************************************************************/
 
 -- Deleting duplicate rows based on all columns
-EXEC dbo.LogSchemaMigrationProgress 'Deleting redundant rows from dbo.TokenText'
+EXEC dbo.LogSchemaMigrationProgress 'Deleting redundant rows from dbo.TokenText';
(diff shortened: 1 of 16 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SearchParameterQueryGenerator.cs
```diff
@@ -258,7 +258,22 @@ protected static SearchParameterQueryGeneratorContext VisitMissingFieldImpl(Miss
                 throw new InvalidOperationException($"Unexpected missing field {expression.FieldName}");
             }
 
-            AppendColumnName(context, column, expression).Append(" IS NULL");
+            switch (expression.FieldName)
+            {
+                case FieldName.TokenSystem:
+                case FieldName.QuantityCode:
+                    context.StringBuilder.Append(" ( ");
+                    AppendColumnName(context, column, expression).Append(" = 0");
```
Ivanidzo4ka, line 266: >0 [](http://example.com/codeflow?start=77&length=1)

nit:
I would suggest to use `NullSystemId` constant here.
I would also suggest to collapse `NullSystemId` and `NullQuantityCodeId` into one variable.

Ivanidzo4ka:
>     internal class BulkQuantitySearchParameterV1RowGenerator : BulkSearchParameterRowGenerator<QuantitySearchValue, BulkQuantitySearchParamTableTypeV1Row>

What about `BulkTokenSearchParameterV1RowGenerator`?

---
Refers to: src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/BulkQuantitySearchParameterV1RowGenerator.cs:12 in 9d0b423. [](commit_id = 9d0b423d445d6b9a13bcdd482f3d918388256d44, deletion_comment = False)

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Tables/DateTimeSearchParam.sql
```diff
@@ -8,7 +8,10 @@
     IsLongerThanADay bit NOT NULL,
     IsHistory bit NOT NULL,
     IsMin bit CONSTRAINT date_IsMin_Constraint DEFAULT 0 NOT NULL,
-    IsMax bit CONSTRAINT date_IsMax_Constraint DEFAULT 0 NOT NULL
+    IsMax bit CONSTRAINT date_IsMax_Constraint DEFAULT 0 NOT NULL,
+    CONSTRAINT PK_DateTimeSearchParam PRIMARY KEY NONCLUSTERED(ResourceTypeId, ResourceSurrogateId, SearchParamId, StartDateTime, EndDateTime)
+    WITH (DATA_COMPRESSION = PAGE) 
+	ON PartitionScheme_ResourceTypeId(ResourceTypeId)
```
Ivanidzo4ka, line 14: >PartitionScheme_ResourceTypeId [](http://example.com/codeflow?start=4&length=30)

Unrelated to content of your PR, but can you enlighten me  about purpose of this file?

We have currentschema.sql, is these files are purely to make reading easier?

apurvabhaleMS: You are right. All the files under /Features/Schema/Sql are used to generate a full schema script file automatically. We don't need to add the currentschema.sql file anymore it gets auto-generated and we still continue to manually add diff.sql file

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SearchParameterQueryGenerator.cs
apurvabhaleMS: Good point I will update this.

apurvabhaleMS:
> What about `BulkTokenSearchParameterV1RowGenerator`?

This PR only contains the SystemId and QuanittyCodeId changes for table QuantitySearchParam. I have another PR coming which contains changes for the other generators.

Ivanidzo4ka:
I would prefer to have them under one PR.
If you think it's to much of a work or PR would have to much files, I can suggest to create feature branch something like `/features/PK_SearchParams_Tables` and make this and future PRs towards that branch, and later have cumulative PR to merge it all together to ~~master~~ main branch.

Main reason why I would prefer it this way: I don't like state of the system where it's half way works one way, and half way in another. Systems like that are really hard to support.

---
In reply to: [1017904868](https://github.com/microsoft/fhir-server/pull/2446#issuecomment-1017904868) [](http://example.com/codeflow?ancestors=1017904868,1017904868)

apurvabhaleMS:
> I would prefer to have them under one PR.

Let's do that! Created a feature branch!

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/26.diff.sql
```diff
@@ -0,0 +1,281 @@
...
+EXEC dbo.LogSchemaMigrationProgress 'Insert an entry with id 0 to dbo.QuantityCode table'
+IF NOT EXISTS (SELECT 1 FROM dbo.QuantityCode WHERE QuantityCodeId = 0 AND value = '')
+BEGIN
+    SET IDENTITY_INSERT dbo.QuantityCode ON;
+
+    Insert INTO dbo.QuantityCode (QuantityCodeId, Value)
+    Values (0, '')
+
+    SET IDENTITY_INSERT dbo.QuantityCode OFF;
+END;
```
rbans96, line 84: nit: I see semicolon(;) after END at few places and not at few places. Should we add (;) after END to be consistent?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/26.diff.sql
```diff
@@ -0,0 +1,281 @@
...
+
+-- Backfill table dbo.QuantitySearchParam with non-null QuantityCodeId value
+EXEC dbo.LogSchemaMigrationProgress 'Back-fill table dbo.QuantitySearchParam'
+UPDATE dbo.QuantitySearchParam
+SET QuantityCodeId = 0 
+WHERE QuantityCodeId IS NULL;
+GO
+
+-- Backfill table dbo.QuantitySearchParam with non-null SystemId value
+EXEC dbo.LogSchemaMigrationProgress 'Back-fill table dbo.QuantitySearchParam'
```
rbans96, line 128: nit: can we add different logging statement than above
maybe EXEC dbo.LogSchemaMigrationProgress 'Back-fill column SystemId into the table dbo.QuantitySearchParam'

rbans96 (approved):

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/26.diff.sql
apurvabhaleMS: Updated 👍

PR #2447, opened 2022-01-20
add import metric notification

peizhou298:
## Description
This PR used to send import operation metric notification using mediator, it only contains the send part, and receive part will be in PaaS side.

## Related issues
Addresses [issue #].

## Testing
Add several tests in both UT and E2E.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportOrchestratorTask.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportOrchestratorTask.cs
@@ -10,6 +10,7 @@ using System.Net;
 using System.Threading;
 using System.Threading.Tasks;
 using EnsureThat;
+using MediatR;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Primitives;
 using Microsoft.Health.Core.Features.Context;
@@ -27,6 +28,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Import
         private const int DefaultPollingFrequencyInSeconds = 3;
         private const long DefaultResourceSizePerByte = 64;
 
+        private readonly IMediator _mediator;
         private ImportOrchestratorTaskInputData _orchestratorInputData;
         private RequestContextAccessor<IFhirRequestContext> _contextAccessor;
         private ImportOrchestratorTaskContext _orchestratorTaskContext;
@@ -40,6 +42,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Import
         private List<(Uri resourceUri, TaskInfo taskInfo)> _runningTasks = new List<(Uri resourceUri, TaskInfo taskInfo)>();
 
         public ImportOrchestratorTask(
+            IMediator mediator,
             ImportOrchestratorTaskInputData orchestratorInputData,
             ImportOrchestratorTaskContext orchestratorTaskContext,
             ITaskManager taskManager,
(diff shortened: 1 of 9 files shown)


src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportTaskMetricsNotification.cs
```diff
@@ -0,0 +1,49 @@
...
+            Id = id;
+            Status = status;
+            CreatedTime = createdTime;
+            EndTime = endTime;
+            DataSize = dataSize;
+            SucceedCount = succeedCount;
+            FailedCount = failedCount;
+        }
+
+        public string FhirOperation { get; }
```
Ivanidzo4ka, line 31: >FhirOperation [](http://example.com/codeflow?start=22&length=13)

Should we set it to `Import`?

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportTaskMetricsNotification.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        public string ResourceType { get; }
```
Ivanidzo4ka, line 33: >ResourceType [](http://example.com/codeflow?start=22&length=12)

Since it's coming from interface, i would assume this is something we would like to populate.

Since import works for all resources, I would suggest to put `Resource` here.

Can be wrong, and it's totally ok to ignore this field.

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportOrchestratorTask.cs
```diff
@@ -267,6 +278,23 @@ private async Task ValidateResourcesAsync(CancellationToken cancellationToken)
...
+        private async Task SendImportMetricsNotification(TaskResult taskResult)
+        {
+            long? succeedCount = _orchestratorTaskContext.ImportResult?.Output?.Sum(output => output.Count);
+            long? failedCount = _orchestratorTaskContext.ImportResult?.Error?.Sum(error => error.Count);
+
+            ImportTaskMetricsNotification importTaskMetricsNotification = new ImportTaskMetricsNotification(
+                _orchestratorInputData.TaskId,
+                taskResult.ToString(),
+                _orchestratorInputData.TaskCreateTime,
+                DateTimeOffset.Now.ToUniversalTime(),
```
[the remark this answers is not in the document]
Ivanidzo4ka: considering how it's used, would it make sense to push this into constructor instead of field we expose?

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportTaskMetricsNotification.cs
peizhou298: yes, it should be set to import ,already set it in constructor.

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportOrchestratorTask.cs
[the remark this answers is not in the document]
peizhou298: Use Clock.UtcNow instead. This field means the end time of the import operation, so may be put it in constructor is not a good choice.

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportTaskMetricsNotification.cs
peizhou298: Import operation may contain many different types of resource, so ignore it maybe a good choice. Anyway, set it to null in the constructor.

Ivanidzo4ka (approved):
:shipit:

Issue #2037, opened 2021-06-18
Resource returned multiple times if it meets both base criteria and _include criteria


moiradillon12:
**Describe the bug**
When running a search that has _include that meets the base criteria and also the include criteria, a single resource is included twice. Example: search for Locations and include partOF, which references a location, the location resource is returned twice. 

**FHIR Version?**
R4

**Data provider?**
CosmosDB

**To Reproduce**
1. Post a Location resource which references another Location resource with partOf.
2. Query : {{fhirurl}}/Location?_include=Location/partOf
3. Same Location resource will be returned twice 

**Expected behavior**
We expected a resource to only be returned once, even if it matches multiple criteria.

**Actual behavior**
Resource meeting multiple criteria is returned twice.

CaitlinV39:
AB#84439

CaitlinV39:
From Zulip on correct behavior:

I don't know of a place that it is specified cleanly, but the Constraints on bundle include a combination of rules that I believe force it:

fullUrl must be unique
fullUrl cannot be version specific
In the comments for Bundle.entry.search.mode, there is the line:

There is only one mode. In some corner cases, a resource may be included because it is both a match and an include. In these circumstances, 'match' takes precedence.

So, while well-hidden, match takes priority.

mikaelweave:
Fixed in #2448

PR #2448, opened 2022-01-20
Matched resources should not also be _included

brendankowitz:
## Description
Adds a filter so that if a resource is in the "matched" set of results, it should not also be there as an "include".

Limitation: this is only for the current page, it doesn't guarantee uniqueness across pages.

## Related issues
Addresses #2037.

## Testing
Adds E2E tests

## FHIR Team Checklist

Patch (bug)

### DIFF
--- a/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/IncludeSearchTests.cs
+++ b/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/IncludeSearchTests.cs
@@ -490,6 +490,25 @@ namespace Microsoft.Health.Fhir.Tests.E2E.Rest.Search
             ValidateBundleUrl(Client.HttpClient.BaseAddress, ResourceType.Patient, query, bundle.Link[0].Url);
         }
 
+        [InlineData("_include")]
+        [InlineData("_revinclude")]
+        [Theory]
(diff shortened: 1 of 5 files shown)


test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/IncludeSearchTests.cs
```diff
@@ -490,6 +490,25 @@ public async Task GivenARevIncludeSearchExpressionWithMultipleIncludes_WhenSearc
             ValidateBundleUrl(Client.HttpClient.BaseAddress, ResourceType.Patient, query, bundle.Link[0].Url);
         }
 
+        [InlineData("_include")]
+        [InlineData("_revinclude")]
+        [Theory]
+        public async Task GivenAnIncludeSearchExpressionWithLocationLinkedToItself_WhenSearched_ThenCorrectBundleShouldBeReturned(string includeType)
```
LTA-Thinking, line 496: Would a test showing that this also works for cyclical references have value? IE: resource 1 references resource 2 and resource 2 references resource 1.

LTA-Thinking (approved):

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/IncludeSearchTests.cs
brendankowitz: Good idea, I think this can only be done with SQL (:iterate)

PR #2453, opened 2022-01-25
Adds primary key for stringSearchParam table

rbans96:
## Description
Adds primary key to StringSearchParam table by adding TextHash column which is not-null.
The logic to add hash of TextOverflow column if its not null otherwise Text column.

This will be merged in primaryKey-part2 feature branch and eventually to main

## Related issues
Addresses [#87736](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Olympus/Stories/?workitem=87736)

## Testing
E2E tests are added

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Tables/StringSearchParam.sql
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Tables/StringSearchParam.sql
@@ -9,7 +9,7 @@ CREATE TABLE dbo.StringSearchParam
     IsHistory bit NOT NULL,
     IsMin bit CONSTRAINT string_IsMin_Constraint DEFAULT 0 NOT NULL,
     IsMax bit CONSTRAINT string_IsMax_Constraint DEFAULT 0 NOT NULL,
-    TextHash varchar(64) COLLATE Latin1_General_100_CI_AI_SC NOT NULL,
+    TextHash binary(32) NOT NULL,
     CONSTRAINT PK_StringSearchParam PRIMARY KEY NONCLUSTERED (ResourceTypeId, ResourceSurrogateId, SearchParamId, TextHash)
     WITH (DATA_COMPRESSION = PAGE)
     ON PartitionScheme_ResourceTypeId(ResourceTypeId),
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Tables/StringSearchParam.sql
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Tables/StringSearchParam.sql
@@ -9,7 +9,7 @@ CREATE TABLE dbo.StringSearchParam
     IsHistory bit NOT NULL,
     IsMin bit CONSTRAINT string_IsMin_Constraint DEFAULT 0 NOT NULL,
     IsMax bit CONSTRAINT string_IsMax_Constraint DEFAULT 0 NOT NULL,
-    TextHash nvarchar(32) COLLATE Latin1_General_100_CI_AI_SC NOT NULL,
+    TextHash varchar(64) COLLATE Latin1_General_100_CI_AI_SC NOT NULL,
     CONSTRAINT PK_StringSearchParam PRIMARY KEY NONCLUSTERED (ResourceTypeId, ResourceSurrogateId, SearchParamId, TextHash)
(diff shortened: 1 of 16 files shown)


apurvabhaleMS (approved):

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Tables/StringSearchParam.sql
```diff
@@ -7,7 +8,11 @@
     TextOverflow nvarchar(max) COLLATE Latin1_General_100_CI_AI_SC NULL,
     IsHistory bit NOT NULL,
     IsMin bit CONSTRAINT string_IsMin_Constraint DEFAULT 0 NOT NULL,
-    IsMax bit CONSTRAINT string_IsMax_Constraint DEFAULT 0 NOT NULL
+    IsMax bit CONSTRAINT string_IsMax_Constraint DEFAULT 0 NOT NULL,
+    TextHash varchar(64) COLLATE Latin1_General_100_CI_AI_SC NOT NULL,
```
Ivanidzo4ka, line 12: >varchar [](http://example.com/codeflow?start=13&length=7)

Since it has constant length, would it make sense to make it char(64).

Ideally I would prefer to have binary(32) to remove any string allocations and just pass byte array back and forth.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/BulkStringSearchParameterV3RowGenerator.cs
```diff
@@ -0,0 +1,50 @@
...
+                overflow = searchValue.String;
+            }
+            else
+            {
+                indexedPrefix = searchValue.String;
+                overflow = null;
+            }
+
+            string computedHash;
+            using (SHA256 sha256Hash = SHA256.Create())
```
Ivanidzo4ka, line 40: >SHA256.Create() [](http://example.com/codeflow?start=39&length=15)

I wonder, how heavy it is to create this object on each row.

Ideally we should create it once, but it's probably mean we need to make generator disposable, and utilize it, which is a lot of work...

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Tables/StringSearchParam.sql
rbans96: Good point, updated.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/BulkStringSearchParameterV3RowGenerator.cs
rbans96: Thanks, updated to implement IDisposable

Ivanidzo4ka: Unfortunately it's a bit trickier than just adding `IDisposable` to a class. Some one need to call it.

For this particular case, in this particular time, it will work fine, since we have only one instance of it, because we inject it with DI as singleton
see `AddSqlServerTableRowParameterGenerators`
so it would be freed up during app closure.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/BulkStringSearchParameterV3RowGenerator.cs
```diff
@@ -0,0 +1,53 @@
...
+                overflow = searchValue.String;
+            }
+            else
+            {
+                indexedPrefix = searchValue.String;
+                overflow = null;
+            }
+
+            var hashText = overflow != null ? overflow : indexedPrefix;
+            byte[] computedHash = _sha256.ComputeHash(Encoding.Unicode.GetBytes(hashText));
```
Ivanidzo4ka, line 41: >byte[] computedHash = _sha256.ComputeHash(Encoding.Unicode.GetBytes(hashText)); [](http://example.com/codeflow?start=12&length=79)

Aren't it pretty? :)

Ivanidzo4ka (approved):
:shipit:

PR #2457, opened 2022-01-27
Enables PUT creates on versioned update

rotodd:
## Description

The `versioned-update` [versioning policy](https://build.fhir.org/codesystem-versioning-policy.html) requires that requests to update a resource specify the most recent version of the resource in the request's `if-match` header.

Previously, PUT creates would result in an error when the versioning policy was set to `versioned-update`. This happened because the server would check for a matching version in the `if-match` header even though no matching version exists.

This PR implements the ability to create a new resource with a PUT request without an `if-match` header when the versioning policy is configured to `versioned-update`.

## Related issues
Addresses [AB#87898](https://microsofthealth.visualstudio.com/Health/_workitems/edit/87898).

## Testing
Adds integration tests and a manual test flow.

## FHIR Team Checklist
- ✅ CI is green before merge

None (bug fix)

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Resources/ResourceHandlerTests.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Resources/ResourceHandlerTests.cs
@@ -208,26 +208,6 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Resources
         }
 
         [Fact]
-        public async Task GivenAFhirMediator_WhenSavingAResourceWithoutETagAndETagIsRequired_ThenPreconditionFailExceptionIsThrown()
-        {
-            _conformanceStatement.Rest.First().Resource.Find(x => x.Type == ResourceType.Patient).UpdateCreate = false;
-
-            var resource = Samples.GetDefaultPatient();
-
-            await Assert.ThrowsAsync<PreconditionFailedException>(async () => await _mediator.UpsertResourceAsync(resource, null));
-        }
-
-        [Fact]
-        public async Task GivenAFhirMediator_WhenSavingAResourceWithoutETagAndETagIsRequiredWithUpdateCreate_ThenPreconditionFailExceptionIsThrown()
-        {
-            var resource = Samples.GetDefaultPatient();
-
-            // Documented for completeness, but under this situation arguably this request should succeed.
-            // Versioned-update + UpdateCreate; When no If-Match header is provided, the request should allow create but not update
-            await Assert.ThrowsAsync<PreconditionFailedException>(async () => await _mediator.UpsertResourceAsync(resource, null));
-        }
-
-        [Fact]
         public async Task GivenAFhirMediator_WhenSavingAResourceWithoutETag_ThenPreconditionFailExceptionIsThrown()
         {
             var resource = Samples.GetDefaultObservation();
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Upsert/UpsertResourceHandler.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Upsert/UpsertResourceHandler.cs
@@ -59,28 +59,11 @@ namespace Microsoft.Health.Fhir.Core.Features.Resources.Upsert
 
             ResourceWrapper resourceWrapper = CreateResourceWrapper(resource, deleted: false, keepMeta: allowCreate);
 
-            UpsertOutcome result = await UpsertAsync(request, resourceWrapper, allowCreate, keepHistory, requireETagOnUpdate, cancellationToken);
+            UpsertOutcome result = await FhirDataStore.UpsertAsync(resourceWrapper, request.WeakETag, allowCreate, keepHistory, cancellationToken, requireETagOnUpdate);
 
             resource.VersionId = result.Wrapper.Version;
 
             return new UpsertResourceResponse(new SaveOutcome(new RawResourceElement(result.Wrapper), result.OutcomeType));
-        }
-
-        private async Task<UpsertOutcome> UpsertAsync(UpsertResourceRequest message, ResourceWrapper resourceWrapper, bool allowCreate, bool keepHistory, bool requireETagOnUpdate, CancellationToken cancellationToken)
-        {
-            UpsertOutcome result;
-
-            try
-            {
-                result = await FhirDataStore.UpsertAsync(resourceWrapper, message.WeakETag, allowCreate, keepHistory, cancellationToken, requireETagOnUpdate);
-            }
-            catch (PreconditionFailedException) when (_modelInfoProvider.Version == FhirSpecification.Stu3)
-            {
-                // The backwards compatibility behavior of Stu3 is to return a Conflict instead of Precondition fail
-                throw new ResourceConflictException(message.WeakETag);
-            }
-
-            return result;
         }
     }
 }
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Upsert/UpsertResourceHandler.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Upsert/UpsertResourceHandler.cs
@@ -53,30 +53,26 @@ namespace Microsoft.Health.Fhir.Core.Features.Resources.Upsert
 
             Resource resource = request.Resource.ToPoco<Resource>();
 
-            if (await ConformanceProvider.Value.RequireETag(resource.TypeName, cancellationToken) && request.WeakETag == null)
-            {
-                throw new PreconditionFailedException(string.Format(Core.Resources.IfMatchHeaderRequiredForResource, resource.TypeName));
-            }
-
             bool allowCreate = await ConformanceProvider.Value.CanUpdateCreate(resource.TypeName, cancellationToken);
             bool keepHistory = await ConformanceProvider.Value.CanKeepHistory(resource.TypeName, cancellationToken);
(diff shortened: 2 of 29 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/UpsertResource_6.sql
```diff
@@ -145,7 +148,14 @@ IF (@previousResourceSurrogateId IS NULL)
...
 ELSE
     BEGIN
-	-- There is a previous version
+        -- There is a previous version
+        IF (@requireETagOnUpdate = 1
+            AND @etag IS NULL)
+            BEGIN
+                -- This is a versioned update and no version was specified
+                THROW 50412, 'Precondition failed', 1;
+            END
```
Ivanidzo4ka, line 157: I don't think this is right place to do.
Imagine doing all the heavy lifting of parsing resource, extracting search values,  filling up tables, sending them to SQL server.

Just to fail on this simple condition.

If we need to fail, fail as soon as possible, as soon as we can check this condition and have all available pieces.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -143,7 +152,15 @@ public async Task<UpsertOutcome> UpsertAsync(ResourceWrapper resource, WeakETag
                     switch (e.Number)
                     {
                         case SqlErrorCodes.PreconditionFailed:
-                            throw new PreconditionFailedException(string.Format(Core.Resources.ResourceVersionConflict, weakETag?.VersionId));
```
Ivanidzo4ka, line 146: Is it worth to add new error code for this case?
I'm not sure logic to check etag==null is capturing all cases

rotodd: I was thinking about it... we currently have SQL error codes mapped to HTTP error codes in [SqlErrorCodes.cs](https://github.com/microsoft/healthcare-shared-components/blob/master/src/Microsoft.Health.SqlServer/Features/Storage/SqlErrorCodes.cs) (in the hsc repo).

We would need to change the convention since both scenarios throw a PreconditionFailedException in this case. Thoughts?

Ivanidzo4ka: I haven't touch Upsert code for awhile so, if you think, condition you have covers all the cases and draws clear line, I'm fine with that.
I'm not insisting, I just see something what potentially can be problematic, and ask question : is this enough?

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Upsert/UpsertResourceHandler.cs
```diff
@@ -53,30 +53,26 @@ public async Task<UpsertResourceResponse> Handle(UpsertResourceRequest request,
 
             Resource resource = request.Resource.ToPoco<Resource>();
 
-            if (await ConformanceProvider.Value.RequireETag(resource.TypeName, cancellationToken) && request.WeakETag == null)
-            {
-                throw new PreconditionFailedException(string.Format(Core.Resources.IfMatchHeaderRequiredForResource, resource.TypeName));
-            }
-
```
rotodd, line 60: This check ensures the version is specified in the `if-match` header when the versioning policy is set to `versioned-update`.

In this PR, the check is moved to data store layer where we know if the upsert operation is an update or a create, allowing us to only throw an error if we are updating.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/26.diff.sql
```diff
@@ -0,0 +1,478 @@
+﻿/*************************************************************
+    This migration passes in a new parameter to the upsert
+    stored procedure. The parameter tracks whether or not a
+    version check is required on an update, which ultimately
+    enables PUT creates when the versioning policy is set to
+    "versioned-update".
+**************************************************************/
```
rotodd, line 7: Modifications made to the `UpsertResource_6` stored procedure are easier to review in `UpsertResource_6.sql`.

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/FhirStorageTests.cs
```diff
@@ -366,45 +366,6 @@ public async Task UpdatingAResourceWithNoHistory_ThenWeCannotAccessHistoricValue
...
-            ResourceElement historyResults = await Mediator.SearchResourceHistoryAsync(KnownResourceTypes.Observation, updateResult.RawResourceElement.Id);
-
-            // The history bundle has both versions because history is kept
-            Bundle bundle = historyResults.ToPoco<Bundle>();
-            Assert.Equal(2, bundle.Entry.Count);
-
-            Assert.Equal(WeakETag.FromVersionId(updateResult.RawResourceElement.VersionId).ToString(), bundle.Entry.Max(entry => entry.Response.Etag));
-            Assert.Equal(WeakETag.FromVersionId(saveResult.RawResourceElement.VersionId).ToString(), bundle.Entry.Min(entry => entry.Response.Etag));
-        }
-
```
rotodd, line 407: These have been moved to `FhirStorageVersioningPolicyTests.cs`.

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Resources/ResourceHandlerTests.cs
```diff
@@ -207,26 +207,6 @@ ResourceWrapper CreateWrapper(ResourceWrapper wrapper)
...
-        [Fact]
-        public async Task GivenAFhirMediator_WhenSavingAResourceWithoutETagAndETagIsRequiredWithUpdateCreate_ThenPreconditionFailExceptionIsThrown()
-        {
-            var resource = Samples.GetDefaultPatient();
-
-            // Documented for completeness, but under this situation arguably this request should succeed.
-            // Versioned-update + UpdateCreate; When no If-Match header is provided, the request should allow create but not update
-            await Assert.ThrowsAsync<PreconditionFailedException>(async () => await _mediator.UpsertResourceAsync(resource, null));
-        }
-
```
rotodd, line 229: These have been converted to integration tests since the logic being tested now lives at the data store layer.

rotodd:
Different versions of the FHIR spec require different error codes to be returned when there are versioning conflicts. This has been accounted for in the upsert logic and automated tests in 8840449 and 564cf6f.

A summary of manual testing can be found [here](https://microsoft-my.sharepoint.com/:x:/g/personal/rotodd_microsoft_com1/EU9rTJGc8hdCjGe7BIFp3CEBNvU0p2QKIXqLfXJlTsD8aQ?e=CtSRmi).

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
rotodd: My bad, it turns out the spec requires a different error code for each scenario. This check is no longer needed and has been removed in 564cf6f. 

I created a table to summarize the different scenarios and their corresponding result codes [here](https://microsoft-my.sharepoint.com/:x:/g/personal/rotodd_microsoft_com1/EU9rTJGc8hdCjGe7BIFp3CEBNvU0p2QKIXqLfXJlTsD8aQ?e=CtSRmi).

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -155,20 +174,36 @@ public async Task<UpsertOutcome> UpsertAsync(ResourceWrapper resource, WeakETag
                             throw new MethodNotAllowedException(Core.Resources.ResourceCreationNotAllowed);
                         case SqlErrorCodes.TimeoutExpired:
                             throw new RequestTimeoutException(Resources.ExecutionTimeoutExpired);
+                        case 50400: // TODO: Add this to SQL error codes in AB#88286
```
rotodd, line 177: The PR with this change is [here](https://github.com/microsoft/healthcare-shared-components/pull/400).

test/Microsoft.Health.Fhir.R4.Tests.Integration/Persistence/FhirStorageVersioningPolicyTests.cs
```diff
@@ -0,0 +1,51 @@
+﻿// -------------------------------------------------------------------------------------------------
```
LTA-Thinking, line 1: What are your thoughts on using the `#if Stu3` method to combine the unit tests into one file?

LTA-Thinking (approved):

test/Microsoft.Health.Fhir.R4.Tests.Integration/Persistence/FhirStorageVersioningPolicyTests.cs
rotodd: Ah, I really like the idea of this. Just tried it out locally, but the `#if Stu3` flag doesn't seem to get picked up by the integration tests. I noticed all other tests that use this are either unit tests or e2e tests. @brendankowitz is this by-design?

rotodd: Going to go ahead and merge the PR for now, since I don't think this should be blocking. However, I am happy to add this in later if it makes sense.

PR #2467, opened 2022-02-02
Handles invalid header on versioned update

rotodd:
## Description
When the versioning policy is set to "versioned-update", we require that the most recent version of the resource is provided in the request's `if-match` header on an update. The specified version must be in ETag format.

Previously, a 500 would be returned if the version was invalid or in an incorrect format. This PR instead returns a 400 Bad Request.

## Related issues
Addresses [AB#87737](https://microsofthealth.visualstudio.com/Health/_workitems/edit/87737).

## Testing
Updated the manual test flow to include this case.

A lot of the versioning policy test setup was done in #2457. An automated test for this scenario will be added once that PR merges and this branch can be rebased with those changes.

## FHIR Team Checklist
- ✅ CI is green before merge

None (bug fix)


src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/WeakETagBinder.cs
```diff
@@ -22,7 +22,15 @@ public Task BindModelAsync(ModelBindingContext bindingContext)
             WeakETag model = null;
             if (!string.IsNullOrWhiteSpace(suppliedWeakETag))
             {
-                model = WeakETag.FromWeakETag(suppliedWeakETag);
+                try
```
LTA-Thinking, line 25: Should this class have unit tests?

rotodd: Hmm I am not opposed to adding them in, but here is my thought process on why I left them out: we do have unit tests for ETag conversions in `WeakETagTests.cs`, and the integration test I'll add once PR [#2457](https://github.com/microsoft/fhir-server/pull/2457) is merged should ensure that a 400 is returned when an invalid version is passed in by the user.

I also noticed our other binder class (`PartialDateTimeBinder.cs`) doesn't have unit tests. Both binder classes only have the `BindModelAsync()` method, and I'm not confident that adding unit tests for it would add much value (I could be wrong, though).

Let me know your thoughts!

LTA-Thinking: Sounds good, I just wanted to make sure there was some test coverage for this.

LTA-Thinking (approved):

LTA-Thinking (approved):

sghino:
The PATCH operation seems to be not possible with the policy version-update. The operation returns  the following error: 400 - A valid if-match header is required for resource type 'Organization'

LTA-Thinking:
@EXPEkesheth, FYI
@sghino would you mind making an issue for this problem? We don't tend to monitor old PRs.

PR #2468, opened 2022-02-02
Record exception if SearchParameterStatus out of sync with resource

feordin:
## Description
It is possible with failures during certain race conditions between multiple FHIR instances to get the SearchParameterStatus out of sync with the underlying SearchParamterResource object.  Currently the service throws an exception, but we will now log the error.

Also added integration tests which replicate the issue.

## Related issues
Addresses [issue [AB#87426](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/87426)].

## Testing
New integration tests added which simulate a second FHIR service instance.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterStatusManager.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterStatusManager.cs
@@ -206,7 +206,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Registry
                     // and there is an entry in the list of updates with a delete status then it indicates
                     // the search parameter was deleted before it was added to this instance, and there is no issue
                     // however if there is no indication that the search parameter was deleted, then there is a problem
-                    throw new UnableToUpdateSearchParameterException(paramStatus.Uri);
+                    _logger.LogError(Resources.UnableToUpdateSearchParameter, paramStatus.Uri);
                 }
             }
 
--- a/src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterStatusManager.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterStatusManager.cs
@@ -12,6 +12,7 @@ using EnsureThat;
 using MediatR;
 using Microsoft.Extensions.Logging;
 using Microsoft.Health.Core;
+using Microsoft.Health.Fhir.Core;
 using Microsoft.Health.Fhir.Core.Features.Definition;
 using Microsoft.Health.Fhir.Core.Features.Search.Parameters;
 using Microsoft.Health.Fhir.Core.Messages.Search;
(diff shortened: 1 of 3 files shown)


src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterStatusManager.cs
```diff
@@ -206,7 +206,7 @@ internal async Task ApplySearchParameterStatus(IReadOnlyCollection<ResourceSearc
                     // and there is an entry in the list of updates with a delete status then it indicates
                     // the search parameter was deleted before it was added to this instance, and there is no issue
                     // however if there is no indication that the search parameter was deleted, then there is a problem
-                    throw new UnableToUpdateSearchParameterException(paramStatus.Uri);
+                    _logger.LogError(Resources.UnableToUpdateSearchParameter, paramStatus.Uri);
```
LTA-Thinking, line 209: Since we aren't throwing an error anymore here, are there any issues with the rest of the method executing?

feordin: Good question!  By allowing the method to continue, any other search parameter updates which are "good" can continue and be unblocked from being sync'd.  It does mean that the update with an issue will not be processed, but the date time will be updated, so it will never be processed.  All we will have is the error log to help diagnose the issue when the affected SearchParameter is not working correctly.

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Features/Operations/Reindex/ReindexJobTests.cs
```diff
@@ -902,14 +902,21 @@ private ResourceWrapper CreateObservationResourceWrapper(string observationId)
             return wrapper;
         }
 
-        private ResourceWrapper CreateSearchParamResourceWrapper(SearchParameter searchParam, bool deleted = false)
+        private ResourceWrapper CreateSearchParamResourceWrapper(SearchParameter searchParam, string id = "searchParam1", bool deleted = false)
```
LTA-Thinking, line 905: You've added this new id parameter, but it looks like it isn't used and the id is still hardcoded to searchParam1.

feordin: It is true, I am only using it on line 570, as only this test case seemed troubled by the same param id.

LTA-Thinking: But this parameter isn't used in the method anywhere. You set it on line 570, but that value is never used.

feordin: You are correct! Sorry.  Removed.

LTA-Thinking (approved):

PR #2491, opened 2022-02-17
Check tag value for history tests

apurvabhaleMS:
## Description
Issue - History tests were failing in the CI pipeline due to multiple tests creating different resources during the same time causing the mismatch with the history query count. (as it pulls all the resources updated during the given timeframe in the test)

This PR filters the response with the matching tag value in the test. Note that the FHIR history search doesn't account for a tag value so adding that filtering in the test. Also, iterating over all the pages for the search query to retrieve all of the resources updated during the given timeframe with the matching tag.

## Related issues
Addresses [[86231](https://dev.azure.com/microsofthealth/Health/_workitems/edit/86231/)].

## Testing
Enabled parallelization and ran multiple instances of FHIR services running the same history tests to make sure there are no failures.

## FHIR Team Checklist

### DIFF
--- a/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/HistoryTests.cs
+++ b/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/HistoryTests.cs
@@ -26,6 +26,14 @@ using Task = System.Threading.Tasks.Task;
 
 namespace Microsoft.Health.Fhir.Tests.E2E.Rest
 {
+    /// <summary>
+    /// This class covers history tests with different supported query parameters
+    /// Other test collections are run parallelly causing some of the tests to fail intermittently as during the same timeframe resources could get created by other tests as well
+    /// We generally use _tag query parameter to filter out the resources but since history search doesn't support _tag, filtering is done explicitly in the test
+    /// While returning the search response default maxItem count is set to 10, due to multiple tests running at the same time this could return entries with different tag code
+    /// hence using the NextLink to keep querying for the next set
+    /// Some tests have Thread.Sleep to avoid query time to fall in future
+    /// </summary>
     [HttpIntegrationFixtureArgumentSets(DataStore.All, Format.All)]
     [CollectionDefinition("History", DisableParallelization = true)]
     [Collection("History")]
@@ -33,7 +41,6 @@ namespace Microsoft.Health.Fhir.Tests.E2E.Rest
     {
         private FhirResponse<Observation> _createdResource;
         private readonly TestFhirClient _client;
-        private readonly string count = "50";
 
         public HistoryTests(HttpIntegrationTestFixture fixture)
         {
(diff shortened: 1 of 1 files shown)


rotodd:
I noticed that this class has the attribute `[CollectionDefinition("History", DisableParallelization = true)]`.

Just curious, why doesn't disabling parallelization at this level prevent resource overlap between tests?

apurvabhaleMS:
> I noticed that this class has the attribute `[CollectionDefinition("History", DisableParallelization = true)]`.
> 
> Just curious, why doesn't disabling parallelization at this level prevent resource overlap between tests?

Good question, parallelization is disabled for the collection name History but the other unit tests from different collections could run together creating different resources and falling in the same time range as the history tests. I believe that's what is overlapping with the history tests.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/HistoryTests.cs
```diff
@@ -127,20 +128,22 @@ public async Task GivenAValueForSince_WhenGettingSystemHistory_TheServerShouldRe
 
             var updatedResource = await CreateResourceWithTag(_createdResource.Resource, tag);
 
-            using FhirResponse<Bundle> readResponse = await _client.SearchAsync($"_history?_tag={tag}&_since={sinceUriString}");
+            using FhirResponse<Bundle> readResponse = await _client.SearchAsync($"_history?_tag={tag}&_since={sinceUriString}&_count={count}");
```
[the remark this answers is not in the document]
apurvabhaleMS: I think it does, I will remove the _tag from the search.
Probably It is a good idea to return the warning to be consistent with others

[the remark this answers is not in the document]
rotodd: Sounds good! I'll make a user story for it and link it here.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/HistoryTests.cs
```diff
@@ -127,20 +128,22 @@ public async Task GivenAValueForSince_WhenGettingSystemHistory_TheServerShouldRe
...
 
-            AssertCount(1, readResponse.Resource.Entry);
+            var readResponseWithMatchingTag = readResponse.Resource.Entry.Where(e => e.Resource.Meta.Tag.Where(t => t.Code == tag).Any()).ToList();
```
rotodd, line 133: Nit: ReSharper is telling me this can be simplified slightly

`var readResponseWithMatchingTag = readResponse.Resource.Entry.Where(e => e.Resource.Meta.Tag.Any(t => t.Code == tag)).ToList();`

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/HistoryTests.cs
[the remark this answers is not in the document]
rotodd: [AB#88880](https://microsofthealth.visualstudio.com/Health/_workitems/edit/88880)

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/HistoryTests.cs
```diff
@@ -26,14 +26,21 @@
...
 namespace Microsoft.Health.Fhir.Tests.E2E.Rest
 {
+    /// <summary>
+    /// This class covers history tests with different supported query parameters
+    /// Other test collections are run parallelly causing some of the tests to fail intermittently as during the same timeframe resources could get created by other tests as well
+    /// We generally use _tag query parameter to filter out the resources but since history search doesn't support _tag, filtering is done explicitly in the test
+    /// While returning the search response default maxItem count is set to 10, due to multiple tests running at the same time this could return entries with different tag code
+    /// hence using the NextLink to keep querying for the next set
+    /// Some tests have Thread.Sleep to avoid query time to fall in future
+    /// </summary>
```
rotodd, line 36: +1

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/HistoryTests.cs
```diff
@@ -222,28 +258,56 @@ public async Task GivenAValueForSinceAndBeforeCloseToLastModifiedTime_WhenGettin
...
-            using FhirResponse<Bundle> allChanges = await _client.SearchAsync($"_history?_tag={tag}&_since={sinceUriString}&_count={count}");
+            FhirResponse<Bundle> response;
+            List<Bundle.EntryComponent> allChanges = new List<Bundle.EntryComponent>();
+            string searchString = $"_history?_since={sinceUriString}";
+            while (!string.IsNullOrEmpty(searchString))
+            {
+                response = await _client.SearchAsync(searchString);
+                allChanges.AddRange(response.Resource.Entry);
+                searchString = response.Resource.NextLink?.ToString();
+            }
```
rotodd, line 269: Nit: this block of code seems to be reused a lot - maybe we could move it to a common method?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/HistoryTests.cs
```diff
@@ -120,58 +127,88 @@ public async Task GivenAValueForSince_WhenGettingSystemHistory_TheServerShouldRe
...
+                response = await _client.SearchAsync(searchString);
+                var readResponseWithMatchingTag = response.Resource.Entry.Where(e => e.Resource.Meta.Tag.Any(t => t.Code == tag)).ToList();
 
-            var readResponseWithMatchingTag = readResponse.Resource.Entry.Where(e => e.Resource.Meta.Tag.Where(t => t.Code == tag).Any()).ToList();
-            AssertCount(1, readResponseWithMatchingTag);
+                readResponse.AddRange(readResponseWithMatchingTag);
+                searchString = response.Resource.NextLink?.ToString();
 
-            var obsHistory = readResponseWithMatchingTag[0].Resource as Observation;
+                if (readResponseWithMatchingTag.Count == 1)
```
rotodd, line 150: Just checking my understanding, are we adding this if-statement to make sure we find the self link that points to the page that has the resource with the matching tag?

rotodd: Sorry, last follow up question! Do we need to iterate over subsequent pages when doing the self link search on line 163 if the link already points the page with the matching tag?

apurvabhaleMS: No problem! I think we do need to iterate over the pages. SelfLink looks like this - http://localhost/_history?_since=2022-02-24T05%3A19%3A51.7570000%2B00%3A00. Also, the tag doesn't really work so there isn't a guarantee to get the right result in one iteration. For example, check the SS below 
image
Before we query for the selfLink I added a bunch of resources and when we query for the first time the matching tag result is not returned. (check the watch for matching tag expression)

rotodd (approved):

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/HistoryTests.cs
rotodd: As discussed, the self link doesn't point to a specific page. It will be set to the initial query string, and we are simply setting it when we find the resource with the matching tag.

rotodd: Backtracking a bit here: each page's self link _does_ point back to that specific page. Every self link except for the self link of the first page of results will not consider new resources that are created after the initial request.

A test I carried out:

1. Created a number of patients
2. Submitted the request: `GET <fhir-server>/Patient/_history`, noting the self link on the first page is set to `GET <fhir-server>/Patient/_history`
3. Used the next link to navigate to the second page of results
4. Saved the self link on the second page of results, noting it includes a continuation token
5. Created more patients
6. Used the self link saved in step 4, noting it points back to that specific page
7. Used the next link to navigate to the third page of results, noting the patients created in step 5 do not show up

PR #2497, opened 2022-02-23
SQL timeout is returning a 500 error

v-mserdar:
## Description
SQL timeout is returning a 500 error

## Related issues
Addresses [86829](https://microsofthealth.visualstudio.com/Health/_workitems/edit/86829/)

## Testing
Testing was manually done since I had to inject a SQL script into the stored procedure being tested. For example, during local debugging I set a breakpoint at a point where the SQLCommand.CommandText was being set, then I revised that text by adding in this specific SQL, "WAITFOR DELAY '00:00:35';". The reason for that is the default SQLCommand.CommandTimeout is 30 seconds and given the script that I added, it will force the current SQL statement to take an additional 35 seconds thereby causing a SQL timeout.

## FHIR Team Checklist


feordin (approved):

src/Microsoft.Health.Fhir.Api/Microsoft.Health.Fhir.Api.csproj
```diff
@@ -26,6 +26,7 @@
     <PackageReference Include="Microsoft.Health.Abstractions" Version="$(HealthcareSharedPackageVersion)" />
     <PackageReference Include="Microsoft.Health.Api" Version="$(HealthcareSharedPackageVersion)" />
     <PackageReference Include="Microsoft.Health.Extensions.DependencyInjection" Version="$(HealthcareSharedPackageVersion)" />
+    <PackageReference Include="Microsoft.Health.SqlServer" Version="$(HealthcareSharedPackageVersion)" />
```
brendankowitz, line 29: @feordin @v-mserdar we should be really careful about the dependencies we introduce between projects. I ideally don't see why and API layer should know anything about SQL.

v-mserdar: @brendankowitz  There is another way we can check for this and that is to replace the SqlException catch and replace it with a Win32Exception as that is the inner exception type. If we do that then we can remove the dependency of concern. If we go this route then the code snippet would check like the following. Please advise on how you want to handle the hardcoded 258 which is the error number that is returned. I take it we will add it to a file as a const?

else if (context.Exception is Win32Exception win32Exception && win32Exception.NativeErrorCode == 258)
{
    context.Result = CreateOperationOutcomeResult(win32Exception.Message, OperationOutcome.IssueSeverity.Error, OperationOutcome.IssueType.Timeout, HttpStatusCode.RequestTimeout);
    context.ExceptionHandled = true;
}

PR #2499, opened 2022-02-25
Personal/peizhou/fiximportperfissue

peizhou298:
## Description
This PR fix the perf issue in import caused by low-effiency distinct function in search parameter generator.

## Related issues
Addresses [issue #].

## Testing
Add a UT for test.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Configs/ImportTaskConfiguration.cs
+++ b/src/Microsoft.Health.Fhir.Core/Configs/ImportTaskConfiguration.cs
@@ -16,6 +16,7 @@ namespace Microsoft.Health.Fhir.Core.Configs
         private const int DefaultSqlCleanResourceBatchSize = 1000;
         private const int DefaultSqlMaxRebuildIndexOperationConcurrentCount = 3;
         private const int DefaultSqlMaxDeleteDuplicateOperationConcurrentCount = 3;
+        private const int DefaultSqlMaxDatatableProcessConcurrentCount = 3;
         private const int DefaultSqlLongRunningOperationTimeoutInSec = 60 * 60 * 2;
         private const int DefaultSqlBulkOperationTimeoutInSec = 60 * 10;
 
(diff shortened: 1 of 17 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/DataGenerator/DateTimeSearchParamsTableBulkCopyDataGenerator.cs
```diff
@@ -72,5 +75,61 @@ internal override void FillSchema(DataTable table)
...
+        internal class BulkDateTimeSearchParamTableTypeV2RowComparer : IEqualityComparer<BulkDateTimeSearchParamTableTypeV2Row>
+        {
+            public bool Equals(BulkDateTimeSearchParamTableTypeV2Row x, BulkDateTimeSearchParamTableTypeV2Row y)
+            {
+                if (x.SearchParamId != y.SearchParamId)
+                {
+                    return false;
+                }
+
+                if (!x.StartDateTime.Equals(y.StartDateTime))
```
LTA-Thinking, line 93: Instead of using the object's Equal method, should we use the static DateTimeOffset.Equals method to avoid potential null pointer exception?

peizhou298: Actually the StartDateTime or EtartDateTime will not be null according to the .sql definition, but from c# code perspective, I think your advice is better, so I turned to use DateTimeOffset.Equals instead.

LTA-Thinking (approved):

src/Microsoft.Health.Fhir.Core/Configs/ImportTaskConfiguration.cs
```diff
@@ -16,6 +16,7 @@ public class ImportTaskConfiguration
         private const int DefaultSqlCleanResourceBatchSize = 1000;
         private const int DefaultSqlMaxRebuildIndexOperationConcurrentCount = 3;
         private const int DefaultSqlMaxDeleteDuplicateOperationConcurrentCount = 3;
+        private const int DefaultSqlMaxDatatableProcessConcurrentCount = 3;
```
SergeyGaluzo, line 19: What was it before?

tongwu-msft: previous it is sequence, it is 1.

tongwu-msft: previous convert to sql format is sequence => 1

SergeyGaluzo: What is the reason for this change?

PR #2504, opened 2022-03-03
Use retryable wrappers instead of using SqlConnection string

apurvabhaleMS:
## Description
Use retryable wrappers instead of using SqlConnection string to create SqlConnection and SqlCommand which fails with transient issues today.

Using scoped SqlConnectionWrapperFactory in SqlServerFhirModel . SqlServerFhirModel is registered as a singleton service and SqlConnectionWrapperFactory is registered as scoped service. In this change, we are using SqlConnectionWrapperFactory in SqlServerFhirModel but we can not consume scoped service or transient service from a singleton. (You should also avoid consuming transient service from a scoped service.)

Your singleton SqlServerFhirModel would get an instance of SqlConnectionWrapperFactory, and every other time you try to get an instance of SqlServerFhirModel since it’s a singleton it will have the same state, and it will always contain the same instance of SqlConnectionWrapperFactory, meaning that SqlConnectionWrapperFactory effectively becomes a singleton inside of SqlServerFhirModel.

What is happening when you consume scoped service from a singleton service is known as a captive dependency. It occurs when a service intended to live for a short(er) amount of time gets held by a service that lives for a more extended amount of time. This almost always suggests that you should change something in your design.

## Related issues
Addresses [[87996](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Olympus/Stories/?workitem=87996)].

## Testing
Existing tests pass and some manually testing

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
@@ -16,7 +16,7 @@ using Microsoft.Health.Abstractions.Data;
 using Microsoft.Health.Fhir.Core.Models;
 using Microsoft.Health.Fhir.SqlServer.Features.Schema;
 using Microsoft.Health.Fhir.SqlServer.Features.Schema.Model;
-using Microsoft.Health.SqlServer;
+using Microsoft.Health.SqlServer.Features.Client;
 using Microsoft.Health.SqlServer.Features.Schema;
 using Microsoft.Health.SqlServer.Features.Storage;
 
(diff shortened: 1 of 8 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/ChangeFeed/SqlServerFhirResourceChangeDataStore.cs
```diff
@@ -98,49 +98,43 @@ public async Task<IReadOnlyCollection<ResourceChangeData>> GetRecordsAsync(long
...
-                using (SqlConnection sqlConnection = await _sqlConnectionFactory.GetSqlConnectionAsync(cancellationToken: cancellationToken))
+                using (SqlConnectionWrapper sqlConnectionWrapper = await _sqlConnectionWrapperFactory.ObtainSqlConnectionWrapperAsync(cancellationToken, true))
+                using (SqlCommandWrapper sqlCommandWrapper = sqlConnectionWrapper.CreateSqlCommand())
                 {
-                    await sqlConnection.OpenAsync(cancellationToken);
                     if (ResourceTypeIdToTypeNameMap.IsEmpty)
                     {
                         lock (ResourceTypeIdToTypeNameMap)
                         {
-                            if (ResourceTypeIdToTypeNameMap.IsEmpty)
```
bbatjargal, line 108: Due to multithread, we need to check the state of ResourceTypeIdToTypeNameMap again after entering the lock statement's block.

apurvabhaleMS: Good catch! I think I took it out by mistake. Readded the check

bbatjargal (approved):

feordin (approved):

PR #2507, opened 2022-03-05
Users/sergal/import

SergeyGaluzo:
## Description
Introduction of high performance import tool.


src/Microsoft.Health.Fhir.Store.Utils/BatchExtensions.cs
```diff
@@ -0,0 +1,138 @@
...
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.Health.Fhir.Store.Utils
+{
+    public static class BatchExtensions
```
zxue, line 14: Any reason this class file and the CancelRequest.cs file are placed in a different folder? Unless the plan is to reuse them, including them in the import project folder would make it easy to download and test the import tool.

tools/Import/Importer.cs
```diff
@@ -0,0 +1,271 @@
...
+            ServicePointManager.DefaultConnectionLimit = 2048;
+
+            if (string.IsNullOrEmpty(fhirEndpoint2))
+            {
+                fhirEndpoint2 = FhirEndpoint;
+            }
+
+            var globalLogPrefix = $"GlobalBlobRange=[{NumberOfBlobsToSkip + 1}-{NumberOfBlobsToImport}]";
+            Console.WriteLine($"{globalLogPrefix}: Starting...");
+            var blobContainerClient = GetContainer(ConnectionString, ContainerName);
```
zxue, line 75: Is it better to define the blobContainerClient as a global variable than a local variable? This way, it doesn't need to be created again in the GetLinesInBlobRange method.

tools/Import/Importer.cs
```diff
@@ -0,0 +1,271 @@
...
+            var blobs = blobContainerClient.GetBlobs().Skip(NumberOfBlobsToSkip).Take(NumberOfBlobsToImport - NumberOfBlobsToSkip).ToList();
```
zxue, line 76: I would consider a filter option, to allow some flexibility.
            var blobs = blobContainerClient.GetBlobs(prefix: FolderOrFileFilter).Skip(NumberOfBlobsToSkip).Take(NumberOfBlobsToImport).ToList();

tools/Import/Importer.cs
```diff
@@ -0,0 +1,271 @@
...
+            catch
+            {
+                Console.WriteLine($"Unable to parse stroage reference or connect to storage account {connectionString}.");
+                throw;
+            }
+        }
+
+        private static void PutResource(string jsonString, ref long resourceCount, ref long epCalls, ref long waits)
+        {
+            ParseJson(jsonString, out var resourceType, out var resourceId);
```
zxue, line 181: Is there any significant performance difference between ParseJson and ParseJsonWithJObject? IndexOf may be faster than JObject parsing, but the latter ensures a valid json object and returns the id and resource type nicely.

tools/Import/Importer.cs
```diff
@@ -0,0 +1,271 @@
...
+                        PutResource(line, ref localResourceCount, ref epCalls, ref waits);
+                        if (swReport.Elapsed.TotalSeconds > ReportingPeriodSec)
+                        {
+                            lock (swReport)
+                            {
+                                if (swReport.Elapsed.TotalSeconds > ReportingPeriodSec)
+                                {
+                                    var resWritten = Interlocked.Read(ref globalResourceCount);
+                                    var resRead = Interlocked.Read(ref resourcesRead);
+                                    Console.WriteLine($"{globalLogPrefix}.Readers={Interlocked.Read(ref readers)}.Writers={Interlocked.Read(ref writers)}.EndPointCalls={Interlocked.Read(ref epCalls)}.Waits={Interlocked.Read(ref waits)}: reads={resRead} writes={resWritten} secs={(int)swWrites.Elapsed.TotalSeconds} read-speed={(int)(resRead / swReads.Elapsed.TotalSeconds)} lines/sec write-speed={(int)(resWritten / swWrites.Elapsed.TotalSeconds)} res/sec");
```
zxue: The reporting feature is really nice. I'd replace globalLogPrefix with localLogPrefix and add "..." to add more info and improve readability on the screen.
Console.WriteLine($"... {localLogPrefix}.Readers={Interlocked.Read(ref readers)}.Writers={Interlocked.Read(ref writers)}.EndPointCalls={Interlocked.Read(ref epCalls)}.Waits={Interlocked.Read(ref waits)}: reads={resRead} writes={resWritten} secs={(int)swWrites.Elapsed.TotalSeconds} read-speed={(int)(resRead / swReads.Elapsed.TotalSeconds)} lines/sec write-speed={(int)(resWritten / swWrites.Elapsed.TotalSeconds)} res/sec");

tools/Import/Importer.cs
```diff
@@ -0,0 +1,271 @@
...
+            BatchExtensions.ExecuteInParallelBatches(blobs, ReadThreads, BlobRangeSize, (reader, blobRangeInt) =>
+            {
+                var localSw = Stopwatch.StartNew();
+                var localResourceCount = 0L;
+                var blobRangeIndex = blobRangeInt.Item1;
+                var blobRange = blobRangeInt.Item2;
+                var localLogPrefix = $"Reader={reader}.BlobRange=[{NumberOfBlobsToSkip + (blobRangeIndex * BlobRangeSize) + 1}-{NumberOfBlobsToSkip + (blobRangeIndex * BlobRangeSize) + blobRange.Count}]";
+                Console.WriteLine($"{localLogPrefix}: Starting...");
+
+                BatchExtensions.ExecuteInParallelBatches(GetLinesInBlobRange(blobRange, localLogPrefix), WriteThreads / ReadThreads, 100, (thread, lineBatch) =>
```
zxue: Is the number of write threads per read thread (or within a read thread)? Is 100 the write batch size? Any significance of WriteThreads / ReadThreads?

tools/Import/Importer.cs
```diff
@@ -0,0 +1,271 @@
...
+        private static readonly HttpClient HttpClient = new HttpClient();
+        private static readonly string ConnectionString = ConfigurationManager.AppSettings["ConnectionString"];
+        private static readonly string FhirEndpoint = ConfigurationManager.AppSettings["FhirEndpoint"];
+        private static string fhirEndpoint2 = ConfigurationManager.AppSettings["FhirEndpoint2"];
+        private static readonly string ContainerName = ConfigurationManager.AppSettings["ContainerName"];
+        private static readonly int BlobRangeSize = int.Parse(ConfigurationManager.AppSettings["BlobRangeSize"]);
+        private static readonly int ReadThreads = int.Parse(ConfigurationManager.AppSettings["ReadThreads"]);
+        private static readonly int WriteThreads = int.Parse(ConfigurationManager.AppSettings["WriteThreads"]);
+        private static readonly int NumberOfBlobsToSkip = int.Parse(ConfigurationManager.AppSettings["NumberOfBlobsToSkip"]);
+        private static readonly int NumberOfBlobsToImport = int.Parse(ConfigurationManager.AppSettings["NumberOfBlobsToImport"]);
```
zxue, line 36: To me the NumberOfBlobsToImport means how many blobs you'd import, excluding the NumberOfBlobsToSkip. I can skip 10000 blobs and import 2000 new blobs. Next time I can skip 12000 blobs and import 5000 blobs. I guess it's just a performance preference.
var globalLogPrefix = $"GlobalBlobRange=[{NumberOfBlobsToSkip + 1}-{NumberOfBlobsToImport + NumberOfBlobsToSkip}]";

tools/Import/Importer.cs
```diff
@@ -0,0 +1,271 @@
...
+                                }
+                            }
+                        }
+                    }
+
+                    Interlocked.Decrement(ref writers);
+                });
+                Console.WriteLine($"{localLogPrefix}: Completed writes. Total={localResourceCount} secs={(int)localSw.Elapsed.TotalSeconds} speed={(int)(localResourceCount / localSw.Elapsed.TotalSeconds)} res/sec");
+            });
+            Console.WriteLine($"{globalLogPrefix}.Readers={readers}.Writers={writers}.EndPointCalls={epCalls}.Waits={waits}: total reads={resourcesRead} total writes={globalResourceCount} secs={(int)swWrites.Elapsed.TotalSeconds} read-speed={(int)(resourcesRead / swReads.Elapsed.TotalSeconds)} lines/sec write-speed={(int)(globalResourceCount / swWrites.Elapsed.TotalSeconds)} res/sec");
```
zxue: I'd provide a summary report by defining a string list and adding reads and writes messages to it.

            var doneMsg = $"{globalLogPrefix}.Readers={readers}.Writers={writers}.EndPointCalls={epCalls}.Waits={waits}: total reads={resourcesRead}; total writes={globalResourceCount}; total secs={(int)swWrites.Elapsed.TotalSeconds}; read-speed={(int)(resourcesRead / swReads.Elapsed.TotalSeconds)} lines/sec; write-speed={(int)(globalResourceCount / swWrites.Elapsed.TotalSeconds)} res/sec";
            summaryReport.Add(doneMsg);
            //Console.WriteLine(doneMsg);
            Console.WriteLine("\n*** Summary Report ***");
            Console.WriteLine($"{DateTime.Now.ToLongDateString()}");
            foreach (var msg in summaryReport) Console.WriteLine(msg);

tools/Import/Importer.cs
```diff
@@ -0,0 +1,271 @@
...
+
+                if (bad && retries < maxRetries)
+                {
+                    retries++;
+                    Interlocked.Increment(ref waits);
+                    Thread.Sleep(networkError ? 1000 : 200 * retries);
+                    Interlocked.Decrement(ref waits);
+                }
+            }
+            while (bad && retries < maxRetries);
```
zxue, line 241: It'd be useful to log lines that didn't get imported.

src/Microsoft.Health.Fhir.Store.Utils/BatchExtensions.cs
SergeyGaluzo: Yes, I plan to use BatchExtentions on other project too.

tools/Import/Program.cs
```diff
@@ -0,0 +1,15 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Fhir.Import
+{
+    public static class Program
```
zxue, line 8: While optional, the AAD authentication module using client id and client secret can add value to the tool.

SergeyGaluzo: I agree. It can be added later in different PR.

src/Microsoft.Health.Fhir.Store.Utils/BatchExtensions.cs
```diff
@@ -0,0 +1,138 @@
...
+            task.Start(TaskScheduler.Default);
+            return task;
+        }
+
+        internal static void ParallelForEach<T>(IEnumerable<T> objects, int threads, Action<int, T> action, CancelRequest cancel = null)
+        {
+            ExecuteInParallelBatches(objects, threads, 1, (thread, batch) => { action(thread, batch.Item2.First()); }, null, cancel);
+        }
+
+        public static void ExecuteInParallelBatches<T>(IEnumerable<T> objects, int threads, int batchSize, Action<int, Tuple<int, IList<T>>> action, int? queueCapacity = null, CancelRequest cancel = null)
```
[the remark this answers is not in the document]
SergeyGaluzo: I added comments in app config.

tools/Import/Importer.cs
SergeyGaluzo: This summary is provided on reporting interval many times during run. I want it to be single line.

tools/Import/Importer.cs
SergeyGaluzo: I changed name to clarify.

tools/Import/Importer.cs
SergeyGaluzo: Write threads is total. Each "reader" gets its equal share of writers. 100 is batch size in internal queueing, all http calls are still single resource calls.

tools/Import/Importer.cs
SergeyGaluzo: Logs are cluttered with lots of retry messages, so it is not easy to navigate without pattern searches anyway. I don't think adding "..." would be that useful.

tools/Import/Importer.cs
SergeyGaluzo: There is huge difference in memory allocation and, hence, perf. I did not measure precisely but gut feeling is 10x.

tools/Import/Importer.cs
SergeyGaluzo: The problem is that retry logic inside Azure client gets exceptions sometimes. We will need to retry that. Creating new containers is not that expensive compared to all the work we do.

tools/Import/Importer.cs
SergeyGaluzo: I redesigned this part.

tools/Import/Importer.cs
JCookit: in general, i'm _very_ afraid of string parsing json.  i'm not convinced we should be offering it as an option, as there are best-practice reasons against it.    JObject has evolved to be pretty light; if we really want to do better, maybe something based on Utf8JsonReader

SergeyGaluzo: Any JSON reader will always loose compared to parsing of first 100 characters of the string.

tools/Import/Import.csproj
```diff
@@ -0,0 +1,25 @@
...
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\Microsoft.Health.Fhir.Store.Utils\Microsoft.Health.Fhir.Store.Utils.csproj" />
+  </ItemGroup>
+
+  <PropertyGroup>
+    <ServerGarbageCollection>true</ServerGarbageCollection>
+    <AssemblyName>Fhir.Import</AssemblyName>
+    <RootNamespace>Microsoft.Health.Fhir.Import</RootNamespace>
```
brendankowitz, line 22: I think we should match the folder / namespace / assembly here.
Perhaps use `Microsoft.Health.Fhir.Importer` or `Microsoft.Health.Fhir.Loader` to differentiate from $import.

tools/Import/Importer.cs
```diff
@@ -0,0 +1,358 @@
...
+                    }
+                }
+
+                return;
+            }
+
+            ParseJsonWithJObject(jsonString, out resourceType, out resourceId);
+        }
+
+        private static void ParseJsonWithJObject(string jsonString, out string resourceType, out string resourceId)
```
brendankowitz, line 323: Might be nicer to use a tuple:
```
private static (string ResourceType, string ResourceId) ParseJsonWithJObject(string jsonString)
```

tools/Import/Importer.cs
```diff
@@ -0,0 +1,358 @@
...
+                var idStart = jsonString.IndexOf(@"""id"":""", StringComparison.OrdinalIgnoreCase) + 6;
+                var idShort = jsonString.Substring(idStart, 50);
+                var idEnd = idShort.IndexOf(@"""", StringComparison.OrdinalIgnoreCase);
+                resourceId = idShort.Substring(0, idEnd);
+                if (string.IsNullOrEmpty(resourceId))
+                {
+                    throw new ArgumentException("Cannot parse resource id with string parser");
+                }
+
+                var rtShort = jsonString.Substring(17, 30);
```
brendankowitz, line 295: Is this reliable?

tools/Import/Importer.cs
```diff
@@ -0,0 +1,358 @@
...
+                    Thread.Sleep(40);
+                    var response = HttpClient.PutAsync(uri, content).Result;
+                    switch (response.StatusCode)
+                    {
+                        case HttpStatusCode.OK:
+                        case HttpStatusCode.Created:
+                            break;
+                        default:
+                            bad = true;
+                            var statusString = response.StatusCode.ToString();
```
brendankowitz: Perhaps just use the enum?
`status != HttpStatusCode.BadGateway`?

tools/Import/Import.csproj
SergeyGaluzo: Changed.

tools/Import/Importer.cs
```diff
@@ -0,0 +1,358 @@
...
+
+            if (!string.IsNullOrEmpty(Endpoint3))
+            {
+                endpoints.Add(Endpoint3);
+            }
+
+            if (!string.IsNullOrEmpty(Endpoint4))
+            {
+                endpoints.Add(Endpoint4);
+            }
```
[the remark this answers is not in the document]
SergeyGaluzo: Do you think it will be easier?

tools/Import/Importer.cs
brendankowitz: There are also forward-only jsonreaders, i.e. https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-use-dom-utf8jsonreader-utf8jsonwriter?pivots=dotnet-6-0#read-from-a-stream-using-utf8jsonreader, so you would be able to parse until you hit these two properties

tools/Import/Importer.cs
SergeyGaluzo: Cool

tools/Import/Importer.cs
SergeyGaluzo: It is for a given file structure where resourcetype is first field. If it is last it will not work, so people can switch to full JObject type conversion.

tools/Import/Importer.cs
```diff
@@ -0,0 +1,358 @@
...
+            {
+                Console.WriteLine($"Failed writing ResourceType={resourceType} ResourceId={resourceId}. Retries={retries} Endpoint={endpoint}");
+            }
+        }
+
+        private static bool IsNetworkError(Exception e)
+        {
+            return e.Message.Contains("connection attempt failed", StringComparison.OrdinalIgnoreCase)
+                   || e.Message.Contains("connected host has failed to respond", StringComparison.OrdinalIgnoreCase)
+                   || e.Message.Contains("operation on a socket could not be performed", StringComparison.OrdinalIgnoreCase);
```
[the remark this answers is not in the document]
SergeyGaluzo: Maybe. It would require investigation. Do you insist in specific exception types?

tools/Import/Importer.cs
SergeyGaluzo: Right.

tools/Import/Importer.cs
```diff
@@ -0,0 +1,358 @@
...
+        }
+
+        private static void ParseJson(string jsonString, out string resourceType, out string resourceId)
+        {
+            if (UseStringJsonParser)
+            {
+                var idStart = jsonString.IndexOf(@"""id"":""", StringComparison.OrdinalIgnoreCase) + 6;
```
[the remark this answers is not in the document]
SergeyGaluzo: OK

tools/Import/Importer.cs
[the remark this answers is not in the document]
SergeyGaluzo: OK

src/Microsoft.Health.Fhir.Store.Utils/BatchExtensions.cs
```diff
@@ -0,0 +1,138 @@
...
+                    var batchList = new List<T>();
+                    foreach (var obj in objects)
+                    {
+                        if (cancel.IsSet)
+                        {
+                            break;
+                        }
+
+                        batchList.Add(obj);
+                        if (batchList.Count == batchSize)
```
zxue, line 79: I'd consider handling batch size = 1, to avoid looping.


                    if (batchSize==1)
                    {
                        batchList = objects.ToList();
                        AddToQueueWithTrapping(queue, batchId, batchList, cancel, workers, threads, workerAction);
                    }
                    else ...

zxue, line 79: I'd consider handling batch size = 1, to avoid looping.


                    if (batchSize==1)
                    {
                        batchList = objects.ToList();
                        AddToQueueWithTrapping(queue, batchId, batchList, cancel, workers, threads, workerAction);
                    }
                    else ...

tools/Importer/Importer.cs
```diff
@@ -0,0 +1,341 @@
...
+            {
+                throw new ArgumentException("FhirEndpoints value is empty");
+            }
+
+            endpoints = Endpoints.Split(";", StringSplitOptions.RemoveEmptyEntries).ToList();
+
+            var globalPrefix = $"RequestedBlobRange=[{NumberOfBlobsToSkip + 1}-{MaxBlobIndexForImport}]";
+            Console.WriteLine($"{globalPrefix}: Starting...");
+            var blobContainerClient = GetContainer(ConnectionString, ContainerName);
+            var blobs = blobContainerClient.GetBlobs().OrderBy(_ => _.Name).Where(_ => _.Name.EndsWith(".ndjson", StringComparison.OrdinalIgnoreCase)).ToList();
```
feordin, line 60: Would it be better to apply the Where before the OrderBy? i.e.:
blobContainerClient.GetBlobs().Where(_ => _.Name.EndsWith(".ndjson", StringComparison.OrdinalIgnoreCase)).OrderBy(_ => _.Name).ToList();

tools/Importer/Importer.cs
```diff
@@ -0,0 +1,341 @@
...
+            Console.WriteLine($"Found ndjson blobs={blobs.Count} in {ContainerName}.");
+            var take = MaxBlobIndexForImport == 0 ? blobs.Count : MaxBlobIndexForImport - NumberOfBlobsToSkip;
+            blobs = blobs.Skip(NumberOfBlobsToSkip).Take(take).ToList();
```
feordin, line 63: If the setting MaxBlobIndexForImport is greater than the blob.Count, does it cause any problem?  Or will Take simply take as many as it can?

feordin (approved):

tools/Importer/Importer.cs
SergeyGaluzo: As many as it can

tools/Importer/Importer.cs
SergeyGaluzo: Good point. In reality this difference is small. I don't want to risk extra failures in build, so suggest to leave as is.

PR #2508, opened 2022-03-05
FHIR resources blend tool

Tom-Markoch:
## Description
Adds FHIR resources blend tool.

## Testing
Manually, tested blend target ratios calculations, creating blend (blob and file), verifying blends (blob and file), importing blends to server via $import, querying server if import is finished and reporting import errors if any.


LTA-Thinking:
Could you add a readme file on how to use the tool and what it does?

feordin (approved):

PR #2509, opened 2022-03-07
Catch SqlExceptions at the Request level

v-mserdar:
## Description
The changes will remove the project dependency that was originally added to address the original issue.  Now we're registering a new class to the startup that handles Request exceptions 

## Related issues
Addresses [86829](https://microsofthealth.visualstudio.com/Health/_workitems/edit/86829)

## Testing
Testing was manually done since I had to inject a SQL script into the stored procedure being tested. For example, during local debugging I set a breakpoint at a point where the SQLCommand.CommandText was being set, then I revised that text by adding in this specific SQL, "WAITFOR DELAY '00:00:35';". The reason for that is the default SQLCommand.CommandTimeout is 30 seconds and given the script that I added, it will force the current SQL statement to take an additional 35 seconds thereby causing a SQL timeout.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Api/Microsoft.Health.Fhir.Api.csproj
+++ b/src/Microsoft.Health.Fhir.Api/Microsoft.Health.Fhir.Api.csproj
@@ -26,7 +26,6 @@
     <PackageReference Include="Microsoft.Health.Abstractions" Version="$(HealthcareSharedPackageVersion)" />
     <PackageReference Include="Microsoft.Health.Api" Version="$(HealthcareSharedPackageVersion)" />
     <PackageReference Include="Microsoft.Health.Extensions.DependencyInjection" Version="$(HealthcareSharedPackageVersion)" />
-    <PackageReference Include="Microsoft.Health.SqlServer" Version="$(HealthcareSharedPackageVersion)" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.Health.Fhir.Core\Microsoft.Health.Fhir.Core.csproj" />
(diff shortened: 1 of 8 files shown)


LTA-Thinking:
It looks like one integration test no longer works:
```
Microsoft.Health.Fhir.Tests.Integration.Persistence.FhirStorageTests(SqlServer).GivenANonexistentResource_WhenUpsertingWithCreateDisabled_ThenAMethodNotAllowedExceptionIsThrown
Assert.Throws() Failure
Expected: typeof(Microsoft.Health.Fhir.Core.Exceptions.MethodNotAllowedException)
Actual:   typeof(Microsoft.Data.SqlClient.SqlException): Resource does not exist and create is not allowed
---- Microsoft.Data.SqlClient.SqlException : Resource does not exist and create is not allowed
```

v-mserdar:
> It looks like one integration test no longer works:
> 
> ```
> Microsoft.Health.Fhir.Tests.Integration.Persistence.FhirStorageTests(SqlServer).GivenANonexistentResource_WhenUpsertingWithCreateDisabled_ThenAMethodNotAllowedExceptionIsThrown
> Assert.Throws() Failure
> Expected: typeof(Microsoft.Health.Fhir.Core.Exceptions.MethodNotAllowedException)
> Actual:   typeof(Microsoft.Data.SqlClient.SqlException): Resource does not exist and create is not allowed
> ---- Microsoft.Data.SqlClient.SqlException : Resource does not exist and create is not allowed
> ```

Yes, I've been working with Brendan trying to figure out how we can get our tests to work correctly and catch this exception.

PR #2513, opened 2022-03-11
Handle search counts greater than int.MaxValue for sql and cosmos

v-mserdar:
Changes support returning a count value that exceeds an int32 from both sql and cosmos

## Description
On large databases (with > 2.3B resources) this query Get {{fhir}}?_summary=count results in the error "Arithmetic overflow error converting expression to data type int." This has been verified as an issue for both sql and cosmos. 
Investigating this issue shows that the underlying bundle object has a property called Total which is what the resulting count gets assigned to. The problem is that property is of type int and since it's owned by the hl7.org spec we're unable to fully account for the root issue. What we can do is gracefully handle it at this point.

For reference:
http://hl7.org/fhir/R4/bundle-definitions.html#Bundle.total
http://hl7.org/fhir/R4/datatypes.html#unsignedInt
TLDR: their specification for the size of an unsignedInt is "Any non-negative integer in the range 0..2,147,483,647"

## Related issues
Addresses [89072](https://microsofthealth.visualstudio.com/Health/_workitems/edit/89072)

## Testing
Since my local resources don't have > 2.3B resources this is how I tested each data source.
I used our rest http files and after getting an access token, I made a GET request like so to make sure we are doing a search for count.
GET {{hostname}}/?_summary=count 
Authorization: Bearer {{bearer.response.body.access_token}}

SQL: I temporarily revised the SqlCommand.CommandText to "SELECT CAST(999999999999 AS bigint)" for this scenario to validate the sql change.
Cosmos: I temporarily revised the QueryBuilder.BuildSqlQuerySpec() so that it only returns "SELECT value 99999999999999 " to validate the cosmos change.

## FHIR Team Checklist

Refs [AB#89072](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/89072)
+semver: skip

### DIFF
--- a/src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
+++ b/src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
@@ -542,6 +542,12 @@ namespace Microsoft.Health.Fhir.CosmosDb.Features.Search
             long count = (await _fhirDataStore.ExecuteDocumentQueryAsync<long>(sqlQuerySpec, feedOptions, continuationToken: null, cancellationToken: cancellationToken)).results.Single();
             if (count > int.MaxValue)
             {
+                _requestContextAccessor.RequestContext.BundleIssues.Add(
+                    new OperationOutcomeIssue(
+                        OperationOutcomeConstants.IssueSeverity.Error,
+                        OperationOutcomeConstants.IssueType.NotSupported,
+                        string.Format(Core.Resources.SearchCountResultsExceedLimit, count, int.MaxValue)));
+
                 throw new InvalidSearchOperationException(string.Format(Core.Resources.SearchCountResultsExceedLimit, count, int.MaxValue));
             }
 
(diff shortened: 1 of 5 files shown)


v-mserdar:
The test output while calling the following endpoint is shown below
GET {{hostname}}/?_summary=count
Authorization: Bearer {{bearer.response.body.access_token}}
image

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
```diff
@@ -539,7 +539,19 @@ private bool IsQuerySelective(QueryPartitionStatistics queryPartitionStatistics)
                 MaxConcurrency = _cosmosConfig.ParallelQueryOptions.MaxQueryConcurrency, // execute counts across all partitions
             };
 
-            return (await _fhirDataStore.ExecuteDocumentQueryAsync<int>(sqlQuerySpec, feedOptions, continuationToken: null, cancellationToken: cancellationToken)).results.Single();
+            long count = (await _fhirDataStore.ExecuteDocumentQueryAsync<long>(sqlQuerySpec, feedOptions, continuationToken: null, cancellationToken: cancellationToken)).results.Single();
```
LTA-Thinking, line 542: This made me worry if we needed a try catch here in case of an overflow on the long, but looking up that max value if a customer has more the 9 quadrillion resources in a database we have other issues.

SergeyGaluzo:
If true count is 5B what will be return in error?

---
In reply to: [1068160340](https://github.com/microsoft/fhir-server/pull/2513#issuecomment-1068160340) [](http://example.com/codeflow?ancestors=1068160340,1068160340)

v-mserdar:
Earlier I added a screen shot to this thread that shows what would be returned when the count exceeds that of int.MaxValue. Hopefully that helps answer your concerns.

PR #2514, opened 2022-03-14
Prevent no impact updates from creating version

apurvabhaleMS:
## Description
If a user updates a record that already exists and nothing changes in the content, the user should get a 200 but nothing should update the record. With this PR we are only covering Cosmos DB changes. (SQL changes might be a little bit more involved so will be covered in different PR). Updated UpsertAsync method to check the existing resource rawData with new resource rawData by ignoring meta.versionId and meta.lastUpdated. If its an absolute match then we simply return Ok with existing resource information without updating VersionId and lastUpdated. If the strings do not match then we proceed with further steps of creating a new version.

We are forming versionId and lastUpdated strings to use String.Replace on raw resource data. Forming lastUpdated with a valid date format is little tricky. We set the value of lastUpdated by using ToString("o") which formats the datetimeoffset to Formats to **2022-03-09T01:40:52.0690000+02:00**. [String serializer](https://github.com/microsoft/fhir-server/blob/main/src/Microsoft.Health.Fhir.Shared.Core/Features/Persistence/RawResourceFactory.cs#:~:text=return%20new%20RawResource(_fhirJsonSerializer.SerializeToString(poco)%2C%20FhirResourceFormat.Json%2C%20keepMeta)%3B) converts this to **2022-03-09T01:40:52.069+02:00** and saves it.

In C#, millisecond vlaue can range from [0 to 999]. Below are some useful examples of how date is stored in rawResource data in DB.
0000000+ -> +, 0010000+ -> 001+, 0100000+ -> 01+, 0180000+ -> 018+, 1000000 -> 1+, 1100000+ -> 11+, 1010000+ -> 101+

## Related issues
Addresses [[88658](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Olympus/Stories/?workitem=88658)].

## Testing
Added new tests and Existing tests passed.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
+++ b/src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
@@ -10,6 +10,7 @@ using System.IO;
 using System.Linq;
 using System.Net;
 using System.Text;
+using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 using EnsureThat;
@@ -34,7 +35,6 @@ using Microsoft.Health.Fhir.CosmosDb.Features.Storage.StoredProcedures.HardDelet
 using Microsoft.Health.Fhir.CosmosDb.Features.Storage.StoredProcedures.Replace;
 using Microsoft.Health.Fhir.ValueSets;
 using Microsoft.IO;
-using Newtonsoft.Json.Linq;
 using Polly;
 using Task = System.Threading.Tasks.Task;
 
@@ -210,11 +210,11 @@ namespace Microsoft.Health.Fhir.CosmosDb.Features.Storage
                     return null;
                 }
 
-                // If not delete then check if its an update with no data change
+                // If not a delete then check if its an update with no data change
                 if (!cosmosWrapper.IsDeleted)
                 {
                     // check if the new resource data is same as existing resource data
-                    if (string.Equals(RemoveVersionIdAndLastUpdatedFromMeta(existingItemResource.RawResource.Data), RemoveVersionIdAndLastUpdatedFromMeta(cosmosWrapper.RawResource.Data), StringComparison.Ordinal))
+                    if (string.Equals(RemoveVersionIdAndLastUpdatedFromMeta(existingItemResource), RemoveVersionIdAndLastUpdatedFromMeta(cosmosWrapper), StringComparison.Ordinal))
                     {
                         // Do not store the duplicate data, for a update with no impact - returning existingItemResource as no updates
                         return new UpsertOutcome(existingItemResource, SaveOutcomeType.Updated);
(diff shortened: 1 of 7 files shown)


src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
```diff
@@ -543,6 +555,20 @@ private void UpdateSortIndex(FhirCosmosResourceWrapper cosmosWrapper)
             }
         }
 
+        private static string RemoveVersionIdAndLastUpdatedFromMeta(string rawResource)
+        {
+            // Removing versionId and lastUpdated from meta
+            // We are ignoring meta.versionId and meta.lastUpdated
+            // if other meta elements like Source/Tag/Profile/Security are updated then we will have to insert a new veriosn for a resource
+
+            JObject jObject = JObject.Parse(rawResource);
```
LTA-Thinking, line 564: Is it possible to remove the version id and last updated time without parsing the resource? If I remember correctly this isn't a very performant operation. We know the values of version and last updated from the top level properties, so it might be possible to do a find/replace to get rid of them.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
```diff
@@ -543,6 +555,28 @@ private void UpdateSortIndex(FhirCosmosResourceWrapper cosmosWrapper)
...
+        {
+            // 0000000+ -> +, 0010000+ -> 001+, 0100000+ -> 01+, 0180000+ -> 018+, 1000000 -> 1+, 1100000+ -> 11+, 1010000+ -> 101+
+            // ToString("o") - Formats to 2022-03-09T01:40:52.0690000+02:00 but serialized value to string in dB is 2022-03-09T01:40:52.069+02:00
+            var formattedDate = date.ToString("o", CultureInfo.InvariantCulture);
+            var milliseconds = formattedDate.Substring(20, 7); // get 0690000
+            var trimmedMilliseconds = milliseconds.TrimEnd('0'); // get 069
+            if (milliseconds.Equals("0000000", StringComparison.Ordinal))
+            {
+                // when date = 2022-03-09T01:40:52.0000000+02:00, value in dB is 2022-03-09T01:40:52+02:00, we need to replace the . after second
+                return formattedDate.Replace("." + milliseconds, trimmedMilliseconds, StringComparison.Ordinal);
```
LTA-Thinking, line 568: Instead of trimmedMilliseconds, could we explicitly use the empty string? I believe if we hit this return statement trimmedMilliseconds will always be the empty string.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/BatchTests.cs
```diff
@@ -65,6 +65,51 @@ public async Task GivenAValidBundle_WhenSubmittingABatch_ThenSuccessIsReturnedFo
...
+
+            Bundle resourceAfterPostingSameBundle = fhirResponseAfterPostingSameBundle.Resource;
+
+            Assert.Equal("201", resourceAfterPostingSameBundle.Entry[0].Response.Status);
+            Assert.Equal("200", resourceAfterPostingSameBundle.Entry[1].Response.Status);
+            Assert.Equal("200", resourceAfterPostingSameBundle.Entry[2].Response.Status);
+            Assert.Equal(resource.Entry[2].Resource.VersionId, resourceAfterPostingSameBundle.Entry[2].Resource.VersionId);
+            Assert.Equal("200", resourceAfterPostingSameBundle.Entry[3].Response.Status);
+            Assert.Equal(resource.Entry[3].Resource.VersionId, resourceAfterPostingSameBundle.Entry[3].Resource.VersionId);
+            Assert.Equal("200", resourceAfterPostingSameBundle.Entry[8].Response.Status);
```
LTA-Thinking, line 105: Could the check for the 8th entry be moved so the entries are checked in order?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/BatchTests.cs
```diff
@@ -65,6 +65,51 @@ public async Task GivenAValidBundle_WhenSubmittingABatch_ThenSuccessIsReturnedFo
...
+
+            // WhenSubmittingABatchTwiceWithNoDataChange_ThenServerShouldNotCreateAVersionSecondTimeAndSendOk
+            using FhirResponse<Bundle> fhirResponseAfterPostingSameBundle = await _client.PostBundleAsync(requestBundle);
+            Assert.NotNull(fhirResponseAfterPostingSameBundle);
+            Assert.Equal(HttpStatusCode.OK, fhirResponseAfterPostingSameBundle.StatusCode);
+
+            Bundle resourceAfterPostingSameBundle = fhirResponseAfterPostingSameBundle.Resource;
+
+            Assert.Equal("201", resourceAfterPostingSameBundle.Entry[0].Response.Status);
+            Assert.Equal("200", resourceAfterPostingSameBundle.Entry[1].Response.Status);
```
LTA-Thinking, line 100: Should entry 1's version id also be checked? It looks like it also wasn't changed.

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/FhirStorageTests.cs
```diff
@@ -721,6 +732,41 @@ public async Task GivenADeletedResource_WhenUpdatingSearchParameterIndexAsync_Th
             }
         }
 
+        [Fact]
+        [FhirStorageTestsFixtureArgumentSets(DataStore.CosmosDb)]
+        public async Task GivenAValidResource_WhenUpdatingAResourceWithSameDataImmaterialKeepHistoryValue_ServerShouldNotCreateANewVersionAndReturnOk()
```
LTA-Thinking, line 737: I feel this test would work better as a theory with the keepHistory value as the input argument. That way the two are tested individually and the duplicate code isn't needed.

LTA-Thinking (dismissed):

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/BatchTests.cs
apurvabhaleMS: Entry 1, is a conditional create and next time there is a single match returning 200 ok with no resource information

feordin (approved):

PR #2519, opened 2022-03-22
Prevent no impact updates from creating version - sql

apurvabhaleMS:
## Description
If a user updates a record that already exists and nothing has changed in the content,  then the user should get a 200 but nothing should update the record. Updated UpsertAsync method to add validations in the code instead of in UpsertResource SP and check the existing resource rawData with new resource rawData by ignoring meta.versionId and meta.lastUpdated. If its an absolute match then we simply return Ok with existing resource information without updating VersionId and lastUpdated. If the strings do not match then we proceed with further steps of creating a new version.

Current login -
Inside a transaction in SP

- 	Read a row with a lock
- 	Validate
- 	Update/Create/Delete

Updated logic -
In code - 
- 	Loop to continue below transaction when the resource is updated by the time we reach the SP
- 	Read the latest resource info
- Validate
- 	Parse, Replace, and Compare resource content with old and new data 
- 	If match return OK with existing resource 
- 	If do not match then calls SP with compared version

-  In SP transaction
      - Read a row with a lock
      - Compare old with existing DB version
      -  If a match (in code comparison was with valid versions) then go ahead with version creation based on if resource Changed
      - VersionMatched = true -> Create new verson
      - If no match (somebody else updated or added new resource) then throw an error for client to retry

Notes -
- IsMetaSet
  - Added logic to check IsMetaSet flag while forming verisonId string (before comparing resource content), if it's false then the version id in rawResource is invalid
  - NewResource has VersionId=1 and IsMetaSet=true
  - SQL
      - Update requests - Saves IsMetaSet = false and does not update the raw Resource data - retaining versionId = 1
  - Cosmos 
    - JSON Update requests saves IsMetaSet = true and updates the rawResource only for JSON request data type versionId is set correctly
    - Other than JSON Update requests saves IsMetaSet = false and does not update the raw Resource for other request data types like XML - retaining versionId = 1

- Can this be possible ?
   - "meta": {                                                                                
        "lastUpdated": "2022-03-23T00:03:05.378+00:00",
        "versionId": "1"
    }  => "meta" : {,}

   - "meta":{"lastUpdated":"2022-03-23T00:03:05.378+00:00"}   => "meta" : {}
   - Above comparison will fail, but this should not happen. Generally we set lastUpdated and versionId both in a meta depending on the flag keepMeta which is set by canUpdateCreate (UpdateCreate from capability statement)
   		- Delete - Set as false
   		- Create - Hardcoded to true
		- Update - Depends on policy (canUpdateOrCreate) 
			  - True = all good
			  - False = Shouldn't be set to false and not toggled**

- We are forming versionId and lastUpdated strings to use String.Replace on raw resource data. Forming lastUpdated with a valid date format is little tricky. We set the value of lastUpdated by using ToString("o") which formats the datetimeoffset to Formats to 2022-03-09T01:40:52.0690000+02:00. [String serializer](https://github.com/microsoft/fhir-server/blob/main/src/Microsoft.Health.Fhir.Shared.Core/Features/Persistence/RawResourceFactory.cs#:~:text=return%20new%20RawResource(_fhirJsonSerializer.SerializeToString(poco)%2C%20FhirResourceFormat.Json%2C%20keepMeta)%3B) converts this to 2022-03-09T01:40:52.069+02:00 and saves it.

  In C#, millisecond value can range from [0 to 999]. Below are some useful examples of how date is stored in rawResource data in DB.
  0000000+ -> +, 0010000+ -> 001+, 0100000+ -> 01+, 0180000+ -> 018+, 1000000 -> 1+, 1100000+ -> 11+, 1010000+ -> 101+
## Related issues
Addresses [89485](https://microsofthealth.visualstudio.com/Health/_workitems/edit/89485)

## Testing
Manual testing
Updated existing tests
Additional tests to cover duplicate data do not creates a new version
If metadata has other properties which were changed during the update then new version should be created

## FHIR Team Checklist


src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/UpsertResource_7.sql
```diff
@@ -119,50 +125,39 @@ FROM   dbo.Resource WITH (UPDLOCK, HOLDLOCK)
...
-            BEGIN
-                -- You can't update a resource with a specified version if the resource does not exist
-				THROW 50404, 'Resource with specified version not found', 1;
-            END
-        IF (@allowCreate = 0)
-            BEGIN
-                THROW 50405, 'Resource does not exist and create is not allowed', 1;
-            END
+        -- There is no previous version
         SET @version = 1;
```
SergeyGaluzo, line 133: >SET @version = 1; [](http://example.com/codeflow?start=8&length=17)

@import sets initial version to 0. It is a discrepancy.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/UpsertResource_7.sql
```diff
@@ -100,7 +104,9 @@ CREATE PROCEDURE dbo.UpsertResource_6
     @tokenQuantityCompositeSearchParams dbo.BulkTokenQuantityCompositeSearchParamTableType_1 READONLY,
     @tokenStringCompositeSearchParams dbo.BulkTokenStringCompositeSearchParamTableType_1 READONLY,
     @tokenNumberNumberCompositeSearchParams dbo.BulkTokenNumberNumberCompositeSearchParamTableType_1 READONLY,
-    @isResourceChangeCaptureEnabled bit = 0
+    @isResourceChangeCaptureEnabled bit = 0,
+    @comparedVersion int = NULL,
+    @isResourceChange bit
```
SergeyGaluzo, line 109: >@isResourceChange bit [](http://example.com/codeflow?start=4&length=21)

Please don't call this stored procedure if code finds resources equal.
Please remove this parameter.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/UpsertResource_7.sql
```diff
@@ -119,50 +125,39 @@ FROM   dbo.Resource WITH (UPDLOCK, HOLDLOCK)
...
+                                COMMIT TRANSACTION;
+                                SELECT -1;
+                                RETURN;
+                            END
+                    END
+                ELSE
+                    BEGIN
+                        -- If not match means the version we compared in the code is not the latest version anymore
+                        -- Go back to code and compare the latest
+                        THROW 50409, 'Resource has been recently updated, please compare the resource content in code for any duplicate updates', 1;
```
SergeyGaluzo, line 157: >been recently updated [](http://example.com/codeflow?start=51&length=21)

Please add "or added"

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/UpsertResource_7.sql
apurvabhaleMS: Updated this with latest commit

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/UpsertResource_7.sql
apurvabhaleMS: Discussed this with Brendan, I will talk with the Import team to figure out if it's possible to start the version from 1 instead of 0

SergeyGaluzo: Why is 1 required?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/UpsertResource_7.sql
```diff
@@ -119,50 +122,24 @@ FROM   dbo.Resource WITH (UPDLOCK, HOLDLOCK)
...
+        -- There is no previous version
         SET @version = 1;
     END
 ELSE
     BEGIN
         -- There is a previous version
-        IF (@requireETagOnUpdate = 1
-            AND @etag IS NULL)
+        -- When not a delete and if @comparedVersion deosn't matche the @previousVersion in the DB
+        IF (@isDeleted = 0 AND @comparedVersion <> @previousVersion) -- When not a delete
```
SergeyGaluzo, line 136: >@comparedVersion <> @previousVersi [](http://example.com/codeflow?start=31&length=34)

:-) This is not C#
DECLARE @comparedVersion int = NULL, @previousVersion int = 1
SELECT CASE WHEN @comparedVersion <> @previousVersion THEN 1 ELSE 0 END
0

apurvabhaleMS: Aha, you are right how about [this](https://github.com/microsoft/fhir-server/pull/2519/commits/ab568485d6a453ba96670b94fe3935ef848f76f2)?

SergeyGaluzo: DECLARE @comparedVersion int = NULL, @previousVersion int = 1
SELECT CASE WHEN @comparedVersion IS NULL OR @comparedVersion <> @previousVersion THEN 1 ELSE 0 END
1

SergeyGaluzo (approved):
:shipit:

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/UpsertResource_7.sql
apurvabhaleMS: Discussed this with Tong, He is going to help make this consistent for import functionality. Small note - SQL always started with version 1 from the beginning. Cosmos also starts with version 1

feordin (approved):

LTA-Thinking (approved):

PR #2557, opened 2022-04-14
Check if Schema is initialized before executing SQL Tasks

apurvabhaleMS:
## Description
When schema is not initialized and the SQL task worker tries to access schema objects, it throws an error as Could not find Store procedure .....
With this PR, we want to prevent throwing unnecessary SQL errors by checking if the schema is initialized before continuing SQL task execution

## Related issues
Addresses [[86887](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Olympus/Stories/?workitem=86887)].

## Testing

- Existing test pass
- For Fhir.Web project, in the launchSettings.json file, set the "SqlServer:SchemaOptions: AutomaticUpdatesEnabled" as false. and run the Fhir server for SQL. With the old code, you will notice the error in the console as could not find the stored procedure... But with the latest changes, you should not see any error related to SQL tasks.

## FHIR Team Checklist


src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerTaskConsumer.cs
```diff
@@ -157,13 +161,16 @@ public async Task<TaskInfo> GetNextMessagesAsync(int taskHeartbeatTimeoutThresho
...
                     };
                 }
-            }
-            catch (SqlException e) when (e.Number == 2812)
-            {
-                _logger.LogWarning(e, "Schema is not initialized - {ex.Message}", e.Message);
-            }
+                catch (SqlException sqlEx)
+                {
+                    if (sqlEx.Number == SqlErrorCodes.NotFound)
```
tongwu-msft, line 166: >if (sqlEx.Number == SqlErrorCodes.NotFound) [](http://example.com/codeflow?start=20&length=43)

I would suspect it would not return notfound, which case would return notfound error?

apurvabhaleMS: I used the catch code from other tasks. I am assuming it catches exception when a resource doesn't exist. Also, if a particular schema version doesn't have the SP

tongwu-msft (approved):

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerTaskConsumer.cs
rbans96: I am still wondering if we need an explicit check for NotFound error here since in all other methods(CompleteAsync() or KeepAliveAsync()), I assume since we are reading the task with specific taskId and runId and if we could not find the task then we are throwing the TaskNotExistException but here we are just checking if the next message exists in the queue so probably we don't need a catch here. @tongwu-msft Please correct me if I have misunderstood anything here

rbans96 (approved):

PR #2563, opened 2022-04-20
Add logging to identify ArgumentException

v-mserdar:
## Description
We have a few customers that have been getting ArgumentExceptions when we have retrieved the RawResource from a Resource. In particular it appears that it's empty. This revision doesn't fix the issue but the goal of it is to provide additional logging so as to better understand what is happening around the time of the exception.

## Related issues
Addresses [90113](https://microsofthealth.visualstudio.com/Health/_workitems/edit/90113).

## Testing
Ran tests locally

## FHIR Team Checklist

Refs [AB#90113](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/90113)
+semver: patch

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
@@ -443,7 +443,7 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Storage
                             rawResource = await _compressedRawResourceConverter.ReadCompressedRawResource(rawResourceStream);
                         }
 
-                        _logger.LogInformation($"{nameof(resourceSurrogateId)}: {resourceSurrogateId}; {nameof(rawResource)} length: {rawResource.Length}");
+                        _logger.LogInformation($"{nameof(resourceSurrogateId)}: {resourceSurrogateId}; {nameof(key.ResourceType)}: {key.ResourceType}; {nameof(rawResource)} length: {rawResource.Length}");
 
                         var isRawResourceMetaSet = sqlDataReader.Read(resourceTable.IsRawResourceMetaSet, 5);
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -441,6 +443,8 @@ public async Task<ResourceWrapper> GetAsync(ResourceKey key, CancellationToken c
                             rawResource = await _compressedRawResourceConverter.ReadCompressedRawResource(rawResourceStream);
                         }
 
+                        _logger.LogInformation($"{nameof(resourceSurrogateId)}: {resourceSurrogateId}; {nameof(rawResource)} length: {rawResource.Length}");
```
LTA-Thinking, line 446: This looks similar to the log in the Search Service. Do you also want the resource type in this information?

v-mserdar: Good catch, I've added the ResourceType to the output.

LTA-Thinking (approved):

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -211,6 +211,8 @@ internal class SqlServerFhirDataStore : IFhirDataStore, IProvideCapability
 
                     stream.Seek(0, 0);
 
+                    _logger.LogInformation($"Upserting {resource.ResourceTypeName} with a stream length of {stream.Length}");
```
brendankowitz, line 214: We shouldn't use string interpolation with structured logging:
e.g.
```
_logger.LogInformation("Upserting {ResourceTypeName} with a stream length of {StreamLength}", resource.ResourceTypeName, stream.Length);
```

brendankowitz: Based on what we know now, this won't log the cases where rawResources are saved with 0 length. The length is ok,  its that the stream has already been read (so its at the end) on the subsequent retries.

PR #2565, opened 2022-04-22
SQL Compartment search refactoring

fhibf:
## Description
- Change Compartment Search to run on top of ReferenceSearchParam.
- Aggregate the execution of multiples CTEs in a single execution. 

## Testing

## FHIR Team Checklist


src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/UnionAllExpression.cs
```diff
@@ -0,0 +1,69 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using EnsureThat;
+
+namespace Microsoft.Health.Fhir.Core.Features.Search.Expressions
+{
+    /// <summary>
+    /// Represents a union all operator.
+    /// </summary>
+    public class UnionAllExpression : Expression, IExpressionsContainer
```
brendankowitz, line 16: UnionAll still sounds very specific for an expression, in most other cases here it would be `UnionExpression` with property `Type.All`. Or is this just a variant of `MultiaryExpression` where the MultiaryOperation is `Union`?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -392,19 +441,47 @@ private void HandleTableKindNormal(SearchParamTableExpression searchParamTableEx
...
         private void HandleTableKindAll(SearchParamTableExpression searchParamTableExpression)
         {
-            StringBuilder.Append("SELECT ")
-                .Append(VLatest.Resource.ResourceTypeId, null).Append(" AS T1, ")
-                .Append(VLatest.Resource.ResourceSurrogateId, null).AppendLine(" AS Sid1")
-                .Append("FROM ").AppendLine(VLatest.Resource);
+            int predecessorIndex = FindRestrictingPredecessorTableExpressionIndex();
 
-            using (var delimited = StringBuilder.BeginDelimitedWhereClause())
+            if (_unionAllVisited && predecessorIndex > 0 && searchParamTableExpression.ChainLevel == 0)
```
brendankowitz, line 446: If you have some insight to what this function does, please add some comments :) method level would be ok too

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -81,18 +82,25 @@ public override object VisitSqlRoot(SqlRootExpression expression, SearchOptions
                     throw new InvalidOperationException("Expected no predicates on the Resource table because of the presence of TableExpressions");
                 }
 
+                // Union All expressions must be executed first than all other expressions. The overral idea is that Union All expressions will
+                // filter the highest group of records, and the following expressions will be executed on top of this group of records.
```
brendankowitz, line 86: Is it possible to have something like `UnionAll (Condition1, Condition2(UnionAll(Condition2, Condition2b))`? i.e. if there it can be nested somewhere else lower in the tree

brendankowitz: I can't think of one since I think most sub expressions will probably my `IN` or `OR`, but if its not supported might make sense to throw an exception in that case

brendankowitz: More for future if we do come up with a scenario

brendankowitz (approved):
LGTM. As discussed I think its worth creating a separate story to consider merging UnionExpression and MultiaryExpression (with MultiaryOperationType.Union)

PR #2573, opened 2022-04-28
generic queue

SergeyGaluzo:
1. Adds JobQueue table and main stored procedures
2. Adds Exporter project to show how SQL works.
3. Adds SQL project (excluded from solution)
4. Shows schema differences in test.


src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Tables/Parameters.sql
```diff
@@ -0,0 +1,65 @@
...
+    ,Char        varchar(4000)    NULL
+    ,Binary      varbinary(max)   NULL
+    
+    ,UpdatedDate datetime         NULL
+    ,UpdatedBy   nvarchar(255)    NULL
+    
+     CONSTRAINT PKC_Parameters_Id PRIMARY KEY CLUSTERED (Id)
+  )
+GO
+CREATE TABLE dbo.ParametersHistory
```
rbans96, line 19: we might have to create a separate sql for this table(ParametersHistory.sql)

SergeyGaluzo: I will do it only if snapshot sql file is generated incorrectly. I don't see it happening yet.

LTA-Thinking (approved):

LTA-Thinking (approved):

PR #2574, opened 2022-04-28
Migrate to .NET 6.0

apurvabhaleMS:
## Description
With this PR, we are migrating FHIR projects to .Net 6
- Update global.json
- Update target frameworks
- Address new analyzer concerns
  1. Updated Microsoft.CodeAnalysis.NetAnalyzers to 6.0.0 version
  2. Remove log messages from resource files
  3. Add exceptions to .editorconfig
  4. Add _SkipUpgradeNetAnalyzersNuGetWarning in Build.Props
- AllowSynchronousIO - By default set to false with latest .Net core. We can set Sync option as true or upgrade all libraries and perform all actions asynchronously. Since we are making Async call may be ok to remove
- Append($"{TableExpressionName(_tableExpressionCounter)}.SortValue ") results in error (due to how string interpolcation works in .Net6) so rearranged 
- Updated Microsoft.SqlServer.DACFx - Integration tests were failing with old nuget (comparision of two schemas retrns empty object)

Error that we see when Microsoft.CodeAnalysis.NetAnalyzers is updated to 6.0.0 version -

- CA1727: Use PascalCase for named placeholders - https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1727 
- CA1848: Use the LoggerMessage delegates https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1848 
- CA2254: Template should be a static expression https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2254 
- CA1849: Call async methods when in an async method https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1849 
- SYSLIB0028: X509Certificate2.PrivateKey is obsolete https://docs.microsoft.com/en-us/dotnet/fundamentals/syslib-diagnostics/syslib0028 
- SYSLIB0021: Derived cryptographic types are obsolete https://docs.microsoft.com/en-us/dotnet/fundamentals/syslib-diagnostics/syslib0021 

## Related issues
Addresses [[AB87195](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Olympus/Stories/?workitem=87195)].

## Testing
Test pass

## FHIR Team Checklist


fhibf (dismissed):

kzhen:
Hi @apurvabhaleMS @fhibf any update on when this is going to be merged into `main`?

brendankowitz:
Hi @kzhen, we need to align and verify this update with a downstream system, so hopefully soon

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
```diff
@@ -600,7 +600,7 @@ public void Build(ICapabilityStatementBuilder builder)
             }
             catch (CosmosException ex)
             {
-                _logger.LogWarning("Failed to obtain provisioned RU throughput. Error: {0}", ex.Message);
+                _logger.LogWarning("Failed to obtain provisioned RU throughput. Error: {MEssage}", ex.Message);
```
PTaladay, line 603: nit: Message is misspelled.

src/Microsoft.Health.Fhir.Azure/AnonymizationConfigurationArtifactProvider.cs
```diff
@@ -165,7 +165,7 @@ private async Task FetchConfigurationFromBlob(string configName, string eTag, St
             BlobClient blob = container.GetBlobClient(configName);
             if (await blob.ExistsAsync(cancellationToken))
             {
-                if (CheckConfigurationIsTooLarge(blob.GetProperties(cancellationToken: cancellationToken).Value.ContentLength))
+                if (CheckConfigurationIsTooLarge(size: blob.GetPropertiesAsync(cancellationToken: cancellationToken).GetAwaiter().GetResult().Value.ContentLength))
```
PTaladay, line 168: Is there a reason why you chose to use GetAwaiter instead of using await? If it is an issue with using it in the method call, then I would just create a temp variable above that if statement and make the async call there. If you stick with the GetAwaiter then there is the potential for deadlocks as it isn't asynchronous.

PTaladay (changes requested):
Overall looks good. Just take a look at the comment about using GetAwaiter and address that before checking in.

PR #2576, opened 2022-04-29
Enforce max string value for SQL columns

feordin:
## Description
This updates healthcare-shared-component which pulls in a change to enforce max string length when inserting values into columns.  Previously the string would be silently truncated.

## Related issues
Addresses [issue [AB#91346](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/91346)].

## Testing
Unit tests added in 

## FHIR Team Checklist

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -26,7 +26,7 @@
   <PropertyGroup>
     <RuntimePackageVersion>5.0.0</RuntimePackageVersion>
     <AspNetPackageVersion>5.0.15</AspNetPackageVersion>
-    <HealthcareSharedPackageVersion>6.1.10</HealthcareSharedPackageVersion>
+    <HealthcareSharedPackageVersion>6.1.21</HealthcareSharedPackageVersion>
     <Hl7FhirVersion>3.8.2</Hl7FhirVersion>
   </PropertyGroup>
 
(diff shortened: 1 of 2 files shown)


test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/CreateTests.cs
```diff
@@ -263,6 +263,19 @@ public async Task GivenAResource_WhenPostingToHttpWithMaliciousNarrative_TheServ
             }
         }
 
+        [Fact]
+        [HttpIntegrationFixtureArgumentSets(DataStore.SqlServer, Format.All)]
```
LTA-Thinking, line 267: Should we have a test to show what happens when a Reindex is run on a database with a value for the new search parameter that is too long?

Directory.Build.props
```diff
@@ -26,7 +26,7 @@
   <PropertyGroup>
     <RuntimePackageVersion>5.0.0</RuntimePackageVersion>
     <AspNetPackageVersion>5.0.15</AspNetPackageVersion>
-    <HealthcareSharedPackageVersion>6.1.10</HealthcareSharedPackageVersion>
+    <HealthcareSharedPackageVersion>6.1.22</HealthcareSharedPackageVersion>
```
rbans96, line 29: we need merge from main as main is already at version higher than this

rbans96, line 29: we need merge from main as main is already at version higher than this

PR #2596, opened 2022-05-10
Sql Search performance improvements 

LTA-Thinking:
## Description
Streamlines SQL queries for basic searches by removing redundant query parameters and speeding up decompression.

## Related issues
Addresses part of [User Story 90388](https://microsofthealth.visualstudio.com/Health/_workitems/edit/90388): Bulk Export Parallelization Improvements Part 1

## Testing
Existing unit and E2E tests

## FHIR Team Checklist

Patch

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlCommandSimplifier.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlCommandSimplifier.cs
@@ -1,0 +1,97 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.Text.RegularExpressions;
+using Microsoft.Data.SqlClient;
+using Microsoft.Extensions.Logging;
+using Microsoft.Health.SqlServer;
+
+namespace Microsoft.Health.Fhir.SqlServer.Features.Search
+{
+    internal static class SqlCommandSimplifier
+    {
+        internal static void RemoveRedundantParameters(IndentedStringBuilder stringBuilder, SqlParameterCollection sqlParameterCollection, ILogger logger)
+        {
+            var commandText = stringBuilder.ToString();
+            if (commandText.Contains("cte", StringComparison.OrdinalIgnoreCase))
+            {
+                return;
+            }
+
+            try
+            {
+                commandText = commandText.Replace("DISTINCT ", string.Empty, StringComparison.OrdinalIgnoreCase);
+                if (!commandText.Contains(" OR ", StringComparison.OrdinalIgnoreCase))
+                {
+                    commandText = RemoveRedundantComparisons(commandText, sqlParameterCollection, '>');
+                    commandText = RemoveRedundantComparisons(commandText, sqlParameterCollection, '<');
+                }
+            }
+            catch (Exception ex)
+            {
+                // Something went wrong simplifing the query, return the query unchanged.
+                logger.LogWarning(ex, "Exception simplifing SQL query");
+                return;
+            }
+
+            stringBuilder.Clear();
+            stringBuilder.Append(commandText);
+        }
+
+        private static string RemoveRedundantComparisons(string commandText, SqlParameterCollection sqlParameterCollection, char operatorChar)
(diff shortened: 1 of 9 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -462,6 +465,75 @@ s.searchParameterInfo.Name switch
...
+            {
+                commandText = commandText.Replace("DISTINCT ", string.Empty, StringComparison.OrdinalIgnoreCase);
+                if (!commandText.Contains(" OR ", StringComparison.OrdinalIgnoreCase))
+                {
+                    commandText = RemoveRedundantComparisons(commandText, sqlParameterCollection, '>');
+                    commandText = RemoveRedundantComparisons(commandText, sqlParameterCollection, '<');
+                }
+            }
+
+            stringBuilder.Clear();
```
fhibf, line 481: As I see, the intention here is to remove DISTINCTs and ORs. 
Does it make sense to keep 'stringBuilder' as it is, and do not changed it? 

Are the changes in lines 481 and 482 required if there are 'ctes'?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -462,6 +465,75 @@ s.searchParameterInfo.Name switch
...
+            stringBuilder.Append(commandText);
+        }
+
+        private static string RemoveRedundantComparisons(string commandText, SqlParameterCollection sqlParameterCollection, char operatorChar)
```
fhibf, line 485: Everything looks good so far. 
My only concern: should we add a new unit test passing a query which contains the expression that must be removed, and then compare with the expected output? 

Maybe moving the logic of methods RemoveRedundantParameters and  RemoveRedundantComparisons to a static class can help us making those tests easier to be written.

fhibf (changes requested):

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
LTA-Thinking: No, that is a good point. If their are CTEs we can just return right away.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/CompressedRawResourceConverter.cs
```diff
@@ -23,15 +22,15 @@ public CompressedRawResourceConverter()
...
         {
-            await using var gzipStream = new GZipStream(compressedResourceStream, CompressionMode.Decompress, leaveOpen: true);
+            using var gzipStream = new GZipStream(compressedResourceStream, CompressionMode.Decompress, leaveOpen: true);
 
             // The current resource encoding uses byte-order marks. The legacy encoding does not, so we provide is as the fallback encoding
             // when there is no BOM
             using var reader = new StreamReader(gzipStream, LegacyResourceEncoding, detectEncodingFromByteOrderMarks: true);
 
-            return await reader.ReadToEndAsync();
+            return reader.ReadToEnd();
```
brendankowitz, line 33: Should we leave a comment for this for future reference?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
LTA-Thinking: The methods now have their own class, and there are unit tests for the simplification steps

fhibf (approved):
Great work.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlCommandSimplifier.cs
```diff
@@ -0,0 +1,97 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text.RegularExpressions;
+using Microsoft.Data.SqlClient;
+using Microsoft.Extensions.Logging;
+using Microsoft.Health.SqlServer;
+
+namespace Microsoft.Health.Fhir.SqlServer.Features.Search
+{
+    internal static class SqlCommandSimplifier
```
brendankowitz, line 15: Since we generate the sql all with our own code, I'm not sure I understand why we also string manipulate the string output?
Now we have Expressions => Writers => Sql Generator => Regex

PR #2620, opened 2022-05-23
Retain DateTimeOffset when parsing JsonPatchDocuments

### LINKED ISSUE #2614: JSON Patch Doesn't Respect Time Zones in Dates
**Describe the bug**
When using JSON Patch with FHIR Server, all input time zone information is dropped and converted to the local server configuration (UTC +00:00 on FHIR Service and API for FHIR). This is not the behavior for any operations with FHIR resources (like FHIR Patch).

**FHIR Version?**
All FHIR versions that support JSON Patch

**Data provider?**
Both

**To Reproduce**
1. PATCH the below payload on a resource and query

```http
PATCH /Patient/728414b1-eeb2-4264-9f6d-333ded7a407d 
Content-Type: application/json-patch+json
[
    {
        "op": "replace",   
        "path": "/deceasedDateTime",
        "value": "2022-05-02T14:00:00+02:00"
    }
]
```

```json
{
    "resourceType": "Patient",
    "id": "728414b1-eeb2-4264-9f6d-333ded7a407d",
    "meta": {
        "versionId": "3",
        "lastUpdated": "2022-05-02T15:07:20.33+00:00"
    },
    "active": false,
    "deceasedDateTime": "2022-05-02T12:00:00+00:00"
}
```

**Expected behavior**
Above request should return with:

```json
{
    "resourceType": "Patient",
    "id": "728414b1-eeb2-4264-9f6d-333ded7a407d",
    "meta": {
        "versionId": "3",
        "lastUpdated": "2022-05-02T15:07:20.33+00:00"
    },
    "active": false,
    "deceasedDateTime": "2022-05-02T14:00:00+02:00"
}
```
**NOTE**:
- Any fix for this should have regression testing to confirm proper behavior for FHIR and JSON payloads with timezones.

**Investigation Notes**
When investigating the bug, I found that the deserialization in the [FHIR Controller](https://github.com/microsoft/fhir-server/blob/main/src/Microsoft.Health.Fhir.Shared.Api/Controllers/FhirController.cs#L420) is incorrectly deserializing dates as local time and dropping any input time zone information. Specifically, `[FromBody] JsonPatchDocument patchDocument` results in a `JsonPatchDocument` without the timezone information.

I believe changing the deserialization settings for JSON payloads will be required. Two options:
- Add setting to the `AddNewtonsoftJson` config [here](https://github.com/microsoft/fhir-server/blob/main/src/Microsoft.Health.Fhir.Shared.Api/Registration/FhirServerServiceCollectionExtensions.cs#L59)
- Use the [JsonPatchInputFormatter](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.formatters.jsonpatchinputformatter?view=aspnetcore-2.2)??

aidanplunkitt:
## Description
The NewtonsoftJson parser's default setting is to deserialize DateTimeOffset values as just DateTimes, which strips UTC offset information. This PR would change that setting to retain the offset, as well as add an E2E test to ensure it's working as expected. (Note: Regression in DateTimeOffset handling in FhirPathPatch should be detected by [this existing case](https://github.com/microsoft/fhir-server/blob/main/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/FhirPathPatchTests.cs#L241).)

## Related issues
Fixes #2614

## Testing
Tested by manual inspection and addition of new E2E REST testcase

## FHIR Team Checklist

Patch (minor code change)


test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/JsonPatchTests.cs
```diff
@@ -301,5 +301,22 @@ public async Task GivenAnExistingResourceWithMismatchingData_WhenPatchingWithPat
...
+        [Fact]
+        [Trait(Traits.Priority, Priority.One)]
+        public async Task GivenAPatchDocumentWithDateTime_WhenDateTimeHasOffset_ThenOffsetShouldBePreserved()
+        {
+            var poco = Samples.GetJsonSample("PatientWithMinimalData").ToPoco<Patient>();
+            FhirResponse<Patient> response = await _client.CreateAsync(poco);
+
+            string dateTimeOffsetString = "2022-05-02T14:00:00+02:00";
+            string patchDocument =
+                $@"[{{""op"":""add"",""path"":""/deceasedDateTime"",""value"":""{dateTimeOffsetString}""}}]";
```
mikaelweave, line 314: Our coding standard for the repo is to not use string interpolation. Can you match the formatting to [this line](https://github.com/microsoft/fhir-server/blob/bb843d0f00d00dfdeb55b3c7d20629ec6f112bbc/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/JsonPatchTests.cs#L151)?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/JsonPatchTests.cs
```diff
@@ -301,5 +301,22 @@ public async Task GivenAnExistingResourceWithMismatchingData_WhenPatchingWithPat
...
+
+            FhirResponse<Patient> patchResponse = await _client.JsonPatchAsync(response.Resource, patchDocument);
+            Patient p = patchResponse.Resource;
+
+            Assert.Equal(dateTimeOffsetString, p.Deceased.ToString());
```
mikaelweave, line 319: I'm slightly worried `ToString` could swallow some issues in the future. Can you also check that `p.Deceased` contains the correct hour and timezone information?

mikaelweave (changes requested):
Nice, elegant find on this one. A couple comments on the tests.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/JsonPatchTests.cs
aidanplunkitt: Done!

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/JsonPatchTests.cs
aidanplunkitt: Good idea, I've added more robust test cases explicitly checking those fields. Let me know if you were thinking of a different approach though.

mikaelweave (dismissed):
Looks great! Thanks for the contribution.

mikaelweave:
@LTA-Thinking @brendankowitz  - this changes the default DateTime parting behavior for non-FHIR payloads. Are there any other non-FHIR REST operations besides JSON patch this could have an adverse effect on?

brendankowitz (approved):
Thank you @aidanplunkitt!

brendankowitz:
> this changes the default DateTime parting behavior for non-FHIR payloads. Are there any other non-FHIR REST operations besides JSON patch this could have an adverse effect on?

@mikaelweave I think this should be ok - it says in the docs - https://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_DateParseHandling.htm that is uses DateTimeOffset instead of DateTime for parsing. That sounds like it should be our preferred default anyway

mikaelweave:
> > this changes the default DateTime parting behavior for non-FHIR payloads. Are there any other non-FHIR REST operations besides JSON patch this could have an adverse effect on?
> 
> @mikaelweave I think this should be ok - it says in the docs - https://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_DateParseHandling.htm that is uses DateTimeOffset instead of DateTime for parsing. That sounds like it should be our preferred default anyway.
> 
> The only adverse situation I can think to check is if there are combinations of FHIR fields that are dates vs datetimes and if passing in Dates with/without offsets causes the model to be invalid

@brendankowitz - good point, didn't think about behavior on date only fields for JSONPatch.

@aidanplunkitt - do you mind adding another two E2E tests that checks the behavior (with and without offsets) to confirm that there are no parsing or data issues? [ActivityDefinition](https://www.hl7.org/fhir/activitydefinition.html) is a good test resource that has date elements without any required fields. The `approvalDate` element is present on the resource on STU3, R4, and R5 as well.

aidanplunkitt:
> @aidanplunkitt - do you mind adding another two E2E tests that checks the behavior (with and without offsets) to confirm that there are no parsing or data issues?
Sure! Happy to add these later this week.

aidanplunkitt:
@mikaelweave @brendankowitz added an extra couple of tests to JsonPatch and FhirPathPatch to regression test Date handling - please let me know if this is what you had in mind or if there is something else you'd like to see checks for!

mikaelweave (approved):
This looks perfect, thank you! Good to see we throw an error when you add time onto a date only field.

PR #2626, opened 2022-05-25
Remove unique index on Resource.ResourceSurrogateId

SergeyGaluzo:
## Description

## Related issues
Addresses [this bug](https://microsofthealth.visualstudio.com/Health/_queries/edit/90278/?triage=true)

## Testing

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/SqlImportOperation.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/SqlImportOperation.cs
@@ -85,7 +85,11 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Storage
         {
             var list = new List<(Table table, Index index)>();
 
-            if (_schemaInformation.Current >= SchemaVersionConstants.AddPrimaryKeyForResourceTable)
+            if (_schemaInformation.Current >= SchemaVersionConstants.RenamedIndexForResourceTable)
+            {
+                ////list.Add((VLatest.Resource, VLatest.Resource.IX_Resource_ResourceSurrogateId));
+            }
+            else if (_schemaInformation.Current >= SchemaVersionConstants.AddPrimaryKeyForResourceTable)
             {
                 list.Add((V25.Resource, V25.Resource.UQIX_Resource_ResourceSurrogateId));
             }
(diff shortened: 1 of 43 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/SqlImportOperation.cs
```diff
@@ -85,11 +85,7 @@ public IReadOnlyList<(Table table, Index index)> UniqueIndexesList()
         {
             var list = new List<(Table table, Index index)>();
 
-            if (_schemaInformation.Current >= SchemaVersionConstants.RenamedIndexForResourceTable)
-            {
-                list.Add((VLatest.Resource, VLatest.Resource.IX_Resource_ResourceSurrogateId));
-            }
```
brendankowitz, line 91: May still be true between versions `SchemaVersionConstants.RenamedIndexForResourceTable` and <34

SergeyGaluzo: It is overkill to support all permutations for functionality that nobody is using... I suggest let it go.

SergeyGaluzo: Added different skip. This will result in this index not being disabled, when it exists. Code will still work, and on final version it will perform.

src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/SqlImportOperation.cs
```diff
@@ -87,7 +87,7 @@ public IReadOnlyList<(Table table, Index index, bool pageCompression)> UniqueInd
 
             if (_schemaInformation.Current >= SchemaVersionConstants.RenamedIndexForResourceTable)
             {
-                list.Add((VLatest.Resource, VLatest.Resource.IX_Resource_ResourceSurrogateId, false));
+                ////list.Add((VLatest.Resource, VLatest.Resource.IX_Resource_ResourceSurrogateId));
             }
```
brendankowitz, line 91: I question is  why isn't the logic something like:
```csharp
  if (_schemaInformation.Current >= SchemaVersionConstants.RenamedIndexForResourceTable && _schemaInformation.Current < 34 /*SchemaVersionConstants.DroppedResourceSurrogateIX*/)
  {
      list.Add((VLatest.Resource, VLatest.Resource.IX_Resource_ResourceSurrogateId));
  }
```

SergeyGaluzo: Because we don't care about how this is handled. As long as it is not broken.

brendankowitz: Ah got it, all good

brendankowitz (approved):

PR #2646, opened 2022-06-06
Add blame to hanging tests

LTA-Thinking:
## Description
Adds a blame statement on test timeout.
Adds a 30 second timeout for all integration and e2e tests.

## Related issues
Addresses where integration tests are hanging for the full 1 hour test run.

## Testing
Existing integration and e2e tests

## FHIR Team Checklist

Skip (Test fixes)

### DIFF
--- a/build/jobs/e2e-tests.yml
+++ b/build/jobs/e2e-tests.yml
@@ -94,7 +94,7 @@ steps:
     displayName: 'E2E ${{ parameters.version }} ${{parameters.appServiceType}}'
     inputs:
       command: test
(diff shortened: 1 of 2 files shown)


rbans96:
Thanks RB for this change, looks good to me. One thing to consider in future since it will generate the dump if the host process is killed and dump might grow into gigabytes. We might include --blame-hang-dump-type as mini or none as by default its full. But for now 'full' makes sense as we want to find out what is causing integration tests to run over 60 mins.

rbans96 (approved):

PR #2688, opened 2022-07-05
FHIRPath Patch Functionality/Test Fixes

mikaelweave:
## Description
While testing FHIRPath Patch on a real-world scenario, I discovered several issues in both the functionality and tests. Essentially the issues I found are:
- Type checking for FHIRPath Patch were incorrect for nested datatypes (like `Patient.link`)
- Unit tests were leveraging `.Matches` on FHIR resources which doesn't compare all properties of an object
- Errors to the client are really vague and make it hard to understand why a patch operation is a bad request.

This PR fixes the above and makes the code more readable by moving operation specific logic into each of the FHIRPath Patch operations.

## Testing
- Fixed existing unit tests and confirmed all are passing
  - Added more detailed testing to confirm the error messages are being returned to the client.
- Upgraded E2E tests
- Manual testing

## FHIR Team Checklist


src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Resources/Patch/FhirPatchDeleteTests.cs
```diff
@@ -19,9 +21,9 @@ public void GivenAFhirPatchDeleteRequest_WhenDeletingPrimitive_ThenPrimitiveShou
         {
             var patchParam = new Parameters().AddDeletePatchParameter("Patient.birthDate");
 
-            Patient patchedPatientResource = (Patient)new FhirPathPatchBuilder(new Patient { BirthDate = "1920-01-01" }, patchParam).Apply();
+            Patient patchedPatientResource = new FhirPathPatchBuilder(new Patient { BirthDate = "1920-01-01" }, patchParam).Apply() as Patient;
 
-            Assert.True(patchedPatientResource.Matches(new Patient()));
+            Assert.Equal(patchedPatientResource.ToJson(), new Patient().ToJson());
```
feordin, line 26: Just curious as to the change from Matches to a string compare on the serialized Json.  Was the Matches method failing?

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Resources/Patch/FhirPatchBuilderTests.cs
```diff
@@ -17,31 +17,35 @@ public void GivenAFhirPatchParameterComponent_WhenOperationInvalid_ThenInvalidOp
         {
             var patchParam = new Parameters().AddPatchParameter("remove", path: "Patient.identifier");
 
-            Assert.Throws<InvalidOperationException>(() => new FhirPathPatchBuilder(new Patient(), patchParam));
+            var exception = Assert.Throws<InvalidOperationException>(() => new FhirPathPatchBuilder(new Patient(), patchParam));
```
feordin, line 20: This is a very complete comparison on the error message.  I wonder if it needs to be quite so complete?  If any part of the error message changes, even just whitespace, then the test would fail.  Perhaps we can assert only a part of the error?

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Patch/FhirPathPatch/Helpers/ElementModelExtensions.cs
```diff
@@ -29,14 +29,56 @@ internal static class ElementModelExtensions
...
+            {
+                var firstResultLocation = nodeList.First().Location;
+                var expectedLocation = firstResultLocation.Remove(firstResultLocation.LastIndexOf('.'));
+
+                // Multiple results are only allowed on collection types
+                if (nodeList.Any(l =>
+                    !l.Definition.IsCollection ||
+                    !l.Location.StartsWith(expectedLocation, StringComparison.InvariantCultureIgnoreCase)))
+                {
+                    throw new InvalidOperationException($"Multiple matches found");
```
feordin, line 65: I am not sure how the error message matches the check.  We fail if the definition is not a collection type, or if the location does not match the expected location.  But, we report an error of "Mulitple matches found"?

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Resources/Patch/FhirPatchDeleteTests.cs
mikaelweave: Matches is quite primitive - it doesn't compare all the elements. 👎🏽 So a lot of tests were testing nothing. 

https://github.com/FirelyTeam/firely-net-sdk/blob/ac40e368b8c09e266cb727edb89c9448b65e178c/src/Hl7.Fhir.Core/Model/Generated/Patient.cs#L245

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Resources/Patch/FhirPatchBuilderTests.cs
mikaelweave: Huh - literally have never thought of that. I don't have much experience in this error. Which of the below would be best?
- Using a string contains to look for the most important part of the error messages
- Creating an exception that is based off of `InvalidOperationException`. This would contain the path and the operation. This would be used to create the exception string. The test would check the fields used to build the string vs the string.

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Patch/FhirPathPatch/Helpers/ElementModelExtensions.cs
mikaelweave: The goal of this method is to provide a "builder like" pattern for the "Operation*.cs" classes which define the logic for each FHIRPath Patch operation. The "idea" was to make these classes more clear about what they are checking for instead of hiding everything in the extensions class. Example:

```
Target = ResourceElement
    .Select(Operation.Path)
    .CheckNoElements()
    .CheckMultipleElements()
    .GetFirstElementNode();
```

I'm feeling these could go in their own class called something like "ElementModelTestExtensions" or something similar. The whole `ElementModelExtensions` class has a bit of code smell (sorry) and could be broken up. I was attempting to minimize the changes in the PR, but happy to split it up if you think it improves the quality of the code base. 😓

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Resources/Patch/FhirPatchBuilderTests.cs
mikaelweave: @feordin I changed these to less restrictive "contains" 😄

feordin (approved):

PR #2727, opened 2022-08-03
Rebuild Metadata

LTA-Thinking:
## Description
Rebuild Metadata document every 4 hours to pull in changes to profiles and search parameters on different instances.

## Related issues
Addresses AB#93492: [Customer Impacting Issue][No Workaround] SupportedProfile was not showed up about 50% of the time when tried to get the metadata.

## Testing

## FHIR Team Checklist

Patch

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Conformance/SystemConformanceProvider.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Conformance/SystemConformanceProvider.cs
@@ -5,8 +5,10 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.Metrics;
 using System.Threading;
 using System.Threading.Tasks;
+using DotLiquid;
 using EnsureThat;
 using MediatR;
 using Microsoft.Extensions.Logging;
@@ -195,24 +197,28 @@ namespace Microsoft.Health.Fhir.Core.Features.Conformance
 
         public override async Task<ResourceElement> GetMetadata(CancellationToken cancellationToken = default)
         {
-            if (_metadata == null)
+            // There is a chance that the BackgroundLoop handler sets _metadata to null between when it is checked and returned, so the value is stored in a local variable.
+            ResourceElement metadata;
+            if ((metadata = _metadata) != null)
             {
-                _ = await GetCapabilityStatementOnStartup(cancellationToken);
-
-                await _metadataSemaphore.WaitAsync(cancellationToken);
-                try
-                {
-                    _metadata = _builder.Build().ToResourceElement();
-                    return _metadata;
-                }
-                finally
-                {
-                    _metadataSemaphore.Release();
-                }
+                return metadata;
             }
 
-            // There is a chance that the BackgroundLoop handler sets metadata to null between when it is checked and returned, but the alternative was leading to deadlocks where the creation of metadata could trigger a rebuild. The rebuild handler had to wait on the metadata semaphore, which wouldn't be released until the metadata could be built. But the metadata builder was waiting on the rebuild handler.
-            return _metadata;
+            _ = await GetCapabilityStatementOnStartup(cancellationToken);
+
+            // The semaphore is only used for building the metadata because claiming it before the GetCapabilityStatementOnStartup was leading to deadlocks where the creation
+            // of metadata could trigger a rebuild. The rebuild handler had to wait on the metadata semaphore, which wouldn't be released until the metadata could be built.
+            // But the metadata builder was waiting on the rebuild handler.
+            await _metadataSemaphore.WaitAsync(cancellationToken);
+            try
+            {
+                _metadata = _builder.Build().ToResourceElement();
+                return _metadata;
+            }
+            finally
+            {
+                _metadataSemaphore.Release();
+            }
         }
     }
 }
--- a/src/Microsoft.Health.Fhir.Core/Features/Conformance/SystemConformanceProvider.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Conformance/SystemConformanceProvider.cs
@@ -38,7 +38,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Conformance
         private static SemaphoreSlim _metadataSemaphore = new SemaphoreSlim(1, 1);
         private readonly List<Action<ListedCapabilityStatement>> _configurationUpdates = new List<Action<ListedCapabilityStatement>>();
         private readonly IOptions<CoreFeatureConfiguration> _configuration;
-        private readonly IKnowSupportedProfiles _supportedProfiles;
+        private readonly ISupportedProfilesStore _supportedProfiles;
         private readonly ILogger _logger;
 
         public SystemConformanceProvider(
@@ -46,7 +46,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Conformance
             ISearchParameterDefinitionManager.SearchableSearchParameterDefinitionManagerResolver searchParameterDefinitionManagerResolver,
             Func<IScoped<IEnumerable<IProvideCapability>>> capabilityProviders,
             IOptions<CoreFeatureConfiguration> configuration,
-            IKnowSupportedProfiles supportedProfiles,
+            ISupportedProfilesStore supportedProfiles,
             ILogger<SystemConformanceProvider> logger)
         {
             EnsureArg.IsNotNull(modelInfoProvider, nameof(modelInfoProvider));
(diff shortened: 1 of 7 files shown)


src/Microsoft.Health.Fhir.Core/Features/Conformance/SystemConformanceProvider.cs
```diff
@@ -86,6 +89,9 @@ public override async Task<ResourceElement> GetCapabilityStatementOnStartup(Canc
                         }
 
                         _listedCapabilityStatement = _builder.Build().ToResourceElement();
+
+                        _rebuilder = new Thread(RebuildCapabilityStatement);
+                        _rebuilder.Start();
```
brendankowitz, line 94: I like this pattern more:

```csharp
CancellationTokenSource _cancellationTokenSource = new CancellationTokenSource();

//in .ctor
_backgroundLoop = Task.Run(BackgroundLoop, CancellationToken.None);

private async Task BackgroundLoop()
{
      while (!_cancellationTokenSource.IsCancellationRequested)
      {
        // work
      }
}
```

HPaas has an example in ContainerManager

src/Microsoft.Health.Fhir.Core/Features/Conformance/SystemConformanceProvider.cs
```diff
@@ -25,14 +25,17 @@ namespace Microsoft.Health.Fhir.Core.Features.Conformance
     public sealed class SystemConformanceProvider
         : ConformanceProviderBase, IConfiguredConformanceProvider, INotificationHandler<RebuildCapabilityStatement>, IDisposable
     {
+        private readonly int _rebuildDelay = 4 * 60 * 60 * 1000; // 4 hours in milliseconds
```
brendankowitz, line 28: Perhaps?
```suggestion
        private readonly TimeSpan _rebuildDelay = TimeSpan.FromHours(4);
```
Then 
`_rebuildDelay.TotalMilliseconds;`

src/Microsoft.Health.Fhir.Core/Features/Validation/IKnowSupportedProfiles.cs
```diff
@@ -15,5 +15,7 @@ public interface IKnowSupportedProfiles
         /// <param name="resourceType">Resource type to get profiles.</param>
         /// <param name="disableCacheRefresh">Should we check server for new updates or get data out of cache.</param>
         IEnumerable<string> GetSupportedProfiles(string resourceType, bool disableCacheRefresh = false);
+
+        void Refresh();
```
brendankowitz, line 19: I know its not your change, but I don't like the name of this class :)

src/Microsoft.Health.Fhir.Core/Features/Conformance/SystemConformanceProvider.cs
```diff
@@ -116,30 +154,32 @@ public void Dispose()
...
+                        break;
+
+                    case RebuildPart.Profiles:
+                        // Update supported profiles;
+                        _builder.SyncProfiles(true);
+                        break;
                 }
+            }
 
+            await _metadataSemaphore.WaitAsync(cancellationToken);
```
brendankowitz, line 180: Not sure I understand why the semaphore shouldn't include fetching the data? This reads that many threads may fetch/sync, but must set to null one at a time. 
I feel like the point of this semaphore is more to stop multiple threads syncing at the same time...like a debounce, thoughts?

src/Microsoft.Health.Fhir.Core/Features/Conformance/SystemConformanceProvider.cs
```diff
@@ -98,6 +104,38 @@ public override async Task<ResourceElement> GetCapabilityStatementOnStartup(Canc
             return _listedCapabilityStatement;
         }
 
+        public async void RebuildCapabilityStatement()
+        {
+            while (_runRebuilder)
+            {
+                Thread.Sleep(_rebuildDelay);
```
v-mserdar, line 111: Should we use task delay instead here?
await Task.Delay(_rebuildDelay).ConfigureAwait(false);

src/Microsoft.Health.Fhir.Core/Features/Conformance/SystemConformanceProvider.cs
```diff
@@ -98,6 +104,38 @@ public override async Task<ResourceElement> GetCapabilityStatementOnStartup(Canc
             return _listedCapabilityStatement;
         }
 
+        public async void RebuildCapabilityStatement()
```
brendankowitz, line 107: We should avoid having `async void` methods

LTA-Thinking: Fixed

src/Microsoft.Health.Fhir.Core/Features/Conformance/SystemConformanceProvider.cs
LTA-Thinking: Replaced

src/Microsoft.Health.Fhir.Core/Features/Validation/IKnowSupportedProfiles.cs
LTA-Thinking: Changed

src/Microsoft.Health.Fhir.Core/Features/Conformance/SystemConformanceProvider.cs
LTA-Thinking: I changed the way this is checking so that when it is disposed it doesn't need to wait the full 4 hours. So this value is still an int, but the delay uses a TimeSpan.

src/Microsoft.Health.Fhir.Core/Features/Conformance/SystemConformanceProvider.cs
LTA-Thinking: Refactored to use this pattern.

src/Microsoft.Health.Fhir.Core/Features/Conformance/SystemConformanceProvider.cs
LTA-Thinking: The issue I was running into was that a deadlock situation was being hit. Because I added a refresh for the profiles to the CapabilityStatementBuilder it triggered a RebuildCapabilityStatement notification from the ServiceProvideProfileValidation. So when metadata was being refreshed and using the semaphore the Handle method also tried to acquire it. The metadata couldn't finish its flow until the notification was handled, and the Handle method couldn't complete until it got the semaphore. Only using the semaphore for when the metadata value was being set removed this issue.
It's not a perfect solution, I'm open to suggestions.

src/Microsoft.Health.Fhir.Core/Features/Conformance/SystemConformanceProvider.cs
```diff
@@ -98,6 +103,41 @@ public override async Task<ResourceElement> GetCapabilityStatementOnStartup(Canc
...
+        public async Task BackgroudLoop()
+        {
+            while (!_cancellationTokenSource.IsCancellationRequested)
+            {
+                for (int i = 0; i < _rebuildDelay; i++)
+                {
+                    await Task.Delay(TimeSpan.FromMinutes(1));
+
+                    if (!_cancellationTokenSource.IsCancellationRequested)
+                    {
```
rbans96, line 115: Should it be if cancellationRequested then return? Or I am missing something here

LTA-Thinking: Yes, thank you. Copy&paste error

src/Microsoft.Health.Fhir.Core/Features/Conformance/SystemConformanceProvider.cs
brendankowitz: Ok, I'm understanding the context a little more now, I'm ok with this

src/Microsoft.Health.Fhir.Core/Features/Conformance/SystemConformanceProvider.cs
```diff
@@ -150,22 +195,24 @@ public async Task Handle(RebuildCapabilityStatement notification, CancellationTo
...
                 }
-
-                return _metadata;
-            }
-            finally
-            {
-                _metadataSemaphore.Release();
             }
+
+            // There is a chance that the BackgroundLoop handler sets metadata to null between when it is checked and returned, but the alternative was leading to deadlocks where the creation of metadata could trigger a rebuild. The rebuild handler had to wait on the metadata semaphore, which wouldn't be released until the metadata could be built. But the metadata builder was waiting on the rebuild handler.
```
brendankowitz, line 214: A pattern here would be to capture the _metadata value in a local variable that gets returned. i.e. 
`if((metadata = _metadata) != null) { return metadata; }`

brendankowitz (approved):

PR #2731, opened 2022-08-08
Fixed issue where ResourceTypeOverride were not being set because of string casing issue

PTaladay:
Noticed an issue where the configuration values that were being sent from the portal were lowercase. The resource types in the FHIR Serivce code are Pascal casing. This resulted in there no longer being a match, so the override would not be set. I updated the dictionary these values get stored in to use the ordinalInvariantCase string comparer so that this is no longer an issue. 

## Description

## Related issues
Addresses [issue #].

## Testing
I updated the unit test we have for resource type override to use a lowercase config value, and validated the override is now set.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Conformance/ConformanceBuilderTests.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Conformance/ConformanceBuilderTests.cs
@@ -73,7 +73,7 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Conformance
             IOptions<CoreFeatureConfiguration> configuration = Substitute.For<IOptions<CoreFeatureConfiguration>>();
             Dictionary<string, string> overrides = new();
             VersioningConfiguration versionConfig = new();
-            versionConfig.ResourceTypeOverrides.Add("Patient", "no-version");
+            versionConfig.ResourceTypeOverrides.Add("patient", "no-version");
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Conformance/ConformanceBuilderTests.cs
```diff
@@ -73,7 +73,7 @@ public void GivenAConformanceBuilder_WhenVersionofResourceIsDifferentFromDefault
             IOptions<CoreFeatureConfiguration> configuration = Substitute.For<IOptions<CoreFeatureConfiguration>>();
             Dictionary<string, string> overrides = new();
             VersioningConfiguration versionConfig = new();
-            versionConfig.ResourceTypeOverrides.Add("Patient", "no-version");
+            versionConfig.ResourceTypeOverrides.Add("patient", "no-version");
```
brendankowitz, line 76: I think this should be a Theory and you should try upper/lower values

LTA-Thinking: I agree, we should also test for irregular casing like "pAtIeNt"

PTaladay: Thanks. I updated the test to a theory and added more test cases to validate.

LTA-Thinking (approved):

PR #2736, opened 2022-08-12
FHIR Schema Manager

apurvabhaleMS:
This PR adds three new projects to FHIR: 
1. SchemaManager
2. SchemaManager.UnitTests.
3. SchemaManager.Console

Schema Manager for FHIR differs from the healthcare-shared-components version by including Managed Identity Client ID and Authentication Type options and removing the required Server URI option. Also, there is only one command now: Apply.
This command uses FhirSchemaClient's GetDiffScriptAsync method which is grabbing the migration script files frommanifest instead of making a call to FHIR service.

## Related issues
Addresses [AB93732](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Olympus/Stories/?workitem=93732)

## Testing
1. Run the apply command on empty database without fhir service running. --> Should apply the schema based on version passed in command (No entry in Instance schema table - if there is no instance registered with the DB then it is safe to pursue the upgrade to the specified version.)
2. Setup a database with version x, keep fhir server running and run the apply command for x+5 --> schema should be upgraded to x+5
3. Unit tests pass
4. Setup a database with version x, Stop the fhir server and run the apply command for x+5 --> This fails right now as we have there is a dependency on InstanceSchema table which gets updated by SchemaJobWorker which will be not running as the fhir server is stopped.

## FHIR Team Checklist


docs/schema-manager.md
```diff
@@ -0,0 +1,85 @@
+# FHIR Schema Manager
+
+### What is it?
+Schema Manager is a command line app that upgrades the schema in your database from one version to the next through migration scripts.
```
rbans96, line 4: should we mention here SQL database
may be your database -> SQL Database?

docs/schema-manager.md
```diff
@@ -0,0 +1,85 @@
+# FHIR Schema Manager
```
rbans96, line 1: This doc is very useful :) Should we somehow provide link to this doc on [sql schema migration guide](https://github.com/microsoft/fhir-server/blob/main/docs/SchemaMigrationGuide.md)

rbans96 (approved):

PR #2739, opened 2022-08-13
collect code coverage

brendankowitz:
## Description
- Collects code coverage for Unit Tests (coverage not collected for Integration or E2E tests)
- Collects symbols as build artifacts

Notes:
Coverage stats will now show up on the overview page:
image Most of the core assemblies have reasonable coverage:
image

## Testing
Adds test coverage collection to unit tests.

## FHIR Team Checklist

Skip

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -36,6 +36,12 @@
         <IsPackable>false</IsPackable>
         <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)\CustomAnalysisRules.Test.ruleset</CodeAnalysisRuleSet>
       </PropertyGroup>
+      <ItemGroup>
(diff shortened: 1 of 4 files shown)


LTA-Thinking (dismissed):

LTA-Thinking:
It looks like we could fine tune the files this is looking at, for example to exclude the generated SQL files. This does look useful though, we should have a meeting to go through the results and see if we need more unit test coverage.

brendankowitz:
> It looks like we could fine tune the files this is looking at, for example to exclude the generated SQL files. This is very useful, we should have a meeting to go through the results and see if we need more unit test coverage.

Good idea, yes we should. Let me see if I can add this in for `*Generated*` which should exclude the generated sql files

LTA-Thinking (approved):

PR #2741, opened 2022-08-16
Added code to pass ProfileValidation header value to kick off profile validation in Bundles

rajithaalurims:
## Description
BundleHandler.cs is modified to pass x-ms-profile-validation in header if it is coming in the request.
Earlier x-ms-profile-validation was missing in header which caused bypassing of validation.

## Related issues
Addresses [AB87248](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Olympus/Stories/?workitem=87248)

## Testing
Tested bundle for Patient resource by passing x-ms-profile-validation as true with all mandatory values. Validation passed.
Tested bundle for Patient resource by passing x-ms-profile-validation as true with all gender missing. Validation failed s expected.

## FHIR Team Checklist


test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/BatchTests.cs
```diff
@@ -169,5 +170,47 @@ private void ValidateOperationOutcome(string actualStatusCode, OperationOutcome
...
         }
+
+        [Theory]
+        [InlineData("true", "201", true, true)]
+        [InlineData("true", "201", false, true)]
+        [InlineData("true", "201", false, false)]
+        [InlineData("true", "400", true, false)]
+        [InlineData("false", "201", true, false)]
+        [InlineData("false", "201", false, true)]
+        public async Task GivenABundle_WithProfileValidationFlag_ReturnsABundleResponse(string headerValue, string returnValue, bool profileValidation, bool withGender)
```
feordin, line 181: A suggestion as this is a bundle validation, would be to add more than one resource to the bundle.  Perhaps another resource which has no validation errors?  Even a third that has an error?  I wonder how it behaves with multiple resources in the bundle.  Do we see a list of validation errors associated with the different resources?

rajithaalurims: Agreed. I added multiple resources to bundle.
First one is missing gender
Second one is a valid Patient profile.
Third one is missing identifier.

feordin (approved):

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/BatchTests.cs
```diff
@@ -169,5 +170,70 @@ private void ValidateOperationOutcome(string actualStatusCode, OperationOutcome
...
+        [Theory]
+        [InlineData("true", "true")]
+        [InlineData("true", "false")]
+        [InlineData("false", "true")]
+        [InlineData("false", "false")]
+        public async Task GivenABundle_WithProfileValidationFlag_ReturnsABundleResponse(string headerValue, string profileValidation)
+        {
+            var bundle = new Hl7.Fhir.Model.Bundle
+            {
+                Type = Bundle.BundleType.Batch,
```
brendankowitz, line 183: We should replicate this test for SQL / Transaction bundles. Difference in behavior should be if one request fails the bundle will fail

rajithaalurims: I replicated and added similar test for Transaction Bundle.

LTA-Thinking (approved):

src/Microsoft.Health.Fhir.Shared.Client/FhirClient.cs
```diff
@@ -493,6 +493,22 @@ public async Task<FhirResponse<Bundle>> PostBundleAsync(Resource bundle, Cancell
             return await CreateResponseAsync<Bundle>(response);
         }
 
+        public async Task<FhirResponse<Bundle>> PostBundleAsyncWithValidationHeader(Resource bundle, string profileValidation, CancellationToken cancellationToken = default)
```
pallar-ms, line 496: Just curious - are we not intending to expose this via the interface IFhirClient?

pallar-ms, line 496: Can this instead be taken as bool arg and then converted to string internally?

rajithaalurims: This is used only in E2E tests where profileValidation is being passed as a string value. Is there any advantage of passing it as boolean? Having it as boolean type needs it to convert to boolean in the test method from where this method is being called and convert back into a string in this method.

pallar-ms: Thank you. This will enable client consumers to access this method using the interface rather than the actual underlying impl obj.

pallar-ms: Given this api is now going to be exposed to consumers using this client, IMO expecting them to pass a string literal instead of a bool is not ideal unless there are strong reasons to have the arg be string. It is more intuitive to pass a bool instead here as this is after all a flag.

brendankowitz: nit: this should be `PostBundleWithValidationHeaderAsync`

Issue #38, opened 2018-09-19
Documentation for RBAC


jackliums:
We should have documentation for how RBAC works.

Acceptance:
- Verify that there is a documentation describing how RBAC works.

PR #2794, opened 2022-09-06
Sort is crashing when Organization (any resource) is searched using chained search param and sort

rajithaalurims:
## Description
Ex:  https://ihealth-pie.fhir.azurehealthcareapis.com/Organization?partof.name=ABC&_sort=name throws Unknown error instead of returning results back.
As part of this [work item](https://microsofthealth.visualstudio.com/Health/_workitems/edit/80138), code was added in SortRewriter class to return null expression when sort field is already part of search filter name, which is causing failure in case of chained expression.
I modified code to use the expression as it is in chained expression part.

## Related issues
AB#94770

## Testing
Created test data and tested below URLs.
https://localhost:44348/Organization?partof.name=Test&_sort=name - should return results without failing
https://localhost:44348/Organization?name=Test&_sort=name - should return results as before
https://localhost:44348/Patient?organization.name=Mi&_sort=name - should return results without failing
https://localhost:44348/Patient?name=Mi&_sort=name - should return results as before

## FHIR Team Checklist


feordin (approved):

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/SortTests.cs
```diff
@@ -762,6 +762,34 @@ public async Task GivenPatientWithManagingOrg_WhenSearchedWithOrgIdentifierAndSo
                 patients.OrderByDescending(x => x.Name.Max(n => n.Family)).Cast<Resource>().ToArray());
         }
 
+        [SkippableFact]
```
fhibf, line 765: Should it be marked as SkippableFact?

src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/ExpressionRewriter.cs
```diff
@@ -38,7 +38,16 @@ public virtual Expression VisitChained(ChainedExpression expression, TContext co
                 return expression;
             }
 
-            return new ChainedExpression(resourceTypes: expression.ResourceTypes, referenceSearchParameter: expression.ReferenceSearchParameter, targetResourceTypes: expression.TargetResourceTypes, reversed: expression.Reversed, expression: visitedExpression);
+            /*SortRewriter will return null expression if it finds _sort param same as on of filter params.
```
fhibf, line 41: I think it's a typo: "as on**e** of"

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/SortTests.cs
```diff
@@ -762,6 +762,34 @@ public async Task GivenPatientWithManagingOrg_WhenSearchedWithOrgIdentifierAndSo
...
+
+            SetPatientInfo(patient, "Seattle", "Robinson", tag);
+            patients.Add((await Client.CreateAsync(patient)).Resource);
+
+            SetPatientInfo(patient, "Portland", "Williams", tag);
+            patients.Add((await Client.CreateAsync(patient)).Resource);
+
+            string query = $"organization.name=MSFT&_sort=name";
+            var response = await Client.SearchAsync(ResourceType.Patient, query);
+            Assert.NotNull(response);
```
fhibf, line 790: I would suggest to add more assert validations on this test, per example: 
- Is the response returning all expected types?
- Is the response status code 200? 
- Are the results properly sorted by name?

src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/ExpressionRewriter.cs
```diff
@@ -38,7 +38,16 @@ public virtual Expression VisitChained(ChainedExpression expression, TContext co
...
+             But in case of chained names we do need to keep it. #94770*/
```
fhibf, line 42: I would suggest removing the Bug ID #94770 from this comment. 
This Bug ID is only usable in our internal repository :)
From an open-source perspective, this ID is not required :)

src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/ExpressionRewriter.cs
fhibf, line 41: > SortRewriter will return null expression if it finds _sort param same as one of filter params. But in case of chained names we do need to keep it.

Is it true for all the scenarios? Can it cause any regression?
Is there any other scenario where the visitedExpression is null, and it must return a null expression?

rajithaalurims: look into ChainedExpression class line #38. It will return error if expression being passed is null. 
Only visitor other than SortRewriter that returns null is RemoveIncludesRewriter.  I have tested samples of requests where RemoveIncludesRewriter is called and it doesnt go through ChainedExpression. So from what I understand, this change shouldn't cause any regression.

src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/ExpressionRewriter.cs
rajithaalurims: done

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/SortTests.cs
rajithaalurims: o. Changed it to Fact.

src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/ExpressionRewriter.cs
rajithaalurims: Fixed

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/SortTests.cs
rajithaalurims: Added code to check all these. Also added another test case of orgs.

fhibf (approved):

PR #2812, opened 2022-09-20
[BugFix][HTTP][ASP.NET] Fix encoded query string parameters handling

fhibf:
## Description
Fix encoded query string parameters handling.
Customers mentioned that when submitting a request with an encoded URI, the query string parameters are not handled properly.

## Testing

When debugging this problem, I've noticed that the previous query string logic was handling the parameter name and the parameter value as a single thing, not as two different concepts (key / value) - as shown in the following picture:

image

After my changes, now FHIR handles to ASP.NET the responsibility of extracting these parameters from a URI. 

image

## FHIR Team Checklist


src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Extensions/HttpRequestExtensionsTests.cs
```diff
@@ -0,0 +1,261 @@
...
+                await writer.FlushAsync();
+                stream.Position = 0;
+                httpContext.Request.Body = stream;
+
+                return httpContext.Request;
+            }
+        }
+
+        /// <summary>
+        /// This methods runs the previous approach adopted to extract query string parameters from an URI.
```
rbans96, line 244: nit methods -> method

rbans96: Just for my understanding, are we planning to decommission this test compare previsous vs new approach once the new approach gets stable?

rbans96 (approved):

src/Microsoft.Health.Fhir.Api/Extensions/HttpRequestExtensions.cs
```diff
@@ -24,5 +27,36 @@ public static bool IsFhirRequest(this HttpRequest request)
...
+                NameValueCollection queryStringParameters =
+                    HttpUtility.ParseQueryString(
+                            request.QueryString.Value);
+
+                List<Tuple<string, string>> queries = new List<Tuple<string, string>>();
+                foreach (string key in queryStringParameters.AllKeys)
+                {
+                    foreach (var queryStringValue in queryStringParameters.GetValues(key))
+                    {
+                        queries.Add(Tuple.Create(key, queryStringValue));
```
brendankowitz, line 50: What was expected outcome of the duplicate query parameters `?name=a&name=b`?
Is it the asp.net default: name=a,b which would result in an OR. Or do we pass both `name` tuples through to the Expression parser, which I'm guessing will result in an AND.

fhibf: I believe that I'm following the old behavior and it'll continue being an **AND**. 

I added a test validating that: in HttpRequestExtensionsTests there is a test method named GivenAnUrlWithDuplicatedParameters_WhenEncoded_ThenCompareTheExpectedResults where I'm testing that behavior.

fhibf: Do you believe this is the expected behavior?

src/Microsoft.Health.Fhir.Api/Extensions/HttpRequestExtensions.cs
```diff
@@ -24,5 +27,36 @@ public static bool IsFhirRequest(this HttpRequest request)
...
+        {
+            EnsureArg.IsNotNull(request, nameof(request));
+
+            if (request.QueryString.HasValue)
+            {
+                NameValueCollection queryStringParameters =
+                    HttpUtility.ParseQueryString(
+                            request.QueryString.Value);
+
+                List<Tuple<string, string>> queries = new List<Tuple<string, string>>();
```
PTaladay, line 45: Just curious here, is there a reason to not just use a Dictionary<string> here instead of List<Tuple<string, string>>?

fhibf: That's a great question.
And there is a reason: If customer uses the query string "?name=FOO&name=BAR", then the final result should have (name, FOO) and (name, BAR), because that's an AND. 

If we adopt the Dictionary, then it would not be possible to handle this case, because the dictionary would not accept two keys with the same value.

PTaladay (dismissed):

brendankowitz (dismissed):
LGTM

PTaladay (approved):

PR #2828, opened 2022-09-26
Adding log output for ReIndex related tests

v-mserdar:
## Description
Revised the way that reindex calls are handled in these tests so that they all call in the same manner. Also adding output for the reindex job as there have been times when this will fail for undetermined reasons besides not completing in the original 20 second limit.
What used to occur was a test that would call to reindex may fail when it doesn't complete in time and then if there were subsequent test that tried to update create a search param they would fail with the message of "Changes to search parameters is not allowed while a reindex job is ongoing".

## Related issues
This is related to [#93223](https://microsofthealth.visualstudio.com/Health/_workitems/edit/93223) such that they're transient issues that do come up in this class.

## Testing
Successfully ran all Microsoft.Health.Fhir.Tests.E2E.Rest.Search.CustomSearchParamTests locally.

## FHIR Team Checklist

### DIFF
--- a/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/CustomSearchParamTests.cs
+++ b/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/CustomSearchParamTests.cs
@@ -5,6 +5,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Net;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/CustomSearchParamTests.cs
```diff
@@ -135,7 +116,7 @@ public async Task GivenANewSearchParam_WhenReindexingComplete_ThenResourcesSearc
             }
             catch (FhirException ex) when (ex.StatusCode == HttpStatusCode.BadRequest && ex.Message.Contains("not enabled"))
             {
-                Skip.If(!_fixture.IsUsingInProcTestServer, "Reindex is not enabled on this server.");
+                _output.WriteLine($"Skipping because reindex is disabled.");
```
fhibf, line 119: I like this approach 🙃😎😁

fhibf (approved):

feordin (approved):

PR #2829, opened 2022-09-27
Check SAS Token in Import

ccfromms:
## Description
In import, if url has token in query, it will throw exception when we validate resource. We plan to add check in importcontroller to check such situation.

## FHIR Team Checklist


src/Microsoft.Health.Fhir.Shared.Api/Controllers/ImportController.cs
```diff
@@ -201,7 +201,7 @@ private void ValidateImportRequestConfiguration(ImportRequest importData)
                     throw new RequestNotValidException(string.Format(Resources.UnsupportedResourceType, item.Type));
                 }
 
-                if (item.Url == null)
+                if (item.Url == null || !string.IsNullOrEmpty(item.Url.Query))
```
LTA-Thinking, line 204: Could you add a test for this scenario in ImportControllerTests?

ccfromms: Ok, I will add test in next commit.

LTA-Thinking (approved):

PR #2836, opened 2022-09-29
Read and parse SMART clinical scopes from the claims

feordin:
## Description
Read the SMART clinical scopes from the claims of user principal.  Parses the resources, user and data action from the scopes and adds them to the FhirRequestContext

## Related issues
Addresses [issue AB#95435].

## Testing
Unit tests were added.  E2E tests will be added once more of the overall scenario is complete.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Api/Features/SMART/SmartClinicalScopesMiddleware.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/SMART/SmartClinicalScopesMiddleware.cs
@@ -23,7 +23,7 @@ namespace Microsoft.Health.Fhir.Api.Features.Smart
         private readonly RequestDelegate _next;
 
         // Regex based on SMART on FHIR clinical scopes v1.0, http://hl7.org/fhir/smart-app-launch/1.0.0/scopes-and-launch-context/index.html#clinical-scope-syntax
-        private readonly Regex clinicalScopeRegEx = new Regex(@"(^|\s+)(?<id>patient|user)(/|\$|.)(?<resource>\*|([a-zA-Z]*))\.(?<accessLevel>read|write|\*)", RegexOptions.Compiled);
+        private static readonly Regex ClinicalScopeRegEx = new Regex(@"(^|\s+)(?<id>patient|user)(/|\$|.)(?<resource>\*|([a-zA-Z]*))\.(?<accessLevel>read|write|\*)", RegexOptions.Compiled);
 
         public SmartClinicalScopesMiddleware(RequestDelegate next)
         {
@@ -34,6 +34,9 @@ namespace Microsoft.Health.Fhir.Api.Features.Smart
 
         public async Task Invoke(HttpContext context, RequestContextAccessor<IFhirRequestContext> fhirRequestContextAccessor, AuthorizationConfiguration authorizationConfiguration)
         {
+            EnsureArg.IsNotNull(fhirRequestContextAccessor, nameof(fhirRequestContextAccessor));
+            EnsureArg.IsNotNull(authorizationConfiguration, nameof(authorizationConfiguration));
+
             if (fhirRequestContextAccessor.RequestContext.Principal != null
                 && authorizationConfiguration.Enabled)
             {
@@ -45,7 +48,7 @@ namespace Microsoft.Health.Fhir.Api.Features.Smart
                 DataActions permittedDataActions = 0;
                 foreach (Claim claim in principal.FindAll(authorizationConfiguration.ScopesClaim))
                 {
-                    var matches = clinicalScopeRegEx.Matches(claim.Value);
+                    var matches = ClinicalScopeRegEx.Matches(claim.Value);
                     foreach (Match match in matches)
                     {
                         fhirRequestContext.AccessControlContext.ClinicalScopes.Add(match.Value);
--- a/src/Microsoft.Health.Fhir.Api/Features/SMART/SmartClinicalScopesMiddleware.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/SMART/SmartClinicalScopesMiddleware.cs
@@ -1,0 +1,83 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Security.Claims;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+using EnsureThat;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Health.Core.Features.Context;
+using Microsoft.Health.Fhir.Core.Configs;
+using Microsoft.Health.Fhir.Core.Features.Context;
+using Microsoft.Health.Fhir.Core.Features.Security;
+
+namespace Microsoft.Health.Fhir.Api.Features.Smart
+{
+    /// <summary>
+    /// Middleware that runs after authentication middleware so the scopes field in the token can be examined for SMART on FHIR clinical scopes
+    /// </summary>
+    public class SmartClinicalScopesMiddleware
+    {
+        private readonly RequestDelegate _next;
+
+        // Regex based on SMART on FHIR clinical scopes v1.0, http://hl7.org/fhir/smart-app-launch/1.0.0/scopes-and-launch-context/index.html#clinical-scope-syntax
+        private readonly Regex clinicalScopeRegEx = new Regex(@"(^|\s+)(?<id>patient|user)(/|\$|.)(?<resource>\*|([a-zA-Z]*))\.(?<accessLevel>read|write|\*)", RegexOptions.Compiled);
+
+        public SmartClinicalScopesMiddleware(RequestDelegate next)
+        {
+            EnsureArg.IsNotNull(next, nameof(next));
+
+            _next = next;
+        }
+
+        public async Task Invoke(HttpContext context, RequestContextAccessor<IFhirRequestContext> fhirRequestContextAccessor, AuthorizationConfiguration authorizationConfiguration)
+        {
+            if (fhirRequestContextAccessor.RequestContext.Principal != null
+                && authorizationConfiguration.Enabled)
+            {
+                var fhirRequestContext = fhirRequestContextAccessor.RequestContext;
+                var principal = fhirRequestContext.Principal;
+                fhirRequestContext.AccessControlContext.ApplyFineGrainedAccessControl = true;
+
+                // examine the scopes claim for any SMART on FHIR clinical scopes
+                DataActions permittedDataActions = 0;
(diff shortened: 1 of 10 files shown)


src/Microsoft.Health.Fhir.Api/Features/SMART/SmartClinicalScopesMiddleware.cs
```diff
@@ -0,0 +1,83 @@
...
+                    var matches = clinicalScopeRegEx.Matches(claim.Value);
+                    foreach (Match match in matches)
+                    {
+                        fhirRequestContext.AccessControlContext.ClinicalScopes.Add(match.Value);
+
+                        var id = match.Groups["id"]?.Value;
+                        var resource = match.Groups["resource"]?.Value;
+                        var accessLevel = match.Groups["accessLevel"]?.Value;
+
+                        switch (accessLevel)
```
fhibf, line 57: Are the access levels always going to be lower case? 
What if the access level is none of the expected values (read, write and *)?

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/SMART/SmartClinicalScopesMiddlewareTests.cs
```diff
@@ -0,0 +1,78 @@
...
+            var expectedPrincipal = new ClaimsPrincipal(claimsIdentity);
+
+            httpContext.User = expectedPrincipal;
+            fhirRequestContext.Principal = expectedPrincipal;
+
+            await _smartClinicalScopesMiddleware.Invoke(httpContext, fhirRequestContextAccessor, authorizationConfiguration);
+
+            Assert.Equal(expectedScopeRestrictions, fhirRequestContext.AccessControlContext.AllowedResourceActions);
+        }
+
```
fhibf, line 58: I would add more validation scenarios, with invalid access levels, IDs, resources, etc.

src/Microsoft.Health.Fhir.Api/Features/SMART/SmartClinicalScopesMiddleware.cs
```diff
@@ -0,0 +1,83 @@
...
+{
+    /// <summary>
+    /// Middleware that runs after authentication middleware so the scopes field in the token can be examined for SMART on FHIR clinical scopes
+    /// </summary>
+    public class SmartClinicalScopesMiddleware
+    {
+        private readonly RequestDelegate _next;
+
+        // Regex based on SMART on FHIR clinical scopes v1.0, http://hl7.org/fhir/smart-app-launch/1.0.0/scopes-and-launch-context/index.html#clinical-scope-syntax
+        private readonly Regex clinicalScopeRegEx = new Regex(@"(^|\s+)(?<id>patient|user)(/|\$|.)(?<resource>\*|([a-zA-Z]*))\.(?<accessLevel>read|write|\*)", RegexOptions.Compiled);
```
fhibf, line 26: I would look for ways to add tests validating this RegEx. 
Maybe extracting it to another class, or including a static method that check if different variations of a claim are clinical scoped.

src/Microsoft.Health.Fhir.Core/Features/Context/AccessControlContext.cs
```diff
@@ -0,0 +1,27 @@
...
+        // Value indicated whether or not fine grained access control policies should be applied
+        public bool ApplyFineGrainedAccessControl { get; set; }
+
+        // the string values that were passed in as scopes
+        public ICollection<string> ClinicalScopes { get; } = new List<string>();
+
+        // A collection of the allowed resource types and which action is allowed on that type
+        public ICollection<ScopeRestriction> AllowedResourceActions { get; } = new List<ScopeRestriction>();
+
+        /// <summary>
```
fhibf, line 22: Should we adopt a standard where all comments in this class are between "<summary>" tags or regular "//"?

src/Microsoft.Health.Fhir.Core/Features/Context/ScopeRestriction.cs
```diff
@@ -0,0 +1,50 @@
...
+namespace Microsoft.Health.Fhir.Core.Features.Context
+{
+    public class ScopeRestriction : IEquatable<ScopeRestriction>
+    {
+        public ScopeRestriction(string resource, DataActions allowedAction, string user)
+        {
+            Resource = resource;
+            AllowedDataAction |= allowedAction;
+            User = user;
+        }
```
fhibf, line 18: Are null values allowed?

src/Microsoft.Health.Fhir.Api/Features/SMART/SmartClinicalScopesMiddleware.cs
erikhoward, line 26: This regex looks like it would support claims in the form of patient$Patient.read. We will be replacing "/" with "$" during app launch. Did not see a test using this form of the claim.

feordin: Yes, I was trying to be flexible with the separator character, as I was suspecting we would need something different than /.  I will add some tests with the $.

src/Microsoft.Health.Fhir.Api/Features/SMART/SmartClinicalScopesMiddleware.cs
feordin: The spec does specify the exact values for the access levels.  If an improper value is part of the scope, it will not match the regex and then no permissions will be granted.

src/Microsoft.Health.Fhir.Core/Features/Context/ScopeRestriction.cs
feordin: I added a check for null for Resource, but we may allow nulls for user.

src/Microsoft.Health.Fhir.Core/Features/Context/AccessControlContext.cs
feordin: updated.

src/Microsoft.Health.Fhir.Api/Features/SMART/SmartClinicalScopesMiddleware.cs
```diff
@@ -0,0 +1,83 @@
...
+        public SmartClinicalScopesMiddleware(RequestDelegate next)
+        {
+            EnsureArg.IsNotNull(next, nameof(next));
+
+            _next = next;
+        }
+
+        public async Task Invoke(HttpContext context, RequestContextAccessor<IFhirRequestContext> fhirRequestContextAccessor, AuthorizationConfiguration authorizationConfiguration)
+        {
+            if (fhirRequestContextAccessor.RequestContext.Principal != null
```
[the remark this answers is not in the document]
feordin: Added a null check.

src/Microsoft.Health.Fhir.Api/Features/SMART/SmartClinicalScopesMiddleware.cs
```diff
@@ -0,0 +1,83 @@
...
+        private readonly Regex clinicalScopeRegEx = new Regex(@"(^|\s+)(?<id>patient|user)(/|\$|.)(?<resource>\*|([a-zA-Z]*))\.(?<accessLevel>read|write|\*)", RegexOptions.Compiled);
+
+        public SmartClinicalScopesMiddleware(RequestDelegate next)
+        {
+            EnsureArg.IsNotNull(next, nameof(next));
+
+            _next = next;
+        }
+
+        public async Task Invoke(HttpContext context, RequestContextAccessor<IFhirRequestContext> fhirRequestContextAccessor, AuthorizationConfiguration authorizationConfiguration)
```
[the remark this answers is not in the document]
feordin: Added null checks.

src/Microsoft.Health.Fhir.Api/Features/SMART/SmartClinicalScopesMiddleware.cs
feordin: I will add more combinations in the unit tests, which will include verifying the regex.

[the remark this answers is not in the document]
feordin: I think you are correct, a static value will avoid the regex being recompiled each instantiation.

src/Microsoft.Health.Fhir.Core/Features/Context/AccessControlContext.cs
```diff
@@ -0,0 +1,33 @@
...
+        /// A collection of the allowed resource types and which action is allowed on that type
+        /// </summary>
+        public ICollection<ScopeRestriction> AllowedResourceActions { get; } = new List<ScopeRestriction>();
+
+        /// <summary>
+        /// A uri which points to a specific fhir resource in this Fhir server, and is associated with the current user
+        /// </summary>
+        public Uri FhirUserClaim { get; set; }
```
mikaelweave, line 31: The FhirUser claim *can* be a list of resources. Someone may be a `patient` but also a `person` to be referenced by other patients (like parent child relationship).

It also could be a URL to another FHIR instance, but I don't think we should worry about that since we would ignore that for access control decisions.

fhibf (approved):

src/Microsoft.Health.Fhir.Api/Features/SMART/SmartClinicalScopesMiddleware.cs
```diff
@@ -0,0 +1,86 @@
...
+{
+    /// <summary>
+    /// Middleware that runs after authentication middleware so the scopes field in the token can be examined for SMART on FHIR clinical scopes
+    /// </summary>
+    public class SmartClinicalScopesMiddleware
+    {
+        private readonly RequestDelegate _next;
+
+        // Regex based on SMART on FHIR clinical scopes v1.0, http://hl7.org/fhir/smart-app-launch/1.0.0/scopes-and-launch-context/index.html#clinical-scope-syntax
+        private static readonly Regex ClinicalScopeRegEx = new Regex(@"(^|\s+)(?<id>patient|user)(/|\$|.)(?<resource>\*|([a-zA-Z]*))\.(?<accessLevel>read|write|\*)", RegexOptions.Compiled);
```
mikaelweave, line 26: Looking forward, is this easily expandable to SMART v2 scopes? We may have to support both at the same time. Would we just run both Regex operations and union the output?

mikaelweave (commented):
A couple small thoughts on how this maps to the SMART spec.

src/Microsoft.Health.Fhir.Api/Features/SMART/SmartClinicalScopesMiddleware.cs
feordin: Yes, while this obviously only handles v1.0 scopes, we will be able to expand it to handle v2.0.  We may need more fine grained DataActions than we currently have.  So it is probably a bigger change than simply expanding the regex and adding the v2 scopes to the parsed result.

PR #2842, opened 2022-09-29
$member-match exception handling updates

v-mserdar:
## Description
$member-match was hiding exceptions by returning an incorrect "No match found." message for any uncaught exceptions.
The new changes will specifically handle the 2 scenarios that were discovered while investigating the bug. 

For SQL dbs it's possible to get a SQLException whereby the query processor has run out of internal resources and can't produce a query plan and thereby throws a sql exception -#8623 .

For CosmosDb it's an error where the subquery was hitting the 1000 results max limit. It will now correctly return the InvalidSearchOperationException with the appropriate message.

## Related issues
Addresses [#95860](https://microsofthealth.visualstudio.com/Health/_workitems/edit/95860) 
Addresses [#95861](https://microsofthealth.visualstudio.com/Health/_workitems/edit/95861) 

## Testing
Used Postman with a $member-match POST. The body of the POST used the json represented on this page here http://hl7.org/fhir/us/davinci-hrex/2020Sep/Parameters-member-match-in.json.html

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/MemberMatch/MemberMatchService.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/MemberMatch/MemberMatchService.cs
@@ -10,6 +10,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using EnsureThat;
 using Hl7.Fhir.Model;
+using Microsoft.Data.SqlClient;
 using Microsoft.Extensions.Logging;
 using Microsoft.Health.Extensions.DependencyInjection;
 using Microsoft.Health.Fhir.Core.Exceptions;
@@ -21,6 +22,7 @@ using Microsoft.Health.Fhir.Core.Features.Search.Expressions;
 using Microsoft.Health.Fhir.Core.Features.Search.Expressions.Parsers;
 using Microsoft.Health.Fhir.Core.Models;
 using Expression = Microsoft.Health.Fhir.Core.Features.Search.Expressions.Expression;
+using SortOrder = Microsoft.Health.Fhir.Core.Features.Search.SortOrder;
 
 namespace Microsoft.Health.Fhir.Core.Features.Operations.MemberMatch
 {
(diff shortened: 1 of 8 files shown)


src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/MemberMatch/MemberMatchService.cs
```diff
@@ -73,12 +75,20 @@ public async Task<ResourceElement> FindMatch(ResourceElement coverage, ResourceE
                 using IScoped<ISearchService> search = _searchServiceFactory();
                 results = await search.Value.SearchAsync(searchOptions, cancellationToken);
             }
-
-            // Generic catch for any exception happened during search.
+            catch (SqlException ex)
```
LTA-Thinking, line 78: Could we change this from a generic SQL exception to the new SqlQueryPlanException? That way we don't need to add the SQL library to Core. We are trying to keep database dependencies out of the non-database packages.

v-mserdar: OK, I'll go ahead and make that change.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlExceptionActionProcessor.cs
```diff
@@ -29,6 +31,10 @@ public Task Execute(TRequest request, TException exception, CancellationToken ca
                 {
                     throw new MethodNotAllowedException(Core.Resources.ResourceCreationNotAllowed);
                 }
+                else if (sqlException.Number == QueryProcessorNoQueryPlan)
```
LTA-Thinking, line 34: Can the QueryProcessorNoQueryPlan number be added to SqlErrorCodes?

v-mserdar: Yes it can. Doing so would just mean waiting for the shared health to get updated and fhir server to pull that in.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/MemberMatch/MemberMatchService.cs
LTA-Thinking: @brendankowitz What are your thoughts here? Matt looked into this and we can't get a specific exception at this point, the action process runs later. So the options are to have this be caught in the general exception catch, or take a dependency on SQL for the exception type. I don't like having it caught in the general exception as that means the user losses useful data.
I also know we have run into this issue before specifically with the SQL exception type.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlExceptionActionProcessor.cs
LTA-Thinking: Oh, that is in Shared Health. Right...
Lets keep it here then for now and make a PR in Shared Health to add that value. We can update it here latter when we pull the new package.

v-mserdar: I created a PR in shared health.

LTA-Thinking (approved):

PR #2846, opened 2022-10-03
Updated Bundle Handler to pass log data for sub calls in Bundle

PTaladay:
## Description
Changed from generic count of each HTTP status code to using BundleSubCallMetric class. This new class captures data needed for logging of the sub calls that happen from a batch or transaction Post bundle. This addresses part of what we need. The  BundleNotificationHandler on the PaaS side will be updated in another PR to accept this new information and log it appropriately. 

This expects the Bundle.Entry.Request for each bundle entry item. 

## Related issues
Addresses [#94094].

## Testing
Updated existing Unit Tests. Still pending testing with PaaS changes to validate data is passed through appropriately. Will run existing E2E and Integration tests as well to ensure no regressions.

## FHIR Team Checklist


src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -242,17 +242,26 @@ public async Task<BundleResponse> Handle(BundleRequest request, CancellationToke
...
         {
-            var apiCallResults = new Dictionary<string, int>();
+            var apiCallResults = new Dictionary<string, List<BundleSubCallMetricData>>();
             foreach (var entry in responseBundle.Entry)
             {
                 var status = entry.Response.Status;
                 if (apiCallResults.ContainsKey(status))
                 {
-                    apiCallResults[status]++;
+                    apiCallResults[status].Add(new BundleSubCallMetricData()
```
LTA-Thinking, line 251: Since the add is the same whether the key exists or not could it be moved out of the if statement to reduce duplication?

LTA-Thinking:
You mentioned that a separate PaaS PR will be needed. It looks like it will have to be the PR that updates the FHIR package version that contains this change. Is that correct?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
PTaladay: Thanks. I cleaned up the duplicate code.

LTA-Thinking (dismissed):

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -242,18 +242,20 @@ public async Task<BundleResponse> Handle(BundleRequest request, CancellationToke
...
-                    apiCallResults[status]++;
+                    apiCallResults[status] = new List<BundleSubCallMetricData>();
                 }
-                else
+
+                apiCallResults[status].Add(new BundleSubCallMetricData()
                 {
-                    apiCallResults[status] = 1;
-                }
+                    FhirOperation = string.Concat(entry?.Request?.Method, entry?.Request?.Url),
```
brendankowitz, line 256: How is this used, does this need a space in between?

rajithaalurims (approved):

PR #2858, opened 2022-10-06
[Tests] Improve FhirException details during pipeline executions

fhibf:
Improving FhirExcception details during pipeline executions: it's common to see "InternalServerError" errors during pipeline execution with no further details. These changes in the FhirExcception will provide more information, like the Uri being consumed, reason phrase, timestamp, etc. These changes will only affect Tests, and the FhirException **is not** used on the FHIR service.  
## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Client/FhirException.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Client/FhirException.cs
@@ -8,6 +8,7 @@ using System.Linq;
 using System.Net;
 using System.Net.Http;
 using System.Net.Http.Headers;
+using System.Text;
 using Hl7.Fhir.Model;
 
 namespace Microsoft.Health.Fhir.Client
@@ -29,7 +30,7 @@ namespace Microsoft.Health.Fhir.Client
 
         public OperationOutcome OperationOutcome => Response.Resource;
 
-        public override string Message => $"{StatusCode}: {OperationOutcome?.Issue?.FirstOrDefault()?.Diagnostics} ({GetOperationId()})";
+        public override string Message => FormatMessage();
 
         public void Dispose()
         {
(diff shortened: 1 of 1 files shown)


src/Microsoft.Health.Fhir.Shared.Client/FhirException.cs
```diff
@@ -46,14 +47,39 @@ protected virtual void Dispose(bool disposing)
...
+            message.Append(StatusCode);
+            if (!string.IsNullOrWhiteSpace(diagnostic))
+            {
+                message.Append(": ").Append(diagnostic);
+            }
+
+            message.Append(" (").Append(operationId).AppendLine(")");
+
+            message.AppendLine("==============================================");
+            message.Append("Url: (").Append(Response.Response.RequestMessage?.Method.Method ?? "NO_HTTP_METHOD_AVAILABLE").Append(") ").AppendLine(Response.Response.RequestMessage?.RequestUri.ToString() ?? "NO_URI_AVAILABLE");
```
LTA-Thinking, line 66: Do we know that the URI is going to have the query values censored out if retrieved this way?

fhibf: That's a good question. 
When I was writing down these changes and checked where FhirException is being used. 
Currently, it's only used in test projects. We do not use it on FHIR server side. 
For this reason, I was thinking that the data exposed in the URI would be only with data from our tests. 
That was my train of thought. 

But, if you think that's a risk, I can change this part of the code to expose only the domain. 
That would be good enough.

LTA-Thinking: Ah, ok. I had this confused with the FhirException class in Core: https://github.com/microsoft/fhir-server/blob/main/src/Microsoft.Health.Fhir.Core/Exceptions/FhirException.cs
We may want to think about having two unrelated classes with the same name.

LTA-Thinking (approved):

PR #2865, opened 2022-10-10
Export tests can extract zip

brendankowitz:
## Description
To save time uploading e2e test assets these directories were zipped. This fixes the Bulk Export E2E tests by unzipping them before execution.

## Testing
Run in pipeline: https://microsofthealthoss.visualstudio.com/FhirServer/_build/results?buildId=24782&view=results

## FHIR Team Checklist

### DIFF
--- a/build/jobs/e2e-tests-extract.yml
+++ b/build/jobs/e2e-tests-extract.yml
@@ -1,0 +1,10 @@
+parameters:
+- name: version
+  type: string
+
+steps:
+  - task: ExtractFiles@1
+    displayName: 'Extract E2E Test Binaries'
+    inputs:
+      archiveFilePatterns: '$(System.ArtifactsDirectory)/IntegrationTests/Microsoft.Health.Fhir.${{ parameters.version }}.Tests.E2E.zip'
+      destinationFolder: '$(Agent.TempDirectory)/E2ETests/'
--- a/build/jobs/run-export-tests.yml
+++ b/build/jobs/run-export-tests.yml
@@ -12,6 +12,9 @@ jobs:
     vmImage: '$(LinuxVmImage)'
   steps:
   - template: e2e-setup.yml
+  - template: e2e-tests-extract.yml
+    parameters:
+      version: ${{parameters.version}}
 
   - task: AzurePowerShell@4
     displayName: 'Set Variables'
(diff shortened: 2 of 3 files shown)


PTaladay (approved):

build/jobs/run-export-tests.yml
```diff
@@ -171,7 +177,7 @@ jobs:
     displayName: 'Export E2E ${{ parameters.version }} SQL'
     inputs:
       command: test
-      arguments: '"**\*${{ parameters.version }}.Tests.E2E*.dll" --filter "FullyQualifiedName~SqlServer&Category=ExportLongRunning"'
```
fhibf, line 174: Should it be an "\" (inverted slash) or the "/" (regular slash)?
I think that it was doing the inverted slash previously.

build/jobs/e2e-tests-extract.yml
```diff
@@ -0,0 +1,10 @@
+parameters:
+- name: version
+  type: string
+
+steps:
+  - task: ExtractFiles@1
+    displayName: 'Extract E2E Test Binaries'
+    inputs:
+      archiveFilePatterns: '$(System.ArtifactsDirectory)/IntegrationTests/Microsoft.Health.Fhir.${{ parameters.version }}.Tests.E2E.zip'
+      destinationFolder: '$(Agent.TempDirectory)/E2ETests/'
```
fhibf, line 10: Should this folder also have the parameters' version as part of the folder name? 
Can it cause conflicts?

build/jobs/run-export-tests.yml
```diff
@@ -171,7 +177,7 @@ jobs:
...
+      arguments: '"$(Agent.TempDirectory)/E2ETests/**/*${{ parameters.version }}.Tests.E2E*.dll" --filter "FullyQualifiedName~SqlServer&Category=ExportLongRunning"'
       workingDirectory: "$(System.ArtifactsDirectory)"
```
fhibf, line 181: Should the parameter name used as argument also contain "Microsoft.Health.Fhir."?

build/jobs/e2e-tests-extract.yml
brendankowitz: From all the paths I can see, we run only one test package on an agent, so doesn't seem like its strictly needed

build/jobs/run-export-tests.yml
brendankowitz: It could, but matching "R4.Tests.E2E.dll" also is pretty specific, in this PR I've only corrected the path to `$(Agent.TempDirectory)/E2ETests`. 
After messing around with the build files recently I'm wanting to do a larger scale cleanup at some point too :)

build/jobs/run-export-tests.yml
brendankowitz: sorry, to clarify, did you mean for the removed code or new code in this comment?

fhibf: Not a big problem. I was just curious that the previous code was using inverted slashes and the new one is using regular slashes. I was wondering which one if the correct. 

But I also see that previous code in e2e-tests.yml is using regular slashes. 

I think everything is fine.

fhibf (approved):

build/jobs/run-export-tests.yml
brendankowitz: I've been moving most things to `/` since its more compatible between windows/linux

PR #2893, opened 2022-10-20
Changes to use existing sqlServer in builds

rbans96:
## Description

## Related issues
Addresses [issue #].

## Testing

## FHIR Team Checklist


build/jobs/provision-deploy.yml
```diff
@@ -22,6 +22,12 @@ parameters:
 - name: schemaAutomaticUpdatesEnabled
   type: string
   default: 'tool'
+- name: sqlServerAdminPassword
+  type: string
+  default: 'test@123'
```
LTA-Thinking, line 27: I don't think we should provide a default password for the SQL server.

LTA-Thinking, line 27: Could we remove the default value for this?

build/jobs/provision-sqlServer.yml
```diff
@@ -0,0 +1,43 @@
...
+- name: resourceGroup
+  type: string
+- name: sqlServerName
+  type: string
+- name: schemaAutomaticUpdatesEnabled
+  type: string
+  default: 'auto'
+- name: sqlServerAdminPassword
+  type: string
+  default: 'test@123'
```
LTA-Thinking, line 12: Could we remove this default value?

build/jobs/provision-deploy.yml
rbans96: Ah missed, done!

rbans96: Just realized that we would need to set default value as provision-deploy.yml is used for cosmosdb site as well. I am setting it to empty value as default which will error out if we don't pass in good value for sql Server

LTA-Thinking (approved):

PR #2896, opened 2022-10-21
[SMART] US Core - Data missing requirement

fhibf:
## Description
US Core - Data missing requirement. 

In this PR we are including the ability of returning a 404 HTTP error if some specific status code are missing.

For the US Core Missing Data requirement, refer to: http://hl7.org/fhir/us/core/STU3.1.1/general-guidance.html#missing-data

For the following status elements no appropriate “unknown” concept code is available, therefore the element must be populated:
- AllergyIntolerance.clinicalStatus
- Condition.clinicalStatus
- DocumentReference.status
- Immunization.status

In this PR I'm including the ability to filter out resources returned by queries as SearchResult using the Filter Pattern: 
a Criteria is used to apply a filtering logic on top of a result, while a Filter controls which Criteria should be applied. 

image

## Screenshots 

image

image

## Related issues
Addresses [issue #].

## Testing

## FHIR Team Checklist
- [x] Tag the PR with **Azure API for FHIR** if this will release to the Azure API for FHIR managed service (CosmosDB or common code related to service)


src/Microsoft.Health.Fhir.Core/Features/Search/Filters/MissingDataFilterCriteria.cs
```diff
@@ -0,0 +1,71 @@
...
+
+using System.Collections.Generic;
+using EnsureThat;
+using Microsoft.Health.Fhir.Core.Features.Persistence;
+
+namespace Microsoft.Health.Fhir.Core.Features.Search.Filters
+{
+    /// <summary>
+    /// Following US Core requirements, <see cref="MissingDataFilterCriteria"/> filters out resources with missing status codes.
+    /// Reference: http://hl7.org/fhir/us/core/STU3.1.1/general-guidance.html#missing-data
```
brendankowitz, line 14: Do we need to be USCore version aware at some point?

fhibf: That's a good question.
Let me check.

src/Microsoft.Health.Fhir.Core/Features/Search/SearchResultFilter.cs
```diff
@@ -0,0 +1,52 @@
...
+namespace Microsoft.Health.Fhir.Core.Features.Search
+{
+    public sealed class SearchResultFilter : ISearchResultFilter
+    {
+        private readonly bool _isUSCoreEnabled;
+        private readonly bool _isSmartUserRequest;
+
+        public SearchResultFilter(IOptions<ImplementationGuidesConfiguration> implementationGuides, RequestContextAccessor<IFhirRequestContext> fhirRequestContextAccessor)
+        {
+            _isUSCoreEnabled = implementationGuides?.Value?.USCore?.MissingData ?? false;
```
feordin, line 22: This class, SearchResultFilter seems to be designed to be expanded in the future to more search filters, if necessary.  Meaning that it is a container for all search filters.  However, it has detailed knowledge of the necessary parameters for MissingDataFilterCriteria.  If I were to add another filter, it seems I would need to update the constructor of this class, SearchResultFilter, to also have new input parameters for my new search filter.  Perhaps we could move the imlementationGuide and RequestContextAccessor to the MissingDataFilterCriteria constructor?  That way SearchResultFilter doesn't have to know about them.  Just an optional tought.

fhibf: Thanks for the suggestion. Let work on it  :)

src/Microsoft.Health.Fhir.Core/Features/Search/Filters/MissingDataFilterCriteria.cs
fhibf: As discussed last week, we are not handling different versions of USCore today.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/BundleFactory.cs
```diff
@@ -174,6 +174,25 @@ private ResourceElement CreateBundle(SearchResult result, Bundle.BundleType type
                 });
             }
 
+            if (result.SearchIssues.Any())
+            {
+                // FERNFE - Should I create one operation outcome for each record not matching? Or should I continue doing similar to the BundleIssues?
```
fhibf, line 179: @feordin should I create one operation outcome for each record not matching? Or should I continue doing similar to the BundleIssues?

feordin: I think we should be similar to the BundleIssues, and give one compiled set of issues.

feordin (approved):

PR #2898, opened 2022-10-21
Refactors JobHost

brendankowitz:
## Description
- Changes IJobs to have state passed into ExecuteAsync method instead of constructor, this allows IJob implementations to be resolved with IoC.
- Simplifies adding new Job and Queue types

## Testing
Fixes existing unit tests

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Web/Startup.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Web/Startup.cs
@@ -4,6 +4,7 @@
 // -------------------------------------------------------------------------------------------------
 
 using System;
+using System.Collections.Generic;
 using Microsoft.ApplicationInsights.Extensibility;
 using Microsoft.AspNetCore.Authentication.JwtBearer;
 using Microsoft.AspNetCore.Builder;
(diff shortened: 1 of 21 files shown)


src/Microsoft.Health.Fhir.Shared.Web/Startup.cs
```diff
@@ -113,6 +115,16 @@ private void AddTaskHostingService(IServiceCollection services)
                 .AsSelf()
                 .AsImplementedInterfaces();
             services.Configure<TaskHostingConfiguration>(options => Configuration.GetSection("TaskHosting").Bind(options));
+
+            IEnumerable<TypeRegistrationBuilder> jobs = services.TypesInSameAssemblyAs<ImportOrchestratorJob>()
```
LTA-Thinking, line 119: How does this TypesInSameAssemblyAs method work? I want to see how easy it will be for future jobs to be automatically picked up.

src/Microsoft.Health.TaskManagement/IJobInfo.cs
```diff
@@ -0,0 +1,11 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.JobManagement;
+
+public interface IJobInfo
```
LTA-Thinking, line 8: I don't like the name since we also have the JobInfo class, which doesn't inherit from this. Maybe something like JobData?

src/Microsoft.Health.Fhir.Shared.Web/Startup.cs
brendankowitz: Any IJob added in the Core assembly (which is where ImportOrchestratorJob is) should be automatically registered. Is we add jobs from other assemblies we'll need to amend the registration

LTA-Thinking (approved):

PR #2903, opened 2022-10-24
Fixes rawResourceStream isn't always disposed

brendankowitz:
## Description
When `continue` is called the `rawResourceStream` doesn't get disposed.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
@@ -363,18 +363,18 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Search
                             out Stream rawResourceStream);
                         numberOfColumnsRead = reader.FieldCount;
 
-                        // If we get to this point, we know there are more results so we need a continuation token
-                        // Additionally, this resource shouldn't be included in the results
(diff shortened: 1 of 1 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -363,18 +363,18 @@ private async Task<SearchResult> SearchImpl(SqlSearchOptions sqlSearchOptions, S
                             out Stream rawResourceStream);
                         numberOfColumnsRead = reader.FieldCount;
 
-                        // If we get to this point, we know there are more results so we need a continuation token
-                        // Additionally, this resource shouldn't be included in the results
-                        if (matchCount >= clonedSearchOptions.MaxItemCount && isMatch)
+                        string rawResource;
+                        await using (rawResourceStream)
```
LTA-Thinking, line 367: Why the change to await using?

LTA-Thinking (approved):

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
brendankowitz: Stream now implements IAsyncDisposable, so we can use this:
image

PR #2911, opened 2022-10-27
Retry sql deadlocks on UpsertAsync

v-mserdar:
## Description
Adding code that will check for a sql deadlock that has occurred in the UpsertAsync method. When this error is raised by sql the code will generate a random value between 10 and 2000 milliseconds that will be used to delay prior to the next attempt.

## Related issues
Addresses [#97711](https://microsofthealth.visualstudio.com/Health/_workitems/edit/97711).

## Testing
Added a test 

## FHIR Team Checklist


src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -129,14 +129,19 @@ internal class SqlServerFhirDataStore : IFhirDataStore, IProvideCapability
...
+            int retryCount = 0;
+            int baseDelay = 5;
+
             while (true)
             {
                 cancellationToken.ThrowIfCancellationRequested();
 
                 // ** We must use CreateNonRetrySqlCommand here because the retry will not reset the Stream containing the RawResource, resulting in a failure to save the data
                 using (SqlConnectionWrapper sqlConnectionWrapper = await _sqlConnectionWrapperFactory.ObtainSqlConnectionWrapperAsync(cancellationToken, true))
                 using (SqlCommandWrapper sqlCommandWrapper = sqlConnectionWrapper.CreateNonRetrySqlCommand())
```
SergeyGaluzo, line 143: >using (SqlCommandWrapper sqlCommandWrapper = sqlConnectionWrapper.CreateNonRetrySqlCommand()) [](http://example.com/codeflow?start=16&length=93)

Has approach with using retryable SQL command and raw resource as binary array considered?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlExceptionActionProcessor.cs
```diff
@@ -33,6 +36,10 @@ public Task Execute(TRequest request, TException exception, CancellationToken ca
                 {
                     throw new SqlQueryPlanException(Core.Resources.SqlQueryProcessorRanOutOfInternalResourcesException);
                 }
+                else if (sqlException.Number == DeadlockVictim)
```
PTaladay, line 39: I thought the plan was also to not expose these errors to the customer. I didn't happen to see any changes related to that here, unless I missed something.

v-mserdar: What message do we want to send to the customer?
Also, what do you think the HTTP status code should be that gets returned?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
v-mserdar: That would definitely be much more work than I thought we were going to do for this iteration of this issue. The [built in list ](https://github.com/dotnet/SqlClient/blob/main/src/Microsoft.Data.SqlClient/src/Microsoft/Data/SqlClient/Reliability/SqlConfigurableRetryFactory.cs)of current retriable errors doesn't include deadlocks and because of that we'll have to create our own. Of course we can do that but again, I thought we were keeping the code churn down for this initial pass.

src/Microsoft.Health.Fhir.Core/Exceptions/DeadlockVictimException.cs
```diff
@@ -0,0 +1,24 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Diagnostics;
+using Microsoft.Health.Fhir.Core.Models;
+
+namespace Microsoft.Health.Fhir.Core.Exceptions
+{
+    public class DeadlockVictimException : FhirException
```
brendankowitz, line 11: Perhaps `TransactionDeadlockException` or `StorageDeadlockException`, thoughts?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlExceptionActionProcessor.cs
brendankowitz: Since its sql contention perhaps 409?

src/Microsoft.Health.Fhir.Core/Exceptions/DeadlockVictimException.cs
v-mserdar: revised to TransactionDeadlockException

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlExceptionActionProcessor.cs
v-mserdar: updated to a 409

v-mserdar: > I thought the plan was also to not expose these errors to the customer. I didn't happen to see any changes related to that here, unless I missed something.

I have revised the message for the deadlock in the case that retries don't work for whatever reason. For those, the message can be viewed in the Resources file that is also part of this PR.

SergeyGaluzo (approved):
:shipit:

PTaladay (approved):

PR #2912, opened 2022-10-27
defrag 3

SergeyGaluzo:


src/Microsoft.Health.Fhir.SqlServer/Features/Watchdogs/DefragWatchdog.cs
```diff
@@ -0,0 +1,490 @@
...
+            using SqlConnectionWrapper conn = await scopedConn.Value.ObtainSqlConnectionWrapperAsync(CancellationToken.None, false);
+            using SqlCommandWrapper cmd = conn.CreateRetrySqlCommand();
+
+            cmd.CommandText = @"
+INSERT INTO dbo.Parameters (Id,Number) SELECT @IsEnabledId, 0
+INSERT INTO dbo.Parameters (Id,Number) SELECT @ThreadsId, 4
+INSERT INTO dbo.Parameters (Id,Number) SELECT @PeriodSecId, 24*3600
+INSERT INTO dbo.Parameters (Id,Number) SELECT @HeartbeatPeriodSecId, 60
+INSERT INTO dbo.Parameters (Id,Number) SELECT @HeartbeatTimeoutSecId, 600
+INSERT INTO dbo.Parameters (Id,Char) SELECT name, 'LogEvent' FROM sys.objects WHERE type = 'p' AND name LIKE '%defrag%'
```
brendankowitz, line 460: Is this needed with the code in [SqlServerFhirModel.cs](https://github.com/microsoft/fhir-server/pull/2912/files#diff-9fd99f52f9c13ba0c89f3de434555c6cd1541a6a787009d713a9c820dd1a6968)?

SergeyGaluzo: Maybe server model is wrong place to do any of it. Where is a good place from which to put TEST only settings? This should be before storage ready is set.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlQueueClient.cs
```diff
@@ -316,5 +335,71 @@ public async Task<bool> KeepAliveJobAsync(JobInfo jobInfo, CancellationToken can
...
+                _ =>
+                {
+                    tokenSource.Cancel();
+                    return Task.CompletedTask;
+                },
+                TaskScheduler.Current);
+
+            try
+            {
+                await Task.WhenAll(actionTask, heartBeatTask);
```
SergeyGaluzo, line 367: >await Task.WhenAll(actionTask, heartBeatTask); [](http://example.com/codeflow?start=16&length=46)

@brendankowitz Currently it waits on heartbeat task and checks its state. This is different from my previous implementation and is not desired, as heartbeats are meant to be fire and forget. How would you suggest to fix this?

brendankowitz: L360 cancels the timer's token as soon as the action is done, this should be immediate. However, it should still be ok to just wait for action.

```suggestion
                await actionTask;
```

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlQueueClient.cs
```diff
@@ -316,5 +335,71 @@ public async Task<bool> KeepAliveJobAsync(JobInfo jobInfo, CancellationToken can
...
+
+        public async Task ArchiveJobsAsync(byte queueType, CancellationToken cancellationToken)
+        {
+            using SqlConnectionWrapper conn = await _sqlConnectionWrapperFactory.ObtainSqlConnectionWrapperAsync(cancellationToken, false);
+            using SqlCommandWrapper cmd = conn.CreateRetrySqlCommand();
+            VLatest.ArchiveJobs.PopulateCommand(cmd, queueType);
+            await cmd.ExecuteNonQueryAsync(cancellationToken);
+        }
+
+        public async Task ExecuteWithHeartbeatAsync(byte queueType, long jobId, long version, Func<CancellationToken, Task> action, TimeSpan heartbeatPeriod, CancellationToken cancellationToken)
```
brendankowitz, line 347: Should this be in queueclient?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlQueueClient.cs
SergeyGaluzo: I replaced all code by original implementation based on Timer. I like its conciseness.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlQueueClient.cs
SergeyGaluzo: Yes, as it deals with heartbeat on jobs in the queue.

brendankowitz (approved):

PR #2913, opened 2022-10-27
Use DataAction check to test for Smart user status

feordin:
## Description
Rather than rely on the role name to determine if a user is a "SMART on FHIR" user, we will use a new specific DataAction which can be assigned to a role.

This also includes the changes to handle the change from * to all for scopes.

## Related issues
Addresses [issue AB#95787].

## Testing
Unit tests, integration and manual E2E tests were updated to check the new DataAction, and the "all" value

## FHIR Team Checklist
- [X] CI is green before merge [Build Status](https://microsofthealthoss.visualstudio.com/FhirServer/_build/latest?definitionId=27&branchName=main) 

### DIFF
--- a/src/Microsoft.Health.Fhir.Api/Features/SMART/SmartClinicalScopesMiddleware.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/SMART/SmartClinicalScopesMiddleware.cs
@@ -15,6 +15,7 @@ using Microsoft.Health.Core.Features.Context;
 using Microsoft.Health.Fhir.Core.Configs;
 using Microsoft.Health.Fhir.Core.Features.Context;
 using Microsoft.Health.Fhir.Core.Features.Security;
+using Microsoft.Health.Fhir.Core.Models;
 
 namespace Microsoft.Health.Fhir.Api.Features.Smart
 {
@@ -26,7 +27,7 @@ namespace Microsoft.Health.Fhir.Api.Features.Smart
         private readonly RequestDelegate _next;
 
         // Regex based on SMART on FHIR clinical scopes v1.0, http://hl7.org/fhir/smart-app-launch/1.0.0/scopes-and-launch-context/index.html#clinical-scope-syntax
-        private static readonly Regex ClinicalScopeRegEx = new Regex(@"(^|\s+)(?<id>patient|user)(/|\$|.)(?<resource>\*|([a-zA-Z]*))\.(?<accessLevel>read|write|\*)", RegexOptions.Compiled);
+        private static readonly Regex ClinicalScopeRegEx = new Regex(@"(^|\s+)(?<id>patient|user)(/|\$|\.)(?<resource>\*|([a-zA-Z]*)|all)\.(?<accessLevel>read|write|\*|all)", RegexOptions.Compiled);
 
         public SmartClinicalScopesMiddleware(RequestDelegate next)
         {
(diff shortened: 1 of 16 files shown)


rajithaalurims (dismissed):
@feordin  
Below is what I understood about the flow. Correct me if I am wrong.

1. "FHIR SMART User" role is going to have "Smart" data action in addition to Read data action.
2. Request Comes in and first you call Authorization Check access to see if User has Smart Data Action.
3. If Yes, then all the needed flags and Scope restrictions are populated for Smart.
4. AuthorizationService gets called again from Handlers/Code that is handling the request.
5. Authorization check now uses the Smart flag populated and continue with the checks.

src/Microsoft.Health.Fhir.Core/Configs/AuthorizationConfiguration.cs
```diff
@@ -22,5 +22,7 @@ public class AuthorizationConfiguration
         public string FhirUserClaim { get; set; } = "fhirUser";
 
         public string SmartUserRole { get; set; } = "smartUser";
```
yoju0121, line 24: Do we need the `SmartUserRole` anymore?

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/SMART/SmartClinicalScopesMiddlewareTests.cs
```diff
@@ -4,17 +4,30 @@
...
 using Microsoft.Health.Fhir.Api.Features.Smart;
+using Microsoft.Health.Fhir.Core.Configs;
 using Microsoft.Health.Fhir.Core.Features.Context;
 using Microsoft.Health.Fhir.Core.Features.Security;
+using Microsoft.Health.Fhir.Core.Features.Security.Authorization;
+using Microsoft.Health.Fhir.Core.Models;
 using Microsoft.Health.Fhir.Core.UnitTests.Features.Context;
 using Microsoft.Health.Fhir.Tests.Common;
 using Microsoft.Health.Test.Utilities;
+using Newtonsoft.Json.Linq;
```
yoju0121, line 30: Should the tests in this file whose names start with `GivenSmartUserRole_` be changed to `GivenSmartDataAction_`?

yoju0121: + should we remove the code in the tests that adds the `smartUser` role? Since the scope reading/applying is now done based on the smart data action.

src/Microsoft.Health.Fhir.Api/Features/SMART/SmartClinicalScopesMiddleware.cs
```diff
@@ -90,13 +98,19 @@ public async Task Invoke(HttpContext context, RequestContextAccessor<IFhirReques
                                     permittedDataActions |= DataActions.Write;
                                     break;
                                 case "*":
+                                case KnownResourceTypes.All:
```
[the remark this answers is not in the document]
feordin: This is true.  It happens to be the same value, but I should rename the constant, or make a new one.

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/SMART/SmartClinicalScopesMiddlewareTests.cs
feordin: You are correct, we should rename things to be more clear.  As far as removing the code for the role, we still need that, because that is the mechanism by which the user gets the smart data action.

rajithaalurims (approved):

rajithaalurims (approved):
We need to still address below changes:
SMARTScopeFhirAuthorizationService should be renames to SmartScopeFhirAuthorizationService
SmartClinicalScopesMiddleware class has unused local variable roles.

PR #2915, opened 2022-10-31
Add same resource as compartment definition to SMART results

feordin:
## Description
When using the compartment definition to filter results for a SMART query, we do not want to prohibit access to the same resource which is used to define the compartment.  In other words, a given Patient should be able to read their own Patient resource.

## Related issues
Addresses [issue AB#95786].

## Testing
Additional unit test and integration tests were added.

## FHIR Team Checklist
- [X] CI is green before merge [Build Status](https://microsofthealthoss.visualstudio.com/FhirServer/_build/latest?definitionId=27&branchName=main) 

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Context/FhirUserClaimParser.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Context/FhirUserClaimParser.cs
@@ -23,8 +23,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Context
             // The URI is of the form <base URL>/<resource type>/<resource id>.
             // The resource type must be one of Patient, Practitioner.  In the future it may be extended to include RelatedPerson.
 
-            if (accessControlContext.FhirUserClaim == null ||
-                accessControlContext.FhirUserClaim.Segments.Length < 2)
+            if (accessControlContext.FhirUserClaim == null)
             {
                 if (errorOnInvalidFhirUserClaim)
                 {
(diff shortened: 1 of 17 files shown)


src/Microsoft.Health.Fhir.Core/Features/Context/FhirUserClaimParser.cs
```diff
@@ -36,8 +35,27 @@ public static class FhirUserClaimParser
...
-            string fhirUserId = accessControlContext.FhirUserClaim.Segments.Last().TrimEnd('/');
-            string fhirUserResourceType = accessControlContext.FhirUserClaim.Segments[accessControlContext.FhirUserClaim.Segments.Length - 2].TrimEnd('/');
+            var fhirUserClaimUri = accessControlContext.FhirUserClaim;
+            if (!accessControlContext.FhirUserClaim.IsAbsoluteUri)
+            {
+                // convert to absolute Uri so that the Uri parsing methods will function
+                fhirUserClaimUri = new Uri(new Uri("http://fhirBase"), accessControlContext.FhirUserClaim);
+            }
+
+            if (fhirUserClaimUri.Segments.Length < 2)
```
rajithaalurims, line 45: Can we also add another || condition to check if Segments length is greater than 2? fhirUserClaimUri.Segments.Length > 2?

feordin: In the case of a fully qualified URL, the segments length may be longer than 2 and still be correct.

PR #2925, opened 2022-11-02
Revised error messages sent to customers

v-mserdar:
## Description
This PR addresses 2 issues for messaging customers.
First, we're removing the exception details that are being sent back to the customer and sending a resource message instead. Second, it was discovered that we didn't handle SqlExceptions for login user failures and with that we're also sending a message to let the customer know they should update the login and retry.

## Related issues
Addresses [#97708](https://microsofthealth.visualstudio.com/Health/_workitems/edit/97708).

## Testing
Added a new test and also confirmed the http result when debugging locally

## FHIR Team Checklist


src/Microsoft.Health.Fhir.Core/Resources.resx
```diff
@@ -678,4 +678,10 @@
   <data name="FhirUserClaimInvalidFormat" xml:space="preserve">
     <value>The fhirUser claim must be a URI with an ending format of &lt;resourceType&gt;/&lt;resourceId&gt;.</value>
   </data>
+  <data name="InternalServerError" xml:space="preserve">
+    <value>An internal error has occurred. If this problem persists, please contact Customer Support.</value>
+  </data>
+  <data name="LoginFailedForUser" xml:space="preserve">
+    <value>Unauthorized Error caused due to failed login. Request to retry with valid credentials.</value>
```
brendankowitz, line 685: We have generally seen this as a transient error when a DB is moved by the loadbalancer, which is going to seem confusing

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlExceptionActionProcessor.cs
```diff
@@ -33,9 +35,13 @@ public Task Execute(TRequest request, TException exception, CancellationToken ca
                 {
                     throw new SqlQueryPlanException(Core.Resources.SqlQueryProcessorRanOutOfInternalResourcesException);
                 }
+                else if (sqlException.Number == LoginFailedForUser)
+                {
+                    throw new LoginFailedForUserException(Core.Resources.LoginFailedForUser);
```
brendankowitz, line 40: Should this message just be the default of this exception?

brendankowitz: For now I think this makes more sense. There isn't much a FHIR Client can do about an error with the database connection which is between the app and sql

PTaladay (approved):

PR #2938, opened 2022-11-08
$export read consistency

SergeyGaluzo:


src/Microsoft.Health.Fhir.Core/Features/Search/SearchService.cs
```diff
@@ -232,6 +232,16 @@ protected SearchService(ISearchOptionsFactory searchOptionsFactory, IFhirDataSto
             throw new NotImplementedException();
         }
 
+        public virtual Task<IReadOnlyList<(long start, long end, long globalStart, long globalEnd)>> GetSurrogateIdRanges(
```
brendankowitz, line 235: I think the naming standards now prefer pascal-case field names on tuples. Though this might also be a good case for returning a `record`.

```suggestion
        public virtual Task<IReadOnlyList<(long Start, long End, long GlobalStart, long GlobalEnd)>> GetSurrogateIdRanges(
```

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
```diff
@@ -80,6 +80,17 @@ public SearchOptions Create(string resourceType, IReadOnlyList<Tuple<string, str
         {
             var searchOptions = new SearchOptions();
 
+            if (queryParameters != null && queryParameters.Any(_ => _.Item1 == KnownQueryParameterNames.GlobalEndSurrogateId && _.Item2 != null))
```
brendankowitz, line 83: I thought `_` is generally used when the variable isn't used

src/Microsoft.Health.Fhir.Core/Features/Search/SearchService.cs
SergeyGaluzo: Sure.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
SergeyGaluzo: In some cases - yes. I use _ in this type of contexts to clearly indicate that its name does not matter.

PR #2949, opened 2022-11-10
$export time travel

SergeyGaluzo:


test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/FhirStorageTests.cs
```diff
@@ -67,6 +68,78 @@ public FhirStorageTests(FhirStorageTestsFixture fixture)
...
+                Tuple.Create(KnownQueryParameterNames.GlobalStartSurrogateId, range.GlobalStart.ToString()),
+                Tuple.Create(KnownQueryParameterNames.StartSurrogateId, range.Start.ToString()),
+            };
+
+            // !!! time travel behavior
+            results = await _fixture.SearchService.SearchAsync(type, queryParameters, CancellationToken.None);
+            Assert.Single(results.Results);
+            resource = results.Results.First().Resource;
+            Assert.Equal("1", resource.Version);
+            Assert.False(resource.IsHistory); // it is returned as current but is marked as history in the database ???
```
LTA-Thinking, line 123: This is strange. We should look into why this is and think about if this is the right behavior.

LTA-Thinking (approved):

PR #2954, opened 2022-11-15
Export work distribution redesign

SergeyGaluzo:
This PR addresses all noted problems:
1.	Hardcoded values for range size and number of ranges. It will limit number of resources per resource type to ~100K.
2.	Duplicate resources: Boundaries of surrogate id ranges returned from my stored procedure are not overlapping by design. Later they are converted to datetime with precision loss. This leads to duplicates.
3.	Not optimal work item size: Currently entire range of surrogate ids is divided in 10 work items. Optimal number of exported resources in a work item is ~100K (it was a result from POC we ran back in May). So this 10 can be optimal for export with ~1M resources. I believe that for smaller exports we will be creating too many small files, and larger exports can become unmanageable due to long running work items. I would suggest changing logic to the original POC design when number of buckets is not preset but naturally derived depending on existing the data.  
4.	Limited parallelism (explicit _since is currently required). 
5. Limited parallelism (explicit _type is currently required). 
6.	Exception: When export is called with explicit time range that do not have matching resources in the database export fails with exception.
7.	Current code relies on enqueueing/registering all work items at once. This is not manageable for large datasets.

### DIFF
--- a/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Export/ExportOrchestratorJobTests.cs
+++ b/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Export/ExportOrchestratorJobTests.cs
@@ -45,6 +45,7 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Operations.Export
 
             var orchestratorJob = GetJobInfoArray(0, orchestratorJobId, false, orchestratorJobId, numExpectedJobs, exportJobType: exportJobType).First();
             var exportOrchestratorJob = new ExportOrchestratorJob(_mockQueueClient, _mockSearchService, _loggerFactory);
+            exportOrchestratorJob.PollingFrequencyInSeconds = 1;
             var result = await exportOrchestratorJob.ExecuteAsync(orchestratorJob, new Progress<string>((result) => { }), CancellationToken.None);
             var jobResult = JsonConvert.DeserializeObject<ExportJobRecord>(result);
             CountOutputFiles(jobResult, numExpectedJobs);
@@ -60,6 +61,7 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Operations.Export
 
             var orchestratorJob = GetJobInfoArray(0, orchestratorJobId, false, orchestratorJobId, numExpectedJobs).First();
             var exportOrchestratorJob = new ExportOrchestratorJob(_mockQueueClient, _mockSearchService, _loggerFactory);
+            exportOrchestratorJob.PollingFrequencyInSeconds = 1;
             var result = await exportOrchestratorJob.ExecuteAsync(orchestratorJob, new Progress<string>((result) => { }), CancellationToken.None);
             var jobResult = JsonConvert.DeserializeObject<ExportJobRecord>(result);
             CountOutputFiles(jobResult, numExpectedJobs);
@@ -75,6 +77,7 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Operations.Export
 
             var orchestratorJob = GetJobInfoArray(0, orchestratorJobId, false, orchestratorJobId, numExpectedJobs).First();
             var exportOrchestratorJob = new ExportOrchestratorJob(_mockQueueClient, _mockSearchService, _loggerFactory);
+            exportOrchestratorJob.PollingFrequencyInSeconds = 1;
             var result = await exportOrchestratorJob.ExecuteAsync(orchestratorJob, new Progress<string>((result) => { }), CancellationToken.None);
             var jobResult = JsonConvert.DeserializeObject<ExportJobRecord>(result);
             CountOutputFiles(jobResult, numExpectedJobs);
@@ -86,13 +89,15 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Operations.Export
             int numExpectedJobs = 10;
             long orchestratorJobId = 10000;
 
-            SetupMockQueue(numExpectedJobs * 2, orchestratorJobId);
+            SetupMockQueue(numExpectedJobs, orchestratorJobId);
 
             var orchestratorJob = GetJobInfoArray(0, orchestratorJobId, false, orchestratorJobId, numExpectedJobs, typeFilter: "Patient,Observation").First();
             var exportOrchestratorJob = new ExportOrchestratorJob(_mockQueueClient, _mockSearchService, _loggerFactory);
+            exportOrchestratorJob.PollingFrequencyInSeconds = 1;
+            exportOrchestratorJob.NumberOfSurrogateIdRanges = 10;
             var result = await exportOrchestratorJob.ExecuteAsync(orchestratorJob, new Progress<string>((result) => { }), CancellationToken.None);
             var jobResult = JsonConvert.DeserializeObject<ExportJobRecord>(result);
-            CountOutputFiles(jobResult, numExpectedJobs * 2);
+            CountOutputFiles(jobResult, numExpectedJobs);
         }
 
         [Fact]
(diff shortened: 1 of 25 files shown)


src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Export/ExportOrchestratorJobTests.cs
```diff
@@ -33,106 +33,113 @@ public class ExportOrchestratorJobTests
         private IQueueClient _mockQueueClient = Substitute.For<IQueueClient>();
         private ILoggerFactory _loggerFactory = new NullLoggerFactory();
 
-        [Theory]
-        [InlineData(ExportJobType.Patient)]
-        [InlineData(ExportJobType.Group)]
-        public async Task GivenANonSystemLevelExportJob_WhenRun_ThenOneProcessingJobShouldBeCreated(ExportJobType exportJobType)
-        {
```
SergeyGaluzo, line 40: I temporarily commented this out as don't fit into new paradigm as is. Not sure whether we should invest in fixing or just remove altogether.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
```diff
@@ -107,9 +107,83 @@ public async Task<ExportJobOutcome> GetExportJobByIdAsync(string id, Cancellatio
             }
 
             var def = jobInfo.Definition;
+            var status = jobInfo.Status;
             var result = jobInfo.Result;
-            var rawJobRecord = result ?? def;
-            return CreateExportJobOutcome(jobId, rawJobRecord, jobInfo.Version, (byte)jobInfo.Status, jobInfo.CreateDate);
+
+            if (status == JobStatus.Completed)
+            {
```
SergeyGaluzo, line 114: This logic was previously in the execute method of orchestrator job.

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Features/Operations/Import/SqlServerIndexesRebuildTests.cs
```diff
@@ -329,7 +329,7 @@ private async Task<string> CompareDatabaseSchemas(string databaseName1, string d
                 {
                     //// Our home grown SQL schema generator does not understand that statements can be formatted differently but contain identical SQL
                     //// Skipping queue objects
-                    var objectsToSkip = new[] { "DequeueJob", "EnqueueJobs", "GetJobs", "GetResourcesByTypeAndSurrogateIdRange", "GetResourceSurrogateIdRanges", "LogEvent", "PutJobCancelation", "PutJobHeartbeat", "PutJobStatus" };
+                    var objectsToSkip = new[] { "DequeueJob", "EnqueueJobs", "GetJobs", "GetResourcesByTypeAndSurrogateIdRange", "GetResourceSurrogateIdRanges", "LogEvent", "PutJobCancelation", "PutJobHeartbeat", "PutJobStatus", "CompartmentAssignment" };
```
SergeyGaluzo, line 332: >"CompartmentAssignment" [](http://example.com/codeflow?start=229&length=23)

This is to fix failing index rebuild tests which do full database compare...

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Export/ExportOrchestratorJobTests.cs
LTA-Thinking: I would say getting at least two tests working would be good, one for a parallel job and one for a single thread job.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportOrchestratorJob.cs
```diff
@@ -21,12 +21,13 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Export
     [JobTypeId((int)JobType.ExportOrchestrator)]
     public class ExportOrchestratorJob : IJob
     {
-        private const int DefaultPollingFrequencyInSeconds = 3;
-        private const int DefaultNumberOfParallelJobs = 10;
+        private const double DefaultPollingIntervalSec = 60;
```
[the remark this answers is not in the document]
SergeyGaluzo: To set smaller than 1 second polling in tests.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
```diff
@@ -107,9 +107,83 @@ public async Task<ExportJobOutcome> GetExportJobByIdAsync(string id, Cancellatio
...
+                            else
+                            {
+                                record.FailureDetails = new JobFailureDetails(record.FailureDetails.FailureReason + "\r\nProcessing job had no results", HttpStatusCode.InternalServerError);
+                            }
+
+                            jobFailed = true;
+                        }
+                    }
+
+                    record.Status = jobFailed ? OperationStatus.Failed : OperationStatus.Completed;
```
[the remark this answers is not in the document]
SergeyGaluzo: Cancellation did not work at all. I fixed that,

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
```diff
@@ -107,9 +107,85 @@ public async Task<ExportJobOutcome> GetExportJobByIdAsync(string id, Cancellatio
...
+                    status = jobFailed ? JobStatus.Failed : JobStatus.Completed;
+                    result = JsonConvert.SerializeObject(record);
+                }
+                else if (cancelledJobsExist)
+                {
+                    status = JobStatus.Cancelled;
```
SergeyGaluzo, line 180: >status = JobStatus.Cancelled; [](http://example.com/codeflow?start=20&length=29)

This is optimistic as not all jobs finished, but should not matter.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
```diff
@@ -137,7 +213,8 @@ public async Task<ExportJobOutcome> UpdateExportJobAsync(ExportJobRecord jobReco
                 }
                 else if (jobRecord.Status == OperationStatus.Canceled)
                 {
-                    await _queueClient.CancelJobByIdAsync((byte)QueueType.Export, jobId, cancellationToken);
+                    var jobWithGroupId = await _queueClient.GetJobByIdAsync((byte)QueueType.Export, jobId, false, cancellationToken);
+                    await _queueClient.CancelJobByGroupIdAsync((byte)QueueType.Export, jobWithGroupId.GroupId, cancellationToken);
```
SergeyGaluzo, line 217: This is the key fir cancellation for a group to work correctly.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportOrchestratorJob.cs
```diff
@@ -42,162 +43,95 @@ public class ExportOrchestratorJob : IJob
...
                         {
-                            record.FailureDetails = new JobFailureDetails("Processing job had no results", System.Net.HttpStatusCode.InternalServerError);
+                            await _queueClient.EnqueueAsync((byte)QueueType.Export, definitions.ToArray(), jobInfo.GroupId, false, false, cancellationToken);
                         }
-                        else
+
+                        if (RaiseTestException)
                         {
-                            record.FailureDetails = new JobFailureDetails(record.FailureDetails.FailureReason + "\r\nProcessing job had no results", HttpStatusCode.InternalServerError);
+                            throw new ArgumentException("Test");
```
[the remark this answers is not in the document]
SergeyGaluzo: I don't see an easy way to accomplish testing on failures. Other methods that use "timing" approach will fail intermittently and make our testing even less reliable than today.

LTA-Thinking (approved):

PR #2972, opened 2022-11-18
Get rid of IEnumerable

SergeyGaluzo:
This PR exposes underlining List that is being built anyway as an output so we don't have confusion what is and is not there.

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Watchdogs/DefragWatchdog.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Watchdogs/DefragWatchdog.cs
@@ -212,11 +212,11 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Watchdogs
             (long groupId, long jobId, long version) id = (-1, -1, -1);
             try
             {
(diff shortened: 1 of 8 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Watchdogs/DefragWatchdog.cs
```diff
@@ -212,11 +212,11 @@ private async Task<(long groupId, long jobId, long version, int activeDefragItem
             (long groupId, long jobId, long version) id = (-1, -1, -1);
             try
             {
-                JobInfo job = (await queueClient.EnqueueAsync(QueueType, new[] { "Defrag" }, null, true, false, cancellationToken)).FirstOrDefault();
+                var jobs = await queueClient.EnqueueAsync(QueueType, new[] { "Defrag" }, null, true, false, cancellationToken);
```
brendankowitz, line 215: The previous call is `FirstOrDefault`, can nulls still be returned here?

SergeyGaluzo: Previous null was only in case of empty list. No, job info cannot be null.

PTaladay (approved):

PR #3014, opened 2022-12-08
[Perf] Memory management improvements.

fhibf:
## Description
This PR contains multiple changes to improve memory consumption. 

Some of these changes are: 
* Disposing objects left alive in memory:
  * AzureContainerRegistryAccessTokenProvider.cs
  * CosmosDbCollectionPhysicalPartitionInfo
  * SecurityProvider
  * FhirCosmosResponseHandler

* Disposing stream under FhirCosmosSerializer.FromStream as the public documentation suggests: https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.cosmos.cosmosserializer.fromstream

* Adding tags to streams created by the RecyclableMemoryStream to help identifying streams created in memory. 

* Replace HttpResponseMessage being accumulated in memory with a new domain class called CosmosResponseMessage.

* Change PerformanceTest for not returning streams from the query being used, instead setting EnableContentResponseOnWrite to false and dealing with the HTTP Status Code.

* Limiting JsonArrayPoolTests to rent byte[] up to 100MB. 

* Adding Garbage Collection calls to bundle operations to help GC collecting leftovers from previous executions. 

* Added logs to different places where streams are managed, in order to understand when some huge allocations are happening.

## Related issues
* Customer suffering with memory leak. 

## Testing
**Adding GC calls to Bundle processing**
Tested bundle execution locally, I've noticed that when pushing a bundle with 500 records FHIR process reached 3.5GB in few minutes. 
image

After adding the GC calls for every 150 records in a bundle, memory improved a lot and now the process memory is around 1.2GB. 
image

When running multiple bundle calls at the same time, the increase in memory is still safe. In my experiments I've ran up to four bundles at the same time, in a single FHIR instance, and it was able to ingest all data properly.
image

## FHIR Team Checklist
- [x] Tag the PR with **Azure API for FHIR** if this will release to the Azure API for FHIR managed service (CosmosDB or common code related to service)

### DIFF
--- a/src/Microsoft.Health.Fhir.Azure/IntegrationDataStore/AzureBlobSourceStream.cs
+++ b/src/Microsoft.Health.Fhir.Azure/IntegrationDataStore/AzureBlobSourceStream.cs
@@ -172,7 +172,7 @@ namespace Microsoft.Health.Fhir.Azure.IntegrationDataStore
 
         private async Task<Stream> DownloadDataFunc(long offset, long length)
         {
-            var stream = new RecyclableMemoryStream(_recyclableMemoryStreamManager);
+            var stream = new RecyclableMemoryStream(_recyclableMemoryStreamManager, tag: nameof(AzureBlobSourceStream));
             await _blobClient.DownloadRangeToStreamAsync(stream, offset, length);
             stream.Position = 0;
 
--- a/src/Microsoft.Health.Fhir.Azure/IntegrationDataStore/AzureBlobSourceStream.cs
+++ b/src/Microsoft.Health.Fhir.Azure/IntegrationDataStore/AzureBlobSourceStream.cs
@@ -18,6 +18,7 @@ namespace Microsoft.Health.Fhir.Azure.IntegrationDataStore
 {
     public class AzureBlobSourceStream : Stream
     {
+        private const int BlobSizeThresholdWarningInBytes = 1000000; // 1MB threshold.
         private const int DefaultConcurrentCount = 3;
         public const int DefaultBlockBufferSize = 8 * 1024 * 1024;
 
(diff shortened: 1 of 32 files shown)


src/Microsoft.Health.Fhir.Azure/IntegrationDataStore/AzureBlobSourceStream.cs
```diff
@@ -172,7 +172,7 @@ private async Task RefreshBlobClientAsync()
 
         private async Task<Stream> DownloadDataFunc(long offset, long length)
         {
-            var stream = new RecyclableMemoryStream(_recyclableMemoryStreamManager);
+            var stream = new RecyclableMemoryStream(_recyclableMemoryStreamManager, tag: nameof(AzureBlobSourceStream));
```
PTaladay, line 175: Wouldn't you want to use a "using" statement here as well? No guarantees when garbage collection would get to it.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/FhirCosmosClientInitializer.cs
```diff
@@ -185,7 +187,7 @@ public override T FromStream<T>(Stream stream)
 
             public override Stream ToStream<T>(T input)
             {
-                MemoryStream stream = _manager.GetStream();
+                MemoryStream stream = _manager.GetStream(tag: nameof(FhirCosmosSerializer));
```
PTaladay, line 190: Do we need a using here as well for this stream?

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Import/ImportResourceParser.cs
```diff
@@ -76,7 +76,7 @@ private static void CheckConditionalReferenceInResource(Resource resource)
 
         private Stream GenerateCompressedRawResource(string rawResource)
         {
-            var outputStream = new RecyclableMemoryStream(_recyclableMemoryStreamManager);
+            var outputStream = new RecyclableMemoryStream(_recyclableMemoryStreamManager, tag: nameof(ImportResourceParser));
```
PTaladay, line 79: Again, use of a using statement here?

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/FhirCosmosClientInitializer.cs
PTaladay: Nevermind, I see it is returned.

src/Microsoft.Health.Fhir.Azure/IntegrationDataStore/AzureBlobSourceStream.cs
PTaladay: Nevermind, I see it gets returned.

PTaladay (approved):

PR #3037, opened 2022-12-22
Revised OperationOutcome message when identifier missing

v-mserdar:
## Description
This change will handle cases where a PUT operation doesn't have the identifier for the resource type included in a url field. When this issue is caught, we'll provide a more meaningful message that states what resource type is missing the identifier.

## Related issues
Addresses [issue #95161](https://microsofthealth.visualstudio.com/Health/_workitems/edit/95161).

## Testing
Created new supporting test

## FHIR Team Checklist

### DIFF
--- a/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ConditionalDeleteTests.cs
+++ b/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ConditionalDeleteTests.cs
@@ -37,6 +37,15 @@ namespace Microsoft.Health.Fhir.Tests.E2E.Rest
             _client = fixture.TestFhirClient;
         }
 
+        [Fact]
+        [Trait(Traits.Priority, Priority.One)]
+        public async Task GivenAnIncompleteSearchParam_WhenDeletingConditionally_TheServerRespondsWithCorrectMessage()
+        {
+            FhirException fhirException = await Assert.ThrowsAsync<FhirException>(() => _client.DeleteAsync($"{_resourceType}?identifier=", CancellationToken.None));
+            Assert.Equal(HttpStatusCode.PreconditionFailed, fhirException.StatusCode);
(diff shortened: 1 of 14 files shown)


test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ConditionalDeleteTests.cs
```diff
@@ -121,7 +130,7 @@ public async Task GivenMatchingResources_WhenDeletingConditionallyWithMultipleFl
 
             await Task.WhenAll(Enumerable.Range(1, create).Select(_ => CreateWithIdentifier(identifier)));
 
-            await GetResourceCount(identifier);
+            // await GetResourceCount(identifier);
```
PTaladay, line 133: If this is commented out why not just removed it? We have the history in the repo.

PTaladay (dismissed):

rajithaalurims (commented):
@v-mserdar 
Changes you added looks good.
I see ConditionalOperationNotSelectiveEnough message used in two more places. ConditionalCreateTests. GivenAResource_WhenCreatingConditionallyWithEmptyIfNoneHeader_TheServerShouldFail method. Isn't this method going to fail now because the message has resource type in it?
Also SearchServiceExtensions class uses this message. We need to change the message to accept Resource Type there.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/ConditionalDeleteTests.cs
v-mserdar: Thanks for finding this @PTaladay . I've removed that line altogether as you have stated. I saw that it was never being used to begin with.

v-mserdar:
> 

Yes, that test does fail - I should've caught that locally. It's now been updated. Thanks for finding it.

rajithaalurims (commented):
Can you also take a look into SearchServiceExtensions class?

rajithaalurims (approved):

PR #3039, opened 2022-12-28
Correct heartbeats

SergeyGaluzo:
This addresses these bugs:
1. convoys on database calls - https://microsofthealth.visualstudio.com/Health/_queries/edit/98697/?triage=true
2. heartbeats after job completion - https://microsofthealth.visualstudio.com/Health/_queries/edit/99441/?triage=true

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlQueueClient.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlQueueClient.cs
@@ -12,12 +12,14 @@ using System.Threading.Tasks;
 using EnsureThat;
 using Microsoft.Data.SqlClient;
 using Microsoft.Extensions.Logging;
+using Microsoft.Health.Fhir.Core.Features.Operations;
 using Microsoft.Health.Fhir.SqlServer.Features.Schema;
 using Microsoft.Health.Fhir.SqlServer.Features.Schema.Model;
 using Microsoft.Health.JobManagement;
 using Microsoft.Health.SqlServer.Features.Client;
 using Microsoft.Health.SqlServer.Features.Schema;
 using Microsoft.Health.SqlServer.Features.Storage;
+using SemVer;
 using JobStatus = Microsoft.Health.JobManagement.JobStatus;
 
 namespace Microsoft.Health.Fhir.SqlServer.Features.Storage
@@ -318,35 +320,33 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Storage
 
         public async Task<bool> KeepAliveJobAsync(JobInfo jobInfo, CancellationToken cancellationToken)
         {
-            using SqlConnectionWrapper sqlConnectionWrapper = await _sqlConnectionWrapperFactory.ObtainSqlConnectionWrapperAsync(cancellationToken, true);
-            using SqlCommandWrapper sqlCommandWrapper = sqlConnectionWrapper.CreateRetrySqlCommand();
-
+            var cancel = false;
             try
             {
+                using SqlConnectionWrapper sqlConnectionWrapper = await _sqlConnectionWrapperFactory.ObtainSqlConnectionWrapperAsync(cancellationToken, true);
+                using SqlCommandWrapper sqlCommandWrapper = sqlConnectionWrapper.CreateNonRetrySqlCommand();
+
                 if (_schemaInformation.Current >= SchemaVersionConstants.ReturnCancelRequestInJobHeartbeat)
                 {
                     VLatest.PutJobHeartbeat.PopulateCommand(sqlCommandWrapper, jobInfo.QueueType, jobInfo.Id, jobInfo.Version, jobInfo.Data, jobInfo.Result, false);
 
                     await sqlCommandWrapper.ExecuteNonQueryAsync(cancellationToken);
-                    return VLatest.PutJobHeartbeat.GetOutputs(sqlCommandWrapper) ?? false;
+                    cancel = VLatest.PutJobHeartbeat.GetOutputs(sqlCommandWrapper) ?? false;
                 }
                 else
                 {
                     V36.PutJobHeartbeat.PopulateCommand(sqlCommandWrapper, jobInfo.QueueType, jobInfo.Id, jobInfo.Version, jobInfo.Data, jobInfo.Result);
 
                     await sqlCommandWrapper.ExecuteNonQueryAsync(cancellationToken);
-                    return (await GetJobByIdAsync(jobInfo.QueueType, jobInfo.Id, false, cancellationToken)).CancelRequested;
+                    cancel = (await GetJobByIdAsync(jobInfo.QueueType, jobInfo.Id, false, cancellationToken)).CancelRequested;
                 }
             }
-            catch (SqlException sqlEx)
+            catch (Exception ex)
             {
-                if (sqlEx.Number == SqlErrorCodes.PreconditionFailed)
-                {
-                    throw new JobNotExistException(sqlEx.Message);
-                }
-
-                throw;
+                _logger.LogError(ex, "Failed to put job heartbeat.");
             }
+
+            return cancel;
         }
 
         public async Task ArchiveJobsAsync(byte queueType, CancellationToken cancellationToken)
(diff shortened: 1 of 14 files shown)


src/Microsoft.Health.TaskManagement/IQueueClient.cs
```diff
@@ -106,5 +99,28 @@ public interface IQueueClient
...
+        /// <summary>
+        /// Executes job with heartbeats. Put heartbeat logic saves Data and Result from provided jobInfo.
+        /// Usage of this method should be discouraged. Please use method with lightweight heartbeats.
+        /// </summary>
+        /// <param name="jobInfo">job info</param>
+        /// <param name="action">job action</param>
+        /// <param name="heartbeatPeriod">heartbeat period in seconds</param>
+        /// <param name="cancellationTokenSource">cancellationTokenSource</param>
+        /// <returns>result</returns>
+        public Task<string> ExecuteJobWithHeartbeats(JobInfo jobInfo, Func<CancellationTokenSource, Task<string>> action, TimeSpan heartbeatPeriod, CancellationTokenSource cancellationTokenSource);
```
skwwt88, line 112: I don't think it is appropriate to put keep heartbeat timer in queue client. Queue client should be only just storage level implementation. let's put the timer in the job hosting.

tongwu-msft, line 112: I don't think it is appropriate to put keep heartbeat timer in queue client. Queue client should be only just storage level implementation. let's put the timer in the job hosting.

tongwu-msft (changes requested):

src/Microsoft.Health.TaskManagement/JobHosting.cs
```diff
@@ -80,11 +65,11 @@ private async Task PullAndProcessJobsAsync(byte queueType, string workerName, Ca
 
                 if (nextJob != null)
                 {
-                    runningJobs.Add(ExecuteJobAsync(nextJob));
+                    runningJobs.Add(ExecuteJobAsync(nextJob, useHeavyHeartbeats));
                 }
                 else
                 {
-                    await intervalDelayTask;
+                    await Task.Delay(TimeSpan.FromSeconds(PollingFrequencyInSeconds), CancellationToken.None);
```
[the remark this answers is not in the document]
SergeyGaluzo: This line 
   await Task.Delay(TimeSpan.FromSeconds(PollingFrequencyInSeconds), CancellationToken.None);
is just a consolidation of previous delay task definition 
   var intervalDelayTask = Task.Delay(TimeSpan.FromSeconds(PollingFrequencyInSeconds), CancellationToken.None);
and its execution
   await intervalDelayTask;
on single line.
This is done for code readability. There is no change in logic here.

src/Microsoft.Health.TaskManagement/IQueueClient.cs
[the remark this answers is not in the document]
SergeyGaluzo: There are 2 main classes here SqlQueueClient and JobHosting. JobHosting holds main job manipulation logic common for $export and $import. When working on DefragWatchdog I first considered JobHosting class for defrag job manipulation logic. I found JobHosting to be too restrictive and not appropriate. 

As a compromise, I created generic ExecuteWithHeartbeats method in SqlQueueClient. This method allows to call already existing in SqlQueueClient PutJobHeartbeat method repeatedly. It provides input for any true business logic as a parameter. This has been approved as part of defrag PR and is in production today.

In this PR I use the same approach and make it consistent. Also I make all heartbeat calls private in SqlQueueClient, as there is no need to publicize them anywhere else.

Though one can consider a new class to hold ExecuteWithHeartbeats, I would advice against it.

src/Microsoft.Health.TaskManagement/JobHosting.cs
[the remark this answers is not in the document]
SergeyGaluzo: Thank you for pointing this out. 

Usually each dequeue operation takes ~10 milliseconds, so for 30 parallel jobs total dequeue time is <1 second. This is much less than 60 seconds polling period. 
   
According to current logic <max processing rate> = <max parallel jobs>/<polling period>. This sounds like artificial limit. 

@tongwu-msft Please confirm.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlQueueClient.cs
```diff
@@ -357,38 +323,61 @@ public async Task ArchiveJobsAsync(byte queueType, CancellationToken cancellatio
...
-                await action(cancellationToken);
+                return await action(cancellationTokenSource);
             }
         }
 
-        private void PutJobHeartbeatFireAndForget(byte queueType, long jobId, long version, CancellationToken cancellationToken)
+        public async Task<string> ExecuteJobWithHeartbeats(JobInfo jobInfo, Func<CancellationTokenSource, Task<string>> action, TimeSpan heartbeatPeriod, CancellationTokenSource cancellationTokenSource)
         {
-            try
+            EnsureArg.IsNotNull(action, nameof(action));
```
PTaladay, line 338: Missing Null check for jobInfo, and TimeSpan?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlQueueClient.cs
```diff
@@ -357,38 +323,61 @@ public async Task ArchiveJobsAsync(byte queueType, CancellationToken cancellatio
             await cmd.ExecuteNonQueryAsync(cancellationToken);
         }
 
-        public async Task ExecuteJobWithHeartbeatAsync(byte queueType, long jobId, long version, Func<CancellationToken, Task> action, TimeSpan heartbeatPeriod, CancellationToken cancellationToken)
+        public async Task<string> ExecuteJobWithHeartbeats(byte queueType, long jobId, long version, Func<CancellationTokenSource, Task<string>> action, TimeSpan heartbeatPeriod, CancellationTokenSource cancellationTokenSource)
```
PTaladay, line 326: Is there a particular reason you removed Async from the method name? That tends to be normal for async methods to add that at the end of the method name.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlQueueClient.cs
```diff
@@ -357,38 +323,61 @@ public async Task ArchiveJobsAsync(byte queueType, CancellationToken cancellatio
...
             {
-                // ignore task cancellation exceptions
+                cancellationTokenSource.Cancel();
             }
-            catch (TaskCanceledException)
+        }
+
+        private void PutJobHeartbeatHeavy(JobInfo jobInfo, CancellationTokenSource cancellationTokenSource)
+        {
+            var cancel = PutJobHeartbeat(jobInfo, cancellationTokenSource.Token).Result;
```
PTaladay, line 357: We should be awaiting here and changing the method to async. Calling ".Result" has a tendency to produce deadlocks because of how the contexts are handled.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlQueueClient.cs
```diff
@@ -357,38 +323,61 @@ public async Task ArchiveJobsAsync(byte queueType, CancellationToken cancellatio
...
             {
-                KeepAliveJobAsync(new JobInfo { QueueType = queueType, Id = jobId, Version = version }, cancellationToken).Wait(cancellationToken);
+                return await action(cancellationTokenSource);
             }
-            catch (AggregateException ex) when (ex.InnerExceptions?.All(x => x is TaskCanceledException || x is OperationCanceledException) == true)
+        }
+
+        private void PutJobHeartbeat(byte queueType, long jobId, long version, CancellationTokenSource cancellationTokenSource)
+        {
+            var cancel = PutJobHeartbeat(new JobInfo { QueueType = queueType, Id = jobId, Version = version }, cancellationTokenSource.Token).Result;
```
PTaladay, line 348: Same here. Should not use ".Result" use await instead and convert to async method call.

src/Microsoft.Health.Fhir.SqlServer/Features/Watchdogs/DefragWatchdog.cs
```diff
@@ -145,19 +148,20 @@ private async Task ExecDefragWithHeartbeatAsync(CancellationToken cancellationTo
...
+                await scopedQueueClient.Value.ExecuteJobWithHeartbeats(
                     QueueType,
                     jobId,
                     job.version,
                     async cancellationSource =>
                     {
                         var split = job.definition.Split(";");
-                        await DefragAsync(split[0], split[1], int.Parse(split[2]), byte.Parse(split[3]) == 1, cancellationSource);
+                        await DefragAsync(split[0], split[1], int.Parse(split[2]), byte.Parse(split[3]) == 1, cancellationSource.Token);
+                        return null;
```
PTaladay, line 159: Recommend returning Task.CompletedTask. If you return null you will most likely get a null reference exception

src/Microsoft.Health.TaskManagement/JobHosting.cs
```diff
@@ -113,137 +98,84 @@ private async Task ExecuteJobAsync(JobInfo jobInfo)
...
-                        await _queueClient.CompleteJobAsync(jobInfo, false, CancellationToken.None);
-                    }
-                    catch (Exception completeEx)
-                    {
-                        _logger.LogError(completeEx, "Job {JobId} failed to complete.", jobInfo.Id);
-                    }
+                var progress = new Progress<string>((result) => { jobInfo.Result = result; });
+
+                var runningJob = useHeavyHeartbeats
+                               ? _queueClient.ExecuteJobWithHeartbeats(
```
PTaladay, line 110: Naming wise would it be better to change this to ExecuteJobWithHeavyHeatbeatsAsync to distinguish it from the other method name? Otherwise, it is just an overloaded method and not entirely clear on the difference between them.

src/Microsoft.Health.Fhir.SqlServer/Features/Watchdogs/DefragWatchdog.cs
SergeyGaluzo: Defrag test passes... I will follow your recommendation anyway...

src/Microsoft.Health.TaskManagement/JobHosting.cs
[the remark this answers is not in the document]
SergeyGaluzo: I entered bug https://microsofthealth.visualstudio.com/Health/_workitems/edit/99455/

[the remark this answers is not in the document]
PTaladay: Makes sense to address in a separate PR.

PTaladay (approved):

PTaladay (dismissed):

PTaladay (dismissed):

src/Microsoft.Health.TaskManagement/JobHosting.cs
```diff
@@ -113,137 +99,120 @@ private async Task ExecuteJobAsync(JobInfo jobInfo)
...
+        public static async Task<string> ExecuteJobWithHeartbeatsAsync(IQueueClient queueClient, byte queueType, long jobId, long version, Func<CancellationTokenSource, Task<string>> action, TimeSpan heartbeatPeriod, CancellationTokenSource cancellationTokenSource)
         {
-            _logger.LogInformation("Start to keep alive job message.");
+            EnsureArg.IsNotNull(queueClient, nameof(queueClient));
+            EnsureArg.IsNotNull(action, nameof(action));
 
-            while (!cancellationToken.IsCancellationRequested)
+            var jobInfo = new JobInfo { QueueType = queueType, Id = jobId, Version = version }; // ignore other data points from job info
+
+            await using (new Timer(async _ => await PutJobHeartbeatAsync(queueClient, jobInfo, cancellationTokenSource), null, TimeSpan.FromSeconds(RandomNumberGenerator.GetInt32(100) / 100.0 * heartbeatPeriod.TotalSeconds), heartbeatPeriod))
```
tongwu-msft, line 191: Any exception in Timer would crash the process, I believe any exception in PutJobHeatbeatAsync would break the process. Repro code here:

            int count = 1;
            await using (new Timer(async _ =>
            {
                Console.WriteLine("Task.Delay(100)");
                if (count++ % 10 == 0)
                {
                    throw new Exception("Test");
                }
                await Task.Delay(100);
            }, null, 1000, 100))
            {
                Console.WriteLine("Test");
                await Task.Delay(TimeSpan.FromMinutes(20));
            }

tongwu-msft (changes requested):

src/Microsoft.Health.TaskManagement/JobHosting.cs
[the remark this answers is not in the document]
SergeyGaluzo: I verified the logic. It works without max mentioned above. Sorry for confusion.

src/Microsoft.Health.TaskManagement/JobHosting.cs
SergeyGaluzo: PutJobHeartbeatAsync has try/catch all block. This should prevent this problem.

SergeyGaluzo: Here is the test code that shows how it works
[Fact]
#pragma warning disable SA1116
#pragma warning disable SA1117
#pragma warning disable SA1513
        public async Task TimerTest()
        {
            var sucess = false;
            var count = 1;
            var bad = 0;
            var good = 0;
            await using (new Timer(async _ =>
            {
                try
                {
                    if (count++ % 2 == 0)
                    {
                        throw new Exception("Test");
                    }
                    await Task.Delay(TimeSpan.FromSeconds(0.01));
                    good++;
                }
                catch
                {
                    bad++;
                }
            }, null, TimeSpan.FromSeconds(0), TimeSpan.FromSeconds(1)))
            {
                await Task.Delay(TimeSpan.FromSeconds(10));
                sucess = true;
            }
            Assert.True(sucess);
            Assert.True(good == 5, $"good: {good} == 5");
            Assert.True(bad == 5, $"bad: {bad} == 5");
        }

SergeyGaluzo: I added redundant try/catch in JobHosting.PutJobHeartbeatAsync.

PTaladay (approved):

tongwu-msft (approved):

PR #3044, opened 2022-12-31
Parallel dequeue

SergeyGaluzo:
Addresses https://microsofthealth.visualstudio.com/Health/_workitems/edit/99473/

### DIFF
--- a/src/Microsoft.Health.TaskManagement/JobHosting.cs
+++ b/src/Microsoft.Health.TaskManagement/JobHosting.cs
@@ -80,7 +80,7 @@ namespace Microsoft.Health.JobManagement
 
             try
             {
-                await Task.WhenAll(workers.ToArray());
+                await Task.WhenAny(workers.ToArray()); // if any worker crashes exit
             }
             catch (Exception ex)
             {
--- a/src/Microsoft.Health.TaskManagement/JobHosting.cs
+++ b/src/Microsoft.Health.TaskManagement/JobHosting.cs
@@ -37,46 +37,50 @@ namespace Microsoft.Health.JobManagement
 
         public int JobHeartbeatTimeoutThresholdInSeconds { get; set; } = Constants.DefaultJobHeartbeatTimeoutThresholdInSeconds;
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.Health.TaskManagement/JobHosting.cs
```diff
@@ -37,46 +37,50 @@ public JobHosting(IQueueClient queueClient, IJobFactory jobFactory, ILogger<JobH
...
+                        }
+
+                        if (nextJob != null)
+                        {
+                            await ExecuteJobAsync(nextJob, useHeavyHeartbeats);
+                        }
+                        else
+                        {
+                            await Task.Delay(TimeSpan.FromSeconds(PollingFrequencyInSeconds), cancellationTokenSource.Token);
+                        }
```
tongwu-msft, line 76: For worker threads, we need to make sure no exception for all operations, I believe redundant try catch is necessary.

SergeyGaluzo: Are you talking about try/catch around ExecuteJobAsync in ExecuteAsync method?
If so, then ExecuteJobAsync is already protected by try/catch logic. I would understand the need for redundant try/catch in ExecuteAsync if ExecuteJobAsync was a method from a different class. But ExecuteJobAsync method is private, so its sole purpose is to be right for JobHosting class. 
Adding redundant try/catch in this particular case does not make sense to me.

tongwu-msft: I mean there could be some other operations like _queueClient.IsInitialized() and possible further more.

SergeyGaluzo: @tongwu-msft Did these problems exist in the original code (before PR)?

tongwu-msft: Previous if there's any exception throw, the JobHosting would stop and the process for PaaS would be able to restart to recover, if we wait for all workers complete then my concern if any exception in 1 worker would only stop this thread and would not recover.

tongwu-msft (approved):

PTaladay (approved):

PR #3051, opened 2023-01-06
New retry logic

Tom-Markoch:
## Description
Adds new retry logic which on failure recreates the SQL connection and re-executes the SQL command.

## Related issues
Addresses #99416, #99196, #93863, #99169.

## Testing
Set of unit tests were created that test retries on SQL failures and connection failures. Manual testing including DB scaling.

## FHIR Team Checklist


src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -293,201 +298,211 @@ private async Task<SearchResult> SearchImpl(SqlSearchOptions sqlSearchOptions, S
...
-            using (SqlCommandWrapper sqlCommandWrapper = sqlConnectionWrapper.CreateRetrySqlCommand())
+            using (SqlCommand sqlCommand = new SqlCommand()) // WARNING, this code will not set sqlCommand.Transaction. Sql transactions via C#/.NET are not supported in this method.
             {
                 var exportTimeTravel = clonedSearchOptions.QueryHints != null && _schemaInformation.Current >= SchemaVersionConstants.ExportTimeTravel;
                 if (exportTimeTravel)
                 {
-                    PopulateSqlCommandFromQueryHints(clonedSearchOptions, sqlCommandWrapper);
-                    sqlCommandWrapper.CommandTimeout = 1200; // set to 20 minutes, as dataset is usually large
+                    PopulateSqlCommandFromQueryHints(clonedSearchOptions, sqlCommand);
+                    sqlCommand.CommandTimeout = 1200; // set to 20 minutes, as dataset is usually large // TODO: read from same place where other timeout is read from (settings?)
```
SergeyGaluzo, line 307: >// TODO: read from same place where other timeout is read from (settings?) [](http://example.com/codeflow?start=104&length=74)

There are many places where individual timeouts per command and per usage are set. Common settings does not sound like good place for this data.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetry.cs
```diff
@@ -0,0 +1,280 @@
...
+        public int MaxRetries { get; set; } = 5;
+
+        public int RetryMillisecondsDelay { get; set; } = 5000;
+
+#pragma warning disable CA1002 // Do not expose generic lists
+        public List<int> RemoveTransientErrors { get; } = new List<int>();
+#pragma warning restore CA1002 // Do not expose generic lists
+
+#pragma warning disable CA1002 // Do not expose generic lists
+        public List<int> AddTransientErrors { get; } = new List<int>();
```
SergeyGaluzo, line 45: >List<int> [](http://example.com/codeflow?start=15&length=9)

Can you expose IList?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetry.cs
```diff
@@ -0,0 +1,280 @@
...
+
+        public delegate T InitializeRowProcessor<T>();
+
+        public delegate void SqlConnectionInitializer(SqlConnection sqlConnection);
+
+        // TODO also define toT action delegate.
+
+        private static bool DefaultIsExceptionRetriable(Exception ex)
+        {
+            if (ex is SqlException sqlEx)
```
SergeyGaluzo, line 149: >if (ex is SqlException sqlEx) [](http://example.com/codeflow?start=12&length=29)

Is this 
Microsoft.Data.SqlClient.SqlException (0x80131904): A connection was successfully established with the server, but then an error occurred during the login process. (provider: TCP Provider, error: 35 - An internal exception was caught)

retried?

SergeyGaluzo, line 149: >if (ex is SqlException sqlEx) [](http://example.com/codeflow?start=12&length=29)

Is this 

Microsoft.Data.SqlClient.SqlException (0x80131904): The connection is broken and recovery is not possible.  The client driver attempted to recover the connection one or more times and all attempts failed.  Increase the value of ConnectRetryCount to increase the number of recovery attempts.

retried?

SergeyGaluzo, line 149: >if (ex is SqlException sqlEx) [](http://example.com/codeflow?start=12&length=29)

Is this

Microsoft.Data.SqlClient.SqlException (0x80131904): A connection was successfully established with the server, but then an error occurred during the login process. (provider: TCP Provider, error: 0 - No error information)

retried?

SergeyGaluzo, line 149: >if (ex is SqlException sqlEx) [](http://example.com/codeflow?start=12&length=29)

Is this 

Microsoft.Data.SqlClient.SqlException (0x80131904): A severe error occurred on the current command.  The results, if any, should be discarded.

retried?

SergeyGaluzo, line 149: Is this

System.InvalidOperationException: Timeout expired.  The timeout period elapsed prior to obtaining a connection from the pool.

retried?

SergeyGaluzo, line 149: Is this

A transport-level error has occurred when receiving results from the server

retried?

SergeyGaluzo, line 149: Is this

Microsoft.Data.SqlClient.SqlException (0x80131904): Connection Timeout Expired.  The timeout period elapsed during the post-login phase. 

retried?

SergeyGaluzo, line 149: Is this

Microsoft.Data.SqlClient.SqlException (0x80131904): A network-related or instance-specific error occurred while establishing a connection to SQL Server. The server was not found or was not accessible

retried?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetry.cs
```diff
@@ -0,0 +1,280 @@
...
+            return false;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public async Task ExecuteSqlWithRetries(SqlCommand sqlCommand, SqlConnectionInitializer sqlConnectionInitializer, ExecuteSqlWithRetriesAction action, CancellationToken cancellationToken)
+        {
+            EnsureArg.IsNotNull(sqlCommand, nameof(sqlCommand));
+            EnsureArg.IsNotNull(action, nameof(action));
+
+            sqlCommand.CommandTimeout = (int)_sqlServerDataStoreConfiguration.CommandTimeout.TotalSeconds;
```
SergeyGaluzo, line 196: >sqlCommand.CommandTimeout = (int)_sqlServerDataStoreConfiguration.CommandTimeout.TotalSeconds; [](http://example.com/codeflow?start=12&length=94)

We should either preserve value on input sql command or provide parameter to override configuration value.

SergeyGaluzo:
I think this change should go through stress test before check in. 
I suggest to use $export to do the testing.

Prerequisites:
1. Successful R4 docker build.
2. Main job queue methods have retries added.

SergeyGaluzo:
It would be helpful to see code changes related to retries in simple case. 
Good method in 
public async Task<JobInfo> DequeueAsync
which we need to do anyway for stress test.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetry.cs
```diff
@@ -0,0 +1,280 @@
...
+            {
+                return true;
+            }
+
+            if (customIsExceptionRetriable(ex))
+            {
+                return true;
+            }
+
+            if (ex.IsRetryable())
```
SergeyGaluzo, line 182: >ex.IsRetryable [](http://example.com/codeflow?start=16&length=14)

Please fix spelling

SergeyGaluzo:
We cannot use retries for individual SQL calls (that is what is implemented so far) when these calls are not idempotent, but we still want to make service level calls idempotent. 
Good example is put resource logic (see SqlServerFhirDataStore.UpsertAsync), where service logic relies on 2 SQL calls - get and and "pure" put. This ensures that service call is idempotent, while each of smaller calls is not.
To handle this and similar situations without major redesign we want to enable retries across large block of C# code which also includes re-building individual SQL commands.
Please add generic method ExecuteWithRetries with Action as input parameter.

SergeyGaluzo:
Once we added generic ExecuteWithRetries method I would reconsider the need to expose granular ProcessSqlDataReaderRowsWithRetries() as public. 
My point is that retry logic can be much simpler, and, at the same time, required changes to add retries will be minimum. 
We can use this generic logic as a simple wrapper on existing code (SearchImple is a good example) almost with no changes.

SergeyGaluzo (changes requested):
:clock1:

brendankowitz:
After discussing with @SergeyGaluzo, I think we took a higher level approach in might be simpler and more generic. i.e. creating a `RetryingSqlServerSearchService` that contains the method level retry logic. This would separate the concerns and would not add any additional complexity to the existing search service. Thoughts?

SergeyGaluzo:
@brendankowitz @Tom-Markoch 
I believe I have simpler view on this. To avoid confusion I expressed this view in this [https://github.com/microsoft/fhir-server/pull/3062](https://github.com/microsoft/fhir-server/pull/3062) PR. Please take a look.

---
In reply to: [1377727759](https://github.com/microsoft/fhir-server/pull/3051#issuecomment-1377727759) [](http://example.com/codeflow?ancestors=1377727759,1377727759)

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
Tom-Markoch: I think things like timings are usually set in configs, not hard coded into the source code.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetry.cs
Tom-Markoch: By the way, how come we have win32 error codes (like 0x80131904) for these but not SqlException error numbers?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetry.cs
Tom-Markoch: This comes from same place currently timeout comes from. So, we are now hardcoding timeouts into the source code?

SergeyGaluzo: Each method that invokes particular SQL command has knowledge about data volume and, therefore, expectations on how long database call should last. Even the same SQL command invoked with different parameters values might yield substantially different results that might require different command timeouts. Only code knows about parameter values, so code should have ability to overwrite default configuration value. This could be simple "hardcoding" on per command basis or more advanced.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetry.cs
SergeyGaluzo: This is just what exception.ToString() returns.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
SergeyGaluzo: See me response below. This is not that trivial.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetry.cs
Tom-Markoch: Then maybe we should stop using simple "ToString()" so we can have the actual Sql error numbers, not win32 errors.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetry.cs
Tom-Markoch: OK, just to make it clear, so we are hardcoding timeouts like above "1200" into the source code?

Tom-Markoch:
I don't like that approach:
1. It unnecessarily executes code that does not have to be executed, so it's lower efficiency.
2. Worse, there may be side-effects with executing some of the code again. This was built to run once per request and there are states that get set once code is executed.
3. Retry logic is defined at multiple places, and code may diverge over time.

Tom-Markoch:
Also, I need clarity on exception handling. Are we really going to use only string compare? Some of raised errors like comm might come from the OS or various libs like .NET. This would mean that from now Fhir server will work on only English version of Windows OS (linux may throw different messages), English version of libs, English version of SQL server.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetry.cs
SergeyGaluzo: That is fine by me. My point is that those exceptions must be still handled.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetry.cs
SergeyGaluzo: On the call/value level - yes.

SergeyGaluzo:
Addressing bullet points:
1. Retries are very infrequent events, so we should not be concerned with performance as long it is not significantly worse than on the first pass.
2. I agree, that we should not use generic retries blindly. We should test them. I just do not see good alternative for generic retries which still can handle multiple SQL calls inside a retry block.
3. I agree, we should not use individual retries inside a generic retry block. All commands inside generic retry block should not have retries on their own.    

---
In reply to: [1379338129](https://github.com/microsoft/fhir-server/pull/3051#issuecomment-1379338129) [](http://example.com/codeflow?ancestors=1379338129,1379338129)

SergeyGaluzo:
Requirements:
1. All exceptions (I put them as comments to this PR) must be retried. BTW Most of them (maybe all) are not part of current list of error numbers.
2. We cannot trust error numbers on exception, as they are not globally unique (same number points to different exceptions, some retriable and some not). Therefore we need to use exception message.

We can argue whether we can rely on only message only or on a combination of number and message, but in any case messages must be used. 

@brendankowitz What are your thoughts? 

---
In reply to: [1379342999](https://github.com/microsoft/fhir-server/pull/3051#issuecomment-1379342999) [](http://example.com/codeflow?ancestors=1379342999,1379342999)

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlServerFhirStorageTestsFixture.cs
```diff
@@ -81,7 +81,14 @@ internal SqlServerFhirStorageTestsFixture(int maximumSupportedSchemaVersion, str
             TestConnectionString = new SqlConnectionStringBuilder(initialConnectionString) { InitialCatalog = _databaseName }.ToString();
 
             var schemaOptions = new SqlServerSchemaOptions { AutomaticUpdatesEnabled = true };
-            var config = Options.Create(new SqlServerDataStoreConfiguration { ConnectionString = TestConnectionString, Initialize = true, SchemaOptions = schemaOptions, StatementTimeout = TimeSpan.FromMinutes(10) });
+            var config = Options.Create(new SqlServerDataStoreConfiguration
+            {
+                ConnectionString = TestConnectionString,
+                Initialize = true,
+                SchemaOptions = schemaOptions,
+                StatementTimeout = TimeSpan.FromMinutes(10),
```
SergeyGaluzo, line 89: >StatementTimeout = TimeSpan.FromMinutes(10), [](http://example.com/codeflow?start=16&length=44)

Now that formatted this such that I can see all :-), curious how statement timeout is different from command timeout...

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -55,6 +57,9 @@ internal class SqlServerSearchService : SearchService
         private readonly BitColumn _isMatch = new BitColumn("IsMatch");
         private readonly BitColumn _isPartial = new BitColumn("IsPartial");
         private readonly SqlConnectionWrapperFactory _sqlConnectionWrapperFactory;
```
SergeyGaluzo, line 59: > private readonly SqlConnectionWrapperFactory _sqlConnectionWrapperFactory; [](http://example.com/codeflow?start=7&length=75)

Can you remove this completely so we have one fully clean class?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -293,209 +306,234 @@ private async Task<SearchResult> SearchImpl(SqlSearchOptions sqlSearchOptions, S
...
+                        SqlCommandSimplifier.RemoveRedundantParameters(stringBuilder, sqlCommand.Parameters, _logger);
 
-                    var resources = new List<SearchResultEntry>(sqlSearchOptions.MaxItemCount);
-                    short? newContinuationType = null;
-                    long? newContinuationId = null;
-                    bool moreResults = false;
-                    int matchCount = 0;
+#pragma warning disable CA2100 // Review SQL queries for security vulnerabilities
+                        sqlCommand.CommandText = stringBuilder.ToString();
+#pragma warning restore CA2100 // Review SQL queries for security vulnerabilities
```
SergeyGaluzo, line 351: >Review SQL queries for security vulnerabilities [](http://example.com/codeflow?start=34&length=47)

Add "no direct user input" in the comments

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetryService.cs
```diff
@@ -0,0 +1,147 @@
...
+
+            if (ex.IsRetriable())
+            {
+                return true;
+            }
+
+            return false;
+        }
+
+        public async Task ExecuteWithRetries(RetriableAction action)
```
SergeyGaluzo, line 123: >public async Task ExecuteWithRetries(RetriableAction action) [](http://example.com/codeflow?start=8&length=60)

Please add back reader method so people do not make common mistakes with IEnumerable and need to call next result.

SergeyGaluzo (changes requested):
:clock1:

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlQueueClient.cs
```diff
@@ -154,125 +153,89 @@ public async Task<IReadOnlyCollection<JobInfo>> DequeueJobsAsync(byte queueType,
...
-                using SqlCommandWrapper sqlCommandWrapper = sqlConnectionWrapper.CreateRetrySqlCommand();
-
-                // cannot use VLatest as it incorrectly asks for optional @InputJobId
-                sqlCommandWrapper.CommandText = "dbo.DequeueJob";
-                sqlCommandWrapper.CommandType = System.Data.CommandType.StoredProcedure;
-                sqlCommandWrapper.Parameters.AddWithValue("@QueueType", queueType);
-                sqlCommandWrapper.Parameters.AddWithValue("@Worker", worker);
-                sqlCommandWrapper.Parameters.AddWithValue("@HeartbeatTimeoutSec", heartbeatTimeoutSec);
-                if (jobId.HasValue)
+            return await _sqlRetryService.ExecuteSqlCommandFuncWithRetries(
```
SergeyGaluzo, line 156: >ExecuteSqlCommandFuncWithRetries [](http://example.com/codeflow?start=42&length=32)

We should use ExecuteSqlReaderWithRetries here which should enforce correct usage of sql reader. 
Current code is incorrect as it does not do reader.NextResultAsync(cancellationToken). 
I added comment about this before. Please address.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlQueueClient.cs
```diff
@@ -154,125 +153,89 @@ public async Task<IReadOnlyCollection<JobInfo>> DequeueJobsAsync(byte queueType,
...
-                }
-            }
-            catch (Exception ex)
-            {
-                _logger.LogError(ex, "EnqueueAsync failed.");
-                if (ex.IsRetryable())
-                {
-                    throw new RetriableJobException(ex.Message, ex);
-                }
+                        return await sqlDataReader.ReadJobInfosAsync(cancellationToken);
```
SergeyGaluzo, line 194: Where is reader.NextResult()?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlQueueClient.cs
```diff
@@ -152,137 +150,101 @@ public async Task<IReadOnlyCollection<JobInfo>> DequeueJobsAsync(byte queueType,
...
+                            throw new JobManagement.JobConflictException(sqlEx.Message);
+                        }
 
-                throw;
-            }
+                        throw;
+                    }
+                },
+                _logger,
+                "EnqueueAsync failed.",
```
SergeyGaluzo, line 216: >"EnqueueAsync failed.", [](http://example.com/codeflow?start=16&length=23)

This message can be predicted. Why do we need this parameter?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlQueueClient.cs
```diff
@@ -340,5 +302,31 @@ public async Task<bool> PutJobHeartbeatAsync(JobInfo jobInfo, CancellationToken
...
+
+            public void PopulateCommand(SqlCommand sqlCommand, byte queueType, long? jobId, global::System.Collections.Generic.IEnumerable<BigintListRow> jobIds, long? groupId, bool? returnDefinition)
+            {
+                sqlCommand.CommandType = global::System.Data.CommandType.StoredProcedure;
+                sqlCommand.CommandText = "dbo.GetJobs";
+                _queueType.AddParameter(sqlCommand.Parameters, queueType);
+                _jobId.AddParameter(sqlCommand.Parameters, jobId);
+                _jobIds.AddParameter(sqlCommand.Parameters, jobIds);
+                _groupId.AddParameter(sqlCommand.Parameters, groupId);
+                _returnDefinition.AddParameter(sqlCommand.Parameters, returnDefinition);
```
SergeyGaluzo, line 328: >_returnDefinition.AddParameter(sqlCommand.Parameters, returnDefinition); [](http://example.com/codeflow?start=16&length=72)

This parameter is optional. Please do not set it.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlQueueClient.cs
```diff
@@ -152,105 +153,64 @@ public async Task<IReadOnlyCollection<JobInfo>> DequeueJobsAsync(byte queueType,
...
-                            throw;
-                        }
-                    }
-                }
+            JobInfo jobInfo = await _sqlRetryService.ExecuteSqlDataReaderFirstRow(
+                sqlCommand,
+                JobInfoExtensions.LoadJobInfo,
+                _logger,
+                "DequeueAsync failed.",
+                cancellationToken);
```
brendankowitz, line 174: This is removing previously added behavior of retrying because of dequeue conflicts

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlQueueClient.cs
```diff
@@ -360,5 +300,29 @@ public async Task<bool> PutJobHeartbeatAsync(JobInfo jobInfo, CancellationToken
 
             return cancel;
         }
+
+        // Class copied from src\Microsoft.Health.Fhir.SqlServer\Features\Schema\Model\VLatest.Generated.cs .
```
brendankowitz, line 304: Can you add the comment why we did this?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetryService.cs
```diff
@@ -0,0 +1,247 @@
...
+
+            if (ex.IsRetriable())
+            {
+                return true;
+            }
+
+            return false;
+        }
+
+        public async Task ExecuteSql(Func<CancellationToken, Task> action, CancellationToken cancellationToken)
```
brendankowitz, line 134: Prefer: Add method documentation on public methods

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetryService.cs
```diff
@@ -0,0 +1,247 @@
...
+            _maxRetries = sqlRetryServiceOptions.Value.MaxRetries;
+            _retryMillisecondsDelay = sqlRetryServiceOptions.Value.RetryMillisecondsDelay;
+
+            _defaultIsExceptionRetriableOff = sqlRetryServiceDelegateOptions.DefaultIsExceptionRetriableOff;
+            _customIsExceptionRetriable = sqlRetryServiceDelegateOptions.CustomIsExceptionRetriable;
+        }
+
+        public delegate bool IsExceptionRetriable(Exception ex);
+
+        private static bool DefaultIsExceptionRetriable(Exception ex)
```
brendankowitz, line 91: This name seems confusing to me

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetryService.cs
```diff
@@ -0,0 +1,247 @@
...
+            EnsureArg.IsNotNull(action, nameof(action));
+            EnsureArg.IsNotNull(logger, nameof(logger));
+            EnsureArg.IsNotNull(logMessage, nameof(logMessage));
+
+            int retry = 0;
+            while (true)
+            {
+                try
+                {
+                    // WARNING, this code will not set sqlCommand.Transaction. Sql transactions via C#/.NET are not supported in this method.
```
brendankowitz, line 170: Should be the method documentation

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
Tom-Markoch: SqlConnectionWrapperFactory not used any more

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
Tom-Markoch: Done

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetryService.cs
Tom-Markoch: Done

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlQueueClient.cs
Tom-Markoch: Logging message if retries. Info log for each retry except last one which is log error.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlQueueClient.cs
Tom-Markoch: Retries are now inside _sqlRetryService.ExecuteSqlDataReaderFirstRow.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlQueueClient.cs
Tom-Markoch: I'll add comment to the source file. This comes from the autogenerated file which uses SqlCcommandWrapper as an input param. We are not using SqlCcommandWrapper any more, we are using SqlCommand. Sergey suggested to just copy autogenerated code here and modify instead of looking how to change autogenerated code.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetryService.cs
Tom-Markoch: Added it to the class doc since it applies to all the methods.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetryService.cs
Tom-Markoch: Added method and class documentation.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetryService.cs
Tom-Markoch: DOne

SergeyGaluzo (dismissed):
:shipit:

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetry.cs
Tom-Markoch: Done

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetry.cs
Tom-Markoch: Done

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetryServiceOptions.cs
```diff
@@ -0,0 +1,36 @@
...
+
+namespace Microsoft.Health.Fhir.SqlServer.Features.Storage
+{
+    /// <summary>
+    /// Provides SQL retry parameters Used to initialize SqlRetryService.
+    /// </summary>
+    public class SqlRetryServiceOptions
+    {
+        /// <summary>Used to bind this class to .NRT options data source.</summary>
+        public const string SqlServer = "SqlServer";
```
fhibf, line 16: Low priority suggestion: you can use "SqlServer = nameof(SqlServer);"

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/JobInfoExtensions.cs
```diff
@@ -39,7 +39,7 @@ public static async Task<IReadOnlyList<JobInfo>> ReadJobInfosAsync(this SqlDataR
             return outcome;
         }
 
-        private static JobInfo LoadJobInfo(SqlDataReader sqlDataReader)
+        public static JobInfo LoadJobInfo(SqlDataReader sqlDataReader)
```
fhibf, line 42: I believe that, as we are making this method public, we should validate if 'sqlDataReader' is not null.

fhibf (approved):

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/JobInfoExtensions.cs
Tom-Markoch: Done

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetryServiceOptions.cs
Tom-Markoch: This is supposed to be a name of a section in a json file. Shouldn't really be tied to any class. Microsoft recommends this pattern for IOptions initialization: https://learn.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-7.0 ,  Also went through all of our startup code, every place uses same pattern: .GetSection(plain-string).Bind(class-instance);

But I found something else that needs to be resolved once FhirServer is fully transitioned to the new retry logic. I left a big TODO comment in the code. Keep in mind that this PR is just a partial transition to the new retries, there's still work to be done, but we needed to get export functionality transitioned first.

fhibf (approved):

PR #3063, opened 2023-01-10
Changes group not found exception to an info log

Tom-Markoch:
## Description
During exports, if nonexistent group is used then exception is logged as info instead of error.

## Related issues
Internal 97939.

## Testing
Unit test was built to verify logging is done as Info.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Export/ExportJobTaskTests.cs
+++ b/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Export/ExportJobTaskTests.cs
@@ -8,6 +8,7 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.Linq.Expressions;
 using System.Net;
+using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 using Hl7.Fhir.ElementModel;
@@ -69,6 +70,10 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Operations.Export
             _cancellationToken = _cancellationTokenSource.Token;
             SetupExportJobRecordAndOperationDataStore();
             _exportJobTask = CreateExportJobTask();
+
+            IModelInfoProvider modelInfoProvider = Substitute.For<IModelInfoProvider>();
+            modelInfoProvider.IsKnownResource("Group").Returns(true);
+            ModelInfoProvider.SetProvider(modelInfoProvider);
         }
 
         [Fact]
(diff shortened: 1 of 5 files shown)


LTA-Thinking (dismissed):

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Export/ExportJobTaskTests.cs
```diff
@@ -986,6 +987,31 @@ public async Task GivenAnExportJobToResumeWithTheTypeParameter_WhenExecuted_Then
             Assert.Equal(OperationStatus.Completed, _exportJobRecord.Status);
         }
 
+        [Fact]
```
fhibf, line 990: Should we create/adapt a test where NoGroupCount is expected to be zero?

src/Microsoft.Health.Fhir.Core/Exceptions/ResourceNotFoundException.cs
```diff
@@ -10,15 +10,18 @@ namespace Microsoft.Health.Fhir.Core.Exceptions
 {
     public class ResourceNotFoundException : FhirException
```
fhibf, line 11: Instead of extending this exception, wouldn't it be better to create a new exception called GroupNotFoundException? 

I'm suggesting that because ResourceNotFoundException is probably used in many other scenarios, and adding a property/parameter called groupNotFound makes it too specific to a single case.

fhibf: GroupNotFoundException would inherit from ResourceNotFoundException.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Export/GroupMemberExtractor.cs
```diff
@@ -96,7 +96,7 @@ private async Task<HashSet<string>> GetGroupPatientIdsHelper(string groupId, Dat
 
             if (groupResource == null)
             {
-                throw new ResourceNotFoundException($"Group {groupId} was not found.");
+                throw new ResourceNotFoundException($"Group {groupId} was not found.", true);
```
fhibf, line 99: Raise GroupNotFoundException?

ResourceNotFoundException is a generic exception today. If we add a new parameter called GroupNotFound, then it'll make the parameter useless for most of all its usages.

fhibf (changes requested):
Waiting for improvements in ResourceNotFoundException

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Export/ExportJobTaskTests.cs
Tom-Markoch: Done

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Export/GroupMemberExtractor.cs
Tom-Markoch: We had long conversation and changed plan altogether.

LTA-Thinking (approved):

fhibf (approved):

PR #3073, opened 2023-01-17
Set merge

SergeyGaluzo:


src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -191,52 +195,33 @@ internal class SqlServerFhirDataStore : IFhirDataStore, IProvideCapability
...
+                                AffectedRows: 0,
+                                RaiseExceptionOnConflict: true,
+                                IsResourceChangeCaptureEnabled: _coreFeatures.SupportsResourceChangeCapture,
+                                tableValuedParameters: _mergeResourcesTvpGeneratorVLatest.Generate(new List<MergeResourceWrapper> { new MergeResourceWrapper(resource, surrId, keepHistory, true) }));
+                            await sqlCommandWrapper.ExecuteNonQueryAsync(cancellationToken);
                         }
                         else
                         {
-                            saveOutcomeType = SaveOutcomeType.Updated;
-                            resource.RawResource.IsMetaSet = false;
```
SergeyGaluzo, line 236: >resource.RawResource.IsMetaSet = false; [](http://example.com/codeflow?start=28&length=39)

This line requires calculation of Meta for updated resource on each resource read. Meta computation is 4x more expensive than unzip. This causes huge performance degradation.

src/Microsoft.Health.Fhir.Core/Features/Persistence/ResourceWrapper.cs
```diff
@@ -75,10 +75,10 @@ protected ResourceWrapper()
         }
 
         [JsonProperty(KnownResourceWrapperProperties.LastModified)]
-        public DateTimeOffset LastModified { get; protected set; }
+        public DateTimeOffset LastModified { get; internal set; }
 
         [JsonProperty(KnownResourceWrapperProperties.RawResource)]
-        public RawResource RawResource { get; set; }
+        public RawResource RawResource { get; internal set; }
```
SergeyGaluzo, line 81: >internal set [](http://example.com/codeflow?start=46&length=12)

I needed to update RawResource and LastModified and this class was already not immutable.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/MergeResources.sql
```diff
@@ -0,0 +1,257 @@
...
+
+  INSERT INTO dbo.Resource 
+         ( ResourceTypeId, ResourceId, Version, IsHistory, ResourceSurrogateId, IsDeleted, RequestMethod, RawResource, IsRawResourceMetaSet, SearchParamHash )
+    SELECT ResourceTypeId, ResourceId, Version, IsHistory, ResourceSurrogateId, IsDeleted, RequestMethod, RawResource, IsRawResourceMetaSet, SearchParamHash
+      FROM @Resources
+  SET @AffectedRows += @@rowcount
+
+  INSERT INTO dbo.ResourceWriteClaim 
+         (          ResourceSurrogateId, ClaimTypeId, ClaimValue )
+    SELECT DISTINCT ResourceSurrogateId, ClaimTypeId, ClaimValue
```
SergeyGaluzo, line 142: >DISTINCT [](http://example.com/codeflow?start=11&length=8)

I hope DISTINCT is temporary and will be removed when bug ([https://microsofthealth.visualstudio.com/Health/_workitems/edit/99866/](https://microsofthealth.visualstudio.com/Health/_workitems/edit/99866/)) is fixed.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -440,11 +425,56 @@ private static string RemoveTrailingZerosFromMillisecondsForAGivenDate(DateTimeO
             return formattedDate.Replace(milliseconds, trimmedMilliseconds, StringComparison.Ordinal);
         }
 
-        private static string RemoveVersionIdAndLastUpdatedFromMeta(ResourceWrapper resourceWrapper)
```
SergeyGaluzo, line 443: >RemoveVersionIdAndLastUpdatedFromMeta [](http://example.com/codeflow?start=30&length=37)

This method creates 2 more large strings. So for any update memory footprint = (2 versions) + 2 extra string per version *2 = 6x.

SergeyGaluzo:
>         private static string RemoveTrailingZerosFromMillisecondsForAGivenDate(DateTimeOffset date)

I am not sure why we need to have this special method. There should be a place that converts date to desired format when we create raw resource string. Why cannot we use that conversion here too?

---
Refers to: src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs:427 in e78dbeb. [](commit_id = e78dbeb04876519e13ca8a406b3a570afdeed8f9, deletion_comment = True)

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/Merge/ResourceListRowGenerator.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        public IEnumerable<ResourceListRow> GenerateRows(IReadOnlyList<MergeResourceWrapper> mergeWrappers)
+        {
+            foreach (var merge in mergeWrappers)
+            {
+                var wrapper = merge.ResourceWrapper;
+                var stream = new MemoryStream();
+                _compressedRawResourceConverter.WriteCompressedRawResource(stream, wrapper.RawResource.Data);
+                stream.Seek(0, 0);
+                yield return new ResourceListRow(_model.GetResourceTypeId(wrapper.ResourceTypeName), merge.ResourceSurrogateId, wrapper.ResourceId, int.Parse(wrapper.Version), merge.HasVersionToCompare, wrapper.IsDeleted, wrapper.IsHistory, merge.KeepHistory, stream, wrapper.RawResource.IsMetaSet, wrapper.Request.Method, wrapper.SearchParameterHash);
```
SergeyGaluzo, line 34: >_model.GetResourceTypeId(wrapper.ResourceTypeName), merge.ResourceSurrogateId [](http://example.com/codeflow?start=49&length=77)

Main difference between old and new is that resource set can have multiple resource types and surrogate id is set before MergeResources call.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Tables/UriSearchParam.sql
```diff
@@ -6,9 +6,11 @@
     Uri varchar(256) COLLATE Latin1_General_100_CS_AS NOT NULL,
     IsHistory bit NOT NULL
 )
-
+GO
+ALTER TABLE dbo.UriSearchParam ADD CONSTRAINT DF_UriSearchParam_IsHistory DEFAULT 0 FOR IsHistory
```
SergeyGaluzo, line 10: This is done in preparation of removing IsHistory column all together.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/49.diff.sql
```diff
@@ -0,0 +1,587 @@
...
+GO
+IF NOT EXISTS (SELECT * FROM sys.types WHERE name = 'TokenTextList')
+CREATE TYPE dbo.TokenTextList AS TABLE
+(
+    ResourceTypeId           smallint NOT NULL
+   ,ResourceSurrogateId      bigint   NOT NULL
+   ,SearchParamId            smallint NOT NULL
+   ,Text                     nvarchar(400) COLLATE Latin1_General_CI_AI NOT NULL
+
+   --PRIMARY KEY (ResourceTypeId, ResourceSurrogateId, SearchParamId, Text)
```
fhibf, line 79: Should we leave this statement as a comment?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/49.diff.sql
```diff
@@ -0,0 +1,587 @@
...
+   ,QuantityCodeId2          int      NULL
+   ,SingleValue2             decimal(18,6) NULL
+   ,LowValue2                decimal(18,6) NULL
+   ,HighValue2               decimal(18,6) NULL
+)
+GO
+IF NOT EXISTS (SELECT * FROM sys.types WHERE name = 'TokenNumberNumberCompositeSearchParamList')
+CREATE TYPE dbo.TokenNumberNumberCompositeSearchParamList AS TABLE
+(
+    ResourceTypeId           smallint NOT NULL
```
fhibf, line 125: Fix tabulation.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/49.diff.sql
```diff
@@ -0,0 +1,587 @@
...
+)
+GO
+IF NOT EXISTS (SELECT * FROM sys.types WHERE name = 'ResourceWriteClaimList')
+CREATE TYPE dbo.ResourceWriteClaimList AS TABLE
+(
+    ResourceSurrogateId      bigint        NOT NULL
+   ,ClaimTypeId              tinyint       NOT NULL
+   ,ClaimValue               nvarchar(128) NOT NULL
+
+   --PRIMARY KEY (ResourceSurrogateId, ClaimTypeId, ClaimValue)
```
fhibf, line 173: Should we leave it as a comment?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/49.diff.sql
```diff
@@ -0,0 +1,587 @@
...
+   ,SystemId1                 int      NULL
+   ,Code1                     varchar(256) COLLATE Latin1_General_100_CS_AS NOT NULL
+   ,CodeOverflow1             varchar(max) COLLATE Latin1_General_100_CS_AS NULL
+   ,StartDateTime2            datetimeoffset(7) NOT NULL
+   ,EndDateTime2              datetimeoffset(7) NOT NULL
+   ,IsLongerThanADay2         bit      NOT NULL
+)
+GO
+IF NOT EXISTS (SELECT * FROM sys.types WHERE name = 'StringSearchParamList')
+CREATE TYPE dbo.StringSearchParamList AS TABLE
```
fhibf, line 155: I've noticed that some of these types don't have primary keys, should all they have it?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/49.diff.sql
SergeyGaluzo: I would prefer not to spend any time on on-off scripts which should go away with adding SQL project.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/49.diff.sql
```diff
@@ -0,0 +1,587 @@
...
+    PRIMARY KEY (ResourceTypeId, ResourceSurrogateId)
+   ,UNIQUE (ResourceTypeId, ResourceId, Version)
+)
+GO
+CREATE OR ALTER PROCEDURE dbo.CaptureResourceIdsForChanges @Ids dbo.ResourceIdForChangesList READONLY
+AS
+set nocount on
+-- This procedure is intended to be called from the MergeResources procedure and relies on its transaction logic
+INSERT INTO dbo.ResourceChangeData 
+       ( ResourceId, ResourceTypeId, ResourceVersion,                                              ResourceChangeTypeId )
```
fhibf, line 295: Extra space characters.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/49.diff.sql
```diff
@@ -0,0 +1,587 @@
...
+  SELECT ResourceId, ResourceTypeId,         Version, CASE WHEN IsDeleted = 1 THEN 2 WHEN Version > 1 THEN 1 ELSE 0 END
+    FROM @Ids
+GO
+
+--DROP PROCEDURE dbo.MergeResourcesBeginTransaction
```
fhibf, line 300: Should we remove this comment?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/49.diff.sql
```diff
@@ -0,0 +1,587 @@
...
+    DELETE FROM dbo.TokenDateTimeCompositeSearchParam WHERE EXISTS (SELECT * FROM @PreviousSurrogateIds WHERE TypeId = ResourceTypeId AND SurrogateId = ResourceSurrogateId)
+    SET @AffectedRows += @@rowcount
+    DELETE FROM dbo.TokenQuantityCompositeSearchParam WHERE EXISTS (SELECT * FROM @PreviousSurrogateIds WHERE TypeId = ResourceTypeId AND SurrogateId = ResourceSurrogateId)
+    SET @AffectedRows += @@rowcount
+    DELETE FROM dbo.TokenStringCompositeSearchParam WHERE EXISTS (SELECT * FROM @PreviousSurrogateIds WHERE TypeId = ResourceTypeId AND SurrogateId = ResourceSurrogateId)
+    SET @AffectedRows += @@rowcount
+    DELETE FROM dbo.TokenNumberNumberCompositeSearchParam WHERE EXISTS (SELECT * FROM @PreviousSurrogateIds WHERE TypeId = ResourceTypeId AND SurrogateId = ResourceSurrogateId)
+    SET @AffectedRows += @@rowcount
+
+    --EXECUTE dbo.LogEvent @Process=@SP,@Mode=@Mode,@Status='Info',@Start=@st,@Rows=@AffectedRows,@Text='Old rows'
```
fhibf, line 461: Should we remove this comment?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/CaptureResourceIdsForChanges.sql
```diff
@@ -0,0 +1,9 @@
+﻿CREATE PROCEDURE dbo.CaptureResourceIdsForChanges @Ids dbo.ResourceIdForChangesList READONLY
+AS
+set nocount on
+-- This procedure is intended to be called from the MergeResources procedure and relies on its transaction logic
+INSERT INTO dbo.ResourceChangeData 
+       ( ResourceId, ResourceTypeId, ResourceVersion,                                              ResourceChangeTypeId )
```
fhibf, line 6: Extra spaces in this line, and line below.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/MergeResources.sql
```diff
@@ -0,0 +1,257 @@
...
+    DELETE FROM dbo.TokenDateTimeCompositeSearchParam WHERE EXISTS (SELECT * FROM @PreviousSurrogateIds WHERE TypeId = ResourceTypeId AND SurrogateId = ResourceSurrogateId)
+    SET @AffectedRows += @@rowcount
+    DELETE FROM dbo.TokenQuantityCompositeSearchParam WHERE EXISTS (SELECT * FROM @PreviousSurrogateIds WHERE TypeId = ResourceTypeId AND SurrogateId = ResourceSurrogateId)
+    SET @AffectedRows += @@rowcount
+    DELETE FROM dbo.TokenStringCompositeSearchParam WHERE EXISTS (SELECT * FROM @PreviousSurrogateIds WHERE TypeId = ResourceTypeId AND SurrogateId = ResourceSurrogateId)
+    SET @AffectedRows += @@rowcount
+    DELETE FROM dbo.TokenNumberNumberCompositeSearchParam WHERE EXISTS (SELECT * FROM @PreviousSurrogateIds WHERE TypeId = ResourceTypeId AND SurrogateId = ResourceSurrogateId)
+    SET @AffectedRows += @@rowcount
+
+    --EXECUTE dbo.LogEvent @Process=@SP,@Mode=@Mode,@Status='Info',@Start=@st,@Rows=@AffectedRows,@Text='Old rows'
```
fhibf, line 131: More comments.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -191,52 +195,33 @@ internal class SqlServerFhirDataStore : IFhirDataStore, IProvideCapability
...
-                            return new UpsertOutcome(existingResource, SaveOutcomeType.Updated);
-                        }
-
-                        resource.Version = newVersion.ToString();
-
-                        SaveOutcomeType saveOutcomeType;
-                        if (newVersion == 1)
-                        {
-                            saveOutcomeType = SaveOutcomeType.Created;
+                            var surrId = (await MergeResourcesBeginTransactionAsync(1, cancellationToken)).MinResourceSurrogateId;
```
fhibf, line 202: Should we always use 1 as the resourceVersionCount?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -440,11 +413,64 @@ private static string RemoveTrailingZerosFromMillisecondsForAGivenDate(DateTimeO
             return formattedDate.Replace(milliseconds, trimmedMilliseconds, StringComparison.Ordinal);
         }
 
-        private static string RemoveVersionIdAndLastUpdatedFromMeta(ResourceWrapper resourceWrapper)
+        private void ReplaceVersionIdAndLastUpdatedInMeta(ResourceWrapper resourceWrapper)
+        {
+            var date = GetJsonValue(resourceWrapper.RawResource.Data, "lastUpdated");
+            string rawResourceData;
+            if (resourceWrapper.Version == "1") // version is already correct
```
fhibf, line 420: Low priority: I see there are plenty of places where there are checks validating if the version of a resource is "1". 

Usually, we try to avoid "Magic Numbers". I would suggest creating a constant in the code called "InitialVersion" or "FirstVersion" with the value "1". Then the code will be more readable in that sense.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -440,11 +413,64 @@ private static string RemoveTrailingZerosFromMillisecondsForAGivenDate(DateTimeO
...
+            else
+            {
+                return input.RawResource.Data
+                            == existing.RawResource.Data
+                                    .Replace($"\"versionId\":\"{existingVersion}\"", $"\"versionId\":\"1\"", StringComparison.Ordinal)
+                                    .Replace($"\"lastUpdated\":\"{existingDate}\"", $"\"lastUpdated\":\"{inputDate}\"", StringComparison.Ordinal);
+            }
+        }
+
+        private string GetJsonValue(string json, string propName)
```
fhibf, line 454: I've noticed this method only works with properties with values between quotes (and no spaces between the ":" char). 
Is that expected?

Per example, the following statement returns an empty string:
string value = GetJsonValue("{ \"version\" : \"1\" }", "version");

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/49.diff.sql
SergeyGaluzo: This is intentional, so INSERT and SELECT column names/expressions are right aligned.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/49.diff.sql
SergeyGaluzo: No. I usually leave some commented code in SQL scripts which helps in debugging from SSMS.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/49.diff.sql
```diff
@@ -0,0 +1,587 @@
...
+
+  EXECUTE dbo.LogEvent @Process=@SP,@Mode=@Mode,@Status='End',@Start=@st,@Rows=NULL,@Text=@TransactionId
+END TRY
+BEGIN CATCH
+  IF error_number() = 1750 THROW -- Real error is before 1750, cannot trap in SQL.
+  EXECUTE dbo.LogEvent @Process=@SP,@Mode=@Mode,@Status='Error';
+  THROW
+END CATCH
+GO
+--DECLARE @TransactionId bigint
```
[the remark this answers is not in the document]
SergeyGaluzo: Helps in debugging.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/49.diff.sql
SergeyGaluzo: No. Once we start using true batches with transaction only in SQL we will uncomment.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
SergeyGaluzo: For current Upsert it is 1. Once we switch to batch it will be number of resource versions in a batch.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
SergeyGaluzo: Will do

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
SergeyGaluzo: True. This works for our current raw resource formatting only which does not have extra spaces.
All this logic should be thrown away once we have good solution for this bug [https://microsofthealth.visualstudio.com/Health/_workitems/edit/99962/](https://microsofthealth.visualstudio.com/Health/_workitems/edit/99962/)

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -469,6 +495,20 @@ public void Build(ICapabilityStatementBuilder builder)
             }
         }
 
+        public async Task<(long TransactionId, long MinResourceSurrogateId)> MergeResourcesBeginTransactionAsync(int resourceVersionCount, CancellationToken cancellationToken)
```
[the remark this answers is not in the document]
SergeyGaluzo: I don't think there will a chance of miscalculation as in MergeResources method it will be substituted by resources.Count.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/MergeResources.sql
SergeyGaluzo: Same as above.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/CaptureResourceIdsForChanges.sql
SergeyGaluzo: This is intentional, so INSERT and SELECT column names/expressions are right aligned.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/49.diff.sql
SergeyGaluzo: Added comment in main code.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/49.diff.sql
SergeyGaluzo: Added comment in main code.

fhibf (approved):

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/Merge/StringSearchParamListRowGenerator.cs
```diff
@@ -0,0 +1,41 @@
...
+                overflow = searchValue.String;
+            }
+            else
+            {
+                indexedPrefix = searchValue.String;
+                overflow = null;
+            }
+
+            row = new StringSearchParamListRow(resourceTypeId, resourceSurrogateId, searchParamId, indexedPrefix, overflow, IsMin: searchValue.IsMin, IsMax: searchValue.IsMax);
+            return results == null || results.Add(new StringSearchParamListRow(resourceTypeId, resourceSurrogateId, searchParamId, indexedPrefix?.ToLowerInvariant(), overflow?.ToLowerInvariant(), IsMin: searchValue.IsMin, IsMax: searchValue.IsMax));
```
fhibf, line 38: As I've mentioned in chat, I would just leave a comment in this part of the code mentioning the reason there are two instances of StringSearchParamListRow being created and the usage of .ToLowerInvariant().

Thanks.

SergeyGaluzo: Added

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/FhirStorageTests.cs
```diff
@@ -144,20 +144,12 @@ private async Task UpdateResource(Patient patient)
         [Fact]
         public async Task GivenAResource_WhenSaving_ThenTheMetaIsUpdated()
```
brendankowitz, line 145:
```suggestion
        public async Task GivenAResource_WhenSaving_ThenTheMetaIsUpdatedWithin100ms()
```

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/SortTests.cs
```diff
@@ -529,6 +530,17 @@ public async Task GivenQueryWithObservationInclude_WhenSearchedWithSortParamOnDa
             await ExecuteAndValidateBundle($"Observation?_tag={tag}&_sort=-date&_include=Observation:subject", false, resources.ToArray());
         }
 
+        [Fact]
+        [HttpIntegrationFixtureArgumentSets(dataStores: DataStore.SqlServer)]
+        public async Task SimpleTestWithCoreLogicToUnderstandHowThingsWork()
```
brendankowitz, line 535: Update name

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/HistoryTests.cs
```diff
@@ -194,6 +195,20 @@ public async Task GivenAValueForSince_WhenGettingSystemHistory_TheServerShouldRe
             Assert.Contains($"Changed by E2E test {tag}", obsHistory.Text.Div);
         }
 
+        [Fact]
+        public async Task ChangedResourceWithinAndOutsideHistoryRange()
```
brendankowitz, line 199: Update name

src/Microsoft.Health.Fhir.Core/Features/Persistence/IFhirDataStore.cs
```diff
@@ -11,6 +11,8 @@ namespace Microsoft.Health.Fhir.Core.Features.Persistence
 {
     public interface IFhirDataStore
     {
+        Task<bool> MergeAsync(IReadOnlyList<ResourceWrapper> resources, CancellationToken cancellationToken);
```
brendankowitz, line 14: As discussed, might not be needed for initial check-in

SergeyGaluzo: Removed

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/HistoryTests.cs
SergeyGaluzo: Changed

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/SortTests.cs
SergeyGaluzo: Done

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/FhirStorageTests.cs
SergeyGaluzo: Done

src/Microsoft.Health.Fhir.Core/Features/Persistence/ResourceWrapperExtended.cs
```diff
@@ -0,0 +1,34 @@
...
+    public class ResourceWrapperExtended
+    {
+        public ResourceWrapperExtended(
+            ResourceWrapper wrapper,
+            bool allowCreate,
+            bool keepHistory,
+            WeakETag weakETag,
+            bool requireETagOnUpdate)
+        {
+            Wrapper = wrapper;
```
fhibf, line 17: Low priority: I think we should add validations to avoid wrapper to be null. Not sure if weakETag can be null as well.

This is a low priority item, if you don't have cycles to fix it, I can fix it in another branch. Just leave a comment.

fhibf: Looking the code I've noticed that weakETag can be null.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
```diff
@@ -108,7 +108,12 @@ public sealed class CosmosFhirDataStore : IFhirDataStore, IProvideCapability
             _modelInfoProvider = modelInfoProvider;
         }
 
-        public async Task<UpsertOutcome> UpsertAsync(
+        public async Task<UpsertOutcome> UpsertAsync(ResourceWrapperExtended resource, CancellationToken cancellationToken)
```
fhibf, line 111: Low priority: add checks to avoid null references of ResourceWrapperExtended.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/SchemaVersion.cs
```diff
@@ -58,5 +58,6 @@ public enum SchemaVersion
         V46 = 46,
         V47 = 47,
         V48 = 48,
+        V49 = 49,
```
fhibf, line 61: Heads up: I believe, due the issue with export and the changes Jared is actively working, version 49 will be released this week. Probably we should include all these changes under version 50.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/Merge/CompartmentAssignmentListRowGenerator.cs
```diff
@@ -0,0 +1,137 @@
...
+            EnsureArg.IsNotNull(model, nameof(model));
+            EnsureArg.IsNotNull(searchParameterTypeMap, nameof(searchParameterTypeMap));
+
+            _model = model;
+            _searchParameterTypeMap = searchParameterTypeMap;
+        }
+
+        public IEnumerable<CompartmentAssignmentListRow> GenerateRows(IReadOnlyList<MergeResourceWrapper> resources)
+        {
+            EnsureInitialized();
```
fhibf, line 39: Low priority: Should we add a validation to avoid null references of "resources"?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/Merge/CompartmentAssignmentListRowGenerator.cs
```diff
@@ -0,0 +1,137 @@
...
+            {
+                var resource = merge.ResourceWrapper;
+                var typeId = _model.GetResourceTypeId(resource.ResourceTypeName);
+
+                var resourceMetadata = new ResourceMetadata(
+                    resource.CompartmentIndices,
+                    resource.SearchIndices?.ToLookup(e => _searchParameterTypeMap.GetSearchValueType(e)),
+                    resource.LastModifiedClaims);
+
+                CompartmentIndices compartments = resourceMetadata.Compartments;
```
fhibf, line 51: If "Compartments" for a single item is null, we should break the looping or move to the next item in the list?
The current code breaks, which causes other items in the list to be ignored. Is that intentional?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -85,172 +89,299 @@ internal class SqlServerFhirDataStore : IFhirDataStore, IProvideCapability
...
+                            if (resource.IsDeleted && existingResource.IsDeleted)
                             {
-                                // Send the existing resource in the response
-                                return new UpsertOutcome(existingResource, SaveOutcomeType.Updated);
+                                // Already deleted - don't create a new version
+                                results.Add(resourceKey, null);
+                                continue;
                             }
+
+                            // check if reosurces are equal if its not a Delete action
```
brendankowitz, line 189:
```suggestion
                            // check if resources are equal if its not a Delete action
```

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -85,172 +89,299 @@ internal class SqlServerFhirDataStore : IFhirDataStore, IProvideCapability
...
+                catch (SqlException e)
+                {
+                    // we cannot retry on connection loss as this call might be in outer transaction.
+                    // TODO: Add retries when set bundle processing is in place.
+                    if (e.Number == SqlErrorCodes.Conflict && retries++ < 10) // retries on conflict should never be more than 1, so it is OK to hardcode.
                     {
-                        _logger.LogCritical("Stream size for resource of type: {ResourceTypeName} is less than 50 bytes, request method: {RequestMethod}", resource.ResourceTypeName, resource.Request.Method);
-                        throw new ServiceUnavailableException();
+                        _logger.LogWarning(e, $"Error from SQL database on {nameof(MergeAsync)} retries={retries}");
+                        await Task.Delay(5000, cancellationToken);
```
brendankowitz, line 238: Any reason behind 5s?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -85,172 +89,299 @@ internal class SqlServerFhirDataStore : IFhirDataStore, IProvideCapability
...
+                }
+                catch (SqlException e)
+                {
+                    // we cannot retry on connection loss as this call might be in outer transaction.
+                    // TODO: Add retries when set bundle processing is in place.
+                    if (e.Number == SqlErrorCodes.Conflict && retries++ < 10) // retries on conflict should never be more than 1, so it is OK to hardcode.
                     {
-                        _logger.LogCritical("Stream size for resource of type: {ResourceTypeName} is less than 50 bytes, request method: {RequestMethod}", resource.ResourceTypeName, resource.Request.Method);
-                        throw new ServiceUnavailableException();
+                        _logger.LogWarning(e, $"Error from SQL database on {nameof(MergeAsync)} retries={retries}");
```
brendankowitz, line 237: nit: Try not to use string interpolation for variables in log messages
```suggestion
                        _logger.LogWarning(e, $"Error from SQL database on {nameof(MergeAsync)} retries={{Retries}}", retries);
```

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/Merge/CompositeSearchParamRowGenerator.cs
```diff
@@ -0,0 +1,83 @@
...
+        private readonly Func<EnumeratorWrapper<ISearchValue>, TSearchValue> _converter = CreateConverterFunc();
+
+        protected CompositeSearchParamRowGenerator(SqlServerFhirModel model, SearchParameterToSearchValueTypeMap searchParameterTypeMap)
+            : base(model, searchParameterTypeMap)
+        {
+        }
+
+        protected override IEnumerable<TSearchValue> ConvertSearchValue(SearchIndexEntry entry)
+        {
+            var compositeSearchValue = (CompositeSearchValue)entry.Value;
```
fhibf, line 31: Should we check if entry is null?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/Merge/DateTimeSearchParamListRowGenerator.cs
```diff
@@ -0,0 +1,50 @@
...
+    internal class DateTimeSearchParamListRowGenerator : MergeSearchParameterRowGenerator<DateTimeSearchValue, DateTimeSearchParamListRow>
+    {
+        private short _lastUpdatedSearchParamId;
+
+        public DateTimeSearchParamListRowGenerator(SqlServerFhirModel model, SearchParameterToSearchValueTypeMap searchParameterTypeMap)
+            : base(model, searchParameterTypeMap)
+        {
+        }
+
+        internal override bool TryGenerateRow(short resourceTypeId, long resourceSurrogateId, short searchParamId, DateTimeSearchValue searchValue, HashSet<DateTimeSearchParamListRow> results, out DateTimeSearchParamListRow row)
```
fhibf, line 23: Should we add a check if searchValue is null?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/Merge/MergeResourceWrapper.cs
```diff
@@ -0,0 +1,40 @@
...
+
+using Microsoft.Health.Fhir.Core.Features.Persistence;
+
+namespace Microsoft.Health.Fhir.SqlServer.Features.Storage
+{
+    internal class MergeResourceWrapper
+    {
+        internal MergeResourceWrapper(ResourceWrapper resourceWrapper, long resourceSurrogateId, bool keepHistory, bool hasVersionToCompare)
+        {
+            ResourceWrapper = resourceWrapper;
```
fhibf, line 14: Should we add checks to avoid null values here?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/Merge/MergeSearchParameterRowGenerator.cs
```diff
@@ -0,0 +1,97 @@
...
+        private bool _isInitialized;
+
+        protected MergeSearchParameterRowGenerator(SqlServerFhirModel model, SearchParameterToSearchValueTypeMap searchParameterTypeMap)
+        {
+            EnsureArg.IsNotNull(model, nameof(model));
+            EnsureArg.IsNotNull(searchParameterTypeMap, nameof(searchParameterTypeMap));
+
+            Model = model;
+            _searchParameterTypeMap = searchParameterTypeMap;
+            _isConvertSearchValueOverridden = GetType().GetMethod(nameof(ConvertSearchValue), BindingFlags.Instance | BindingFlags.NonPublic).DeclaringType != typeof(SearchParameterRowGenerator<TSearchValue, TRow>);
```
fhibf, line 30: I like this.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/Merge/MergeSearchParameterRowGenerator.cs
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        protected SqlServerFhirModel Model { get; }
+
+        public virtual IEnumerable<TRow> GenerateRows(IReadOnlyList<MergeResourceWrapper> resources)
```
fhibf, line 35: Add a check to ensure "resources" is not null.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/Merge/ReferenceTokenCompositeSearchParamListRowGenerator.cs
```diff
@@ -0,0 +1,52 @@
...
+            SearchParameterToSearchValueTypeMap searchParameterTypeMap)
+            : base(model, searchParameterTypeMap)
+        {
+            _referenceRowGenerator = referenceRowGenerator;
+            _tokenRowGenerator = tokenRowGenerator;
+        }
+
+        internal override bool TryGenerateRow(short resourceTypeId, long resourceSurrogateId, short searchParamId, (ReferenceSearchValue component1, TokenSearchValue component2) searchValue, HashSet<ReferenceTokenCompositeSearchParamListRow> results, out ReferenceTokenCompositeSearchParamListRow row)
+        {
+            if (_referenceRowGenerator.TryGenerateRow(resourceTypeId, resourceSurrogateId, searchParamId, searchValue.component1, null, out var reference1Row) &&
```
fhibf, line 30: Is there a reason we are passing null to TryGenerateRow instead of passing "results"? 
Not sure if I'm missing it, but probably it'll would be required to avoid duplicates.

fhibf: If that's intentional, then we probably need a comment mentioning the reason why.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/Merge/ResourceListRowGenerator.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        public IEnumerable<ResourceListRow> GenerateRows(IReadOnlyList<MergeResourceWrapper> mergeWrappers)
+        {
+            foreach (var merge in mergeWrappers)
+            {
+                var wrapper = merge.ResourceWrapper;
+                using var stream = new RecyclableMemoryStream(_memoryStreamManager, tag: nameof(ResourceListRowGenerator));
+                _compressedRawResourceConverter.WriteCompressedRawResource(stream, wrapper.RawResource.Data);
+                stream.Seek(0, 0);
+                yield return new ResourceListRow(_model.GetResourceTypeId(wrapper.ResourceTypeName), merge.ResourceSurrogateId, wrapper.ResourceId, int.Parse(wrapper.Version), merge.HasVersionToCompare, wrapper.IsDeleted, wrapper.IsHistory, merge.KeepHistory, stream, wrapper.RawResource.IsMetaSet, wrapper.Request.Method, wrapper.SearchParameterHash);
```
fhibf, line 36: Something we should tests it's the usage of stream at this point. 
As we are creating it around a using block, it's possible that at the end of the method it'll dispose the stream. 
And at the time the ResourceListRow.RawResource uses it, it'll be disposed. 
Let's check that.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/Merge/ResourceWriteClaimListRowGenerator.cs
```diff
@@ -0,0 +1,57 @@
...
+            {
+                var resource = merge.ResourceWrapper;
+
+                var resourceMetadata = new ResourceMetadata(
+                    resource.CompartmentIndices,
+                    resource.SearchIndices?.ToLookup(e => _searchParameterTypeMap.GetSearchValueType(e)),
+                    resource.LastModifiedClaims);
+
+                IReadOnlyCollection<KeyValuePair<string, string>> writeClaims = resourceMetadata.WriteClaims;
+                if (writeClaims == null)
```
fhibf, line 41: Should we break if the writeClaims are null? Shouldn't we continue processing the other elements under resources?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/Merge/ResourceWriteClaimListRowGenerator.cs
```diff
@@ -0,0 +1,57 @@
...
+                {
+                    yield break;
+                }
+
+                var results = new HashSet<ResourceWriteClaimListRow>();
+                foreach (var claim in writeClaims)
+                {
+                    if (results.Add(new ResourceWriteClaimListRow(merge.ResourceSurrogateId, _model.GetClaimTypeId(claim.Key), claim.Value?.ToLowerInvariant())))
```
fhibf, line 49: I believe the usage of ToLowerInvariant() is expected here, as a solution to the sensitive case issue we discussed last week.

fhibf: I'm resolving this comment, as you have mentioned in StringSearchParamListRowGenerator that this behavior is common in other places.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/Merge/ReferenceTokenCompositeSearchParamListRowGenerator.cs
fhibf: I've noticed the same behavior in other places, like:

TokenStringCompositeSearchParamListRowGenerator 
TokenQuantityCompositeSearchParamListRowGenerator 
TokenNumberNumberCompositeSearchParamListRowGenerator 
TokenDateTimeCompositeSearchParamListRowGenerator

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
SergeyGaluzo: As you can see, I kept the same behavior as on original Upsert. Are you saying existing code is bad?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/SchemaVersion.cs
SergeyGaluzo: Sure. There is nothing to discuss as Jared's fix will go first.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/Merge/CompartmentAssignmentListRowGenerator.cs
SergeyGaluzo: Again, it is copy/past from current code. I don't see the need to add checks. I can try making classes internal, do it is obvious that they rely on external validation.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/Merge/CompartmentAssignmentListRowGenerator.cs
SergeyGaluzo: This is the behavior on the previous code.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/Merge/CompositeSearchParamRowGenerator.cs
SergeyGaluzo: This is the behavior of the previous code.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/Merge/DateTimeSearchParamListRowGenerator.cs
SergeyGaluzo: This is the behavior of the previous code.

src/Microsoft.Health.Fhir.Core/Features/Persistence/IFhirDataStore.cs
```diff
@@ -11,13 +11,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Persistence
...
     public interface IFhirDataStore
     {
-        Task<UpsertOutcome> UpsertAsync(
-            ResourceWrapper resource,
-            WeakETag weakETag,
-            bool allowCreate,
-            bool keepHistory,
-            CancellationToken cancellationToken,
-            bool requireETagOnUpdate = false);
+        Task<UpsertOutcome> UpsertAsync(ResourceWrapperExtended resource, CancellationToken cancellationToken);
```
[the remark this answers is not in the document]
SergeyGaluzo: I think name should contain ResoureWrapper.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -85,172 +89,299 @@ internal class SqlServerFhirDataStore : IFhirDataStore, IProvideCapability
...
-                // ** We must use CreateNonRetrySqlCommand here because the retry will not reset the Stream containing the RawResource, resulting in a failure to save the data
-                using (SqlConnectionWrapper sqlConnectionWrapper = await _sqlConnectionWrapperFactory.ObtainSqlConnectionWrapperAsync(cancellationToken, true))
-                using (SqlCommandWrapper sqlCommandWrapper = sqlConnectionWrapper.CreateNonRetrySqlCommand())
-                using (var stream = new RecyclableMemoryStream(_memoryStreamManager, tag: nameof(SqlServerFhirDataStore)))
+                try
                 {
-                    // Read the latest resource
-                    var existingResource = await GetAsync(new ResourceKey(resource.ResourceTypeName, resource.ResourceId), cancellationToken);
+                    // ignore input resource version to get latest version from the store
+                    var existingResources = (await GetAsync(resources.Select(_ => _.Wrapper.ToResourceKey(true)).Distinct().ToList(), cancellationToken)).ToDictionary(_ => _.ToResourceKey(true), _ => _);
```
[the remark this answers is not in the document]
SergeyGaluzo: we don't agree.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -85,172 +89,299 @@ internal class SqlServerFhirDataStore : IFhirDataStore, IProvideCapability
...
+                        int? eTag = weakETag == null
+                            ? null
+                            : (int.TryParse(weakETag.VersionId, out var parsedETag) ? parsedETag : -1); // Set the etag to a sentinel value to enable expected failure paths when updating with both existing and nonexistent resources.
+
+                        var resource = resourceExt.Wrapper;
+                        var resourceKey = resource.ToResourceKey(); // keep input version in the results to allow processing multiple versions per resource
+                        existingResources.TryGetValue(resource.ToResourceKey(true), out var existingResource);
+
+                        // Check for any validation errors
+                        if (existingResource != null && eTag.HasValue && eTag.ToString() != existingResource.Version)
```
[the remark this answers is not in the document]
SergeyGaluzo: @brendankowitz Do you realize thar you are suggesting change in previous code behavior?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -85,172 +89,299 @@ internal class SqlServerFhirDataStore : IFhirDataStore, IProvideCapability
...
+                                // Don't bother marking the resource as deleted since it already does not exist.
+                                results.Add(resourceKey, null);
+                                continue;
+                            }
 
-                        if (eTag.HasValue && eTag != null)
-                        {
-                            // You can't update a resource with a specified version if the resource does not exist
-                            if (weakETag != null)
+                            if (eTag.HasValue && eTag != null)
```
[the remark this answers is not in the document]
SergeyGaluzo: I will remove garbage copy/pasted from previous code as is.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
SergeyGaluzo: I will remove garbage copy/pasted from previous code as is.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
SergeyGaluzo: Retrying instantly does not make sense. 5 seconds is how SQL reacts on deadlocks.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
SergeyGaluzo: will change

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/Merge/MergeSearchParameterRowGenerator.cs
SergeyGaluzo: Not sure what...

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/Merge/ReferenceTokenCompositeSearchParamListRowGenerator.cs
SergeyGaluzo: It is intentional. I will put comment.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/Merge/ResourceListRowGenerator.cs
SergeyGaluzo: Ts code works and passes stress tests. What else should be check?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/Merge/ResourceWriteClaimListRowGenerator.cs
SergeyGaluzo: This is behavior of the code it was copied/pasted from.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/Merge/MergeSearchParameterRowGenerator.cs
SergeyGaluzo: This is internal class.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/Merge/MergeResourceWrapper.cs
SergeyGaluzo: This is internal class.

src/Microsoft.Health.Fhir.Core/Features/Persistence/ResourceWrapperExtended.cs
SergeyGaluzo: I would avoid using default value null for week tag. This brought us to not testing not null value anywhere. I prefer explicit values. People should make caution decision on passing null.

src/Microsoft.Health.Fhir.Core/Features/Persistence/IFhirDataStore.cs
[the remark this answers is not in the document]
fhibf: Usually, I name those objects adding the suffix "Context".
It helps understanding that a specific object has values used in the context of some type of operation. 

In this case, per example, some suggestions would be: 
- "UpsertContext"
- "ResourceOperationContext"
- "ResourceWrapperOperationContext"

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/Merge/CompartmentAssignmentListRowGenerator.cs
fhibf: @brendankowitz we've noticed few cases of loops breaking in the middle if a null condition is true. 
Should we instead of breaking the loop in the middle continue analyzing the remaining items in the collection?

src/Microsoft.Health.Fhir.Core/Features/Persistence/IFhirDataStore.cs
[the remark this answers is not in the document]
brendankowitz: I like ResourceWrapperOperation[Context] or ResourceWrapperContainer

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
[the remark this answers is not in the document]
brendankowitz: Never too late to improve :)

src/Microsoft.Health.Fhir.Core/Features/Persistence/IFhirDataStore.cs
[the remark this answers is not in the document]
SergeyGaluzo: ResourceWrapperOperation it is

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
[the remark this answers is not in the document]
SergeyGaluzo: Changed.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/Merge/CompartmentAssignmentListRowGenerator.cs
SergeyGaluzo: I replaced yield break by continue.

src/Microsoft.Health.Fhir.Core/Features/Persistence/ResourceWrapperOperation.cs
```diff
@@ -0,0 +1,34 @@
...
+    public class ResourceWrapperOperation
+    {
+        public ResourceWrapperOperation(
+            ResourceWrapper wrapper,
+            bool allowCreate,
+            bool keepHistory,
+            WeakETag weakETag,
+            bool requireETagOnUpdate)
+        {
+            Wrapper = wrapper;
```
brendankowitz, line 17:
```suggestion
            Wrapper = EnsureArg.IsNotNull(wrapper, nameof(wrapper));
```

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
```diff
@@ -108,7 +108,44 @@ public sealed class CosmosFhirDataStore : IFhirDataStore, IProvideCapability
...
+            var results = new List<ResourceWrapper>();
+            if (keys == null || keys.Count == 0)
+            {
+                return results;
+            }
+
+            foreach (var key in keys)
+            {
+                results.Add(await GetAsync(key, cancellationToken));
+            }
```
brendankowitz, line 138: Guessing the syntax here, but why not use a parallel select?

```suggestion
            results.AddRange(await Task.WhenAll(keys.Select(async key => await GetAsync(key, cancellationToken))));
```

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Upsert/UpsertResourceHandler.cs
```diff
@@ -59,7 +59,8 @@ public async Task<UpsertResourceResponse> Handle(UpsertResourceRequest request,
 
             ResourceWrapper resourceWrapper = CreateResourceWrapper(resource, deleted: false, keepMeta: allowCreate);
 
-            UpsertOutcome result = await FhirDataStore.UpsertAsync(resourceWrapper, request.WeakETag, allowCreate, keepHistory, cancellationToken, requireETagOnUpdate);
+            ////UpsertOutcome result = await FhirDataStore.UpsertAsync(resourceWrapper, request.WeakETag, allowCreate, keepHistory, cancellationToken, requireETagOnUpdate);
```
brendankowitz, line 62: nit: commented code

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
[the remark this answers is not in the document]
brendankowitz: `_` has a documented use as a discard https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/discards. _Discards are placeholder variables that are intentionally unused in application code_, which makes its use in these scenarios confusing. 

I think .net does have pretty extensive and reasonable documentation for best practices. Keeping to them generally means we usually won't get into trouble like this as the language continues to evolve.
https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/general-naming-conventions#word-choice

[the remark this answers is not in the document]
SergeyGaluzo: changed

src/Microsoft.Health.Fhir.Core/Features/Persistence/ResourceWrapperOperation.cs
SergeyGaluzo: changed

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
SergeyGaluzo: changed

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Upsert/UpsertResourceHandler.cs
SergeyGaluzo: removed

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/MergeResources.sql
```diff
@@ -0,0 +1,259 @@
...
+
+  IF @IsResourceChangeCaptureEnabled = 1 --If the resource change capture feature is enabled, to execute a stored procedure called CaptureResourceChanges to insert resource change data.
+  BEGIN
+    DECLARE @Ids dbo.ResourceIdForChangesList
+    INSERT INTO @Ids
+           ( ResourceTypeId, ResourceId, Version, IsDeleted )
+      SELECT ResourceTypeId, ResourceId, Version, IsDeleted
+        FROM @Resources
+        WHERE IsHistory = 0
+    EXECUTE dbo.CaptureResourceIdsForChanges @Ids
```
ms-teli, line 241: This may just for my own knowledge, what's the reason behind that we buffer the data in the type ResourceIdForChangesList comparing to passing the parameters directly to the Stored Procedure, any benefit?

SergeyGaluzo: There are multiple rows, so we pass parameter as TVP. It is as direct as it could be short of inserting to change data table directly.

ms-teli (approved):

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/MergeResources.sql
SergeyGaluzo: After talking directly... We decided to use @Resources parameter for change data stored proc.

ms-teli (approved):
Looks good from FHIR event's perspective.

fhibf (approved):

PR #3076, opened 2023-01-20
Add intermediate export stage

LTA-Thinking:
## Description
When a long running export job is run there is no intermediate states recorded. This means that if the job is stopped and restarted it has to start from the beginning. This PR adds saving of intermediate stages to long running jobs by having the heartbeat record results every 10 minutes.

## Related issues
Addresses [Bug 99078](https://microsofthealth.visualstudio.com/Health/_workitems/edit/99078): Single threaded export does not save intermediate states

## Testing
Manual testing and new unit test.

## FHIR Team Checklist

Patch

### DIFF
--- a/src/Microsoft.Health.TaskManagement.UnitTests/JobHostingTests.cs
+++ b/src/Microsoft.Health.TaskManagement.UnitTests/JobHostingTests.cs
@@ -277,6 +277,55 @@ namespace Microsoft.Health.JobManagement.UnitTests
         }
 
         [Fact]
+        public async Task GivenJobRunning_WhenLongTimeHasPassed_ThenResultsShouldBeRecorded()
+        {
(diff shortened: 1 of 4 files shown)


SergeyGaluzo (changes requested):
:clock1:

SergeyGaluzo:
Heavy heartbeats is a leftover of wrong design used by $import. It leads to "N-square" problem, where N is number of saves, in case when each save includes results of previous saves. 

Good example is maintaining full list of file urls written by $export. Assume 10K files, 10 new files are added every 10 minutes, 1000 saves. Average number of file urls per save is 5K. Total data written 5K * 1K = 5M file urls, instead of just 10K.

I missed adding deprecated attribute on heavy heartbeats method. I will add it.

Please change logic to save incremental states (10 new file urls in the above example) as separate completed jobs using group id of the main parent job.

Issue #3100, opened 2023-02-02
Narrative HTML validation ignores some errors


michael-wilson-au:
**Describe the bug**
The HTML validation of a resource narrative can ignore errors that should be raised. 

If the narrative contains html errors that should be ignored the validation function exits without completing the rest of the validation which may find additional errors that shouldn't be ignored.  

**FHIR Version?**
Core

**Data provider?**
N/A

**To Reproduce**
1. Create a FHIR resource (i.e. Patient) with a text narrative containing html errors that can be ignored by the narrative html validator as well as html errors that cannot be ignored:
```
{
  "resourceType": "Patient",
  "id": "example",
  "text": {
    "status": "generated",
    "div": "<div><!-- Comment ended with a dash. This error should be ignored --->\"><not_valid_tag>This tag should return validation error</not_valid_tag></div>"
  },
  "name": [
    {
      "family": "Windsor",
      "given": [
        "Peter",
        "James"
      ]
    }
  ]
}
```
2.     Create the resource POST {{fhirurl}}/Patient

**Expected behavior**
400 Bad Request response for "Illegal element name 'NOT_VALID_TAG'."

**Actual behavior**
201 Created response

EXPEkesheth:
Thanks for reporting the issue, we will take a look and respond

michael-wilson-au:
I have submitted PR #3101 to resolve this issue

PR #3101, opened 2023-02-02
Narrative HTML validation fix

michael-wilson-au:
## Description
Updated the NarrativeHtmlSanitizer validator to not exit when HTML parse errors are to be ignored.  It now allows the validator to continue checking for errors after ignoring a HTML parse error. 

## Related issues
Addresses #3100 

## Testing
New unit test added to NarrativeHtmlSanitizerTests to check that errors can still be raised after a ignored HTML parse error is found.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Validation/Narratives/NarrativeHtmlSanitizerTests.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Validation/Narratives/NarrativeHtmlSanitizerTests.cs
@@ -62,7 +62,7 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Validation.Narratives
         }
 
         [Theory]
-        [InlineData("<div><!-- Comment ended with a dash. This error should be ignored --->><not_valid_tag>This tag should return validation error</not_valid_tag></div>")]
+        [InlineData("<div><!-- Comment ended with a dash. This error should be ignored ---><not_valid_tag>This tag should return validation error</not_valid_tag></div>")]
         public void GivenHtmlWithIgnoredAndEnforecedErrors_WhenSanitizingHtml_ThenAValidationErrorIsReturned(string val)
         {
             var results = _sanitizer.Validate(val);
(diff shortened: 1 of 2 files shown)


michael-wilson-au:
@microsoft-github-policy-service agree

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Validation/Narratives/NarrativeHtmlSanitizerTests.cs
```diff
@@ -61,6 +61,15 @@ public void GivenHtmlWithDivAndText_WhenSanitizingHtml_ThenValidationIsSuccessfu
             Assert.Empty(results);
         }
 
+        [Theory]
+        [InlineData("<div><!-- Comment ended with a dash. This error should be ignored ---><not_valid_tag>This tag should return validation error</not_valid_tag></div>")]
```
LTA-Thinking, line 65: Instead of making a new test, could you move this scenario to the theory on line 35? That theory's name doesn't fit the scenarios it is running. It is for invalid narratives, not just empty divs. If you could update its name to match what it is testing that would be great.

michael-wilson-au: Removed the new test and updated the existing test as suggested in latest commit

LTA-Thinking:
This PR has shown an interesting issue with some of the implementation guides and how they follow FHIR standards. We have a few copies of implementation resources for testing, and they are using external image sources in their narratives. This is not allowed by the FHIR standard or the validator, so they are now failing in our E2E tests. For example: [US Core Birthsex](https://github.com/microsoft/fhir-server/blob/main/src/Microsoft.Health.Fhir.Tests.Common/TestFiles/Stu3/StructureDefinition-us-core-birthsex.json)
I'm looking into what the best way forward here is.

michael-wilson-au:
I have also come across the issue with external references to source images in the narrative and it seems to be quite widespread. It is prevalent in HL7 Base resources, for example the Patient resource:
http://hl7.org/fhir/r4/patient.profile.json
`<img src=\"help16.png\"`

I think that strict enforcement of image source will need to be removed from narrative validation in the FHIR server as there are too many instances where the implementation guide hasn't been followed and the image source validation is preventing the creation of these resources.

LTA-Thinking:
I asked this narrative image source question on Zulip. For now we can't accept this PR into main. @michael-wilson-au if you want to keep working on this issue it needs to filter out these image source errors. If not we can point this PR to a feature branch where one of our team members can pick up the work of filtering the exceptions when time allows.

Thank you for the contribution. I'm sorry we can't accept it as is.

LTA-Thinking:
I'm going to merge this PR to a feature branch and add the changes we need to it from there.

PR #3131, opened 2023-02-16
Test Fixes + ConformanceProvider Thrdsf + Retriable DequeueAsync

fhibf:
## Description
- Adding thread safe validations on top of SystemConformanceProvider.
- Change SearchTestBase to handle continuation tokens and compare results across multiple pages. 
- Creating a new class to handle retriable blocks of code. 
- Changes in PatientEverythingService to avoid Phase 3 runs and additional empty pages.
- Exposing "_GetActivityId()_" as a public method to help exposing more details from FhirClientExceptions.
- SqlQueueClient.DequeueAsync - Added logic to retry retriable exceptions. 
- ConditionalDeleteTests - Exposing more details in case of failures.
- DateSearchTests - Exposing more details in case of failures.
- IncludeSearchTests - Exposing more details in case of failures.

## Related issues
Addresses [issue #].

## Testing
- Tests executed through local and pipeline test validations. 

## FHIR Team Checklist


PTaladay (dismissed):

global.json
```diff
@@ -1,5 +1,5 @@
 {
     "sdk": {
-        "version": "7.0.101"
+        "version": "7.0.103"
```
brendankowitz, line 3: We should update this and https://github.com/microsoft/fhir-server/blob/main/build/docker/Dockerfile#L1, maybe in a different PR. Is there a good way to automate this?

fhibf: Thanks for the comment.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Everything/PatientEverythingService.cs
```diff
@@ -154,10 +154,15 @@ public class PatientEverythingService : IPatientEverythingService
                         ? await SearchCompartment(resourceId, parentPatientId, since, types, encodedInternalContinuationToken, token, cancellationToken)
                         : await SearchCompartmentWithoutDate(resourceId, parentPatientId, since, types, encodedInternalContinuationToken, token, cancellationToken);
 
-                    if (!searchResult.Results.Any())
+                    // Starting from FHIR R5, "Devices" are included as part of Compartment Search.
+                    // Previous versions of FHIR should still query for "Devices" explicitly.
+                    if (ModelInfoProvider.Version < FhirSpecification.R5)
```
brendankowitz, line 159: Inject IModelInfoProvider is possible since its more declarative as a dependency

src/Microsoft.Health.Fhir.Tests.Common/TestRuntime.cs
```diff
@@ -0,0 +1,106 @@
...
+using System;
+using System.IO;
+using System.Linq;
+using System.Net.Http;
+using System.Net.Sockets;
+using System.Threading.Tasks;
+
+namespace Microsoft.Health.Fhir.Tests.Common
+{
+    public static class TestRuntime
```
brendankowitz, line 15: Perhaps a bigger change, but have wondered if this should be built into our runner somehow? https://github.com/microsoft/fhir-server/blob/main/src/Microsoft.Health.Extensions.Xunit/CustomXunitTestFrameworkExecutor.cs#L177

fhibf: I've added an extension class for IClassFixture.All classes implementing IClassFixture can now have the Retry logic. My only concern in adding it directly to CustomXunitTestFrameworkExecutor, was that probably we would lose control of all requests and maybe ingest more records than required, run methods more time than need, and that could cause some unexpected behaviors.  I think the way how it's right now it's more controllable/manageable.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/SearchTestsBase.cs
```diff
@@ -225,29 +244,47 @@ protected void ValidateBundle(Bundle bundle, bool sort, bool invalidSortParamete
...
                 }
+            }
 
-                throw new XunitException(sb.ToString());
+            var fhirJsonSerializer = new FhirJsonSerializer(new SerializerSettings() { AppendNewLine = false, Pretty = false });
+            using var sw = new StringWriter(sb);
+            sb.AppendLine("Actual collection as below -");
+            foreach (var element in actualResources)
+            {
+                sb.AppendLine(fhirJsonSerializer.SerializeToString(element));
```
brendankowitz, line 278: Is this different to the extension method `element.ToJson()`?

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlQueueClientTests.cs
```diff
@@ -322,36 +328,42 @@ public async Task ExecuteWithHeartbeats()
...
+
+                    Assert.Equal(job.Id, jobInt.Id);
+                    Assert.True(dequeueDate >= execDate, $"dequeue:{dequeueDate} >= exec:{execDate}");
                 },
-                cancel.Token);
-            Task.WaitAll(execTask, dequeueTask);
-
-            Assert.Equal(job.Id, jobInt.Id);
-            Assert.True(dequeueDate >= execDate, $"dequeue:{dequeueDate} >= exec:{execDate}");
+                additionalRetriableExceptions: _retriableExceptions);
```
brendankowitz, line 366: I think this test does highlight that the underlying code doesn't handle this condition very well. In the above, a Heartbeat event can be written while the Dequeue operation is processing causing a Pre-condition failure. I'm wondering if we should just build this logic into Dequeue to retry when this this happens. Thoughts?

fhibf: Yeap. I see your suggestion, and I think it makes total sense. 
And I don't see places catching the RetriableJobException raised by Dequeue.

fhibf: Added a logic to retry retriable exceptions up to three times, unless the exception is not retriable.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlQueueClient.cs
```diff
@@ -152,41 +152,61 @@ public async Task<IReadOnlyCollection<JobInfo>> DequeueJobsAsync(byte queueType,
 
         public async Task<JobInfo> DequeueAsync(byte queueType, string worker, int heartbeatTimeoutSec, CancellationToken cancellationToken, long? jobId = null)
         {
-            try
+            const int MaxNumberOfIcMs = 3;
```
brendankowitz, line 155: 😆 IcMs?

fhibf: HAHAHA I was probably speaking about IcMs with someone. Fixing it.

brendankowitz (approved):
:shipit:

PR #3163, opened 2023-03-08
reindex for large dbs

v-mserdar:
## Description
Changes include using new search logic that bases the major emphasis on ResourceSurrogateIds.
Reindex scenarios to handle include the following:
* Getting a count of resources when doing a force reindex operation (does not use SearchParamHash)
* Getting a count of resources when doing a regular reindex operation (uses SearchParamHash)
* Getting resources when doing a force reindex operation (does not use SearchParamHash)
* Getting resources when doing a regular reindex operation (uses SearchParamHash)

When doing a force reindex, it's actually fairly straightforward because we can easily get the min/max ResourceSurrogateIds and count of resources in one SQL call. Using the min and max ResourceSurrogateIds along with the ResourceType to then search within params to get the resources that will be reindexed.
 
When doing a regular reindex, we need to use the SearchParamHash which adds complexity because there is not a SQL index on that field. Therefore we need to collect the ResourceSurrogateIds in batches based on the reindex parameter "maximumNumberOfResourcesPerQuery". This field is internally limited in our code base to between 1 and 5000 inclusive and has been tested on the large db to execute quickly even when using the upper bound of 5000.

Also made some reindex test fixes/enhancements so that we don't make calls to the db when we already know there are 0 resources. I'm hoping that this change can stop the transient errors that we saw for a few of the reindex tests.

## Related issues
Addresses [issue #100712](https://microsofthealth.visualstudio.com/Health/_workitems/edit/100712).

## Testing
Successfully ran the reindex integration tests and reindex unit tests.

## FHIR Team Checklist


src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/CreateReindexRequestHandler.cs
```diff
@@ -92,23 +87,10 @@ public async Task<CreateReindexResponse> Handle(CreateReindexRequest request, Ca
 
                     searchParameterResourceTypes.UnionWith(searchParameterInfo.BaseResourceTypes);
                 }
```
v-mserdar, line 89: This was removed because adding a guid as an indicator for the search param hash was incorrect to begin with and wasn't noticed until doing this work.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/Models/ReindexJobRecord.cs
```diff
@@ -119,10 +121,10 @@ protected ReindexJobRecord()
         public ConcurrentDictionary<ReindexJobQueryStatus, byte> QueryList { get; private set; } = new ConcurrentDictionary<ReindexJobQueryStatus, byte>();
 
         [JsonProperty(JobRecordProperties.ResourceCounts)]
-        public ConcurrentDictionary<string, int> ResourceCounts { get; private set; } = new ConcurrentDictionary<string, int>();
+        public ConcurrentDictionary<string, SearchResultReindex> ResourceCounts { get; private set; } = new ConcurrentDictionary<string, SearchResultReindex>();
```
brendankowitz, line 124: This creates the possibility that previous records won't deserialize correctly, did we test this scenario?

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -594,18 +621,39 @@ private async Task CheckJobCompletionStatus()
...
+            {
+                // if queryStatus has value then it's a continuation otherwise first time for the resource type
+                long startResourceSurrogateId = Math.Max(queryStatus.StartResourceSurrogateId, searchResultReindex.CurrentResourceSurrogateId);
+
+                queryParametersList.AddRange(new[]
+                {
+                    Tuple.Create(KnownQueryParameterNames.EndSurrogateId, searchResultReindex.EndResourceSurrogateId.ToString()),
+                    Tuple.Create(KnownQueryParameterNames.StartSurrogateId, startResourceSurrogateId.ToString()),
+                    Tuple.Create(KnownQueryParameterNames.GlobalEndSurrogateId, "0"),
+                });
```
brendankowitz, line 642: Feeling like this would be a useful extension i.e. queryParametersList.AddSurrogateIdRange(min, max, global)

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/Models/ReindexJobRecord.cs
v-mserdar: Great catch on this. I have now added a converter to address this situation as well as a test that demonstrates that an older version of the ReindexJobRecord json can still be deserialized correctly.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
v-mserdar: Well after searching the code base there's only 1 location, the one you have highlighted, that only adds those 3 values. The other locations are adding 4 or 5 parameters. The closest candidate is ExportJobTask. Let me know what you think about it.

feordin (approved):

PR #3168, opened 2023-03-09
Set up CI with Azure Pipelines

apurvabhaleMS:


security-build-pipelines.yml
```diff
@@ -0,0 +1,170 @@
...
+      inputs:
+        userProvideBuildInfo: 'autoMsBuildInfo'
+        setupCommandlinePicker: 'vs2022'
+        rulesetName: 'Multiple'
+        multipleRulesetRequired: true
+        multipleRulesetRecommended: true
+        customRuleset: 'CustomAnalysisRules.ruleset'
+        suppressionFileForCompilerWarnings: 'RoslynSuppressions.txt'
+      env:
+        SYSTEM_ACCESSTOKEN: $(System.AccessToken)
```
brendankowitz, line 125: Should the missing steps just be added into: https://github.com/microsoft/fhir-server/blob/main/build/jobs/analyze.yml?

This template already runs in PR and CI pipelines

apurvabhaleMS: Right that's where I started from https://github.com/microsoft/fhir-server/pull/3165/files
Trying to see if we can directly assign the full path to build.exe or if we chose to do autoMsBuildInfo then we will have to run the build again

PR #3169, opened 2023-03-10
Moves Cosmos Export to QueueClient

brendankowitz:
## Description
- Moves Cosmos Export to QueueClient
- Existing Cosmos Export jobs will continue to run under the old model (LegacyExportJobWorkerBackgroundService) to avoid disruption or file name changes

**Warning:** File names in Cosmos export will change for new jobs.

## Related issues
Addresses #3248

## Testing
Existing export tests
Manual testing 
 - Queueing jobs in main, checking out and running in this branch for continuity 

## FHIR Team Checklist


LTA-Thinking:
I think the legacy export code should be removed before merging this in. Having two code paths active at once will just create confusing code flow and unpredictable functionality. I've tested this locally without it and it works. I don't see any reason why it would still be needed.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/ResourceSurrogateIdHelper.cs
LTA-Thinking, line 1: Can this file be removed? It doesn't look like it is doing any work, it is just pass through methods.

LTA-Thinking:
The updated code doesn't seem able to find export jobs made the old way when a user tries to hit the job status endpoint.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Operations/CosmosFhirOperationDataStore.cs
```diff
@@ -86,45 +88,64 @@ public sealed class CosmosFhirOperationDataStore : IFhirOperationDataStore
...
-            var hash = JsonConvert.SerializeObject(hashObject).ComputeHash();
-
-            var resultFromHash = await GetExportJobByHashAsync(hash, cancellationToken);
-            if (resultFromHash != null)
+            try
             {
-                return resultFromHash;
+                return base.GetExportJobByIdAsync(id, cancellationToken);
             }
+            catch (JobNotFoundException)
```
LTA-Thinking, line 97: When a job isn't found Cosmos is throwing an aggregate exception that is wrapping the job not found exception. This isn't being caught right now.

brendankowitz: Hopefully this is fixed. Headpalm moment on the task await

src/Microsoft.Health.Fhir.CosmosDb/Features/Operations/Export/CosmosExportOrchestratorJob.cs
```diff
@@ -65,17 +65,17 @@ private static ExportJobRecord CreateExportRecord(ExportJobRecord record, long g
...
                         record.RollingFileSizeInMB,
                         record.RequestorClaims,
-                        since == null ? record.Since : since,
-                        till == null ? record.Till : till,
-                        startSurrogateId,
-                        endSurrogateId,
-                        globalStartSurrogateId,
-                        globalEndSurrogateId,
+                        since: null,
+                        till: null,
```
LTA-Thinking, line 74: Since and Till should be copied from the original record.

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Operations/Export/ExportProcessingJobTests.cs
```diff
@@ -19,7 +19,7 @@
 using NSubstitute;
 using Xunit;
 
-namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Operations.Export
+namespace Microsoft.Health.Fhir.SqlServer.UnitTests.Features.Operations.Export
```
LTA-Thinking, line 22: Namespace should be updated to Core.

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Operations/Export/ExportProcessingJobTests.cs
LTA-Thinking, line 1: Could this still be in the Core project instead of Shared Core?

src/Microsoft.Health.Fhir.Core/Features/Operations/FhirOperationDataStoreBase.cs
```diff
@@ -0,0 +1,254 @@
...
+        return CreateExportJobOutcome(jobId, result ?? def, jobInfo.Version, (byte)status, jobInfo.CreateDate);
+    }
+
+    public virtual async Task<ExportJobOutcome> UpdateExportJobAsync(ExportJobRecord jobRecord, WeakETag eTag, CancellationToken cancellationToken)
+    {
+        EnsureArg.IsNotNull(jobRecord, nameof(jobRecord));
+
+        if (jobRecord.Status != OperationStatus.Canceled)
+        {
+            throw new NotSupportedException($"Calls to this method with status={jobRecord.Status} are deprecated.");
```
LTA-Thinking, line 167: This exception is preventing legacy jobs from completing. The legacy jobs will still call this method to complete or fail, but won't get to the JobNotExistException below that will get caught. This exception also needs to be caught in the CosmosFhirOperationDataStore and call the legacy method.

src/Microsoft.Health.Fhir.Core/Features/Operations/FhirOperationDataStoreBase.cs
```diff
@@ -0,0 +1,254 @@
...
+            {
+                new EnumLiteralJsonConverter(),
+            },
+        };
+    }
+
+    public async Task<ExportJobOutcome> CreateExportJobAsync(ExportJobRecord jobRecord, CancellationToken cancellationToken)
+    {
+        var clone = jobRecord.Clone();
+        clone.QueuedTime = DateTime.Parse("1900-01-01");
```
PTaladay, line 46: Is there a reason we don't use something like DateTime.UtcNow? for the QueuedTime?

LTA-Thinking: This field is left over from the old way we handled export records. The standardized job records have queued time in them, so the specific export job record doesn't need it anymore. But we can't remove it yet because it would break backwards compatibility. So we set it to a value that is obviously wrong so that there isn't confusion on which queued time field is correct.

src/Microsoft.Health.Fhir.Core/Features/Operations/FhirOperationDataStoreBase.cs
```diff
@@ -0,0 +1,254 @@
...
+        var def = JsonConvert.SerializeObject(clone, _jsonSerializerSettings);
+        var results = await _queueClient.EnqueueAsync((byte)QueueType.Export, new[] { def }, null, false, clone.Status == OperationStatus.Completed, cancellationToken);
```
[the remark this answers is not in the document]
LTA-Thinking: That parameter allows for a job to be queued that has already been completed. I believe the thought was to allow for records to be recorded to track partial job completion. @SergeyGaluzo can you elaborate on the isCompleted parameter?

src/Microsoft.Health.Fhir.Core/Features/Operations/FhirOperationDataStoreBase.cs
brendankowitz: Thanks for finding this, updated the logic for detecting legacy jobs in the Cosmos method

LTA-Thinking (approved):

LTA-Thinking (approved):

PR #3191, opened 2023-03-22
Add metadata to health check

LTA-Thinking:
## Description
Adds a check to make sure the metadata document is available to the health check. Since the capabilities statement (aka metadata) is required by most endpoints if it is not available the service should be marked as unhealthy.

## Related issues
Addresses [issue #].

## Testing
Manual.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Conformance/ConformanceHealthCheck.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Conformance/ConformanceHealthCheck.cs
@@ -45,7 +45,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Conformance
                 }
                 else
                 {
-                    _logger.LogWarning("Capability statement was null.");
+                    _logger.LogError("Capability statement was null.");
 
                     return HealthCheckResult.Unhealthy("Failed to retrieve the capability statement.");
                 }
@@ -58,13 +58,13 @@ namespace Microsoft.Health.Fhir.Core.Features.Conformance
                     return HealthCheckResult.Healthy("Successfully retrieved the capability statement.");
                 }
 
-                _logger.LogWarning(ex, "Failed to retrieve the capability statement.");
+                _logger.LogError(ex, "Failed to retrieve the capability statement.");
 
                 return HealthCheckResult.Unhealthy("Failed to retrieve the capability statement.");
             }
             catch (Exception ex)
             {
-                _logger.LogWarning(ex, "Failed to retrieve the capability statement.");
+                _logger.LogError(ex, "Failed to retrieve the capability statement.");
 
                 return HealthCheckResult.Unhealthy("Failed to retrieve the capability statement.");
             }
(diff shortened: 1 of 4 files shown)


rajithaalurims (approved):

src/Microsoft.Health.Fhir.Core/Features/Conformance/ConformanceHealthCheck.cs
```diff
@@ -0,0 +1,73 @@
...
+            }
+            catch (ArgumentNullException ex)
+            {
+                if (ex.StackTrace.Contains("UrlHelperFactory", StringComparison.InvariantCultureIgnoreCase))
+                {
+                    // If the metadata document hasn't been initialized yet before health check is called the Url Helper factory will not have an action context from this call. The system is still healthy.
+                    return HealthCheckResult.Healthy("Successfully retrieved the capability statement.");
+                }
+
+                _logger.LogWarning(ex, "Failed to retrieve the capability statement.");
```
fhibf, line 61: Shouldn't we log these errors as exceptions?

src/Microsoft.Health.Fhir.Core/Features/Conformance/ConformanceHealthCheck.cs
```diff
@@ -0,0 +1,73 @@
...
+                    _logger.LogWarning("Capability statement was null.");
+
+                    return HealthCheckResult.Unhealthy("Failed to retrieve the capability statement.");
+                }
+            }
+            catch (ArgumentNullException ex)
+            {
+                if (ex.StackTrace.Contains("UrlHelperFactory", StringComparison.InvariantCultureIgnoreCase))
+                {
+                    // If the metadata document hasn't been initialized yet before health check is called the Url Helper factory will not have an action context from this call. The system is still healthy.
```
brendankowitz, line 57: Url Helper factory doesn't have an ActionContext I think because /health/check isn't an MVC route

fhibf: I think adding a MVC route to /health/check will be a more complex task. 
I've noticed how we are doing that for /metadata in FhirController, but that's something we should test more carefully.

LTA-Thinking: I don't think we should add a MVC route for health/check. Reading the documentation on health check endpoints, they are meant to have minimal dependencies and operate quickly. I think having it just check the cached metadata is sufficient for health/check, it shouldn't be trying to construct the metadata document. The special case we have here allows it to detect if the metadata document can't be constructed, but still allows the health check to be fast and reliable in a healthy system.

fhibf: I think we should spend more time finding a solution to this problem.
I don't like the check if the exception came from _UrlHelperFactory_, just because if another dependency to _ActionContext_ raises the _ArgumentNullException_ it'll break the /health/check.

LTA-Thinking:
This needs rework if we are going to add it.

PR #3198, opened 2023-03-27
default throttling values for both cosmosdb and sqlserver

v-mserdar:
## Description
This allows a default value for the throttling middleware for both cosmosdb and sqlserver to be set.

## Related issues
Addresses [issue #100374](https://microsofthealth.visualstudio.com/Health/_workitems/edit/100374).

## Testing
Successfully ran all tests in ThrottlingMiddlewareTests.cs

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Api/Features/Throttling/ThrottlingMiddleware.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/Throttling/ThrottlingMiddleware.cs
@@ -11,12 +11,15 @@ using System.Threading;
 using System.Threading.Tasks;
 using EnsureThat;
 using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Microsoft.Health.Abstractions.Exceptions;
 using Microsoft.Health.Fhir.Api.Configs;
 using Microsoft.Health.Fhir.Api.Features.Headers;
 using Microsoft.Health.Fhir.Core.Configs;
+using Microsoft.Health.Fhir.Core.Features;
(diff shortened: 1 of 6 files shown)


src/Microsoft.Health.Fhir.Api/Features/Throttling/ThrottlingMiddleware.cs
```diff
@@ -82,7 +88,19 @@ public sealed class ThrottlingMiddleware : IAsyncDisposable, IDisposable
             }
 
             // snapshot the configuration values to reduce the number of instructions that need to execute in the lock.
-            _concurrentRequestLimit = configuration.ConcurrentRequestLimit;
+            if (_baseConfiguration["DataStore"].Equals(KnownDataStores.CosmosDb, StringComparison.OrdinalIgnoreCase))
+            {
+                _concurrentRequestLimit = configuration.ConcurrentRequestLimit == 0 ? (int)ThrottlingLimitDefault.Gen1 : Math.Max(configuration.ConcurrentRequestLimit, (int)ThrottlingLimitDefault.Gen1);
```
PTaladay, line 93: Do we necessarily need to have the == 0 check? The Math.Max between the default and the config value should be enough. For example, if it is 0 then it would take the default value for gen1 instead.

v-mserdar: Change has been made.

PTaladay: Also, wouldn't we want to check against the "_concurrentRequestLimit" instead of calling the configuration.ConcurrentRequestLimit? I saw a comment regarding minimizing checking of the configuration directly and this is one of those values that they take a snapshot of.

PTaladay (approved):

PR #3205, opened 2023-03-29
Export intermediate state

LTA-Thinking:
## Description
Adds intermediate states back to non-parallel jobs. Jobs not running in parallel will now create new jobs for each page of results, allowing interrupted jobs to restart from the last successfully completed page of results instead of starting from the beginning.
This will slightly change the names of output files to bring them inline with the naming convention used for jobs run in parallel.
This should not have a noticeable impact on performance.

## Related issues
Addresses [issue #].

## Testing

## FHIR Team Checklist

Patch (reason)


SergeyGaluzo (approved):
:shipit:

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportProcessingJob.cs
```diff
@@ -74,11 +78,29 @@ public Task<string> ExecuteAsync(JobInfo jobInfo, IProgress<string> progress, Ca
...
+                    // 3. The export job has reached a terminal state
+                    // This section is only reached in cases 1 & 2.
+                    // For jobs running in parallel such that each job only processes one page of results this section will never be reached as there are no filters and no continuation tokens.
+                    // For single threaded jobs this section will complete a job after every page of results is processed and make a new job for the next page.
+                    // This will allow for saving intermediate states of the job.
+
+                    var definition = JsonConvert.DeserializeObject<ExportJobRecord>(jobInfo.Definition);
+                    definition.Progress = exportJobRecord.Progress;
+                    await _queueClient.EnqueueAsync((byte)QueueType.Export, new[] { JsonConvert.SerializeObject(definition) }, jobInfo.GroupId, false, false, innerCancellationToken);
+                    throw new JobSegmentCompletedException();
```
brendankowitz, line 100: This is a little strange to me. I don't think I've seen success via exception before :)

LTA-Thinking: It isn't the best pattern I will admit, but it makes this still work with the old Cosmos system. Once your PR is finished so Cosmos uses the new JobHosting I can refactor ExportJobTask to handle this better.

src/Microsoft.Health.TaskManagement/JobHosting.cs
```diff
@@ -204,6 +206,7 @@ await using (new Timer(async _ => await PutJobHeartbeatAsync(queueClient, jobInf
             }
         }
 
+        [Obsolete("Heartbeats should only update timestamp, results should only be written when job reaches terminal state.")]
```
brendankowitz, line 209: Still not sure on this. We update 1 field instead of 2. Still curious if this is more a design motive vs perf motive

SergeyGaluzo: It is not about how many fields are updated, but about how many times entire job record in the Result column is update and with what data. If thus record contains all intermediate states N and each state is saved as soon as it arrived, it leads to N*N problem (that is how much data is saved overall). This should be discouraged, hence obsolete tag.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportProcessingJob.cs
```diff
@@ -74,11 +78,29 @@ public Task<string> ExecuteAsync(JobInfo jobInfo, IProgress<string> progress, Ca
...
                 TaskScheduler.Current);
 
-            Task<ExportJobOutcome> UpdateExportJob(ExportJobRecord exportJobRecord, WeakETag weakETag, CancellationToken innerCancellationToken)
+            async Task<ExportJobOutcome> UpdateExportJob(ExportJobRecord exportJobRecord, WeakETag weakETag, CancellationToken innerCancellationToken)
             {
                 record = exportJobRecord;
                 progress.Report(JsonConvert.SerializeObject(exportJobRecord));
-                return Task.FromResult(new ExportJobOutcome(exportJobRecord, weakETag));
+
+                if (record.Status != OperationStatus.Completed && record.Status != OperationStatus.Failed)
```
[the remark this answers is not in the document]
LTA-Thinking: Queued shouldn't be possible for a job that is being worked on. When a job is dequeued it is put into the running state.
Running is the expected state for a job that enters this statement.
A job in Canceled state can't be being worked on. That state is only set on jobs in the database.
Paused and Unknown aren't used for Export jobs anywhere.

I could change it to only allow the Running state instead of excluding completed and failed.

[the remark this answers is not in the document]
SergeyGaluzo: I like to have only Running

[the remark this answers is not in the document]
LTA-Thinking: Updated

src/Microsoft.Health.TaskManagement/JobHosting.cs
brendankowitz: > about how many times entire job record in the Result column is updated and with what data.

I think the preference here is that Result should not contain the definition, it should only be data or state created or obtained by the action of the job.

> that is how much data is saved overall, actually square(N)/2 to be exact). This should be discouraged, hence obsolete tag.

If Result was updated new distinct values at each logical job checkpoint -- so assuming result is only updated when you would potentially insert a new record. Why is this worse than inserting an entirely new record that also duplicates most of the fields of the previous record?

SergeyGaluzo: The goal is never save distinct list of any values to a job, but rather each of those values separately. If we do that amount of data saved is proportional to N (not square(N)).

LTA-Thinking:
OoM fix didn't work. It is just crashing and not getting to the error handling.

SergeyGaluzo (dismissed):
:shipit:

PTaladay (approved):

PR #3234, opened 2023-04-17
Selective SearchParameter Update Logic

PTaladay:
## Description
Adds handler for update of search parameter status to either supported or disabled (pendingDisable saved to datastore), and accompanying unit tests.

Moved new methods in the SearchParameterStatusManager to extension methods and out of the interface since only one implementation needs to use them. 

Fixed issue where we always load search parameters from the initial flat files and overwrite what is in the datastore. 

## Related issues
Addresses [issue AB#102147, AB#101609].

## Testing
Added unit test to coverage validation logic in controller and handler. The update status calls to the status manager already have tests.

Manual calls while debugging an instance of the FHIR Server to validate the results come back as expected.

Example with Disable coming back as pendingDisable

## FHIR Team Checklist


rajithaalurims (commented):
@PTaladay 
I have a question about one scenario.

- In current version of code, customers can create Search parameter at base Resource level which gets applied to all resources.
- What is the expectation from customer on updating/searching status of those parameters? 
- Can customers disable status for a particular resource?
- What is the behavior when customers send update/search status request at base resource level?

**About test result screen shots:**

- And also, in the screen shot you attached for Bundle test scenario, where one search param is invalid, error message displays status instead of Uri.
- In the case of unsupported search parameter, I think we should show the same error message for both individual/Bundle PUTs. In the screenshots, I see two different messages in the response.

PTaladay:
> @PTaladay I have a question about one scenario.
> 
> * In current version of code, customers can create Search parameter at base Resource level which gets applied to all resources.
> * What is the expectation from customer on updating/searching status of those parameters?
> * Can customers disable status for a particular resource?
> * What is the behavior when customers send update/search status request at base resource level?
> 
> **About test result screen shots:**
> 
> * And also, in the screen shot you attached for Bundle test scenario, where one search param is invalid, error message displays status instead of Uri.
> * In the case of unsupported search parameter, I think we should show the same error message for both individual/Bundle PUTs. In the screenshots, I see two different messages in the response.

1.  This endpoint does not deal with creating or deleting search parameters. That is still handled within the searchparameter endpoint. To my knowledge, that search parameter at the base level would be applied to all of the resource types so we should expect to see resourcename-blah, resourcename2-blah, etc. 
2. Customers can disable for a particular resource. I'll double check with Jared, but right now the customer would need to query for all search parameters by resource type in the search portion of $status. That result can be copied and updated and sent to the put request. I can go and add in the disabling by resource type directly.
3. Can you give me an example so I can validate? Not sure what is meant by base resource level. 
4. For the status showing in the error message, that is a bug. I may not have updated that screenshot since I made those code changes. I'll go double check that. 
5. There are two different cases represented here. The first is the status that the customer is trying to use is not one we allow, such as "Enabled". The second is a customer trying to make a change to a search parameter that is in a deleted state. This we don't allow regardless of whether the status they sent is valid or not.

feordin:
> > @PTaladay I have a question about one scenario.
> > 
> > * In current version of code, customers can create Search parameter at base Resource level which gets applied to all resources.
> > * What is the expectation from customer on updating/searching status of those parameters?
> > * Can customers disable status for a particular resource?
> > * What is the behavior when customers send update/search status request at base resource level?
> > 
> > **About test result screen shots:**
> > 
> > * And also, in the screen shot you attached for Bundle test scenario, where one search param is invalid, error message displays status instead of Uri.
> > * In the case of unsupported search parameter, I think we should show the same error message for both individual/Bundle PUTs. In the screenshots, I see two different messages in the response.
> 
> 1. This endpoint does not deal with creating or deleting search parameters. That is still handled within the searchparameter endpoint. To my knowledge, that search parameter at the base level would be applied to all of the resource types so we should expect to see resourcename-blah, resourcename2-blah, etc.
> 2. Customers can disable for a particular resource. I'll double check with Jared, but right now the customer would need to query for all search parameters by resource type in the search portion of $status. That result can be copied and updated and sent to the put request. I can go and add in the disabling by resource type directly.
> 3. Can you give me an example so I can validate? Not sure what is meant by base resource level.
> 4. For the status showing in the error message, that is a bug. I may not have updated that screenshot since I made those code changes. I'll go double check that.
> 5. There are two different cases represented here. The first is the status that the customer is trying to use is not one we allow, such as "Enabled". The second is a customer trying to make a change to a search parameter that is in a deleted state. This we don't allow regardless of whether the status they sent is valid or not.

@PTaladay @rajithaalurims For the question about a SearchParameter that applies to multiple resource types:  As we specify a SearchParameter by URL to change its status, then any change to a SearchParameter to disable it or enable it would apply to all the resource types.  If it is a base resource type, then all the "children" types that are in the hierarchy would be affected.  If a SearchParameter applies to multiple resource types, then you cannot simply disable that SearchParameter for one resource type.  You disable or enable the SearchParameter entirely.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/SearchParameterState/SearchParameterStateUpdateHandler.cs
```diff
@@ -0,0 +1,184 @@
...
+                }
+                else if (searchParameterInfo.Status.Equals(SearchParameterStatus.Deleted) || searchParameterInfo.Status.Equals(SearchParameterStatus.Unsupported))
+                {
+                    invalidSearchParameters.Add(new OperationOutcomeIssue(
+                        OperationOutcomeConstants.IssueSeverity.Error,
+                        OperationOutcomeConstants.IssueType.NotSupported,
+                        detailsText: string.Format(Core.Resources.SearchParameterDeleted, uri)));
+                }
+                else
+                {
```
feordin, line 92: Just a thought, make a new variable statusKey.  If status == disabled, statusKey = pendingDisabled.  Then use statusKey for all the following lines, which may remove one level of nested if statements.

feordin (dismissed):
Just left a suggestion for possibly simplifying one piece of code.

feordin (approved):

feordin (approved):

PR #3240, opened 2023-04-18
Refactor for IConvertDataTemplateProviders

ShaunDonn:
Separated logic for default template use case and custom template use case into a DefaultTemplateProvider and a CustomTemplateProvider, which inherits from the DefaultTemplateProvider.


ShaunDonn2:
Draft PR no longer relevant. Solution was merged through [Convert Refactor for Missing External MI Case #3263](https://github.com/microsoft/fhir-server/pull/3263).

PR #3250, opened 2023-04-21
changes to support a list for ScopesClaim

v-mserdar:
## Description
Revised ScopesClaim from a string to a list to support the following:

For certain scenarios AAD includes clinical scopes values in the roles claim.  We should include examining that claim for scopes.  We can examine any value in the claim using the regex for clinical scopes and then include them in the collection.

For any scenario where the caller to the FHIR Api is an app, rather than a user, scenario such as bulk export, with a scope of system.read.all then  we are not interpreting the clinical scope currently.  But as bulk export has only the one scope of system.read.all and we don't currently support any other system scopes, it is not currenly a problem. If not addressed right now, we will forget and will be tough to tackle later

## Related issues
Addresses [issue #98439](https://microsofthealth.visualstudio.com/Health/_workitems/edit/98439).

## Testing
Successfully ran the SmartClinicalScopesMiddlewareTests

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Api/Features/SMART/SmartClinicalScopesMiddleware.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/SMART/SmartClinicalScopesMiddleware.cs
@@ -77,84 +77,87 @@ namespace Microsoft.Health.Fhir.Api.Features.Smart
 
                     // examine the scopes claim for any SMART on FHIR clinical scopes
                     DataActions permittedDataActions = 0;
-                    foreach (Claim claim in principal.FindAll(authorizationConfiguration.ScopesClaim))
+                    foreach (string singleScope in authorizationConfiguration.ScopesClaim)
                     {
-                        var matches = ClinicalScopeRegEx.Matches(claim.Value);
-                        foreach (Match match in matches)
+                        foreach (Claim claim in principal.FindAll(singleScope))
                         {
-                            fhirRequestContext.AccessControlContext.ClinicalScopes.Add(match.Value);
-
-                            var id = match.Groups["id"]?.Value;
-                            var resource = match.Groups["resource"]?.Value;
-                            var accessLevel = match.Groups["accessLevel"]?.Value;
-
-                            switch (accessLevel)
+                            var matches = ClinicalScopeRegEx.Matches(claim.Value);
+                            foreach (Match match in matches)
                             {
-                                case "read":
-                                    permittedDataActions |= DataActions.Read;
-                                    break;
-                                case "write":
-                                    permittedDataActions |= DataActions.Write;
-                                    break;
-                                case "*":
-                                case AllDataActions:
-                                    permittedDataActions |= DataActions.Read | DataActions.Write | DataActions.Export;
-                                    break;
-                            }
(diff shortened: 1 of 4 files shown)


src/Microsoft.Health.Fhir.Api/Features/SMART/SmartClinicalScopesMiddleware.cs
```diff
@@ -77,84 +77,87 @@ public SmartClinicalScopesMiddleware(RequestDelegate next, ILogger<SmartClinical
 
                     // examine the scopes claim for any SMART on FHIR clinical scopes
                     DataActions permittedDataActions = 0;
-                    foreach (Claim claim in principal.FindAll(authorizationConfiguration.ScopesClaim))
+                    foreach (string singleScope in authorizationConfiguration.ScopesClaim)
```
feordin, line 80: I think this foreach includes too much of the file.  It also wraps up the fhirUserClaim processing which doesn't need to be repeated multiple times.  It doesn't look like there is any harm in running it more than needed, but I feel this could be a source of errors later.  I think we should have this foreach iterate over all the claims in the list, and then perhaps build one concatenated string of the values from the different claims, then this longer string could be passed to the regex to identify all the scopes.

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/SMART/SmartClinicalScopesMiddlewareTests.cs
```diff
@@ -70,18 +70,22 @@ public async Task GivenSmartScope_WhenInvoked_ThenScopeParsedandAddedtoContext(s
 
             var fhirUserClaim = new Claim(authorizationConfiguration.FhirUserClaim, "https://fhirServer/Patient/foo");
             var rolesClaim = new Claim(authorizationConfiguration.RolesClaim, "smartUser");
-            var scopesClaim = new Claim(authorizationConfiguration.ScopesClaim, scopes);
-            var claimsIdentity = new ClaimsIdentity(new List<Claim>() { scopesClaim, rolesClaim, fhirUserClaim });
-            var expectedPrincipal = new ClaimsPrincipal(claimsIdentity);
 
-            httpContext.User = expectedPrincipal;
-            fhirRequestContext.Principal = expectedPrincipal;
+            foreach (string singleClaim in authorizationConfiguration.ScopesClaim)
```
feordin, line 74: This is a good change, where each of the different scopes are tested separately, but we also need some new tests where the different scopes are spread across the multiple claims.

For example, claim scope = "patient.Patient.read" and role claim = practitioner.Observation.write"  In the end both are picked up and applied.

feordin (changes requested):
I added couple of comments about reducing the bounds the new foreach, and also adding a new unit test.

feordin (changes requested):
I realized that we need one more change for this PR.  The value in appsettings.json is currently this:  "ScopesClaim": "scope" 

Now that the setting is update to List<string> , the single string in the appsettings does not parse in and work.

It should be changed to:
"ScopesClaim": [ "scope" ]

feordin (approved):

Issue #2818, opened 2022-09-22
SQL Exception 8623 The query processor ran out of internal resources and could not produce a query plan


akatakritos:
**Describe the bug**
Hi, we're running this service in Azure PaaS as the "Azure Health Data Services - FHIR" and recently started getting this exception on a query that used to work fine. This is reproducible in both our QA and Production environments.

> Operation was unsuccessful, and returned status InternalServerError. OperationOutcome: Overall result: FAILURE (1 errors and 0 warnings)
>[ERROR] (no details)(further diagnostics: SqlException number: 8623; Exception: The query processor ran out of internal resources and could not produce a query plan. This is a rare event and only expected for extremely complex queries or queries that reference a very large number of tables or partitions. Please simplify the query. If you believe you have received this message in error, contact Customer Support Services for more information.). 

GET https://host/ServiceRequest?_count=20&_include=ServiceRequest%3Apatient&_include=ServiceRequest%3Aspecimen&_sort=-_lastUpdated&resultStatus=Closed&intent=original-order&status=active%2Ccompleted%2Cunknown&collectionDate=gt2022-05-31&collectionDate=lt2022-05-31T23:59&subject%3APatient.address-state%3Aexact=AL%2CAS%2CAZ%2CAR%2CCA%2CCO%2CCT%2CDE%2CDC%2CFL%2CGA%2CGU%2CHI%2CID%2CIL%2CIN%2CIA%2CKS%2CKY%2CLA%2CME%2CMD%2CMA%2CMI%2CMN%2CMS%2CMO%2CMT%2CNE%2CNV%2CNH%2CNJ%2CNM%2CNY%2CNC%2CND%2CMP%2COH%2COK%2COR%2CPA%2CPR%2CRI%2CSC%2CSD%2CTN%2CTX%2CUT%2CVT%2CVI%2CVA%2CWA%2CWV%2CWI%2CWY%2CAK%2CUNKNOWN&category%3Anot=Toxicology&_total=accurate 

I'm not quite sure if its related to the number of parameters or the combination of them. Removing `category:not` still fails, but removing one of the others causes it to pass.

Some of those search params are custom ones, but from my udnerstanding of the FHIR Server architecture, theres no real difference in database schema design between a custom and built in SearchParam. But they are defined as follows:

[...]

PR #3267, opened 2023-04-28
Alter SQL generator to use INNER JOIN for ReferenceSearchParam

feordin:
## Description
At times with complex FHIR queries, the resulting SQL query is very complex and SQL server can timeout attempting to generate the query plan.  We use an EXISTS clause to link References to the target type, but it reduces the burden on SQL server if we use an INNER JOIN.

This change alters the SQL query generator to use an INNER JOIN for ReferenceSearchParams.

## Related issues
Addresses [issue #2818].

## Testing
This was shown to reproduce with member-match and that test has been added.  Also, the specific query mentioned in #2818 was tested manually, and shown to have a 50% reduction in time to produce a query plan.

## FHIR Team Checklist

### DIFF
--- a/docs/rest/ComplexSQLQueryServiceRequest.http
+++ b/docs/rest/ComplexSQLQueryServiceRequest.http
@@ -1,0 +1,129 @@
+@hostname = localhost:44348
+
+### Test rest client
+https://{{hostname}}/metadata
+
+### Get the globalAdminServicePrincipal to be able to POST test data
+# @name bearer
+POST https://{{hostname}}/connect/token
+content-type: application/x-www-form-urlencoded
+
+grant_type=client_credentials
+&client_id=globalAdminServicePrincipal
+&client_secret=globalAdminServicePrincipal
+&scope=patient/Observation.read fhirUser fhir-api user/Encounter.*
+
+### Post 1st custom search parameter
+POST https://{{hostname}}/SearchParameter
+content-type: application/json
+Authorization: Bearer {{bearer.response.body.access_token}}
+
+{
+      "resourceType" : "SearchParameter",
+      "id" : "ServiceRequest-resultStatus",
+      "extension" : [{
+        "url" : "http://hl7.org/fhir/StructureDefinition/structuredefinition-standards-status",
+        "valueCode" : "trial-use"
+      }],
+      "url" : "http://hl7.org/fhir/SearchParameter/ServiceRequest-resultStatus",
+      "version" : "4.0.1",
+      "name" : "resultStatus",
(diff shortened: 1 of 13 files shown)


SergeyGaluzo:
Changing from EXISTS to JOIN for M:M tables introduced more ambiguity for SQL. We can use JOIN for queries that work today. We can try it only as last resort for query SQL cannot generate plan for.

feordin:
> Changing from EXISTS to JOIN for M:M tables introduced more ambiguity for SQL, and can have performance impact. 
> I suggest not to use JOIN for queries that work today. We can try it only as last resort for query SQL cannot generate plan for.   

For the query in question, the customer currently is blocked as their query regularly has issues completing the query plan generation.  On my dev box, with no data, it requires about 6 seconds to generate the query plan.  With this change it is reduced to 3 seconds.

I could change this PR to be more targeted and limit the conditions when the JOIN is used.

SergeyGaluzo:
When M:M is involved using JOIN in filtering introduces duplicates in the result set which need to be filtered later. In this respect JOIN is strategically incorrect, because it increases chances  that inefficient plan is generated. Creating "bad" plan faster should not be considered as good.
I don't see another way to deal with this problem besides having 2 paths: 1 - with EXISTS, and 2 - only if #1 fails - JOIN.   

---
In reply to: [1532023315](https://github.com/microsoft/fhir-server/pull/3267#issuecomment-1532023315) [](http://example.com/codeflow?ancestors=1532023315,1532023315)

feordin:
> When M:M is involved using JOIN in filtering introduces duplicates in the result set which need to be filtered later. In this respect JOIN is strategically incorrect, because it increases chances that inefficient plan is generated. Creating "bad" plan faster should not be considered as good. I don't see another way to deal with this problem besides having 2 paths: 1 - with EXISTS, and 2 - only if 1 fails - JOIN.
> 
> In reply to: [1532023315](https://github.com/microsoft/fhir-server/pull/3267#issuecomment-1532023315)

However, in the case of failure, the result is a timeout.  How long do we wait?  How do we distinguish between a long running query, and a query that is still attempting to generate a query plan?

SergeyGaluzo:
That is a good point. 
What we know is that there is a limited number of EXISTS SQL can handle. 
Can't we see how many we are going to add, if it is too many switch to JOINs?

---
In reply to: [1532238689](https://github.com/microsoft/fhir-server/pull/3267#issuecomment-1532238689) [](http://example.com/codeflow?ancestors=1532238689,1532238689)

feordin:
build issues, will close and recreate this PR.

PR #3280, opened 2023-05-04
Throw specific authentication exceptions

rajithaalurims:
## Description
After our latest changes to Authentication library migration, we are throwing a wrapper exception (S2SAuthenticationException) instead of specific individuals like SecurityTokenInvalidIssuerException, SecurityTokenInvalidAudienceException.

I have added middleware extension code to intercept the error and throw specific errors in case of invalid audience and issuer.

## Related issues
Addresses [issue #[102851](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Olympus/Stories/?workitem=102851)].

## Testing
Unit tests.
Tested happy scenario in Postman.

## FHIR Team Checklist


src/Microsoft.Health.Fhir.Api/Features/Context/FhirAuthenticationExceptionHandlerMiddleware.cs
```diff
@@ -0,0 +1,51 @@
...
+                {
+                    throw new SecurityTokenInvalidAudienceException(ex.InnerException.Message, ex.InnerException);
+                }
+                else if (ex.InnerException is SecurityTokenInvalidIssuerException)
+                {
+                    throw new SecurityTokenInvalidIssuerException(ex.InnerException.Message, ex.InnerException);
+                }
+                else
+                {
+                    throw ex;
```
fhibf, line 46: I would replace "throw ex;" with "throw;".

"throw;" will preserve the original call stack.

src/Microsoft.Health.Fhir.Api/Features/Context/FhirAuthenticationExceptionHandlerMiddleware.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            try
+            {
+                await _next(context);
+            }
+            catch (S2SAuthenticationException ex)
+            {
+                if (ex.InnerException is SecurityTokenInvalidAudienceException)
+                {
+                    throw new SecurityTokenInvalidAudienceException(ex.InnerException.Message, ex.InnerException);
```
brendankowitz, line 38: Could you just re-throw the inner exception here?
```suggestion
                    throw ex.InnerException;
```

src/Microsoft.Health.Fhir.Api/Features/Context/FhirAuthenticationExceptionHandlerMiddleware.cs
rajithaalurims: VS suggested that actually and I thought I changed it to just throw to keep the call stack. Good catch. Fixed it now.

src/Microsoft.Health.Fhir.Api/Features/Context/FhirAuthenticationExceptionHandlerMiddleware.cs
rajithaalurims: Yes. changed it now.

src/Microsoft.Health.Fhir.Api/Features/Context/FhirAuthenticationExceptionHandlerMiddleware.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            try
+            {
+                await _next(context);
+            }
+            catch (S2SAuthenticationException ex)
+            {
+                if (ex.InnerException is SecurityTokenInvalidAudienceException || ex.InnerException is SecurityTokenInvalidIssuerException)
+                {
+                    throw ex.InnerException;
```
brendankowitz, line 37: Sorry I'm iterating here, but even better:

```suggestion
                    ExceptionDispatchInfo.Capture(ex.InnerException).Throw();
```

This should allow a similar behavior to what Fernando mentioned and restore the original execution context and call stack, instead of looking like it stopped in this Middleware.

src/Microsoft.Health.Fhir.Api/Features/Context/FhirAuthenticationExceptionHandlerMiddleware.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        public async Task Invoke(HttpContext context)
+        {
+            try
+            {
+                await _next(context);
+            }
+            catch (S2SAuthenticationException ex)
+            {
+                if (ex.InnerException is SecurityTokenInvalidAudienceException || ex.InnerException is SecurityTokenInvalidIssuerException)
```
fhibf, line 36: I'm curious if SecurityTokenInvalidAudienceException  and SecurityTokenInvalidIssuerException are the only exception types we should add here. 

Probably we should  if ex.InnerException is a [SecurityTokenException](https://learn.microsoft.com/en-us/dotnet/api/microsoft.identitymodel.tokens.securitytokenexception?view=msal-web-dotnet-latest) , then it should be true.

fhibf: SecurityTokenException is the base class for SecurityTokenInvalidIssuerException, SecurityTokenInvalidAudienceException , and other token related exceptions.

src/Microsoft.Health.Fhir.Api/Features/Context/FhirAuthenticationExceptionHandlerMiddleware.cs
```diff
@@ -0,0 +1,48 @@
...
+            }
+            catch (Exception ex)
+            {
+                if (ex.InnerException is SecurityTokenException)
+                {
+                    ExceptionDispatchInfo.Capture(ex.InnerException).Throw();
+                }
+                else
+                {
+                    throw;
```
fhibf, line 43: @brendankowitz I believe the intension is to use _ExceptionDispatchInfo.Capture(...).Throw()_ in both situations to keep the original call stack. 
The difference would be: if the inner exception is a SecurityTokenException, then we would throw the inner exception (using _ExceptionDispatchInfo.Capture(...).Throw()_ ), otherwise, throw the original exception (using _ExceptionDispatchInfo.Capture(...).Throw()_ ).

fhibf (approved):

PR #3291, opened 2023-05-10
Updated Reindex to use Pending States for Search Parameters

PTaladay:
## Description
Updated reindex operation to use "PendingDelete" and "PendingDisable". Both now update to their respective final states of "Deleted" and "Disabled" and remove the no longer needed search parameter references.

## Related issues
Addresses [issue AB#101606].

## Testing
Ran through and updated existing Reindex tests along with manually testing the pending states using the new $status operation before calling reindex. 

## FHIR Team Checklist


src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -766,17 +773,32 @@ private async Task<(int totalCount, List<string> resourcesTypes)> CalculateTotal
...
 
-        private async Task UpdateParametersAndCompleteJob()
+        private async Task UpdateParametersAndCompleteJob(CancellationToken cancellationToken)
         {
             // here we check if all the resource types which are base types of the search parameter
             // were reindexed by this job.  If so, then we should mark the search parameters
             // as fully reindexed
             var fullyIndexedParamUris = new List<string>();
             var reindexedResourcesSet = new HashSet<string>(_reindexJobRecord.Resources);
+            var searchParamStatusCollection = await _searchParameterStatusManager.GetAllSearchParameterStatus(cancellationToken);
```
feordin, line 783: Seeing this request to update the current status reminds me that perhaps we need to block any requests to change SearchParam status while a reindex job is running.

We have this check when a user creates or updates a CustomSearch parameter, as we don't want them introducing problems for the currently running reindex job.  We should think about adding this same restriction for the $status operation.  This is currently done for Search Parameters in this file: \fhir-server\src\Microsoft.Health.Fhir.Shared.Core\Features\Search\Parameters\SearchParameterValidator.cs.

This is not something that needs to be done in this PR, but we should capture the need to analyze this.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -766,17 +773,32 @@ private async Task<(int totalCount, List<string> resourcesTypes)> CalculateTotal
...
+                {
+                    _logger.LogInformation("Reindex job updating the status of the fully indexed search, id: {Id}, parameter: '{ParamUri}' to Disabled.", _reindexJobRecord.Id, searchParam);
+                    await _searchParameterStatusManager.UpdateSearchParameterStatusAsync(new List<string>() { searchParamInfo.Url.ToString() }, SearchParameterStatus.Disabled, cancellationToken);
+                }
+                else if (spStatus == Search.Registry.SearchParameterStatus.PendingDelete)
+                {
+                    _logger.LogInformation("Reindex job updating the status of the fully indexed search, id: {Id}, parameter: '{ParamUri}' to Deleted.", _reindexJobRecord.Id, searchParam);
+                    await _searchParameterStatusManager.UpdateSearchParameterStatusAsync(new List<string>() { searchParamInfo.Url.ToString() }, SearchParameterStatus.Deleted, cancellationToken);
+                }
+                else if (reindexedResourcesSet.IsSupersetOf(searchParamInfo.BaseResourceTypes) && (spStatus == SearchParameterStatus.Supported || spStatus == SearchParameterStatus.Enabled))
```
feordin, line 801: Do we need a similar supserset check on the pendingDisable or PendingDelete search parameters, in case the reindex job was run for only a subset of the resource types for the pending SearchParameters?

feordin (changes requested):
Added a couple of questions, for clarification.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
PTaladay: That is a good point. I will make note of it.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
PTaladay: This change is just to make sure that the pending disable and pending delete don't get set to enabled, which was happening previously. This just makes sure they don't get included in the list for search parameter updates to a status of enabled. Maybe I'm not understanding what you are asking.

feordin: On line 801, we only update supported -> Enabled if the resource types that were reindexed, were the complete list of resource types the search parameter applied to.

But on lines 791 and 796, we just check the status of the search parameter being in a pending state.  It's possible that we have a SearchParameter X  which applies to resource types 1,2 and 3.  Then we set X to pendingDisable.  Then we run a reindex job for only resource type 2.  resource types 1 and 3 still need to be cleaned up, but it hasn't happened yet.  We don't want to mark SearchParameter X to Disabled, because it's not really finished yet.  We should leave it as pending.

PTaladay: I see, that makes sense. Let me look into that as we should have the same type of check.

PTaladay: Ok, I moved it to the beginning, and only update any of the statuses if that checks out first. Thanks for the catch.

feordin (approved):

PR #3292, opened 2023-05-10
Only Allow AHDS for Selectable SearchParameters

PTaladay:
## Description
Added in a runtime configuration for Azure Api for FHIR and Azure Health Data Services that share a common interface. This explicitly sets whether selectable search parameters can be used. Configuration is passed through DI.

## Related issues
Addresses [issue AB#103193].

## Testing
Added unit test and manually ran tests to verify as well.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Api/Controllers/SearchParameterController.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Api/Controllers/SearchParameterController.cs
@@ -13,18 +13,17 @@ using MediatR;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Options;
 using Microsoft.Health.Api.Features.Audit;
-using Microsoft.Health.Fhir.Api.Configs;
 using Microsoft.Health.Fhir.Api.Extensions;
 using Microsoft.Health.Fhir.Api.Features.ActionResults;
 using Microsoft.Health.Fhir.Api.Features.Filters;
 using Microsoft.Health.Fhir.Api.Features.Routing;
 using Microsoft.Health.Fhir.Core.Configs;
 using Microsoft.Health.Fhir.Core.Exceptions;
-using Microsoft.Health.Fhir.Core.Features;
 using Microsoft.Health.Fhir.Core.Features.Operations;
 using Microsoft.Health.Fhir.Core.Features.Operations.SearchParameterState;
 using Microsoft.Health.Fhir.Core.Features.Search.Registry;
 using Microsoft.Health.Fhir.Core.Messages.SearchParameterState;
+using Microsoft.Health.Fhir.Core.Registration;
 using Microsoft.Health.Fhir.ValueSets;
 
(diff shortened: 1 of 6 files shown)


src/Microsoft.Health.Fhir.Shared.Api/Controllers/SearchParameterController.cs
```diff
@@ -35,14 +37,19 @@ public class SearchParameterController : Controller
...
         {
             EnsureArg.IsNotNull(mediator, nameof(mediator));
             EnsureArg.IsNotNull(coreFeatures?.Value, nameof(coreFeatures));
+            EnsureArg.IsNotNull(throttlingConfig?.Value, nameof(throttlingConfig));
 
             _mediator = mediator;
             _coreFeaturesConfig = coreFeatures.Value;
+
+            // Used to prevent Azure Api for Fhir from using this controller.
+            _isAzureApiForFhir = throttlingConfig.Value.DataStore.Equals(KnownDataStores.CosmosDb, StringComparison.OrdinalIgnoreCase);
```
fhibf, line 52: Is this the official way to check if a FHIR instance is running Gen1 or Gen2?

feordin: It is a good question, but it seems that this may be the only place that we save the datastore value in the config objects.  I think it will work.

PTaladay: No, but we do use it to determine the throttling config default for the concurrency limit between gen1 and gen2. Otherwise I'm not aware of any "official" way to determine it other than the datastore.

fhibf: I think that our solution is based on using dependency injection to solve this type of problem, and not have this type of "conditional expressions". 
PRobably, it would be the ideal to have an interface that has two different implementation per data source, and it's initialized during Startup.

PTaladay: Reworked this based on Fernando's suggestion. Thanks!

src/Microsoft.Health.Fhir.Shared.Api/Controllers/SearchParameterController.cs
```diff
@@ -35,14 +36,19 @@ public class SearchParameterController : Controller
     {
         private readonly IMediator _mediator;
         private readonly CoreFeatureConfiguration _coreFeaturesConfig;
+        private bool _isSelectSearchParameterSupported;
 
-        public SearchParameterController(IMediator mediator, IOptions<CoreFeatureConfiguration> coreFeatures)
+        public SearchParameterController(IMediator mediator, IOptions<CoreFeatureConfiguration> coreFeatures, IFhirRuntimeConfiguration fhirConfig)
```
fhibf, line 41: Fancy! 🤩

fhibf (approved):

PR #3318, opened 2023-05-31
Ensure $convert template providers read from cache

wi-y:
## Description
When receiving many $convert-data requests concurrently it is possible for many template providers to be created. Each template provider has a cache where the contents of the template are cached. In the case where the templates have not yet been cached or the cache has expired, and when receiving many requests at once, it is possible for each template provider to request template data from the ACR which results in a lot of pulls from the ACR. To address this, we will be caching the template provider and limiting the number of threads that can create a new template provider to 1. This ensures that at most only one provider will be created and will read from the ACR. All other requests will read from the cache.

## Related issues
Addresses work item#104027.

## Testing
Before change: 
Created a console app that sends 1000 concurrent $convert-data requests.
The application code was observed to pull the same image from ACR many times and was eventually blocked from reading from ACR. Many $convert-data requests fail with HTTP 500 errors when unable to pull from ACR

After change:
Send 1000 $convert-data requests
The application code reads from ACR once and all $convert-data requests are fulfilled.

Patch

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Operations/ConvertData/ContainerRegistryTemplateProvider.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Operations/ConvertData/ContainerRegistryTemplateProvider.cs
@@ -17,6 +17,7 @@ using Microsoft.Health.Fhir.Core.Configs;
 using Microsoft.Health.Fhir.Core.Features.Operations.ConvertData.Models;
 using Microsoft.Health.Fhir.Core.Messages.ConvertData;
 using Microsoft.Health.Fhir.TemplateManagement;
+using Microsoft.Health.Fhir.TemplateManagement.ArtifactProviders;
 using Microsoft.Health.Fhir.TemplateManagement.Exceptions;
 
 namespace Microsoft.Health.Fhir.Core.Features.Operations.ConvertData
@@ -28,6 +29,8 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.ConvertData
         private readonly ITemplateCollectionProviderFactory _templateCollectionProviderFactory;
         private readonly ConvertDataConfiguration _convertDataConfig;
         private readonly MemoryCache _cache;
+        private MemoryCache _templateProviderCache;
+        private SemaphoreSlim _templateProviderFactorySemaphore;
         private readonly ILogger<ContainerRegistryTemplateProvider> _logger;
 
         public ContainerRegistryTemplateProvider(
@@ -49,6 +52,9 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.ConvertData
                 SizeLimit = _convertDataConfig.CacheSizeLimit,
             });
             _templateCollectionProviderFactory = new TemplateCollectionProviderFactory(_cache, Options.Create(_convertDataConfig.TemplateCollectionOptions));
+
+            _templateProviderCache = new MemoryCache(new MemoryCacheOptions());
+            _templateProviderFactorySemaphore = new SemaphoreSlim(1, 1);
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


src/Microsoft.Health.Fhir.Core/Features/Operations/ConvertData/ContainerRegistryTemplateProvider.cs
```diff
@@ -73,7 +79,31 @@ async Task<string> TokenEntryFactory(ICacheEntry entry)
 
             try
             {
-                var provider = _templateCollectionProviderFactory.CreateTemplateCollectionProvider(request.TemplateCollectionReference, accessToken);
+                /*
+                   First try to get the template provider from the cache.
+                   If template provider is not in cache, then limit the number of threads that can create new providers to 1, and then create a new provider.
+                   This is to limit the amount of providers created in a multi-threaded scenario, so that we do not have multiple providers pulling the same image from the ACR.
```
ShaunDonn, line 85: Just curious, I know you mentioned that the ACR eventually starts throwing 500s when there are many requests for the same image. Does this not happen if there are many requests for different images within the same ACR?

wi-y: It seems related to the total number of requests, or total bandwidth pulling from the ACR in general. So it seems to still throw an error even when pulling different images.

[The ACR service tier limits/throttling docs](https://learn.microsoft.com/en-us/azure/container-registry/container-registry-skus) do state that an HTTP 429 should be returned when too many requests are made, but it appears that is not really what is being observed. It could be that the server (ACR) is returning the appropriate error and the client (ACR SDK client) is not handling it for whatever reason. Or that the server is just denying the connection. But in any case we probably should be just making 1 (or maybe a few) requests to pull from the ACR instead of potentially hundreds. I think that was the intent with the caching in the original implementation but think it just needed to be tweaked slightly to account for concurrency.

For reference the error is:

Pull Image Failed.System.Net.Http.HttpRequestException: The SSL connection could not be established, see inner exception.
---> System.IO.IOException:  Received an unexpected EOF or 0 bytes from the transport stream.
at System.Net.Security.SslStream.ReceiveBlobAsync[TIOAdapter](TIOAdapter adapter)
at System.Net.Security.SslStream.ForceAuthenticationAsync[TIOAdapter](TIOAdapter adapter, Boolean receiveFirst, Byte[] reAuthenticationData, Boolean isApm)
at System.Net.Http.ConnectHelper.EstablishSslConnectionAsyncCore(Boolean async, Stream stream, SslClientAuthenticationOptions sslOptions, CancellationToken cancellationToken)
--- End of inner exception stack trace ---
at Microsoft.Rest.RetryDelegatingHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
at System.Net.Http.HttpClient.SendAsyncCore(HttpRequestMessage request, HttpCompletionOption completionOption, Boolean async, Boolean emitTelemetryStartStop, CancellationToken cancellationToken)
at Microsoft.Azure.ContainerRegistry.BlobOperations.GetWithHttpMessagesAsync(String name, String digest, Dictionary`2 customHeaders, CancellationToken cancellationToken)
at Microsoft.Azure.ContainerRegistry.BlobOperationsExtensions.GetAsync(IBlobOperations operations, String name, String digest, CancellationToken cancellationToken)
at Microsoft.Health.Fhir.TemplateManagement.Client.AcrClient.PullBlobAsStreamAcync(String name, String digest, CancellationToken cancellationToken)

src/Microsoft.Health.Fhir.Core/Features/Operations/ConvertData/ContainerRegistryTemplateProvider.cs
```diff
@@ -73,7 +79,31 @@ async Task<string> TokenEntryFactory(ICacheEntry entry)
...
+                */
+                ITemplateCollectionProvider provider = _templateProviderCache.Get(request.TemplateCollectionReference) as TemplateCollectionProvider;
```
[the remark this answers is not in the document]
wi-y: Thanks for the review. Yeah it would be better to cast to the interface `ITemplateCollectionProvider` rather than over casting to `TemplateCollectionProvider`. Updated the code to cast to `ITemplateCollectionProvider`

ShaunDonn (approved):

src/Microsoft.Health.Fhir.Core/Features/Operations/ConvertData/ContainerRegistryTemplateProvider.cs
```diff
@@ -28,6 +29,8 @@ public class ContainerRegistryTemplateProvider : IConvertDataTemplateProvider, I
         private readonly ITemplateCollectionProviderFactory _templateCollectionProviderFactory;
         private readonly ConvertDataConfiguration _convertDataConfig;
         private readonly MemoryCache _cache;
+        private MemoryCache _templateProviderCache;
```
fhibf, line 32: Should we mark these private attributes are readonly? As I see, they are only assigned in the constructor.

wi-y: yes thanks for the suggestion, I've updated these attributes to be readonly.

fhibf (approved):

PR #3322, opened 2023-06-01
Bulk Delete

LTA-Thinking:
## Description
Adds the $bulk-delete operation.  `DELETE \$bulk-delete`
This operation can be called at the system level or the resource type level.
As with our other async jobs it returned a Content-Location header showing the endpoint where the job record can be retrieved from. This endpoint can be used to get the current status of the job (GET) or to cancel it (DELETE).

This job is limited by the integer max value due to reporting the ids of deleted resources. This means that jobs should be limited to impact no more than ~2 billion resources, probably less than 500 million to be safe.

Added to capabilities statement.
Added notification on job competition.
Added unit tests.
Cosmos DB is now supported.
Rest examples added

Supports the following parameters:
- _hardDelete: Flag for whether to remove the resource from the database. Default false.
- _purgeHistory: Can be used with hardDelete to just purge historical records of existing resources. Default false.
- _reportIds: Flag for whether the final report should include the ids of all the resources deleted in this job. Default false.
- \<search parms\>: Any parameters used for a normal search operation can be added.

## Testing
Manually tested the three types of bulk delete with every type of versioning policy.
New E2E tests
New unit tests

## FHIR Team Checklist

Feature


src/Microsoft.Health.Fhir.Api/Features/Routing/RouteNames.cs
```diff
@@ -70,5 +70,9 @@ internal static class RouteNames
         internal const string PostSearchParameterState = "PostSearchParameterState";
 
         internal const string UpdateSearchParameterState = "UpdateSearchParameterState";
+
+        internal const string GetBulkDeleteStatusById = "GetBulkDeleteStatusById";
```
fhibf, line 74: Is it possible to replace this literal with nameof(GetBulkDeleteStatusById)?

It would be something like:
internal const string GetBulkDeleteStatusById = nameof(GetBulkDeleteStatusById);

Just suggesting that as an option to minimize possible bugs due a small typo. 

The same applies to the constant below.

src/Microsoft.Health.Fhir.Core/Features/Conformance/SystemConformanceProvider.cs
```diff
@@ -134,6 +167,20 @@ public override async Task<ResourceElement> GetCapabilityStatementOnStartup(Canc
...
+        public bool IsBackgroundJob()
+        {
+            try
+            {
+                _urlResolver.ResolveMetadataUrl(false);
+                return false;
+            }
+            catch (Exception ex)
+            {
+                _logger.LogInformation(ex, "null arg");
```
fhibf, line 179: Should we add more information to this log?

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkDelete/BulkDeleteOrchestratorJob.cs
```diff
@@ -0,0 +1,65 @@
...
+            ISearchService searchService)
+        {
+            EnsureArg.IsNotNull(queueClient, nameof(queueClient));
+            EnsureArg.IsNotNull(searchService, nameof(searchService));
+
+            _queueClient = queueClient;
+            _searchService = searchService;
+        }
+
+        public async Task<string> ExecuteAsync(JobInfo jobInfo, IProgress<string> progress, CancellationToken cancellationToken)
```
fhibf, line 37: Should we add logging mentioning the state of the job? 
I believe it'll help us debugging problems in prod.

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkDelete/BulkDeleteProcessingJob.cs
```diff
@@ -0,0 +1,78 @@
...
+    public class BulkDeleteProcessingJob : IJob
+    {
+        private readonly Func<IScoped<IDeleter>> _deleterFactory;
+        private readonly RequestContextAccessor<IFhirRequestContext> _contextAccessor;
+
+        public BulkDeleteProcessingJob(
+            Func<IScoped<IDeleter>> deleterFactory,
+            RequestContextAccessor<IFhirRequestContext> contextAccessor)
+        {
+            _deleterFactory = deleterFactory;
```
fhibf, line 34: Should we check null values as this is a public class?

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkDelete/BulkDeleteProcessingJob.cs
```diff
@@ -0,0 +1,78 @@
...
+
+                _contextAccessor.RequestContext = fhirRequestContext;
+
+                using IScoped<IDeleter> deleter = _deleterFactory();
+                var itemsDeleted = await deleter.Value.DeleteMultipleAsync(new ConditionalDeleteResourceRequest(definition.Type, (IReadOnlyList<Tuple<string, string>>)definition.SearchParameters, definition.DeleteOperation, -1), cancellationToken);
+                var result = new BulkDeleteResult();
+                result.ResourcesDeleted.Add(definition.Type, itemsDeleted);
+                return JsonConvert.SerializeObject(result);
+            }
+            catch (Exception)
```
fhibf, line 68: Probably we can replace this try/catch/finally with a simpler catch/finally as the catch block is quite simple.

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkDelete/Mediator/CancelBulkDeleteHandler.cs
```diff
@@ -0,0 +1,83 @@
...
+using Microsoft.Health.Fhir.Core.Features.Security;
+using Microsoft.Health.Fhir.Core.Models;
+using Microsoft.Health.JobManagement;
+using Newtonsoft.Json;
+
+namespace Microsoft.Health.Fhir.Core.Features.Operations.BulkDelete.Mediator
+{
+    public class CancelBulkDeleteHandler : IRequestHandler<CancelBulkDeleteRequest, CancelBulkDeleteResponse>
+    {
+        private IAuthorizationService<DataActions> _authorizationService;
```
fhibf, line 28: I believe these two attributes can be set as readonly.

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkDelete/Mediator/CreateBulkDeleteHandler.cs
```diff
@@ -0,0 +1,80 @@
...
+using Microsoft.Health.Fhir.Core.Messages.Delete;
+using Microsoft.Health.Fhir.Core.Models;
+using Microsoft.Health.JobManagement;
+using Newtonsoft.Json;
+
+namespace Microsoft.Health.Fhir.Core.Features.Operations.BulkDelete.Mediator
+{
+    public class CreateBulkDeleteHandler : IRequestHandler<CreateBulkDeleteRequest, CreateBulkDeleteResponse>
+    {
+        private IAuthorizationService<DataActions> _authorizationService;
```
fhibf, line 30: I think these private attributes can be set as readonly.

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkDelete/Mediator/CreateBulkDeleteHandler.cs
```diff
@@ -0,0 +1,80 @@
...
+        private RequestContextAccessor<IFhirRequestContext> _contextAccessor;
+        private ISearchService _searchService;
+
+        public CreateBulkDeleteHandler(
+            IAuthorizationService<DataActions> authorizationService,
+            IQueueClient queueClient,
+            RequestContextAccessor<IFhirRequestContext> contextAccessor,
+            ISearchService searchService)
+        {
+            _authorizationService = authorizationService;
```
fhibf, line 41: Probably we should add validations to avoid null references.

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkDelete/Mediator/CreateBulkDeleteHandler.cs
```diff
@@ -0,0 +1,80 @@
...
+            _queueClient = queueClient;
+            _contextAccessor = contextAccessor;
+            _searchService = searchService;
+        }
+
+        public async Task<CreateBulkDeleteResponse> Handle(CreateBulkDeleteRequest request, CancellationToken cancellationToken)
```
fhibf, line 47: Probably we should add a check to ensure request is not null.

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkDelete/Mediator/CreateBulkDeleteHandler.cs
```diff
@@ -0,0 +1,80 @@
...
+            }
+
+            var definitions = new List<string>();
+            var processingDefinition = new BulkDeleteDefinition(JobType.BulkDeleteOrchestrator, request.DeleteOperation, request.ResourceType, searchParameters, _contextAccessor.RequestContext.Uri.ToString(), _contextAccessor.RequestContext.BaseUri.ToString());
+            definitions.Add(JsonConvert.SerializeObject(processingDefinition));
+            var jobInfo = await _queueClient.EnqueueAsync((byte)QueueType.BulkDelete, definitions.ToArray(), null, false, false, cancellationToken);
+
+            // Bad, fix this to be better
+            if (jobInfo.Count < 1)
+            {
```
fhibf, line 73: For this condition, should we add a more detailed comment? A work-item id? More details to the exception mentioning that there was not Job Info available? Something that would make DRI's life easier.

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkDelete/Mediator/GetBulkDeleteHandler.cs
```diff
@@ -0,0 +1,113 @@
...
+using Microsoft.Health.Fhir.Core.Features.Security;
+using Microsoft.Health.Fhir.Core.Models;
+using Microsoft.Health.JobManagement;
+using Newtonsoft.Json;
+
+namespace Microsoft.Health.Fhir.Core.Features.Operations.BulkDelete.Mediator
+{
+    public class GetBulkDeleteHandler : IRequestHandler<GetBulkDeleteRequest, GetBulkDeleteResponse>
+    {
+        private IAuthorizationService<DataActions> _authorizationService;
```
fhibf, line 25: Should we set these private attributes as readonly?

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkDelete/Mediator/GetBulkDeleteHandler.cs
```diff
@@ -0,0 +1,113 @@
...
+        private IQueueClient _queueClient;
+
+        public GetBulkDeleteHandler(
+            IAuthorizationService<DataActions> authorizationService,
+            IQueueClient queueClient)
+        {
+            _authorizationService = authorizationService;
```
fhibf, line 32: Check for null references?

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkDelete/Mediator/GetBulkDeleteHandler.cs
```diff
@@ -0,0 +1,113 @@
...
+            var failureResultCode = HttpStatusCode.OK;
+
+            foreach (var job in jobs)
+            {
+                if (job.Status == JobStatus.Failed)
+                {
+                    failed = true;
+                    succeeded = false;
+
+                    // TODO: fix exception logging
```
fhibf, line 58: Should we fix the exception logging in this PR? Do you want to improve the "detailsText" with something more specific?

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Delete/DeleteResourceHandler.cs
```diff
@@ -20,14 +21,18 @@ namespace Microsoft.Health.Fhir.Core.Features.Resources.Delete
 {
     public class DeleteResourceHandler : BaseResourceHandler, IRequestHandler<DeleteResourceRequest, DeleteResourceResponse>
     {
+        private IDeleter _deleter;
```
fhibf, line 24: Should we set this attribute as readonly?

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Delete/Deleter.cs
```diff
@@ -0,0 +1,113 @@
...
+            int itemsDeleted = 0;
+            bool deleteAll = request.MaxDeleteCount < 0;
+
+            // Delete the matched results...
+            while (matchedResults.Any() || !string.IsNullOrEmpty(ct))
+            {
+                var resultsToDelete = deleteAll ? matchedResults : matchedResults.Take(request.MaxDeleteCount - itemsDeleted);
+                foreach (SearchResultEntry item in resultsToDelete)
+                {
+                    var result = await DeleteAsync(new DeleteResourceRequest(request.ResourceType, item.Resource.ResourceId, request.DeleteOperation), cancellationToken);
```
fhibf, line 91: I believe that in this case, if these are all soft deletes, we can use MergeAsync method and delete all resources in a single database operation. That would be great in terms of performance.

fhibf (commented):
Few suggestions provided.

src/Microsoft.Health.Fhir.Api/Features/ActionResults/JobResult.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.Collections.Generic;
+using System.Net;
+using EnsureThat;
+using Hl7.Fhir.Model;
+using Microsoft.Health.Fhir.Core.Extensions;
+using Microsoft.Health.Fhir.Core.Models;
+
+namespace Microsoft.Health.Fhir.Api.Features.ActionResults
+{
+    public class JobResult : ResourceActionResult<Parameters>
```
brendankowitz, line 16: Does Parameters go back to the user? Should we consider using (Bundle.Entry.Response) from Batch-Response bundle type from https://www.hl7.org/fhir/async-bundle.html#3.2.6.2.4.0.3?

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Delete/Deleter.cs
```diff
@@ -0,0 +1,113 @@
...
+                    itemsDeleted++;
+                }
+
+                if (!string.IsNullOrEmpty(ct) && (request.MaxDeleteCount - itemsDeleted > 0 || deleteAll))
+                {
+                    (matchedResults, ct) = await _searchService.ConditionalSearchAsync(
+                        request.ResourceType,
+                        request.ConditionalParameters,
+                        deleteAll ? -1 : request.MaxDeleteCount - itemsDeleted,
```
brendankowitz, line 100: Should we convert this to an optional parameter, since we are making it _optional_

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkDelete/BulkDeleteOrchestratorJob.cs
LTA-Thinking: Good idea. I'm adding it to JobHosting so it is generic for all job types.

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkDelete/Mediator/CreateBulkDeleteHandler.cs
LTA-Thinking: Fixed the code and removed comment.

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkDelete/Mediator/GetBulkDeleteHandler.cs
LTA-Thinking: Yes, this was a note to remind me to fix this in this PR.

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Delete/Deleter.cs
LTA-Thinking: None of the places it is used would benefit from a default value. All the places that use it need to specify the value.

src/Microsoft.Health.Fhir.Api/Features/ActionResults/JobResult.cs
LTA-Thinking: A Batch-Response could be used to wrap the Parameters object, but the fields in Bundle.Entry.Response wouldn't allow me to send back all the relevant data.
I was trying to base this on our response for a reindex job.

src/Microsoft.Health.Fhir.Core/Features/Conformance/SystemConformanceProvider.cs
LTA-Thinking: Changed

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkDelete/Mediator/GetBulkDeleteHandler.cs
LTA-Thinking: I'm fixing the error message to be something usable, but the error handing of async jobs needs to be reworked. Right now there is no way to deserialize the objects stored when error happen because the formatting isn't standardized. This will be a larger effort than I want to do as part of this story as it will involve reworking error handling for Import and Export as well.
I made a new story to track this work: [User Story 104493](https://microsofthealth.visualstudio.com/Health/_workitems/edit/104493): Rework async job error handling

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -584,15 +584,16 @@ private string GetJsonValue(string json, string propName, bool isExisting)
             var startIndex = json.IndexOf($"\"{propName}\":\"", StringComparison.Ordinal);
             if (startIndex == -1)
             {
-                _logger.LogError($"Cannot parse {propName} value from {(isExisting ? "existing" : "input")} {json}");
+                // I think this should be a warning because it happens every time a resource is deleted. Maybe even info.
```
LTA-Thinking, line 587: Is this an agreeable change? If so I'll remove the comment.

LTA-Thinking:
I need to see how this interacts with SMART roles.

LTA-Thinking:
> I need to see how this interacts with SMART roles.

I think this is fine. SMART can't assign any Delete roles, so there shouldn't be an issue.

src/Microsoft.Health.Fhir.Api/Features/Routing/RouteNames.cs
LTA-Thinking: Updated the entire file.

docs/rest/BulkDelete.http
```diff
@@ -0,0 +1,63 @@
...
+
+< ./Data/ExportDataBatch.json
+
+### Record Bulk Delete content location
+@bulkDeleteLocation = {{bulkDelete.response.headers.Content-Location}}
+
+### Create Bulk Delete request
+# @name bulkDelete
+DELETE https://{{hostname}}/$bulk-delete
+Authorization: Bearer {{bearer.response.body.access_token}}
```
brendankowitz, line 35: Should these require the Prefer: async header?

src/Microsoft.Health.Fhir.Api/Features/ActionResults/JobResult.cs
brendankowitz: Makes sense, though reindex is the one we invented and not following spec guidelines :)

src/Microsoft.Health.Fhir.Api/Features/Routing/RouteNames.cs
```diff
@@ -7,68 +7,74 @@ namespace Microsoft.Health.Fhir.Api.Features.Routing
 {
     internal static class RouteNames
     {
-        internal const string Metadata = "Metadata";
+        internal const string Metadata = nameof(Metadata);
```
brendankowitz, line 10: Trying to reason if this is a change for the better, it makes sense that it can be the same name as the property, though the compile-time output will be the same. The advantage of `nameof` is that is responds to refactoring within the IDE tools, but in this case doesn't seem desired

src/Microsoft.Health.Fhir.Core/Extensions/BulkDeleteMediatorExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using EnsureThat;
+using MediatR;
+using Microsoft.Health.Fhir.Core.Features.Operations.BulkDelete.Mediator;
+using Microsoft.Health.Fhir.Core.Messages.Delete;
+
+namespace Microsoft.Health.Fhir.Core.Extensions
```
brendankowitz, line 15: Love the idea of extensions per feature, should they also be kept in the feature folder?

src/Microsoft.Health.Fhir.Core/Extensions/CommonModelExtensions.cs
```diff
@@ -0,0 +1,50 @@
...
+using System.Linq;
+using EnsureThat;
+using Hl7.Fhir.Model;
+using Microsoft.Health.Fhir.Core.Models;
+
+namespace Microsoft.Health.Fhir.Core.Extensions
+{
+    public static class CommonModelExtensions
+    {
+        public static OperationOutcome.IssueComponent ToPoco(this OperationOutcomeIssue issue)
```
brendankowitz, line 17: Did this already exist somewhere before?

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkDelete/BulkDeleteOrchestratorJob.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            EnsureArg.IsNotNull(jobInfo, nameof(jobInfo));
+            EnsureArg.IsNotNull(progress, nameof(progress));
+
+            BulkDeleteDefinition definition = JsonConvert.DeserializeObject<BulkDeleteDefinition>(jobInfo.Definition);
```
brendankowitz, line 42: Since this is so common, I was thinking we should have extensions for this like:
```csharp
var request = jobInfo.DeserializeDefinition<BulkDeleteDefinition>();
var jobResult = jobInfo.DeserializeResult<BulkDeleteResult>();
```

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkDelete/BulkDeleteOrchestratorJob.cs
```diff
@@ -0,0 +1,65 @@
...
+                    definitions.Add(JsonConvert.SerializeObject(processingDefinition));
+                }
+            }
+            else
+            {
+                var processingDefinition = new BulkDeleteDefinition(JobType.BulkDeleteProcessing, definition.DeleteOperation, definition.Type, definition.SearchParameters, definition.Url, definition.BaseUrl);
+                definitions.Add(JsonConvert.SerializeObject(processingDefinition));
+            }
+
+            await _queueClient.EnqueueAsync((byte)QueueType.BulkDelete, definitions.ToArray(), jobInfo.GroupId, false, false, cancellationToken);
```
brendankowitz, line 61: I think a simple extension can improve the syntax here to:
```csharp
await _queueClient.EnqueueAsync(QueueType.BulkDelete, groupId: jobInfo.GroupId, cancellationToken: cancellationToken, definitions: definitions);
```
See: https://github.com/microsoft/fhir-server/blob/fbce445fd1d0689d835b43f85e1da1060e87626b/src/Microsoft.Health.Fhir.Core/Features/Operations/QueueClientExtensions.cs

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkDelete/BulkDeleteProcessingJob.cs
```diff
@@ -0,0 +1,83 @@
...
+                    correlationId: jobInfo.Id.ToString(),
+                    requestHeaders: new Dictionary<string, StringValues>(),
+                    responseHeaders: new Dictionary<string, StringValues>())
+                {
+                    IsBackgroundTask = true,
+                };
+
+                _contextAccessor.RequestContext = fhirRequestContext;
+
+                using IScoped<IDeleter> deleter = _deleterFactory();
```
brendankowitz, line 67: nit: old school habits, but prefer `_deleterFactory.Invoke();`

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
```diff
@@ -87,9 +89,25 @@ internal class FhirCosmosSearchService : SearchService
...
         }
 
-        public override Task<IReadOnlyList<(short ResourceTypeId, string Name)>> GetUsedResourceTypes(CancellationToken cancellationToken)
+        public override async Task<IReadOnlyList<string>> GetUsedResourceTypes(CancellationToken cancellationToken)
         {
-            throw new NotImplementedException();
+            QueryDefinition sqlQuerySpec = new QueryDefinition(@"SELECT DISTINCT VALUE r.resourceTypeName
+                FROM root r
+                WHERE r.isSystem = false
+                ORDER BY r.resourceTypeName");
```
brendankowitz, line 97: Do we know how this will perform on large databases? Something I'd like to find out

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Create/CreateResourceHandler.cs
```diff
@@ -59,7 +59,7 @@ public async Task<UpsertResourceResponse> Handle(CreateResourceRequest request,
 
             await _referenceResolver.ResolveReferencesAsync(resource, _referenceIdDictionary, resource.TypeName, cancellationToken);
 
-            ResourceWrapper resourceWrapper = CreateResourceWrapper(resource, deleted: false, keepMeta: true);
+            ResourceWrapper resourceWrapper = ResourceWrapperFactory.CreateResourceWrapper(resource, ResourceIdProvider, deleted: false, keepMeta: true);
```
brendankowitz, line 62: This looks like it should be an extension method

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Delete/ConditionalDeleteResourceHandler.cs
```diff
@@ -26,6 +26,7 @@ public class ConditionalDeleteResourceHandler : BaseResourceHandler, IRequestHan
     {
         private readonly ISearchService _searchService;
         private readonly IMediator _mediator;
+        private readonly IDeleter _deleter;
```
brendankowitz, line 29: `IDeletionService`? Still thinking on this one also 😄

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Delete/Deleter.cs
```diff
@@ -0,0 +1,134 @@
...
+
+            return new ResourceKey(key.ResourceType, key.Id, version);
+        }
+
+        public async Task<int> DeleteMultipleAsync(ConditionalDeleteResourceRequest request, CancellationToken cancellationToken)
+        {
+            (IReadOnlyCollection<SearchResultEntry> matchedResults, string ct) = await _searchService.ConditionalSearchAsync(request.ResourceType, request.ConditionalParameters, request.MaxDeleteCount, cancellationToken);
+
+            int itemsDeleted = 0;
+            bool deleteAll = request.MaxDeleteCount < 0;
```
brendankowitz, line 84: This still seems scary, if by some mistake a negative is passed we delete everything

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Delete/Deleter.cs
```diff
@@ -0,0 +1,134 @@
...
+                            return new ResourceWrapperOperation(deletedWrapper, true, keepHistory, null, false, false, bundleOperationId: null);
+                        }).ToList(),
+                        cancellationToken);
+                    itemsDeleted = resultsToDelete.Count();
+                }
+                else
+                {
+                    foreach (SearchResultEntry item in resultsToDelete)
+                    {
+                        await DeleteAsync(new DeleteResourceRequest(request.ResourceType, item.Resource.ResourceId, request.DeleteOperation), cancellationToken);
```
brendankowitz, line 111: Should this be in parallel to some degree?

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Delete/Deleter.cs
```diff
@@ -0,0 +1,134 @@
...
+                    await _fhirDataStore.HardDeleteAsync(key, request.DeleteOperation == DeleteOperation.PurgeHistory, cancellationToken);
+                    break;
+                default:
+                    throw new ArgumentOutOfRangeException(nameof(request));
+            }
+
+            return new ResourceKey(key.ResourceType, key.Id, version);
+        }
+
+        public async Task<int> DeleteMultipleAsync(ConditionalDeleteResourceRequest request, CancellationToken cancellationToken)
```
brendankowitz, line 79: Thought: If we had `IAsyncEnumerable<ResourceKey>` we could pass them out as we delete them

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
mikaelweave: I think if you remove the `ORDER BY` this will perform fine.

I'm testing (*adopting* 😄 ) this function for parallel Gen 1 export. I have a 500GB collection with 50K RU. The orig query above appears to be scanned record by record and the query times out for me for subsequent results after the first.

Removing the `ORDER BY` results in near instant execution. I can get more precise info if needed.

src/Microsoft.Health.Fhir.Core/Extensions/BulkDeleteMediatorExtensions.cs
LTA-Thinking: That makes sense. Keep all the files for the feature in one place.

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkDelete/BulkDeleteOrchestratorJob.cs
LTA-Thinking: It is hard to find all the paterns like this, but I think I got most of them replaced.

docs/rest/BulkDelete.http
LTA-Thinking: I'll add this. Do we know why the Reindex job requires the respond-sync value for Prefer? ValidateReindexRequestFilterAttribute

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
LTA-Thinking: Mikael and I talked about this. We have the issue where the request will fail to return a continuation token if we don't include an `ORDER BY` with a `DISTINCT`. I'm still looking into solutions.

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Create/CreateResourceHandler.cs
LTA-Thinking: It is, it is part of ResourceWrapperFactoryExtensions.

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Delete/Deleter.cs
LTA-Thinking: We could, but what benefit would that do us? We don't want to save the ids because it could be a large ammount of data.

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Delete/Deleter.cs
LTA-Thinking: Changed it to be a nullable int.

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Operations/OperationsCapabilityProviderTests.cs
```diff
@@ -42,28 +42,19 @@ public OperationsCapabilityProviderTests()
...
-                _coreFeatureConfiguration.SupportsSelectableSearchParameters = false;
-                ListedCapabilityStatement listedCapabilityStatement = new ListedCapabilityStatement();
-                OperationsCapabilityProvider provider = new OperationsCapabilityProvider(_operationsOptions, _featureOptions, _coreFeatureOptions, _urlResolver);
+                var provider = new OperationsCapabilityProvider(_operationsOptions, _featureOptions, _coreFeatureOptions, _urlResolver);
                 ICapabilityStatementBuilder builder = Substitute.For<ICapabilityStatementBuilder>();
                 provider.Build(builder);
 
-                builder.Received(2).Apply(Arg.Any<Action<ListedCapabilityStatement>>());
+                builder.Received(added ? 1 : 0)
+                    .Apply(Arg.Is<Action<ListedCapabilityStatement>>(x => x.Method.Name == nameof(OperationsCapabilityProvider.AddSelectableSearchParameterDetails)));
```
brendankowitz, line 57: @PTaladay - In case you're also fixing this, I think this should be more reliable for a specific operation

src/Microsoft.Health.Fhir.Core/Configs/CoreFeatureConfiguration.cs
```diff
@@ -76,5 +76,10 @@ public class CoreFeatureConfiguration
         /// Gets or sets a value indicating whether the server supports the $status operation for SearchParameters.
         /// </summary>
         public bool SupportsSelectableSearchParameters { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether the server supports the $bulk-delete.
+        /// </summary>
+        public bool SupportsBulkDelete { get; set; }
```
LTA-Thinking, line 83: Why do we want this configuration parameter when we have the one in the job host configuration?

brendankowitz: This blocks the API call and excludes it from the CapabilityStatement. Looking forward I think it still might be easier to delete this setting and clean up the checks after its enabled.

brendankowitz (approved):
:shipit:

PR #3337, opened 2023-06-07
Add keep last updated parameter to merge

SergeyGaluzo:
Adding parameter to merge to regenerate last updated without overlaps if values are not provided on input.


LTA-Thinking:
Can tests be added to show that _lastUpdated is being correctly persisted or generated.

LTA-Thinking (approved):

PR #3355, opened 2023-06-15
adding New Feature without the hyphen character for valid check

v-mserdar:
## Description
To handle when a label that is "New Feature" is added that it passes the metadata validator.

## FHIR Team Checklist

### DIFF
--- a/.github/workflows/validate-prs.yml
+++ b/.github/workflows/validate-prs.yml
@@ -44,7 +44,8 @@ jobs:
                labels.includes("Bug") == false &&
                labels.includes("Dependencies") == false && 
                labels.includes("Enhancement") == false &&
-               labels.includes("New-Feature") == false) {
+               labels.includes("New Feature") == false &&
+               labels.includes("New-Feature") == false)) {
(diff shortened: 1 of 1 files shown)


LTA-Thinking (dismissed):

LTA-Thinking:
I feel we don't really need both. Would it make sense to just remove the one with the hyphen? The one without the hyphen has a description and looks better to a reader.

v-mserdar:
> I feel we don't really need both. Would it make sense to just remove the one with the hyphen? The one without the hyphen has a description and looks better to a reader.

Can we just remove it without causing issues? What happens to ones that have that label attached?
If there's no issue with my questions, is it OK for me to just remove it from this page here?
https://github.com/microsoft/fhir-server/labels?page=2&sort=name-asc

LTA-Thinking:
The label with the hyphen is only a day old, I'm the only one using it I think. I made it because my bulk delete PR wouldn't pass the check. We don't use the new feature tag much so I don't think it has come up before. I find it funny both of us ran into the same issue within a day of each other. Thanks for taking the better approach to fixing it :)

PR #3360, opened 2023-06-17
SQL without transaction next

SergeyGaluzo:
1. retriable merge with no tran
2. cosmetic in sql client
3. self retriable calls in store
4. transaction watchdog
5. TransactionId added to the Resource table.

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetryService.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetryService.cs
@@ -5,6 +5,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Data;
 using System.Threading;
 using System.Threading.Tasks;
 using EnsureThat;
@@ -245,7 +246,7 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Storage
                     await action(sqlCommand, cancellationToken);
                     if (retry > 0)
                     {
-                        await TryLogToDatabase($"ExecuteSql.{sqlCommand.CommandText}", "Warn", $"retries={retry} error={lastException}", start, cancellationToken);
+                        await TryLogToDatabase($"SqlRetryService:{sqlCommand.CommandText}", "Warn", $"retries={retry} error={lastException}", start, cancellationToken);
                     }
 
                     return;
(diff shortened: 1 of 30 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetryService.cs
```diff
@@ -334,5 +346,31 @@ public async Task ExecuteSql<TLogger>(SqlCommand sqlCommand, Func<SqlCommand, Ca
...
+                using var conn = await _sqlConnectionBuilder.GetSqlConnectionAsync(initialCatalog: null, cancellationToken: cancellationToken).ConfigureAwait(false);
+                conn.RetryLogicProvider = null;
+                await conn.OpenAsync(cancellationToken);
+                cmd.Connection = conn;
+
+                await cmd.ExecuteNonQueryAsync(cancellationToken);
+            }
+            catch
+            {
+                // do nothing;
```
fhibf, line 372: Does it make sense to, in case of failure, log the error using our regular logging framework?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetryServiceOptions.cs
```diff
@@ -24,7 +24,7 @@ public class SqlRetryServiceOptions
         public const string SqlServer = "SqlServer";
 
         /// <summary>Maximum number of (re)try attempts on a retriable error.</summary>
-        public int MaxRetries { get; set; } = 5;
+        public int MaxRetries { get; set; } = 100;
```
fhibf, line 27: Isn't 100 to high?
If an operation is failing more than 10 times in a row, I believe there is something more critical happening. 
My only concern is that we can hide a more serious and complex problem setting max retries to 100.

fhibf (approved):

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetryServiceOptions.cs
SergeyGaluzo: These are truly retriable exceptions. Consider backend process. It should never fail because of retriable exception. Hence high number. This is different from retrying blindly.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetryService.cs
SergeyGaluzo: I don't think so. When we do Log.Whatever we do our best effort, and do not try to log if log itself fails.

PR #3365, opened 2023-06-20
Migrate Reindex Job to QueueClient

PTaladay:
## Description
Moves ReindexJobTask from a background worker service to now use the queue client for handling parallelization of jobs. Removed/cleaned up components that either never worked or are no longer necessary such as the throttling controller, reindexJobWorker, etc. 

## Related issues
Addresses [issue AB#102362, AB#103503, AB#102363].

## Testing
Existing unit tests update, running integration tests and manually running reindex and validating data after run to ensure logic is still correct. 

## FHIR Team Checklist


src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlQueueClient.cs
```diff
@@ -221,6 +221,31 @@ public async Task<IReadOnlyList<JobInfo>> GetJobsByIdsAsync(byte queueType, long
...
+                sqlCommandWrapper.CommandText = "SELECT TOP 1 JobId FROM [dbo].[JobQueue] WITH (NOLOCK) WHERE QueueType = @QueueType AND Status in (0, 1) ORDER BY JobId DESC";
+                sqlCommandWrapper.Parameters.AddWithValue("@QueueType", queueType);
+                await using SqlDataReader sqlDataReader = await sqlCommandWrapper.ExecuteReaderAsync(cancellationToken);
+
+                var result = await sqlDataReader.ReadJobInfoAsync(cancellationToken);
+                return result?.Id.ToString();
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex, "PeekAsync failed.");
```
v-mserdar, line 239: Do we want to differentiate between LogError for "real" errors and LogWarning for retriable errors?

v-mserdar:
When I run this locally and do a GET $reindex/{id} I don't see any progress which is the current functionality. For long running reindex jobs users won't know the reindex job is making any progress. 
For example, the current reindex updates the following fields on regular intervals:
lastModified
totalResourcesToReindex
resourcesSuccessfullyReindexed
progress
resources with counts

In sql, I see that the definition saved for my Patient resources that I reindexed doesn't save the "CountReindexed" - it remains at 0. Also the *ResourceSurrogateId fields in that same json don't show any progress
See the documentation for the full list of possible fields https://learn.microsoft.com/en-us/azure/healthcare-apis/fhir/how-to-run-a-reindex#how-to-check-the-status-of-a-reindex-job

I also discovered that if you do a GET $reindex/{id} where the id is valid but it's not the id based on the returned Content-Location that it returns an error within the 400 response. See the example out here. the id I passed was part of the same group but wasn't the initial entry - see second image below for sql screenshot. I used id=5 succesfully but id=7 gives the error returned. You can tell the json is different just by the small view of the definition column.
image

image

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlQueueClient.cs
PTaladay: Thanks. I updated logs to be more inline with actual errors vs warnings and informational.

src/Microsoft.Health.TaskManagement.UnitTests/TestQueueClient.cs
```diff
@@ -226,5 +226,10 @@ public Task<bool> PutJobHeartbeatAsync(JobInfo jobInfo, CancellationToken cancel
 
             return Task.FromResult(cancel);
         }
+
+        public Task<string> PeekAsync(byte queueType, CancellationToken cancellationToken)
+        {
+            return Task.FromResult(jobInfos.Where(j => j.QueueType == queueType && (j.Status == JobStatus.Created || j.Status == JobStatus.Running)).ToList().FirstOrDefault().Id.ToString());
```
v-mserdar, line 232: This throws when the jobInfos.Count == 0. You'll need to revise as:
FirstOrDefault()?.Id.ToString()

PTaladay: Thanks, I updated it.

PR #3381, opened 2023-07-06
Added Status and $Status operation to CapabilityStatement

PTaladay:
## Description
Updated capability statement to display search parameter status in the documentation field. $status operation is also displayed only when feature flag is enabled. 

I used the documentation field because the firely sdk doesn't allow extending on the search param component or any component that relates to the search parameter.

## Related issues
Addresses [issue AB#10442].

## Testing
Ran through running the metadata endpoint with the different configuration options. 

## FHIR Team Checklist


src/Microsoft.Health.Fhir.Shared.Api/Controllers/OperationDefinitionController.cs
```diff
@@ -33,19 +33,23 @@ public class OperationDefinitionController : Controller
...
         private readonly OperationsConfiguration _operationConfiguration;
         private readonly FeatureConfiguration _featureConfiguration;
+        private readonly CoreFeatureConfiguration _coreFeatureConfiguration;
 
         public OperationDefinitionController(
             IMediator mediator,
             IOptions<OperationsConfiguration> operationsConfig,
-            IOptions<FeatureConfiguration> featureConfig)
+            IOptions<FeatureConfiguration> featureConfig,
+            IOptions<CoreFeatureConfiguration> coreFeatureConfig)
```
abiisnn, line 42: What's the difference between `featureConfig` and `coreFeatureConfig`?

abiisnn (commented):
🍒 

Do you think you can add a test to make sure it's called correctly?

src/Microsoft.Health.Fhir.Shared.Api/Controllers/OperationDefinitionController.cs
PTaladay: I think the original intent was to separate fhir spec features vs fhir server features? Either way I needed to add the coreFeatureConfiguration to check the selectablesearchparameter feature flag.

src/Microsoft.Health.Fhir.Core/Data/OperationDefinition/search-parameter-status.json
```diff
@@ -0,0 +1,32 @@
+{
+    "resourceType": "OperationDefinition",
+    "id": "status",
+    "url": "[base]/OperationDefinition/search-parameter-status",
+    "version": "1.0.0",
+    "name": "Selectable search parameters",
+    "status": "active",
+    "kind": "operation",
+    "description": "Allows for disabling of search parameters that are not being utilized in order to save space in the datastore. This operation is asynchronous as defined in the [FHIR Asynchronous Request Pattern](http://hl7.org/fhir/async.html)",
```
rajithaalurims, line 9: nit: this allows customers to either disable or turn it back on as "supported" from disabled state correct? Can we change the description to add something about updating status to supported as well?

src/Microsoft.Health.Fhir.Api/Features/Routing/RouteNames.cs
```diff
@@ -65,6 +65,8 @@ internal static class RouteNames
 
         internal const string PurgeHistoryDefinition = "PurgeHistoryDefinition";
 
+        internal const string SearchParameterStatusOperationDefinition = "SearchParameterStatusOperationDefinition";
```
fhibf, line 68: Low priority: You can replace the static string with nameof(SearchParameterStatusOperationDefinition).

fhibf (approved):

src/Microsoft.Health.Fhir.Api/Features/Routing/RouteNames.cs
PTaladay: I can add that as a fix in another PR.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -668,9 +668,10 @@ public void Build(ICapabilityStatementBuilder builder)
         {
             EnsureArg.IsNotNull(builder, nameof(builder));
 
-            builder.PopulateDefaultResourceInteractions()
-                .SyncSearchParameters()
-                .AddGlobalSearchParameters()
+            var task = Task.Run(() => builder.PopulateDefaultResourceInteractions()
+                .SyncSearchParametersAsync(CancellationToken.None));
+            task.Wait();
```
brendankowitz, line 673: Should we be using: https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.runsynchronously?view=net-7.0 here?

PR #3382, opened 2023-07-06
Disabling check for conditional references

SergeyGaluzo:
$Import - Disabling check for conditional references in incremental mode

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Import/ImportResourceParser.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Import/ImportResourceParser.cs
@@ -72,7 +72,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Import
 
                 if (reference.Reference.Contains('?', StringComparison.Ordinal))
                 {
-                    throw new NotSupportedException("Conditional reference is not supported for initial $import.");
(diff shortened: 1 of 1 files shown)


src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Import/ImportResourceParser.cs
```diff
@@ -67,7 +72,7 @@ private static void CheckConditionalReferenceInResource(Resource resource)
 
                 if (reference.Reference.Contains('?', StringComparison.Ordinal))
                 {
-                    throw new NotSupportedException("Conditional reference is not supported for $import.");
+                    throw new NotSupportedException("Conditional reference is not supported for initial $import.");
```
mikaelweave, line 75: My only suggestion would be to use the enum name vs "initial" in this string. If more import modes are added later, this won't have to be updated.

mikaelweave (changes requested):
One small comment - rest of the PR looks good. Will approve after the comment is addressed.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Import/ImportResourceParser.cs
SergeyGaluzo: Fixed.

mikaelweave (approved):
Looks good!

PR #3398, opened 2023-07-11
Invisible history records

SergeyGaluzo:
This PR provides ability to replace current events change capturing system, currently based on writing resource keys for changed resources to user tables, by querying Resource table based on processed transactions (old events system is still not removed and works AS-IS).

To enable above we need to hold resource versions for the event retention period, so events system can query Resource table and data is not missed. This works out of the box with versioned data policy and absence of hard deletes. 

To cover no-version policy and hard deletes PR code delays physical deletes. It marks historical resource version as "invisible", i.e. sets raw resource to 0xF. As far as data size, setting raw resource to single byte is equivalent to retaining only resource keys in the old events tables, but without actual writes (other name of this feature is "events with no writes"). Note that these updates are scoped only to no-version & events enabled case. This magic 0xF value allows to filter out these resources from search (hence "invisible" name).

To finally remove these invisible historical resource keys, a new watchdog "InvisibleHistoryCleanupWatchdog" was added. 

Main logic to set raw resource to 0xF is in the couple of lines added to MergeResources and HardDeleteResource stored procedures.


feordin:
@SergeyGaluzo Just a general question Sergey, is there any problem with this change and the constraint we have on the Resource table to detect empty resources?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/HardDeleteResource.sql
```diff
@@ -0,0 +1,71 @@
...
+    DELETE dbo.TokenTokenCompositeSearchParam WHERE ResourceTypeId = @ResourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @SurrogateIds)
+    DELETE dbo.TokenDateTimeCompositeSearchParam WHERE ResourceTypeId = @ResourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @SurrogateIds)
+    DELETE dbo.TokenQuantityCompositeSearchParam WHERE ResourceTypeId = @ResourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @SurrogateIds)
+    DELETE dbo.TokenStringCompositeSearchParam WHERE ResourceTypeId = @ResourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @SurrogateIds)
+    DELETE dbo.TokenNumberNumberCompositeSearchParam WHERE ResourceTypeId = @ResourceTypeId AND ResourceSurrogateId IN (SELECT ResourceSurrogateId FROM @SurrogateIds)
+  END
+  
+  IF @@trancount > 0 COMMIT TRANSACTION
+
+  IF @IsResourceChangeCaptureEnabled = 1 EXECUTE dbo.MergeResourcesCommitTransaction @TransactionId
```
feordin, line 64: If this fails will it go to the catch and try to roll back a transaction that has already been comitted?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Tables/Resource.sql
```diff
@@ -19,6 +20,7 @@
 ALTER TABLE dbo.Resource SET ( LOCK_ESCALATION = AUTO )
 
 CREATE INDEX IX_ResourceTypeId_TransactionId ON dbo.Resource (ResourceTypeId, TransactionId) WHERE TransactionId IS NOT NULL ON PartitionScheme_ResourceTypeId (ResourceTypeId)
+CREATE INDEX IX_ResourceTypeId_HistoryTransactionId ON dbo.Resource (ResourceTypeId, HistoryTransactionId) WHERE HistoryTransactionId IS NOT NULL ON PartitionScheme_ResourceTypeId (ResourceTypeId)
```
feordin, line 23: Will this be a long running index creation for existing customers?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/HardDeleteResource.sql
SergeyGaluzo: Yes, it will go to catch. No, it will not try to rollback as catch has @@trancount check.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Tables/Resource.sql
SergeyGaluzo: It will be long running, but indexes are created with online option, so there are no affect to customers.

SergeyGaluzo:
No. Resource is technically not empty (one byte 0xF), and it passes constraint check.

---
In reply to: [1634916506](https://github.com/microsoft/fhir-server/pull/3398#issuecomment-1634916506) [](http://example.com/codeflow?ancestors=1634916506,1634916506)

feordin (approved):

PR #3400, opened 2023-07-14
Fixed bug with force reindex and when no work to be done.

PTaladay:
## Description
Force Reindex was not setting the search parameter status for search parameters listed in the TargetSearchParameterTypes.

In the case where there is no work to be done we should be updating the search parameters to their applicable final state.

## Related issues
Addresses [issue AB#105357].

## Testing
Added unit tests and manual testing

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
@@ -190,7 +190,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Reindex
             // Find search parameters not in a final state such as supported, pendingDelete, pendingDisable.
             List<SearchParameterStatus> validStatus = new List<SearchParameterStatus>() { SearchParameterStatus.Supported, SearchParameterStatus.PendingDelete, SearchParameterStatus.PendingDisable };
             var searchParamStatusCollection = await _searchParameterStatusManager.GetAllSearchParameterStatus(cancellationToken);
-            var possibleNotYetIndexedParams = _searchParameterDefinitionManager.AllSearchParameters.Where(sp => validStatus.Contains(searchParamStatusCollection.First(p => p.Uri == sp.Url).Status) || sp.IsSearchable == false || sp.SortStatus == SortParameterStatus.Supported);
+            var possibleNotYetIndexedParams = _searchParameterDefinitionManager.AllSearchParameters.Where(sp => validStatus.Contains(searchParamStatusCollection.First(p => p.Uri == sp.Url).Status) || sp.IsSearchable == true || sp.SortStatus == SortParameterStatus.Supported);
             var notYetIndexedParams = new List<SearchParameterInfo>();
 
             var resourceList = new HashSet<string>();
--- a/src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
@@ -218,6 +218,12 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Reindex
             else if (_reindexJobRecord.ForceReindex)
             {
                 resourceList.UnionWith(_reindexJobRecord.SearchParameterResourceTypes);
+
+                // Adding these in so they get included in search param status updates.
+                foreach (var searchParam in _reindexJobRecord.TargetSearchParameterTypes)
+                {
+                    _reindexJobRecord.SearchParams.Add(searchParam);
+                }
             }
             else
             {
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -232,6 +238,12 @@ private async Task<bool> TryPopulateNewJobFields(CancellationToken cancellationT
                 }
             }
 
+            // save the list of search parameters to the reindexjob document
+            foreach (var searchParams in notYetIndexedParams.Select(p => p.Url.OriginalString))
+            {
```
v-mserdar, line 243: We should probably do a sanity check here.

                if (_reindexJobRecord.SearchParams.Contains(searchParams))
                {
                    continue;
                }

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -190,7 +190,7 @@ private async Task<bool> TryPopulateNewJobFields(CancellationToken cancellationT
             // Find search parameters not in a final state such as supported, pendingDelete, pendingDisable.
             List<SearchParameterStatus> validStatus = new List<SearchParameterStatus>() { SearchParameterStatus.Supported, SearchParameterStatus.PendingDelete, SearchParameterStatus.PendingDisable };
             var searchParamStatusCollection = await _searchParameterStatusManager.GetAllSearchParameterStatus(cancellationToken);
-            var possibleNotYetIndexedParams = _searchParameterDefinitionManager.AllSearchParameters.Where(sp => validStatus.Contains(searchParamStatusCollection.First(p => p.Uri == sp.Url).Status) || sp.IsSearchable == false || sp.SortStatus == SortParameterStatus.Supported);
+            var possibleNotYetIndexedParams = _searchParameterDefinitionManager.AllSearchParameters.Where(sp => validStatus.Contains(searchParamStatusCollection.First(p => p.Uri == sp.Url).Status) || sp.IsSearchable == true || sp.SortStatus == SortParameterStatus.Supported);
```
feordin, line 193: I am not sure if the change to sp.IsSearchable == true is correct.  It seems that now the overall logic is:

Where:
1) any status is one of the three validStatus (supported, pendingDelete or pendingDisbable
or 2) SortStatus == Supported
or 3) IsSearchable == true

That would indicate that any SearchParameter with an IsSearchable of true will be included in the job.  But, we don't want to include it unless it is in one of the three states.

Perhaps the flag of IsSearchable should be included together with SortStatus?  I am not certain why this flag is part of the logic.

PTaladay: I reverted back to false as it included way more than previously.

PTaladay: It is a good point that we may no longer need it. In the event that say a customer upgrades from one FHIR schema to the next and there happens to be search parameters that are now included then they would need to be marked as supported to get reindexed, instead of unsupported.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexJobTask.cs
```diff
@@ -240,6 +258,8 @@ private async Task<bool> TryPopulateNewJobFields(CancellationToken cancellationT
                     OperationOutcomeConstants.IssueType.Informational,
                     Core.Resources.NoSearchParametersNeededToBeIndexed));
                 _reindexJobRecord.CanceledTime = Clock.UtcNow;
+
+                await UpdateParametersAndCompleteJob(cancellationToken);
```
rajithaalurims, line 262: Why are we adding this line?  It comes here only if there are NO search parameters and no resources to index.
Example: 
If there is a SP 'ABC' with status = Supported and it doesn't have any resources to reindex.
notYetIndexedParams will have 'ABC' in it.  this condition will not be true if (!notYetIndexedParams.Any() && resourceList.Count == 0).

rajithaalurims: Can you please confirm if you tested below scenario with your code changes?

- Pick a valid resource type, that doesn't have resources in your DB yet.
- Create a custom SP and refer to the above resource type as Base Resource.
- Run a full reindex.
- Test the status of the parameter.
- Repeat the same steps for a new SP but reindex targeted SP and check the status of the parameter.

PTaladay: Thanks. I removed that one.  I did re test those scenarios and they look good.

rajithaalurims (approved):

PR #3408, opened 2023-07-20
SqlStoreClient class and SqlRetryService signatures plus

SergeyGaluzo:
To increase portability/testability of FHIR codebase this PR creates lightweight initializer of SqlRetryService class and separates simple (no business logic) SQL wrappers to new SqlStoreClient class.

Both these are used in the new EventsReader exe for perf testing.

Also fixing SqlRetryService method signatures.


tools/EventsReader/EventsReader.csproj
```diff
@@ -0,0 +1,19 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <AssemblyName>Fhir.EventsReader</AssemblyName>
```
brendankowitz, line 5: Internal tools should be "Microsoft.Health.Internal"

PR #3479, opened 2023-08-11
History separation

SergeyGaluzo:


src/Microsoft.Health.Fhir.SqlServer/Features/Schema/SchemaVersionConstants.cs
```diff
@@ -7,9 +7,9 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Schema
 {
     public static class SchemaVersionConstants
     {
-        public const int Min = (int)SchemaVersion.V63;
-        public const int Max = (int)SchemaVersion.V63;
-        public const int MinForUpgrade = (int)SchemaVersion.V58; // this is used for upgrade tests only
+        public const int Min = (int)SchemaVersion.V64;
```
brendankowitz, line 10: Should be v60, or no greater than Max-1.

brendankowitz (changes requested):

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/SchemaVersionConstants.cs
SergeyGaluzo: I don't think I can make new code to work with old database schema.

SergeyGaluzo: Addressed

PR #3486, opened 2023-08-18
Convert - remove postprocess exception details from logs

wi-y:
## Description
Previously when a convert postprocess operation failed we would log the details of the failure, which included the part of input message payload which failed to be processed. We also returned the details of the input message in the HTTP response.

Since the input message payload could contain sensitive information, we should not include this information in the logs. Now, in this PR we will no longer log the input message payload in the logs. We will only state that an error has occurred during postprocessing. We will still return the same error details (which includes the partial input message payload that errored) in the HTTP response so that the user has some insight into what caused the error.

## Related issues
Addresses 106437.

## Testing
Unit tests

## FHIR Team Checklist

Patch

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Operations/ConvertData/ConvertDataEngine.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Operations/ConvertData/ConvertDataEngine.cs
@@ -95,7 +95,15 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.ConvertData
                     throw new ConvertDataTimeoutException(Core.Resources.ConvertDataOperationTimeout, convertException.InnerException);
                 }
 
-                _logger.LogInformation(convertException, "Convert data failed.");
+                if (convertException is PostprocessException)
+                {
+                    _logger.LogInformation("A postprocess exception occurred", "Convert data failed.");
+                }
+                else
+                {
+                    _logger.LogInformation(convertException, "Convert data failed.");
+                }
+
                 throw new ConvertDataFailedException(string.Format(Core.Resources.ConvertDataFailed, convertException.Message), convertException);
             }
             catch (Exception ex)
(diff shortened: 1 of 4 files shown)


dustinburson (dismissed):

rogordon01 (approved):

src/Microsoft.Health.Fhir.Core/Features/Operations/ConvertData/ConvertDataEngine.cs
```diff
@@ -95,7 +95,15 @@ private string GetConvertDataResult(ConvertDataRequest convertRequest, ITemplate
...
+                if (convertException is PostprocessException)
+                {
+                    _logger.LogInformation($"An exception of type {convertException.GetType()} occurred during postprocessing. {convertException.StackTrace}", "Convert data failed.");
+                }
+                else
+                {
+                    _logger.LogInformation(convertException, "Convert data failed.");
+                }
+
                 throw new ConvertDataFailedException(string.Format(Core.Resources.ConvertDataFailed, convertException.Message), convertException);
```
LTA-Thinking, line 107: You might want to look in the OperationOutcomeExceptionFilterAttribute to make sure the exception message isn't getting logged there.

LTA-Thinking: Confirmed with Will this is a 400 response code so it wouldn't get logged there.

LTA-Thinking (approved):

src/Microsoft.Health.Fhir.Core/Features/Operations/ConvertData/ConvertDataEngine.cs
wi-y: For reference, this is the line of code that handles the ConvertDataFailedException as an HTTP 400: https://github.com/microsoft/fhir-server/blob/00facae70124c40d5f05fcd26fd504f98cbaced5/src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/OperationOutcomeExceptionFilterAttribute.cs#L159

Which means that the code that logs exceptions >= HTTP 500 would not be executed here:
https://github.com/microsoft/fhir-server/blob/00facae70124c40d5f05fcd26fd504f98cbaced5/src/Microsoft.Health.Fhir.Shared.Api/Features/Filters/OperationOutcomeExceptionFilterAttribute.cs#L269C25-L269C25

Issue #3230, opened 2023-04-13
FHIR Search `_count=0` should be same as `_summary=count`


mikaelweave:
**Describe the bug**
When running a FHIR query with `_count=0`, I get a HTTP 400 response. This is not expected according to the FHIR spec.

`http://<redacted>.azurewebsites.net/Encounter?_count=0`

**FHIR Version?**
Found on R4, possibly applicable to all.

**Data provider?**
Found on SQL, possibly applicable to all.

**To Reproduce**
1. Load a FHIR Server with encounter data.
2. Run a `_count=0` on encounters (`http://<redacted>.azurewebsites.net/Encounter?_count=0`)

**Expected behavior**
According to the FHIR Spec, `_count=0` should be the same as `_summary=count`.

![CleanShot 2023-04-13 at 10 10 19](https://user-images.githubusercontent.com/753437/231834279-e807403b-b48b-4646-bf7d-a2a2c7282477.png)

**Actual behavior**
HTTP 400 response.

![CleanShot 2023-04-13 at 10 11 27](https://user-images.githubusercontent.com/753437/231834577-51b6e26e-8239-4bc5-b7dc-104079912ef9.png)

AB#102605

LTA-Thinking:
This is a bug on the Firely library. Their SearchParam validator requires _count to be >0.
https://github.com/FirelyTeam/firely-net-sdk/blob/develop/src/Hl7.Fhir.Base/Rest/SearchParams.cs#L117

We need to add a filter to remove the _count param and replace it with the _summary param before this line: https://github.com/microsoft/fhir-server/blob/main/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs#L193

LTA-Thinking:
I think this is a good first issue. The fix should be limited to one file, is easy to test manually, and easy to add a unit test for.

brendankowitz:
Done

PR #3491, opened 2023-08-24
Fix FHIR Search query with _count=0 and SqlCustomQueryTest issue

mahajan-xor:
## Description
Fix FHIR Serach query to return result same as _summary=count when passing _count=0.
Previously, 400 response was there for _count=0.
Fix SqlCustomQueryTest issue.

Updated following logic to fix this issue.

1. Adds logic to remove the _count param and replaces it with _summary param in SearchOptionsFactory .
2. Updates GetSummaryTypeOrDefault method in HttpContextExtensions class to return summary type.

## Related issues
Addresses [issue #3230 , #105699].

## Testing
Adds unit test under SearchOperationsFactoryTests
Adds unit test under HttpContextExtensionsTests

## FHIR Team Checklist


src/Microsoft.Health.Fhir.Shared.Api/Features/Formatters/HttpContextExtensions.cs
```diff
@@ -25,6 +25,16 @@ public static SummaryType GetSummaryTypeOrDefault(this HttpContext context)
             {
                 return summary;
             }
+            else if (string.IsNullOrWhiteSpace(query))
+            {
+                var res = context.Request.Query[KnownQueryParameterNames.Count].FirstOrDefault();
+
+                if (!string.IsNullOrWhiteSpace(res) && Convert.ToInt32(res) == 0 &&
```
brendankowitz, line 32: Use a TryParse since this is user input and might not be valid

mahajan-xor: Done . Used TryParse.

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Formatters/HttpContextExtensionsTests.cs
```diff
@@ -42,6 +42,17 @@ public void GivenARequestWithCapsSummaryType_WhenSerializingTheResponse_ThenTheC
             Assert.Equal(SummaryType.Data, summary);
         }
 
+        [Fact]
+        public void GivenARequestWithCountType_WhenSerializingTheResponse_ThenTheCorrectSummeryTypeIsApplied()
```
rbans96, line 46: nit: Summary

mahajan-xor: Done

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlCustomQueryTests.cs
```diff
@@ -49,16 +49,13 @@ public async Task GivenASqlQuery_IfAStoredProcExistsWithMatchingHash_ThenStoredP
             // set the wait time to 1 second
             CustomQueries.WaitTime = 1;
 
-            // Create a query with a known hash
             var queryParameters = new[]
             {
-                Tuple.Create("participating-organization.identifier", $"TAB%7C910000290"),
-                Tuple.Create(SearchParameterNames.Include, $"OrganizationAffiliation:location"),
-                Tuple.Create("participating-organization.type", $"practice"),
```
brendankowitz, line 57: Who was this test changed?

mahajan-xor: Unit test case was failing, needed modification to pass unit test.

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/SearchOptionsFactoryTests.cs
```diff
@@ -110,6 +110,22 @@ public void GivenACount_WhenCreated_ThenCorrectMaxItemCountShouldBeSet()
...
+            var queryParameters = new[]
+            {
+               Tuple.Create("_count", "0"),
+            };
+
+            SearchOptions options = CreateSearchOptions(
+            resourceType: resourceType.ToString(),
+            queryParameters: queryParameters);
+
+            Assert.NotNull(options);
```
LTA-Thinking, line 126: Can you add a check for the summary being count?

LTA-Thinking:
What happens if a user specifies both _count=0 and _summary? This looks like it will take whichever is later in the query parameter  array. I'm going to run a test on this.

LTA-Thinking:
> What happens if a user specifies both _count=0 and _summary? This looks like it will take whichever is later in the query parameter array. I'm going to run a test on this.

Yeah, these queries give different results:
`https://localhost:44348/Patient?_summary=text&_count=0`
`https://localhost:44348/Patient?_count=0&_summary=text`

The first one gives the number of Patients and nothing else.
The second one gives the text summary.

The FHIR spec doesn't say what should be done when multiple summary parameters are given. I feel this should be an error.

Also, we have a bug. When returning just the count of resources we don't return a self link. But [section 3.2.1.7.5](https://www.hl7.org/fhir/search.html#_summary) of the FHIR search spec says we should.

I know we have an existing issue about the order of search parameters affecting results. @brendankowitz thoughts?

LTA-Thinking:
Nevermind, I found the section in the [spec](https://www.hl7.org/fhir/search.html#modifyingresults). It says this is left up to implementations, but they recommend returning an error.

> Note that with the exception of _include and _revinclude, search result parameters SHOULD only appear once in a search. If such a parameter appears more than once, the behavior is undefined and a server MAY treat the situation as an error 

@mahajan-xor I don't think this needs to be part of this PR as it is a wider issue in our service. It is a bug though, and we should track it. I'll make an item for it.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
```diff
@@ -184,6 +184,17 @@ public SearchOptions Create(string resourceType, IReadOnlyList<Tuple<string, str
                         throw new BadRequestException(string.Format(Core.Resources.InvalidTotalParameter, query.Item2, SupportedTotalTypes));
                     }
                 }
+                else if (query.Item1 == KnownQueryParameterNames.Count && Convert.ToInt32(query.Item2) == 0)
```
brendankowitz, line 187: Is there a testcase with `_count=a` or `_count=1.1`?

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/SearchOptionsFactoryTests.cs
mahajan-xor: Done

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/SearchOptionsFactoryTests.cs
```diff
@@ -110,6 +110,41 @@ public void GivenACount_WhenCreated_ThenCorrectMaxItemCountShouldBeSet()
...
+            var queryParameters = new[]
+            {
+               Tuple.Create("_count", value),
+            };
+
+            System.FormatException exception = Assert.Throws<System.FormatException>(() => CreateSearchOptions(
+            resourceType: resourceType.ToString(),
+            queryParameters: queryParameters));
+
+            Assert.Equal("Input string was not in a correct format.", exception.Message);
```
LTA-Thinking, line 145: The expected string doesn't match the actual string.

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/SearchOptionsFactoryTests.cs
LTA-Thinking: I was meaning to check the options.Summary field to ensure it got populated.

mahajan-xor: As discussed added check for options.CountOnly field

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/SearchOptionsFactoryTests.cs
mahajan-xor: As discussed we have analyzed to check type of exception instead of checking actual exception message string

LTA-Thinking (approved):

PR #3503, opened 2023-08-29
[Fix] Limit the number of resources loaded in FHIR Model

fhibf:
## Description
Limiting the number of resources loaded from dbo.System in FHIR Model. 

## Related issues
Addresses [issue #].

## Testing

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
@@ -252,7 +252,7 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Storage
                         COMMIT TRANSACTION
     
                         -- result set 5
-                        SELECT Value, SystemId from dbo.System;
+                        SELECT TOP 10000 Value, SystemId from dbo.System;
 
                         -- result set 6
                         SELECT Value, QuantityCodeId FROM dbo.QuantityCode";


src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
```diff
@@ -252,7 +252,7 @@ BEGIN TRANSACTION
                         COMMIT TRANSACTION
     
                         -- result set 5
-                        SELECT Value, SystemId from dbo.System;
+                        SELECT TOP 10000 Value, SystemId from dbo.System;
```
brendankowitz, line 255: Agree it won't crash, but will this still result in more bugs since data will be missing?

fhibf: Thanks for the comment! I was going to reach out to you for discussion! :D 
Yes, it can. 

I have some options in mind: 
* We limit this query and unblock the account failing in prod. 
* We add to our backlog the dynamic load into memory.

In my mind, having this info added dynamically in memory makes more sense, but we need to unblock a prod account to minimize IcMs. 

From the discussion with the team, having more than 10k resources in dbo.System does not sound correct. Even our biggest customers don't have that number of namespaces.

fhibf:
Closing this PR, as we replaced it with Memory Cache.

PR #3477, opened 2023-08-09
QueueClient handles storage notification

rbans96:
## Description
SqlQueueClient handles StorageInitializedNotification

## Related issues
Addresses [issue #].

## Testing
Describe how this change was tested.

## FHIR Team Checklist
- **Update the title** of the PR to be succinct and less than 65 characters
- **Add a milestone** to the PR for the sprint that it is merged (i.e. add S47)
- Tag the PR with the type of update: **Bug**, **Build**, **Dependencies**, **Enhancement**, **New-Feature** or **Documentation**
- Tag the PR with **Open source**, **Azure API for FHIR** (CosmosDB or common code) or **Azure Healthcare APIs** (SQL or common code) to specify where this change is intended to be released.
- [ ] CI is green before merge [Build Status](https://microsofthealthoss.visualstudio.com/FhirServer/_build/latest?definitionId=27&branchName=main) 
- Review [squash-merge requirements](https://github.com/microsoft/fhir-server/blob/master/SquashMergeRequirements.md)

### Semver Change ([docs](https://github.com/microsoft/fhir-server/blob/master/docs/Versioning.md))
Patch|Skip|Feature|Breaking (reason)

PR #3508, opened 2023-09-01
Revert QueueClient handles storage notification

rbans96:
Reverts microsoft/fhir-server#3477

### DIFF
--- a/build/pr-pipeline.yml
+++ b/build/pr-pipeline.yml
@@ -75,7 +75,7 @@ stages:
       parameters:
         codeCoverage: false
         unitTest: false
-        componentGovernance: true
+        componentGovernance: false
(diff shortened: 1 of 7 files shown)


build/pr-pipeline.yml
```diff
@@ -75,7 +75,7 @@ stages:
       parameters:
         codeCoverage: false
         unitTest: false
-        componentGovernance: true
+        componentGovernance: false
```
LTA-Thinking, line 78: We can't do this, this is a required step now.

LTA-Thinking (changes requested):
We can't disable CG

LTA-Thinking (dismissed):

rbans96:
This is merged as part of other PR

PR #3520, opened 2023-09-12
Fix for Searches not Checking Search Parameter Status

PTaladay:
## Description
Searches now will only include search parameters that are in an "Enabled" state. Those that are not enabled will be added to the unsupported search parameters results, but allow the search to continue on.

Did some refactoring for the searchablesearchparameterdefinitonmanager as it was only being used for reindex which no longer uses that logic to determine if a reindex is needed.

## Related issues
Addresses [issue AB#108891].

## Testing
Added/updated existing tests to verify changes.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Definition/SearchableSearchParameterDefinitionManager.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Definition/SearchableSearchParameterDefinitionManager.cs
@@ -81,9 +81,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Definition
                 return parameter;
             }
 
-            return null;
-
-            // throw new SearchParameterNotSupportedException(resourceType, code);
+            throw new SearchParameterNotSupportedException(resourceType, code);
         }
 
         public SearchParameterInfo GetSearchParameter(string definitionUri)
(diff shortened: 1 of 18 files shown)


src/Microsoft.Health.Fhir.Core/Features/Definition/SearchableSearchParameterDefinitionManager.cs
```diff
@@ -148,5 +156,16 @@ private bool UsePartialSearchParams(SearchParameterInfo parameter)
...
         }
+
+        private bool IsEnabled(SearchParameterInfo parameter)
+        {
+            if (parameter.Code == "_type")
+            {
+                return true;
+            }
+
+            var searchParameterStatuses = _statusManager.GetAllSearchParameterStatus(default).ConfigureAwait(false).GetAwaiter().GetResult();
```
feordin, line 167: Will this line cause a trip to the DB everytime we check if a searchparameter IsEnabled?  I just want to understand if there are any negative perf implications.

PTaladay: Correct, it would. I forgot to pull that outside of the select statement to reduce it to calling once for all of them. Really it would be better to refactor in this area to combine the status management with the definition manager and just have it use the in memory collection that gets updated periodically already.

PR #3523, opened 2023-09-18
Reenable Queue Client in Cosmos

LTA-Thinking:
## Description
Fixes an issue with queue client in Cosmos so it can be reenabled.

## Related issues
Addresses [issue #].

## Testing

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportProcessingJob.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportProcessingJob.cs
@@ -83,7 +83,8 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Export
                 record = exportJobRecord;
                 progress.Report(JsonConvert.SerializeObject(exportJobRecord));
 
-                if (record.Status == OperationStatus.Running)
+                if (record.Status == OperationStatus.Running
+                    && (record.ExportType != ExportJobType.All || record.Filters.Any()))
                 {
                     // Update Export Job is called in three scenarios:
                     // 1. A page of search results is processed and there is a continuation token.
@@ -94,9 +95,16 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Export
                     // For single threaded jobs this section will complete a job after every page of results is processed and make a new job for the next page.
                     // This will allow for saving intermediate states of the job.
 
-                    var definition = jobInfo.DeserializeDefinition<ExportJobRecord>();
-                    definition.Progress = exportJobRecord.Progress;
-                    await _queueClient.EnqueueAsync(QueueType.Export, innerCancellationToken, jobInfo.GroupId, definitions: definition);
+                    record.Status = OperationStatus.Completed;
+                    var existingJobs = await _queueClient.GetJobByGroupIdAsync(QueueType.Export, jobInfo.GroupId, false, cancellationToken);
+
+                    // Checks that a followup job has not already been made.
+                    if (!existingJobs.Any((job) => job.Id > jobInfo.Id))
+                    {
+                        var definition = jobInfo.DeserializeDefinition<ExportJobRecord>();
+                        definition.Progress = exportJobRecord.Progress;
+                        await _queueClient.EnqueueAsync(QueueType.Export, innerCancellationToken, jobInfo.GroupId, definitions: definition);
+                    }
 
                     // TODO: When legacy export support is removed from Cosmos remove this exception and refactor ExportJobTask to expect only one page of results will be processed.
                     throw new JobSegmentCompletedException();
--- a/src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportProcessingJob.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportProcessingJob.cs
@@ -83,8 +83,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Export
                 record = exportJobRecord;
                 progress.Report(JsonConvert.SerializeObject(exportJobRecord));
 
-                if (record.Status == OperationStatus.Running
-                    && (record.ExportType != ExportJobType.All || record.Filters.Any()))
+                if (record.Status == OperationStatus.Running)
                 {
                     // Update Export Job is called in three scenarios:
                     // 1. A page of search results is processed and there is a continuation token.
--- a/src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Queues/CosmosQueueClient.cs
+++ b/src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Queues/CosmosQueueClient.cs
@@ -7,6 +7,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Net;
+using System.Runtime;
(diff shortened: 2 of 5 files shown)


src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportProcessingJob.cs
```diff
@@ -94,9 +95,16 @@ record = exportJobRecord;
                     // For single threaded jobs this section will complete a job after every page of results is processed and make a new job for the next page.
                     // This will allow for saving intermediate states of the job.
 
-                    var definition = jobInfo.DeserializeDefinition<ExportJobRecord>();
-                    definition.Progress = exportJobRecord.Progress;
-                    await _queueClient.EnqueueAsync(QueueType.Export, innerCancellationToken, jobInfo.GroupId, definitions: definition);
+                    record.Status = OperationStatus.Completed;
```
mikaelweave, line 98: The handler for `JobSegmentCompletedException()` sets the completion status, do we need this here?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportProcessingJob.cs
```diff
@@ -83,7 +83,8 @@ async Task<ExportJobOutcome> UpdateExportJob(ExportJobRecord exportJobRecord, We
                 record = exportJobRecord;
                 progress.Report(JsonConvert.SerializeObject(exportJobRecord));
 
-                if (record.Status == OperationStatus.Running)
+                if (record.Status == OperationStatus.Running
```
mikaelweave, line 86: I think this statement may be needed for system level export as well (without filters). I tested this for perf and system level export execution went into this block to queue a new job w/ the continuation token / throw a `JobSegmentException` to set the current job in memory to complete.

This logic is opaque - does this change also work for system level exports without filters?

mikaelweave (commented):
Some questions about the code flow for both of us.

I also wonder why we are throwing a `JobSegmentCompletedException` vs logging a warning / calling that code. I think this area could be improved in general but out of scope of the PR. 😄

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportProcessingJob.cs
```diff
@@ -94,9 +95,16 @@ record = exportJobRecord;
...
                     // This will allow for saving intermediate states of the job.
 
-                    var definition = jobInfo.DeserializeDefinition<ExportJobRecord>();
-                    definition.Progress = exportJobRecord.Progress;
-                    await _queueClient.EnqueueAsync(QueueType.Export, innerCancellationToken, jobInfo.GroupId, definitions: definition);
+                    record.Status = OperationStatus.Completed;
+                    var existingJobs = await _queueClient.GetJobByGroupIdAsync(QueueType.Export, jobInfo.GroupId, false, cancellationToken);
+
+                    // Checks that a followup job has not already been made.
+                    if (!existingJobs.Any((job) => job.Id > jobInfo.Id))
```
brendankowitz, line 102: qq, why was the new definition always different when we requeued?

https://github.com/microsoft/fhir-server/blob/90c537ee52e64cd041bb2354de2374d747f68844/src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Queues/CosmosQueueClient.cs#L66-L77

LTA-Thinking: My guess is because of the CT included in the progress. I haven't paid attention to Cosmos CTs to see if they change over time.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportProcessingJob.cs
LTA-Thinking: That shouldn't have been happening. In system level export each job should only be handling one page of data, there shouldn't be a continuation token.
Ug, this may be another partitioning thing.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportProcessingJob.cs
LTA-Thinking: I added it here in case we hit an exception later. Once the follow up job is created we don't want the code to think the current job is still running. It is probably redundant, but I don't think it hurts anything.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportProcessingJob.cs
mikaelweave: Could be mistaken but this is my mental model from the perf work.

In the new Queue Client implementation,`CosmosExportOrchestratorJob.cs` kicks off a single Export Job for a system level export that may have tens millions of resources. I didn't see any code in the **Cosmos** side to split this up. SQL side does this great.

The Cosmos side was working with the new queue client implementation by creating a single export record and then iterating (mostly single worker doing something at once) over the result until there is no continuation token. The `JobSegmentException` was crucial to this - I remember my logs filling up like crazy.

LTA-Thinking: Ah, you are correct. I was thinking of the SQL pattern, I forgot we had setup Cosmos to just use one job.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Queues/CosmosQueueClient.cs
```diff
@@ -453,7 +454,7 @@ public async Task CompleteJobAsync(JobInfo jobInfo, bool requestCancellationOnFa
                 .WithParameter("@groupId", groupId.ToString())
                 .WithParameter("@queueType", queueType);
 
-            var response = await ExecuteQueryAsync(sqlQuerySpec, 100, cancellationToken);
+            var response = await ExecuteQueryAsync(sqlQuerySpec, 100, queueType, cancellationToken);
```
mikaelweave, line 457: QQ - does this limit the number of responses to only 100? For very large export operations, could this cause an issue?

I saw an issue when building the export response with this code as well - it will only fetch the export file links for the first one hundred group jobs it finds.

```c#
var response = await ExecuteQueryAsync(sqlQuerySpec, 100, cancellationToken);
```eems to limit the number of responses to only 100. For large export operations, could this cause an issue? Would this new "safety" logic be moot for larger exports?

I think I saw an issue when building the export response with this code as well - it will only fetch the export file links for the first 100 group jobs it finds.

```c#
var response = await ExecuteQueryAsync(sqlQuerySpec, 100, cancellationToken);
```

mikaelweave (commented):
Looks good, I just have one question before approving.

mikaelweave (approved):
Looks great! 🦾

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Queues/CosmosQueueClient.cs
brendankowitz: Sorry I didn't reply here, but no specific reason to limit. I think 100 felt high but not infinite

PR #3533, opened 2023-09-22
Updates .NET and docker images

brendankowitz:
## Description
Updates .NET
Switch to Mariner docker images

## Testing
Builds / pipelines should succeed

## FHIR Team Checklist


build/docker/Dockerfile
```diff
@@ -1,8 +1,9 @@
-FROM mcr.microsoft.com/dotnet/sdk:7.0.400-alpine3.17 AS build
-
+FROM mcr.microsoft.com/dotnet/sdk:7.0.401-cbl-mariner2.0-amd64 AS build
```
mikaelweave, line 1: Should we be pinning an architecture here vs using `mcr.microsoft.com/dotnet/sdk:7.0.401-cbl-mariner2.0`?

mikaelweave (commented):
One question on this

build/docker/Dockerfile
fhibf: Should instead of "sdk:7.0.401-cbl-mariner2.0-amd64" we use "sdk:7.0.401-alpine3.18"?

fhibf: Reference: https://hub.docker.com/_/microsoft-dotnet-nightly-sdk/

mikaelweave: @fhibf - we discussed mariner a couple weeks ago in standup. The idea is mariner should run better on Azure than alpine images since it's tuned. Mariner also has security tweaks which makes sense for healthcare industry.

mikaelweave:
Okay, that was ill-advised. 😓 The build environments do not have that version of .NET installed yet. Reverting the commit. I added a screenshot of the local issue, but if it's not happening in the build pipeline I don't think it should block this PR.

CleanShot 2023-10-11 at 09 29 37@2x

brendankowitz:
@mikaelweave I thought that should be ok, we install .net based on the global.json value

mikaelweave:
@brendankowitz it was the code scanning GitHub based pipelines. Looks like we don't install there but use what's on the agents?? I tried this hours after `402` was released. Good to see the agents update quickly.

https://github.com/microsoft/fhir-server/actions/runs/6474809135/job/17580541170

CleanShot 2023-10-12 at 10 12 28@2x

PTaladay:
> @brendankowitz it was the code scanning GitHub based pipelines. Looks like we don't install there but use what's on the agents?? I tried this hours after `7.0.402` was released. Good to see the agents update quickly.
> 
> https://github.com/microsoft/fhir-server/actions/runs/6474809135/job/17580541170
> 
> CleanShot 2023-10-12 at 10 12 28@2x

We are using managed images for our agents that are not in github, but in ADO and are One Engineering recommended images. So however quickly those managed images get updated is how quickly we can utilize new versions of ASP.NET.

mikaelweave:
Makes sense, thank you @PTaladay. 😄

PTaladay (approved):

PR #3541, opened 2023-10-04
Fix syntax check for : when wildcard is used

feordin:
## Description
When searching with an _include and wildcards, the following query would fail:
https://{{fhirServer}}/AllergyIntolerance?patient=foo&_include=*

The FHIR server would require a  : character, for example:
https://{{fhirServer}}/AllergyIntolerance?patient=foo&_include=:*

This fix allows only an * to be present after the = for an _include or _revinclude

## Related issues
Addresses [issue AB#108759].

## Testing
Additional test was added with the new syntax.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/Parsers/ExpressionParser.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/Parsers/ExpressionParser.cs
@@ -72,6 +72,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Expressions.Parsers
             SearchParameterInfo refSearchParameter = null;
             bool wildCard = false;
 
+            // check before split if this is a wildcard match
             if (valueSpan.Equals("*".AsSpan(), StringComparison.InvariantCultureIgnoreCase))
             {
                 refSearchParameter = null;
@@ -81,6 +82,13 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Expressions.Parsers
             if (!TrySplit(SearchSplitChar, ref valueSpan, out ReadOnlySpan<char> originalType) && !wildCard)
             {
                 throw new InvalidSearchOperationException(isReversed ? Core.Resources.RevIncludeMissingType : Core.Resources.IncludeMissingType);
+            }
+
+            // check after split if this is a wildcard match
+            if (valueSpan.Equals("*".AsSpan(), StringComparison.InvariantCultureIgnoreCase))
+            {
+                refSearchParameter = null;
+                wildCard = true;
             }
 
             if (resourceTypes.Length == 1 && resourceTypes[0].Equals(KnownResourceTypes.DomainResource, StringComparison.OrdinalIgnoreCase))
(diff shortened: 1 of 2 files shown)


test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/IncludeSearchTests.cs
```diff
@@ -273,6 +273,17 @@ public async Task GivenAnIncludeSearchExpression_WhenSearched_DoesNotIncludeUnty
                 Fixture.ObservationWithUntypedReferences);
         }
 
+        [Fact]
+        public async Task GivenAnIncludeSearchExpressionWithAllResources_WhenSearched_DoesNotIncludeUntypedReferences()
+        {
+            string query = $"_id={Fixture.ObservationWithUntypedReferences.Id}&_include=*";
```
brendankowitz, line 279: What does _include=:* do now, is it the same?
Also do we have tests for _include=Observation:* too? (i.e. the other permutations of this syntax)

feordin: Good questions!  The behavior between the two syntax is the same.  We just don't error out if the : is present,.
Yes, we do have a test for the other syntax, the new test case I added was modified from the test with the :

brendankowitz (dismissed):
:shipit:

src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/Parsers/ExpressionParser.cs
```diff
@@ -68,11 +68,29 @@ public Expression Parse(string[] resourceTypes, string key, string value)
...
+            bool wildCard = false;
+
+            // check before split if this is a wildcard match
+            if (valueSpan.Equals("*".AsSpan(), StringComparison.InvariantCultureIgnoreCase))
+            {
+                refSearchParameter = null;
+                wildCard = true;
+            }
+
+            if (!TrySplit(SearchSplitChar, ref valueSpan, out ReadOnlySpan<char> originalType) && !wildCard)
```
brendankowitz, line 82: If `valueSpan == "*"` then we shouldn't need to check for the split char, so perhaps:

```suggestion
            else if (!TrySplit(SearchSplitChar, ref valueSpan, out ReadOnlySpan<char> originalType) && !wildCard)
```

src/Microsoft.Health.Fhir.Core/Features/Search/Expressions/Parsers/ExpressionParser.cs
```diff
@@ -68,11 +68,29 @@ public Expression Parse(string[] resourceTypes, string key, string value)
...
             {
                 throw new InvalidSearchOperationException(isReversed ? Core.Resources.RevIncludeMissingType : Core.Resources.IncludeMissingType);
             }
 
+            // check after split if this is a wildcard match
+            if (valueSpan.Equals("*".AsSpan(), StringComparison.InvariantCultureIgnoreCase))
+            {
+                refSearchParameter = null;
+                wildCard = true;
+            }
```
brendankowitz, line 92: Then remove the after check
```suggestion
```

feordin: The second check is still needed as valueSpan is passed by ref and the value updated to be just the '*' if the incoming string is something like:  Observation:* .  But I updated it to shortcircuit if wildcard has already been set to true.

brendankowitz (dismissed):

brendankowitz (approved):

PR #3544, opened 2023-10-09
Add Export Max Count parameter

LTA-Thinking:
## Description
Add a new parameter _maxCount to the export operation to allow users to specify how many resources are fetched per page. This allows users with small resources to increase throughput, and users with large resources to avoid out of memory exceptions.

## Related issues
Addresses [Bug 110476](https://microsofthealth.visualstudio.com/Health/_workitems/edit/110476): Export OOM

## Testing
Manual and existing tests.

## FHIR Team Checklist

Patch


src/Microsoft.Health.Fhir.Shared.Api/Controllers/ExportController.cs
```diff
@@ -97,6 +98,7 @@ public class ExportController : Controller
             [FromQuery(Name = KnownQueryParameterNames.TypeFilter)] string typeFilter,
             [FromQuery(Name = KnownQueryParameterNames.Format)] string formatName,
             [FromQuery(Name = KnownQueryParameterNames.IsParallel)] bool isParallel = true,
+            [FromQuery(Name = KnownQueryParameterNames.MaxCount)] uint maxCount = 0,
```
mikaelweave, line 101: QQ - why does `maxCount` default to zero on system level export but not default on the others?

mikaelweave (commented):
One question on controller defaults. Also, does this need test coverage? Not sure where it would go - maybe a very simple test in `CreateExportRequestHandlerTests`?

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ExportController.cs
LTA-Thinking: It is because the bottom parameters (typeParameter and idParameter) can't have default values. I moved those parameters to the top and set default values. I did test that the lack of default values didn't make a difference, but it is good for consistency.

mikaelweave (commented):
Looks good! 🦾 Oops forgot to approve will do now.

mikaelweave (approved):

PR #3547, opened 2023-10-11
Restore resourcetype index hint to SQL Query generator

feordin:
## Description
A check for schema version was added to the SQL Query generator which caused the generator to no longer add a hint about an index to use when querying the resource table.  Failing to include this hint has caused poor performance in some queries as SQL fails to use the index and does a scan on the resource table.

Removing the check for the schema version will cause the hint to be added to the generated SQL once again.

## Related issues
Addresses [issue AB#104994].

## Testing
Existing tests were run, and the problematic query was manually executed.  (baseUrl/Slot?_count=1000)

## FHIR Team Checklist


SergeyGaluzo (approved):
:shipit:

PTaladay (approved):

PR #3564, opened 2023-10-18
Added support for extension_fhirUser claim for SMART auth

namalu:
## Description
This PR adds support for extension_fhirUser token claim which can be used with Azure B2C to map FHIR user resources with B2C user accounts. This PR also removes support for sending FHIR user data in a request header.

Milestone S126

## Testing
Manual testing was performed to ensure these changes work with Azure B2C.
Unit tests have been updated to ensure code quality and guard against regression.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Api/Features/SMART/SmartClinicalScopesMiddleware.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/SMART/SmartClinicalScopesMiddleware.cs
@@ -134,15 +134,13 @@ namespace Microsoft.Health.Fhir.Api.Features.Smart
 
                     if (includeFhirUserClaim)
                     {
+                        // Check if the "fhirUser" claim is present.
                         var fhirUser = principal.FindFirstValue(authorizationConfiguration.FhirUserClaim);
                         if (string.IsNullOrEmpty(fhirUser))
                         {
-                            // look for the fhirUser info in a header
-                            if (context.Request.Headers.ContainsKey(KnownHeaders.FhirUserHeader)
-                                && context.Request.Headers.TryGetValue(KnownHeaders.FhirUserHeader, out var hValue))
-                            {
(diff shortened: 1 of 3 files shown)


src/Microsoft.Health.Fhir.Api/Features/SMART/SmartClinicalScopesMiddleware.cs
```diff
@@ -134,15 +134,13 @@ public SmartClinicalScopesMiddleware(RequestDelegate next, ILogger<SmartClinical
...
                         var fhirUser = principal.FindFirstValue(authorizationConfiguration.FhirUserClaim);
                         if (string.IsNullOrEmpty(fhirUser))
                         {
-                            // look for the fhirUser info in a header
-                            if (context.Request.Headers.ContainsKey(KnownHeaders.FhirUserHeader)
-                                && context.Request.Headers.TryGetValue(KnownHeaders.FhirUserHeader, out var hValue))
-                            {
-                                fhirUser = hValue.ToString();
-                            }
+                            // The "fhirUser" claim is not present, check if the "extension_fhirUser" claim is present.
```
feordin, line 141: We did discuss that if both the extension_fhirUser and fhirUser claim were both present and did not agree that we should throw an exception.

It does seem clean here that we simply prefer fhirUser, and only check extension_fhirUser if necessary.  I am open to leaving it as is, or adding the exception logic.

namalu: I'm inclined to leave this as is. Because the claims are generated and signed by the issuer, I don't think having a preference for one claim over another raises a security concern. Also, other than significant human error, I would imagine that if both claims did exist they would likely point to the same FHIR resource.

feordin (approved):

PR #3584, opened 2023-11-03
Add Parallel Cosmos Export

mikaelweave:
## Description
Adds parallel export to CosmosDB by resource type and feed range. This export does not work like SQL because CosmosDB doesn't work like SQL. It's costly to get date ranges in CosmosDB

Performance is >100K resources per second on a 100K RU CosmosDB wth 300 million documents. 20 node P2V2 App Service. Exporting history gets closer to 200K resources per second.

## Related issues
Addresses AB112054.

## Testing
LOTS of local testing and testing in Azure OSS. I've mainly been testing with 50 million docs in my CosmosDB.
Xoriant has also been testing with a 300 million doc db.

## FHIR Team Checklist


src/Microsoft.Health.Fhir.Core/Features/KnownQueryParameterNames.cs
```diff
@@ -90,6 +90,8 @@ public static class KnownQueryParameterNames
 
         public const string IgnoreSearchParamHash = "_ignoreSearchParamHash";
 
+        public const string FeedRange = "_feedRange";
```
mikaelweave, line 93: I struggled with the decision to make this a query parameter or not. In the end I did because the surrogate id parameters are similar.

If this shouldn't be a query param should the surrogate id params be here either?

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Features/Operations/Export/CosmosDbExportTests.cs
```diff
@@ -0,0 +1,174 @@
...
+        }
+
+        [Fact]
+        public async Task ExportWorkRegistration()
+        {
+            try
+            {
+                await PrepareData(3333); // 1111 patients + 1111 observations + 1111 claims.
+
+                var coordJob = new CosmosExportOrchestratorJob(_queueClient, _searchService);
```
mikaelweave, line 62: I have a couple more small changes to make here

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportProcessingJob.cs
```diff
@@ -95,17 +97,27 @@ record = exportJobRecord;
...
-                    if (!existingJobs.Any((job) => job.Id > jobInfo.Id))
+                    // Only queue new jobs if group has no canceled jobs. This ensures canceled jobs won't continue to queue new jobs (Cosmos cancel is not 100% reliable).
+                    if (!existingJobs.Any(job => job.Status == JobStatus.Cancelled || job.CancelRequested))
                     {
                         var definition = jobInfo.DeserializeDefinition<ExportJobRecord>();
-                        definition.Progress = exportJobRecord.Progress;
-                        await _queueClient.EnqueueAsync(QueueType.Export, innerCancellationToken, jobInfo.GroupId, definitions: definition);
+
+                        // Checks that a follow up job has not already been made. Extra checks are needed for parallel jobs by parallelization factors.
+                        var newerJobs = existingJobs.Where(existingJob =>
```
LTA-Thinking, line 108: Nit: this could just be an Any since the contents of newerJobs is never used.

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
```diff
@@ -449,23 +455,59 @@ private async Task<Expression> RecurseChainedExpression(ChainedExpression expres
...
+
+                    // Large async operations also need the full result set. Accounting for the till factor to achieve that.
+                    if (searchOptions.IsLargeAsyncOperation)
+                    {
+                        mustNotExceedMaxItemCount = false;
+                        feedOptions.MaxItemCount = (int)(feedOptions.MaxItemCount / CosmosFhirDataStore.ExecuteDocumentQueryAsyncMinimumFillFactor);
+
+                        try
+                        {
+                            queryFeedRange = searchOptions.FeedRange is not null ? FeedRange.FromJsonString(searchOptions.FeedRange) : queryFeedRange;
```
LTA-Thinking, line 495: Nit: Setting queryFeedRange equal to itself when the only possible value at this point is null seems unnecessarily confusing. I think just setting it equal to null would be clearer.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Queues/CosmosQueueClient.cs
```diff
@@ -331,6 +331,11 @@ public async Task<bool> PutJobHeartbeatAsync(JobInfo jobInfo, CancellationToken
             var job = jobs.Single();
             JobDefinitionWrapper item = job.MatchingJob.Single();
 
+            if (item.CancelRequested)
+            {
+                return true; // If the job is cancelled, the job data doesn't need to be updated.
```
LTA-Thinking, line 336: I wouldn't say this is a safe assumption to make. Even if the job is cancelled it may have data that needs to be updated to show what point the job got to before being cancelled.

src/Microsoft.Health.Fhir.Core/Features/KnownQueryParameterNames.cs
mikaelweave: Given all the params I've been adding in this area (😂), the only other solution that comes to mind is separating these out from being passed as "query parameters" to another object that has "system request modifiers" as properties. Cosmos and SQL could have their own derived classes (no `FeedRange` for SQL for example).

Not sure if it's worth the work. Thinking of something like

*SystemRequestModifiers.cs*
``` csharp
{
    public class SystemRequestModifiers
    {
        public bool UseSmartCompartmentDefinitions {get; set; }
        public ResourceTypeVersion useSmartCompartmentDefinitions {get; set; }
        public bool IsAsyncOperation {get; set; }
        ....
    }
}
```

*SqlSystemRequestModifiers.cs*
``` csharp
{
    public class SqlSystemRequestModifiers : SystemRequestModifiers
    {
        public long GlobalStartId {get; set; }
        public long GlobalEndIt {get; set; }
        public long GlobalSurrogateStartId {get; set; }
        public long GlobalSurrogateEndId {get; set; }
    }
}
```

*CosmosSystemRequestModifiers.cs*
``` csharp
{
    public class CosmosSystemRequestModifiers : SystemRequestModifiers
    {
        public FeedRange FeedRange {get; set; }
    }
}
```

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Queues/CosmosQueueClient.cs
mikaelweave: Good call - I don't think this will impact perf of the cancel call but will run a test tomorrow to make sure. If the `CancelRequested` flag is set, the cancellation job won't be trying to change the document at the same time.

The issue is the heartbeats, completed jobs, new jobs and the cancels all at the same time was causing the parallel threads to compete in their optimistic concurrency. I may have gone overboard trying to fix it. 😫

src/Microsoft.Health.Fhir.CosmosDb/Resources.resx
```diff
@@ -1,17 +1,17 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <root>
-  <!--
-    Microsoft ResX Schema
-
+  <!--
```
mikaelweave, line 3: Will fix whatever happened here - been switching between OS X and Windows...

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/CosmosDbFhirStorageTestsFixture.cs
```diff
@@ -155,14 +157,14 @@ public async Task InitializeAsync()
                 CollectionInitializationSemaphore.Release();
             }
 
-            var cosmosDocumentQueryFactory = new CosmosQueryFactory(cosmosResponseProcessor, NullFhirCosmosQueryLogger.Instance);
+            _cosmosQueryFactory = new CosmosQueryFactory(cosmosResponseProcessor, NullFhirCosmosQueryLogger.Instance);
```
mikaelweave, line 160: This could be a merge issue. I will track down the conflicting change as I don't believe I modified this file.

src/Microsoft.Health.Fhir.CosmosDb/Features/Search/FhirCosmosSearchService.cs
mikaelweave: Agreed - thanks. Sometimes one liners are not the right approach, haha.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportProcessingJob.cs
```diff
@@ -95,17 +97,27 @@ record = exportJobRecord;
...
+                        });
+
+                        if (!newerJobsExist)
+                        {
+                            definition.Progress = exportJobRecord.Progress;
+                            await _queueClient.EnqueueAsync(QueueType.Export, innerCancellationToken, jobInfo.GroupId, definitions: definition);
+                        }
                     }
 
-                    // TODO: When legacy export support is removed from Cosmos remove this exception and refactor ExportJobTask to expect only one page of results will be processed.
```
LTA-Thinking, line 108: I would still like to clean this up at some point, but I guess Cosmos being removed will happen sooner than that.

src/Microsoft.Health.Fhir.CosmosDb/Features/Operations/Export/CosmosExportOrchestratorJob.cs
```diff
@@ -20,10 +24,14 @@ namespace Microsoft.Health.Fhir.CosmosDb.Features.Operations.Export
     public class CosmosExportOrchestratorJob : IJob
     {
         private readonly IQueueClient _queueClient;
+        private ISearchService _searchService;
 
-        public CosmosExportOrchestratorJob(IQueueClient queueClient)
+        public CosmosExportOrchestratorJob(
+            IQueueClient queueClient,
+            ISearchService searchService)
```
LTA-Thinking, line 31: This should be a Func<IScoped<ISearchService>>

src/Microsoft.Health.Fhir.CosmosDb/Features/Operations/Export/CosmosExportOrchestratorJob.cs
```diff
@@ -36,7 +44,43 @@ public async Task<string> ExecuteAsync(JobInfo jobInfo, IProgress<string> progre
...
+                    {
+                        var processingRecord = CreateExportRecord(
+                                record,
+                                jobInfo.GroupId,
+                                resourceType: resourceType,
+                                feedRange: partitionRange);
+
+                        string[] definitions = [JsonConvert.SerializeObject(processingRecord)];
+
+                        await _queueClient.EnqueueAsync((byte)QueueType.Export, definitions, jobInfo.GroupId, false, false, cancellationToken);
```
LTA-Thinking, line 79: To make sure I'm understanding this correctly: one job is being created for every resource type / physical partition combination?

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportProcessingJob.cs
mikaelweave: I agree - though our current data schema with Cosmos current GA features means the exported rows would need to be traversed twice. The current implementation is what the Cosmos team recommended but open to change it in the future (or now if it's important).

If our date/time format was true ISO 8601, we could use a swath of date time functions (like `DateTimeBin`) that would approach SQL windowing.

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Export/ExportProcessingJobTests.cs
```diff
@@ -133,16 +133,79 @@ public async Task GivenAnExportJob_WhenItFinishesAPageOfResultsAndAFollowupJobEx
...
+
+            var followUpJob = GenerateJobInfo(GenerateJobRecord(OperationStatus.Running, resourceType: resourceType2, feedRange: feedRange2));
+            followUpJob.Id = runningJob.Id + 1;
+            queueClient.JobInfos.Add(followUpJob);
+
+            var taskResult = await processingJob.ExecuteAsync(runningJob, progress, CancellationToken.None);
+
+            Assert.True(queueClient.JobInfos.Count == 2);
+            Assert.DoesNotContain(_progressToken, queueClient.JobInfos[0].Definition);
+            Assert.Equal(followUpJob, queueClient.JobInfos[0]);
```
LTA-Thinking, line 161: Can an assert be added to show the second job is a follow up to the running job? Also, the name of followUpJob maybe should be changed in this test since it is a follow up to a parallel job, not the running job.

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Export/ExportProcessingJobTests.cs
```diff
@@ -133,16 +133,79 @@ public async Task GivenAnExportJob_WhenItFinishesAPageOfResultsAndAFollowupJobEx
...
+
+            var existingCanceledJob = GenerateJobInfo(GenerateJobRecord(OperationStatus.Running));
+            existingCanceledJob.Id = runningJob.Id - 2;
+            existingCanceledJob.Status = JobStatus.Cancelled;
+            queueClient.JobInfos.Add(existingCanceledJob);
+
+            var existingCanceledPendingJob = GenerateJobInfo(GenerateJobRecord(OperationStatus.Running));
+            existingCanceledPendingJob.Id = runningJob.Id - 1;
+            existingCanceledPendingJob.CancelRequested = true;
+            queueClient.JobInfos.Add(existingCanceledPendingJob);
```
LTA-Thinking, line 187: Should these be two different theories instead of both being part of one test? IE: one for cancelled and one for cancel requested?

src/Microsoft.Health.Fhir.CosmosDb/Features/Operations/Export/CosmosExportOrchestratorJob.cs
mikaelweave: Correct - these will queue up have different execution periods. If a database is heavy on Observations, the observation export will take the longest to run (depending on partition distribution). The larger the database the more parallel the job can execute.

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Features/Operations/Export/CosmosDbExportTests.cs
```diff
@@ -0,0 +1,174 @@
...
+            _fixture = fixture;
+            _testOutputHelper = testOutputHelper;
+            _searchService = _fixture.GetService<ISearchService>();
+            _operationDataStore = _fixture.GetService<CosmosFhirOperationDataStore>();
+            _queueClient = _fixture.GetService<IQueueClient>();
+            _fhirStorageTestHelper = _fixture.GetService<IFhirStorageTestHelper>();
+        }
+
+        [Fact]
+        public async Task ExportWorkRegistration()
```
LTA-Thinking, line 56: Should this be one test? It looks like it is three different tests.

src/Microsoft.Health.Fhir.CosmosDb.UnitTests/Features/Operations/Export/ExportOrchestratorJobTests.cs
```diff
@@ -0,0 +1,268 @@
...
+using Microsoft.Health.Test.Utilities;
+using Newtonsoft.Json;
+using NSubstitute;
+using Xunit;
+
+namespace Microsoft.Health.Fhir.CosmosDb.UnitTests.Features.Operations.Export
+{
+    [Trait(Traits.OwningTeam, OwningTeam.Fhir)]
+    [Trait(Traits.Category, Categories.Export)]
+    public class ExportOrchestratorJobTests
```
LTA-Thinking, line 28: This should be CosmosExportOrchestratorJobTests

mikaelweave: Will change SQL as well - initially was trying to mirror SQL tests.

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Features/Operations/Export/CosmosDbExportTests.cs
mikaelweave: Will investigate - I used the same format as the SQL tests but agree it doesn't seem right.

LTA-Thinking (approved):

PR #3593, opened 2023-11-14
[Bundle] Fix racing condition during parallel processing

fhibf:
Under racing conditions, it's possible that the same record (_< resourceType >/< id >_) is present in two bundles running at the same time: one record is updated first than the second, increasing the version of the record, and then updated once more by the second bundle. As the version is higher than the expected, the local identifier in the current thread does not match, and an alternative search for the combination < resourceType >/< id > is required.

## Testing
A new integration tests was created to simulate this scenario. It's a test simulating 4 different bundles updating the same 4 records at the same time, and causing the concurrency issue to happen. 

## FHIR Team Checklist


src/Microsoft.Health.Fhir.Core/Features/Persistence/Orchestration/BundleOrchestratorOperation.cs
```diff
@@ -132,26 +133,55 @@ public async Task<UpsertOutcome> AppendResourceAsync(ResourceWrapperOperation re
...
+                    // One record is updated first than the second, increasing the version of the record, and then updated once more by the second bundle.
+                    // As the version is higher than the expected, the local identifier in the current thread does not match, and an alternative search for
+                    // the combination <resourceType>/<id> is required.
+                    var ingestedResourcesById = ingestedResources.Where(i => i.Key.ResourceType == identifier.ResourceType && i.Key.Id == identifier.Id).ToList();
+
+                    int countOfResourcesFound = ingestedResourcesById.Count;
+                    if (countOfResourcesFound == 1)
                     {
-                        throw dataStoreOperationOutcome.Exception;
+                        // Edge Case scenario: resource was updated by another bundle concurrently, but it was also updated by the current bundle, increasing its version.
```
abiisnn, line 151: Where are we **"increasing the version"**? or you mean **"updating the version"**?

abiisnn: Can we add a log to know the resource was sent twice and we have already updated it, just for our records or debugging? ofc if you think it's helpful

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/BundleEdgeCaseTests.cs
```diff
@@ -91,5 +96,133 @@ public async Task GivenABundleWithConditionalUpdateByReference_WhenExecutedWithM
...
+                }
+            }
+
+            Assert.True(
+                validatedResources == totalNumberOfExpectedPatients,
+                $"Number of expected patients ({totalNumberOfExpectedPatients}) is different than the final number of validated resources ({validatedResources}).");
+
+            Assert.True(
+                uniquePatientIds.Count == numberOfPatientsPerBundle,
+                $"Total mumber of created patients ({numberOfPatientsPerBundle}) is different than the number ingested during the test {uniquePatientIds}.");
```
abiisnn, line 194: Can you add an "Assert" for the final version of the resource? which is the expected output after the parallel operations?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/BundleEdgeCaseTests.cs
```diff
@@ -91,5 +96,133 @@ public async Task GivenABundleWithConditionalUpdateByReference_WhenExecutedWithM
...
+        }
```
abiisnn, line 195: Would you mind adding a test for to check that it does not loops forever? ➿

abiisnn (commented):
Thanks for fixing it Fernando, and for explaining everything in comments. Just a couple of questions!

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/BundleEdgeCaseTests.cs
abiisnn: Also I do see 3 tests cases when countOfResources == 0, countOfResources == 1 and countOfResources > 1, is it possible to add a test case per case or do you think we are confident with one tests for all of them?

src/Microsoft.Health.Fhir.Core/Features/Persistence/Orchestration/BundleOrchestratorOperation.cs
```diff
@@ -132,26 +133,55 @@ public async Task<UpsertOutcome> AppendResourceAsync(ResourceWrapperOperation re
...
+
+                // Attempt 1: Retrieve from the list of merged records, the one with the same identifier from current thread.
+                if (!ingestedResources.TryGetValue(identifier, out dataStoreOperationOutcome))
                 {
-                    if (!dataStoreOperationOutcome.IsOperationSuccessful)
+                    // Attemp 2: Edge case scenario:
+                    // Under racing conditions, it's possible that the same record (<resourceType>/<id>) is present in two bundles running at the same time.
+                    // One record is updated first than the second, increasing the version of the record, and then updated once more by the second bundle.
+                    // As the version is higher than the expected, the local identifier in the current thread does not match, and an alternative search for
+                    // the combination <resourceType>/<id> is required.
```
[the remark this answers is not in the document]
fhibf: **Same Resource ID**
If there are two resources with the same content, but different resource IDs, then there is no issue.

src/Microsoft.Health.Fhir.Core/Features/Persistence/Orchestration/BundleOrchestratorOperation.cs
fhibf: The version is increased on SQL Server side. 

I can add a log to this edge case scenario mentioning we handled it.

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Persistence/Orchestration/BundleOrchestratorOperationTests.cs
```diff
@@ -107,6 +108,56 @@ public void GivenABatchOperation_WhenAppendedMultipleResourcesInParallelWaitForA
             Assert.Equal(numberOfResources, operation.CurrentExpectedNumberOfResources);
         }
 
+        [Fact]
+        public async Task GivenABatchOperation_WhenJustOneResourcedIsAppendedAndAllOtherResourcesAreHanging_ThenCancelTheOperationAfter120SecondsToAvoidRunningForever()
+        {
+            // Long running test.
+
+            // A security exit clause in Bundle Orchestrator Operation throws a TaskCanceledException after 100 seconds waiting for recourses.
```
abiisnn, line 116: NIT: Resources

abiisnn (dismissed):
🍒

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/BundleEdgeCaseTests.cs
```diff
@@ -91,5 +96,139 @@ public async Task GivenABundleWithConditionalUpdateByReference_WhenExecutedWithM
                 Assert.Equal("2", bundleResponse2.Resource.Entry[i].Resource.Meta.VersionId);
             }
         }
+
+        [Fact]
+        [HttpIntegrationFixtureArgumentSets(dataStores: DataStore.SqlServer)]
```
LTA-Thinking, line 101: Why is this only for SQL? The logic looks like it would work for Cosmos as well.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/BundleEdgeCaseTests.cs
```diff
@@ -31,6 +35,7 @@ public BundleEdgeCaseTests(HttpIntegrationTestFixture fixture)
         }
 
         [Fact]
+        [HttpIntegrationFixtureArgumentSets(dataStores: DataStore.CosmosDb)]
```
LTA-Thinking, line 38: Why is this one only for Cosmos?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/BundleEdgeCaseTests.cs
fhibf: That's a good question. I created this test focused on a bug that only happened in Gen2 due the way how MergeAsync works. But it makes sense to keep it running for Gen1 and Gen2.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/BundleEdgeCaseTests.cs
fhibf: That's a good question. I created this test focused on a scenario for Gen1. But it makes sense to keep it running for both generations.

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Persistence/Orchestration/BundleOrchestratorOperationTests.cs
```diff
@@ -107,6 +105,56 @@ public void GivenABatchOperation_WhenAppendedMultipleResourcesInParallelWaitForA
...
+
+            // A security exit clause in Bundle Orchestrator Operation throws a TaskCanceledException after 100 seconds waiting for resources.
+            // This security exit clause was added to avoid a looping running forever, while resources are not appended.
+
+            // In this test, to simulate that scenario, a bundle operation is created expecting 10 resources, but only one is appended.
+            // The other 9 resources are never appened, forcing the looping to keep waiting for the remaining resources.
+            // This security exit clause will be activated after 120 seconds, cancelling the operation.
+
+            const int numberOfResources = 10;
+            const int maxWaitingTimeInSeconds = 150;
```
rajithaalurims, line 121: nit: Is this supposed to be 120?

fhibf: I set it to 150 because it's supposed to be higher than 100s. 
The reason is: the internal cancellation will happen at 100s, if the external cancellation happens with 150, then it means that the internal one did not happen. I just added some additional time to the external cancellation. 

By the way, I just noticed that I need to update the comment with 100, instead of 120. 
I'll also decrease the external cancellation from 150s to 120s. It does not need to be as high as 150s.

feordin (approved):

rajithaalurims (approved):

PR #3603, opened 2023-11-27
Check for Disabled Search Parameters in Searches

PTaladay:
## Description
This PR adds in validation for search parameters that are not in an enabled state. If there is a search parameter that is in the search that is not enabled then a bundle issue is created and it is added to the unsupported search parameters in the search options object that is returned from the SearchOptionsFactory. It also adds additional tests to cover include and revinclude scenarios. 

Notes:
The check does include a call out to the database in order to grab the search parameter statuses. This is because the SearchParameterStatusManager does not cache these results, unlike the SearchParameterDefinitonManager which does. There is a user story to refactor the way Search Parameters are handled and kept in sync with the datastore such that it minimizes trips to the databases, but also stays in sync. 
  The performance impact with having a call to gather statuses for each search request is minimal as the search parameter table has under 1500 rows. Once we have the refactoring done we can move away from the db call added here.

I also added the check at the end of the Create method call and not earlier where it is going through the searchExpressions.Parameters since at that point the include and reinclude expressions have not been parsed at that time. Adding to the end of the method call ensures that all expressions are parsed. 

## Related issues
Addresses [issue AB#108891, AB#108894].
-- Refactoring user story [AB#104464]

## Testing

## FHIR Team Checklist


src/Microsoft.Health.Fhir.Core/Resources.resx
```diff
@@ -736,4 +736,8 @@
     <value>A resource should only appear once in each Bundle.</value>
     <comment>Error message for a duplicate resource key in the same bundle</comment>
   </data>
+  <data name="SearchParameterNotEnabledErrorMessage" xml:space="preserve">
+    <value>The search parameter: '{0}' is currently set as: '{1}', which is not a valid state and can't be used in this search. Please run reindex if state is shown as supported to include this search parameter in your query. The search will run without this search parameter, but may not return what you are expecting.</value>
+    <comment>{NumberedPlaceHolder="{0}", "{1}"}</comment>
+  </data>
```
LTA-Thinking, line 742: Should we include reference to the x-ms-use-partial-indices header?

PR #3619, opened 2023-12-07
Version bug fix

SergeyGaluzo:
Fixes https://microsofthealth.visualstudio.com/Health/_workitems/edit/112948


abiisnn (dismissed):
I have pulled your branch and tested that now it's expecting as expected. Thanks for fixing it. 🍒

LTA-Thinking (approved):

PR #3623, opened 2023-12-10
Remove clutter from SQL generator

SergeyGaluzo:


src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -1261,14 +1265,10 @@ private void AppendIntersectionWithPredecessorUsingInnerJoin(IndentedStringBuild
                 // To simplify query plan generation, if we are intersecting with the Reference search param table, we will use an inner join
                 // rather than an EXISTS clause.  We have see that this significanlty reduces the query plan generation time for
                 // complex queries
```
LTA-Thinking, line 1267: The comment needs updating to reflect the use of a join instead of an inner join.

LTA-Thinking (approved):

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
SergeyGaluzo: INNER JOIN == JOIN. One can refer to JOIN as INNER JOIN.

PR #3656, opened 2024-01-04
Removing search param and resource type ids from sql parameters

SergeyGaluzo:
Fixes https://microsofthealth.visualstudio.com/Health/_workitems/edit/112586/

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Search/HashingSqlQueryParameterManager.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Search/HashingSqlQueryParameterManager.cs
@@ -45,7 +45,7 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Search
         /// Whether this parameter should be included in the hash of the overall parameters.
         /// If true, this parameter will prevent other identical queries with a different value for this parameter from re-using the query plan.
         /// </param>
-        /// <returns>The SQL parameter.</returns>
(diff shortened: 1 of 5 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Search/HashingSqlQueryParameterManager.cs
```diff
@@ -45,13 +46,13 @@ public HashingSqlQueryParameterManager(SqlQueryParameterManager inner)
         /// If true, this parameter will prevent other identical queries with a different value for this parameter from re-using the query plan.
         /// </param>
         /// <returns>The SQL parameter.</returns>
-        public SqlParameter AddParameter<T>(Column<T> column, T value, bool includeInHash)
+        public object AddParameter<T>(Column<T> column, T value, bool includeInHash)
```
feordin, line 49: Why did we change the return type?

SergeyGaluzo: I want to have a single place in the code where we can make a decision what to return @pXX or value itself. This allows to keep logic in all other places as it is today, with slight modification that instead of .ParameterName we send object itself, so implicit .ToString() on SqlParameter returns same name as before, but on integer value, it returns value itself.

feordin (approved):

PR #3657, opened 2024-01-04
Fix Cosmos Bulk Delete

LTA-Thinking:
## Description
Adds a throttler to Cosmos Bulk Delete and fixes a bug with auth tokens expiring after an hour.

## Related issues
Addresses [Bug 113641](https://microsofthealth.visualstudio.com/Health/_workitems/edit/113641): Gen 1 Bulk Delete auth tokens expire after an hour

## Testing

## FHIR Team Checklist


src/Microsoft.Health.Fhir.Core.UnitTests/SubstituteHelpers.cs
```diff
@@ -0,0 +1,21 @@
...
+
+using System;
+using Microsoft.Health.Extensions.DependencyInjection;
+using NSubstitute;
+
+namespace Microsoft.Health.Fhir.Core.UnitTests
+{
+    public class SubstituteHelpers
+    {
+        public static Func<IScoped<T>> ScopedFunc<T>(T obj)
```
brendankowitz, line 14: Looks pretty similar to:

https://github.com/microsoft/fhir-server/blob/f0776e2471f3340f163e494220b1cfbdf94147ce/src/Microsoft.Health.Fhir.Core.UnitTests/Extensions/CreateMockedScopeExtensions.cs#L13-L18

I think in other places the usage was something like `Method(param1, () => obj.CreateMockScope())`

src/Microsoft.Health.Fhir.CosmosDb/Features/Operations/Reindex/CosmosBackgroundJobThrottleController.cs
```diff
@@ -30,40 +34,52 @@ public class ReindexJobCosmosThrottleController : IReindexJobThrottleController
...
         {
-            EnsureArg.IsNotNull(reindexJobRecord, nameof(reindexJobRecord));
+            EnsureArg.IsNotNull(jobRecord, nameof(jobRecord));
+
+            if (jobRecord.TargetDataStoreUsagePercentage != null &&
+                    jobRecord.TargetDataStoreUsagePercentage > 0)
+            {
+                using (IScoped<IFhirDataStore> store = _fhirDataStoreFactory.Invoke())
+                {
+                    _provisionedRUThroughput = await store.Value.GetProvisionedDataStoreCapacityAsync(cancellationToken);
```
brendankowitz, line 61: @feordin is this the call that doesn't work in paas? I also don't think we can read these values under Cosmos/MI

src/Microsoft.Health.Fhir.CosmosDb/Features/Operations/Reindex/CosmosBackgroundJobThrottleController.cs
```diff
@@ -104,9 +120,10 @@ public double UpdateDatastoreUsage()
             if (_initialized)
             {
                 var requestContext = _fhirRequestContextAccessor.RequestContext;
-                Debug.Assert(
-                    requestContext.Method.Equals(OperationsConstants.Reindex, StringComparison.OrdinalIgnoreCase),
-                    "We should not be here with FhirRequestContext that is not reindex!");
+
+                // Debug.Assert(
+                //    requestContext.Method.Equals(OperationsConstants.Reindex, StringComparison.OrdinalIgnoreCase),
+                //    "We should not be here with FhirRequestContext that is not reindex!");
```
brendankowitz, line 126: nit: should this still be commented?

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Delete/DeletionService.cs
```diff
@@ -51,20 +55,22 @@ public class DeletionService : IDeletionService
         public DeletionService(
             IResourceWrapperFactory resourceWrapperFactory,
             Lazy<IConformanceProvider> conformanceProvider,
-            IFhirDataStore fhirDataStore,
-            ISearchService searchService,
+            Func<IScoped<IFhirDataStore>> fhirDataStoreFactory,
+            Func<IScoped<ISearchService>> searchServiceFactory,
```
brendankowitz, line 59: This should work to resolve a new instance, but every invoke leaves a floating IScope<> object reference in the root container

src/Microsoft.Health.Fhir.Core.UnitTests/SubstituteHelpers.cs
LTA-Thinking: Darn it, I thought we already had an extension for this but I couldn't find it.

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Delete/DeletionService.cs
LTA-Thinking: I'm working to review your PR so that it shouldn't be an issue anymore.

src/Microsoft.Health.Fhir.CosmosDb/Features/Operations/Reindex/CosmosBackgroundJobThrottleController.cs
feordin: @brendankowitz I think you are correct that with MI we would need a control pane role to access this information, at least a read-only role.  I cannot remember the details with this method in PaaS, if our Comsos DB access key allowed permission to read  this information.

LTA-Thinking: Is this something that isn't working right now? This code is what was already there, I just moved it around.

LTA-Thinking: I've reverted all the throttling changes. The deletion service already has a retry on the delete, but it doesn't look like there is any retry on the search at any level. Due to the high usage I've seen for this service I'm going to increase the backoff for the retry in the deletion service.

feordin (approved):

PR #3663, opened 2024-01-11
Enable use of read only replicas

SergeyGaluzo:
Addresses https://microsofthealth.visualstudio.com/Health/_workitems/edit/113795/


src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetryService.cs
```diff
@@ -383,5 +383,102 @@ public async Task TryLogEvent(string process, string status, string text, DateTi
...
+                    {
+                        conn = await sqlConnectionBuilder.GetSqlConnectionAsync(initialCatalog: null, cancellationToken: cancel).ConfigureAwait(false);
+                    }
+                    else if (replicaTrafficRatio == 1)
+                    {
+                        conn = await sqlConnectionBuilder.GetReadOnlySqlConnectionAsync(initialCatalog: null, cancellationToken: cancel).ConfigureAwait(false);
+                    }
+                    else
+                    {
+                        var useWriteConnection = Interlocked.Increment(ref _usageCounter) % (int)(1 / (1 - _replicaTrafficRatio)) == 1; // examples for ratio -> % divider = { 0.9 -> 10, 0.8 -> 5, 0.75 - 4, 0.67 - 3, 0.5 -> 2, <0.5 -> 1}
```
brendankowitz, line 430: Should we consider: 

```suggestion
                        var useWriteConnection = unchecked(Interlocked.Increment(ref _usageCounter)) % (int)(1 / (1 - _replicaTrafficRatio)) == 1; // examples for ratio -> % divider = { 0.9 -> 10, 0.8 -> 5, 0.75 - 4, 0.67 - 3, 0.5 -> 2, <0.5 -> 1}
```
Assuming we might not overflow with long, but this should reset to 0 if we do.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetryService.cs
```diff
@@ -383,5 +383,102 @@ public async Task TryLogEvent(string process, string status, string text, DateTi
...
+                // must be set before opening connection to take effect. Therefore we must reset it to null here before opening the connection.
+                conn.RetryLogicProvider = null; // To remove this line _sqlConnectionBuilder in healthcare-shared-components must be modified.
+                await conn.OpenAsync(cancel);
+
+                return conn;
+            }
+
+            private double GetReplicaTrafficRatio(ISqlConnectionBuilder sqlConnectionBuilder)
+            {
+                if (_lastUpdated.HasValue && (DateTime.UtcNow - _lastUpdated.Value).TotalSeconds < 600)
```
brendankowitz, line 447: Should 600 be a named const? Looks like a TrafficRatioCacheDuration?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetryService.cs
SergeyGaluzo: Will do.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetryService.cs
```diff
@@ -383,5 +383,102 @@ public async Task TryLogEvent(string process, string status, string text, DateTi
...
+                    conn.Open();
+                    using var cmd = new SqlCommand("IF object_id('dbo.Parameters') IS NOT NULL SELECT Number FROM dbo.Parameters WHERE Id = 'ReplicaTrafficRatio'", conn);
+                    var value = cmd.ExecuteScalar();
+                    return value == null ? 0 : (double)value;
+                }
+                catch (SqlException)
+                {
+                    return 0;
+                }
+            }
```
brendankowitz, line 481: Should we have a ParametersDataStore or something similar to standardize access to this table?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetryService.cs
SergeyGaluzo: Fixed

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetryService.cs
SergeyGaluzo: I am not sure. We can consider specific tables if you feel they provide significant benefit. In any rate, I would suggest to do it as a separate work item.

tools/EventsReader/SqlConnectionBuilder.cs
```diff
@@ -29,5 +29,13 @@ public async Task<SqlConnection> GetSqlConnectionAsync(string initialCatalog = n
             await Task.CompletedTask;
             return new SqlConnection(_connectionString);
         }
+
+        public async Task<SqlConnection> GetReadOnlySqlConnectionAsync(string initialCatalog = null, int? maxPoolSize = null, CancellationToken cancellationToken = default)
+        {
+            await Task.CompletedTask;
+            var builder = new SqlConnectionStringBuilder(_connectionString);
+            builder.ApplicationIntent = ApplicationIntent.ReadOnly;
+            return new SqlConnection(builder.ToString());
```
LTA-Thinking, line 38: You can use a Task.FromResult instead of having the await Task.CompletedTask.

SergeyGaluzo: It should be functionally identical. I tested this code, and suggest to leave it as is.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetryService.cs
```diff
@@ -383,5 +383,103 @@ public async Task TryLogEvent(string process, string status, string text, DateTi
...
+
+            private static double GetReplicaTrafficRatioFromDatabase(ISqlConnectionBuilder sqlConnectionBuilder)
+            {
+                try
+                {
+                    using var conn = sqlConnectionBuilder.GetSqlConnection();
+                    conn.RetryLogicProvider = null;
+                    conn.Open();
+                    using var cmd = new SqlCommand("IF object_id('dbo.Parameters') IS NOT NULL SELECT Number FROM dbo.Parameters WHERE Id = 'ReplicaTrafficRatio'", conn);
```
LTA-Thinking, line 474: Do we have anywhere we are documenting these Parameter table values? There are now a few flags we can control from there and I think we should document them in a common location.

LTA-Thinking (approved):

PR #3667, opened 2024-01-11
Extend the max parallelism flag

fhibf:
## Description
* Renaming the HTTP flag "x-conditionalquery-processing-logic" to a more generic name. 
* Extending the use of the max parallelism flag to more (GET) scenarios. 

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/KnownHeaders.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/KnownHeaders.cs
@@ -31,7 +31,7 @@ namespace Microsoft.Health.Fhir.Core.Features
         public const string CustomAuditHeaderPrefix = "X-MS-AZUREFHIR-AUDIT-";
         public const string FhirUserHeader = "x-ms-fhiruser";
 
-        // #conditionalQueryParallelism - Header used to activate parallel conditional-query processing.
+        // #maxQueryParallelism - Header used to activate parallel conditional-query processing.
         public const string QueryProcessingLogic = "x-ms-query-processing-logic";
     }
 }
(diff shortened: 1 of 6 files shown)


src/Microsoft.Health.Fhir.Shared.Api/Controllers/FhirController.cs
```diff
@@ -272,6 +272,8 @@ private FhirResult ToSaveOutcomeResult(SaveOutcome saveOutcome)
         [AuditEventType(AuditEventSubType.Read)]
         public async Task<IActionResult> Read(string typeParameter, string idParameter)
         {
+            SetupRequestContextWithConditionalQueryMaxParallelism();
```
brendankowitz, line 275: At this point it should probably be a filter

src/Microsoft.Health.Fhir.Core/Features/KnownHeaders.cs
```diff
@@ -31,7 +31,7 @@ public static class KnownHeaders
         public const string CustomAuditHeaderPrefix = "X-MS-AZUREFHIR-AUDIT-";
         public const string FhirUserHeader = "x-ms-fhiruser";
 
-        // #conditionalQueryParallelism - Header used to activate parallel conditional-query processing.
-        public const string ConditionalQueryProcessingLogic = "x-conditionalquery-processing-logic";
+        // #maxQueryParallelism - Header used to activate parallel conditional-query processing.
+        public const string QueryProcessingLogic = "x-ms-query-processing-logic";
```
brendankowitz, line 35: Its a conventions thats hard to break :D but with the link above: https://datatracker.ietf.org/doc/html/rfc6648

I know the naming is supposed to be generic, but could it be something like "x-ms-optimize-query-logic"?

Really, the purpose is to try to reduce latency at the cost of RUs... `x-ms-query-latency-over-efficiency`?

PR #3683, opened 2024-01-22
[Dock Template] Application Insights Workpace and S360 tag

fhibf:
## Description

## Related issues
Addresses [issue #].

## Testing

## FHIR Team Checklist

### DIFF
--- a/samples/templates/default-azuredeploy-docker.json
+++ b/samples/templates/default-azuredeploy-docker.json
@@ -242,6 +242,7 @@
         "appServiceResourceId": "[resourceId('Microsoft.Web/sites', variables('serviceName'))]",
         "securityAuthenticationEnabled": "[and(not(empty(parameters('securityAuthenticationAuthority'))),not(empty(parameters('securityAuthenticationAudience'))))]",
         "deployAppInsights": "[parameters('deployApplicationInsights')]",
+        "appInsightsWorkspaceName": "[concat('AppInsights-', variables('serviceName'), '-ws')]",
         "appInsightsName": "[concat('AppInsights-', variables('serviceName'))]",
         "storageBlobDataContributerRoleId": "[concat('/subscriptions/', subscription().subscriptionId, '/providers/Microsoft.Authorization/roleDefinitions/', 'ba92f5b4-2d11-453d-a403-e96b0029c9fe')]",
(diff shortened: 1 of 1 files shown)


samples/templates/default-azuredeploy-docker.json
```diff
@@ -404,20 +405,34 @@
...
+            "tags": {
+                "Privacy.Asset.NonPersonal": ""
+            }
+        },
+        {
             "type": "Microsoft.Insights/components",
+            "apiVersion": "2020-02-02",
+            "name": "[variables('appInsightsName')]",
             "location": "[parameters('applicationInsightsLocation')]",
             "condition": "[variables('deployAppInsights')]",
```
LTA-Thinking, line 421: The workspace should also depend on this condition.

LTA-Thinking (approved):

samples/templates/default-azuredeploy-docker.json
```diff
@@ -404,20 +405,35 @@
...
+            "dependsOn": [
+                "[resourceId('Microsoft.OperationalInsights/workspaces', variables('appInsightsWorkspaceName'))]"
+            ],
             "kind": "web",
             "tags": {
                 "[concat('hidden-link:', variables('appServiceResourceId'))]": "Resource",
                 "displayName": "AppInsightsComponent",
-                "FhirServerSolution": "[parameters('solutionType')]"
+                "FhirServerSolution": "[parameters('solutionType')]",
+                "Privacy.Asset.NonPersonal": ""
```
brendankowitz, line 431: is this needed in OSS?

fhibf: When this script is executed, it creates App Insights resources that can trigger S360 alerts in our subscriptions. 
To minimize the number of S360 alerts I thought about adding this tag during the creation of these App Insights.

mikaelweave: I don't think this tag is correct for all OSS deployments. We should check w/ Joyce, but my understanding is If you deploy a test OSS solution and you login with your IP address, upn, etc that is still personal data even if it's only employee data.

I would recommend exposing an "additionalTags" variable that is an object and combining it with this tag object. I've seen this done in the past

PR #3693, opened 2024-01-29
[Jobs] Extending existing logs

fhibf:
* Extending existing logs

## FHIR Team Checklist


src/Microsoft.Health.Fhir.SqlServer/Features/Operations/JobLoggerExtension.cs
```diff
@@ -0,0 +1,107 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime;
+using Microsoft.Extensions.Logging;
+using Microsoft.Health.Fhir.SqlServer.Features.Operations;
+using Microsoft.Health.JobManagement;
+
+namespace Microsoft.Health.Fhir.SqlServer.Features.Operations
```
mikaelweave, line 14: Could this be in a common area where CosmosDB jobs could use it too?

mikaelweave (commented):
One specific question with the code and one general question - when querying these logs in the future will we be able to query by a column (like JobId) in the `Logs` table or will this information be in the message?

src/Microsoft.Health.Fhir.SqlServer/Features/Operations/JobLoggerExtension.cs
fhibf: That would be applicable for the old jobs, is that the intension?

mikaelweave: Here is an example on the *new* job framework that occurs on CosmosDB. 

https://github.com/microsoft/fhir-server/blob/53f5153126e604573441c61a82b02d713b9c6ebe/src/Microsoft.Health.Fhir.CosmosDb/Features/Operations/Export/CosmosExportOrchestratorJob.cs#L25

mikaelweave (approved):

PR #3696, opened 2024-01-31
Removing dead code from $import

SergeyGaluzo:


src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/ImportOrchestratorJob.cs
```diff
@@ -109,34 +105,11 @@ public async Task<string> ExecuteAsync(JobInfo jobInfo, IProgress<string> progre
...
-
-                if (currentResult.Progress == ImportOrchestratorJobProgress.PreprocessCompleted)
-                {
-                    await ExecuteImportProcessingJobAsync(progress, jobInfo, inputData, currentResult, cancellationToken);
-                    currentResult.Progress = ImportOrchestratorJobProgress.SubJobsCompleted;
-                    progress.Report(JsonConvert.SerializeObject(currentResult));
-
-                    _logger.LogJobInformation(jobInfo, "SubJobs Completed.");
-                }
+                await ExecuteImportProcessingJobAsync(jobInfo, inputData, currentResult, cancellationToken);
```
mikaelweave, line 111: Since the orchestrator is long running, what happens if the job gets restarted (either through a crash, deployment, etc)? Will the import start again and queue up duplicate child jobs?

src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/ImportOrchestratorJob.cs
```diff
@@ -109,34 +105,11 @@ public async Task<string> ExecuteAsync(JobInfo jobInfo, IProgress<string> progre
...
-
-                    currentResult.Progress = ImportOrchestratorJobProgress.InputResourcesValidated;
-                    progress.Report(JsonConvert.SerializeObject(currentResult));
-
-                    _logger.LogJobInformation(jobInfo, "Input Resources Validated.");
-                }
-
-                if (currentResult.Progress == ImportOrchestratorJobProgress.InputResourcesValidated)
-                {
-                    await _importOrchestratorJobDataStoreOperation.PreprocessAsync(cancellationToken);
```
mikaelweave, line 124: Are we no longer disabling indexing for initial load $import??

mikaelweave (commented):
A couple questions about the core idea of the PR.

src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/ImportOrchestratorJob.cs
SergeyGaluzo: Import orchestrator will be restarted. It will try enqueue the same jobs, but because jobs definitions are identical to previous ones, jobs will not be enqueued, but old ones will be returned back.

src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/ImportOrchestratorJob.cs
SergeyGaluzo: Yes. We are not using it in PROD for multiple reasons, so I remove it.

src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/ImportOrchestratorJob.cs
mikaelweave: Sick - that's awesome. So if our orchestrators produce the same result, we don't need all this progress work. One minor suggestion would be to add a comment for those unfamiliar with the code, but not required at all.

src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/ImportOrchestratorJob.cs
mikaelweave: This makes the future `$import` orchestrator changes much easier.

src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/ImportOrchestratorJob.cs
SergeyGaluzo: If add, then it is per each queue type. Too many places...

mikaelweave (dismissed):
Love how this simplifies the orchestrator. Nice work.

brendankowitz (approved):

PR #3708, opened 2024-02-12
Tools changes to support bundles test

SergeyGaluzo:


tools/BlobRewriter/BlobRewriter.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <AssemblyName>Microsoft.Health.Internal.Fhir.BlobRewriter</AssemblyName>
+    <AssemblyName>Fhir.BlobRewriter</AssemblyName>
```
fhibf, line 5: May we keep the original assembly name?

tools/EventsReader/EventsReader.csproj
```diff
@@ -2,7 +2,7 @@
   
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <AssemblyName>Microsoft.Health.Internal.Fhir.EventsReader</AssemblyName>
+    <AssemblyName>Fhir.EventsReader</AssemblyName>
```
fhibf: May we keep the original assembly name?

tools/Exporter/Exporter.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <AssemblyName>Microsoft.Health.Internal.Fhir.Exporter</AssemblyName>
+    <AssemblyName>Fhir.Exporter</AssemblyName>
```
fhibf: The same here.

tools/Importer/BearerTokenHandler.cs
```diff
@@ -150,5 +178,23 @@ public void Dispose(bool disposing)
 
             _disposed = true;
         }
+
+        private static void PrintClaimsFromJwt(string jwtToken)
```
fhibf, line 182: Do we need to print this information aways? Can we have a flag to enable this print or not. Just thinking that we can expose someone's information if that person decides to use this tool.

SergeyGaluzo: It is very handy for test tool. Yes.

tools/BlobRewriter/BlobRewriter.csproj
SergeyGaluzo: It is very inconvenient to have long exe name with all not distinct letters in front. This way I don't need to type much.

tools/Exporter/Exporter.csproj
SergeyGaluzo: Keeping original because name is referenced.

PR #3720, opened 2024-02-19
Import bundles

SergeyGaluzo:
To avoid confusion...

Currently the only API to ingest the set of resources in real time is to POST Bundle. This API has performance and stability issues, which prompted this new work. 

My goal was to extend efficient $import logic to real time calls. We should not, at least now, add any extra logic beyond $import, which might have performance impact.

$import uses ndjson as input format. Very efficient. Hence new endpoint should use this format. There is a demand, for customer convenience, support bundle format, so it was enabled too. By no means, this endpoint is a replacement of current generic bundles processing logic. 

We should avoid inheriting any complexity allowed by generic bundle format. For example, $import logic (so new endpoint logic) requires resource id. We should avoid attempts to extend functionality by enabling identity search to identify resource. Maybe in future, but not now.

To avoid confusion (added because of bundle format) about what functionality is supported, I suggest refraining from using bundle in endpoint name. I suggest clarifying intent in the documentation. 

This endpoint will process batch in a single database call, hence it inherently equivalent to the transaction bundle. It also either PUTs all input resources or none. This makes output response trivial (no need to have complex output logic generic bundles have).

When using bundle format, it would be nice to ignore all bundle components except Resource. We should be clear about it in the documentation. BTW I am not sure that we need to make documentation change now, we can wait till first customer provides feedback.

https://microsofthealth.visualstudio.com/Health/_workitems/edit/117904


src/Microsoft.Health.Fhir.Core/Messages/BulkImport/ImportBundleRequest.cs
```diff
@@ -0,0 +1,28 @@
...
+using Microsoft.Health.Fhir.Core.Features.Operations.Import.Models;
+using Microsoft.Health.Fhir.Core.Features.Persistence;
+
+namespace Microsoft.Health.Fhir.Core.Messages.Import
+{
+    public class ImportBundleRequest : IRequest<ImportBundleResponse>
+    {
+        public ImportBundleRequest(IList<ImportResource> resources)
+        {
+            Resources = resources;
```
fhibf, line 20: We should add a validation and check if resources are not null.

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Controllers/ImportControllerTests.cs
```diff
@@ -100,6 +100,7 @@ private ImportController GetController(ImportTaskConfiguration bulkImportConfig)
                 _urlResolver,
                 optionsOperationConfiguration,
                 optionsFeatures,
+                null,
```
fhibf, line 103: We can replace null with Substitute.For<IResourceWrapperFactory resourceWrapperFactory>()

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ImportController.cs
```diff
@@ -79,9 +93,80 @@ public class ImportController : Controller
...
+                        {
+                            throw new RequestNotValidException(string.Format(Resources.ResourcesMustBeUnique, key));
+                        }
+
+                        resources.Add(importResource);
+                        index++;
+                    }
+                }
+            }
+            catch (Exception)
```
fhibf, line 156: This exception is too broad. 
If any type of exception is thrown, then it'll return a "Unable to parse resource at index=", given that it's not necessarily and parsing issue.

fhibf: Per example, when a RequestNotValidException is thrown with a ResourcesMustBeUnique or an InvalidBundleEntry, the details of those exceptions are lost.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ImportController.cs
```diff
@@ -79,9 +93,80 @@ public class ImportController : Controller
...
+            {
+                throw new RequestNotValidException(string.Format(Resources.ParsingError, $"Unable to parse resource at index={index}"));
+            }
+
+            var request = new ImportBundleRequest(resources);
+            var response = await _mediator.ImportBundleAsync(request.Resources, HttpContext.RequestAborted);
```
fhibf, line 162: How are we handling exceptions here? If a SQL exception is thrown, are we exposing the exception and all details on it?

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ImportController.cs
```diff
@@ -79,9 +93,80 @@ public class ImportController : Controller
...
+            var result = new ImportBundleResult(response.LoadedResources, HttpStatusCode.OK);
+            result.Headers["LoadedResources"] = response.LoadedResources.ToString();
+            await _mediator.Publish(new ImportBundleMetricsNotification(startDate, DateTime.UtcNow, response.LoadedResources), CancellationToken.None);
```
fhibf, line 165: For regular bundles we expose metrics with the results of all API calls. 
We should do the same here.

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportBundleRequestHandler.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        public async Task<ImportBundleResponse> Handle(ImportBundleRequest request, CancellationToken cancellationToken)
+        {
+            EnsureArg.IsNotNull(request, nameof(request));
+            if (await _authorizationService.CheckAccess(DataActions.Write, cancellationToken) != DataActions.Write)
+            {
+                throw new UnauthorizedFhirActionException();
+            }
+
+            var input = request.Resources.Select(_ => new ResourceWrapperOperation(_.ResourceWrapper, true, false, null, false, false, null)).ToList();
```
fhibf, line 51: Are we handling or rejecting conditional PUTs in this context?
If a customer sends a request to update a resource with the following information, I believe we should handle it properly. 

```
            "request": {
                "method": "PUT",
                "url": "Patient?identifier=|11788759296812"
            }
```

src/Microsoft.Health.Fhir.Core/Messages/BulkImport/ImportBundleResponse.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        public ImportBundleResponse(int count)
+        {
+            LoadedResources = count;
+        }
+
+        /// <summary>
+        /// Number of loaded resources
+        /// </summary>
+        public int LoadedResources { get; }
```
fhibf, line 18: We should discuss about the expected outcomes of these bundles.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ImportController.cs
```diff
@@ -79,9 +93,80 @@ public class ImportController : Controller
...
+            _resourceWrapperFactory = resourceWrapperFactory;
             _logger = logger;
         }
 
+        [HttpPost]
+        [Route(KnownRoutes.ImportBundle)]
+        [AuditEventType(AuditEventSubType.ImportBundle)]
+        public async Task<IActionResult> ImportBundle()
+        {
+            var sw = Stopwatch.StartNew();
```
fhibf, line 105: We are not considering X-Provenance.

src/Microsoft.Health.Fhir.Core/Messages/BulkImport/ImportBundleRequest.cs
SergeyGaluzo: This is not possible in current code. Why do we need to do this?

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ImportController.cs
```diff
@@ -79,9 +93,80 @@ public class ImportController : Controller
             _urlResolver = urlResolver;
             _features = features.Value;
             _mediator = mediator;
+            _resourceWrapperFactory = resourceWrapperFactory;
```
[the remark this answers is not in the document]
SergeyGaluzo: There is a test that would fail if wd do this.

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Controllers/ImportControllerTests.cs
SergeyGaluzo: Good idea

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ImportController.cs
SergeyGaluzo: What exception type do you suggest?

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ImportController.cs
SergeyGaluzo: I am using default logic. I don't think we expose details to the client in it.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ImportController.cs
SergeyGaluzo: There is single API call. Single transaction in SQL. All valid resources are loaded by definition. What is the problem?

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportBundleRequestHandler.cs
SergeyGaluzo: Current code does not support conditional PUTs. I do not plan to have generic logic.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ImportController.cs
SergeyGaluzo: Please elaborate.

SergeyGaluzo:
> I believe that before merging this new endpoint, we should discuss the boundaries of this new ingestion.
> 
> Customers relying on Bundles today are already expecting the same bundle response structure, operation outcomes, errors handling, conditional operations, POST, Deletes, etc.

I intentionally created separate endpoint to make sure that customers are making cautious decision when they start using it. This endpoint is not a generic bundle processing endpoint, so there should not be expectations that input and output are the same.

src/Microsoft.Health.Fhir.Core/Messages/BulkImport/ImportBundleRequest.cs
SergeyGaluzo: Done.

src/Microsoft.Health.Fhir.Core/Messages/BulkImport/ImportBundleResponse.cs
SergeyGaluzo: Not sure what you mean. If all resources are valid, all resources are loaded, either updated, or created. Therefore outcome is simpler than for generic bundle.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ImportController.cs
```diff
@@ -79,9 +93,80 @@ public class ImportController : Controller
...
+                        resources.Add(importResource);
+                        line = await reader.ReadLineAsync();
+                        index++;
+                    }
+                }
+                else
+                {
+                    var str = await reader.ReadToEndAsync();
+                    var resource = await fhirParser.ParseAsync(str);
+                    var bundle = resource.ToResourceElement().ToPoco<Bundle>();
```
brendankowitz, line 136: This bypasses the MVC InputFormatters - https://github.com/microsoft/fhir-server/blob/main/src/Microsoft.Health.Fhir.Shared.Api/Features/Formatters/FhirJsonInputFormatter.cs. 
I think this should be two different Action methods that delineate by media-type instead of trying to parse everything

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ImportController.cs
```diff
@@ -79,9 +93,80 @@ public class ImportController : Controller
...
+            var keys = new HashSet<ResourceKey>();
+            var fhirParser = new FhirJsonParser();
+            var importParser = new ImportResourceParser(fhirParser, _resourceWrapperFactory);
+            var index = 0L;
+            using var reader = new StreamReader(Request.Body, Encoding.UTF8);
+            try
+            {
+                if (Request.ContentType != null && Request.ContentType.Contains("application/fhir+ndjson", StringComparison.OrdinalIgnoreCase))
+                {
+                    var line = await reader.ReadLineAsync();
```
brendankowitz, line 117: What is the stream is really large, should we have some validation or limits?

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ImportController.cs
SergeyGaluzo: I changed the exception. Please keep in mind that current logic is "fail fast", it exits on first found problem. This is intentional.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ImportController.cs
```diff
@@ -79,9 +93,80 @@ public class ImportController : Controller
...
+                        line = await reader.ReadLineAsync();
+                        index++;
+                    }
+                }
+                else
+                {
+                    var str = await reader.ReadToEndAsync();
+                    var resource = await fhirParser.ParseAsync(str);
+                    var bundle = resource.ToResourceElement().ToPoco<Bundle>();
+                    foreach (var entry in bundle.Entry)
```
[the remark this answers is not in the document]
SergeyGaluzo: I am trying to avoid deep validation. Instead, we validate Resource only, and ignore the rest. We should be clear about it.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ImportController.cs
SergeyGaluzo: I will make the change.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ImportController.cs
SergeyGaluzo: If I understand correctly, it would require 2 endpoints, something like: $import/ndjson and $import/batch. I am not a fan of this idea. I will try implementing checks for stream size you mentioned below.

brendankowitz: MVC natively supports routing to the correct action based on many factors such as URL, method, headers etc. The url can be the same, I'm more referring to something like https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.consumesattribute?view=aspnetcore-8.0, where for the same action, one can consume fhir+json, while the other consumes ndjson

SergeyGaluzo: Main consumption is ndjson. We would need to put restrictions on size there. Will default MVC restrictions work? Will they limit content size to something that our service will not choke on?

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ImportController.cs
```diff
@@ -79,9 +93,80 @@ public class ImportController : Controller
             _urlResolver = urlResolver;
             _features = features.Value;
             _mediator = mediator;
+            _resourceWrapperFactory = EnsureArg.IsNotNull(resourceWrapperFactory, nameof(resourceWrapperFactory));
             _logger = logger;
         }
 
+        [HttpPost]
+        [Route(KnownRoutes.ImportBundle)]
```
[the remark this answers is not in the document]
SergeyGaluzo: I will update once we pass the discussion point.

PR #3723, opened 2024-02-22
Add content-type constraint for the $convert endpoint.

ms-teli:
## Description
Add content-type constraint for the $convert endpoint.

## Related issues
We have found that customers can use other MIME types on the payload which can result in parsing errors.

## Testing
Tested locally, the FHIR server started successfully and the $convert endpoint only accepts application/json after the code change.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Api/Controllers/ConvertDataController.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Api/Controllers/ConvertDataController.cs
@@ -29,6 +29,7 @@ namespace Microsoft.Health.Fhir.Api.Controllers
 {
     [ServiceFilter(typeof(AuditLoggingFilterAttribute))]
     [ServiceFilter(typeof(OperationOutcomeExceptionFilterAttribute))]
+    [Consumes("application/json")]
     [ValidateResourceTypeFilter]
     [ValidateModelState]
     public class ConvertDataController : Controller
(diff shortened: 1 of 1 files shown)


src/Microsoft.Health.Fhir.Shared.Api/Controllers/ConvertDataController.cs
```diff
@@ -29,6 +29,7 @@ namespace Microsoft.Health.Fhir.Api.Controllers
 {
     [ServiceFilter(typeof(AuditLoggingFilterAttribute))]
     [ServiceFilter(typeof(OperationOutcomeExceptionFilterAttribute))]
+    [Consumes("application/json")]
```
brendankowitz, line 32: Does this work with `application/fhir+json` too?

We have an attribute that goes through a lot of the checks with `[ServiceFilter(typeof(ValidateFormatParametersAttribute))]`

ms-teli: > Does this work with `application/fhir+json` too?
> 
> We have an attribute that goes through a lot of the checks with `[ServiceFilter(typeof(ValidateFormatParametersAttribute))]`

Good point. It actually should accept both  "application/fhir+json" and "application/json" conforming this spec: https://hl7.org/fhir/parameters.html

It would be a good idea to add [ServiceFilter(typeof(ValidateFormatParametersAttribute))], but if only adding that would not mitigate this specific issue. The .net core parser gets hit first. Any idea?

brendankowitz (approved):

PR #3729, opened 2024-02-28
[Bugfix] HeaderDictionary - Support concurrent operations

fhibf:
## Description
It was noticed that per day, four or five requests fail due attempts to access/modify the same HTTP header at the same time. 
As a way to avoid this issue happening, we are now simplifying the access to the HeaderDictionary and switching to concurrent dictionaries in order to avoid concurrent access to the same resources. 

[AB#117368](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/117368)


mikaelweave (dismissed):
🦾

mikaelweave (approved):
🦾🦾

PR #3747, opened 2024-03-12
[Tests] Refactoring test accounts

fhibf:
## Description
Refactoring tests to use Service Principals instead of User Accounts. 

[AB#117408](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/117408)

## Testing
Changes in tests were tested using our pipelines.

## FHIR Team Checklist


mikaelweave (changes requested):
One area of the `Add-AadTestAuthEnvironment.ps1` script seems  incorect or the comments are just confusing me. Can you look at this area and ensure it's correct? If so, update the comment a bit more to make it clearer? Thanks!

feordin:
I just wanted to double check that we will already have, or be able to create, all these apps for each of the environments that we run the tests: OSS, dogfood, INT, PROD, GOV.

fhibf:
> I just wanted to double check that we will already have, or be able to create, all these apps for each of the environments that we run the tests: OSS, dogfood, INT, PROD, GOV.

@feordin , I believe we'll be able to create these accounts in other regions as well. 

The use of service principals is not something new and in this PR we are expanding it to more scenarios. 
For this reason, I believe that running these tests in other environments will be fine. 

Probably, we need to double-check, some adjusts in other repositories scripts will be required to ensure these accounts are created properly.

feordin (approved):

mikaelweave (approved):
🦾

PR #3749, opened 2024-03-13
Move import writes to store class

SergeyGaluzo:
This is refactoring only, no functional changes. This should facilitate honoring input last updated and version. 

https://microsofthealth.visualstudio.com/Health/_workitems/edit/117398/

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/SqlImporter.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/SqlImporter.cs
@@ -53,7 +53,6 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Operations.Import
                 await _model.EnsureInitialized();
 
                 long succeededCount = 0;
-                long failedCount = 0;
                 long processedBytes = 0;
                 long currentIndex = -1;
                 var importErrorBuffer = new List<string>();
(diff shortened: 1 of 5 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/SqlImporter.cs
```diff
@@ -85,135 +85,21 @@ await foreach (ImportResource resource in inputChannel.Reader.ReadAllAsync(cance
...
-            var dups = resources.Where(r => string.IsNullOrEmpty(r.ImportError)).Except(loaded).Except(conflicts);
-            AppendErrorsToBuffer(dups, conflicts, errors);
-
-            succeededCount += loaded.Count;
-            failedCount += errorResources.Count() + dups.Count() + conflicts.Count;
+            errors.AddRange(resources.Where(r => !string.IsNullOrEmpty(r.ImportError)).Select(r => r.ImportError));
+            failedCount += errors.Count;
+            //// exclude resources with parsing error (ImportError != null)
+            var validResources = resources.Where(r => string.IsNullOrEmpty(r.ImportError)).ToList();
+            var results = _store.ImportResourcesAsync(validResources, importMode, cancellationToken).Result;
```
PTaladay, line 92: Calling "Result" here will most likely run into a deadlock at some point. It isn't recommended to do this. I realize in the previous version of the code it was calling "Wait" and not returning any values.  I would recommend making the calling method async and follow that up through the code to ensure it doesn't run into issues. If you call "Result" here without any kind of wait there is a high chance it won't have the method call complete, so your results will be inconsistent.

Syntactically the old code and this refactor look similar functionally, but they will behave differently when ran.

SergeyGaluzo: OK. Fixed.

PTaladay (approved):

PR #3753, opened 2024-03-14
[Metrics] FHIR metric emitter

fhibf:
## Description
FHIR metric emitter.

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Logging/DefaultMetricEmitter.cs
+++ b/src/Microsoft.Health.Fhir.Core/Logging/DefaultMetricEmitter.cs
@@ -1,0 +1,26 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Fhir.Core.Logging
+{
+    /// <summary>
+    /// <see cref="DefaultMetricEmitter"/> does not emit any metrics.
+    /// It's a default implementation to support the case where no metrics are required.
+    /// </summary>
+    public sealed class DefaultMetricEmitter : IFhirMetricEmitter
+    {
+        public void EmitBundleLatency(long latencyInMilliseconds)
+        {
+        }
+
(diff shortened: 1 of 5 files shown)


src/Microsoft.Health.Fhir.Core/Logging/Metrics/DefaultMetricEmitter.cs
```diff
@@ -0,0 +1,28 @@
...
+
+using Microsoft.Health.Fhir.Core.Logging.Metrics;
+
+namespace Microsoft.Health.Fhir.Core.Logging.Metrics
+{
+    /// <summary>
+    /// <see cref="DefaultMetricEmitter"/> does not emit any metrics.
+    /// It's a default implementation to support the case where no metrics are required.
+    /// </summary>
+    public sealed class DefaultMetricEmitter : IFhirMetricEmitter
```
brendankowitz, line 14: Would it makes sense to have a real metric implementation that works with OSS? https://learn.microsoft.com/en-us/dotnet/core/diagnostics/metrics-instrumentation#get-a-meter-via-dependency-injection

src/Microsoft.Health.Fhir.Core/Logging/Metrics/CrudMetricHandler.cs
```diff
@@ -0,0 +1,26 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using EnsureThat;
+
+namespace Microsoft.Health.Fhir.Core.Logging.Metrics
+{
+    public sealed class CrudMetricHandler
```
brendankowitz, line 10: Sorry to back track here, but I don't know if it makes sense to have both a handler and an empty IFhirMetricEmitter, I think we only should have of those, no need for both. I like the filter approach though, encapsulates the logic nicely.

src/Microsoft.Health.Fhir.Core/Logging/Metrics/DefaultMetricEmitter.cs
fhibf: Added real implementations using IMeterFactory.

brendankowitz (approved):

PR #3763, opened 2024-03-20
Import Bad Request Duplicate Files

mikaelweave:
## Description
Changes `$import` behavior to immediately return with bad request if duplicate files are submitted.

## Related issues
AB#1117694

## Testing
Included new unit test and E2E test.
Tested locally.

## FHIR Team Checklist


LTA-Thinking (dismissed):

SergeyGaluzo (changes requested):
:clock1:

src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/ImportOrchestratorJob.cs
```diff
@@ -412,7 +412,7 @@ private async Task<IList<long>> EnqueueProcessingJobsAsync(IEnumerable<InputReso
             {
                 const string message = "Duplicate file detected in list of files to import.";
                 _logger.LogJobError(ex, orchestratorInfo, message);
-                throw new JobExecutionException(message, ex);
+                throw new ImportProcessingException(message, ex);
```
SergeyGaluzo, line 415: > throw new ImportProcessingException(message, ex); [](http://example.com/codeflow?start=15&length=50)

What is the reason to register $import? Why not to return bad request from import controller before registration?

SergeyGaluzo:
>             catch (SqlException ex) when (ex.Number == 2627)

I think it is a wrong place to check whether input list is unique. We should return 400 from import controller without registration.

---
Refers to: src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/ImportOrchestratorJob.cs:411 in a78b1b7. [](commit_id = a78b1b7102b1eaa12467e588741cbbb10670dd6b, deletion_comment = False)

mikaelweave:
> > ```
> >         catch (SqlException ex) when (ex.Number == 2627)
> > ```
> 
> I think it is a wrong place to check whether input list is unique. We should return 400 from import controller without registration.
> 
> Refers to: src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/ImportOrchestratorJob.cs:411 in [a78b1b7](https://github.com/microsoft/fhir-server/commit/a78b1b7102b1eaa12467e588741cbbb10670dd6b). [](commit_id = [a78b1b7](https://github.com/microsoft/fhir-server/commit/a78b1b7102b1eaa12467e588741cbbb10670dd6b), deletion_comment = False)

Makes sense - should be a bad request from the start. I will add this to the controller and remove the check from the job itself.

SergeyGaluzo:
I think it is fixed by https://github.com/microsoft/fhir-server/pull/3796

mikaelweave:
> I think it is fixed by #3796

In general, yes you are correct. However I also implemented a check at the `POST /$import` level as suggested in the PR comments. This was not fixed by #3796. 😄

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ImportController.cs
```diff
@@ -198,6 +198,14 @@ private void ValidateImportRequestConfiguration(ImportRequest importData)
                 throw new RequestNotValidException(string.Format(Resources.ImportRequestValueNotValid, nameof(input)));
             }
 
+            var duplicateInputUrls = input.GroupBy(item => item.Url).Where(group => group.Count() > 1).Select(group => group.Key);
+
+            if (duplicateInputUrls.Any())
+            {
+                var duplicateUrlString = string.Join(", ", duplicateInputUrls);
+                throw new RequestNotValidException(string.Format(Resources.ImportRequestDuplicateInputFiles, duplicateUrlString));
+            }
```
SergeyGaluzo, line 207: Please remove dup exception logic from orchestrator.

mikaelweave: done thx for calling that out

SergeyGaluzo (approved):
:shipit:

LTA-Thinking (approved):

PR #3771, opened 2024-03-25
Prevent reporting dups on worker restart

SergeyGaluzo:
Add extra checks to prevent reporting dups. Fixes
https://microsofthealth.visualstudio.com/Health/_workitems/edit/118113

### DIFF
--- a/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Import/ImportTests.cs
+++ b/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Import/ImportTests.cs
@@ -334,24 +334,6 @@ namespace Microsoft.Health.Fhir.Tests.E2E.Rest.Import
             Assert.Equal(GetLastUpdated("2001"), result.Resource.Meta.LastUpdated); // version 1 imported
         }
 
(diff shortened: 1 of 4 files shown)


test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Import/ImportTests.cs
```diff
@@ -332,6 +334,94 @@ public async Task GivenIncrementalLoad_WhenOutOfOrder_ThenCurrentDatabaseVersion
             Assert.Equal(GetLastUpdated("2001"), result.Resource.Meta.LastUpdated); // version 1 imported
         }
 
+        [Fact(Skip = "requires new stored procedures")]
```
LTA-Thinking, line 337: Which new stored procedures? Is there a work item number that can be put here so we know when these can be enabled?

SergeyGaluzo: Not now. It depends on results of future discussion on whether behaviors in the skipped tests are desirable or not.

LTA-Thinking (approved):

LTA-Thinking (approved):

PR #3781, opened 2024-03-27
Add support for partial delete success

LTA-Thinking:
## Description
If Cosmos DB runs out of RUs while performing a delete operation it will roll back the transaction. This can lead to a situation where it is impossible to delete a resource because it has too much history to delete in one pass. This PR add the ability to optionally support partial delete success. This allows for the service to delete as much of a resource as it can, and leave the rest for a later delete operation. By rerunning the delete operation eventually the resource will be deleted.

## Related issues
Addresses [AB#118312](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/118312)

## Testing
Manually tested. Working on unit tests. E2E tests aren't possible as the scenario is hard to manufacture. 

## FHIR Team Checklist

Patch

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Exceptions/PartialDeleteSuccessException.cs
+++ b/src/Microsoft.Health.Fhir.Core/Exceptions/PartialDeleteSuccessException.cs
@@ -1,0 +1,22 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Health.Abstractions.Exceptions;
+
+namespace Microsoft.Health.Fhir.Core.Exceptions
+{
+    public class PartialDeleteSuccessException : MicrosoftHealthException
+    {
+        public PartialDeleteSuccessException(int numberOfResourceVersionsDeleted)
+            : base(Resources.PartialDeleteSuccess.Replace("{0}", numberOfResourceVersionsDeleted.ToString(), StringComparison.Ordinal))
+        {
+        }
+    }
+}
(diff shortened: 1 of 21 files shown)


src/Microsoft.Health.Fhir.Core/Exceptions/PartialDeleteSuccessException.cs
```diff
@@ -0,0 +1,22 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Health.Abstractions.Exceptions;
+
+namespace Microsoft.Health.Fhir.Core.Exceptions
+{
+    public class PartialDeleteSuccessException : MicrosoftHealthException
```
brendankowitz, line 15: In the context of an exception, I still prefer the terminology of "Incomplete" over "PartialSuccess" 🤔

src/Microsoft.Health.Fhir.Core/Exceptions/PartialDeleteSuccessException.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        public PartialDeleteSuccessException(int numberOfResourceVersionsDeleted)
+            : base(Resources.PartialDeleteSuccess.Replace("{0}", numberOfResourceVersionsDeleted.ToString(), StringComparison.Ordinal))
```
brendankowitz, line 18: Should this be using string.Format?

src/Microsoft.Health.Fhir.Core/Exceptions/PartialDeleteSuccessException.cs
LTA-Thinking: I like the name better, but I'm hesitant on Request Too Costly. We currently handle those with a 401, and I think 429 is the correct response for this. The request isn't bad, we just are unable to make it work in one pass.

LTA-Thinking: Switched and changed it to use code 413, request too large.

feordin (approved):

PR #3790, opened 2024-03-29
[DRAFT] Allow for "duplicate" search parameters

feordin:
## Description
There are published IG's with Search Parameters that have identical code values and expression as existing Search Parameters.  We want to allow these parameters without causing any reindexing or ambiguous behaviors when searching using the duplicated code values.

This PR also blocks the addition of any Search Parameter resource with a conflicting code value that has a different expression.  As this is a bug that currently puts the service into an unstable state.

## Related issues
Addresses [issue AB#117004].

## Testing
Some additional unit tests have been created, as well some manual testing.  More to come.

## FHIR Team Checklist


src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/SearchParameterOperations.cs
```diff
@@ -55,42 +60,51 @@ public class SearchParameterOperations : ISearchParameterOperations
...
-                (bool Supported, bool IsPartiallySupported) supportedResult = _searchParameterSupportResolver.IsSearchParameterSupported(searchParameterInfo);
+                    (bool Supported, bool IsPartiallySupported) supportedResult = _searchParameterSupportResolver.IsSearchParameterSupported(searchParameterInfo);
 
-                if (!supportedResult.Supported)
-                {
-                    throw new SearchParameterNotSupportedException(searchParameterInfo.Url);
-                }
+                    if (!supportedResult.Supported)
+                    {
+                        throw new SearchParameterNotSupportedException(searchParameterInfo.Url);
```
rajithaalurims, line 91: Customer sees this message "The search parameter with definition URL '{0}' is not supported.".  I know this change is not related to this PR but in this scenario, error is thrown if definition has any issues. Do you think we can use better error message to indicate customer that something is wrong with the definition they have sent?

PR #3813, opened 2024-04-12
New BundleHandler.RequestContexts must honor original requests CT

apurvabhaleMS:
## Description
Ensure new request contexts for BundleHandler are honoring original requests cancellation tokens.
If the request is timed-out due to client httprequest timed-out or due to request cancellation by client - FHIR managed service customers would still see 500 as they do today. FHIR service would behave as below -

For Transactions - (This is what we do today as well if the transaction is failed due to any reason)
- If any request failed with 408 (first, midway or last), log the statistics upto where we have collected them 
- Do not publish metric notification (No need to worry about the billing as metrics are not published and transaction is rolledback) 
- Final status of the transaction would be last failed operation status which is 408 in this case
 
For Batches -
- Statistics would have an information about how many requests actually failed due to time-out with status code 408 
- Publish the metric notification with count of passed + timed-out requests (check if we need to count 408 requests for billing or not)
- Final status of the batch is always 200

Note - 
- Statistics are always collected
- For batches, to provide the satistics of the request that failed with 408, we are not exactly stopping the creation of new context and invokation of sub request. If the cancellation is requested then the sub request would fail immediately as soon as it is invoked/called

## Related issues
Addresses [AB105685](https://microsofthealth.visualstudio.com/Health/_workitems/edit/105685)

## Testing

## FHIR Team Checklist


brendankowitz (approved):
:shipit:

src/Microsoft.Health.Fhir.Core/Features/Persistence/Orchestration/BundleOrchestratorOperation.cs
```diff
@@ -112,6 +112,7 @@ public async Task<UpsertOutcome> AppendResourceAsync(ResourceWrapperOperation re
             {
                 if (!_resources.Any())
                 {
+                    cancellationToken.ThrowIfCancellationRequested();
```
fhibf, line 115: ❤️❤️🤓🤓

fhibf (approved):

PR #3816, opened 2024-04-15
Prep work for schema update - Iteration 1

rajithaalurims:
## Description
This PR has code to support reading from Int32/Int64 for Version column on Resource table. This is to support backward compatibility for future schema update release. Since we cannot read same column twice with sequential access, this PR has change to remove sequential access on Readers.
Issue [AB#118978](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Olympus/Stories/?workitem=118978)
## Related issues
Addresses [issue [118978]([url](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Olympus/Stories/?workitem=118978))].

## Testing
Tested POST/PUT/Import/Export with Resource table Version int and bigint.

## FHIR Team Checklist


src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -360,7 +360,7 @@ private async Task<SearchResult> SearchImpl(SqlSearchOptions sqlSearchOptions, s
 
                         LogSqlCommand(sqlCommand);
 
-                        using (var reader = await sqlCommand.ExecuteReaderAsync(CommandBehavior.SequentialAccess, cancellationToken))
+                        using (var reader = await sqlCommand.ExecuteReaderAsync(cancellationToken))
```
SergeyGaluzo, line 363: >using (var reader = await sqlCommand.ExecuteReaderAsync(cancellationToken)) [](http://example.com/codeflow?start=24&length=75)

Please put comment to revert it back

rajithaalurims: added

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlStoreClient.cs
```diff
@@ -130,7 +130,19 @@ private static ResourceWrapper ReadResourceWrapper(SqlDataReader reader, bool re
             var resourceTypeId = reader.Read(VLatest.Resource.ResourceTypeId, 0);
             var resourceId = reader.Read(VLatest.Resource.ResourceId, 1);
             var resourceSurrogateId = reader.Read(VLatest.Resource.ResourceSurrogateId, 2);
-            var version = reader.Read(VLatest.Resource.Version, 3);
+            int version;
+            bool versionIsLong = true;
```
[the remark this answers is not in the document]
rajithaalurims: Yes. Done

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -807,7 +807,18 @@ private SqlSearchOptions UpdateSort(SqlSearchOptions searchOptions, Expression s
         {
             resourceTypeId = reader.Read(VLatest.Resource.ResourceTypeId, 0);
             resourceId = reader.Read(VLatest.Resource.ResourceId, 1);
-            version = reader.Read(VLatest.Resource.Version, 2);
+            bool versionIsLong = true;
```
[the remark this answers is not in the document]
rajithaalurims: Yes.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -71,6 +71,7 @@
         private readonly ISqlQueryHashCalculator _queryHashCalculator;
         private static ResourceSearchParamStats _resourceSearchParamStats;
         private static object _locker = new object();
+        private static bool versionIsLong = true;
```
SergeyGaluzo, line 74: >versionIsLong [](http://example.com/codeflow?start=28&length=13)

_

SergeyGaluzo (dismissed):
:shipit:

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -807,7 +810,17 @@
         {
             resourceTypeId = reader.Read(VLatest.Resource.ResourceTypeId, 0);
             resourceId = reader.Read(VLatest.Resource.ResourceId, 1);
-            version = reader.Read(VLatest.Resource.Version, 2);
+            retryVersion:
+            try
+            {
+                version = versionIsLong ? (int)reader.GetInt64(2) : reader.GetInt32(2);
```
SergeyGaluzo, line 816: >version = versionIsLong ? (int)reader.GetInt64(2) : reader.GetInt32(2); [](http://example.com/codeflow?start=16&length=71)

Consider adding unit test that flips/flops version

SergeyGaluzo:
>                     var version = reader.Read(table.Version, 3);

Here is on more place to put cast switch

---
Refers to: src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlStoreClient.cs:113 in 5f84614. [](commit_id = 5f846146b7a92f1550caa0eb7843591d477c5753, deletion_comment = False)

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlStoreClient.cs
```diff
@@ -130,7 +131,18 @@
...
+            retryVersion:
+            try
+            {
+                version = versionIsLong ? (int)reader.GetInt64(3) : reader.GetInt32(3);
+            }
+            catch (InvalidCastException)
+            {
+                versionIsLong = !versionIsLong;
+                goto retryVersion;
+            }
```
SergeyGaluzo, line 144: Please add static method to read version and share it across all 4 places.

rajithaalurims: Done

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/ResourceVersion.cs
```diff
@@ -0,0 +1,32 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using Microsoft.Data.SqlClient;
+
+namespace Microsoft.Health.Fhir.SqlServer.Features.Schema
+{
+    public static class ResourceVersion
```
SergeyGaluzo, line 11: >ResourceVersion [](http://example.com/codeflow?start=24&length=15)

Consider just static method on SqlStoreClient. No need to have separate class.

SergeyGaluzo (approved):
:shipit:

mikaelweave (approved):
🦾

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/ResourceVersion.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        private static bool versionIsLong = true;
+
+        public static int GetVersion(SqlDataReader reader, int index)
+        {
+            int version = 0;
+            retryVersion:
+            try
+            {
+                version = versionIsLong ? (int)reader.GetInt64(index) : reader.GetInt32(index);
```
cunninghamjc, line 21: Is there a way to get the type from the SqlDataReader instead?

rajithaalurims: SqlDataReader uses the type defined in the model by default. Type defined in the model and DB should be same unless we have a code rollback where Model has type as Int32 and DB has type as Int64. This code is to take care of compatibility issue in case of code rollback.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/ResourceVersion.cs
```diff
@@ -0,0 +1,32 @@
...
+            }
+            catch (InvalidCastException)
+            {
+                versionIsLong = !versionIsLong;
+                goto retryVersion;
```
brendankowitz, line 26: We should probably avoid using a `goto` here, or pretty much anywhere, ever. There is pretty strong industry consensus that if using a goto there is likely a better/simpler solution.

brendankowitz (changes requested):

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/ResourceVersion.cs
cunninghamjc: SqlDataReader is returned by ExecuteReaderAsync().  Where does it get the model from?

brendankowitz: I don't think `(int)reader.GetInt64(index)` will be compatible if it actually does return a long since it won't fit into an int.

rajithaalurims: @cunninghamjc 
When we read, we do pass column type. See example below. That is how reader gets the type information.
reader.Read(VLatest.Resource.ResourceId, 1);

rajithaalurims: @brendankowitz 
you are right, (int)reader.GetInt64(index) won't be compatible if version actually have a long value. But we are not going to allow customers to send version as long until we are done with iteration 2.
This code is in place to handle rollback situation. Even though column type is long, existing values can fit it into int. This code will be cleaned up in iteration 3 to read it as long.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/ResourceVersion.cs
rajithaalurims: Removed goto

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/ResourceVersion.cs
```diff
@@ -0,0 +1,36 @@
...
+using System;
+using Microsoft.Data.SqlClient;
+
+namespace Microsoft.Health.Fhir.SqlServer.Features.Schema
+{
+    public static class ResourceVersion
+    {
+        private static bool versionIsLong = true;
+
+        public static int GetVersion(SqlDataReader reader, int index)
```
brendankowitz, line 15: Curious also why we can't leverage the SchemaInformation for this? e.g.

```csharp
public static long GetVersion(SqlDataReader reader, int index, SchemaInformation schemaInformation)
{
    if (schemaInformation.Current >= SchemaVersionConstants.VersionAsLong)
    {
        return reader.GetInt64(index);
    }
    else
    {
        return reader.GetInt32(index);
    }
}
```

rajithaalurims: Discussed that option as well. With that approach, there might be a possibility of a gap between actual schema update and schemainformation object reflecting that.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/ResourceVersion.cs
```diff
@@ -0,0 +1,36 @@
...
+                    retry = false;
+                }
+                catch (InvalidCastException)
+                {
+                    versionIsLong = !versionIsLong;
+                }
+            }
+
+            return version;
+        }
```
brendankowitz, line 34: I think @cunninghamjc mentioned this, if we check the type it seems like it should be as simple as an if/else? This seems like its worth testing, this is probably the most efficient implementation here.

```suggestion
        public static int GetVersion(SqlDataReader reader, int index)
        {
            if (reader.GetFieldType(index) == typeof(long))
            {
                return (int)reader.GetInt64(index);
            }
        
            return reader.GetInt32(index);
        }
```

SergeyGaluzo: Please keep in mind that this is a throwaway code, so we should not be spending any extra time to make it super nice. 
BTW As far as efficiency, current code has single if based on single static bool and does not require checking reader metadata.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/ResourceVersion.cs
cunninghamjc: In iteration 2, what happens if an upgraded instance of a FHIR service accepts a long and writes it to the DB, then a non-upgraded instance of the FHIR service reads it back out?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/ResourceVersion.cs
brendankowitz: The cyclomatic complexity of a single `if` statement is far lower than a `try`, a `ternary` , a `while` loop, a switching `bool`, you also have 2 local variable allocations.
You could cache the metadata check without any of the above also.

In the current implement if for some reason `version = versionIsLong ? (int)reader.GetInt64(index) : reader.GetInt32(index);` throws an `InvalidCastException` because it doesn't return an int or int64, its an infinate loop.

Even if its temporary work it, doesn't mean we should ignore all best practices. The very fact its added to handle an edge case and doesn't have a test feels a little uncomfortable.

cunninghamjc: Does this approach also allow us to keep sequential access, since we wouldn't have to read the column twice?

SergeyGaluzo: No, it does not. But this is irrelevant, as "read twice" happens once in the lifetime when schema is switched.

SergeyGaluzo: That "In the current implement if for some reason version = versionIsLong ? (int)reader.GetInt64(index) : reader.GetInt32(index); throws an InvalidCastException because it doesn't return an int or int64, its an infinate loop." is a good point. We can ensure that we retry only once.

SergeyGaluzo: On this "The cyclomatic complexity of a single if statement is far lower than a try, a ternary , a while loop, a switching bool, you also have 2 local variable allocations."
It is all true, but if we compare to reading compressed resource and unzipping it, it goes to a wash.

brendankowitz (approved):
No blockers from me

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/ResourceVersion.cs
rajithaalurims: @cunninghamjc @brendankowitz code is modified to check the type from SqlDataReader. Please take a look.
@cunninghamjc schema update starts after deploying applications. So there will not be a scenario few instances are on new bits and few are on old.
But there is a scenario where after releasing iteration 2 and consuming long version, if we rollback the code, we will run into issue. To avoid that issue,
We will not advertise this new change until we feel comfortable.
And also, we will try avoiding having several major changes going in the same release as iteration2 to minimize reasons for rollback.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/ResourceVersion.cs
rajithaalurims: Simplified the code reading version based on type.

PR #3838, opened 2024-05-03
Return bad request on too many surrogate ids

SergeyGaluzo:
When >80K resources with the same last updated are ingested, code fails with
Current - 500
After fix - 400 with good message

fixes https://microsofthealth.visualstudio.com/Health/_workitems/edit/120027/

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/ImportProcessingJob.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/ImportProcessingJob.cs
@@ -12,6 +12,7 @@ using System.Threading.Tasks;
 using Azure;
 using EnsureThat;
 using MediatR;
+using Microsoft.Data.SqlClient;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Primitives;
 using Microsoft.Health.Core.Features.Audit;
@@ -22,6 +23,7 @@ using Microsoft.Health.Fhir.Core.Features.Context;
 using Microsoft.Health.Fhir.Core.Features.Operations;
 using Microsoft.Health.Fhir.Core.Features.Operations.Import;
 using Microsoft.Health.JobManagement;
+using Microsoft.Health.SqlServer.Features.Storage;
 using Newtonsoft.Json;
 
 namespace Microsoft.Health.Fhir.SqlServer.Features.Operations.Import
@@ -31,6 +33,7 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Operations.Import
     {
         private const string CancelledErrorMessage = "Import processing job is canceled.";
         internal const string DefaultCallerAgent = "Microsoft.Health.Fhir.Server";
+        internal const string SurrogateIdsErrorMessage = "Unable to generate internal IDs. If the lastUpdated meta element is provided as input, reduce the number of resources with the same up to millisecond lastUpdated below 10,000.";
 
         private readonly IMediator _mediator;
         private readonly IQueueClient _queueClient;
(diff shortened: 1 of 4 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -444,10 +444,10 @@ async Task<(List<ImportResource> Loaded, List<ImportResource> Conflicts)> Import
                 }
                 else if (importMode == ImportMode.IncrementalLoad)
                 {
-                    // dedup by last updated - keep all versions when version and lastUpdated exist on record.
+                    // dedup by last updated - take first version for single last updated, prefer large version.
                     var inputsDedupped = resources
-                        .GroupBy(_ => _.ResourceWrapper.ToResourceDateKey(_model.GetResourceTypeId, ignoreVersion: !_.KeepVersion || !_.KeepLastUpdated))
-                        .Select(_ => _.First())
+                        .GroupBy(_ => _.ResourceWrapper.ToResourceDateKey(_model.GetResourceTypeId, ignoreVersion: true))
+                        .Select(_ => _.OrderByDescending(_ => _.ResourceWrapper.Version).First())
```
mikaelweave, line 450: Is this a change of behavior? Will this no longer allow customers to have multiple versions with the same last updated?

src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/ImportProcessingJob.cs
```diff
@@ -31,6 +33,7 @@ public class ImportProcessingJob : IJob
     {
         private const string CancelledErrorMessage = "Import processing job is canceled.";
         internal const string DefaultCallerAgent = "Microsoft.Health.Fhir.Server";
+        internal const string SurrogateIdsErrorMessage = "Unable to generate internal IDs. If the lastUpdated meta element is provided as input, reduce the number of resources with the same up to millisecond lastUpdated below 10,000.";
```
mikaelweave, line 36: NIT - consider resource file if appropriate. Ignore if not.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -459,8 +459,8 @@ async Task<(List<ImportResource> Loaded, List<ImportResource> Conflicts)> Import
 
                 async Task HandleIncrementalVersionedImport(List<ImportResource> inputs, bool useReplicasForReads)
                 {
-                    // Dedup by version via ToResourceKey - only keep first occurance of a version in this batch.
-                    var inputsWithVersion = inputs.Where(_ => _.KeepVersion).GroupBy(_ => _.ResourceWrapper.ToResourceKey()).Select(_ => _.First()).ToList();
+                    // Dedup by version via ToResourceKey - prefer latest dates.
+                    var inputsWithVersion = inputs.Where(_ => _.KeepVersion).GroupBy(_ => _.ResourceWrapper.ToResourceKey()).Select(_ => _.OrderByDescending(_ => _.ResourceWrapper.LastModified.DateTime).First()).ToList();
```
mikaelweave, line 463: Nice - I like this change. Makes more sense than "first in file". 🦾

mikaelweave (commented):
Some questions but in general PR looks good.

Will approve after hearing back about change near:
`// dedup by last updated - take first version for single last updated, prefer large version.`

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
SergeyGaluzo: Yes. It is a fix for a bug. This only affects cases with incorrect input data.

mikaelweave (approved):
🦾

PR #3840, opened 2024-05-07
Negative versions

SergeyGaluzo:


src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/GetResourceVersions.sql
```diff
@@ -13,15 +13,24 @@ BEGIN TRY
...
-                     WHEN isnull(U.Version, 1) - isnull(L.Version, 0) > 1 THEN isnull(U.Version, 1) - 1 
-                     ELSE 0 
+        -- set version to 0 if there is no gap available, or lastUpdated is already used. It would indicate conflict for the caller.
+        ,Version = CASE
+                     -- ResourceSurrogateId is generated from lastUpdated only without extra bits at the end. Need to ckeck interval (0..79999) on resource id level.
+                     WHEN EXISTS (SELECT * 
+                                    FROM dbo.Resource B WITH (INDEX = IX_Resource_ResourceTypeId_ResourceId_Version)
+                                    WHERE B.ResourceTypeId = A.ResourceTypeId 
+                                      AND B.ResourceId = A.ResourceId 
+                                      AND B.ResourceSurrogateId BETWEEN A.ResourceSurrogateId AND A.ResourceSurrogateId + 79999
```
LTA-Thinking, line 23: Why 79999? Is there significance to this number?

SergeyGaluzo: Sure there is. This is baked into surrogate id definition. Look at ResourceSurrogateIdUniquifierSequence

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Import/ImportTests.cs
```diff
@@ -183,14 +183,119 @@ public async Task GivenIncrementalLoad_80KSurrogateIds_BadRequestIsReturned()
...
+            var id = Guid.NewGuid().ToString("N");
+            var ndJson1 = CreateTestPatient(id);
+            var ndJson2 = CreateTestPatient(id, DateTimeOffset.UtcNow.AddDays(-1));
+            var location = (await ImportTestHelper.UploadFileAsync(ndJson1 + ndJson2, _fixture.StorageAccount)).location;
+            var request = CreateImportRequest(location, ImportMode.IncrementalLoad, false, true);
+            await ImportCheckAsync(request, null, 0);
+
+            var history = await _client.SearchAsync($"Patient/{id}/_history");
+            Assert.Single(history.Resource.Entry);
+            Assert.Equal("1", history.Resource.Entry[0].Resource.VersionId);
```
LTA-Thinking, line 197: I think I'm missing something here. Why aren't both versions included?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Import/ImportTests.cs
```diff
@@ -183,14 +183,119 @@ public async Task GivenIncrementalLoad_80KSurrogateIds_BadRequestIsReturned()
...
+            var ndJson1 = CreateTestPatient(id);
+            var ndJson2 = CreateTestPatient(id, DateTimeOffset.UtcNow.AddDays(1));
+            var location = (await ImportTestHelper.UploadFileAsync(ndJson1 + ndJson2, _fixture.StorageAccount)).location;
+            var request = CreateImportRequest(location, ImportMode.IncrementalLoad, false, true);
+            await ImportCheckAsync(request, null, 0);
+
+            var history = await _client.SearchAsync($"Patient/{id}/_history");
+            Assert.Equal(2, history.Resource.Entry.Count);
+            Assert.Equal("1", history.Resource.Entry[0].Resource.VersionId);
+            Assert.Equal("-1", history.Resource.Entry[1].Resource.VersionId);
```
LTA-Thinking, line 213: Why doesn't this import as versions 1 & 2?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Import/ImportTests.cs
```diff
@@ -183,14 +183,119 @@ public async Task GivenIncrementalLoad_80KSurrogateIds_BadRequestIsReturned()
...
+
+            // check history
+            result = await _client.VReadAsync<Patient>(ResourceType.Patient, id, "1");
+            Assert.Equal(GetLastUpdated("2001"), result.Resource.Meta.LastUpdated);
+            result = await _client.VReadAsync<Patient>(ResourceType.Patient, id, "-1");
+            Assert.Equal(GetLastUpdated("2002"), result.Resource.Meta.LastUpdated);
+        }
+
+        [Fact]
+        public async Task GivenIncrementalLoad_MultipleInputVersionsOutOfOrder1NotExplicit_ResourceNotExisting_NoGap_DoNotAllowNegative()
```
LTA-Thinking, line 291: Should this test check the import response to confirm it is telling the user the version in ndJson2 was skipped?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Import/ImportTests.cs
```diff
@@ -311,6 +416,38 @@ public async Task GivenIncrementalLoad_SameLastUpdated_DifferentVersions_Resourc
             Assert.Equal(GetLastUpdated("2002"), result.Resource.Meta.LastUpdated);
         }
 
+        [Fact]
+        public async Task GivenIncrementalLoad_SameLastUpdated_DifferentVersions_DifferentImports_ResourceExisting()
```
LTA-Thinking, line 420: Same question as above, should the response be checked to show a message about ndJson1 is being returned?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Import/ImportTests.cs
SergeyGaluzo: Both resources have the same content. Version with explicit last updated is loaded first. When we attempt to load version with implicit last updated we check current database state. We see that there is version with identical content, and we return - loaded. Hence, both input records were loaded :-), but only one version is actually present in the database.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Import/ImportTests.cs
SergeyGaluzo: Record with explicit last updated is loaded first, It takes version = 1. BTW It is from future. When record with implicit last updated is worked on, version 2 cannot be used because it would break our rule "version on current record is always max, so we can just increment in normal circumstances", Therefore -1.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Import/ImportTests.cs
SergeyGaluzo: We are checking response on line 299. It says one record was not loaded. Subsequently we check for first and third records. Hence only second record is not loaded.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Import/ImportTests.cs
SergeyGaluzo: The check is on line 437. It tells that one record was not loaded. That import contained only one record, hence we know that it was not loaded.

LTA-Thinking (approved):

LTA-Thinking (dismissed):

LTA-Thinking (approved):

LTA-Thinking (approved):

LTA-Thinking (approved):

LTA-Thinking (approved):

docs/BulkImport.md
```diff
@@ -105,7 +105,8 @@ Content-Type:application/fhir+json
 | Parameter Name      | Description | Card. |  Accepted values |
 | ----------- | ----------- | ----------- | ----------- |
 | inputFormat      | String representing the name of the data source format. Currently only FHIR NDJSON files are supported. | 1..1 | ```application/fhir+ndjson``` |
-| mode      | Import mode. Currently only initial load mode is supported. | 1..1 | For initial import use ```InitialLoad``` mode value. For incremental import mode use ```IncrementalLoad``` mode value. If no mode value is provided, IncrementalLoad mode value is considered by default. |
+| mode      | Import mode. | 0..1 | For initial import use ```InitialLoad``` mode value. For incremental import mode use ```IncrementalLoad``` mode value. If no mode value is provided, IncrementalLoad mode value is considered by default. |
+| allowNegativeVersions | Allows FHIR server assigning negative versions for out of order processed records with explicit lastUpdated value and no version specified. | 0..1 | To enable this feature pass true. By default it is false. |
```
EXPEkesheth, line 109: Can we update the flag name to "resource-versionagnostic"? So when it is set to true, this feature is enabled.

EXPEkesheth, line 109: Should the description state -" Ingests records in FHIR service based on lastUpdated value when no version is specified. In case of records ingested out of order, negative version values can be assigned.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -379,7 +378,7 @@ await using (new Timer(async _ => await _sqlStoreClient.MergeResourcesPutTransac
             return results;
         }
 
-        internal async Task<IReadOnlyList<string>> ImportResourcesAsync(IReadOnlyList<ImportResource> resources, ImportMode importMode, CancellationToken cancellationToken)
+        internal async Task<IReadOnlyList<string>> ImportResourcesAsync(IReadOnlyList<ImportResource> resources, ImportMode importMode, bool allowNegativeVersions, CancellationToken cancellationToken)
```
mikaelweave, line 381: Note: If we enhance this in the future with more logic, we should evaluate breaking this single unit into multiple units as long units are harder to debug/enhance without breaking.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Import/ImportTests.cs
mikaelweave: @SergeyGaluzo - should this be true even if the content of the resource is the same (except for explicit vs implicit last updated)? What would the behavior be if these resources were imported "in order" or "not late"?

mikaelweave (approved):
A couple small comments but approved 🦾

docs/BulkImport.md
SergeyGaluzo: I don't think so. 

"Ingests records in FHIR service based on lastUpdated value when no version is specified" 
is not limited to this new flag. This behavior exists even without it. Only in cases when there is no "room" for incoming record in positive version space this flag changes "drop on the floor with conflict" to "ingest with negative version".

I am open for suggestions though.

SergeyGaluzo: This sounds less counter intuitive for me.

SergeyGaluzo:
@mikaelweave in reference to https://github.com/microsoft/fhir-server/pull/3840#discussion_r1607359492

When customer specifies last updated, he/she takes control over versioning. Only if last updated is not specified, we can do checks on data content. If we do opposite, we can run into the problem, when according to current database state a record does not need to be written, but if other late arrivals come, it makes this decision invalid.

mikaelweave:
> @mikaelweave in reference to [#3840 (comment)](https://github.com/microsoft/fhir-server/pull/3840#discussion_r1607359492)
> 
> When customer specifies last updated, he/she takes control over versioning. Only if last updated is not specified, we can do checks on data content. If we do opposite, we can run into the problem, when according to current database state a record does not need to be written, but if other late arrivals come, it makes this decision invalid.

I understand - if resources are imported specifying last updated, then we encounter scenarios where this is required. Since we import latest first (by last updated) in a batch, the latest one here takes version #1.

docs/BulkImport.md
EXPEkesheth: In future, we decided to got with character support for version. How will this flag consumed?

SergeyGaluzo: When we decide to support version as string, "version agnostic" imports will be supported without any special flags. This will be their default behavior as there will be not any version contiguity requirements. When and if this happens, this flag will not be needed and should be deprecated.

EXPEkesheth: Ok looks good to me

EXPEkesheth (dismissed):

LTA-Thinking (approved):

mikaelweave (approved):
small NIT but looks great 😄

PR #3841, opened 2024-05-07
Return 500 for requests that timed-out for  now

apurvabhaleMS:
## Description
Until we decide on the billing part, let's return 500 for requests that timed-out
For Transactions - (This is what we do today as well if the transaction is failed due to any reason)
- If any request failed with 408 (first, midway or last), log the statistics upto where we have collected them with 500 instead of 408
- Do not publish metric notification (No need to worry about the billing as metrics are not published and transaction is rolledback)
- Final status of the transaction would be last failed operation status which is 500 in this case

For Batches -
- Statistics would have an information about how many requests actually failed due to time-out with status code 500
- Publish the metric notification with count of passed + timed-out requests with 500 as status code
- Final status of the batch is always 200

## Related issues
Addresses [AB119603](https://microsofthealth.visualstudio.com/Health/_workitems/edit/119603)

## Testing
1. 
2. .

## FHIR Team Checklist


apurvabhaleMS:
Synced offline and decided that this change is not needed

PR #3842, opened 2024-05-07
Enable export for SMART clinical scope : read

feordin:
## Description
This adds an allowed DataAction of Export when a SMART scope of "read" is requested.  This change is being made because prior to this it was necessary to grant "write" privileges to a SMART user if they needed export.  This was too much privilege for an export request.

Note that there is still another layer of protection before a SMART user could actually initiate an export job.  They will need to be a member of the FHIR Export role in RBAC in addition to requesting the "read" SMART clinical scope.

## Related issues
Addresses [issue [AB#120097](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/120097)].

## Testing
Unit tests were updated.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Api/Features/SMART/SmartClinicalScopesMiddleware.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/SMART/SmartClinicalScopesMiddleware.cs
@@ -99,7 +99,7 @@ namespace Microsoft.Health.Fhir.Api.Features.Smart
                         switch (accessLevel)
                         {
                             case "read":
-                                permittedDataActions = DataActions.Read;
+                                permittedDataActions = DataActions.Read | DataActions.Export;
                                 break;
                             case "write":
                                 permittedDataActions = DataActions.Write;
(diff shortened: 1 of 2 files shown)


LTA-Thinking:
> Note that there is still another layer of protection before a SMART user could actually initiate an export job. They will need to be a member of the FHIR Export role in RBAC in addition to requesting the "read" SMART clinical scope.

If they still need the RBAC roll what value is the SMART roll? Wouldn't they be able to export without the SMART roll if they had the RBAC roll?

feordin:
> > Note that there is still another layer of protection before a SMART user could actually initiate an export job. They will need to be a member of the FHIR Export role in RBAC in addition to requesting the "read" SMART clinical scope.
> 
> If they still need the RBAC roll what value is the SMART roll? Wouldn't they be able to export without the SMART roll if they had the RBAC roll?

It is true that once they are members of the Export role they have export permissions, but if that same user is also a member of the SMART role, the actions they take are then limited by the SMART clinical scopes in their token.  Right now, we have to grant write permissions in the SMART clinical scope in order for the user to take advantage of their Export privileges.

LTA-Thinking (approved):

Issue #3846, opened 2024-05-08
Unable to retrieve Patient resource with numerous versions from search


tswelsh:
Apologies in advance as this is a bit long and a bit vague.  I wasn't sure whether it is a bug or a question, but something slightly dodgy appears to have happened so I plumped for bug.

**Describe the bug**
We have a Patient resource that has had multiple updates over time.  When searching for it using the identifier search param, the search returns no results even though it definitely should.  I'm not sure if the fault here is with the SQL query that does the search, or if the data in this particular DB has got into a funny state.

The Patient resource in question has 12 versions.  The query we are trying is equivalent to `/Patient?identifier=https://local.id.system|123456`.  All 12 versions have the identifier, and retrieving the Patient directly by logical id also shows that it definitely has that identifier.  Digging deeper, the generated SQL query looks like this:

```
SET STATISTICS IO ON;
SET STATISTICS TIME ON;

DECLARE @FilteredData AS TABLE (T1 smallint, Sid1 bigint, IsMatch bit, IsPartial bit, Row int)
;WITH
cte0 AS
(
    SELECT ResourceTypeId AS T1, ResourceSurrogateId AS Sid1
    FROM dbo.TokenSearchParam
    WHERE SearchParamId = 861
        AND SystemId = @p0
        AND Code = @p1 
        AND ResourceTypeId = 81 -- Patient
)
,cte1 AS
(
    SELECT DISTINCT TOP (@p2) T1, Sid1, 1 AS IsMatch, 0 AS IsPartial 
    FROM cte0
    ORDER BY T1 ASC, Sid1 ASC
)
/* HASH sNtlrPkDO22+3QwZFia3EKOx6dlmdmiMQYxXfq57GD4= params=@p0,@p1 */
SELECT DISTINCT r.ResourceTypeId, r.ResourceId, r.Version, r.IsDeleted, r.ResourceSurrogateId, r.RequestMethod, CAST(IsMatch AS bit) AS IsMatch, CAST(IsPartial AS bit) AS IsPartial, r.IsRawResourceMetaSet, r.SearchParamHash, r.RawResource
FROM dbo.Resource r
     JOIN cte1 ON r.ResourceTypeId = cte1.T1 AND r.ResourceSurrogateId = cte1.Sid1
WHERE IsHistory = 0 
ORDER BY r.ResourceTypeId ASC, r.ResourceSurrogateId ASC
```

[...]

PR #3847, opened 2024-05-08
Cleanup search parameters on historical resources

LTA-Thinking:
## Description
Adds a SQL Schema version that cleans up search parameters on historical resources.

## Related issues
Addresses [AB#102248](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/102248)
#3846 

## FHIR Team Checklist

Patch


SergeyGaluzo:
Is not it too late? We have prior versions that rely on deleted history. They will not be able to apply if go in version order. Should we fix prior version instead?

LTA-Thinking:
> Is not it too late? We have prior versions that rely on deleted history. They will not be able to apply if they go in version order. Should we fix prior version instead?

Good point, it should be in both. That way people who have already updated will still get it, and people who have yet to update get it sooner.

SergeyGaluzo:
I don't see how version 73 can be successful with deleting history first. If so, I don't see the need to have special version to do this. We just need to fix 73.

abiisnn (approved):

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/80.diff.sql
```diff
@@ -0,0 +1 @@
+Execute dbo.DeleteHistory 0,0,0
```
fhibf, line 1: What is the reason for this statement?

LTA-Thinking: What has happened is that after Schema 76 the IsHistory column was removed from search parameter tables. This column hadn't been being used since v48, but for any customers who had data older than that there could be historical data still in those tables. That data was preventing current data from being returned in search requests because it was filling up the count of resources the search was looking for. 
This stored procedure was written as part of v48 to clean up this data, but it was never actually run. This change explicitly runs this stored proc to finally clean up this unneeded data. It does not delete the historical resources, just the search parameters connected to them.

PR #3859, opened 2024-05-15
Update $convert-data error log

pallar-ms:
## Description
Update $convert-data error log. Currently, the entire exception is logged in some cases (except post processing exceptions) which might not be ideal in case the message has text from the input data being processed.
Updating the log to only mention exception type, FHIRConverterErrorCode and stack trace.

## Related issues
Addresses [AB#120207](https://microsofthealth.visualstudio.com/Health/_workitems/edit/120207).

## Testing

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Operations/ConvertData/ConvertDataEngine.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Operations/ConvertData/ConvertDataEngine.cs
@@ -96,7 +96,6 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.ConvertData
                 }
 
                 _logger.LogError($"Convert data failed. An exception of type {convertException.GetType()} occurred with error code - {convertException.FhirConverterErrorCode}. {convertException.StackTrace}");
-
                 throw new ConvertDataFailedException(string.Format(Core.Resources.ConvertDataFailed, convertException.Message), convertException);
             }
             catch (Exception ex)
--- a/src/Microsoft.Health.Fhir.Core/Features/Operations/ConvertData/ConvertDataEngine.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Operations/ConvertData/ConvertDataEngine.cs
@@ -95,14 +95,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.ConvertData
                     throw new ConvertDataTimeoutException(Core.Resources.ConvertDataOperationTimeout, convertException.InnerException);
                 }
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.Health.Fhir.Core/Features/Operations/ConvertData/ConvertDataEngine.cs
```diff
@@ -95,15 +95,7 @@ private string GetConvertDataResult(ConvertDataRequest convertRequest, ITemplate
...
-                if (convertException is PostprocessException)
-                {
-                    _logger.LogInformation($"An exception of type {convertException.GetType()} occurred during postprocessing. {convertException.StackTrace}", "Convert data failed.");
-                }
-                else
-                {
-                    _logger.LogInformation(convertException, "Convert data failed.");
-                }
-
+                _logger.LogError($"Convert data failed. An exception of type {convertException.GetType()} occurred with error code - {convertException.FhirConverterErrorCode}. {convertException.StackTrace}");
```
dustinburson, line 98: One thing we are losing is the an explicit message that the error occurred during post processing.  We should be able to infer that from the stack trace but it may still be nice to include a post processing indicator in the log for clarity.

dustinburson (dismissed):

src/Microsoft.Health.Fhir.Core/Features/Operations/ConvertData/ConvertDataEngine.cs
pallar-ms: The type already has that information.
e.g. of how it was logged - An exception of type Microsoft.Health.Fhir.Liquid.Converter.Exceptions.PostprocessException occurred during postprocessing. 

Yeah if we have derived classes for PostProcessException later, then yes this would make sense, but that is also why I added the FhirConverterErrorCode as that is meant to give more granular nsights on when/what error caused this exception. So I feel that should lead us to the exact point.

pallar-ms: But I can also add that check if we feel the exception type is not enough to indicate that.

wi-y (approved):

wi-y (approved):

LTA-Thinking (approved):
Approving to lift FHIR Owner's block.

pallar-ms:
@LTA-Thinking , thanks for taking a look. Question - the PR build & deploy seems to be failing at an unrelated change to this PR, and I am also seeing other PRs blocked by the same error. Do you know if anyone is taking a look into that? What is the best place to bring this up and track status?

PR #3890, opened 2024-05-30
Refactor last updated deduping after UTC change

SergeyGaluzo:
Addresses https://microsofthealth.visualstudio.com/Health/_workitems/edit/120854/

### DIFF
--- a/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Import/ImportTests.cs
+++ b/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Import/ImportTests.cs
@@ -205,9 +205,9 @@ IF (SELECT count(*) FROM EventLog WHERE Process = 'MergeResourcesCommitTransacti
             request = CreateImportRequest(location, ImportMode.IncrementalLoad, false);
             await ImportCheckAsync(request, null, 0); // reported loaded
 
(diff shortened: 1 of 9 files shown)


test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Import/ImportTests.cs
```diff
@@ -276,7 +276,7 @@ public async Task GivenIncrementalLoad_MultipleImportsWithSameLastUpdatedAndImpl
             ndJson = CreateTestPatient(id, baseDate.AddYears(-1));
             location = (await ImportTestHelper.UploadFileAsync(ndJson, _fixture.StorageAccount)).location;
             request = CreateImportRequest(location, ImportMode.IncrementalLoad, false, true);
-            await ImportCheckAsync(request, null, 1);
+            await ImportCheckAsync(request, null, 0);
```
LTA-Thinking, line 279: So this is now considered a duplicate (and thus not an error)?

SergeyGaluzo: The logic was faulty. Now when code sees input data that is identical to prior loaded (key is last updated only, as there is no version) it considers it already loaded, so no errors.

LTA-Thinking (approved):

PR #3894, opened 2024-06-03
Reflect current code state for $import in the doc

SergeyGaluzo:
Several corrections/clarifications

### DIFF
--- a/docs/BulkImport.md
+++ b/docs/BulkImport.md
@@ -10,7 +10,7 @@ The Bulk import feature enables importing FHIR data in the NDJSON format to the 
 ## Prerequisites
 
 * NDJSON FHIR data to be imported.
-* You may have multiple NDJSON files per resource type.
+* You may have multiple NDJSON files.
 
 ### Current limitations
 
--- a/docs/BulkImport.md
+++ b/docs/BulkImport.md
@@ -10,7 +10,7 @@ The Bulk import feature enables importing FHIR data in the NDJSON format to the 
 ## Prerequisites
 
 * NDJSON FHIR data to be imported.
-* You may have multiple NDJSON files.
+* You may have multiple NDJSON files in single import. There is a soft limit on maximum number of files, it exceeds 10000.
 
 ### Current limitations
 
--- a/docs/BulkImport.md
+++ b/docs/BulkImport.md
@@ -320,7 +320,8 @@ Below are some errors you may encounter:
 1. Deploy the FHIR server, SQL Server database, and the storage account in the same region to avoid data movement across regions.
 1. The optimal NDJSON file size for import is >=50MB (or >=20K resources, no upper limit). Consider combining smaller files together.
 1. For optimal performance total size of files in single import should be large (>=100GB or >=100M resources, no upper limit). 
-1. Though multiple parallel imports are supported, best performance can be achieved for single import with the same payload as in multiple parallel imports. There is no limit on number of files in single import (tested with 50K files). 
+1. Though multiple parallel imports are supported, best performance can be achieved for single import with the same payload as in multiple parallel imports.
+1. There is no limit on number of files in single import (tested with 50K files). Small number of files (up to single file) is preferred. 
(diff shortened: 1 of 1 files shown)


docs/BulkImport.md
```diff
@@ -3,7 +3,7 @@
 There are two modes of $import supported today-
 
 1. Initial mode is intended to load FHIR resources into an empty FHIR server. Initial mode only supports CREATE operations and, when enabled, blocks API writes to the FHIR server.
-1. Incremental mode is optimized to load data into FHIR server periodically and doesn't block writes via API. It also allows you to load lastUpdated and versionId from resource Meta (if present in resource JSON).
+1. Incremental mode is optimized to load data into FHIR server periodically and doesn't block writes via API. It also allows you to load lastUpdated, or both lastUpdated and versionId, from resource Meta (if present in resource JSON). There are no performance differences between incremental and initial modes.
```
EXPEkesheth, line 7: Xoriant in their run identified performance difference. Lets double click offline

docs/BulkImport.md
```diff
@@ -15,8 +15,8 @@ The Bulk import feature enables importing FHIR data in the NDJSON format to the
 
 ### Current limitations
 
-* Conditional references in resources are not supported for initial mode import.
-* If multiple resources share the same resource ID, then only one of those resources will be imported at random and an error will be logged corresponding to the remaining resources sharing the ID.
+* In initial mode, resources with conditional references will be logged as errors. In incremental mode, resources with conditional references will be loaded but references will not be resolved.
+* If multiple resources share the same resource ID, in initial mode, only one of those resources will be imported at random and an error will be logged corresponding to the remaining resources sharing the ID. In incremental mode, if lastUpdated is not provided, behavior is the same as in the initial. If lastUpdated is provided, then resource versions are deduplicated by lastUpdated, and only duplicates are reported as errors, the rest is loaded.
```
EXPEkesheth, line 19: If lastUpdated is provided, then resource versions are deduplicated by lastUpdated, and only duplicates are reported as errors, the rest is loaded. --- is this true with allowed version flag ? In case there are multiple resources with same version id and different lastUpdated, then error will be reported?

SergeyGaluzo: 1. This is "universally" true irrespective of allow negative versions flag.
2. If there are multiple records for the same resource with different lastUpdated and the same version, then version conflict will be reported.

docs/BulkImport.md
SergeyGaluzo: In my tests on hospital data with ~1B resources performed on correctly scaled FHIR server and database I saw zero performance differences. This is as expected, because slight differences in compute on FHIR server related to processing modes do not change the database t-log bottleneck, and this bottleneck identifies the performance of import.

docs/BulkImport.md
EXPEkesheth: For statement "In incremental mode, if lastUpdated is not provided, behavior is the same as in the initial."  -  Initial mode import creates resources and not updates. So if lastUpdated is not provided but incremental version ids are provided, then the resource is updated.  

Should we update the statement?

EXPEkesheth: "If there are multiple records for the same resource with different lastUpdated and the same version, then version conflict will be reported." - Version conflict will be reported on the randomly selected resource ?

SergeyGaluzo: The answer is "not predictable" record. The reason is that answer is dependent on order of processing across batches, and only in a single batch any order can be guaranteed.

SergeyGaluzo: > For statement "In incremental mode, if lastUpdated is not provided, behavior is the same as in the initial." - Initial mode import creates resources and not updates. So if lastUpdated is not provided but incremental version ids are provided, then the resource is updated.
> 
> Should we update the statement?

Input versions are ignored if lastUpdated is not provided. This has existed since day one for incremental mode. There is no need to update the statement.

EXPEkesheth (approved):

PR #3900, opened 2024-06-05
Allow datetime values to be treated as strings

rogordon01:
Currently, default Newtonsoft JsonSerialization settings are used when performing Json deserialization. This impacts datetime values, as the default behavior is to convert them into .Net `DateTime` objects. This causes unwanted formatting issues and leads to loss of TimeZone information on the converted data. 

This PR allows end users to instruct Newtonsoft to simply treat dates as strings during deserialization. This allows the datetime information to be preserved as-is. They can then be assured that any filters are applied on the original value, and not a value modified by Newtownsoft. 

End users must opt-in to this new behavior by specifying a new Parameter.

Current behavior:
image

When enabled:
image

## Description

## Related issues
Addresses [AB#120779](https://microsofthealth.visualstudio.com/Health/_workitems/edit/120779).

## Testing

## FHIR Team Checklist


ChaChaChaChami (approved):
Makes sense.

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Operations/ConvertData/ConvertDataEngineTests.cs
```diff
@@ -92,7 +93,7 @@ public async Task GivenCcdaConvertDataRequest_WithADefaultTemplates_CorrectResul
         public async Task GivenJsonConvertDataRequest_WithADefaultTemplates_CorrectResultShouldReturn()
         {
             var convertDataEngine = GetDefaultEngine();
-            var request = GetJsonRequestWithDefaultTemplates();
+            var request = GetJsonRequestWithDefaultTemplates(true);
```
pallar-ms, line 96: nit: add param name for values passed as is.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ConvertDataController.cs
```diff
@@ -76,6 +78,7 @@ public async Task<IActionResult> ConvertData([FromBody] Parameters inputParams)
             string inputData = ReadStringParameter(inputParams, ConvertDataProperties.InputData);
             string templateCollectionReference = ReadStringParameter(inputParams, ConvertDataProperties.TemplateCollectionReference);
             string rootTemplate = ReadStringParameter(inputParams, ConvertDataProperties.RootTemplate);
+            bool treatDatesAsStrings = ReadBoolParameter(inputParams, ConvertDataProperties.JsonDeserializationTreatDatesAsStrings);
```
pallar-ms, line 81: Just to confirm, this won't error if the request doesn't have this param present?

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ConvertDataController.cs
```diff
@@ -180,6 +189,7 @@ private static HashSet<string> GetSupportedParams()
                 ConvertDataProperties.InputDataType,
                 ConvertDataProperties.TemplateCollectionReference,
                 ConvertDataProperties.RootTemplate,
+                ConvertDataProperties.JsonDeserializationTreatDatesAsStrings,
```
pallar-ms, line 192: By adding it in the supported param list, upon validation of params we are enforcing this parameter is provided in the request, right? Is that intended or can this simply not be provided in the request and default if so. If it is the former, I suppose we will need an api version update.

rogordon01: The supported param list contains all of the params that this controller works with. So if a parameter provided by the end user doesn't exist within this list then the user will get an error. It doesn't enforce that these params need to _exist_, though. The various `ReadXXXParameter` method seem to do that.  

As you've stated, I've opted to allow a default value to be used if one isn't provided. I added a new `ReadBoolParameter` which will return `false` if the parameter is not present.

pallar-ms (dismissed):

src/Microsoft.Health.Fhir.Core/Data/OperationDefinition/convert-data.json
```diff
@@ -43,6 +43,14 @@
             "max": "1",
             "documentation": "The root template to use while transforming the data. Supported default values - `ADT_A01`, `OML_O21`, `ORU_R01`, `VXU_V04` for input data type `Hl7v2`, `CCD`, `ConsultationNote`, `DischargeSummary`, `HistoryandPhysical`, `OperativeNote`, `ProcedureNote`, `ProgressNote`, `ReferralNote`, `TransferSummary` for input data type `Ccda`, `ExamplePatient`, `Stu3ChargeItem` for data type `Json`, and use resource name as root template for `Fhir`",
             "type": "string"
+        },
+        {
+            "name": "jsonDeserializationTreatDatesAsStrings",
+            "use": "in",
+            "min": 1,
```
LTA-Thinking, line 50: Since this isn't a required field the min should be 0.

rogordon01: Done

LTA-Thinking (approved):

Issue #3920, opened 2024-06-14
When searching with _security:not , deleted resources are included in the search results.


klimkjar:
**Describe the bug**
When the _security:not search parameter is specified in a search operation, IDs for soft-deleted resources are included in the search results. I've reproduced the behaviour with a fresh install of version 4.0.160 , but it appears at least as far back as in version 3.4.124 .

**FHIR Version?**
R4

**Data provider?**
SQL Server

**To Reproduce**
1. POST a Patient resource to the fhir-server instance. The contents are not important, for repro purposes even an empty Patient resource is enough.
2. DELETE the resource that was created in step 1.
3. SEARCH for resources using the `_security:not` parameter, for instance `/Patient?_security:not=V` . It needs to be a generic search, if the ID for the resource is specified with `_id=...` the behaviour does not appear, but adding `_lastUpdated=...` to limit the results if testing on a server with lots of existing data works.

**Expected behavior**
The deleted resource does not appear in the list of results.

**Actual behavior**
The deleted resource appears in the list as an empty resource (containing only the id and meta.lastUpdated fields).

feordin:
Thank you for the bug report.  I have also been able to repro the issue and confirmed it.  We'll get it on the list to fix.

EXPEkesheth:
@feordin - can this be closed?

feordin:
Yes, this is resolved.

PR #3925, opened 2024-06-14
Add IsDeleted check in search queries

feordin:
## Description
We do not have a check for resources being in the deleted state on search queries.  This can result in deleted resources being returned in result sets under some conditions.

## Related issues
Addresses [issue #3920].  DevOps [issue [AB#121553](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/121553)]

## Testing
New E2E test added.

## FHIR Team Checklist

### DIFF
--- a/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/DeleteTests.cs
+++ b/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/DeleteTests.cs
@@ -76,12 +76,12 @@ namespace Microsoft.Health.Fhir.Tests.E2E.Rest
         [Fact]
         public async Task GivenADeletedResource_WhenSearchWithNotModifier_ThenDeletedResourcesShouldNotBeReturned()
         {
-            using FhirResponse<Flag> response = await _client.CreateAsync(Samples.GetJsonSample("Flag").ToPoco<Flag>());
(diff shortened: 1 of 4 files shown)


test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/DeleteTests.cs
```diff
@@ -73,6 +73,19 @@ async Task<FhirClientException> ExecuteAndValidateGoneStatus(Func<Task> action)
             }
         }
 
+        [Fact]
+        public async Task GivenADeletedResource_WhenSearchWithNotModifier_ThenDeletedResourcesShouldNotBeReturned()
```
LTA-Thinking, line 77: Is there a way to write this test using tags? I don't like relying on us never using the Flag resource type in the future, it looks like it could cause hard to debug failures down the road.

LTA-Thinking (approved):

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/DeleteTests.cs
feordin: I did think about that, but I can't use the tag value later to search, because the resources will be deleted, and the tags should also be deleted at that point.

PR #3928, opened 2024-06-17
[Metrics] New components to track HTTP failures and exceptions.

fhibf:
## Description
 New components to emit signals to track HTTP failures and exceptions.

## Workitem
[AB#117202](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/117202)

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Logging/Metrics/DefaultFailureMetricHandler.cs
+++ b/src/Microsoft.Health.Fhir.Core/Logging/Metrics/DefaultFailureMetricHandler.cs
@@ -11,21 +11,21 @@ namespace Microsoft.Health.Fhir.Core.Logging.Metrics
 {
     public sealed class DefaultFailureMetricHandler : BaseMeterMetricHandler, IFailureMetricHandler
     {
-        private readonly Counter<long> _counterErrors;
-        private readonly Counter<long> _counterHttpErrors;
+        private readonly Counter<long> _counterExceptions;
+        private readonly Counter<long> _counterHttpFailures;
 
         public DefaultFailureMetricHandler(IMeterFactory meterFactory)
             : base(meterFactory)
         {
-            _counterErrors = MetricMeter.CreateCounter<long>("Failures.Exceptions");
-            _counterHttpErrors = MetricMeter.CreateCounter<long>("Failures.HttpFailures");
+            _counterExceptions = MetricMeter.CreateCounter<long>("Failures.Exceptions");
+            _counterHttpFailures = MetricMeter.CreateCounter<long>("Failures.HttpFailures");
         }
 
         public void EmitException(IExceptionMetricNotification notification)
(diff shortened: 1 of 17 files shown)


src/Microsoft.Health.Fhir.Core/Logging/Metrics/DefaultFailureMetricHandler.cs
```diff
@@ -0,0 +1,44 @@
...
+
+using System.Collections.Generic;
+using System.Diagnostics.Metrics;
+using EnsureThat;
+
+namespace Microsoft.Health.Fhir.Core.Logging.Metrics
+{
+    public sealed class DefaultFailureMetricHandler : BaseMeterMetricHandler, IFailureMetricHandler
+    {
+        private readonly Counter<long> _counterErrors;
```
fhibf, line 14: _counterException

src/Microsoft.Health.Fhir.Shared.Api/Features/Exceptions/BaseExceptionMiddleware.cs
```diff
@@ -105,13 +112,30 @@ public async Task Invoke(HttpContext context)
...
             await result.ExecuteResultAsync(new ActionContext { HttpContext = context });
         }
+
+        private void EmitHttpFailureMetricInCaseOfError(HttpContext context, bool doesOperationOutcomeHaveError)
+        {
+            if (context.Response.StatusCode >= 500 || doesOperationOutcomeHaveError)
+            {
+                string operationName = context.Request.GetOperationName(includeRouteValues: false);
+
+                _failureMetricHandler.EmitHttpFailure(
```
mikaelweave, line 136: Does it make sense to have additional info in this metric like HTTP code? Or since they are all >= 500 perhaps it won't tell us much? 

I'm thinking we way want to measure 503s vs 500s but not sure how this is going to be used downstream.

mikaelweave (approved):
One question but otherwise looks great / clean. 🦾

src/Microsoft.Health.Fhir.Shared.Api/Features/Exceptions/BaseExceptionMiddleware.cs
fhibf: Yeap. I had the same thinking when developing this. 

The current metric tracks HTTP failures (>=500).
I think that from the DRI perspective, if a high volume of 5**s are happening, you don't care if they are 500, 503 or 504. 
It should probably act fast, because it's a problem on our side. 

That was my line of thought.

mikaelweave: Makes sense - we can always get more info from logs. 👍🏽

PR #3950, opened 2024-07-04
Cosmos managed Identity Authentication Changes

v-ajajvanu:
## Description
Made changes to support Managed Identity Authentication for Cosmos DB instead of using Secrets.

## Related issues
user story [AB#122541](https://microsofthealth.visualstudio.com/Health/_workitems/edit/122541)

## Testing
Tested by creating FHIR server instance with code running from publishing new images to ACR that contain managed identity authentication changes.

## FHIR Team Checklist


src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosDbCollectionPhysicalPartitionInfo.cs
```diff
@@ -33,19 +35,23 @@ internal class CosmosDbCollectionPhysicalPartitionInfo : IRequireInitializationO
         private readonly IHttpClientFactory _httpClientFactory;
         private readonly ILogger<CosmosDbCollectionPhysicalPartitionInfo> _logger;
         private readonly CancellationTokenSource _backgroundLoopCancellationTokenSource = new();
+        private readonly IAccessTokenProvider _aadTokenProvider;
         private Task _backgroundLoopTask;
 
         public CosmosDbCollectionPhysicalPartitionInfo(
+            IAccessTokenProvider aadTokenProvider,
```
abiisnn, line 42: What "add" in "aadTokenProvider" stands for?

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosDbCollectionPhysicalPartitionInfo.cs
```diff
@@ -92,46 +98,18 @@ private async Task<int> GetPhysicalPartitionCount(CancellationToken cancellation
...
-                }
-                else
-                {
-                    Ensure.That(host, $"{nameof(CosmosDataStoreConfiguration)}.{nameof(CosmosDataStoreConfiguration.Host)}").IsNotNullOrEmpty();
-                }
-            }
-            else if (string.IsNullOrWhiteSpace(key))
-            {
-                Ensure.That(key, $"{nameof(CosmosDataStoreConfiguration)}.{nameof(CosmosDataStoreConfiguration.Key)}").IsNotNullOrEmpty();
-            }
```
abiisnn, line 112: I think we should keep the same behavior but using the new **tokenProvider**.

We might need to add a new variable called "**token**" in [CosmosDBLocalEmuletor](https://github.com/microsoft/fhir-server/blob/fb9836bb14817f50977860217ff69a034493e90d/src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosDbLocalEmulator.cs#L8) and make sure we clean up "**key**" in case it is not needed anymore. If "token" is not longer needed (because now you use _new DefaultAzureCredential()_), then can we clean up all key references?

I think we can keep the same structure because Ensure.That() will throw an exception in case host or token are null or white space, as now, it will just check **host** and not **token**.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosDbCollectionPhysicalPartitionInfo.cs
```diff
@@ -186,6 +164,23 @@ private static string GenerateAuthToken(string verb, string resourceType, string
             return $"type=master&ver=1.0&sig={signature}";
         }
 
+        private async Task<string> GenerateAccessToken(string host, CancellationToken cancellationToken)
```
abiisnn, line 167: Can we check that **host** is not null, empty or white space?

abiisnn (commented):
Left a couple of comments, thanks for fixing it!!

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosDbCollectionPhysicalPartitionInfo.cs
abiisnn: Do we still need "[key](https://github.com/microsoft/fhir-server/blob/fb9836bb14817f50977860217ff69a034493e90d/src/Microsoft.Health.Fhir.CosmosDb.Core/Configs/CosmosDataStoreConfiguration.cs#L15)"?

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosDbCollectionPhysicalPartitionInfo.cs
v-ajajvanu: @abiisnn "aad" stands here for Azure Active Directory.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosDbCollectionPhysicalPartitionInfo.cs
v-ajajvanu: @abiisnn We are already checking for "host" as not null and empty or white space in line number 101.
Only after the check is successful, GenerateAccessToken function will get invoked.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosDbCollectionPhysicalPartitionInfo.cs
v-ajajvanu: @abiisnn We had removed "Key" reference from everywhere.
 Also, in CosmosDBLocalEmulator class, we do not need token property as we are using DefaultAzureCredentials, and it generates token internally.

v-ajajvanu:
@abiisnn We had worked on the suggestions provided and changes are committed from our end. If possible, could you please review the same and provide feedbacks.

brendankowitz:
Covered by changes in #4026. The investigation/changes for `CosmosDbCollectionPhysicalPartitionInfo` was helpful in getting this working under MI

PR #3972, opened 2024-07-18
Enhancement(import):Import should error for SearchParameter

nathanlemma:
## Description
US 121465: Import should report "SearchParameter" type resources as error
Handle the following cases
- If the type = "SearchParameter", the $import request should return HTTP 400 Bad Request
- During import processing, for mixed files, error if a search parameter is encountered.

## Related issues
Addresses [[issue #121465](https://microsofthealth.visualstudio.com/Health/_workitems/edit/121465/)].

## Testing

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Import/ImportResourceLoaderTests.cs
+++ b/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Import/ImportResourceLoaderTests.cs
@@ -183,7 +183,7 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Operations.Import
             Func<long, long> idGenerator = (i) => i;
             ImportResourceLoader loader = new ImportResourceLoader(integrationDataStoreClient, importResourceParser, serializer, NullLogger<ImportResourceLoader>.Instance);
 
-            (Channel<ImportResource> outputChannel, Task importTask) = loader.LoadResources("http://dummy", 0, (int)1e9, "SearchParameter", ImportMode.InitialLoad, CancellationToken.None);
+            (Channel<ImportResource> outputChannel, Task importTask) = loader.LoadResources("http://dummy", 0, (int)1e9, null, ImportMode.InitialLoad, CancellationToken.None);
 
             int errorCount = 0;
             await foreach (ImportResource resource in outputChannel.Reader.ReadAllAsync())
--- a/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Import/ImportResourceLoaderTests.cs
+++ b/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Import/ImportResourceLoaderTests.cs
@@ -137,6 +137,54 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Operations.Import
         }
 
         [Fact]
+        public async Task GivenResourceLoader_WhenLoadResourcesWithSearchParameterResourceType_ThenResourcesWithSearchParameterTypeShouldBeSkipped()
+        {
+            string errorMessage = "SearchParameter is not a supported resource type.";
+            using MemoryStream stream = new MemoryStream();
+            using StreamWriter writer = new StreamWriter(stream);
+            await writer.WriteLineAsync("test");
+            await writer.FlushAsync();
+
+            stream.Position = 0;
+
+            IIntegrationDataStoreClient integrationDataStoreClient = Substitute.For<IIntegrationDataStoreClient>();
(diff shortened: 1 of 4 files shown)


src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportResourceLoader.cs
```diff
@@ -151,6 +151,11 @@ private ImportResource ParseImportRawContent(string resourceType, (string Conten
                 {
                     throw new FormatException("Resource type not match.");
                 }
+
+                if (!string.IsNullOrEmpty(resourceType) && resourceType.Equals("SearchParameter", StringComparison.Ordinal))
```
mikaelweave, line 155: We need to be checking `importResource.ResourceWrapper?.ResourceTypeName` here to see if the imported resource is a SearchParameter. The logic in the controller should handle this case.

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportResourceLoader.cs
```diff
@@ -151,6 +151,11 @@ private ImportResource ParseImportRawContent(string resourceType, (string Conten
...
+                {
+                    throw new ArgumentException("SearchParameter is not a supported resource type.");
```
mikaelweave, line 157: NIT - maybe something like `SearchParameter resources cannot be processed by import.`

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Import/ImportResourceLoaderTests.cs
```diff
@@ -136,6 +136,54 @@ await foreach (ImportResource resource in outputChannel.Reader.ReadAllAsync())
             Assert.Equal(1, errorCount);
         }
 
+        [Fact]
+        public async Task GivenResourceLoader_WhenLoadResourcesWithSearchParameterResourceType_ThenResourcesWithSearchParameterTypeShouldBeSkipped()
```
mikaelweave, line 140: Given the suggested change in ImportResourceLoader, perhaps you can mock `ImportResourseParser` in the test like the `VerifyResourceLoaderAsync` method does to return a SearchParameter resource

mikaelweave (changes requested):

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/ImportResourceLoader.cs
```diff
@@ -151,6 +151,11 @@ private ImportResource ParseImportRawContent(string resourceType, (string Conten
                 {
                     throw new FormatException("Resource type not match.");
                 }
+
+                if (!string.IsNullOrEmpty(resourceType) && importResource.ResourceWrapper != null && importResource.ResourceWrapper.ResourceTypeName.Equals("SearchParameter", StringComparison.Ordinal))
```
mikaelweave, line 155: We don't need to check `resourceType` as it may be null/empty for import requests that don't specify the resource type (files that have mixed resource types).

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Import/ImportResourceLoaderTests.cs
```diff
@@ -136,6 +136,67 @@ await foreach (ImportResource resource in outputChannel.Reader.ReadAllAsync())
...
+                .Returns(callInfo =>
+                {
+                    Exception ex = (Exception)callInfo[1];
+                    return ex.Message;
+                });
+
+            Func<long, long> idGenerator = (i) => i;
+            ImportResourceLoader loader = new ImportResourceLoader(integrationDataStoreClient, importResourceParser, serializer, NullLogger<ImportResourceLoader>.Instance);
+
+            (Channel<ImportResource> outputChannel, Task importTask) = loader.LoadResources("http://dummy", 0, (int)1e9, "SearchParameter", ImportMode.InitialLoad, CancellationToken.None);
```
mikaelweave, line 186: I think we want the `resourceType` parameter which is currently set to `SearchParameter` to be null. This case should be caught in the controller so we really want to test when the input resource type is a `SearchParameter` instead.

mikaelweave (changes requested):
A couple small changes, then we'll be good 👍🏽

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Import/ImportResourceLoaderTests.cs
nathanlemma: I just pushed some changes, let me know if this addresses the scenario we need or we can sync up otherwise.

mikaelweave (approved):
Looks good!! Thanks!

PR #3977, opened 2024-07-22
Log Connection.Open() duration

SergeyGaluzo:
Fixes https://microsofthealth.visualstudio.com/Health/_workitems/edit/123215/


src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlRetry/SqlRetryService.cs
```diff
@@ -464,6 +468,7 @@ public async Task<SqlConnection> GetConnection(ISqlConnectionBuilder sqlConnecti
                 // must be set before opening connection to take effect. Therefore we must reset it to null here before opening the connection.
                 conn.RetryLogicProvider = null; // To remove this line _sqlConnectionBuilder in healthcare-shared-components must be modified.
                 await conn.OpenAsync(cancel);
+                logger.LogInformation("Opened connection to the database in {ConnectionOpenSeconds} seconds.", sw.Elapsed.TotalSeconds);
```
LTA-Thinking, line 471: Could this include if the connection is to a read or write replica? I know it doesn't have much use now, but it could be helpful in the future.

SergeyGaluzo: Added

LTA-Thinking (approved):

PR #3982, opened 2024-07-23
Change Mediator in Export to use own IServiceProvider

mikaelweave:
## Description
Wraps `_mediator` in Export with `Func<IScoped<>>` so it has its own `IServiceProvider`. This will fix exceptions where the shared `IServiceProvider` is disposed prematurely.

## Related issues
[AB#121830](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/121830)

## Testing

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
@@ -43,7 +43,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Export
         private readonly IResourceToByteArraySerializer _resourceToByteArraySerializer;
         private readonly IExportDestinationClient _exportDestinationClient;
         private readonly IResourceDeserializer _resourceDeserializer;
-        private readonly IScoped<IMediator> _mediator;
+        private readonly Func<IScoped<IMediator>> _mediatorFactory;
         private readonly RequestContextAccessor<IFhirRequestContext> _contextAccessor;
         private readonly ILogger _logger;
 
@@ -83,7 +83,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Export
             _resourceDeserializer = resourceDeserializer;
             _exportDestinationClient = exportDestinationClient;
             _anonymizerFactory = anonymizerFactory;
-            _mediator = mediatorFactory.Invoke();
+            _mediatorFactory = mediatorFactory;
             _contextAccessor = contextAccessor;
             _logger = logger;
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -83,7 +83,7 @@ public class ExportJobTask : IExportJobTask
             _resourceDeserializer = resourceDeserializer;
             _exportDestinationClient = exportDestinationClient;
             _anonymizerFactory = anonymizerFactory;
-            _mediator = mediator;
+            _mediator = mediatorFactory.Invoke();
```
LTA-Thinking, line 86: I think we should wait to invoke this until it is used. The factory is what should be stored here.

mikaelweave: Makes sense. I did this approach because I was worried the scope would be disposed already haha. But we can adjust later if needed.

mikaelweave: fixed, thanks RB 👍🏽

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ExportJobTask.cs
```diff
@@ -60,7 +60,7 @@ public class ExportJobTask : IExportJobTask
             IExportDestinationClient exportDestinationClient,
             IResourceDeserializer resourceDeserializer,
             IScoped<IAnonymizerFactory> anonymizerFactory,
-            IMediator mediator,
+            Func<IScoped<IMediator>> mediatorFactory,
```
brendankowitz, line 63: Thought/question: if the parent scope is disposed, would the child scope also be disposed here?

mikaelweave: I don't know to be honest. :( ChatGPT says:

> No, if the parent IServiceProvider is disposed, the child scope created with CreateScope will not be disposed automatically. You need to dispose of the child scope explicitly.

When prompted further:
> When you create a scope using CreateScope() on an IServiceProvider, the created scope is independent of the parent IServiceProvider. If the parent IServiceProvider is disposed, it does not automatically dispose of the child scopes. Each scope created using CreateScope() needs to be explicitly disposed of to ensure proper cleanup of any IDisposable services that were resolved within that scope.

> This behavior is confirmed in the .NET dependency injection guidelines and various discussions on Stack Overflow​ ([Stack Overflow](https://stackoverflow.com/questions/43244316/iserviceprovider-garbage-collection-disposal))​​ ([MS Learn](https://learn.microsoft.com/en-us/dotnet/core/extensions/dependency-injection-guidelines))​​ ([Stack Overflow](https://stackoverflow.com/questions/70864055/call-iservicescopefactory-createscope-in-a-scoped-service))​. For more information, you can refer to the official Microsoft documentation on dependency injection.

brendankowitz: Still not sure I trust it :P

image

mikaelweave:
This is wrong, closing.

PR #3991, opened 2024-07-28
Add database setting to reuse query plans

SergeyGaluzo:
https://microsofthealth.visualstudio.com/Health/_workitems/edit/123488/

FYI Does not contain replacement of EXISTS by JOIN.

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Storage/ProcessingFlag.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Storage/ProcessingFlag.cs
@@ -1,0 +1,70 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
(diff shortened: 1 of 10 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Storage/ProcessingFlag.cs
```diff
@@ -0,0 +1,70 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Data.SqlClient;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Health.Fhir.SqlServer.Features.Storage
+{
+    internal class ProcessingFlag<TLogger>
```
brendankowitz, line 17: Why is `ProcessingFlag` class using another class's logger?

SergeyGaluzo: I just kept current behavior when moved ProcessingFlag from private to internal. Does it hurt?

LTA-Thinking (approved):

PR #4473, opened 2024-08-28
Fixing an IsDeleted issue on import/history.

v-iyamauchi:
## Description
The PR addresses two issues.
- IsDeleted() for Bundle.EntryComponent returns false on a deleted resource imported.
- A _history response doesn't have the request and response section for a deleted resource imported.

## Related issues
Addresses [issue #120287].

[Bug 120287](https://microsofthealth.visualstudio.com/Health/_workitems/edit/120287): Bundle.EntryComponent IsDeleted() method does not return TRUE for records deleted via $import

## Testing
Tested $import/_history scenarios by locally running the service.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/BundleFactory.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/BundleFactory.cs
@@ -74,7 +74,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Search
                 if (!hasVerb && string.Equals("Import", r.Resource.Request?.Method, StringComparison.OrdinalIgnoreCase))
                 {
                     hasVerb = true;
-                    httpVerb = resource.IsDeleted() ? Bundle.HTTPVerb.DELETE : Bundle.HTTPVerb.POST;
+                    httpVerb = resource.IsDeleted() ? Bundle.HTTPVerb.DELETE : Bundle.HTTPVerb.PUT;
                 }
 
                 resource.FullUrlElement = new FhirUri(_urlResolver.ResolveResourceWrapperUrl(r.Resource));
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/BundleFactory.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Search/BundleFactory.cs
@@ -71,6 +71,12 @@ namespace Microsoft.Health.Fhir.Core.Features.Search
                 }
 #endif
 
+                if (!hasVerb && string.Equals("Import", r.Resource.Request?.Method, StringComparison.OrdinalIgnoreCase))
+                {
+                    hasVerb = true;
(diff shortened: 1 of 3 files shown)


SergeyGaluzo (commented):
It would be simpler to always trust Resource.IsDeleted flag. Can we do it?

SergeyGaluzo (changes requested):
Should the verb for not deleted resource from $import be PUT?

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/BundleFactory.cs
```diff
@@ -71,6 +71,12 @@ public ResourceElement CreateHistoryBundle(SearchResult result)
                 }
 #endif
 
+                if (!hasVerb && string.Equals("Import", r.Resource.Request?.Method, StringComparison.OrdinalIgnoreCase))
+                {
+                    hasVerb = true;
+                    httpVerb = resource.IsDeleted() ? Bundle.HTTPVerb.DELETE : Bundle.HTTPVerb.PUT;
```
brendankowitz, line 77: Should/does the import code do this logic now too?

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/RawBundleEntryComponent.cs
```diff
@@ -15,11 +15,14 @@ namespace Microsoft.Health.Fhir.Shared.Core.Features.Search
...
     public class RawBundleEntryComponent : Bundle.EntryComponent
     {
+        private readonly ResourceWrapper _resourceWrapper;
+
         public RawBundleEntryComponent(ResourceWrapper resourceWrapper)
         {
             EnsureArg.IsNotNull(resourceWrapper, nameof(resourceWrapper));
 
             ResourceElement = new RawResourceElement(resourceWrapper);
+            _resourceWrapper = resourceWrapper;
```
brendankowitz, line 25: `_resourceWrapper` and `ResourceElement` represent the same object but `ResourceElement` isn't immutable. Are there any references where these could get out of sync?

v-iyamauchi: Fixed by changing to save just the isDeleted status.

SergeyGaluzo (approved):

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/BundleFactory.cs
SergeyGaluzo: Currently $import saves all its resources with Resource.Method=Import. This is nice as it allows to distinguish records from $import and API.

v-iyamauchi:
> Should the verb for not deleted resource from $import be PUT?

Fixed as suggested.

rajithaalurims (approved):

PR #4654, opened 2024-09-27
Update AAD token request target resource uri for ACR access

ShaunDonn2:
## Description
In alignment with [this documentation](https://dev.azure.com/msazure/AzureContainerRegistry/_wiki/wikis/ACR%20Specs/480000/TrustedServicesPatterns), we should use the ACR-specific target resource URI rather than the generic ARM URI in order to get the "xms_az_tm" claim on the token that we need to authenticate as a Trusted Service with ACR. This value is passed in and used as the scopes value for the token request in the [AuthenticationProvider](https://microsofthealth.visualstudio.com/Health/_git/health-external-mi?path=/src/Authentication/Providers/AuthenticationProvider.cs&version=GBmain&line=100&lineEnd=100&lineStartColumn=17&lineEndColumn=68&lineStyle=plain&_a=contents).

This PR effectively updates the name of the config field as "armResourceManagerId" is not descriptive of the purpose of this field, which is to determine the target resource URI for getting access tokens for the ACR. The "management.azure.com" value will still work here, so we will keep it as the default so as not to break any OSS or other scenarios. We will override this value in our PaaS code to point to the ACR-specific value which enabled Trusted Services scenarios.

## Related issues
Addresses bug [126662](https://microsofthealth.visualstudio.com/Health/_workitems/edit/126662).

## Testing
Validated this change by passing in the values of a prod Credential Bundle and the updated target resource URI into the token request in the [AuthenticationProvider](https://microsofthealth.visualstudio.com/Health/_git/health-external-mi?path=/src/Authentication/Providers/AuthenticationProvider.cs&version=GBmain&line=87&lineEnd=104&lineStartColumn=9&lineEndColumn=1&lineStyle=plain&_a=contents). With the generic ARM URI, the token comes back without the xms_az_tm claim; with the ACR-specific URI, the request comes back with the claim included.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Azure/ContainerRegistry/AzureContainerRegistryAccessTokenProvider.cs
+++ b/src/Microsoft.Health.Fhir.Azure/ContainerRegistry/AzureContainerRegistryAccessTokenProvider.cs
@@ -59,7 +59,7 @@ namespace Microsoft.Health.Fhir.Azure.ContainerRegistry
         {
             EnsureArg.IsNotNullOrEmpty(registryServer, nameof(registryServer));
 
-            var aadResourceUri = new Uri(_convertDataConfiguration.AcrTargetResourceUri);
+            var aadResourceUri = _convertDataConfiguration.AcrTargetResourceUri;
             string aadToken;
             try
             {
--- a/src/Microsoft.Health.Fhir.Azure/ContainerRegistry/AzureContainerRegistryAccessTokenProvider.cs
+++ b/src/Microsoft.Health.Fhir.Azure/ContainerRegistry/AzureContainerRegistryAccessTokenProvider.cs
@@ -33,6 +33,8 @@ namespace Microsoft.Health.Fhir.Azure.ContainerRegistry
         private const string ExchangeAcrRefreshTokenUrl = "oauth2/exchange";
         private const string GetAcrAccessTokenUrl = "oauth2/token";
 
+        private static readonly Uri AcrTargetResourceUri = new Uri("https://containerregistry.azure.net/");
+
         private readonly IAccessTokenProvider _aadTokenProvider;
         private readonly IHttpClientFactory _httpClientFactory;
         private readonly ConvertDataConfiguration _convertDataConfiguration;
@@ -59,7 +61,7 @@ namespace Microsoft.Health.Fhir.Azure.ContainerRegistry
         {
             EnsureArg.IsNotNullOrEmpty(registryServer, nameof(registryServer));
 
-            var aadResourceUri = _convertDataConfiguration.AcrTargetResourceUri;
+            var aadResourceUri = AcrTargetResourceUri;
             string aadToken;
             try
             {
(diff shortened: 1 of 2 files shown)


rainyjyu:
Does the ACR audience need to be different for GovCloud?

rainyjyu (commented):
Does the ACR audience need to be parameterized in config for GovCloud or other non-PublicCloud environments?

rainyjyu (dismissed):

ShaunDonn2:
> Does the ACR audience need to be parameterized in config for GovCloud or other non-PublicCloud environments?

I was planning on doing this in the Health-PaaS repo, where we had the old ARM value parameterized, but I just saw in the docs that the value I have added is constant across Azure Clouds. I will still need to update the value in Health-PaaS, but it seems that we can have the same value for all environments.

abiisnn (dismissed):

src/Microsoft.Health.Fhir.Azure/ContainerRegistry/AzureContainerRegistryAccessTokenProvider.cs
```diff
@@ -59,7 +59,7 @@ public async Task<string> GetTokenAsync(string registryServer, CancellationToken
         {
             EnsureArg.IsNotNullOrEmpty(registryServer, nameof(registryServer));
 
-            var aadResourceUri = new Uri(_convertDataConfiguration.ArmResourceManagerId);
+            var aadResourceUri = _convertDataConfiguration.AcrTargetResourceUri;
```
ms-teli, line 62: Is it the only place where it has reference?

ms-teli (approved):

src/Microsoft.Health.Fhir.Azure/ContainerRegistry/AzureContainerRegistryAccessTokenProvider.cs
ShaunDonn2: @ms-teli yes, this is the only place in this code, but thinking more about it, I think we should keep the ArmResourceManagerId in the ConvertDataConfiguration and just add this new property, rather than swapping them. That lowers the risk of a breaking change and also can be useful if we need that value in future.

abiisnn (approved):

src/Microsoft.Health.Fhir.Azure/ContainerRegistry/AzureContainerRegistryAccessTokenProvider.cs
```diff
@@ -33,6 +33,8 @@ public class AzureContainerRegistryAccessTokenProvider : IContainerRegistryToken
         private const string ExchangeAcrRefreshTokenUrl = "oauth2/exchange";
         private const string GetAcrAccessTokenUrl = "oauth2/token";
 
+        private static readonly Uri AcrTargetResourceUri = new Uri("https://containerregistry.azure.net/");
```
ms-teli, line 36: It would make more sense to add a property in the ConvertDataConfiguration for this const value: "https://containerregistry.azure.net/" if it won't break anything.

ShaunDonn2: @ms-teli after speaking with Dustin, I am going to keep the value as the default "management.azure.com" value in order to avoid breaking any OSS scenarios. We will override this value in PaaS for the ACR-specific URI. I created a new field name in the config because "armResourceManagerId" does not make sense given the new URI value and its purpose (to acquire a token for ACR).

ms-teli (approved):

abiisnn (approved):

PR #4663, opened 2024-10-09
Allow reindex with unlimited empty pages

mikaelweave:
## Description
Removes artificial limitation of reindex where there is large historical datasets or customer wants limited query size.

Customer has tested my reindex fix from the last release and it's not quite right. It doesn't allow for many sequential historical/deleted resources with default query size. The reindex still says completed incorrectly in this case. This change adds an outer loop to continue getting surrogate ranges until there is one w/ data.

## Related issues
[AB#124883](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/124883)

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
@@ -926,8 +926,9 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Search
             }
 
             var queryHints = searchOptions.QueryHints;
-            long startId = long.Parse(queryHints.First(_ => _.Param == KnownQueryParameterNames.StartSurrogateId).Value);
-            long endId = long.Parse(queryHints.First(_ => _.Param == KnownQueryParameterNames.EndSurrogateId).Value);
+            long globalStartId = long.Parse(queryHints.First(_ => _.Param == KnownQueryParameterNames.StartSurrogateId).Value);
+            long globalEndId = long.Parse(queryHints.First(_ => _.Param == KnownQueryParameterNames.EndSurrogateId).Value);
+            long queryStartId = globalStartId;
 
             SearchResult results = null;
             IReadOnlyList<(long StartId, long EndId)> ranges;
@@ -935,7 +936,7 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Search
             do
             {
                 // Get surrogate ID ranges
-                ranges = await GetSurrogateIdRanges(resourceType, startId, endId, searchOptions.MaxItemCount, 50, true, cancellationToken);
+                ranges = await GetSurrogateIdRanges(resourceType, queryStartId, globalEndId, searchOptions.MaxItemCount, 50, true, cancellationToken);
 
                 foreach (var range in ranges)
                 {
@@ -958,11 +959,14 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Search
 
                     _logger.LogInformation("For Reindex, empty data page encountered. Resource Type={ResourceType} StartId={StartId} EndId={EndId}", resourceType, range.StartId, range.EndId);
                 }
+
+                // If no resources are found in the group of surrogate id ranges, move forward the starting point.
+                queryStartId = ranges[-1].EndId + 1;
             }
             while (ranges is not null && ranges.Count > 0); // Repeat until there are no more ranges to scan. Needed to advance through large contigous history.
 
             // Return empty result when no resources are found in the given range provided by queryHints.
-            _logger.LogInformation("No surrogate ID ranges found. Resource Type={ResourceType} StartId={StartId} EndId={EndId}", resourceType, startId, endId);
+            _logger.LogInformation("No surrogate ID ranges found. Resource Type={ResourceType} StartId={StartId} EndId={EndId}", resourceType, globalStartId, globalEndId);
             return new SearchResult(0, []);
         }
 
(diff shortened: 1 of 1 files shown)


SergeyGaluzo (dismissed):

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -930,12 +930,16 @@ protected async override Task<SearchResult> SearchForReindexInternalAsync(Search
...
 
             SearchResult results = null;
-            IReadOnlyList<(long StartId, long EndId)> ranges = await GetSurrogateIdRanges(resourceType, startId, endId, searchOptions.MaxItemCount, 50, true, cancellationToken);
+            IReadOnlyList<(long StartId, long EndId)> ranges;
 
-            if (ranges?.Count > 0)
+            do
             {
+                // Get surrogate ID ranges
+                ranges = await GetSurrogateIdRanges(resourceType, startId, endId, searchOptions.MaxItemCount, 50, true, cancellationToken);
```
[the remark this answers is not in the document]
SergeyGaluzo: +1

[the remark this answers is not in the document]
mikaelweave: Good catch - added logic to use a query start id that is advanced at the end of the do...while loop.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -948,20 +953,24 @@ protected async override Task<SearchResult> SearchForReindexInternalAsync(Search
...
+                        return results;
                     }
 
                     _logger.LogInformation("For Reindex, empty data page encountered. Resource Type={ResourceType} StartId={StartId} EndId={EndId}", resourceType, range.StartId, range.EndId);
                 }
+
+                // If no resources are found in the group of surrogate id ranges, move forward the starting point.
+                if (ranges.Any())
+                {
+                    queryStartId = ranges.Max(x => x.EndId) + 1;
```
brendankowitz, line 966: Do we need to be careful of queryStartId >= globalEndId? (would the +1 make this possible)

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -926,16 +926,21 @@ protected async override Task<SearchResult> SearchForReindexInternalAsync(Search
             }
 
             var queryHints = searchOptions.QueryHints;
-            long startId = long.Parse(queryHints.First(_ => _.Param == KnownQueryParameterNames.StartSurrogateId).Value);
-            long endId = long.Parse(queryHints.First(_ => _.Param == KnownQueryParameterNames.EndSurrogateId).Value);
+            long globalStartId = long.Parse(queryHints.First(_ => _.Param == KnownQueryParameterNames.StartSurrogateId).Value);
+            long globalEndId = long.Parse(queryHints.First(_ => _.Param == KnownQueryParameterNames.EndSurrogateId).Value);
```
brendankowitz, line 930: :see_no_evil: nit: using `_` has a convention of being the same name as [discard](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/discards) variables

mikaelweave: I only changed the variable names here for clarity but am happy to fix this in another PR if needed. 😄

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
mikaelweave: No - these get translated to a `BETWEEN` in the SQL stored procedure. If start > end this will return no results from SQL.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -948,20 +953,24 @@ protected async override Task<SearchResult> SearchForReindexInternalAsync(Search
...
-                        break;
+                        _logger.LogInformation("For Reindex, Resource Type={ResourceType} Count={Count} MaxResourceSurrogateId={MaxResourceSurrogateId}", resourceType, results.TotalCount, results.MaxResourceSurrogateId);
+                        return results;
                     }
 
                     _logger.LogInformation("For Reindex, empty data page encountered. Resource Type={ResourceType} StartId={StartId} EndId={EndId}", resourceType, range.StartId, range.EndId);
                 }
+
+                // If no resources are found in the group of surrogate id ranges, move forward the starting point.
+                if (ranges.Any())
```
brendankowitz, line 964: Another question :) based on the check below: `while (ranges is not null`

Does the nullability apply here too?
```suggestion
                if (ranges?.Any() == true)
```

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
mikaelweave: Actually fixed this since pipelines need to run again. 👍🏽

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
mikaelweave: nullability doesn't apply to either, good catch. removed from the while statement. 👍🏽

brendankowitz: I think consistency, null is either ok or it isn't :)

PR #4686, opened 2024-10-17
add ordering to reindex surrogate id fetch

mikaelweave:
## Description
When an empty page is encountered w/ reindex, the next page processed isn't the next page - it skips records. The order by ensures pages of data are processed in order.

## Related issues
[AB#130486](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/130486)

## Testing
Debugging database, seeing unordered list causing data to be skipped. Manual testing on DB planned.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
@@ -936,7 +936,10 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Search
             do
             {
                 // Get surrogate ID ranges
-                ranges = await GetSurrogateIdRanges(resourceType, queryStartId, globalEndId, searchOptions.MaxItemCount, 50, true, cancellationToken);
+                ranges = await GetSurrogateIdRanges(resourceType, queryStartId, globalEndId, searchOptions.MaxItemCount, 10, true, cancellationToken);
+
+                // Order the ranges by start id as they come back unordered. This ensures records aren't skipped during reindex.
+                ranges = ranges.OrderBy(x => x.StartId).ToList();
 
                 foreach (var range in ranges)
                 {


src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -936,7 +936,10 @@ protected async override Task<SearchResult> SearchForReindexInternalAsync(Search
             do
             {
                 // Get surrogate ID ranges
-                ranges = await GetSurrogateIdRanges(resourceType, queryStartId, globalEndId, searchOptions.MaxItemCount, 50, true, cancellationToken);
+                ranges = await GetSurrogateIdRanges(resourceType, queryStartId, globalEndId, searchOptions.MaxItemCount, 10, true, cancellationToken);
```
LTA-Thinking, line 939: Why reduce the number of ranges to 10?

brendankowitz (approved):

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
mikaelweave: My thinking behind this is the "happy path" for most reindex operations doesn't encounter empty data ranges. Reducing the ranges to 10 will slightly reduce the CPU consumption used by general reindex operations while also not querying the database for every range.

I'm not solid on these numbers - since this is temporary code I did what felt right vs any perf tested. 

For a standard reindex which looks at 100 records at a time, there would need to be 1000 concurrent records of only history to activate the "do/while" part of the loop w/ additional database queries. 1000 seems like a round number and sorting 10 records seems like a lower CPU lift.

mikaelweave: I think the term [SWAG](https://en.wikipedia.org/wiki/Scientific_wild-ass_guess) is appropriate.

LTA-Thinking (approved):

mikaelweave:
CleanShot 2024-10-17 at 13 43 11

Confirmed by reindex on problematic db

PR #4698, opened 2024-10-24
Add REST examples for patching security tag

mikaelweave:
## Description
Adds examples created during investigation of patching to security tag

## Related issues
[AB#129850](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/129850)

## Testing
Local rest client

### DIFF
--- a/docs/rest/FhirPatchRequests.http
+++ b/docs/rest/FhirPatchRequests.http
@@ -338,3 +338,64 @@ Authorization: Bearer {{bearer.response.body.access_token}}
         }
     ]
 }
+
+### Add a security tag. This will always add the elements to the end of the array if the array exists or not.
+PATCH https://{{hostname}}/Patient/{{patient.response.body.id}}
+content-type: application/fhir+json
+Authorization: Bearer {{bearer.response.body.access_token}}
+
+{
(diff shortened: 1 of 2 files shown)


docs/rest/FhirPatchRequests.http
```diff
@@ -337,4 +337,65 @@ Authorization: Bearer {{bearer.response.body.access_token}}
...
+          "valueCoding": {
+            "system": "http://example.org/security-system",
+            "code": "SECURITY_TAG_CODE",
+            "display": "New Security Tag Display"
+          }
+        }
+      ]
+    },
+    {
+      "name": "operation",
```
LTA-Thinking, line 376: What does having 2 values add to the example?

mikaelweave: It shows that one operation doesn't replace the array if it already exists. Would it make sense in two requests?

LTA-Thinking: Ah I see. That makes sense.

LTA-Thinking (approved):

PR #4702, opened 2024-10-25
Fix Cancel Import Request

abiisnn:
## Description
Currently, the Orchestrator job completes before the processing jobs are completed, it will incorrectly not allow cancel of an import because it thinks the job is done. This is incorrect as the Orchestrator is done but not the child Processing Jobs,.

This PR is fixing this issue, now we check all jobs to check the status of all jobs.

## Related issues
Addresses [issue #124093](https://microsofthealth.visualstudio.com/Health/_workitems/edit/124093/)

## Testing
Added a test to make sure it first returns Accepted status to cancel request, which means the import job has been cancelled, and then make sure we received Conflict status if we have already cancelled the job.

The existing test was making sure that the cancel operation works if all jobs have already completed (with just one patient) it was happening.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Import/CancelImportRequestHandlerTests.cs
+++ b/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Import/CancelImportRequestHandlerTests.cs
@@ -74,7 +74,8 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Operations.BulkImport
         {
             List<JobInfo> jobs = await SetupAndExecuteCancelImportAsync(jobStatus, HttpStatusCode.Accepted);
 
-            await _queueClient.Received(1).CancelJobByGroupIdAsync((byte)Core.Features.Operations.QueueType.Import, jobs[0].Id, _cancellationToken);
+            var groupJobId = jobs[0].GroupId;
+            await _queueClient.Received(1).CancelJobByGroupIdAsync((byte)Core.Features.Operations.QueueType.Import, groupJobId, _cancellationToken);
         }
 
         [Fact]
--- a/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Import/CancelImportRequestHandlerTests.cs
+++ b/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Import/CancelImportRequestHandlerTests.cs
@@ -70,7 +70,7 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Operations.BulkImport
         [InlineData(JobStatus.Cancelled)]
         public async Task GivenAFhirMediator_WhenCancelingExistingBulkImportJobThatHasAlreadyBeenCanceled_ThenAcceptedStatusCodeShouldBeReturned(JobStatus taskStatus)
         {
-            List<JobInfo> jobs = SetupBulkImportJob(taskStatus, true);
+            SetupBulkImportJob(taskStatus, true);
             CancelImportResponse response = await _mediator.CancelImportAsync(JobId, _cancellationToken);
 
             Assert.NotNull(response);
--- a/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Import/CancelImportRequestHandlerTests.cs
+++ b/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Import/CancelImportRequestHandlerTests.cs
@@ -4,6 +4,7 @@
 // -------------------------------------------------------------------------------------------------
 
 using System;
+using System.Collections.Generic;
 using System.Net;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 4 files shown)


test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Import/ImportTests.cs
```diff
@@ -1507,6 +1507,63 @@ public async Task GivenImportWithCancel_ThenTaskShouldBeCanceled()
...
+                InputFormat = "application/fhir+ndjson",
+                InputSource = new Uri("https://other-server.example.org"),
+                StorageDetail = new ImportRequestStorageDetail() { Type = "azure-blob" },
+                Input = patients,
+                Mode = ImportMode.InitialLoad.ToString(),
+            };
+
+            Uri checkLocation = await ImportTestHelper.CreateImportTaskAsync(_client, request);
+
+            await Task.Delay(TimeSpan.FromMinutes(1));
```
mikaelweave, line 1545: We could reduce test wait time by polling the Content-Location and wait for a body to be returned.

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/CancelImportRequestHandler.cs
```diff
@@ -43,19 +45,41 @@ public async Task<CancelImportResponse> Handle(CancelImportRequest request, Canc
...
+                if (job.Status != JobStatus.Completed)
+                {
+                    allComplete = false;
+                }
+            }
+
+            // If the job is already completed, failed or cancelled, return conflict status.
+            if (conflict || allComplete)
             {
                 throw new OperationFailedException(Core.Resources.ImportOperationCompleted, HttpStatusCode.Conflict);
```
abiisnn, line 76: I tried to use `return new CancelImportResponse(HttpStatusCode.Conflict);` but looks like it process the request as 500 instead of 400, is it correct given import behavior?

abiisnn (commented):
/azp run

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Import/CancelImportRequestHandlerTests.cs
```diff
@@ -66,14 +66,26 @@ public async Task GivenAFhirMediator_WhenCancelingExistingBulkImportJobThatHasAl
             Assert.Equal(HttpStatusCode.Conflict, operationFailedException.ResponseStatusCode);
         }
 
+        [Theory]
+        [InlineData(JobStatus.Cancelled)]
+        public async Task GivenAFhirMediator_WhenCancelingExistingBulkImportJobThatHasAlreadyBeenCanceled_ThenAcceptedStatusCodeShouldBeReturned(JobStatus taskStatus)
```
abiisnn, line 71: This is new behavior, because if a tastk has been already canceled, then we will return accepted because we will continue trying to cancel.

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/CancelImportRequestHandler.cs
```diff
@@ -43,19 +45,43 @@ public async Task<CancelImportResponse> Handle(CancelImportRequest request, Canc
...
+            // We need to check the status of all jobs
+            IReadOnlyList<JobInfo> jobs = await _queueClient.GetJobByGroupIdAsync(QueueType.Import, request.JobId, false, cancellationToken);
 
-            if (jobInfo == null)
+            if (jobs == null || jobs.Count == 0)
             {
                 throw new ResourceNotFoundException(string.Format(Core.Resources.ImportJobNotFound, request.JobId));
             }
 
-            if (jobInfo.Status == JobManagement.JobStatus.Completed || jobInfo.Status == JobManagement.JobStatus.Cancelled || jobInfo.Status == JobManagement.JobStatus.Failed)
```
abiisnn, line 53: Please check that now we are not returning CONFLICT if the job has been already cancelled.

Now, if the job has been already cancelled, we will retry cancel as Sergey suggested.

SergeyGaluzo:
I don't think jobs == null condition needs to be checked.

SergeyGaluzo:
I think it would be easier to use Linq instead of foreach.

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Import/CancelImportRequestHandlerTests.cs
mikaelweave: A change in behavior may require giving customers 1 month notice before deploying to prod. We should check w./ PM team

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/CancelImportRequestHandler.cs
mikaelweave: just need to check w/ PM on behavior change

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Import/ImportTests.cs
```diff
@@ -1494,17 +1494,68 @@ public async Task GivenImportWithCancel_ThenTaskShouldBeCanceled()
...
+                        Type = "Patient",
+                    },
+                },
+                Mode = ImportMode.InitialLoad.ToString(),
+            };
+
+            Uri checkLocation = await ImportTestHelper.CreateImportTaskAsync(_client, request);
+
+            // Wait for import job to complete
+            await Task.Delay(TimeSpan.FromSeconds(10));
```
mikaelweave, line 1542: I worry this could create a flake test in the future. 

Could we check the status of the import before moving forward in the test?

We could even loop and wait again if it's not complete (up to 5 times or so).

mikaelweave (approved):
approved with some suggestions

mikaelweave (approved):

abiisnn (commented):
/azp run

mikaelweave (approved):

PR #4708, opened 2024-10-31
Truncate ref resource id to fit into 64 bytes

SergeyGaluzo:


src/Microsoft.Health.Fhir.SqlServer/Features/Storage/TvpRowGeneration/Merge/ReferenceSearchParamListRowGenerator.cs
```diff
@@ -24,7 +28,7 @@ internal override bool TryGenerateRow(short resourceTypeId, long resourceRecordI
                 searchParamId,
                 searchValue.BaseUri?.ToString(),
                 searchValue.ResourceType == null ? null : Model.GetResourceTypeId(searchValue.ResourceType),
-                searchValue.ResourceId,
+                searchValue.ResourceId[..Math.Min(searchValue.ResourceId.Length, _maxLength)], // Truncate to fit the column size
```
LTA-Thinking, line 31: We can't do this. Truncating a search value will lead to unintuitive search behavior which could result in customers getting records when they don't want them or not getting them when they do. I think we should be throwing an error in this case. It is a breaking change, but I think it is better than trying to work in a way the customer won't expect.

SergeyGaluzo: Implicit truncation is happening today. So this PR is not changing this behavior. It just prevents failures. Also we are still able to search, we just return more than expected sometimes. I will add a comment to PR.

LTA-Thinking (approved):

PR #4715, opened 2024-11-07
[Bundle] Simplify profile refresh during bundle execution

fhibf:
## Description
Simplify profile refresh during bundle execution. 
In case a bundle contains changes on ValueSet, StructureDefinition and/or CodeSystem, no profile refreshes are executed until the bundle is completed. 

This strategy will improve the performance of bundle when handling changes on these resource types. 

Bundles were previously delayed due the multiple refreshes and changes on the state of the service. 

## Related issues
[AB#132200](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/132200)

## Testing
Testing the improvement, it's noticeable the gain in performance. 
With the issue, one of the tests took 160 seconds to execute, and with the fix it only took 2.4 seconds to process the entire bundle. 

image

## FHIR Team Checklist


src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
```diff
@@ -220,9 +219,12 @@ await Parallel.ForEachAsync(resources, parallelOptions, async (resource, innerCt
...
         public async Task<UpsertOutcome> UpsertAsync(ResourceWrapperOperation resource, CancellationToken cancellationToken)
         {
-            bool isBundleOperation = _bundleOrchestrator.IsEnabled && resource.BundleResourceContext != null;
+            bool isBundleParallelOperation =
+                _bundleOrchestrator.IsEnabled &&
+                resource.BundleResourceContext != null &&
+                resource.BundleResourceContext.IsParallelBundle;
 
-            if (isBundleOperation)
+            if (isBundleParallelOperation)
```
feordin, line 227: I am not sure I understand this change.  Why do we check this only for parallel bundles if before it applied whether or not the bundle was a parallel bundle?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -828,14 +839,23 @@ private static void SetupContexts(
...
+                }
 
-                // Assign the HTTP Verb operation associated with the request as part of the downstream request.
-                newFhirRequestContext.RequestHeaders.Add(BundleOrchestratorNamingConventions.HttpHeaderBundleResourceHttpVerb, httpVerb.ToString());
+                // Assign the current Bundle Orchestrator Operation ID as part of the downstream request.
+                newFhirRequestContext.RequestHeaders.Add(BundleOrchestratorNamingConventions.HttpInnerBundleRequestHeaderOperationTag, bundleOrchestratorOperation.Id.ToString());
+            }
+            else
+            {
+                // Assign an empty Bundle Orchestrator Operation ID.
```
feordin, line 857: Do we not need an orchestrator id for sequential bundles?

fhibf: Nope. Sequential bundles do not rely on bundle orchestrator. 
Bundle orchestrator is a concept only used by parallel bundles.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
fhibf: Previously, BundleResourceContext was a concept only used by parallel bundles. 

The use of BundleResourceContext had to be expanded to support other parts of the service to identify if the requests were executed under the context of a bundle or not. 

My changes are expanding BundleResourceContext to all types of bundles (sequential and parallel), but it's now differentiated by the addition of the ProcessingLogic property to BundleResourceContext.

That's why I added the validation: to validate that the context is requesting the bundle to be executed in parallel, otherwise it should follow the previous logic and run sequentially.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
feordin: Ok!  Thanks.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
feordin: Sounds good!

feordin (approved):

PR #4737, opened 2024-11-21
Add new SQL hash calculation

LTA-Thinking:
## Description
The SQL query hash has been including parameter values when it shouldn't, which is causing custom queries to be parameter value specific. This is undesired as custom queries should apply regardless of parameter values. This PR changes the way query hashes are generated to remove parameter values.

This will change the hashes of all requests, which will render current custom queries invalid.

## Related issues
Addresses [Bug 131113](https://microsofthealth.visualstudio.com/Health/_workitems/edit/131113): SQL query hashes have parameter value hashes included

## Testing

## FHIR Team Checklist


feordin (dismissed):

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlSearchOptions.cs
```diff
@@ -39,5 +44,26 @@ public SqlSearchOptions(SearchOptions searchOptions)
         /// Performs a shallow clone of this instance
         /// </summary>
         public SqlSearchOptions CloneSqlSearchOptions() => (SqlSearchOptions)MemberwiseClone();
+
+        public string GetHash()
+        {
+            // I don't think UnsupportedSearchParams is needed as it isn't used in the actual search
```
abiisnn, line 50: Is this comment meant to be committed? 🍒

LTA-Thinking: I should reword it, but I think the point is helpful to understand why I didn't include that value in the hash.

LTA-Thinking:
Waiting to merge until we can discuss the impact to existing custom queries.

PR #4745, opened 2024-12-06
Better exception handling 

fhibf:
## Description
Adding better exception handling for a few situations: 
* Extending _PreconditionFailedException_ to support inner exceptions. 
* Raising _PreconditionFailedException_ when an operation fails due a partition reaching 20GB. That should replace HTTP500s with HTTP400s, indicating to customer that a failure on their end caused the issue. 
* Adding hew FHIR Exception type called _TransactionFailureException_ to replace odd _InvalidOperationExceptions_ raised in case of Cosmos DB transaction failures. 
* Adding more logs for when FHIR profiles are refreshed by _BundleHandler_. Additionally, logging for how long the refresh ran and handling cases when the refresh fails. Fails during the profile refresh should not affect the response of bundles. 
* Handling null reference exceptions at Bundle _SanitizeString_. I noticed those errors recently, and I want to avoid the noise caused by these errors. These errors are not currently affecting the execution of bundles. 

## Related issues
[AB#134896](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/134896).

## Testing


src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
```diff
@@ -199,6 +200,17 @@ await Parallel.ForEachAsync(resources, parallelOptions, async (resource, innerCt
                     var result = new DataStoreOperationOutcome(upsertOutcome);
                     results.AddOrUpdate(identifier, _ => result, (_, _) => result);
                 }
+                catch (CosmosException cme) when (cme.StatusCode == HttpStatusCode.Forbidden && cme.SubStatusCode == 1014)
+                {
+                    // https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/troubleshoot-forbidden#partition-key-exceeding-storage
+
+                    var exceptionWithNoTechnicalDetails = new PreconditionFailedException("Partition reached maximum size.");
+                    var exceptionForLogging = new PreconditionFailedException(exceptionWithNoTechnicalDetails.Message, cme);
```
LTA-Thinking, line 208: Why are two exceptions needed here? We don't strip out the technical details in other exceptions, they aren't included in the response message.

fhibf: Thanks for confirming @LTA-Thinking . I'm gonna refactor that part. My concern in this scenario was about sharing too much information.

LTA-Thinking (dismissed):

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
```diff
@@ -204,12 +204,8 @@ await Parallel.ForEachAsync(resources, parallelOptions, async (resource, innerCt
...
                     // https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/troubleshoot-forbidden#partition-key-exceeding-storage
 
-                    var exceptionWithNoTechnicalDetails = new PreconditionFailedException("Partition reached maximum size.");
-                    var exceptionForLogging = new PreconditionFailedException(exceptionWithNoTechnicalDetails.Message, cme);
-
-                    _logger.LogWarning(exceptionForLogging, "PreconditionFailed: Cosmos DB partition reached max size of 20GB.");
-
-                    results.TryAdd(identifier, new DataStoreOperationOutcome(exceptionWithNoTechnicalDetails));
+                    _logger.LogWarning(cme, "PreconditionFailed: Partition reached maximum size.");
+                    results.TryAdd(identifier, new DataStoreOperationOutcome(new PreconditionFailedException("Partition reached maximum size.")));
```
brendankowitz, line 208: Exception messages should be in Resource files if possible

feordin (approved):

PR #4752, opened 2024-12-10
Disable JobHosting workers for disabled jobs

mikaelweave:
## Description
- Adds method to disable JobHosting queues for export and import if they are disabled or have no storage account configured.

## Related issues
[AB#134785](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/134785)

## Testing
- Added tests
- Debugged during startup to confirm behavior.

## FHIR Team Checklist


src/Microsoft.Health.Fhir.Core/Configs/OperationsConfiguration.cs
```diff
@@ -22,5 +24,27 @@ public class OperationsConfiguration
         public IntegrationDataStoreConfiguration IntegrationDataStore { get; set; } = new IntegrationDataStoreConfiguration();
 
         public ImportTaskConfiguration Import { get; set; } = new ImportTaskConfiguration();
+
+        /// <summary>
+        /// Removes queues based on the enabled status of the operations.
+        /// </summary>
+        public void RemoveDisabledQueues()
```
brendankowitz, line 31: Maybe a silly question, why do we add them then remove instead of skipping to add?

mikaelweave: I thought the same thing initially. 😆 The main reasoning is this approach allows default configuration applied in `appsettings.json` which is parsed to the configuration objects. I think this is the correct approach vs applying overrides via environment variables/appsettings overrides in the hosting environment which have to be maintained separately.

So once we parse the defaults into the configuration class, why have the removal logic here? I'm thinking keeping it in this class encapsulates the logic here vs polluting the registration classes with operation specific logic. It just exposes a hook to be called.

I was on the fence about implementation on this - happy to change the approach if there are thought a different direction.

mikaelweave: @brendankowitz - I could use a custom `ConfigurationProvider` instead like below. Thoughts?? I think this would intercept the configuration of the class.

https://github.com/microsoft/fhir-server/blob/cb53a355a88c50456c4542484a96512642b54b1a/src/Microsoft.Health.Fhir.Api/Features/Binders/DictionaryExpansionConfigurationProvider.cs#L22

brendankowitz: In thinking about it, perhaps this is calling more to consolidate these settings into one place. Right now we enable/disable in one place, configure job settings in another and remove as a post step. It might make more sense to move all these into a single config which would make this more discoverable / maintainable

src/Microsoft.Health.Fhir.Core/Configs/ImportJobConfiguration.cs
```diff
@@ -3,19 +3,21 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
+using Microsoft.Health.Fhir.Core.Features.Operations;
+
 namespace Microsoft.Health.Fhir.Core.Configs
 {
-    public class ImportTaskConfiguration // This class name is inconistent with others which use Job instrad of Task.
+    public class ImportJobConfiguration : HostingBackgroundServiceQueueItem
```
mikaelweave, line 10: **Author Context:** This class was renamed due to the removed comment - we are aligning the names of our configuration as cleanup as part of this PR.

src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/ImportOrchestratorJob.cs
```diff
@@ -45,7 +45,7 @@
         private readonly IMediator _mediator;
         private readonly RequestContextAccessor<IFhirRequestContext> _contextAccessor;
         private readonly IQueueClient _queueClient;
-        private ImportTaskConfiguration _importConfiguration;
+        private ImportJobConfiguration _importConfiguration;
```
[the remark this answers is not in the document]
mikaelweave: not in scope for rename

src/Microsoft.Health.Fhir.Api/Features/Operations/Import/InitialImportLockMiddleware.cs
```diff
@@ -20,7 +20,7 @@
     public sealed class InitialImportLockMiddleware
     {
         private RequestDelegate _next;
-        private ImportTaskConfiguration _importTaskConfiguration;
+        private ImportJobConfiguration _importTaskConfiguration;
```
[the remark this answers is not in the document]
mikaelweave: not in scope for rename

src/Microsoft.Health.Fhir.Api/Features/BackgroundJobService/HostingBackgroundService.cs
```diff
@@ -72,10 +74,12 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
                     await Task.Delay(TimeSpan.FromSeconds(1), cancellationTokenSource.Token);
                 }
 
-                foreach (var operation in _operationsConfiguration.HostingBackgroundServiceQueues)
+                var enabledQueueConfigs = GetEnabledQueueConfigs();
+
+                foreach (var queueConfig in enabledQueueConfigs)
```
mikaelweave, line 79: **Author Context:** Look below at `OperationsConfiguration.cs`. We are combining the job hosting queue configuration with the job configuration as @brendankowitz suggested via inheritance. Due to this, we are using reflection to find the jobs to host vs the list.

PTaladay (approved):

src/Microsoft.Health.Fhir.Core/Configs/OperationsConfiguration.cs
mikaelweave: Fixed!! :)

PR #4785, opened 2025-01-17
Add SQL performance option

LTA-Thinking:
## Description
Adds support for the x-ms-query-cache header for Gen2. This gives the user the option to control if SQL uses query caching on the request level. This can enable, disable, or have SQL run both and use the better performing one. This helps in situations where a user has some queries that perform better with query caching and some that perform better without it.

## Related issues
Addresses [User Story 137387](https://microsofthealth.visualstudio.com/Health/_workitems/edit/137387): DDMS: SQL search with and without cache

## Testing

## FHIR Team Checklist

Patch


src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -223,8 +223,53 @@ public override async Task<SearchResult> SearchAsync(SearchOptions searchOptions
...
+
+                    var tryWithQueryCache = SearchImpl(sqlSearchOptions, true, token);
+                    var tryWithoutQueryCache = SearchImpl(sqlSearchOptions, false, token);
+
+                    var result = await Task.WhenAny(tryWithQueryCache, tryWithoutQueryCache);
+                    await tokenSource.CancelAsync();
+
+                    _logger.LogInformation("First search completed in {ElapsedMilliseconds}ms, query cache enabled: {QueryCacheEnabled}.", stopwatch.ElapsedMilliseconds, result == tryWithQueryCache);
+                    return await result;
+                }
```
brendankowitz, line 258: Do we want to expose this on such a granular level?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -223,8 +223,53 @@ public override async Task<SearchResult> SearchAsync(SearchOptions searchOptions
...
+                var useQueryCache = Convert.ToString(useQueryCacheObj);
+                if (string.Equals(useQueryCache, "true", StringComparison.OrdinalIgnoreCase))
+                {
+                    return await SearchImpl(sqlSearchOptions, true, cancellationToken);
+                }
+                else if (string.Equals(useQueryCache, "false", StringComparison.OrdinalIgnoreCase))
+                {
+                    return await SearchImpl(sqlSearchOptions, false, cancellationToken);
+                }
+                else if (string.Equals(useQueryCache, "both", StringComparison.OrdinalIgnoreCase))
```
brendankowitz, line 242: Should this be an enum?

brendankowitz:
I think the description is now out of date for "x-ms-query-latency-over-efficiency"

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
LTA-Thinking: I didn't see a way to let the user control this setting with less visibility.
I know we try and keep the database operations opaque to the user, but since we've seen this setting can dramatically affect query performance I think it would be a good lever to be able to pull if needed.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/QueryCacheSetting.cs
```diff
@@ -0,0 +1,15 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Fhir.SqlServer.Features.Search
+{
+    internal class QueryCacheSetting
+    {
+        public const string Enabled = "enabled";
```
feordin, line 10: nit: Maybe an enumeration

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -223,8 +223,53 @@ public override async Task<SearchResult> SearchAsync(SearchOptions searchOptions
...
-        private async Task<SearchResult> SearchImpl(SqlSearchOptions sqlSearchOptions, CancellationToken cancellationToken)
+        private async Task<SearchResult> RunSearch(SqlSearchOptions sqlSearchOptions, CancellationToken cancellationToken)
         {
+            var fhirContext = _requestContextAccessor.RequestContext;
+            if (fhirContext != null
+                && fhirContext.Properties.TryGetValue(KnownQueryParameterNames.QueryCaching, out object useQueryCacheObj)
+                && useQueryCacheObj != null)
+            {
+                var useQueryCache = Convert.ToString(useQueryCacheObj);
+                if (string.Equals(useQueryCache, QueryCacheSetting.Enabled, StringComparison.OrdinalIgnoreCase))
```
feordin, line 234: nit: perhaps reading the value once and convert to enum instead of multiple string comparisions

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -357,7 +402,7 @@ await _sqlRetryService.ExecuteSql(
                                 new HashingSqlQueryParameterManager(new SqlQueryParameterManager(sqlCommand.Parameters)),
                                 _model,
                                 _schemaInformation,
-                                _reuseQueryPlans.IsEnabled(_sqlRetryService),
+                                reuseQueryPlans,
```
feordin, line 405: Why this change?

LTA-Thinking: This allows the setting to use or not use query plans to be controlled from an input parameter, which is then controlled by the header value parsing code.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/QueryCacheSetting.cs
feordin: It's true you would need something like: Enum.TryParse(x, out QueryCacheSettingsEnum cacheSetting);

But more of a suggestion.

feordin (approved):

PR #4807, opened 2025-01-29
Using connection mode from Cosmos Configuration

rajithaalurims:
## Description
Default connection mode in Cosmos Configuration is "Direct". This PR has changes to use connection mode based on the value in configuration file.

## Related issues
Addresses [issue # [140094](https://microsofthealth.visualstudio.com/Health/_workitems/edit/140094)].

## Testing
Confirmed that the connection mode is set based on the configuration file.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/FhirCosmosClientInitializer.cs
+++ b/src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/FhirCosmosClientInitializer.cs
@@ -131,8 +131,16 @@ namespace Microsoft.Health.Fhir.CosmosDb.Features.Storage
                 new CosmosClientBuilder(host, _cosmosAccessTokenProviderFactory.Invoke().TokenCredential) :
                 new CosmosClientBuilder(host, key);
 
+            if (configuration.ConnectionMode == ConnectionMode.Gateway)
+            {
+                builder.WithConnectionModeGateway();
+            }
+            else
+            {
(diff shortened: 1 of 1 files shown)


brendankowitz (approved):

rbans96:
Looks good - we should monitor latency with this change as it adds an extra hop since requests go through the gateway.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/FhirCosmosClientInitializer.cs
```diff
@@ -131,8 +131,16 @@ private CosmosClient CreateCosmosClientInternal(CosmosDataStoreConfiguration con
                 new CosmosClientBuilder(host, _cosmosAccessTokenProviderFactory.Invoke().TokenCredential) :
                 new CosmosClientBuilder(host, key);
 
+            if (configuration.ConnectionMode == ConnectionMode.Gateway)
```
rbans96, line 134: Should we add a logging statement to ensure that connectionMode is correctly populated from our config? If there are other means to confirm that, feel free to ignore

rajithaalurims: In case of exceptions, we do see client configuration which displays connection mode. 
Example: "ConnectionMode":"Direct"
To save time of PR build, I am going to leave it as it is.

Issue #4750, opened 2024-12-10
Remove LegacyExportJobWorker


brendankowitz:
**User story**
As a Developer, I want to remove [LegacyExportJobWorker](https://github.com/microsoft/fhir-server/blob/main/src/Microsoft.Health.Fhir.Core/Features/Operations/Export/LegacyExportJobWorker.cs) and related config, settings and tests so that our code is more maintainable.

**Acceptance criteria**
1. QueueClient Export jobs is on as the default export worker
2. All legacy code is removed

**Also see**
* [ILegacyExportOperationDataStore](https://github.com/microsoft/fhir-server/blob/main/src/Microsoft.Health.Fhir.Core/Features/Operations/Export/ILegacyExportOperationDataStore.cs)
* [appsettings.json#L185](https://github.com/microsoft/fhir-server/blob/cb53a355a88c50456c4542484a96512642b54b1a/src/Microsoft.Health.Fhir.Shared.Web/appsettings.json#L185)

PR #4838, opened 2025-02-24
Remove legacy export job worker

paorodma-ms:
## Description
Remove all classes, stored procedures and configurations related to Legacy export job worker.
https://microsofthealth.visualstudio.com/Health/_workitems/edit/145009

## Related issues
Addresses issue #4750.

## Testing
Unit tests were executed.

## FHIR Team Checklist


src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Operations/CosmosFhirOperationDataStore.cs
```diff
@@ -82,207 +73,28 @@ public CosmosFhirOperationDataStore(
...
 
         public override async Task<ExportJobOutcome> GetExportJobByIdAsync(string id, CancellationToken cancellationToken)
         {
-            if (IsLegacyJob(id))
-            {
-                // try old job records
-                var oldJobs = (ILegacyExportOperationDataStore)this;
-                return await oldJobs.GetLegacyExportJobByIdAsync(id, cancellationToken);
-            }
-
```
brendankowitz, line 120: @LTA-Thinking FYI, just want to confirm if you think this should be ok to delete? Its been a long while since old jobs were created

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Operations/CosmosFhirOperationDataStore.cs
```diff
@@ -82,207 +73,28 @@ public CosmosFhirOperationDataStore(
...
         {
-            if (_cosmosDataStoreConfiguration.UseQueueClientJobs)
-            {
-                return await base.CreateExportJobAsync(jobRecord, cancellationToken);
-            }
-
-            // try old job records
-            var oldJobs = (ILegacyExportOperationDataStore)this;
-            return await oldJobs.CreateLegacyExportJobAsync(jobRecord, cancellationToken);
+            return await base.CreateExportJobAsync(jobRecord, cancellationToken);
```
brendankowitz, line 85: If we are only calling the base, we shouldn't need the override at all

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Operations/CosmosFhirOperationDataStore.cs
LTA-Thinking: I think it would be better to keep the ability to see old job records, since this is the only part of the cleanup that would be a breaking change.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Operations/CosmosFhirOperationDataStore.cs
paorodma-ms: True, I have removed the three affected methods.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Operations/CosmosFhirOperationDataStore.cs
paorodma-ms: ok, I will bring back the needed deleted objects

LTA-Thinking (approved):

PR #4845, opened 2025-02-26
Enable schema compare test locally and make normalization generic

SergeyGaluzo:


test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlServerSchemaUpgradeTests.cs
```diff
@@ -44,16 +46,21 @@ namespace Microsoft.Health.Fhir.Tests.Integration.Persistence
...
         }
 
-        [Fact(Skip = "Issue connecting with SQL workload identity & custom auth provider. AB#122858")]
+        [Fact]
         public async Task GivenTwoSchemaInitializationMethods_WhenCreatingTwoDatabases_BothSchemasShouldBeEquivalent()
         {
+            // previously test was skipped with (Skip = "Issue connecting with SQL workload identity & custom auth provider. AB#122858")
+            if (!KnownEnvironmentVariableNames.SqlServerConnectionString.Contains("(local)", StringComparison.OrdinalIgnoreCase))
+            {
+                await Task.CompletedTask;
```
LTA-Thinking, line 61: What is this line doing? This should just be a no op.

SergeyGaluzo: This test cannot run in pipelined because of workload identity issue, so it is enabled only when run locally, for which we set server=(local).

LTA-Thinking (approved):

PR #4850, opened 2025-03-03
Increase max valueset size to 20,000

feordin:
## Description
This increases the maximum expansion for a valueset from 500 to 20,000 codes.  This is needed for a customer in production which uses a large in memory valuest.

## Related issues
Addresses [issue [AB#145937](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/145937)].

## Testing
Manual testing

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Validation/ProfileValidator.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Validation/ProfileValidator.cs
@@ -9,6 +9,7 @@ using EnsureThat;
 using Hl7.Fhir.ElementModel;
 using Hl7.Fhir.Model;
 using Hl7.Fhir.Specification.Source;
+using Hl7.Fhir.Specification.Terminology;
 using Hl7.Fhir.Validation;
 using Microsoft.Extensions.Options;
 using Microsoft.Health.Fhir.Core.Configs;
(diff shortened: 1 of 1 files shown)


src/Microsoft.Health.Fhir.Shared.Core/Features/Validation/ProfileValidator.cs
```diff
@@ -38,12 +39,20 @@ public ProfileValidator(IProvideProfilesForValidation profilesResolver, IOptions
 
         private Validator GetValidator()
         {
+            var expanderSettings = new ValueSetExpanderSettings
+            {
+                MaxExpansionSize = 20000, // Set your desired max expansion size here
```
LTA-Thinking, line 44: Where did we get the 20,000 from? Was that the default value for Firely until we upgraded?

mikaelweave, line 44: Can we set this from a configuration somewhere?

feordin: The current valueset expansion that is being used is ~17,000.  The 20,000 give a bit of a buffer, but I don't want much larger than necessary.

feordin: We could add a setting, and make it configurable.  But I wanted a quick change to address this issue.  Then we can consider a more in depth change.

LTA-Thinking (approved):

PR #4863, opened 2025-03-12
Merge throttling (waits in code)

SergeyGaluzo:
Addresses
https://microsofthealth.visualstudio.com/Health/_workitems/edit/146794/

Uses application name on the connection string to track MergeResources calls. 
Latency increase on each merge call is ~2 milliseconds.
If number of concurrent calls is greater than threshold (default is 256) throws exception in MergeResourcesBeginTransaction stored procedure. On this exception code retries with exponential backoff.
In future PR retries should be replaced by code throwing 429 to the client,
Even with current waits both latency and throughput are significantly improved. 

image


src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlStoreClient.cs
```diff
@@ -243,9 +245,20 @@ internal async Task<long> MergeResourcesGetTransactionVisibilityAsync(Cancellati
...
                 catch (Exception e)
                 {
+                    var sqlEx = e as SqlException;
+                    if (sqlEx != null && sqlEx.Number == SqlStoreErrorCodes.MergeResourcesConcurrentCallsIsAboveOptimal)
+                    {
+                        _logger.LogWarning(e, $"Error on {nameof(MergeResourcesBeginTransactionAsync)}: MergeResources concurrent calls is above optimal. Delay={delayOnOverloadMilliseconds} milliseconds.");
+
+                        // TODO: Prepare to throw 429 instead of wait/delay when bundle code is ready
+                        await Task.Delay(delayOnOverloadMilliseconds, cancellationToken);
+                        delayOnOverloadMilliseconds = (int)(delayOnOverloadMilliseconds * (2 + (RandomNumberGenerator.GetInt32(1000) / 1000.0)));
```
LTA-Thinking, line 255: Should we set a maximum wait time and maximum number of retries before throwing an exception?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -717,7 +718,18 @@ internal async Task MergeResourcesWrapperAsync(long transactionId, bool singleTr
             new TokenStringCompositeSearchParamListTableValuedParameterDefinition("@TokenStringCompositeSearchParams").AddParameter(cmd.Parameters, new TokenStringCompositeSearchParamListRowGenerator(_model, new TokenSearchParamListRowGenerator(_model, _searchParameterTypeMap), new StringSearchParamListRowGenerator(_model, _searchParameterTypeMap), _searchParameterTypeMap).GenerateRows(mergeWrappers));
             new TokenNumberNumberCompositeSearchParamListTableValuedParameterDefinition("@TokenNumberNumberCompositeSearchParams").AddParameter(cmd.Parameters, new TokenNumberNumberCompositeSearchParamListRowGenerator(_model, new TokenSearchParamListRowGenerator(_model, _searchParameterTypeMap), new NumberSearchParamListRowGenerator(_model, _searchParameterTypeMap), _searchParameterTypeMap).GenerateRows(mergeWrappers));
             cmd.CommandTimeout = 300 + (int)(3600.0 / 10000 * (timeoutRetries + 1) * mergeWrappers.Count);
-            await cmd.ExecuteNonQueryAsync(cancellationToken);
+
+            if (enlistInTransaction)
```
LTA-Thinking, line 722: What is this doing differently than just passing false to ObtainSqlConnectionWrapperAsync?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlStoreClient.cs
SergeyGaluzo: If we set it relatively low value appropriate to API calls, then background processes start failing. If we set to value appropriate for background jobs it will never happen.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
SergeyGaluzo: ObtainSqlConnectionWrapperAsync does not allow setting app name, because it was too difficult to make appropriate change in shared.

In general, enlisting in C# transaction is evil, so we should stay away from using it in all cases we can, and eventually get rid of it.

LTA-Thinking (approved):

LTA-Thinking (approved):

rajithaalurims (approved):

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/87.diff.sql
```diff
@@ -0,0 +1,49 @@
...
+    EXECUTE sys.sp_sequence_get_range @sequence_name = 'dbo.ResourceSurrogateIdUniquifierSequence', @range_size = @Count, @range_first_value = @FirstValueVar OUT, @range_last_value = @LastValueVar OUT
+    SET @SequenceRangeFirstValue = convert(int,@FirstValueVar)
+    IF @SequenceRangeFirstValue > convert(int,@LastValueVar)
+      SET @FirstValueVar = NULL
+  END
+
+  SET @TransactionId = datediff_big(millisecond,'0001-01-01',sysUTCdatetime()) * 80000 + @SequenceRangeFirstValue
+
+  INSERT INTO dbo.Transactions
+         (  SurrogateIdRangeFirstValue,   SurrogateIdRangeLastValue,                      HeartbeatDate )
```
fhibf, line 40: I see that line 40 and 41 have these continuous spaces ("                      "). Is that a formatting issue?

fhibf (approved):

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlStoreClient.cs
```diff
@@ -243,9 +245,20 @@ internal async Task<long> MergeResourcesGetTransactionVisibilityAsync(Cancellati
                 }
                 catch (Exception e)
                 {
+                    var sqlEx = e as SqlException;
+                    if (sqlEx != null && sqlEx.Number == SqlStoreErrorCodes.MergeResourcesConcurrentCallsIsAboveOptimal)
+                    {
+                        _logger.LogWarning(e, $"Error on {nameof(MergeResourcesBeginTransactionAsync)}: MergeResources concurrent calls is above optimal. Delay={delayOnOverloadMilliseconds} milliseconds.");
+
+                        // TODO: Prepare to throw 429 instead of wait/delay when bundle code is ready
+                        await Task.Delay(delayOnOverloadMilliseconds, cancellationToken);
```
fhibf, line 254: Should we add a max number of retries, similar to the timeouts? 
Or should that be included as part of the bundle handling part of the code?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/87.diff.sql
SergeyGaluzo: This is because INSERT column list is aligned with SELECT list.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlStoreClient.cs
SergeyGaluzo: I don't think so. Please look at my answer to RB's comment.

PR #4867, opened 2025-03-14
Upstream changes to SQL layer to call the blob library for write

gunjitchhhatwal:
## Description
- Introduced a feature flag to determine if blob storage should be used
- Removed .net 6.0. This has been done in main already. Any associated changes can be brought in when we rebase from main post DB schema optimizations has gone in.
- Added method for registering blob store and calling that during start up if blob feature is enabled.

## Related issues
https://microsofthealth.visualstudio.com/Health/_workitems/edit/144577

## Testing
Manual. Will add integration tests once read path is available. It is under dev as of now. 

## FHIR Team Checklist


src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlAdlsClient.cs
```diff
@@ -44,7 +45,7 @@
                         }
 
                         var db = sqlRetryService.Database.Length < 50 ? sqlRetryService.Database : sqlRetryService.Database.Substring(0, 50);
-                        _adlsContainerName = $"fhir-adls-{db.Replace("_", "-", StringComparison.InvariantCultureIgnoreCase).ToLowerInvariant()}";
+                        _adlsContainerName = "fhirresources";
```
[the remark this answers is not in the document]
gunjitchhhatwal: This is a temporary change I had to do to read the resource for the integration test. When we have the read path available, this piece of code will be removed.

src/Microsoft.Health.Fhir.Core/Configs/CoreFeatureConfiguration.cs
```diff
@@ -86,5 +88,10 @@ public class CoreFeatureConfiguration
         /// Whether the service supports SQL read only replicas.
         /// </summary>
         public bool SupportsSqlReplicas { get; set; } = false;
+
+        /// <summary>
+        /// Gets or sets a value indicating whether storing FHIR resources in blob storage is enabled or not.
+        /// </summary>
+        public bool SupportsRawResourceInBlob { get; set; } = false;
```
feordin, line 95: Should this be part of the SqlServer configuration instead of the core configurtion?  As blob store only applies if you have Sql?

src/Microsoft.Health.Fhir.Shared.Web/Startup.cs
```diff
@@ -130,6 +131,14 @@ private void AddDataStore(IServiceCollection services, IFhirServerBuilder fhirSe
                     Configuration?.GetSection(SqlServerDataStoreConfiguration.SectionName).Bind(config);
                 });
                 services.Configure<SqlRetryServiceOptions>(Configuration.GetSection(SqlRetryServiceOptions.SqlServer));
+
+                // Only register blob store if the feature is enabled
+                CoreFeatureConfiguration coreFeatureConfiguration = new();
```
feordin, line 136: Similar to previous comment, if we move this setting to be part of the SQL configuration, we would not need to bring in CoreFeatureConfiguration

src/Microsoft.Health.Fhir.Core/Configs/CoreFeatureConfiguration.cs
gunjitchhhatwal: I was thinking about that and chose to keep it here if in future there was a move to have an orchestration layer that called into blob and SQL layers in which case it would have access to this configuration. But I do see your point that is not applicable to Cosmos and could live in SQL layer. Do you think it should matter considering this offering is only being considered for certain managed customers who can only use SQL Server?

src/Microsoft.Health.Fhir.Shared.Web/Startup.cs
gunjitchhhatwal: This I can move inside the SQL registration method so that app startup is unaware of this detail. Will make this change.

Directory.Build.props
```diff
@@ -19,7 +19,7 @@
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
     <RepositoryUrl>https://github.com/microsoft/fhir-server</RepositoryUrl>
     <RunSettingsFilePath>$(MSBuildThisFileDirectory)\CodeCoverage.runsettings</RunSettingsFilePath>
-    <TargetFrameworks>net9.0;net8.0;net6.0</TargetFrameworks>
```
gunjitchhhatwal, line 22: I have removed support for ease of development as it has been removed from main as well. Any other changes related to this will be merged in when we rebase from main once DB schema optimizations has been merged to main.

build/jobs/e2e-tests.yml
```diff
@@ -88,7 +88,7 @@ steps:
     displayName: 'E2E ${{ parameters.version }} ${{parameters.appServiceType}}'
     inputs:
       command: test
-      arguments: '"$(Agent.TempDirectory)/E2ETests/**/*${{ parameters.version }}.Tests.E2E*.dll" --blame-hang-timeout 7m --filter "FullyQualifiedName~${{parameters.appServiceType}}&Category!=ExportLongRunning"'
+      arguments: '"$(Agent.TempDirectory)/E2ETests/**/*${{ parameters.version }}.Tests.E2E*.dll" --blame-hang-timeout 7m --filter "FullyQualifiedName~${{parameters.appServiceType}}&Category!=ExportLongRunning&DisplayName!~BlobEnabled"'
```
gunjitchhhatwal, line 91: This is a temporary change until we have the read path enabled. I am tracking a separate integration and e2e test task. Will add these back when i take that up.

src/Microsoft.Health.Fhir.R4.Web/Properties/launchSettings.json
```diff
@@ -19,7 +19,25 @@
             },
             "applicationUrl": "https://localhost:44348/"
         },
-        "SqlServer": {
+        "SqlServerBlobDisabled": {
```
gunjitchhhatwal, line 22: As of now both settings related to blob enabled and blob disabled are the same since the we are skipping blob enabled test in the pipeline. Will make additional changes to it when I take the task related to e2e test including both scenarios.

src/Microsoft.Health.Fhir.Core/Configs/CoreFeatureConfiguration.cs
feordin: Ok, let's keep it there for future proof :).

PR #4873, opened 2025-03-18
Converting IdentityServer4 to OpenIddict.

tarunmathew12:
## Description
The PR replaces use of IdentityServer4  with OpenIddict libraries.

## Related issues
Addresses [issue #123727].
[User Story 123727](https://microsofthealth.visualstudio.com/Health/_workitems/edit/123727): Fix vulnerability issues in IdentityServer4 package referenced by Fhir OSS projects.

## Testing
Tested by running the existing E2E/integration test and also manual testing against local Fhir server.

## FHIR Team Checklist


Directory.Packages.props
```diff
@@ -119,6 +118,12 @@
     <PackageVersion Include="Newtonsoft.Json.Schema" Version="4.0.1" />
     <PackageVersion Include="Newtonsoft.Json" Version="13.0.3" />
     <PackageVersion Include="NSubstitute" Version="5.3.0" />
+    <PackageVersion Include="Microsoft.EntityFrameworkCore.InMemory" Version="9.0.0" />
+    <PackageVersion Include="OpenIddict.Abstractions" Version="4.0.0" />
```
kevinchalet, line 122: Hey 👋🏻 

OpenIddict maintainer here.

FYI, OpenIddict 4.0 shipped in 2022 and is no longer supported. Instead, you may want to reference OpenIddict 6.1.1, which is the latest version available at the time of writing. These 2 migration guides can help you move smoothly to the 6.x branch:
  - https://documentation.openiddict.com/guides/migration/40-to-50
  - https://documentation.openiddict.com/guides/migration/50-to-60

All the best.

tarunmathew12: Thank you, Kevin. We will check it out.

tarunmathew12: Hi @kevinchalet, we are letting you know of an issue we found in OpenIddict since we can't post the issue in your github. We are seeing an unhandled exception from OpenIddict as we run this [test](https://github.com/microsoft/fhir-server/blob/698a1fb783f67f5b0e90026b362b4ea04316fd34/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/BasicSearchTests.cs#L190)  that sends a really long request url against our server. The exception is likely thrown [here](https://github.com/openiddict/openiddict-core/blob/d9e44860f60ed929c2ff5a1892bb2460d32d608d/src/OpenIddict.Server.AspNetCore/OpenIddictServerAspNetCoreHandlers.cs#L108), before the request reaches our service. Could you please take a look? We will disable this test until the issue is resolved. Thank you.
 
Microsoft.AspNetCore.Diagnostics.ExceptionHandlerMiddleware: Error: An unhandled exception has occurred while executing the request.

System.UriFormatException: Invalid URI: The Uri string is too long.
   at System.Uri.CreateThis(String uri, Boolean dontEscape, UriKind uriKind, UriCreationOptions& creationOptions)
   at System.Uri..ctor(String uriString, UriKind uriKind)
   at OpenIddict.Server.AspNetCore.OpenIddictServerAspNetCoreHandlers.ResolveRequestUri.HandleAsync(ProcessRequestContext context)
   at OpenIddict.Server.OpenIddictServerDispatcher.DispatchAsync[TContext](TContext context)
   at OpenIddict.Server.OpenIddictServerDispatcher.DispatchAsync[TContext](TContext context)
------------- [more]

kevinchalet: Hey,

> Hi @kevinchalet, we are letting you know of an issue we found in OpenIddict since we can't post the issue in your github.

Microsoft sponsored the project last year, so you can actually open a ticket there 😃 

> We are seeing an unhandled exception from OpenIddict as we run this test that sends a really long request url against our server. The exception is likely thrown here, before the request reaches our service. 

As shown by the stack trace, that exception isn't thrown by OpenIddict itself but by the BCL `Uri` class, that enforces a ~65K limit: https://github.com/dotnet/runtime/blob/551379a185a011789e29a9d326804eb15995f2c7/src/libraries/System.Private.Uri/src/System/Uri.cs#L43. It's not something we can control as it's an internal and non-configurable limit.

That said, 100K characters (even "just 65K" 😄) is insanely high and you'll always hit another limit when running against a real server (e.g IIRC, Kestrel enforces a "8K-ASCII-encoded-bytes" limit for the whole HTTP request line). Are you sure you need to deal with such large URIs? The name of the test suggests you're using HTTP POST: aren't you flowing the payload in the request form/body instead of the query string?

tarunmathew12: The point of this test is to ensure our service doesn't crash with a really long url. With IdentityServer4, we were able to test this scenario. But, we will contact the test owner and convert the test to UT as we think the case is targeting some of our internal components parsing the request url. Thank you for the input.

kevinchalet: > With IdentityServer4, we were able to test this scenario.

It most likely didn't use the BCL `System.Uri` type 😃 

I'll tweak that handler to use `Uri.TryCreate()` instead of the `Uri` constructor: you still won't be able to use OpenID Connect endpoints with URIs larger than ~65K (which really makes no sense), but at least, it won't interfere with your own API endpoints.

tarunmathew12: Thank you, @kevinchalet , we will pick up your change later.

kevinchalet: @tarunmathew12 OpenIddict 6.2.0 shipped today and includes the fix you wanted: https://github.com/openiddict/openiddict-core/releases/tag/6.2.0.

tarunmathew12: Thank you, we will update it soon.

src/Microsoft.Health.Fhir.R4.Api/Microsoft.Health.Fhir.R4.Api.csproj
```diff
@@ -17,6 +17,8 @@
     <PackageReference Include="MediatR" />
     <PackageReference Include="Microsoft.AspNetCore.Mvc.NewtonsoftJson" />
     <PackageReference Include="Microsoft.Extensions.Http.Polly" />
+    <PackageReference Include="OpenIddict.Abstractions" />
+    <PackageReference Include="OpenIddict.AspNetCore" />
```
brendankowitz, line 21: Is this needed in API layer?

kevinchalet:
> We will try again, but if you have any sample code please point us to it.

Here's an example: https://github.com/openiddict/openiddict-samples/blob/479b3ba5cd5ac3f3549dacc5e5cd352f117527d9/samples/Kalarba/Kalarba.Server/Startup.cs#L59-L96 (it's for the resource owner password credentials grant but the same exact approach also works for the client credentials grant you're using here).

If you get a 404 response, make sure you're not calling `EnableTokenEndpointPassthrough()` (otherwise, control flow is returned to ASP.NET Core, which ultimately returns a 404 response if no MVC or minimal action matches the request path).

tarunmathew12:
> > We will try again, but if you have any sample code please point us to it.
> 
> Here's an example: https://github.com/openiddict/openiddict-samples/blob/479b3ba5cd5ac3f3549dacc5e5cd352f117527d9/samples/Kalarba/Kalarba.Server/Startup.cs#L59-L96 (it's for the resource owner password credentials grant but the same exact approach also works for the client credentials grant you're using here).
> 
> If you get a 404 response, make sure you're not calling `EnableTokenEndpointPassthrough()` (otherwise, control flow is returned to ASP.NET Core, which ultimately returns a 404 response if no MVC or minimal action matches the request path).
> 
> Note: you don't have to define the event handler inline: you can create a dedicated class that implements `IOpenIddictServerHandler<HandleTokenRequestContext>` and use the `builder.UseScopedHandler<...>()` or `builder.UseSingletonHandler<...>()` APIs to register it as a service.

Thank you @kevinchalet, we will take a look.

tarunmathew12:
> > > We will try again, but if you have any sample code please point us to it.
> > 
> > 
> > Here's an example: https://github.com/openiddict/openiddict-samples/blob/479b3ba5cd5ac3f3549dacc5e5cd352f117527d9/samples/Kalarba/Kalarba.Server/Startup.cs#L59-L96 (it's for the resource owner password credentials grant but the same exact approach also works for the client credentials grant you're using here).
> > If you get a 404 response, make sure you're not calling `EnableTokenEndpointPassthrough()` (otherwise, control flow is returned to ASP.NET Core, which ultimately returns a 404 response if no MVC or minimal action matches the request path).
> > Note: you don't have to define the event handler inline: you can create a dedicated class that implements `IOpenIddictServerHandler<HandleTokenRequestContext>` and use the `builder.UseScopedHandler<...>()` or `builder.UseSingletonHandler<...>()` APIs to register it as a service.
> 
> Thank you @kevinchalet, we will take a look.

Hi @kevinchalet, we tried your suggestion and confirmed that it worked when running our service locally and sending some requests. But, we noticed our E2E tests failing because they use our internal auth controller that gets a token from a token provider. We noticed this controller no longer getting called after removing EnableTokenEndpointPassthrough (as you suggested). It seems we still need our OpenIddict auth controller (OpenIddictAuthorizationController) to make both scenarios work. Please let us know if you have any other suggestions for the issue. Thank you.

kevinchalet:
> We noticed this controller no longer getting called after removing EnableTokenEndpointPassthrough (as you suggested). It seems we still need our OpenIddict auth controller (OpenIddictAuthorizationController) to make both scenarios work.

When the pass-through mode is not enabled, flow control is not returned to ASP.NET Core for the token endpoint and the controller action is never called (so it's basically useless and should be removed).

Is the code for that test visible somewhere?

tarunmathew12:
> > We noticed this controller no longer getting called after removing EnableTokenEndpointPassthrough (as you suggested). It seems we still need our OpenIddict auth controller (OpenIddictAuthorizationController) to make both scenarios work.
> 
> When the pass-through mode is not enabled, flow control is not returned to ASP.NET Core for the token endpoint and the controller action is never called (so it's basically useless and should be removed).
> 
> Is the code for that test visible somewhere?

This is the controller that calls the OpenIddict controller to get a token. We have a configuration to enable/disable use of this controller and it is enabled by default.
 
[fhir-server/src/Microsoft.Health.Fhir.Api/Controllers/AadSmartOnFhirProxyController.cs at main · microsoft/fhir-server](https://github.com/microsoft/fhir-server/blob/main/src/Microsoft.Health.Fhir.Api/Controllers/AadSmartOnFhirProxyController.cs)

kevinchalet:
Is there a specific test I can debug?

I just cloned the solution locally and many of the tests are failing (mainly due to culture issues, probably because of some missing `CultureInfo.InvariantCulture` parameters 😄), so it's like finding a needle in a haystack.

tarunmathew12:
> Is there a specific test I can debug?
> 
> I just cloned the solution locally and many of the tests are failing (mainly due to culture issues, probably because of some missing `CultureInfo.InvariantCulture` parameters 😄), so it's like finding a needle in a haystack.

You can run any test under any version of Fhir to reproduce the issue. (e.g. I am running this [one](https://github.com/microsoft/fhir-server/blob/89d88a2b9670a2429bf6c76c352affe6756ac177/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/BasicSearchTests.cs#L1124) for no particular reason.) You will need at least one type of data stores (SQL/Cosmos) mentioned in this [wiki](https://github.com/microsoft/fhir-server/wiki/Build-and-debug-locally) to run the test. Please note that we have the asp.net core control flow [enabled](https://github.com/microsoft/fhir-server/blob/9338337e93801faa20d8c39f1d7852e756b7d23f/src/Microsoft.Health.Fhir.Api.OpenIddict/Extensions/DevelopmentIdentityProviderRegistrationExtensions.cs#L116) in the repo you cloned. I have a change converting the openiddict controller to a class that implements IOpenIddictServerHandler locally. If you need the change, I will push it to another branch if you need it. Thank you.

image

kevinchalet:
Thanks for the instructions.

There's something I don't understand: how does that "AadSmartOnFhirProxy" thing work?

I see you configured OpenIddict to handle both `/AadSmartOnFhirProxy/authorize` and `/AadSmartOnFhirProxy/token` requests - routes that are shared with `AadSmartOnFhirProxyController` AFAICT - but there's no code at all in `AadSmartOnFhirProxyController` that will ever ask OpenIddict to return a proper OIDC response (i.e an authorization response-  e.g with an authorization code when using the code flow - and a token response containing an access token that was generated by OpenIddict).

Essentially, all you're currently doing with that setup is asking OpenIddict to validate authorization and token requests while still ignoring all that and returning your own arbitrary response:
  - For the authorization endpoint, a browser-interactive HTTP 302 response that sends the user agent to AAD.
  - For the token endpoint, a token response proxied from AAD via a backchannel HTTP call.

IMHO, it doesn't make a lot of sense. If you expect OpenIddict to handle the `/AadSmartOnFhirProxy/authorize` and `/AadSmartOnFhirProxy/token` routes, then it should take care of both the request validation **and** the response generation.

Are you trying to make that "AadSmartOnFhirProxy" thing conditional? Perhaps for using AAD in production while having an OpenIddict-based fallback for development or something like that? If so, it's probably the calls to `options.SetTokenEndpointUris(...)` and `options.SetAuthorizationEndpointUris(...)` that should be conditional.

tarunmathew12:
> Thanks for the instructions.
> 
> There's something I don't understand: how does that "AadSmartOnFhirProxy" thing work?
> 
> I see you configured OpenIddict to handle both `/AadSmartOnFhirProxy/authorize` and `/AadSmartOnFhirProxy/token` requests - routes that are shared with `AadSmartOnFhirProxyController` AFAICT - but there's no code at all in `AadSmartOnFhirProxyController` that will ever ask OpenIddict to return a proper OIDC response (i.e an authorization response- e.g with an authorization code when using the code flow - and a token response containing an access token that was generated by OpenIddict).
> 
> Essentially, all you're currently doing with that setup is asking OpenIddict to validate authorization and token requests while still ignoring all that and returning your own arbitrary response:
> 
> * For the authorization endpoint, a browser-interactive HTTP 302 response that sends the user agent to AAD.
> * For the token endpoint, a token response proxied from AAD via a backchannel HTTP call.
> 
> IMHO, it doesn't make a lot of sense. If you expect OpenIddict to handle the `/AadSmartOnFhirProxy/authorize` and `/AadSmartOnFhirProxy/token` routes, then it should take care of both the request validation **and** the response generation.
> 
> Are you trying to make that "AadSmartOnFhirProxy" thing conditional? Perhaps for using AAD in production while having an OpenIddict-based fallback for development or something like that? If so, it's probably the calls to `options.SetTokenEndpointUris(...)` and `options.SetAuthorizationEndpointUris(...)` that should be conditional.

Hi @kevinchalet , I am not so familiar with [Smart on Fhir](https://learn.microsoft.com/en-us/azure/healthcare-apis/fhir/smart-on-fhir), but I think it is more involved than what AadSmartOnFhirProxyController does. Our goal is to make local debugging and automation run work as before - like with IdentityServer4. The only way we found so far is to have OpenIddictAuthorizationController we have in the repo, so AadSmartOnFhirProxyController can get a token from the controller. This flow will be broken when we disable the token pass-through in OpenIddict. We are asking the team if they have any concern with having OpenIddictAuthorizationController. Will keep you posted.

kevinchalet:
> Our goal is to make local debugging and automation run work as before - like with IdentityServer4.

I'm not an IdSrv expert (for obvious reasons :trollface:) but AFAICT, you're not using the same configuration at all:

  - In the main branch, you're using IdSrv with a single in-memory client only allowed to use the ROPC and client credentials grants + the default endpoint configuration (i.e the hardcoded `/connect/token` route that IdSrv always handles for you since it doesn't the pass-through mode OpenIddict gives you).

  - In this PR, you've configured OpenIddict to handle token requests posted to both `/connect/token` and `/AadSmartOnFhirProxy/token`, which differs from the main branch where IdSrv only takes care of requests posted to `/connect/token`. You've also configured OpenIddict to handle authorization requests posted to `/AadSmartOnFhirProxy/authorize`, which isn't something the main branch does at all (no IdSrv client is allowed to use an interactive flow like the code flow in that branch, BTW).

> The only way we found so far is to have OpenIddictAuthorizationController we have in the repo, so AadSmartOnFhirProxyController can get a token from the controller. This flow will be broken when we disable the token pass-through in OpenIddict. We are asking the team if they have any concern with having OpenIddictAuthorizationController. Will keep you posted.

I strongly suspect bringing back that controller is not the right answer, you have a more fundamental issue that requires shedding some light on the design of that feature before making any appropriate decision.

If you want something that behaves the same way as IdSrv in the main branch, you should disable the authorization endpoint (i.e remove `options.SetAuthorizationEndpointUris("/AadSmartOnFhirProxy/authorize")`) and remove the route collision for the token endpoint by calling `options.SetTokenEndpointUris("/connect/token")` instead of `options.SetTokenEndpointUris("/connect/token", "/AadSmartOnFhirProxy/token")`.

brendankowitz (approved):
LGTM, works well enough for dev purposes

feordin (approved):

feordin (approved):

Directory.Packages.props
kevinchalet: For the record, the limits discussed here were removed in .NET 10. More information here: https://github.com/dotnet/docs/issues/47602.

PR #4890, opened 2025-03-27
Reject search request with resourceType with wrong casing.

tarunmathew12:
## Description
The PR addresses these two issues:
- The search request with the resourceType with wrong casing returns 500 response.

## Related issues
Addresses [issue #148216].

[Bug 148216](https://microsofthealth.visualstudio.com/Health/_workitems/edit/148216): API calls with lower case Resource Types returning 500

## Testing
Tested manually with local FHIR service, and by adding some E2E tests. 

## FHIR Team Checklist


test/Microsoft.Health.Fhir.Shared.Tests.Integration/Features/Smart/SmartSearchTests.cs
```diff
@@ -692,6 +693,29 @@
...
+            Skip.If(
+                ModelInfoProvider.Instance.Version != FhirSpecification.R4 &&
+                ModelInfoProvider.Instance.Version != FhirSpecification.R4B,
+                "This test is only valid for R4 and R4B");
+
+            try
+            {
+                var response = await _fixture.GetResourceHandler.Handle(
+                    new GetResourceRequest(new ResourceKey(resourceType, "smart-patient-B"), bundleResourceContext: null),
+                    CancellationToken.None);
```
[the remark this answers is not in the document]
tarunmathew12: Fixed.

rajithaalurims (dismissed):

rajithaalurims (approved):

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Get/GetResourceHandler.cs
```diff
@@ -74,6 +74,11 @@ public async Task<GetResourceResponse> Handle(GetResourceRequest request, Cancel
             }
             else
             {
+                if (KnownDataStores.IsSqlServerDataStore(FhirDataStore.DataStoreType) && !string.IsNullOrWhiteSpace(key?.ResourceType) && !ModelInfoProvider.IsKnownResource(key.ResourceType, false))
```
brendankowitz, line 77: Why does GetResourceHandler need to know if its a sqldatastore?

tarunmathew12: CosmosFhirDataStore throws "ResourceNotFound" exception in the case of wrong casing resource type. I think the right exception would be "ResourceNotSupported" exception. I'm checking the data store type to fix the issue for SQL preserving the current behavior for Cosmos. Please let me know if we should make it consistent and which exception to throw.

mikaelweave: Can we have this check for both Gen1 and Gen2? Is there any downside to running this for both datastores?

Also, if we change the behavior for Gen2, we may need to merge to give customers enough advance notice to check their code.

src/Microsoft.Health.Fhir.Core/Features/Persistence/IFhirDataStore.cs
```diff
@@ -11,6 +11,8 @@ namespace Microsoft.Health.Fhir.Core.Features.Persistence
 {
     public interface IFhirDataStore
     {
+        string DataStoreType { get; }
```
mikaelweave, line 14: I'm not sure we need this but open to discussion. Can we just inject `IFhirRuntimeConfiguration` via DI?

See here for injection
https://github.com/microsoft/fhir-server/blob/132cd07f97d47ee82dc5c5bd7799b41a990c1edf/src/Microsoft.Health.Fhir.Shared.Api/Controllers/ExportController.cs#L73

There is a datastore on that object we could use.

mikaelweave (commented):
A couple questions to understand before next step of review.

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Get/GetResourceHandler.cs
tarunmathew12: Gen1 currently throws ResourceNotFoundException. If no problem changing to throwing ResourceNotSupportedException, I will update the PR accordingly (that will automatically address your other comment on adding DataStoreType in IFhirDataStore.)
 
Yes, I agree with you about notifying customers for this change (Gen2 currently returns 500, but still good to notify them).

mikaelweave: Do `ResourceNotSupported` and `ResourceNotFound` result in the same HTTP result code or different result codes? If they are the same HTTP response code, we can change Cosmos to match.

Since SQL is returning 500 that's a bug so no need to notify first.

tarunmathew12: Both will result in 400 (with different error messages). I think ResourceNotSupported in this case is more accurate than ResourceNotFound since the reason for 400 is an invalid resource type. I will update the PR based on your feedback. Thank you, Mikael!

tarunmathew12: Fixed as suggested.

src/Microsoft.Health.Fhir.Core/Features/Persistence/IFhirDataStore.cs
tarunmathew12: Removed the property.

PR #4898, opened 2025-04-01
Delete only applicable types from search param

feordin:
## Description
When running a delete search parameter operation (which also runs on an update), we were removing all related search parameters that matched on the same "code" value, even those for other resource types.

This change fixes that issue.

## Related issues
Addresses [issue [AB#117004](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/117004)].

## Testing
Manual testing

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterDefinitionManager.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterDefinitionManager.cs
@@ -267,11 +267,15 @@ namespace Microsoft.Health.Fhir.Core.Features.Definition
                 throw new ResourceNotFoundException(string.Format(Core.Resources.CustomSearchParameterNotfound, url));
             }
 
-            // for search parameters with a base resource type we need to delete the search parameter
-            // from all derived types as well, so we iterate across all resources
-            foreach (var resourceType in TypeLookup.Keys)
+            // find all derived resources from the list of base resources
+            var allResourceTypes = GetDerivedResourceTypes(searchParameterInfo.BaseResourceTypes);
+            foreach (var resourceType in allResourceTypes)
             {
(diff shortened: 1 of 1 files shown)


src/Microsoft.Health.Fhir.Core/Features/Definition/SearchParameterDefinitionManager.cs
```diff
@@ -267,11 +267,15 @@ public void DeleteSearchParameter(string url, bool calculateHash = true)
...
-            foreach (var resourceType in TypeLookup.Keys)
+            // find all derived resources from the list of base resources
+            var allResourceTypes = GetDerivedResourceTypes(searchParameterInfo.BaseResourceTypes);
+            foreach (var resourceType in allResourceTypes)
             {
                 TypeLookup[resourceType].TryRemove(searchParameterInfo.Code, out var removedParam);
+                if (removedParam.Url != searchParameterInfo.Url)
+                {
+                    _logger.LogError("Error, Search Param {RemovedParam} removed from Search Param Definitoni manager.  It does not match deleted Search Param {Url}", removedParam.Url, url);
+                }
```
rajithaalurims, line 278: @feordin  do we need to change CalculateSearchParameterHash method as well to update hash for only resource types that are affected?

feordin: @rajithaalurims CalculateSearchParameterHash can be called when any change occurs to the Search Params and it will reproducible calculate the hash value for each resource type.  If no changes have occurred for a specific resource type, it won't be affected.  So we shouldn't need any changes here.

rajithaalurims (approved):

PR #4937, opened 2025-04-21
fix invisible trans watchdog

SergeyGaluzo:
Limit number of transactions to process in a single batch to 10000 to avoid timeouts.

fixes https://microsofthealth.visualstudio.com/Health/_workitems/edit/151722 and 
https://microsofthealth.visualstudio.com/Health/_workitems/edit/148858

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Watchdogs/InvisibleHistoryCleanupWatchdog.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Watchdogs/InvisibleHistoryCleanupWatchdog.cs
@@ -42,40 +42,46 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Watchdogs
 
         public override double PeriodSec { get; internal set; } = 3600;
 
+        public string RetentionPeriodDaysId => $"{Name}.RetentionPeriodDays";
+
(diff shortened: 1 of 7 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Watchdogs/InvisibleHistoryCleanupWatchdog.cs
```diff
@@ -42,40 +42,46 @@ internal InvisibleHistoryCleanupWatchdog()
...
+                    }
+
+                    var cleanedUpToTranId = transToClean.Max(_ => _.TransactionId);
+                    await UpdateLastCleanedUpTransactionId(cleanedUpToTranId);
+                    _logger.LogInformation($"{Name}: cleaned up to {cleanedUpToTranId}. transactions={totalTrans} removed rows={totalRows}");
+                }
+
+                transToCleanCount = transToClean.Count;
             }
+            while (transToCleanCount > 0);
```
feordin, line 82: Do we need any kind of limit on this loop?  We always want it to run through all transactions, no matter the number?

SergeyGaluzo: Yes we do. I don't see the need to limit.

PR #4962, opened 2025-04-30
Fix small stats bug

SergeyGaluzo:


mikaelweave (approved):

src/Microsoft.Health.Fhir.SqlServer/Features/Watchdogs/CleanupEventLogWatchdog.cs
```diff
@@ -241,11 +248,11 @@ ORDER BY
 
         private async Task DropTmpProceduresAsync(CancellationToken cancellationToken)
         {
-            using var cmd = new SqlCommand("IF object_id('tmp_GetMaxSurrogateId') IS NOT NULL DROP PROCEDURE dbo.tmp_GetMaxSurrogateId");
+            using var cmd = new SqlCommand("DROP PROCEDURE dbo.tmp_GetMaxSurrogateId");
```
mikaelweave, line 251: Do we need to drop these only if they exist? Would it hurt to leave this defense check in place?

mikaelweave (approved):

src/Microsoft.Health.Fhir.SqlServer/Features/Watchdogs/CleanupEventLogWatchdog.cs
SergeyGaluzo: As a general rule, any code should have as little protection as necessary. This particular workflow guarantees that resources are always present when we attempt to drop. Previous "if exist" logic was hiding a bug when the same drop command was used unintentionally twice but, because of "if exists" did not cause any failures, so this was noticed only later, and hence this fix.

PR #4995, opened 2025-05-15
Fix missing headers and content for conditional create.

tarunmathew12:
## Description
The change will address an issue where conditional create fails to return the content and appropriate headers when the condition provided matches a resource already existing. Also, a handling of the Prefer header in a request is added for returning the content in a requested form.

## Related issues
Addresses [issue #146261].
[User Story 146261](https://microsofthealth.visualstudio.com/Health/_workitems/edit/146261): As an Api Developer I'd like Content-Location, Etag and LastModified headers returned on Conditional Create and Conditional Update requests

## Testing
Tested manually and also by adding e2e test.

## FHIR Team Checklist


src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Resources/ResourceHandlerTests_ConditionalCreate.cs
```diff
@@ -56,7 +56,8 @@
 
             UpsertResourceResponse result = await _mediator.Send<UpsertResourceResponse>(message);
 
-            Assert.Null(result);
+            Assert.NotNull(result?.Outcome);
+            Assert.Equal(SaveOutcomeType.MatchFound, result.Outcome.Outcome);
```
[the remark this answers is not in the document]
tarunmathew12: Won't fix: both result and Outcome shouldn't be null

src/Microsoft.Health.Fhir.Shared.Api/Controllers/FhirController.cs
```diff
@@ -196,17 +193,71 @@ public async Task<IActionResult> ConditionalCreate([FromBody] Resource resource)
...
+
+                            result = minimalResult;
+                        }
+
+                        break;
+
+                    case ReturnPreference.OperationOutcome:
+                        {
+                            var operationOutcome = new OperationOutcome();
+                            var message = createResponse.Outcome.Outcome == SaveOutcomeType.Created ? "A new resource was created." : "A resource already exists.";
```
brendankowitz, line 231: User text should be in resx

src/Microsoft.Health.Fhir.Shared.Api/Controllers/FhirController.cs
```diff
@@ -196,17 +193,71 @@ public async Task<IActionResult> ConditionalCreate([FromBody] Resource resource)
...
+            var result = FhirResult.Create(resourceElement, statusCode)
                 .SetETagHeader()
                 .SetLastModifiedHeader()
                 .SetLocationHeader(_urlResolver);
+            if (preferHeader != null)
+            {
+                switch (preferHeader)
+                {
+                    case ReturnPreference.Minimal:
+                        {
```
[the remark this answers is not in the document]
tarunmathew12: Fixed as suggested.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/FhirController.cs
```diff
@@ -196,17 +193,65 @@ public async Task<IActionResult> ConditionalCreate([FromBody] Resource resource)
...
-            return FhirResult.Create(response, HttpStatusCode.Created)
+            IResourceElement resourceElement = createResponse.Outcome.RawResourceElement;
+            var statusCode = createResponse.Outcome.Outcome == SaveOutcomeType.Created ? HttpStatusCode.Created : HttpStatusCode.OK;
+            var result = FhirResult.Create(resourceElement, statusCode)
                 .SetETagHeader()
                 .SetLastModifiedHeader()
                 .SetLocationHeader(_urlResolver);
+            if (preferHeader != null)
+            {
+                switch (preferHeader)
```
brendankowitz: I think the changes in this method could use some improvement, for clarity, reuse and maintenance reasons:

1. Could we consider moving this logic to `FhirResult.Create(resourceElement, statusCode, returnPreference)` (happy to discuss alternatives if this doesn't work).
2. The switch statement does not cover the "default" condition, if it doesn't need to, then I'd suggest it should be an `if` statement instead of a switch with a silent fall through.

In principle we should aim to keep controller code to be as minimal as possible, which is why most methods offload into mediator at the start and FhirResult methods handle most of the return/result options.

src/Microsoft.Health.Fhir.Core/Extensions/PreferHeaderExtensions.cs
```diff
@@ -51,5 +51,33 @@ internal static bool GetIsStrictHandlingEnabled(this RequestContextAccessor<IFhi
 
             return null;
         }
+
+        internal static ReturnPreference? GetReturnPreferenceValue(this RequestContextAccessor<IFhirRequestContext> contextAccessor)
```
brendankowitz, line 55: I see this is implicitly covered, but can we add a unit test for just this method?

src/Microsoft.Health.Fhir.Shared.Api/Controllers/FhirController.cs
v-isyamauchi-gh: Fixed as suggested.

src/Microsoft.Health.Fhir.Core/Extensions/PreferHeaderExtensions.cs
v-isyamauchi-gh: Added UTs for two methods in the prefer header extensions.

src/Microsoft.Health.Fhir.Core/Extensions/PreferHeaderExtensions.cs
```diff
@@ -51,5 +51,33 @@ internal static bool GetIsStrictHandlingEnabled(this RequestContextAccessor<IFhi
...
+            {
+                var returnValue = values.SelectMany(x => x.Split(',', StringSplitOptions.TrimEntries)).FirstOrDefault(x => x.StartsWith("return=", StringComparison.OrdinalIgnoreCase));
+                if (returnValue?.Any() ?? false)
+                {
+                    returnValue = returnValue.Substring("return=".Length);
+
+                    if (string.IsNullOrWhiteSpace(returnValue) ||
+                        !Enum.TryParse(returnValue, true, out ReturnPreference returnPreference))
+                    {
+                        throw new BadRequestException(string.Format(
```
mikaelweave, line 70: It seems the values of the prefer header are case sensitive. Is this correct? I think this is fine, mainly curious.

mikaelweave (approved):

src/Microsoft.Health.Fhir.Core/Extensions/PreferHeaderExtensions.cs
v-isyamauchi-gh: Per offline conversation with Mikael, we are okay with comparing the header value in the case-insensitive way.

PR #4999, opened 2025-05-20
Bulk Delete - add option for immutable resource types

paorodma-ms:
## Description
Added a parameter that can receive a list of resources to exclude from the bulk delete operation.

## Related issues
[Addresses issue #139076.](https://microsofthealth.visualstudio.com/Health/_workitems/edit/139076)

## Testing
Unit tests were rerun and using http bulk delete file with a local instance.

## FHIR Team Checklist


docs/rest/BulkDelete.http
```diff
@@ -67,3 +66,27 @@ Authorization: Bearer {{bearer.response.body.access_token}}
...
+GET {{hostname}}/Patient
+Authorization: Bearer {{bearer.response.body.access_token}}
+
+### Get observations
+GET {{hostname}}/Observation
+Authorization: Bearer {{bearer.response.body.access_token}}
+
+### Get groups
+GET {{hostname}}/Group
+Authorization: Bearer {{bearer.response.body.access_token}}
```
LTA-Thinking, line 80: NIT: I don't think these GETs are needed in this file.

docs/rest/BulkDelete.http
```diff
@@ -67,3 +66,27 @@ Authorization: Bearer {{bearer.response.body.access_token}}
...
+# @name bulkDelete
+DELETE https://{{hostname}}/$bulk-delete?excludedResourceTypes=patient
+Prefer: respond-async
+Authorization: Bearer {{bearer.response.body.access_token}}
+
+### Bulk Delete everything
+# @name bulkDelete
+DELETE https://{{hostname}}/$bulk-delete
+Prefer: respond-async
+Authorization: Bearer {{bearer.response.body.access_token}}
```
LTA-Thinking, line 92: I intentionally didn't include this example because of how dangerous it is to run it.

docs/rest/BulkDelete.http
```diff
@@ -67,3 +66,27 @@ Authorization: Bearer {{bearer.response.body.access_token}}
...
+GET {{hostname}}/Observation
+Authorization: Bearer {{bearer.response.body.access_token}}
+
+### Get groups
+GET {{hostname}}/Group
+Authorization: Bearer {{bearer.response.body.access_token}}
+
+### Bulk Delete everything except patients
+# @name bulkDelete
+DELETE https://{{hostname}}/$bulk-delete?excludedResourceTypes=patient
```
LTA-Thinking, line 84: Can you add the _tag=oldData filter to the search to prevent unintended data being deleted.

docs/rest/BulkDelete.http
paorodma-ms: Removed.

docs/rest/BulkDelete.http
paorodma-ms: Removed.

docs/rest/BulkDelete.http
paorodma-ms: Done

LTA-Thinking (approved):
There are no warnings if the resource type is misspelled, but I think this is ok to move forward with. That may be a future enhancement.

PR #5007, opened 2025-05-22
Updating the error message during conditional reference

v-rachitsh:
## Description
The PR updates the error message of conditional references when multiple resources are found. Instead of stating "does not resolve to a resource," the message will now indicate that the reference resolved to multiple resources.

## Related issues
Addresses [issue #77964](https://microsofthealth.visualstudio.com/Health/_workitems/edit/77964).

## Testing
These changes were tested on the OSS FHIR server, and we received the expected error output.

## FHIR Team Checklist


src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/ResourceReferenceResolver.cs
```diff
@@ -74,6 +74,10 @@ public async Task ResolveReferencesAsync(Resource resource, IDictionary<string,
                         {
```
LTA-Thinking, line 74: This check needs to be changed to be where results equal 0 instead of not equal to 1, otherwise the new else if won't be reached.

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/ResourceReferenceResolver.cs
```diff
@@ -74,6 +74,10 @@ public async Task ResolveReferencesAsync(Resource resource, IDictionary<string,
...
                             throw new RequestNotValidException(string.Format(Core.Resources.InvalidConditionalReference, reference.Reference));
                         }
+                        else if (results.Count > 1)
```
LTA-Thinking, line 77: The count needs to be the matched results, so this needs the same filter as the if statement above it.

LTA-Thinking (approved):

PR #5011, opened 2025-05-30
ADR on eventual consistency

SergeyGaluzo:
This is the result of our discussions on Wednesday-Friday this week - introduction of eventual consistency in FHIR, currently for $import only

### DIFF
--- a/docs/arch/adr-2505-eventual-consistency.md
+++ b/docs/arch/adr-2505-eventual-consistency.md
@@ -1,0 +1,20 @@
+# ADR 2505: Enable eventual consistency for maintaining search param indexes in sync with the Resource table
+Labels: [SQL](https://github.com/microsoft/fhir-server/labels/Area-SQL)
+
+## Context
+Efficiency of writes to FHIR SQL storage is higly dependent on duration of SQL transactions. Long SQL transactions cause significant blocking between parallel ingestion processes and reduce overall write throughput. This is usually observered for large hyperscale databases (>1TB). Currently FHIR SQL supports strong consistency when updates in the Resource table and the search param tables are included in the same SQL transaction. This makes SQL transactions long.
+
+## Decision
+We will implement eventual consistency for maintaining search param indexes. Updates of resources will still run in single SQL transaction as today (strong consistency), because they involve deletion of previous search param rows. For resource creates only, we will run all writes in separate SQL transactions, starting with insert into the Resource table. In case of success this is equivalent to current strong consistency behavior. In case of interruptions search param indexes might not be completely in sync with corresponding row in the Resource table. This will be addressed by the Transaction watchdog, which will roll transaction forward based on data committed in the Resource table.
+
+Initially, this change will be enabled for $import only. We will introduce new flag "EventualConsistency" which will accept true/false values with default value false (strong consistency, current behavior).
+
+## Status
+Accepted
+
+## Consequences
+### Benefits:
+- Improved write througput for large hyperscale databases (>1TB)
+
+### Neutral Effects:
+- No impact on current customer logic.


fhibf (approved):

docs/arch/adr-2505-eventual-consistency.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Status
+Accepted
+
+## Consequences
+### Benefits:
+- Improved write througput for large hyperscale databases (>1TB)
+
+### Neutral Effects:
+- No impact on current customer logic.
```
mikaelweave, line 20: Can you explain the zero impact a bit more? A common scenario is as below - can you confirm there is no risk to changed behavior?

1. Patient is created in FHIR with an AllergyIntollerance in a transaction bundle.
2. Application reads FHIR after save. This application loads the patient by ID (GET) and on the patient display page it runs a query for any AllergyIntolerance resources with this patient.

Thanks!

SergeyGaluzo: The plan is to run writes with eventual consistency only with user consent. In $import it is a new parameter called "EventualConsistency" with default value FALSE. In future similar behavior can be introduced in bundles. In any case, all current writes will run with eventual consistency OFF. This is our current behavior, so customers will not see anything new, if they do not specifically ask for it, hence "no impact on current customer logic."

rajithaalurims:
@SergeyGaluzo Can you explain what happens in this scenario with EventualConsistency flag ON?

1. Customer runs incremental import loading some resources.
2. Customer also kicked off Reindex.
3. If Reindex finishes indexing first, how does it impact the logic that loads XXXSearchParameters for the import in a separate transaction.

docs/arch/adr-2505-eventual-consistency.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Context
+Efficiency of writes to FHIR SQL storage is higly dependent on duration of SQL transactions. Long SQL transactions cause significant blocking between parallel ingestion processes and reduce overall write throughput. This is usually observered for large hyperscale databases (>1TB). Currently FHIR SQL supports strong consistency when updates in the Resource table and the search param tables are included in the same SQL transaction. This makes SQL transactions long.
+
+## Decision
+We will implement eventual consistency for maintaining search param indexes. Updates of resources will still run in single SQL transaction as today (strong consistency), because they involve deletion of previous search param rows. For resource creates only, we will run all writes in separate SQL transactions, starting with insert into the Resource table. In case of success this is equivalent to current strong consistency behavior. In case of interruptions search param indexes might not be completely in sync with corresponding row in the Resource table. This will be addressed by the Transaction watchdog, which will roll transaction forward based on data committed in the Resource table.
```
brendankowitz, line 8: So the changes are?
1. Turning off transactions held from c# code
2. Import uses transaction table id on imported data
3. Watchdog ensures consistency and rolls transactions forward

SergeyGaluzo:
@rajithaalurims As long as definitions of search parameters are the same for reindex and update of XXXSearchParam tables, update after reindex will be NOOP.

docs/arch/adr-2505-eventual-consistency.md
SergeyGaluzo: C# transactions... 
$import does not use C# transactions, so not applicable to what we plan to do next. 
In future, if we want to extend eventual consistency usage for bundles and other API calls, we will not do this for cases where C# transactions are needed. This will work as it is today. For cases where all PUTs in bundle are aggregated into a single SQL call, we can use the same method $import will be using (new input parameter).

Today all writes, $import and any calls via API including, are already using Transactions table, so no changes here.

Today Transaction watchdog already deals with timeout transactions and rolls them forward, so no changes here.

mikaelweave (approved):

PR #5022, opened 2025-06-13
Bulk-update

apurvabhaleMS:
## Description
Adds the $bulk-update operation. PATCH\$bulk-update
This operation can be called at the system level or the resource type level.
As with our other async jobs it returned a Content-Location header showing the endpoint where the job record can be retrieved from. This endpoint can be used to get the current status of the job (GET) or to cancel it (DELETE).

This job is limited by the integer max value due to reporting the ids of deleted resources. This means that jobs should be limited to impact no more than ~2 billion resources, probably less than 500 million to be safe.

Added to capabilities statement.
Added notification on job competition.
Added unit tests.
Rest examples added

Supports the following parameters:

_isParallel: Flag for whether to run subjobs parallely. Default true.
<search parms>: Any parameters used for a normal search operation can be added.

## Related issues
Addresses [AB146980](https://microsofthealth.visualstudio.com/Health/_workitems/edit/146980).

## Testing

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Upsert/BulkUpdateService.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Upsert/BulkUpdateService.cs
@@ -669,7 +669,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Persistence
         {
             if (results == null)
             {
-                return new Dictionary<string, long>();
+                results = new Dictionary<string, long>();
             }
 
             foreach (var newResult in newResults)


src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Queues/CosmosQueueClient.cs
```diff
@@ -333,6 +334,13 @@ public async Task<IReadOnlyList<JobInfo>> GetJobByGroupIdAsync(byte queueType, l
         return jobs.SelectMany(job => job.ToJobInfo(job.Definitions)).ToList();
     }
 
+    /// <inheritdoc />
+    public async Task<IReadOnlyList<JobInfo>> GetActiveJobsByQueueTypeAsync(byte queueType, CancellationToken cancellationToken)
```
LTA-Thinking: Is Cosmos going to support bulk update?

src/Microsoft.Health.Fhir.Shared.Api/Controllers/BulkUpdateController.cs
```diff
@@ -0,0 +1,119 @@
...
+            CreateBulkUpdateResponse result = await _mediator.BulkUpdateAsync(typeParameter, searchParameters, parameters, isParallel ?? true, HttpContext.RequestAborted);
+
+            var response = JobResult.Accepted();
+            response.SetContentLocationHeader(_urlResolver, OperationsConstants.BulkUpdate, result.Id.ToString());
+            return response;
+        }
+
+        private void CheckIfOperationIsSupported()
+        {
+            if (!_fhirRuntimeConfiguration.DataStore?.Equals(KnownDataStores.SqlServer, StringComparison.OrdinalIgnoreCase) ?? true)
```
LTA-Thinking: Since you also have a configuration flag for if this is supported you can also check that here.

src/Microsoft.Health.Fhir.Core/Data/OperationDefinition/bulk-update.json
```diff
@@ -0,0 +1,15 @@
...
+    "version": "1.0.0",
+    "name": "Bulk Update",
+    "status": "active",
+    "kind": "operation",
+    "description": "Updates all data from a FHIR server that matches the search parameters. Any supported search parameter can be inluded as query string parameters. This operation is asynchronous as defined in the [FHIR Asynchronous Request Pattern](http://hl7.org/fhir/async.html)",
+    "code": "bulk-update",
+    "system": true,
+    "type": true,
+    "instance": false,
+    "parameter": []
```
LTA-Thinking: Need to fill in parameters.

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkUpdate/BulkUpdateOrchestratorJob.cs
```diff
@@ -0,0 +1,223 @@
...
+                var atLeastOneWorkerJobRegistered = false;
+
+                // For system level bulk update, get all the resource types that are used in the system and enqueue jobs for each resource type-surrogateId range combination.
+                // For ResourceType based bulk updates query with includes and revinclude could return multiple resourceTypes so we will do the generic search without surrogateId ranges
+                var resourceTypes = await searchService.Value.GetUsedResourceTypes(cancellationToken);
+                resourceTypes = resourceTypes
+                    .Where(x => !_excludedResourceTypes.Contains(x))
+                    .ToList();
+
+                resourceTypes = resourceTypes.OrderByDescending(x => string.Equals(x, "Observation", StringComparison.OrdinalIgnoreCase)).ToList(); // true first, so observation is processed as soon as
```
LTA-Thinking: The Observation first thing was a leftover from export where we had noticed that observation was the most common resource type in general. This meant it would start getting processed first as it would take the longest. I don't know if this is a pattern worth continuing, but it also doesn't hurt anything.

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkUpdate/BulkUpdateOrchestratorJob.cs
```diff
@@ -0,0 +1,223 @@
...
+            }
+            else if (definition.IsParallel && !string.IsNullOrEmpty(definition.Type))
+            {
+                // read all the data from conditional search using Continuation token and create processing definitions and enqueue them by storing continuation token as well
+                IReadOnlyCollection<SearchResultEntry> searchResults;
+                string ct;
+                string ict;
+                string prevCT = null;
+                var searchParams = definition.SearchParameters.ToList();
+                (searchResults, ct, ict) = await searchService.Value.ConditionalSearchAsync(
```
LTA-Thinking: Why use conditional search instead of regular search?

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkUpdate/BulkUpdateOrchestratorJob.cs
```diff
@@ -0,0 +1,223 @@
...
+
+            BulkUpdateDefinition definition = jobInfo.DeserializeDefinition<BulkUpdateDefinition>();
+            var surrogateIdRangeSize = (int)definition.MaximumNumberOfResourcesPerQuery;
+
+            _logger.LogJobInformation(jobInfo, "Loading job by Group Id.");
+            var groupJobs = await _queueClient.GetJobByGroupIdAsync(QueueType.BulkUpdate, jobInfo.GroupId, true, cancellationToken);
+
+            // If system level Parallel bulk update then create sub jobs by resourceType-surrogateId ranges
+            using var searchService = _searchService.Invoke();
+            if (definition.IsParallel && string.IsNullOrEmpty(definition.Type))
```
LTA-Thinking: Why is definition.Type being used as the splitting condition? It looks like definition.SearchParameters would make a better splitting condition, since if all the user specified was type and not any search parameters we could still split by surrogate id.

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkUpdate/BulkUpdateOrchestratorJob.cs
```diff
@@ -0,0 +1,223 @@
...
+
+                while (searchResults.Any() || !string.IsNullOrEmpty(prevCT))
+                {
+                    _logger.LogJobInformation(jobInfo, "Creating bulk update definition (3).");
+                    var processingRecord = CreateProcessingDefinition(definition, searchService.Value, cancellationToken, definition.Type, prevCT, false);
+
+                    _logger.LogJobInformation(jobInfo, "Enqueuing bulk update job (3).");
+                    await _queueClient.EnqueueAsync(QueueType.BulkUpdate, cancellationToken, groupId: jobInfo.GroupId, definitions: processingRecord);
+
+                    if (ct is null && ict is null)
```
LTA-Thinking: This is dangerous. If the last page of results (IE: ct is null) has an includes continuation token this will cause the search to run again, but with null as the continuation token. This will get the first page of results again and cause the job creation to enter an infinite loop.

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkUpdate/BulkUpdateProcessingJob.cs
```diff
@@ -0,0 +1,129 @@
...
+                }
+                catch (IncompleteOperationException<BulkUpdateResult> ex)
+                {
+                    resourcesUpdated = ex.PartialResults.ResourcesUpdated;
+                    result = ex.PartialResults;
+                    result.Issues.Add(ex.Message);
+                    exception = ex;
+                }
+
+                await _mediator.Publish(new BulkUpdateMetricsNotification(jobInfo.Id, resourcesUpdated.Sum(resource => resource.Value)), cancellationToken);
```
LTA-Thinking: When is resourcesUpdated set?

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkUpdate/BulkUpdateProcessingJob.cs
```diff
@@ -0,0 +1,129 @@
...
+                var tillTime = new PartialDateTime(jobInfo.CreateDate);
+                var queryParametersList = new List<Tuple<string, string>>();
+                queryParametersList.AddRange(definition.SearchParameters);
+
+                queryParametersList.Add(Tuple.Create(KnownQueryParameterNames.Count, definition.MaximumNumberOfResourcesPerQuery.ToString(CultureInfo.InvariantCulture)));
+                queryParametersList.Add(Tuple.Create(KnownQueryParameterNames.LastUpdated, $"le{tillTime}"));
+
+                if (definition.GlobalEndSurrogateId != null) // no need to check individually as they all should have values if anyone does
+                {
+                    queryParametersList.Add(Tuple.Create(KnownQueryParameterNames.Type, definition.Type));
```
LTA-Thinking, line 110: Why does type only need to be added when the surrogate id filters are used?

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/CustomFhirJsonSerializer.cs
```diff
@@ -0,0 +1,38 @@
...
+using Hl7.Fhir.Serialization;
+using Microsoft.Health.Fhir.Core.Features.Operations;
+using Microsoft.Health.Fhir.Core.Features.Operations.Import;
+using Microsoft.Health.Fhir.Core.Features.Resources.Patch;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Health.Fhir.Core.Features.Operations
+{
+    public class CustomFhirJsonSerializer<T> : ICustomFhirJsonSerializer<T>
```
LTA-Thinking: Why is this class needed instead of using the existing resource serializer and deserializer?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlQueueClient.cs
```diff
@@ -213,6 +213,13 @@ public async Task<IReadOnlyList<JobInfo>> GetJobsByIdsAsync(byte queueType, long
             return await cmd.ExecuteReaderAsync(_sqlRetryService, JobInfoExtensions.LoadJobInfo, _logger, cancellationToken, "GetJobsByIdsAsync failed.");
         }
 
+        public async Task<IReadOnlyList<JobInfo>> GetActiveJobsByQueueTypeAsync(byte queueType, CancellationToken cancellationToken)
```
SergeyGaluzo: I am not sure what the reason for adding this method is. If it just to prevent more than one bulk-update job running, you should use forceOneActiveJobGroup on Enqueue method.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlQueueClient.cs
```diff
@@ -277,5 +284,12 @@ private static void PopulateGetJobsCommand(SqlCommand cmd, byte queueType, long?
                 cmd.Parameters.AddWithValue("@ReturnDefinition", returnDefinition.Value);
             }
         }
+
+        private static void PopulateGetActiveJobsCommand(SqlCommand cmd, byte queueType)
```
SergeyGaluzo: I am not sure why this is a separate method. Looks like it is called from single place?

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Queues/CosmosQueueClient.cs
apurvabhaleMS: no

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkUpdate/BulkUpdateProcessingJob.cs
apurvabhaleMS: For surrogate id stored proc we need type in queryParam list
for other scenarios we use request.type

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkUpdate/Handlers/GetBulkUpdateHandler.cs
```diff
@@ -0,0 +1,210 @@
...
+                issues.Add(new OperationOutcomeIssue(OperationOutcomeConstants.IssueSeverity.Error, OperationOutcomeConstants.IssueType.Exception, detailsText: "Encountered an unhandled exception. The job will be marked as failed."));
+                return new GetBulkUpdateResponse(fhirResults, issues, failureResultCode);
+            }
+            else if (succeeded)
+            {
+                return new GetBulkUpdateResponse(fhirResults, issues, HttpStatusCode.OK);
+            }
+            else
+            {
+                issues.Add(new OperationOutcomeIssue(
```
LTA-Thinking, line 211: Since this isn't failing the job when it gets an error, this message needs to be returned as long as the job is still running.
Currently if a processing job fails the status endpoint returns a 500 with no indication that the rest of the processing jobs are still running.

docs/rest/BulkUpdate.http
```diff
@@ -0,0 +1,497 @@
...
+                    }
+                }
+            ]
+        }
+    ]
+}
+
+###
+## Add link (used in patient merge)
+PATCH https://{{hostname}}/Patient/$bulk-update/{{patient.response.body.id}}
```
LTA-Thinking: What is this testing? I don't see an endpoint in the controller to match a URL like this one

docs/rest/BulkUpdate.http
```diff
@@ -0,0 +1,497 @@
...
+                        "code": "SECURITY_TAG_CODE",
+                        "display": "Updated Security Tag Display"
+                    }
+                }
+            ]
+        }
+    ]
+}
+
+### Get the status of the bulk update operation
```
LTA-Thinking: Could you move the GET and DELETE to the top of the document so they are easier to find? 
Also, could you add a name to all the bulk update calls and use a reference in these urls instead of a hard coded number. There are examples in bulk delete. This will make it easier to run the samples.

docs/rest/BulkUpdate.http
apurvabhaleMS: I will fix this

src/Microsoft.Health.Fhir.Shared.Client/FhirClient.cs
```diff
@@ -590,7 +605,7 @@ public async Task<FhirResponse<Parameters>> WaitForReindexStatus(Uri reindexJobU
             return reindexJobResult;
         }
 
-        public async Task<FhirResponse<Parameters>> WaitForBulkDeleteStatus(Uri bulkDeleteJobUri)
+        public async Task<FhirResponse<Parameters>> WaitForBulkJobStatus(string jobType, Uri bulkDeleteJobUri)
```
LTA-Thinking: Nit: Since this isn't about bulk delete anymore can you update the Uri name?

src/Microsoft.Health.TaskManagement/JobHosting.cs
```diff
@@ -195,6 +195,31 @@ private async Task ExecuteJobAsync(JobInfo jobInfo)
...
             }
+            catch (JobExecutionSoftFailureException ex)
+            {
+                if (ex.IsCustomerCaused)
+                {
+                    _logger.LogJobWarning(ex, jobInfo, "Job failed due to customer caused issue.", jobInfo.Id, jobInfo.GroupId, jobInfo.QueueType);
+                }
+                else
+                {
+                    _logger.LogJobError(ex, jobInfo, "Job failed.", jobInfo.Id, jobInfo.GroupId, jobInfo.QueueType);
```
LTA-Thinking: Can you add something to the messages to show they come from a soft failure.

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Patch/FhirPathPatch/Operations/OperationUpsert.cs
```diff
@@ -0,0 +1,99 @@
...
+
+                if (targetAtName.Count == 0) // If the element doesn't exist
+                {
+                    // If we couldn't find an element or a collection then, we can ADD it
+                    try
+                    {
+                        Target = ResourceElement
+                                    .Select(Operation.Path)
+                                    .RequireOneOrMoreElements()
+                                    .RequireMultipleElementsInSameCollection()
```
LTA-Thinking: What does this line do that means we have to get the target a second time?

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkUpdate/Handlers/CancelBulkUpdateHandler.cs
```diff
@@ -54,9 +59,34 @@ public async Task<CancelBulkUpdateResponse> Handle(CancelBulkUpdateRequest reque
...
+                var needRefresh = false;
+
+                // Check if any completed job in the group updated a profile resource type
+                var profileTypes = _supportedProfiles.GetProfilesTypes();
+
                 foreach (var job in jobs)
                 {
-                    if (job.Status == JobStatus.Failed || job.Status == JobStatus.Cancelled)
+                    BulkUpdateResult bulkUpdateResult;
+                    bool softDeleted = false;
```
LTA-Thinking: I don't know if soft deleted is the best variable name here. It makes me think of soft deleting a resource. I think here you are using it to test if a job had a soft failure.

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/BulkUpdate/CancelBulkUpdateHandlerTests.cs
```diff
@@ -0,0 +1,227 @@
...
+                    new JobInfo()
+                    {
+                        Status = JobStatus.Completed,
+                    },
+                },
+                HttpStatusCode.Conflict);
+        }
+
+        [Fact]
+        public async Task GivenFailedBulkUpdateJob_WhenCancelationIsRequested_ThenConflictIsReturned()
```
LTA-Thinking, line 83: Can there also be a test for a soft failure?

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Patch/FhirPathPatch/Operations/OperationUpsert.cs
apurvabhaleMS: I checked in new code. But this line will only allow multiple elements if they are a collection and are from the same collection

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkUpdate/Handlers/CancelBulkUpdateHandler.cs
apurvabhaleMS: yes, I agree :)

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/BulkUpdate/GetBulkUpdateHandlerTests.cs
```diff
@@ -0,0 +1,588 @@
...
+                        {
+                            Status = JobStatus.Running,
+                        },
+                        0),
+                },
+                new GetBulkUpdateResponse(ToParameters(resultsDictionary).ToArray(), issues, System.Net.HttpStatusCode.InternalServerError));
+        }
+
+        [Fact]
+        public async Task GivenSoftFailedBulkUpdateJob_WhenStatusRequestedAndJobIsComplete_ThenStatusIsReturned()
```
LTA-Thinking, line 412: This is a duplicate of the test GivenSoftFailedBulkUpdateJob_WhenStatusRequested_ThenStatusIsReturned on line 182.

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkUpdate/BulkUpdateOrchestratorJob.cs
```diff
@@ -0,0 +1,295 @@
...
+                            rows = definitions.Count;
+                            if (rows > 0)
+                            {
+                                _logger.LogJobInformation(jobInfo, "Enqueuing bulk update job (1).");
+                                await _queueClient.EnqueueAsync(QueueType.BulkUpdate, cancel, groupId: jobInfo.GroupId, definitions: definitions.ToArray());
+                            }
+                        }
+                    });
+                }
+                else if (definition.IsParallel)
```
LTA-Thinking, line 149: This case isn't rerun safe. If the orchestrator job crashes in the middle of executing it may create multiple identical processing jobs.

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkUpdate/BulkUpdateProcessingJob.cs
```diff
@@ -0,0 +1,203 @@
...
+
+                // Get all jobs for the group
+                var jobs = (await _queueClient.GetJobByGroupIdAsync(QueueType.BulkUpdate, jobInfo.GroupId, true, cancellationToken)).ToList();
+
+                // Filter out the current job and group job, and keep only active subjobs
+                var activeJobs = jobs.Where(j => j.Id != jobInfo.Id && j.Id != jobInfo.GroupId &&
+                    (j.Status == JobStatus.Created || j.Status == JobStatus.Running)).ToList();
+
+                // Only proceed if this is the last active subjob in the group
+                if (activeJobs.Count == 0)
```
LTA-Thinking, line 164: Could you add a log to show it made it into this if statement? Just to say it has detected this is the last job and it is checking the profile types. I think it will make debugging easier to know if it made it into this check.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/BulkUpdateTests.cs
```diff
@@ -0,0 +1,658 @@
...
+            BulkUpdateResult expectedResults = new BulkUpdateResult();
+            expectedResults.ResourcesUpdated.Add("Patient", 1);
+            expectedResults.ResourcesUpdated.Add("Observation", 1);
+            expectedResults.ResourcesUpdated.Add("Encounter", 1);
+            await MonitorBulkUpdateJob(response.Content.Headers.ContentLocation, expectedResults);
+        }
+
+        [SkippableFact]
+        [HttpIntegrationFixtureArgumentSets(DataStore.SqlServer, Format.Json)]
+        public async Task GivenBulkUpdateJobWithMoreThanOnePageOfIncludeResultsAndIsParallelTrue_WhenCompleted_ThenIncludedResultsAreUpdated()
```
LTA-Thinking: Could this and the test below it be a Theory instead of two Facts?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/BulkUpdateTests.cs
```diff
@@ -0,0 +1,658 @@
...
+
+        public BulkUpdateTests(HttpIntegrationTestFixture fixture)
+        {
+            _fixture = fixture;
+            _httpClient = fixture.HttpClient;
+            _fhirClient = fixture.TestFhirClient;
+        }
+
+        [SkippableFact]
+        public async Task GivenVariousResourcesOfDifferentTypesAndIsParallelTrue_WhenBulkUpdated_ThenAllAreUpdated()
```
LTA-Thinking: Could this test and the one below it be a Theory instead of two Facts?

docs/rest/BulkUpdate.http
LTA-Thinking, line 1: I love all the examples you've provided here. This is really helpful.

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/BulkUpdate/BulkUpdateProcessingJobTests.cs
```diff
@@ -0,0 +1,472 @@
...
+            _updater.UpdateMultipleAsync(Arg.Any<string>(), Arg.Any<string>(), Arg.Any<int>(), Arg.Any<bool>(), Arg.Any<uint>(), Arg.Any<bool>(), Arg.Any<IReadOnlyList<Tuple<string, string>>>(), Arg.Any<BundleResourceContext>(), Arg.Any<CancellationToken>())
+                .Returns(args => substituteResults);
+
+            var result = JsonConvert.DeserializeObject<BulkUpdateResult>(await _processingJob.ExecuteAsync(jobInfo, CancellationToken.None));
+            Assert.Equal(3, result.ResourcesUpdated["Patient"]);
+            Assert.Equal(1, result.ResourcesUpdated["Observation"]);
+            Assert.Single(result.ResourcesIgnored);
+            Assert.Equal(2, result.ResourcesIgnored["CarePlan"]);
+
+            await _updater.ReceivedWithAnyArgs(1).UpdateMultipleAsync(Arg.Any<string>(), Arg.Any<string>(), Arg.Any<int>(), Arg.Any<bool>(), Arg.Any<uint>(), Arg.Any<bool>(), Arg.Any<IReadOnlyList<Tuple<string, string>>>(), Arg.Any<BundleResourceContext>(), Arg.Any<CancellationToken>());
```
LTA-Thinking: Since the below test checks that true is passed when an includes CT is present, can this test check that false is passed when there is no includes CT?

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/BulkUpdate/BulkUpdateProcessingJobTests.cs
```diff
@@ -0,0 +1,472 @@
...
+            var result = JsonConvert.DeserializeObject<BulkUpdateResult>(await _processingJob.ExecuteAsync(jobInfo, CancellationToken.None));
+            Assert.Single(result.ResourcesUpdated);
+            Assert.Equal(3, result.ResourcesUpdated["Patient"]);
+
+            await _updater.ReceivedWithAnyArgs(1).UpdateMultipleAsync(Arg.Any<string>(), Arg.Any<string>(), Arg.Any<int>(), Arg.Any<bool>(), Arg.Any<uint>(), Arg.Any<bool>(), Arg.Any<IReadOnlyList<Tuple<string, string>>>(), Arg.Any<BundleResourceContext>(), Arg.Any<CancellationToken>());
+            await _mediator.Received(1).Publish(Arg.Is<BulkUpdateMetricsNotification>(n => n.JobId == jobInfo.Id && n.ResourcesUpdated == 3), Arg.Any<CancellationToken>());
+        }
+
+        [Fact]
+        public async Task GivenProcessingJobLastOneFromGroupAndProfileResourceTypeIsUpdatedInPreviouslyCompleteJob_WhenJobIsRun_ThenProfilesAreRefreshed()
```
LTA-Thinking: Could this test and the one below it be made into a theory?

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/BulkUpdate/BulkUpdateOrchestratorJobTests.cs
```diff
@@ -0,0 +1,613 @@
...
+            };
+
+            await _orchestratorJob.ExecuteAsync(jobInfo, CancellationToken.None);
+            await _queueClient.ReceivedWithAnyArgs(1).EnqueueAsync(Arg.Any<byte>(), Arg.Any<string[]>(), Arg.Any<long?>(), false, Arg.Any<CancellationToken>());
+            await _searchService.DidNotReceiveWithAnyArgs().GetUsedResourceTypes(Arg.Any<CancellationToken>());
+
+            // Checks that 6 processing jobs were queued
+            var calls = _queueClient.ReceivedCalls();
+            var definitions = calls
+                .Skip(1) // Skip the first call (index 0)
```
LTA-Thinking: Why does it need to skip the first call?

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/BulkUpdate/BulkUpdateOrchestratorJobTests.cs
```diff
@@ -0,0 +1,613 @@
...
+                GroupId = 1,
+                Definition = JsonConvert.SerializeObject(definition),
+                CreateDate = DateTime.Now,
+            };
+
+            await _orchestratorJob.ExecuteAsync(jobInfo, CancellationToken.None);
+            await _queueClient.ReceivedWithAnyArgs(1).EnqueueAsync(Arg.Any<byte>(), Arg.Any<string[]>(), Arg.Any<long?>(), false, Arg.Any<CancellationToken>());
+            await _searchService.DidNotReceiveWithAnyArgs().GetUsedResourceTypes(Arg.Any<CancellationToken>());
+
+            // Checks that 6 processing jobs were queued
```
LTA-Thinking: Doesn't this test only have 1 processing job?

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/BulkUpdate/BulkUpdateOrchestratorJobTests.cs
```diff
@@ -0,0 +1,613 @@
...
+            Assert.Single(definitions);
+
+            var actualDefinition = JsonConvert.DeserializeObject<BulkUpdateDefinition>(definitions[0]);
+            Assert.Null(actualDefinition.Type);
+            Assert.Null(actualDefinition.StartSurrogateId);
+            Assert.Null(actualDefinition.EndSurrogateId);
+        }
+
+        [Fact]
+        public async Task GivenBulkUpdateJob_WhenIsParallelIsTrueWithNoSearchParamAndGetUsedResourceTypesReturnsEmpty_ThenNoProcessingJobsAreEnqueued()
```
LTA-Thinking: Have you tested how the GET bulk update endpoint works for a job that had no processing jobs?

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Upsert/BulkUpdateService.cs
```diff
@@ -0,0 +1,657 @@
...
+        private readonly IResourceWrapperFactory _resourceWrapperFactory;
+        private readonly Lazy<IConformanceProvider> _conformanceProvider;
+        private readonly IScopeProvider<IFhirDataStore> _fhirDataStoreFactory;
+        private readonly IScopeProvider<ISearchService> _searchServiceFactory;
+        private readonly ResourceIdProvider _resourceIdProvider;
+        private readonly FhirRequestContextAccessor _contextAccessor;
+        private readonly IAuditLogger _auditLogger;
+        private readonly CoreFeatureConfiguration _configuration;
+        private readonly ILogger<BulkUpdateService> _logger;
+        private readonly HashSet<string> _excludedResourceTypes = ["SearchParameter", "StructureDefinition"];
```
LTA-Thinking: Should this be in a constants file? This list is defined in at least three files and I'm concerned about them getting out of sync for some reason.

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Upsert/BulkUpdateService.cs
```diff
@@ -0,0 +1,657 @@
...
+                    customHeaders: null,
+                    operationType: string.Empty,
+                    callerAgent: DefaultCallerAgent,
+                    additionalProperties: updateAdditionalProperties);
+            });
+
+            return auditTask;
+        }
+
+        private bool AreIncludeResultsTruncated()
```
LTA-Thinking, line 658: Since this is now being used in multiple places it may be better to move this to the context accessor, or somewhere else, as a method in a common location. Not needed for this PR, but just thinking about future improvements.

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Upsert/BulkUpdateService.cs
```diff
@@ -0,0 +1,657 @@
...
+
+                    pageOne = false;
+
+                    // Keep reading the next page of results if there are more results to process and when it is not a continuation token level job
+                    if (!string.IsNullOrEmpty(ct) && readNextPage)
+                    {
+                        var cloneList = new List<Tuple<string, string>>(conditionalParameters);
+                        if (isIncludesRequest)
+                        {
+                            // for includerequests, keep ct as is and remove the includesContinuationToken and add new includesContinuationToken to the cloneList
```
LTA-Thinking: nit: missing space

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Upsert/BulkUpdateService.cs
```diff
@@ -0,0 +1,657 @@
...
+                        }
+
+                        searchResult = await Search(resourceType, isIncludesRequest, cloneList, cancellationToken);
+                        ict = searchResult.IncludesContinuationToken;
+
+                        // For resources that are included
+                        if (!isIncludesRequest && !string.IsNullOrEmpty(ict) && AreIncludeResultsTruncated() && readNextPage)
+                        {
+                            // run a search for included results
+                            (resourceTypesUpdated, finalBulkUpdateResult) = await HandleIncludedResources(resourceType, fhirPatchParameters, parallelThreads - updateTasks.Count, readNextPage, maximumNumberOfResourcesPerQuery, conditionalParameters, bundleResourceContext, ct, ict, resourceTypesUpdated, finalBulkUpdateResult, cancellationToken);
```
LTA-Thinking: Do the included results need to be handled here, or will they be handled by line 166 on the next pass of the while loop?

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Upsert/BulkUpdateService.cs
```diff
@@ -0,0 +1,657 @@
...
+            _contextAccessor = EnsureArg.IsNotNull(contextAccessor, nameof(contextAccessor));
+            _auditLogger = EnsureArg.IsNotNull(auditLogger, nameof(auditLogger));
+            _logger = EnsureArg.IsNotNull(logger, nameof(logger));
+            _configuration = EnsureArg.IsNotNull(configuration.Value, nameof(configuration));
+        }
+
+        public async Task<BulkUpdateResult> UpdateMultipleAsync(string resourceType, string fhirPatchParameters, int parallelThreads, bool readNextPage, uint maximumNumberOfResourcesPerQuery, bool isIncludesRequest, IReadOnlyList<Tuple<string, string>> conditionalParameters, BundleResourceContext bundleResourceContext, CancellationToken cancellationToken)
+        {
+            IReadOnlyCollection<SearchResultEntry> searchResults;
+            bool tooManyIncludeResults = false;
```
LTA-Thinking: I don't think this bool is used anywhere.

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Upsert/BulkUpdateService.cs
```diff
@@ -0,0 +1,657 @@
...
+            {
+                totalResources[group.Key] = totalResources.TryGetValue(group.Key, out var existing)
+                    ? existing + group.Value
+                    : group.Value;
+            }
+
+            // Add resources ignored by resource type by filtering out the excluded resource types
+            foreach (var kvp in resourcesPerPage.Where(kvp => _excludedResourceTypes.Contains(kvp.Key)))
+            {
+                if (resourcesIgnored.TryGetValue(kvp.Key, out long existingValue))
```
LTA-Thinking: The two dictionary updates should use the same strategy to count their values. The first one uses a one line ? while this uses an if/else.

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Upsert/BulkUpdateService.cs
```diff
@@ -0,0 +1,657 @@
...
+            }
+
+            // Filter the resourcesPerPage by removing the excluded resource types
+            foreach (var resource in resourcesIgnored)
+            {
+                resourcesPerPage.Remove(resource.Key);
+            }
+
+            // Filter the resourcesPerPage by removing the commonPatchFailures
+            foreach (var resource in commonPatchFailures)
```
LTA-Thinking: Should the count of resources filtered by common patch failures be added to the count of ignored resources? I'm trying to think of how this can be reported to the customer since we are filtering resources that the search returned.

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Upsert/BulkUpdateService.cs
```diff
@@ -0,0 +1,657 @@
...
+                    if (patchedResources.Any())
+                    {
+                        updateTasks.Add(UpdateResourcePage(patchedResources, resourceType, bundleResourceContext, searchResults, finalBulkUpdateResult, cancellationTokenSource.Token));
+                        if (updateTasks.Any((task) => task.IsFaulted || task.IsCanceled))
+                        {
+                            break;
+                        }
+
+                        resourceTypesUpdated = AppendUpdateResults(resourceTypesUpdated, updateTasks.Where(x => x.IsCompletedSuccessfully).Select(task => task.Result));
+                        AppendUpdateResults(finalBulkUpdateResult.ResourcesUpdated as Dictionary<string, long>, updateTasks.Where(x => x.IsCompletedSuccessfully).Select(task => task.Result));
```
LTA-Thinking: Why does AppendUpdateResults need to be called twice? I'm not seeing where resourceTypesUpdated is used.

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/BulkUpdate/BulkUpdateOrchestratorJobTests.cs
apurvabhaleMS: Because queueClient receives the first call to get the jobs by group id > GetJobByGroupIdAsync

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Upsert/BulkUpdateService.cs
apurvabhaleMS: We are counting the common patch failures and individual patch failures in ResourcesPatchFailed dictionary, check FinalizePatchResultsAndAuditAsync

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/BulkUpdate/BulkUpdateOrchestratorJobTests.cs
```diff
@@ -254,13 +255,14 @@ public async Task GivenBulkUpdateJob_WhenSearchParameterIsGivenAndIsParallelIsTr
             };
 
             await orchestratorJobWithTruncatedIssue.ExecuteAsync(jobInfo, CancellationToken.None);
-            await _queueClient.ReceivedWithAnyArgs(2).EnqueueAsync(Arg.Any<byte>(), Arg.Any<string[]>(), Arg.Any<long?>(), false, Arg.Any<CancellationToken>());
+            await _queueClient.ReceivedWithAnyArgs(1).EnqueueAsync(Arg.Any<byte>(), Arg.Any<string[]>(), Arg.Any<long?>(), false, Arg.Any<CancellationToken>());
             await _searchService.DidNotReceiveWithAnyArgs().GetUsedResourceTypes(Arg.Any<CancellationToken>());
 
             // Checks that 6 processing jobs were queued
```
LTA-Thinking: Nit: I think there are only 2 processing jobs

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Upsert/BulkUpdateService.cs
```diff
@@ -0,0 +1,659 @@
...
+            _conformanceProvider = EnsureArg.IsNotNull(conformanceProvider, nameof(conformanceProvider));
+            _fhirDataStoreFactory = EnsureArg.IsNotNull(fhirDataStoreFactory, nameof(fhirDataStoreFactory));
+            _searchServiceFactory = EnsureArg.IsNotNull(searchServiceFactory, nameof(searchServiceFactory));
+            _resourceIdProvider = EnsureArg.IsNotNull(resourceIdProvider, nameof(resourceIdProvider));
+            _contextAccessor = EnsureArg.IsNotNull(contextAccessor, nameof(contextAccessor));
+            _auditLogger = EnsureArg.IsNotNull(auditLogger, nameof(auditLogger));
+            _logger = EnsureArg.IsNotNull(logger, nameof(logger));
+        }
+
+        public async Task<BulkUpdateResult> UpdateMultipleAsync(string resourceType, string fhirPatchParameters, int parallelThreads, bool readNextPage, uint maximumNumberOfResourcesPerQuery, bool isIncludesRequest, IReadOnlyList<Tuple<string, string>> conditionalParameters, BundleResourceContext bundleResourceContext, CancellationToken cancellationToken)
```
LTA-Thinking: Is the maximumNumberOfResourcesPerQuery parameter used? I see it being passed around to different methods, but I never see it used anywhere.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/BulkUpdateTests.cs
```diff
@@ -492,67 +492,51 @@ public async Task GivenBulkUpdateJobWithMoreThanOnePageOfIncludeResultsWithDiffe
...
+            using HttpResponseMessage response = await SendBulkUpdateRequest(tag, patchRequest, "Patient/$bulk-update", new Dictionary<string, string>() { { "_isParallel", isParallel.ToString() } });
             Assert.Equal(HttpStatusCode.Accepted, response.StatusCode);
 
             BulkUpdateResult expectedResults = new BulkUpdateResult();
             expectedResults.ResourcesUpdated.Add("Patient", 2005);
-            expectedResults.ResourcesUpdated.Add("Group", 1);
             await MonitorBulkUpdateJob(response.Content.Headers.ContentLocation, expectedResults);
 
             // For Ignored resources
             patchRequest = new Parameters()
```
LTA-Thinking: Is this still needed without Group resources?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/BulkUpdateTests.cs
```diff
@@ -492,67 +492,51 @@ public async Task GivenBulkUpdateJobWithMoreThanOnePageOfIncludeResultsWithDiffe
...
-            expectedResultsPatchFailed.ResourcesUpdated.Add("Group", 1);
             await MonitorBulkUpdateJob(responsePatchFailed.Content.Headers.ContentLocation, expectedResultsPatchFailed);
         }
 
         [SkippableTheory]
         [InlineData(true)]
         [InlineData(false)]
         [HttpIntegrationFixtureArgumentSets(DataStore.SqlServer, Format.Json)]
-        public async Task GivenBulkUpdateJobWithMoreThanOnePageOfResultsWithDifferentScenarios_WhenCompleted_ThenResultsAreMatched(bool isParallel)
+        public async Task GivenBulkUpdateJobWithMoreThanOnePageOfIncludeResultsWithDifferentScenarios_WhenCompleted_ThenResultsAreMatched(bool isParallel)
```
LTA-Thinking: This test looks like it has four tests inside of it: normal, ignored, failed, and a more complex include. Should these all be in one test or split up?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/BulkUpdateTests.cs
apurvabhaleMS: No as I am running the bulk-update on Patients. With new split tests it will be more clear

LTA-Thinking (approved):

LTA-Thinking (approved):

PR #5027, opened 2025-06-18
Bulk Delete: Remove references

LTA-Thinking:
## Description
Adds a parameter to bulk-delete to allow for the removal of references to deleted resources.

## Related issues
Addresses [User Story 139078](https://microsofthealth.visualstudio.com/Health/_workitems/edit/139078)

## Testing
New and existing tests

## FHIR Team Checklist

Patch

### DIFF
--- a/docs/BulkImport.md
+++ b/docs/BulkImport.md
@@ -106,6 +106,7 @@ Content-Type:application/fhir+json
 | inputFormat      | String representing the name of the data source format. Currently only FHIR NDJSON files are supported. | 1..1 | ```application/fhir+ndjson``` |
 | mode      | Import mode. | 0..1 | For initial import use ```InitialLoad``` mode value. For incremental import mode use ```IncrementalLoad``` mode value. If no mode value is provided, IncrementalLoad mode value is considered by default. |
 | allowNegativeVersions | Allows FHIR server assigning negative versions for resource records with explicit lastUpdated value and no version specified when input does not fit in contiguous space of positive versions existing in the store. | 0..1 | To enable this feature pass true. By default it is false. |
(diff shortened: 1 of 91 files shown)


src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Delete/ReferenceRemover.cs
```diff
@@ -0,0 +1,38 @@
...
+            var namedChildren = resource.NamedChildren;
+            foreach (var child in namedChildren)
+            {
+                var childValue = child.Value;
+                if (childValue.TypeName == "Reference")
+                {
+                    var reference = (ResourceReference)childValue;
+                    if (reference.Reference.Contains(target, StringComparison.OrdinalIgnoreCase))
+                    {
+                        reference.Reference = null;
```
feordin, line 27: Just double checking that a null value as the reference is valid.  Can we have an empty value for the reference property?

LTA-Thinking: Yes, reference just requires that one of the fields reference, identifier, or display has a value.

feordin: Yes, agreed, the reference property is optional.  By setting it to null will the property no longer be included in the serialized json?  I think it could be an issue if we end up with something like "reference" : ""

LTA-Thinking: No, it is fully removed from the serialized json.

rbans96 (approved):

mikaelweave (approved):

mikaelweave:
Reviewed merge commit - approved on top of last approval.

PR #5030, opened 2025-06-21
[Do not review] Remove lazy

SergeyGaluzo:


src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlStoreClient.cs
```diff
@@ -129,21 +129,18 @@ public async Task<IReadOnlyList<ResourceWrapper>> GetAsync(IReadOnlyList<Resourc
...
+                rawResource = _invisibleResource;
             }
             else
             {
-                rawResource = new Lazy<string>(() =>
-                {
-                    using var rawResourceStream = new MemoryStream(rawResourceBytes);
-                    return decompress(rawResourceStream);
-                });
+                using var rawResourceStream = new MemoryStream(rawResourceBytes);
```
brendankowitz, line 142: Might be interesting to look at RecyclableMemoryStream here

SergeyGaluzo: I don't think there is a problem with it, as entire data reader execution is very fast. We are spending time elsewhere...

PR #5042, opened 2025-07-10
Added ManagedIdentityCredentialOptions to AzureAccessTokenProvider

PTaladay:
## Description
ManagedIdentityCredentialOptions was added to the AzureAccessTokenProvider so that the AuthorityHost can be set depending on what cloudEnvironment environment variable is set to. Defaults to AzurePublicCloud.

## Related issues
Addresses [issue #AB143654].

## Testing
Tested locally

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Operations/AzureAccessTokenProvider.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Operations/AzureAccessTokenProvider.cs
@@ -23,6 +23,10 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations
         {
             EnsureArg.IsNotNull(logger, nameof(logger));
 
+            ManagedIdentityCredentialOptions managedIdentityCredentialOptions = new ManagedIdentityCredentialOptions
+            {
+                AuthorityHost = GetAuthorityHost(),
+            };
             _azureServiceTokenProvider = new ManagedIdentityCredential();
             _logger = logger;
         }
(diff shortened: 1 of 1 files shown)


src/Microsoft.Health.Fhir.Core/Features/Operations/AzureAccessTokenProvider.cs
```diff
@@ -23,6 +23,10 @@ public AzureAccessTokenProvider(ILogger<AzureAccessTokenProvider> logger)
         {
             EnsureArg.IsNotNull(logger, nameof(logger));
 
+            ManagedIdentityCredentialOptions managedIdentityCredentialOptions = new ManagedIdentityCredentialOptions
```
LTA-Thinking, line 26: Should this value be used in the ManagedIdentityCredential created on line 30?

src/Microsoft.Health.Fhir.Core/Features/Operations/AzureAccessTokenProvider.cs
```diff
@@ -61,5 +65,16 @@ public async Task<string> GetAccessTokenForResourceAsync(Uri resourceUri, Cancel
...
         }
+
+        private static Uri GetAuthorityHost()
+        {
+            string cloudEnvironment = Environment.GetEnvironmentVariable("cloudEnvironment");
+
+            return cloudEnvironment switch
+            {
+                "AzureUsGovernment" => AzureAuthorityHosts.AzureGovernment,
+                "AzureCloud" or _ => AzureAuthorityHosts.AzurePublicCloud,
```
LTA-Thinking, line 76: I agree with copilot on this one, I don't think it is necessary to explicitly reference "AzureCloud" and the default case since they go to the same place. I think just having a default case that sets the value to AzurePublicCloud would be more readable.

PTaladay:
Not going to implement into OSS code. Will change to WP codebase.

PR #5050, opened 2025-07-16
 Improve Import Error Handling for Database Constraint Violations

v-shahzad:
## Description
This change introduces a pre-flight validation step within the $import data pipeline to proactively detect and report database constraint violations on a per-resource basis.

## Related issues
Addresses [AB# 151709](https://microsofthealth.visualstudio.com/Health/_workitems/edit/151709)

## Testing
For now it just contains sample change for one constraint to get approval for the approach

## FHIR Team Checklist


src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -693,12 +713,174 @@ async Task MergeUnversioned(List<ImportResource> inputs, bool keepLastUpdated, b
...
+                    var tokenNumberNumberCompositeSearchParamListRowGenerator = new TokenNumberNumberCompositeSearchParamListRowGenerator(_model, tokenSearchParamRowGenerator, new NumberSearchParamListRowGenerator(_model, _searchParameterTypeMap), _searchParameterTypeMap);
+                    var tokenDateTimeCompositeSearchParamListRowGenerator = new TokenDateTimeCompositeSearchParamListRowGenerator(_model, tokenSearchParamRowGenerator, new DateTimeSearchParamListRowGenerator(_model, _searchParameterTypeMap), _searchParameterTypeMap);
+                    var referenceTokenCompositeSearchParamListRowGenerator = new ReferenceTokenCompositeSearchParamListRowGenerator(_model, new ReferenceSearchParamListRowGenerator(_model, _searchParameterTypeMap), tokenSearchParamRowGenerator, _searchParameterTypeMap);
+
+                    // Traverse through the resources to check which resources are causing constraint violation
+                    foreach (var resource in resources)
```
feordin: What kind of impact do we think this will have on $import performance?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -705,12 +725,174 @@ async Task MergeUnversioned(List<ImportResource> inputs, bool keepLastUpdated, b
...
```
SergeyGaluzo: If I read it correctly this code will call Merge for each of input resources. This will have devastating effect on performance.
I agree that it is better provide info on resource level, but not with negative affect on perf.
We should either do "constraint" checks in C#, or just say that batch staring with this resource and ending with this failed with constraint violation, raise bad request, and let customer figure it out.

SergeyGaluzo (changes requested):
Adding checks before calling merge adds extra load on app compute for each normal merge call. I suggest running this check only if merge fails with constraint violation. This way there should be **zero** effect on performance for case when there are not constraint errors.

I do not like having any code duplicated (C# constraint code and the database). Can we get away with reporting some details of SQL error and attributing it to the resource lines in the input file for the batch we processed? This might not be that precise, but we will not have any code duplication.

We also should abort import on first constraint error and should not spend app compute on analyzing all the input resources. 

Please add a test for this functionality.

v-shahzad:
Hi @SergeyGaluzo , Thanks for the detailed suggestions. I’ll start implementing these changes accordingly.

SergeyGaluzo:
> Hi @SergeyGaluzo , Thanks for the detailed suggestions. I’ll start implementing these changes accordingly.

@v-shahzad I would consider "no code duplication" approach if possible. This sounds very attractive,

SergeyGaluzo:
@v-shahzad Please remove any constraint validation from C#.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -554,11 +554,15 @@ List<string> GetErrors(IReadOnlyCollection<ImportResource> dups, IReadOnlyCollec
                 }
                 catch (SqlException ex) when (ex.Number == SqlStoreErrorCodes.ConstraintViolation && !ex.IsRetriable())
                 {
-                    // Post-validation: identify all resources that violate length/overflow constraints.
-                    var postCheck = ValidateConstraintViolations(resources);
-
-                    // Any resource flagged gets reported as conflict; we DO NOT attempt partial re-merge.
-                    conflicts.AddRange(postCheck.Conflicts.Where(c => !conflicts.Contains(c)));
+                    // Put a generic message that this resource was in the batch where one or more resource failed due to constraint violation
```
SergeyGaluzo: Error message should contain some details from SQL. See below example
CREATE TABLE SG_Tbl (Id int NOT NULL PRIMARY KEY, Data1 varchar(100) CHECK (datalength(Data1) > 3), Data2 varchar(100) NOT NULL)

INSERT INTO SG_Tbl SELECT 1, '11', '11'
The INSERT statement conflicted with the CHECK constraint "CK__SG_Tbl__Data1__3726238F". The conflict occurred in database "HS_16PS_PSMT_OFF_DCRS_OFF_Small_Allocations_MoreCompression", table "dbo.SG_Tbl", column 'Data1'.
We should look at the definition of check constraint in system catalog and provide it to the user
Message should look like "Error on batch with offset xxx rows from YYY to ZZZ. INSERT statement conflicted with the CHECK constraint datalength(Data1) > 3 column Data1"

INSERT INTO SG_Tbl SELECT 1, '1111', NULL
Cannot insert the value NULL into column 'Data2', table 'HS_16PS_PSMT_OFF_DCRS_OFF_Small_Allocations_MoreCompression.dbo.SG_Tbl'; column does not allow nulls. INSERT fails.
Message should look like "Error on batch with offset xxx rows from YYY to ZZZ. Cannot insert the value NULL into column 'Data2'; column does not allow nulls"

INSERT INTO SG_Tbl SELECT 1, '1111', ''
INSERT INTO SG_Tbl SELECT 1, '1111', ''
Violation of PRIMARY KEY constraint 'PK__SG_Tbl__3214EC07F5A7D2C8'. Cannot insert duplicate key in object 'dbo.SG_Tbl'. The duplicate key value is (1).
Message should look like "Error on batch with offset xxx rows from YYY to ZZZ. Violation of PRIMARY KEY constraint (Id). The duplicate key value is (1)"

SergeyGaluzo (changes requested):

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
v-shahzad: @SergeyGaluzo  Thank you for the suggestions. I will implement the required changes and update the PR for review.

v-shahzad:
Hi @SergeyGaluzo , I have added the constraint details to the first resource of the batch with the format 

Error on batch with offset 0 rows from 1 to 3. Insert failed due to the constraint 'CHK_TokenSearchParam_CodeOverflow' on 'TokenSearchParam'. Details: (len([Code])=(256) OR [CodeOverflow] IS NULL).  

Please suggest me if it matches your expectation or if you would like any changes.
 Thank you

SergeyGaluzo:
> Hi @SergeyGaluzo , I have added the constraint details to the first resource of the batch with the format
> 
> Error on batch with offset 0 rows from 1 to 3. Insert failed due to the constraint 'CHK_TokenSearchParam_CodeOverflow' on 'TokenSearchParam'. Details: (len([Code])=(256) OR [CodeOverflow] IS NULL).
> 
> Please suggest me if it matches your expectation or if you would like any changes. Thank you

@v-shahzad  This is much better. 
1. The problem I see in adding to the first resource in a batch is a potential confusion for the customer. They might think that error is on that specific resource. Do we need to add any resource level errors at all? Can't we just raise exception and not record errors in the import error log?
2. We should avoid putting constraint and table names into the error message. In this case we know that it is a search parameter of type token, so something like this "Insert failed due to the constraint violation for token search parameter. Details: (len(Code]=(256) OR CodeOverflow IS NULL)". You can remove [] to simplify reading.

v-shahzad:
> > Hi @SergeyGaluzo , I have added the constraint details to the first resource of the batch with the format
> > Error on batch with offset 0 rows from 1 to 3. Insert failed due to the constraint 'CHK_TokenSearchParam_CodeOverflow' on 'TokenSearchParam'. Details: (len([Code])=(256) OR [CodeOverflow] IS NULL).
> > Please suggest me if it matches your expectation or if you would like any changes. Thank you
> 
> @v-shahzad This is much better.
> 
> 1. The problem I see in adding to the first resource in a batch is a potential confusion for the customer. They might think that error is on that specific resource. Do we need to add any resource level errors at all? Can't we just raise exception and not record errors in the import error log?
> 2. We should avoid putting constraint and table names into the error message. In this case we know that it is a search parameter of type token, so something like this "Insert failed due to the constraint violation for token search parameter. Details: (len(Code]=(256) OR CodeOverflow IS NULL)". You can remove [] to simplify reading.

Thanks for the feedback, @SergeyGaluzo . I’ve made the suggested changes. Please let me know if any further updates are needed.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -479,68 +485,86 @@ List<string> GetErrors(IReadOnlyCollection<ImportResource> dups, IReadOnlyCollec
...
-                if (importMode == ImportMode.InitialLoad)
-                {
-                    var inputsDedupped = resources.GroupBy(_ => _.ResourceWrapper.ToResourceKey(true)).Select(_ => _.OrderBy(_ => _.ResourceWrapper.LastModified).First()).ToList();
-                    var current = new HashSet<ResourceKey>((await GetAsync(inputsDedupped.Select(_ => _.ResourceWrapper.ToResourceKey(true)).ToList(), cancellationToken)).Select(_ => _.ToResourceKey(true)));
-                    loaded.AddRange(inputsDedupped.Where(i => !current.TryGetValue(i.ResourceWrapper.ToResourceKey(true), out _)));
-                    await Merge(loaded, false, useReplicasForReads);
-                }
-                else if (importMode == ImportMode.IncrementalLoad)
+
+                // Adding try-catch to chekc if Contraint violation erro is happening here
```
SergeyGaluzo: add r in erro**r**

src/Microsoft.Health.Fhir.SqlServer/Resources.resx
```diff
@@ -26,36 +26,36 @@
...
+
     mimetype: application/x-microsoft.net.object.soap.base64
-    value   : The object must be serialized with 
+    value   : The object must be serialized with
             : System.Runtime.Serialization.Formatters.Soap.SoapFormatter
             : and then encoded with base64 encoding.
 
     mimetype: application/x-microsoft.net.object.bytearray.base64
-    value   : The object must be serialized into a byte array 
+    value   : The object must be serialized into a byte array
```
SergeyGaluzo: Please make sure that only added lines are marked as changed in diff.

SergeyGaluzo:
Please add import tests for all expected types of constraint violations.

SergeyGaluzo:
We do not plan to have foreign keys, please remove this section from SQL query.

SergeyGaluzo:
Please evaluate whether violation of PK/Unique constraints are possible in import. If not, remove corresponding section from SQL query.

v-shahzad:
Hi @SergeyGaluzo , I’ve removed the checks for primary key and foreign key constraints since we’re only catching exceptions related to overflow constraints. I’ll create a separate user story and PR for the test cases.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -469,7 +471,11 @@ List<string> GetErrors(IReadOnlyCollection<ImportResource> dups, IReadOnlyCollec
 
                 foreach (var resource in conflicts)
```
feordin, line 472: We discussed as a team, and want to keep this simple.  If SQL reports an exception during import.  We capture the exception and associated with the entire batch that had an issue.  So instead of checking each resource, and attempting to identify the specific resource, we'll let the logic remain in SQL and simplify the C# code.  It will be a more generic error the user, but gives us less complexity for what is an error case.

rbans96:
Closing this PR as its 342 days old, branch is not cleaned up.

PR #5052, opened 2025-07-18
Add architectural doc for duplicated clinical notes

feordin:
## Description
Add architectural notes/design doc for duplicated clinical notes feature.

## Related issues
Addresses [issue [AB#158013](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/158013)].

## Testing

## FHIR Team Checklist

### DIFF
--- a/docs/arch/adr-fhir-resource-duplication-design.md
+++ b/docs/arch/adr-fhir-resource-duplication-design.md
@@ -131,10 +131,42 @@ This approach intercepts search operations and dynamically generates resources o
 
 #### Flow
 1. **Search Interception**: When searching for DocumentReference or DiagnosticReport:
-   - Execute the original search query
-   - Execute a parallel search for the other resource type with attachment URLs
-   - Transform found resources of the other type into the requested type
-   - Merge results and return unified response
+   - When a search query comes in, it will be analyzed for any DiagnosticReport or DocumentReference resource type to be included in the results.  If only one of those types is being queried, the other type will be added
+    - When the other type is added it will be added with the same search criteria, and in addition, if not already present the new type will also have two more search criteria
+      - Must have a patient reference
+      - Must have a reference to a url
+    - Once the search results come back, the resources in the results will be examined.  Any items in the result set which are of the "other" type will be removed from the search results, and replaced with a dynamically generated resource of the primary type.
+      - This dynamically generated resource will not have an ID
+      - This dynamically generated resource will have a source property populated with a link to resource that was used to generate it.
+
+#### Example:
+
+```mermaid
+sequenceDiagram
+    participant Client
+    participant FHIR_Service
+    participant Storage
+
+    Note over Client,Storage: Search Interception Flow for DocumentReference Query
+
+    Client->>FHIR_Service: GET DocumentReference?patient=123
+    
+    Note over FHIR_Service: Analyze search query
+    Note over FHIR_Service: Rewrite query to include both DocumentReference<br/>and DiagnosticReport with patient reference<br/>and url constraints
+    
+    FHIR_Service->>Storage: Query (DocumentReference?patient=123) OR<br/>(DiagnosticReport?patient=123&presentedForm.url:exists=true)
+    Storage-->>FHIR_Service: Combined result set with both resource types
+    
+    loop For each DiagnosticReport in results
+        Note over FHIR_Service: Remove DiagnosticReport from result set
+        Note over FHIR_Service: Generate DocumentReference from DiagnosticReport
+        Note over FHIR_Service: Add generated DocumentReference to result set
+    end
+    
+    Note over FHIR_Service: Apply paging and count limits to final result set
+    
+    FHIR_Service-->>Client: Bundle with DocumentReference resources<br/>(original + generated from DiagnosticReport)
+```
 
 2. **Resource Operations**: Normal create/update/delete operations (no special logic)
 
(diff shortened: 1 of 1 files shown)


rajithaalurims:
@feordin I have couple of questions on the ask:

**Question 1:**
We have 3 DocumentReeference resources for a Patient 123. 2 of those are Clinical notes and 1 is just Patient's insurance card.
We have 3 DiagosticReport resources for a Patient 123

For this query Get /DocumentReference?patient=123 - We should return all 6 resources, correct?
For this query Get /DiagosticReport?patient=123 - We should return all 3 DRs and only 2 DocumentReference that are of type Clinical Notes. Is this correct?

**Question 2:**
Do we always search for both Diagnostic Report and Document Reference for all search criteria or do we need to support this only when these are queried for a Patient?
For ex, does this query search for DiagnosticReport resources associated with that Encounter?
GET [base]/DocumentReference?encounter=Encounter/67890

**Question 3:**
Do we have to convert DiagnosticReports --> DocumentReference or DocumentReference --> DiagnosticReport in the search results based on what customer is queried for?
My understanding is we can return different resource types in a bundle.  We do not have to convert resource types.

**Question 4:**
Can you share details of how these 2 queries will change?
GET [base]/DiagnosticReport?patient=123&_revinclude=*
GET [base]/DiagnosticReport?patient=123&_include=*

**Question 4:**
In the second sequence diagram, for Get /DocumentReference query, I see you documented expected query as "Query (DocumentReference?patient=123) OR
(DiagnosticReport?patient=123&presentedForm.url:exists=true)"
What will be the logic when customer queries for DiagnosticReport?

feordin:
> @feordin I have couple of questions on the ask:
> 
> **Question 1:** We have 3 DocumentReeference resources for a Patient 123. 2 of those are Clinical notes and 1 is just Patient's insurance card. We have 3 DiagosticReport resources for a Patient 123
> 
> For this query Get /DocumentReference?patient=123 - We should return all 6 resources, correct? For this query Get /DiagosticReport?patient=123 - We should return all 3 DRs and only 2 DocumentReference that are of type Clinical Notes. Is this correct?
> 
> **Question 2:** Do we always search for both Diagnostic Report and Document Reference for all search criteria or do we need to support this only when these are queried for a Patient? For ex, does this query search for DiagnosticReport resources associated with that Encounter? GET [base]/DocumentReference?encounter=Encounter/67890
> 
> **Question 3:** Do we have to convert DiagnosticReports --> DocumentReference or DocumentReference --> DiagnosticReport in the search results based on what customer is queried for? My understanding is we can return different resource types in a bundle. We do not have to convert resource types.
> 
> **Question 4:** Can you share details of how these 2 queries will change? GET [base]/DiagnosticReport?patient=123&_revinclude=* GET [base]/DiagnosticReport?patient=123&_include=*
> 
> **Question 4:** In the second sequence diagram, for Get /DocumentReference query, I see you documented expected query as "Query (DocumentReference?patient=123) OR (DiagnosticReport?patient=123&presentedForm.url:exists=true)" What will be the logic when customer queries for DiagnosticReport?

Excellent questions!

1) It is true that the spec indicates that we need to return clinical notes and ignore other types of references.  In your example we ignore the insurance card.  I'm not sure we can reliably implement that part of the requirement. It is dependent on our server having detailed knowledge of the attachments.  We may be able to use SNOMED codes on the attachment, if present, to filter the results, but I think this should be a nice to have.  In your example where we have 3 DocRefs, and 3 DiagnosticReports and the query is for DocumentReference, Yes, we return 6 items.  They will all be DocumentReference resources.  We will search for the DiagnosticReports (and assuming they meet the search criteria) they will initially be included in the search results.  Then we will iterate over the search results, removing the DiagnosticReports and replacing them with generated DocumentReference resources which contain the same URL reference.

2) Yes, we only generate new resources when there is a patient reference.  This will need to be a part of the search criteria any time we search for the "other" resource type.  Another criteria we will add is that the resource contains a URL reference to an external document.  Those two criteria will be in addition to whatever the user specifies in the query string.

3) We have to convert the resource types.  While FHIR allows us to return multiple resource types in a bundle, the user asked for DocRefs, then the user gets DocRefs.  If the user asks for DiagReport, then they get DiagReport.

4) Yes, the diagram was an example, you could switch DocumentReference and DiagnosticReport for the opposite scenario.

evachen96:
@feordin , option 2 mentions the performance impact on search, do you foresee performance impact on option 1 during create/update event?

v-isyamauchi-gh:
For approach 2, how do we handle _include/_revinclude in search? Also, search with the wild character such as "_include=*"?

v-isyamauchi-gh:
How do we ensure the page size in the search result be honored as we create counter resources on the fly and add them to the search result?

v-isyamauchi-gh:
Assuming search operations being more frequently executed operations, how would approach 2 (additional pre- and post-search processes) impact the overall customer experience as opposed to approach 1 taking additional store space for counter resources?

v-isyamauchi-gh:
There are some mandatory fields that must be specified in both resource types. How do we determine which values to use when creating a counter source? Also, how can we make counter resources more searchable maybe using some fields?

https://www.hl7.org/fhir/R4/documentreference.html
https://www.hl7.org/fhir/R4/diagnosticreport.html

v-isyamauchi-gh:
The Hl7 doc says the following. How can we recognize which one to duplicate?

"Not all scanned information stored through DocumentReference will be exposed through DiagnosticReport since DocumentReference stores other non-clinical information. For example, DocumentReference can point to an insurance card."

https://hl7.org/fhir/us/core/STU6.1/clinical-notes.html#fhir-resources-to-exchange-clinical-notes

v-isyamauchi-gh:
How do we handle a request on a duplicate resource in approach 2? For example, a customer wants to update, delete, or search again a duplicate resource returned in previous search?

feordin:
> How do we ensure the page size in the search result be honored as we create counter resources on the fly and add them to the search result?

I think we do this by updating the search expression tree in the search service and add the query for the "other" resource type.  The search for both resource types will then be included in the page count, and the continuation tokens.

feordin:
> How do we handle a request on a duplicate resource in approach 2? For example, a customer wants to update, delete, or search again a duplicate resource returned in previous search?

When we return a "generated" resource, it will not have a valid ID.  So a customer cannot update or delete it.

feordin:
The previous questions does bring up a thought of my own.  If we generate and store "duplicates" following option 1.  What happens when a customer deletes the "duplicate" resource?  Not the original?

Do we delete the original?  If not, if we preserve the original.  Later when the customer runs a query, the "duplicate" will not be there, and we won't return it in the search, so we won't meet the requirement.  Is that acceptable?

feordin:
> For approach 2, how do we handle _include/_revinclude in search? Also, search with the wild character such as "_include=*"?

Any time we _include or _revinclude either DiagReport or DocRef, we would include the other.

PR #5054, opened 2025-07-22
Use Debug Logging

LTA-Thinking:
## Description
This PR introduces a few new logs into validation and switches some existing logs to use Debug instead of Infromation.

The new logs are to help with diagnosing the issue in the story linked below.
The switch to Debug logging is to address the overabundance of logs we have right now. The logs selected for demotion to Debug are primarily ones that occur automatically without any user input and don't report any changes to the service state. These are mostly background jobs saying they have done nothing because there is no work for them at the moment. These can be useful to show the background process is still running, but most of the time they are just clutter and we have other logs from startup and shutdown of background jobs that will show whether they are running.
Some others selected for demotion are SQL logs that output a log of metadata on calls that is not helpful most of the time. These have complimenting logs that send the metadata at a higher logging level if they are outside of normal bounds.

This has already helped me with local testing as changing my logging level to "Information" in appsettings.Development.json made it so only logs related to the requests I was making were sent. Previously the terminal was filled by logs for background tasks that weren't doing anything.

## Related issues
Addresses [User Story 154657](https://microsofthealth.visualstudio.com/Health/_workitems/edit/154657)

## Testing
Checked local logs.

## FHIR Team Checklist

Patch


src/Microsoft.Health.Fhir.SqlServer/Features/Watchdogs/Watchdog.cs
```diff
@@ -77,9 +79,19 @@ private async Task OnNextTickAsync(CancellationToken cancellationToken)
...
-            using (_logger.BeginTimedScope($"{Name}.OnNextTickAsync"))
+            var stopwatch = new Stopwatch();
+            stopwatch.Start();
+
+            await RunWorkAsync(cancellationToken);
+
+            if (DateTime.UtcNow - _lastLog > TimeSpan.FromHours(1))
             {
-                await RunWorkAsync(cancellationToken);
+                _lastLog = DateTime.Now;
```
jestradaMS: This should be DateTime.UtcNow for consistency.

src/Microsoft.Health.Fhir.SqlServer/Features/Watchdogs/WatchdogLease.cs
```diff
@@ -103,7 +103,7 @@ protected async Task OnNextTickAsync(CancellationToken cancellationToken)
                 _leaseEndTime = isAcquired ? leaseEndTime : _leaseEndTime;
             }
 
-            _logger.LogInformation($"WatchdogLease.RunAsync: Completed acquire: resource=[{_watchdogName}] worker=[{_worker}] period={PeriodSec} timeout={_leaseTimeoutSec} leaseEndTime=[{leaseEndTime:s}] isAcquired={isAcquired} currentHolder=[{currentHolder}].");
+            _logger.LogDebug($"WatchdogLease.RunAsync: Completed acquire: resource=[{_watchdogName}] worker=[{_worker}] period={PeriodSec} timeout={_leaseTimeoutSec} leaseEndTime=[{leaseEndTime:s}] isAcquired={isAcquired} currentHolder=[{currentHolder}].");
```
jestradaMS, line 106: Should we log info if  "isAcquired=true"?

LTA-Thinking: That would still create a lot of logs. Most of the watchdogs have leases that last an hour, but Transaction watchdog's only lasts 20 seconds.

jestradaMS (approved):
Looks good! Would be helpful to add a doc around how to enable debug for an account if we don't already have one.

PR #5056, opened 2025-07-25
Index all token values, regardless of casing

LTA-Thinking:
## Description
If token search field has an array of values and two or more of those values differ only in casing, then only the first value in the array will be stored as a searchable parameter. This change fixes that so that all the values in the array are stored and searchable, regardless of casing.

## Related issues
Addresses [Bug 160947](https://microsofthealth.visualstudio.com/Health/_workitems/edit/160947)

## Testing
A new test was added to cover this scenario.

## FHIR Team Checklist

Patch

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/TokenSearchValue.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/TokenSearchValue.cs
@@ -107,9 +107,9 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.SearchValues
                 return false;
             }
 
-            return string.Equals(System, tokenSearchValueOther.System, StringComparison.OrdinalIgnoreCase) &&
-                string.Equals(Code, tokenSearchValueOther.Code, StringComparison.OrdinalIgnoreCase) &&
-                string.Equals(Text, tokenSearchValueOther.Text, StringComparison.OrdinalIgnoreCase);
+            return string.Equals(System, tokenSearchValueOther.System, StringComparison.Ordinal) &&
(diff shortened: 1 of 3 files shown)


src/Microsoft.Health.Fhir.Core/Features/Search/SearchValues/TokenSearchValue.cs
```diff
@@ -107,9 +107,9 @@ public bool Equals([AllowNull] ISearchValue other)
                 return false;
             }
 
-            return string.Equals(System, tokenSearchValueOther.System, StringComparison.OrdinalIgnoreCase) &&
-                string.Equals(Code, tokenSearchValueOther.Code, StringComparison.OrdinalIgnoreCase) &&
-                string.Equals(Text, tokenSearchValueOther.Text, StringComparison.OrdinalIgnoreCase);
+            return string.Equals(System, tokenSearchValueOther.System, StringComparison.Ordinal) &&
+                string.Equals(Code, tokenSearchValueOther.Code, StringComparison.Ordinal) &&
+                string.Equals(Text, tokenSearchValueOther.Text, StringComparison.Ordinal);
```
brendankowitz, line 115: Maybe this needs a comment for future us 😃 along the lines of "using string comparison to match database collation". Does this have any impact on Cosmos? Might be worth a quick check

LTA-Thinking: The new E2E test runs on both DBs. 
Yes, this bug is also in Cosmos. 
The search parameters before the change:
image

and after the change:
image

brendankowitz (approved):

PR #5075, opened 2025-08-04
Handle version 0 on racing conditions

LTA-Thinking:
## Description
When importing a large amount of data with the same last updated time an error can happen when two versions are created with version 0. This causes a key conflict when the resources are sent to the merge method.
The root of this is a check where if a resource is made at the same time as an existing version of the same resource it is given version 0. This should be marked as a conflict, but instead it is added to the list of resources to add if allowNegativeVersions is set. If two versions of the same resource with the same last updated are processed in the same subjob then the key conflict can happen.

## Related issues
Addresses [Bug 165395](https://microsofthealth.visualstudio.com/Health/_workitems/edit/165395)

## Testing
Manual

## FHIR Team Checklist

Patch


SergeyGaluzo (approved):

SergeyGaluzo:
There is line 
   conflicts.Add(input); // no version slot available and negative versions are not allowed
Its comment will become not accurate. Please change to 
// (no version slot available and negative versions are not allowed) or a conflict (returned version from thew database is zero)

SergeyGaluzo (dismissed):

fhibf (approved):

PR #5097, opened 2025-08-15
Implemented App and DB level concurrency management for Search parameter upates

jestradaMS:
## Description
Resolves critical race conditions in SearchParameter operations that occurred when multiple concurrent requests attempted to create, update, or delete the same SearchParameter resource simultaneously. This implementation introduces a robust two-layer concurrency control strategy combining application-level pessimistic locking with database-level optimistic concurrency.

## Related issues
Addresses race condition issues where concurrent SearchParameter updates would fail unpredictably, particularly in high-load environments with multiple FHIR server instances.

AB#164155

## Testing

## FHIR Team Checklist

Feature - New concurrency management capabilities with backward compatibility

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/SearchParameterOperations.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/SearchParameterOperations.cs
@@ -60,64 +60,73 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Parameters
 
         public async Task AddSearchParameterAsync(ITypedElement searchParam, CancellationToken cancellationToken)
         {
-            try
-            {
-                // We need to make sure we have the latest search parameters before trying to add
-                // a search parameter. This is to avoid creating a duplicate search parameter that
-                // was recently added and that hasn't propogated to all fhir-server instances.
-                await GetAndApplySearchParameterUpdates(cancellationToken);
+            var searchParameterWrapper = new SearchParameterWrapper(searchParam);
+            var searchParameterUrl = searchParameterWrapper.Url;
 
-                // verify the parameter is supported before continuing
-                var searchParameterWrapper = new SearchParameterWrapper(searchParam);
-                var searchParameterInfo = new SearchParameterInfo(searchParameterWrapper);
-
-                if (searchParameterInfo.Component?.Any() == true)
+            await SearchParameterConcurrencyManager.ExecuteWithLockAsync(
+                searchParameterUrl,
+                async () =>
                 {
-                    foreach (SearchParameterComponentInfo c in searchParameterInfo.Component)
+                    try
                     {
-                        c.ResolvedSearchParameter = _searchParameterDefinitionManager.GetSearchParameter(c.DefinitionUrl.OriginalString);
+                        // We need to make sure we have the latest search parameters before trying to add
+                        // a search parameter. This is to avoid creating a duplicate search parameter that
+                        // was recently added and that hasn't propogated to all fhir-server instances.
+                        await GetAndApplySearchParameterUpdates(cancellationToken);
+
+                        // verify the parameter is supported before continuing
+                        var searchParameterInfo = new SearchParameterInfo(searchParameterWrapper);
+
+                        if (searchParameterInfo.Component?.Any() == true)
+                        {
+                            foreach (SearchParameterComponentInfo c in searchParameterInfo.Component)
+                            {
+                                c.ResolvedSearchParameter = _searchParameterDefinitionManager.GetSearchParameter(c.DefinitionUrl.OriginalString);
+                            }
+                        }
+
+                        (bool Supported, bool IsPartiallySupported) supportedResult = _searchParameterSupportResolver.IsSearchParameterSupported(searchParameterInfo);
+
+                        if (!supportedResult.Supported)
+                        {
+                            throw new SearchParameterNotSupportedException(string.Format(Core.Resources.NoConverterForSearchParamType, searchParameterInfo.Type, searchParameterInfo.Expression));
+                        }
+
+                        // check data store specific support for SearchParameter
+                        if (!_dataStoreSearchParameterValidator.ValidateSearchParameter(searchParameterInfo, out var errorMessage))
+                        {
+                            throw new SearchParameterNotSupportedException(errorMessage);
+                        }
+
+                        _logger.LogInformation("Adding the search parameter '{Url}'", searchParameterWrapper.Url);
+                        _searchParameterDefinitionManager.AddNewSearchParameters(new List<ITypedElement> { searchParam });
+
+                        await _searchParameterStatusManager.AddSearchParameterStatusAsync(new List<string> { searchParameterWrapper.Url }, cancellationToken);
                     }
-                }
+                    catch (FhirException fex)
+                    {
+                        _logger.LogError(fex, "Error adding search parameter.");
+                        fex.Issues.Add(new OperationOutcomeIssue(
+                            OperationOutcomeConstants.IssueSeverity.Error,
+                            OperationOutcomeConstants.IssueType.Exception,
+                            Core.Resources.CustomSearchCreateError));
 
-                (bool Supported, bool IsPartiallySupported) supportedResult = _searchParameterSupportResolver.IsSearchParameterSupported(searchParameterInfo);
+                        throw;
+                    }
+                    catch (Exception ex)
+                    {
+                        _logger.LogError(ex, "Error adding search parameter.");
+                        var customSearchException = new ConfigureCustomSearchException(Core.Resources.CustomSearchCreateError);
+                        customSearchException.Issues.Add(new OperationOutcomeIssue(
+                            OperationOutcomeConstants.IssueSeverity.Error,
+                            OperationOutcomeConstants.IssueType.Exception,
+                            ex.Message));
 
-                if (!supportedResult.Supported)
-                {
-                    throw new SearchParameterNotSupportedException(string.Format(Core.Resources.NoConverterForSearchParamType, searchParameterInfo.Type, searchParameterInfo.Expression));
-                }
-
-                // check data store specific support for SearchParameter
-                if (!_dataStoreSearchParameterValidator.ValidateSearchParameter(searchParameterInfo, out var errorMessage))
-                {
-                    throw new SearchParameterNotSupportedException(errorMessage);
-                }
-
-                _logger.LogInformation("Adding the search parameter '{Url}'", searchParameterWrapper.Url);
-                _searchParameterDefinitionManager.AddNewSearchParameters(new List<ITypedElement> { searchParam });
-
-                await _searchParameterStatusManager.AddSearchParameterStatusAsync(new List<string> { searchParameterWrapper.Url }, cancellationToken);
-            }
-            catch (FhirException fex)
-            {
-                _logger.LogError(fex, "Error adding search parameter.");
-                fex.Issues.Add(new OperationOutcomeIssue(
-                    OperationOutcomeConstants.IssueSeverity.Error,
-                    OperationOutcomeConstants.IssueType.Exception,
-                    Core.Resources.CustomSearchCreateError));
-
-                throw;
-            }
-            catch (Exception ex)
-            {
-                _logger.LogError(ex, "Error adding search parameter.");
-                var customSearchException = new ConfigureCustomSearchException(Core.Resources.CustomSearchCreateError);
-                customSearchException.Issues.Add(new OperationOutcomeIssue(
-                    OperationOutcomeConstants.IssueSeverity.Error,
-                    OperationOutcomeConstants.IssueType.Exception,
-                    ex.Message));
-
-                throw customSearchException;
-            }
+                        throw customSearchException;
+                    }
+                },
+                _logger,
+                cancellationToken);
         }
 
         /// <summary>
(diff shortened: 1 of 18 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Tables/SearchParam.sql
```diff
@@ -7,5 +7,6 @@
     WITH (DATA_COMPRESSION = PAGE),
     Status                  varchar(20)                 NULL,
     LastUpdated             datetimeoffset(7)           NULL,
-    IsPartiallySupported    bit                         NULL
+    IsPartiallySupported    bit                         NULL,
+    RowVersion              rowversion                  NOT NULL
```
SergeyGaluzo: datetimeoffset(7) has 0.1 microsecond precision. Why it cannot be used to concurrency control?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/95.diff.sql
```diff
@@ -0,0 +1,78 @@
+/*************************************************************
```
SergeyGaluzo, line 1: Is there a reason for this file to be here?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Types/Types.sql
```diff
@@ -5,6 +5,14 @@
     IsPartiallySupported bit NOT NULL
 )
 
+CREATE TYPE dbo.SearchParamTableType_3 AS TABLE
```
SergeyGaluzo: Please change name to SearchParamList

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/UpsertSearchParams.sql
```diff
@@ -16,17 +17,43 @@
...
+DECLARE @conflictedRows TABLE (
+    Uri VARCHAR(128) COLLATE Latin1_General_100_CS_AS NOT NULL
+);
+
+-- Check for concurrency conflicts first using LastUpdated
+INSERT INTO @conflictedRows (Uri)
+SELECT sp.Uri 
+FROM @searchParams sp
+INNER JOIN dbo.SearchParam existing ON sp.Uri = existing.Uri
+WHERE sp.LastUpdated IS NOT NULL
```
SergeyGaluzo: Current SearchParam table definition is incorrect. There should not be any reason for having NULL value in LastUpdated. Please change table definition, corresponding types and code.

Same goes for status and "support" flag

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/UpsertSearchParams.sql
```diff
@@ -16,17 +17,43 @@
...
+    SELECT @conflictMessage = CONCAT('Optimistic concurrency conflict detected for search parameters: ', 
+                                   STRING_AGG(Uri, ', '))
+    FROM @conflictedRows;
+    
+    ROLLBACK TRANSACTION;
+    THROW 50001, @conflictMessage, 1;
+END
+
 -- Acquire and hold an exclusive table lock for the entire transaction to prevent parameters from being added or modified during upsert.
 MERGE INTO dbo.SearchParam WITH (TABLOCKX)
```
SergeyGaluzo: Transaction is serializable. Is TABLOCKX required?

src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterConcurrencyException.cs
```diff
@@ -0,0 +1,40 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.Health.Fhir.Core.Features.Search.Registry
+{
+    /// <summary>
+    /// Exception thrown when an optimistic concurrency conflict occurs during search parameter updates.
+    /// </summary>
+    public class SearchParameterConcurrencyException : Exception
```
LTA-Thinking, line 15: Is this exception used anywhere? I don't see any references to it in this PR.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
```diff
@@ -397,16 +397,23 @@ FROM dbo.SearchParam INNER JOIN @searchParamStatuses as sps
...
+                    ? "dbo.SearchParamTableType_3"
+                    : "dbo.SearchParamTableType_2";
+
                 var tableValuedParameter = new SqlParameter
                     {
                         ParameterName = "searchParamStatuses",
                         SqlDbType = SqlDbType.Structured,
                         Value = collection,
                         Direction = ParameterDirection.Input,
-                        TypeName = "dbo.SearchParamTableType_2",
```
LTA-Thinking, line 409: I think this will be the problem with backwards compatibility. The old code will pass in a type that the new stored procedure isn't expecting as an input.

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlServerFhirStorageTestsFixture.cs
```diff
@@ -353,11 +354,6 @@ object IServiceProvider.GetService(Type serviceType)
                 return _testHelper;
             }
 
-            if (serviceType.IsInstanceOfType(this))
```
LTA-Thinking, line 357: Why remove this? It doesn't look useful at all, but I'm curious why it caught your attention.

jestradaMS: Had GitHub copilot refactor and add tests and it made this change. When I questioned it, I got the below, which seems reasonable after reviewing it as well.

Service Locator Anti-Pattern: The IsInstanceOfType(this)check essentially makes the fixture return itself when asked for services it implements. This is considered an anti-pattern because it creates tight coupling and makes the dependency injection container return the container itself rather than properly registered services.

Explicit Service Registration: Instead of relying on a generic "return this if I implement the interface" approach, the code now explicitly checks for each service type and returns the appropriate instance. This makes the dependencies more explicit and easier to understand.

Better Maintainability: By explicitly listing each service type the fixture can provide, it becomes clearer what services are available and makes the code more maintainable and less prone to unexpected behavior.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Tables/SearchParam.sql
jestradaMS: Modified to use LastUpdated per this review.

src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterConcurrencyException.cs
jestradaMS: Used in [SqlServerSearchParameterStratusDataStore](https://github.com/microsoft/fhir-server/pull/5097/files#diff-a4d7d40ccbe52c3d7812308c5383a1d981e829c471de16c83cb94b0b0195c6c6R210)

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/UpsertSearchParams.sql
jestradaMS: Moved up above initial selects as we discussed

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/UpsertSearchParams.sql
jestradaMS: Removed is not null condition. Did not see anything regarding IsPartiallySupported. Please comment on line # if I missed something.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/95.diff.sql
jestradaMS: fixed white space diff in 94.sql as we discussed. This comment was made on wrong file.

src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterConcurrencyManager.cs
```diff
@@ -0,0 +1,148 @@
...
+            logger?.LogDebug("Acquiring lock for search parameter: {SearchParameterUri}", searchParameterUri);
+
+            await semaphore.WaitAsync(cancellationToken);
+
+            try
+            {
+                logger?.LogDebug("Lock acquired for search parameter: {SearchParameterUri}", searchParameterUri);
+                return await function();
+            }
+            finally
```
LTA-Thinking, line 62: This cleanup logic looks correct to me, but I would not be surprised if there is another edge case I'm not seeing.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SqlServerSearchParameterStatusDataStore.cs
```diff
@@ -155,15 +161,59 @@ public async Task UpsertStatuses(IReadOnlyCollection<ResourceSearchParameterStat
...
+                    var refreshedStatuses = await GetSearchParameterStatuses(cancellationToken);
+                    var refreshedDict = refreshedStatuses.ToDictionary(s => s.Uri.OriginalString, s => s);
+
+                    // Update our statuses with fresh LastUpdated values
+                    foreach (var status in currentStatuses)
                     {
-                        status.Status = SearchParameterStatus.Disabled;
+                        if (refreshedDict.TryGetValue(status.Uri.OriginalString, out var refreshed))
+                        {
+                            status.LastUpdated = refreshed.LastUpdated;
```
LTA-Thinking, line 203: This is having a conflict returned from SQL saying that the version we have is out of date as shown by the last updated time not matching. This then just overrides the last updated time locally to match the DB and tries again. What does this gain over just not checking the last updated at all? It looks like we are just ignoring any other change in the DB, we aren't trying to resolve the conflict with the DB.

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/SearchParameterOperations.cs
```diff
@@ -60,64 +60,73 @@ public SearchParameterOperations(
...
-                customSearchException.Issues.Add(new OperationOutcomeIssue(
-                    OperationOutcomeConstants.IssueSeverity.Error,
-                    OperationOutcomeConstants.IssueType.Exception,
-                    ex.Message));
-
-                throw customSearchException;
-            }
+                        throw;
+                    }
+                    catch (Exception ex) when (!(ex is FhirException))
```
LTA-Thinking: Why is this addition needed? It is implied by this catch being after the one for FhirException.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Types/Types.sql
jestradaMS: Done and added to it's own .sql file under Types as discussed.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Types/SearchParamList.sql
```diff
@@ -0,0 +1,12 @@
+--DROP TYPE dbo.SearchParamList
+GO
+CREATE TYPE dbo.SearchParamList AS TABLE
+(
+    Uri                      varchar(128) COLLATE Latin1_General_100_CS_AS NOT NULL
+   ,Status                   varchar(20) NOT NULL
+   ,IsPartiallySupported     bit NOT NULL
+   ,LastUpdated              datetimeoffset(7) NULL
```
SergeyGaluzo: LastUpdated should be NOT NULL. Right?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/UpsertSearchParamsWithOptimisticConcurrency.sql
```diff
@@ -0,0 +1,81 @@
...
+    Action VARCHAR (20)  NOT NULL);
+
+-- Declare table to collect concurrency conflicts
+DECLARE @conflictedRows TABLE (
+    Uri VARCHAR(128) COLLATE Latin1_General_100_CS_AS NOT NULL
+);
+
+-- Adding WITH TABLOCKX higher up in transaction per PR 5097 review comments
+-- Acquire and hold an exclusive table lock for the entire transaction to prevent parameters from being added or modified during upsert.
+SELECT TOP 0 * FROM dbo.SearchParam WITH (TABLOCKX);
```
SergeyGaluzo: No need to put tablockx here. You can put it on next statement (insert into conflicts)

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/UpsertSearchParamsWithOptimisticConcurrency.sql
```diff
@@ -0,0 +1,81 @@
...
+--     Given a set of search parameters, creates or updates the parameters.
+--     Implements optimistic concurrency control using LastUpdated.
+--
+-- PARAMETERS
+--     @searchParams
+--         * The updated existing search parameters or the new search parameters
+--
+-- RETURN VALUE
+--     The IDs and URIs of the search parameters that were inserted (not updated).
+--
```
SergeyGaluzo, line 18: Please remove the comment

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SqlServerSearchParameterStatusDataStore.cs
SergeyGaluzo: Should not we raise an exception in case of conflict and let customer deal with it?

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/SearchParameterOperations.cs
jestradaMS: removed. was added to satisfy a codeql flag for generic exception.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
jestradaMS: Add logic for backwards compatibility as discussed

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/UpsertSearchParamsWithOptimisticConcurrency.sql
jestradaMS: Not sure why to remove, but I did update the comment to be accurate. Why should it be removed otherwise?

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Search/Registry/SearchParameterConcurrencyManagerTests.cs
```diff
@@ -0,0 +1,362 @@
...
+            // Assert
+            Assert.Equal(concurrentOperations, results.Length);
+            Assert.Equal(concurrentOperations, executionOrder.Count);
+            Assert.Equal(concurrentOperations, executionStartTimes.Count);
+            Assert.Equal(concurrentOperations, executionEndTimes.Count);
+
+            // Verify operations executed sequentially (no overlap)
+            for (int i = 1; i < executionStartTimes.Count; i++)
+            {
+                // Each operation should start after the previous one ended
```
LTA-Thinking: Should this also check execution order?

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Search/Registry/SearchParameterConcurrencyManagerTests.cs
```diff
@@ -0,0 +1,362 @@
...
+            Assert.Equal(operationsPerUri, uri2ExecutionTimes.Count);
+
+            // Verify that operations on different URIs can execute concurrently
+            // by checking that at least one operation from URI2 started before all URI1 operations completed
+            var minUri1Time = uri1ExecutionTimes[0];
+            var minUri2Time = uri2ExecutionTimes[0];
+
+            // Both should start around the same time (within a reasonable window)
+            var timeDifference = Math.Abs((minUri1Time - minUri2Time).TotalMilliseconds);
+            Assert.True(timeDifference < 50, $"Operations on different URIs should start concurrently. Time difference: {timeDifference}ms");
```
LTA-Thinking: I'm a little nervous about this test being flaky, but I think it is ok to try it out.

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Search/Registry/SearchParameterConcurrencyManagerTests.cs
```diff
@@ -0,0 +1,362 @@
...
+            // Assert
+            await Assert.ThrowsAsync<TaskCanceledException>(() => operationTask);
+
+            // Verify lock is released
+            await Task.Delay(10); // Allow cleanup
+            Assert.Equal(0, SearchParameterConcurrencyManager.ActiveLockCount);
+        }
+
+        [Fact]
+        public async Task GivenLongRunningOperations_WhenExecutingManySequentiallyOnSameUri_ThenMemoryIsReclaimed()
```
LTA-Thinking, line 361: Nit: They are executing on different uris.

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Search/Registry/SearchParameterConcurrencyManagerTests.cs
```diff
@@ -0,0 +1,362 @@
...
+            var tasks = new Task[threadCount];
+            for (int i = 0; i < threadCount; i++)
+            {
+                tasks[i] = Task.Run(async () =>
+                {
+                    barrier.SignalAndWait(); // Ensure all threads start simultaneously
+
+                    await SearchParameterConcurrencyManager.ExecuteWithLockAsync(TestUri1, async () =>
+                    {
+                        var currentCount = Interlocked.Increment(ref executionCount);
```
LTA-Thinking, line 409: Doesn't the interlocked nullify the test results? It is adding another locking layer so it wouldn't be clear if the concurrency manager was working.

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Search/Registry/SearchParameterConcurrencyTests.cs
```diff
@@ -0,0 +1,155 @@
...
+using Microsoft.Health.Fhir.Core.Features.Search.Registry;
+using Microsoft.Health.Fhir.Tests.Common;
+using Microsoft.Health.Test.Utilities;
+using Xunit;
+
+namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Search.Registry
+{
+    [Trait(Traits.OwningTeam, OwningTeam.Fhir)]
+    [Trait(Traits.Category, Categories.Search)]
+    public class SearchParameterConcurrencyTests
```
LTA-Thinking: This class has a lot of overlap with SearchParameterConcurrencyManagerTests. Do we need the double coverage?

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SearchParameterOptimisticConcurrencyIntegrationTests.cs
```diff
@@ -0,0 +1,494 @@
...
+            {
+                // Cleanup
+                await _testHelper.DeleteSearchParameterStatusAsync(testUri);
+            }
+        }
+
+        /// <summary>
+        /// Determines if this is a SQL Server fixture with schema version 94 or higher (optimistic concurrency support).
+        /// </summary>
+        private bool IsSqlServerWithOptimisticConcurrency()
```
LTA-Thinking: When will this return false? The test class is flagged as only running on SQL Server and the SqlServerSearchParameterStatusDataStore was in the fixture before this change.

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SearchParameterOptimisticConcurrencyIntegrationTests.cs
```diff
@@ -0,0 +1,494 @@
...
+                var staleUpdate = new ResourceSearchParameterStatus
+                {
+                    Uri = createdStatus.Uri,
+                    Status = SearchParameterStatus.Supported,
+                    IsPartiallySupported = true,
+                    LastUpdated = createdStatus.LastUpdated, // This is now stale
+                };
+
+                // Act - This should succeed due to retry mechanism
+                await _fixture.SearchParameterStatusDataStore.UpsertStatuses(new[] { staleUpdate }, CancellationToken.None);
```
LTA-Thinking, line 175: Is there any way to test that the retry actually happened? This is showing that the upsert worked, but it doesn't show why it is working.

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SearchParameterOptimisticConcurrencyIntegrationTests.cs
```diff
@@ -0,0 +1,494 @@
...
+            }
+        }
+
+        [SkippableFact]
+        public void GivenMaxRetriesExceeded_WhenConcurrencyConflictsPersist_ThenThrowsConcurrencyException()
+        {
+            // Skip if not SQL Server or schema version < 94
+            Skip.If(!IsSqlServerWithOptimisticConcurrency(), "Schema version 94+ required for optimistic concurrency");
+
+            // This test would need to be implemented with a custom data store that simulates
```
LTA-Thinking: Nit: duplicate line

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SearchParameterOptimisticConcurrencyIntegrationTests.cs
```diff
@@ -0,0 +1,494 @@
...
+
+                Assert.Equal(2, successes); // Both should succeed due to retry mechanism
+                Assert.Equal(0, failures);
+
+                // Verify that one of the updates was applied (whichever succeeded last)
+                var finalStatuses = await _fixture.SearchParameterStatusDataStore.GetSearchParameterStatuses(CancellationToken.None);
+                var finalStatus = finalStatuses.First(s => s.Uri.ToString() == testUri);
+
+                // The final status should be one of the two updates
+                Assert.True(
```
LTA-Thinking: Can this check that whichever status is the final status has the last updated time from the corresponding result?

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SearchParameterOptimisticConcurrencyIntegrationTests.cs
```diff
@@ -0,0 +1,494 @@
...
+            }
+            finally
+            {
+                // Cleanup
+                await _testHelper.DeleteSearchParameterStatusAsync(testUri);
+            }
+        }
+
+        [SkippableFact]
+        public async Task GivenOptimisticConcurrencyDetection_WhenLastUpdatedChangesBeforeUpdate_ThenRetryMechanismHandlesConflict()
```
LTA-Thinking, line 328: What is this test checking that is different from the test on line 143?

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SearchParameterOptimisticConcurrencyIntegrationTests.cs
```diff
@@ -0,0 +1,494 @@
...
+            finally
+            {
+                // Cleanup
+                await _testHelper.DeleteSearchParameterStatusAsync(existingTestUri);
+                await _testHelper.DeleteSearchParameterStatusAsync(newTestUri);
+            }
+        }
+
+        [SkippableFact]
+        public async Task GivenSchemaVersionBelow94_WhenUpserting_ThenLastUpdatedIsIgnoredGracefully()
```
LTA-Thinking: I can see the value in having tested this, but I don't know if it should be in the codebase permanently. I'm afraid this will become clutter after we deploy the change and verify no rollback is needed.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/UpsertSearchParamsWithOptimisticConcurrency.sql
```diff
@@ -0,0 +1,88 @@
+﻿/*************************************************************
```
SergeyGaluzo, line 1: Please remove all text before CREATE PROCEDURE. This meaningless text is prepended in all system tables, so it is hard to debug.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/UpsertSearchParamsWithOptimisticConcurrency.sql
```diff
@@ -0,0 +1,88 @@
...
+    Stored procedures - UpsertSearchParams_2
```
SergeyGaluzo: Please remove all the text before CREATE PROCEDURE. This meaningless text is prepended in all system tables, so it is hard to debug.

SergeyGaluzo (approved):

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/UpsertSearchParamsWithOptimisticConcurrency.sql
jestradaMS: Adding comments is common practice in general as well as within our repo currently. Additionally, after reviewing, none of this text is carried over into SQL Stored Procedure definition.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/UpsertSearchParamsWithOptimisticConcurrency.sql
jestradaMS: Adding comments is common practice in general as well as within our repo currently. Additionally, after reviewing, none of this text is carried over into SQL Stored Procedure definition.

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Search/Registry/SearchParameterConcurrencyManagerTests.cs
jestradaMS: Fair, made updates to take a different approach for these scenarios.

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Search/Registry/SearchParameterConcurrencyManagerTests.cs
jestradaMS: Different sync levels as well as Interlocked is atomic and does not use "locks" like SemaphoreSlim and therefore should not create any conflicts or contention.

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Search/Registry/SearchParameterConcurrencyTests.cs
jestradaMS: Combined into single file and cleaned up tests.

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SearchParameterOptimisticConcurrencyIntegrationTests.cs
jestradaMS: Agreed, this is not necessary. Removed it and the skip logic.

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SearchParameterOptimisticConcurrencyIntegrationTests.cs
jestradaMS: removed all skips and logic for this per a different comment

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SearchParameterOptimisticConcurrencyIntegrationTests.cs
jestradaMS: Tests were rewritten and this was removed in process.

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SearchParameterOptimisticConcurrencyIntegrationTests.cs
jestradaMS: removed during re-write.

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlServerSearchParameterInitializationTests.cs
```diff
@@ -72,7 +72,7 @@ public async Task GivenADatabaseWithSearchParametersDisabled_WhenInitializing_Th
             });
     }
 
-    [Fact]
+    [Fact(Skip = "No longer valid")]
```
LTA-Thinking: Should this just be deleted then?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/95.diff.sql
```diff
@@ -73,5 +76,68 @@ FROM   dbo.SearchParam AS searchParam
...
+END
+
+IF NOT EXISTS (SELECT * FROM sys.columns WHERE object_id = OBJECT_ID('dbo.SearchParam') AND name = 'IsPartiallySupported')
+BEGIN
+    ALTER TABLE dbo.SearchParam ADD IsPartiallySupported bit NULL;
+END
+
+-- Step 2: Update any NULL values with appropriate defaults
+UPDATE dbo.SearchParam 
+SET Status = 'Enabled'
```
LTA-Thinking: Is Enabled the safe default value?

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SearchParameterOptimisticConcurrencyIntegrationTests.cs
jestradaMS: Leaving as is for now as discussed

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/95.diff.sql
jestradaMS: Updated to disabled for a safter value from null

LTA-Thinking (approved):

PR #5099, opened 2025-08-16
Ability to provide bytes to read value in import

SergeyGaluzo:


src/Microsoft.Health.Fhir.Core/Features/Routing/KnownRoutes.cs
```diff
@@ -65,6 +65,7 @@ internal class KnownRoutes
         public const string Import = "$import";
         public const string ImportDataOperationDefinition = OperationDefinition + "/" + OperationsConstants.Import;
         public const string ImportJobLocation = OperationsConstants.Operations + "/" + OperationsConstants.Import + "/" + IdRouteSegment;
+        public const string ImportJobLocationWithDetails = OperationsConstants.Operations + "/" + OperationsConstants.Import + "/" + IdRouteSegment + "/true";
```
LTA-Thinking, line 68: This feels like it should be a query parameter instead of a url segment.
IE: `_operations/import/1?_details=true`

SergeyGaluzo: OK. Fixed.

PR #5144, opened 2025-09-15
Add silent meta update

LTA-Thinking:
## Description
Adds a query parameter for PUT updates that allows changes that only update metadata fields to happen without changing the last updated time or the version.
The parameter is `_silent-meta`
ex: `PUT <fhir base>/Patient/12456?_silent-meta=true`

## Related issues
Addresses [User Story 167335](https://microsofthealth.visualstudio.com/Health/_workitems/edit/167335)

## Testing
Manual testing so far, unit and e2e tests to be added

## FHIR Team Checklist


SergeyGaluzo:
Are any of these meta fields searchable?

LTA-Thinking:
Yes, tags are a default search parameter in meta and other custom search parameters can be created.

SergeyGaluzo:
@LTA-Thinking, @brendankowitz 
Planned "meta update" performs updates of resources in place without version and last updated change. This is a violation of current FHIR storage principal - append only. I think this functionality should not be supported.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -949,6 +959,91 @@ private bool ExistingRawResourceIsEqualToInput(RawResource input, RawResource ex
...
 
+        private bool ChangesAreOnlyInMetadata(ResourceWrapper inputWrapper, ResourceWrapper existingWrapper)
+        {
+            var inputResource = (Hl7.Fhir.Model.Resource)_resourceDeserializer.Deserialize(inputWrapper).ResourceInstance;
+            var existingResource = (Hl7.Fhir.Model.Resource)_resourceDeserializer.Deserialize(existingWrapper).ResourceInstance;
+
+            var inputMeta = inputResource.Meta;
+            var existingMeta = existingResource.Meta;
+
+            if (inputMeta.Equals(existingMeta))
```
feordin: Does this Equal comparison use the EqualValues method that you defined in the extension?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -949,6 +959,91 @@ private bool ExistingRawResourceIsEqualToInput(RawResource input, RawResource ex
             }
         }
 
+        private bool ChangesAreOnlyInMetadata(ResourceWrapper inputWrapper, ResourceWrapper existingWrapper)
+        {
+            var inputResource = (Hl7.Fhir.Model.Resource)_resourceDeserializer.Deserialize(inputWrapper).ResourceInstance;
+            var existingResource = (Hl7.Fhir.Model.Resource)_resourceDeserializer.Deserialize(existingWrapper).ResourceInstance;
```
feordin: Do we need to deserialize here?  Are these resources already deserialized elsewhere while processing the request?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -949,6 +959,91 @@ private bool ExistingRawResourceIsEqualToInput(RawResource input, RawResource ex
...
+            var inputMeta = inputResource.Meta;
+            var existingMeta = existingResource.Meta;
+
+            if (inputMeta.Equals(existingMeta))
+            {
+                return false; // Difference is in a non-meta field
+            }
+            else
+            {
+                var inputChildren = inputResource.NamedChildren.GetEnumerator();
```
feordin: I see we have a method: ExistingRawResourceIsEqualToInput, which replaces some values in the meta field, and then simply compares the two rawresources.  Could we do something similar where we replace the entire meta field from the incoming with the existing and then simply compare the entire resource?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
SergeyGaluzo: I think this comparison will be slow. This is how we did it MergeAsync before I added quick string comparisons. Can we use something similar here?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
SergeyGaluzo: @feordin :-) I added my comment above before reading yours...

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
LTA-Thinking: No, this is an optimistic check to see if the two are identical references. It will only return true if inputMeta and existingMeta are the same object.

LTA-Thinking:
I changed the name to Meta History as the only thing it now does is decide to store or not store a historical version.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
LTA-Thinking: I changed the way this is handled, it now just replaces the meta field with empty.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
LTA-Thinking: Changed to avoid deserialization.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
feordin: Ok :)

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -64,6 +66,7 @@ internal class SqlServerFhirDataStore : IFhirDataStore, IProvideCapability
         private readonly SchemaInformation _schemaInformation;
         private readonly IModelInfoProvider _modelInfoProvider;
         private readonly IImportErrorSerializer _importErrorSerializer;
+        private readonly IResourceDeserializer _resourceDeserializer;
```
feordin: Is this still needed? Or can it now be removed?

brendankowitz (commented):
# Code Review: PR #5144 - Add Silent Meta Update

## Overview

This PR adds a `_meta-history` query parameter for PUT requests that allows metadata-only changes to be made without creating a historical version of the resource. This is an optimization to reduce database bloat when only metadata (like tags) is updated.

## Critical Issue - Potential PATCH Regression and Data Loss

### **HIGH SEVERITY: Default Value Mismatch Causes History Loss for All PATCH Operations**

**Location:** `UpsertResourceRequest.cs:18`

```csharp
public UpsertResourceRequest(ResourceElement resource, BundleResourceContext bundleResourceContext = null, 
    WeakETag weakETag = null, bool metaHistory = false)  // ⚠️ Default is FALSE
```

**Location:** `PatchResourceHandler.cs:82`

```csharp
return await _mediator.Send<UpsertResourceResponse>(
    new UpsertResourceRequest(patchedResource, request.BundleResourceContext, request.WeakETag), 
    cancellationToken);  // ⚠️ metaHistory not passed, uses default FALSE
```

**The Problem:**
1. `UpsertResourceRequest` constructor defaults `metaHistory = false`
2. `PatchResourceHandler` doesn't pass the `metaHistory` parameter
3. When `metaHistory = false` AND `ChangesAreOnlyInMetadata()` returns `true`, **NO HISTORY IS KEPT**

**Location:** `SqlServerFhirDataStore.cs:304-307`
```csharp
else if (!resourceExt.MetaHistory && ChangesAreOnlyInMetadata(resource, existingResource))
{
    metaHistory = false;  // Prevents history creation!
}
```

**Location:** `SqlServerFhirDataStore.cs:350`
```csharp
mergeWrappersWithVersions.Add((new MergeResourceWrapper(resource, resourceExt.KeepHistory && metaHistory, ...
```

### **Impact Scenario:**
Any PATCH operation that only modifies metadata (tags, security labels, etc.) will **silently skip creating a historical version**, even though:
- The user never requested this behavior
- PATCH should always create history (standard FHIR behavior)
- The ADR explicitly states: *"This is planned to be added to PATCH...in the future"* - meaning it was NOT intended for PATCH yet

### **Inconsistency: ResourceWrapperOperation vs UpsertResourceRequest Defaults**

| Class | `metaHistory` Default |
|-------|----------------------|
| `ResourceWrapperOperation` | `true` (safe) |
| `UpsertResourceRequest` | `false` (dangerous) |

This inconsistency is confusing and error-prone.

---

## Additional Issues

### 2. Fragile JSON String Parsing

**Location:** `StringExtensions.cs:154-207` (`GetJsonSection`)

```csharp
public static string GetJsonSection(this string input, int startingIndex)
```

This method parses JSON by tracking brackets and quotes manually. Risks:
- Unicode escapes not handled (`\u0022` for quote)
- Edge cases with nested objects containing `"meta":` in string values could cause incorrect parsing
- Returns empty string for unmatched brackets instead of throwing (silent failures)

**Location:** `SqlServerFhirDataStore.cs:974`
```csharp
var inputMeta = inputData.GetJsonSection(inputMetaStartIndex);
inputDataWithoutMeta = inputData.Replace(inputMeta, string.Empty, StringComparison.Ordinal);
```

Replacing the meta section content but leaving `"meta":{}` structure could cause comparison mismatches.

### 3. Inconsistent Default in FhirController vs UpsertResourceRequest

**Location:** `FhirController.cs:228`
```csharp
[FromQuery(Name = KnownQueryParameterNames.MetaHistory)] bool metaHistory = true
```

The controller defaults to `true`, but:
- `UpsertResourceRequest` defaults to `false`
- This means explicitly calling the constructor without the parameter behaves **opposite** to the HTTP API

### 4. Unused Import

**Location:** `StringExtensions.cs:9`
```csharp
using Hl7.FhirPath.Sprache;  // Not used in file
```

---

## Recommendations

### Critical Fix Required:
1. **Change `UpsertResourceRequest.metaHistory` default to `true`** to match safe behavior and `ResourceWrapperOperation`
2. **Update `PatchResourceHandler` to explicitly pass `metaHistory: true`** to ensure PATCH always creates history

### Suggested Fix:
```csharp
// UpsertResourceRequest.cs
public UpsertResourceRequest(ResourceElement resource, BundleResourceContext bundleResourceContext = null, 
    WeakETag weakETag = null, bool metaHistory = true)  // Change to TRUE
```

```csharp
// PatchResourceHandler.cs - Be explicit
return await _mediator.Send<UpsertResourceResponse>(
    new UpsertResourceRequest(patchedResource, request.BundleResourceContext, request.WeakETag, metaHistory: true), 
    cancellationToken);
```

---

## Summary

| Category | Rating |
|----------|--------|
| **Code Correctness** | ❌ Critical bug - PATCH silently loses history |
| **Project Conventions** | ⚠️ Inconsistent defaults between classes |
| **Performance** | ✅ Good - Reduces unnecessary DB writes |
| **Test Coverage** | ⚠️ Missing tests for PATCH with meta-only changes |
| **Security** | ⚠️ Data integrity risk (audit trail loss) |

**Verdict:** This PR contains a **critical regression** where PATCH operations modifying only metadata will silently lose historical versions. This should be patched immediately.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5159, opened 2025-09-23
Simplified query logic for id api requests

PTaladay:
## Description
Added check for simple id read queries. If check is true then use the dbo.ReadResource stored procedure instead of SQL query generation logic in order to avoid improper query execution plan generation.

## Related issues
Addresses [issue [AB#151824](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/151824)].

## Testing
Added unit tests to validate change in functionality.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
@@ -1793,18 +1793,15 @@ SELECT isnull(min(ResourceSurrogateId), 0), isnull(max(ResourceSurrogateId), 0),
             SearchParameterExpression resourceTypeExpression = null;
             SearchParameterExpression resourceIdExpression = null;
 
-            foreach (var tableExpression in expression.ResourceTableExpressions)
+            foreach (var searchParamExpression in expression.ResourceTableExpressions.OfType<SearchParameterExpression>())
             {
-                if (tableExpression is SearchParameterExpression searchParamExpression)
+                if (searchParamExpression.Parameter.Name == SearchParameterNames.ResourceType)
                 {
-                    if (searchParamExpression.Parameter.Name == SearchParameterNames.ResourceType)
-                    {
-                        resourceTypeExpression = searchParamExpression;
-                    }
-                    else if (searchParamExpression.Parameter.Name == SearchParameterNames.Id)
-                    {
-                        resourceIdExpression = searchParamExpression;
-                    }
+                    resourceTypeExpression = searchParamExpression;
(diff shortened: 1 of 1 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -711,6 +716,91 @@ private static void PopulateSqlCommandFromQueryHints(SqlCommand command, short r
...
+            {
+                return false;
+            }
+
+            // Populate command to use ReadResource stored procedure
+            command.CommandType = CommandType.StoredProcedure;
+            command.CommandText = "dbo.ReadResource";
+            command.Parameters.AddWithValue("@resourceTypeId", resourceTypeId);
+            command.Parameters.AddWithValue("@resourceId", resourceId);
+            command.Parameters.AddWithValue("@version", DBNull.Value); // null for latest version
```
feordin, line 779: Does this work if you specify a version?  It looks like we do not attempt to extract the version, which means, I think that this will fall back to generating SQL rather than using the stored proc.  Perhaps that is fine, or perhaps we should update this to try and extract the version?

SergeyGaluzo:
I did not find ReadResource.sql. Does this stored procedure exist? Commonly used and more generic one is GetResources. Please use GetResources stored procedure. 
There are already existing GetAsync methods in the SqlServerFhirDataStore class. We should use them in search instead of creating new ones.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
PTaladay: If you specify a version it goes directly to the data store and completely bypasses the search service. The only time when that isn't true is when ApplyFineGrainedAccessControl is enabled, which would then fall back to generating SQL. If we want to make this change to take into account versioned calls then this will be a more significant change as it will require than just extracting the version and passing it as it appears that the search service was only designed to handle latest version.

feordin (approved):

SergeyGaluzo:
How is it possible that all tests pass with the code that uses not existing stored procedure?

SergeyGaluzo (changes requested):
Please address my comments.

PTaladay:
> How is it possible that all tests pass with the code that uses not existing stored procedure?

I didn't catch a change that the AI had made. I'm currently in the process of testing with the suggestion you provided, and doing additional validation to make sure that this code isn't actually skipped.

PR #5168, opened 2025-09-26
Change profile locking

LTA-Thinking:
## Description
Changes the way profiles are loaded by the validator to avoid long waits on locks when the cache isn't expired. 
A user reported long wait times on some creates, and we traced the waits to validation. These are happening randomly a small percentage of the time and we have been unable to trace the delay to an issue with the profiles or resources. This is another attempt to remove a possible cause of the latency.

## Related issues
Addresses [User Story 154657](https://microsofthealth.visualstudio.com/Health/_workitems/edit/154657)

## Testing
Existing E2E and Unit tests. User scenario has proven impossible to reproduce so it is unknown if this will fix the issue.

## FHIR Team Checklist

Patch


src/Microsoft.Health.Fhir.Shared.Core/Features/Validation/ServerProvideProfileValidation.cs
```diff
@@ -33,15 +33,19 @@ namespace Microsoft.Health.Fhir.Core.Features.Validation
     /// <summary>
     /// Provides profiles by fetching them from the server.
     /// </summary>
+#pragma warning disable CA1001 // Types that own disposable fields should be disposable. ServerProvideProfileValidation is a Singleton class.
```
brendankowitz: This can be disposable, singletons are still tracked by the DI container, but will only get cleaned up when the root container shuts down

LTA-Thinking: What is the benefit of making it disposable though? There is no chance of a memory leak since it is being shut down with the entire instance.  I'd rather not add code that doesn't do anything.

src/Microsoft.Health.Fhir.Core/Features/Conformance/ICapabilityStatementBuilder.cs
```diff
@@ -53,7 +53,7 @@ public interface ICapabilityStatementBuilder
         /// Updates capability statement to lastest supported profiles by pulling them from database.
         /// </summary>
         /// <param name="disableCacheRefresh">Disable pull from database and check cached version in memory. Needed to prevent circular calls to sync profiles.</param>
-        ICapabilityStatementBuilder SyncProfiles(bool disableCacheRefresh = false);
+        Task<ICapabilityStatementBuilder> SyncProfiles(bool disableCacheRefresh = false);
```
fhibf: Could we rename this method to SyncProfilesSync and expected a CancellationToken?

src/Microsoft.Health.Fhir.Core/Features/Conformance/IProvideCapability.cs
```diff
@@ -14,6 +16,6 @@ public interface IProvideCapability
         /// Allows this component to configure the server's CapabilityStatement.
         /// </summary>
         /// <param name="builder">Instance of the CapabilityStatementBuilder.</param>
-        void Build(ICapabilityStatementBuilder builder);
+        Task Build(ICapabilityStatementBuilder builder);
```
fhibf: Is it possible to rename this method to BuildAsync and expect a cancellation token?

src/Microsoft.Health.Fhir.Core/Features/Validation/ISupportedProfilesStore.cs
```diff
@@ -14,7 +15,7 @@ public interface ISupportedProfilesStore
         /// </summary>
         /// <param name="resourceType">Resource type to get profiles.</param>
         /// <param name="disableCacheRefresh">Should we check server for new updates or get data out of cache.</param>
-        IEnumerable<string> GetSupportedProfiles(string resourceType, bool disableCacheRefresh = false);
+        Task<IEnumerable<string>> GetSupportedProfiles(string resourceType, bool disableCacheRefresh = false);
```
fhibf: Is it possible to rename this method to GetSupportedProfilesAsync and expect a cancellation token?

src/Microsoft.Health.Fhir.Shared.Core/Features/Validation/ServerProvideProfileValidation.cs
```diff
@@ -156,22 +167,27 @@ private async Task<List<ArtifactSummary>> GetSummaries()
...
+            }
+
             using (var memoryStream = new MemoryStream(Encoding.UTF8.GetBytes(summary.Origin)))
             using (var navStream = new JsonNavigatorStream(memoryStream))
             {
                 if (navStream.Seek(summary.Position))
                 {
                     if (navStream.Current != null)
                     {
-                        // TODO: Cache this parsed resource, to prevent parsing again and again
```
fhibf, line 173: Should we keep the TODO for in future implement a caching logic?

src/Microsoft.Health.Fhir.Shared.Core/Features/Validation/ServerProvideProfileValidation.cs
```diff
@@ -33,15 +33,19 @@ namespace Microsoft.Health.Fhir.Core.Features.Validation
...
+#pragma warning disable CA1001 // Types that own disposable fields should be disposable. ServerProvideProfileValidation is a Singleton class.
     public sealed class ServerProvideProfileValidation : IProvideProfilesForValidation
+#pragma warning restore CA1001 // Types that own disposable fields should be disposable. ServerProvideProfileValidation is a Singleton class.
     {
         private static HashSet<string> _supportedTypes = new HashSet<string>() { "ValueSet", "StructureDefinition", "CodeSystem" };
+
+        private readonly SemaphoreSlim _cacheSemaphore = new SemaphoreSlim(1, 1);
         private readonly Func<IScoped<ISearchService>> _searchServiceFactory;
         private readonly ValidateOperationConfiguration _validateOperationConfig;
+        private readonly Dictionary<string, Resource> _resourcesByUri = new Dictionary<string, Resource>();
```
fhibf: As this is a Dictionary and ServerProvideProfileValidation is a single instance running in memory, should we limit the number of resources that can be added to _resourcesByUri?

A while ago we added support to DotNet memory caching in the FHIR service. Should we rely on it?

src/Microsoft.Health.Fhir.Shared.Core/Features/Validation/ServerProvideProfileValidation.cs
LTA-Thinking: I added a cache as part of this change.

src/Microsoft.Health.Fhir.Shared.Core/Features/Validation/ServerProvideProfileValidation.cs
brendankowitz: I don't think we should ignore the dispose because we assume the components usage, it should enable a cleaner shutdown should anything actually be waiting

src/Microsoft.Health.Fhir.Core/Features/Storage/MostRecentlyUsedCache.cs
```diff
@@ -0,0 +1,75 @@
...
+            _cache.Clear();
+        }
+
+        public bool TryGetValue(string key, out T item)
+        {
+            if (_cache.TryGetValue(key, out item))
+            {
+                // Move the accessed item to the end to mark it as recently used
+                _cache.Remove(key);
+                _cache[key] = item;
```
brendankowitz: I don't think .net dictionary guarantees order. Perhaps look at the IFhirMemoryCache interface we've used in some of the other refactoring?

LTA-Thinking: This was specifically to avoid IMemoryCache as it requires knowledge of the size of the resource, which we don't have with Resource. I'll see what I can do to guaranty the order.

LTA-Thinking: Changed order handling

fhibf (approved):

fhibf (dismissed):

fhibf (approved):

brendankowitz:
Doesn't have to be this PR, but I think this is another low hanging fruit fix too in [ProfileValidator](https://github.com/microsoft/fhir-server/blob/6fd529d2d871865f07a38e3fe46baa223a3fe1a6/src/Microsoft.Health.Fhir.Shared.Core/Features/Validation/ProfileValidator.cs#L80)
image

PR #5180, opened 2025-10-07
Add reference resource id back to hash

SergeyGaluzo:
There are cases when number of rows in the ReferenceSearchParam are very different for different ReferenceResourceId values. Reusing plan for different values is a problem. We should not be caching these plans.

Customer who might still benefit form plan caching can use explicit caching headers.


feordin (approved):
We previously changed the hashing to remove referenceResourceId to cause an increase in query plan caching.  However, now that we see customers with highly variable cardinality in the references, it makes sense to add this back to the hash value, which will cause the query plan cache to be invalid.  



The possible negative of an increased percentage of queries having their query plan recalculated is the CPU time spent regenerating query plans that may have been valid.  As we now have given customer's more control via the header, and Fernando's new feature to automatically include or not the hash value, I agree with Sergey that it makes sense to add this value back into the hash.  It gives the opportunity to cache or not based on the referenceId.

PR #5222, opened 2025-11-04
Disallow job creation after cancellation

LTA-Thinking:
## Description
Currently there is a window of time where jobs can be created in a job group after the group has been cancelled. This can happen if an orchestrator job is running when a cancellation request is submitted. The orchestrator will be marked for cancellation, but it will still create new jobs that aren't marked for cancellation. This means jobs will continue to run after they should have been cancelled.

## Related issues
Addresses [Bug 175373](https://microsofthealth.visualstudio.com/Health/_workitems/edit/175373)

## Testing
Integration test

## FHIR Team Checklist

Patch

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/EnqueueJobs.sql
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/EnqueueJobs.sql
@@ -37,7 +37,7 @@ BEGIN TRY
       RAISERROR('There are other active job groups',18,127)
 
     IF @GroupId IS NOT NULL AND EXISTS (SELECT * FROM dbo.JobQueue WHERE QueueType = @QueueType AND JobId = @GroupId AND CancelRequested = 1)
-      RAISERROR('The specified job group is cancelled',18,128)
+      RAISERROR('The specified job group is cancelled',18,127)
 
     SET @MaxJobId = isnull((SELECT TOP 1 JobId FROM dbo.JobQueue WHERE QueueType = @QueueType ORDER BY JobId DESC),0)
   
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlQueueClient.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlQueueClient.cs
@@ -165,11 +165,11 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Storage
             {
                 return await sqlCommand.ExecuteReaderAsync(_sqlRetryService, JobInfoExtensions.LoadJobInfo, _logger, cancellationToken);
             }
-            catch (SqlException sqlEx) when (forceOneActiveJobGroup && sqlEx.State == 127 && sqlEx.Message.Contains("cancel", StringComparison.OrdinalIgnoreCase))
+            catch (SqlException sqlEx) when (forceOneActiveJobGroup && sqlEx.State == 127 && sqlEx.Message.Contains("other active job", StringComparison.OrdinalIgnoreCase))
             {
                 throw new JobConflictException(sqlEx.Message);
             }
-            catch (SqlException sqlEx) when (sqlEx.State == 128)
+            catch (SqlException sqlEx) when (sqlEx.State == 127 && sqlEx.Message.Contains("cancel", StringComparison.OrdinalIgnoreCase))
             {
                 _logger.LogError(sqlEx, "EnqueueAsync failed due to job orchestrator being cancelled.");
                 throw new OperationCanceledException("Job has been cancelled.", sqlEx);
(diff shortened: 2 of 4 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/EnqueueJobs.sql
```diff
@@ -36,6 +36,9 @@ BEGIN TRY
     IF @ForceOneActiveJobGroup = 1 AND EXISTS (SELECT * FROM dbo.JobQueue WHERE QueueType = @QueueType AND Status IN (0,1) AND (@GroupId IS NULL OR GroupId <> @GroupId))
       RAISERROR('There are other active job groups',18,127)
 
+    IF @GroupId IS NOT NULL AND EXISTS (SELECT * FROM dbo.JobQueue WHERE QueueType = @QueueType AND GroupId = @GroupId AND (Status = 4 OR CancelRequested = 1))
```
mikaelweave: This is smart - there is no way to ensure this race condition won't happen except inside a transaction in the db.

mikaelweave (dismissed):
Makes sense - I know we have handled this in explicit job types in the past (like CosmosDB export) but this should be on the queue client level.

Even if we create jobs in the orchestrator vs continuation jobs, I could see how we can encounter this edge case. For customers, this means cancelled jobs will cancel faster, correct?

LTA-Thinking:
Previously if a job slipped through like this it would run to completion. The orchestrator would be cancelled, but all the processing jobs would run. So in a sense it will "cancel faster", but that is compared to it previously not cancelling at all :)

rajithaalurims (dismissed):

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlQueueClient.cs
```diff
@@ -165,7 +165,7 @@ public async Task<IReadOnlyList<JobInfo>> EnqueueAsync(byte queueType, string[]
             {
                 return await sqlCommand.ExecuteReaderAsync(_sqlRetryService, JobInfoExtensions.LoadJobInfo, _logger, cancellationToken);
             }
-            catch (SqlException sqlEx) when (forceOneActiveJobGroup && sqlEx.State == 127)
+            catch (SqlException sqlEx) when ((forceOneActiveJobGroup && sqlEx.State == 127) || sqlEx.State == 128)
```
mikaelweave: NIT: This or case is a bit hard to read. I thought `forceOneActiveJobGroup ` was a condition for both error codes. Maybe reordering could help?

Code is fine as is - this is more focused on readability.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/EnqueueJobs.sql
SergeyGaluzo: We don't have an index to support this query.

SergeyGaluzo (changes requested):
Please resolve my comments

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/EnqueueJobs.sql
SergeyGaluzo: Can we just check 

(SELECT * FROM dbo.JobQueue WHERE QueueType = @QueueType AND **JobId** = @**GroupId** AND CancelRequested = 1)

instead of full group check?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/EnqueueJobs.sql
```diff
@@ -36,6 +36,9 @@ BEGIN TRY
...
+      RAISERROR('The specified job group is cancelled',18,128)
```
SergeyGaluzo: Might need change depending on my suggestion above

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/EnqueueJobs.sql
LTA-Thinking: Yes, that would work.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/EnqueueJobs.sql
LTA-Thinking: I think this message still works. While only the orchestrator is being checked, if any job in the group is cancelled the entire group should be cancelled.

SergeyGaluzo (dismissed):

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlQueueClient.cs
```diff
@@ -169,6 +169,11 @@ public async Task<IReadOnlyList<JobInfo>> EnqueueAsync(byte queueType, string[]
             {
                 throw new JobConflictException(sqlEx.Message);
             }
+            catch (SqlException sqlEx) when (sqlEx.State == 128)
```
SergeyGaluzo: 128 is a generic state that might exist for other not skippable errors. Please either raise very specific error or check error message.

SergeyGaluzo (changes requested):
Please address my comment about SQL error

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlQueueClient.cs
SergeyGaluzo: I added check for cancel

PR #5242, opened 2025-11-18
Fix search with include and sort

LTA-Thinking:
## Description
When using `_sort` and `_include`/`_revinclude` together there are issues when there are enough included resources to create an include continuation token. 

- If sorting by last updated descending the included results won't be in the bundle and no include continuation token will be returned. 
- If sorting by any other field it will cause a 500 if there are enough results with the sort field to fill a page of data.
- If during a search the number of matched results with the sort value isn't enough to fill a page, but is enough to generate an includes continuation token, the includes continuation token will be lost when the second search is run to get results that don't contain the sort value.
- If an includes continuation token is used with a sort field it will get data for both matches that contain the sort field and those that don't contain the sort field regardless of which type of matched result generated the includes continuation token.

This PR fixes all of these issues.

## Related issues
Addresses [Bug 177120](https://microsofthealth.visualstudio.com/Health/_workitems/edit/177120)

## Testing
Manual testing and new E2E tests.

## FHIR Team Checklist

Patch

### DIFF
--- a/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/SortTests.cs
+++ b/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Search/SortTests.cs
@@ -1202,6 +1202,7 @@ namespace Microsoft.Health.Fhir.Tests.E2E.Rest.Search
         [InlineData(10, "-birthdate")]
         [InlineData(5, "-birthdate")]
         [InlineData(4, "-birthdate")]
+        [HttpIntegrationFixtureArgumentSets(dataStores: DataStore.SqlServer)]
         public async Task GivenPatientsWithIncludedResources_WhenSearchedWithSortAndInclude_ThenTheSecondPhaseContinuationTokenIsReturned(int includesCount, string sort)
         {
             var tag = Guid.NewGuid().ToString();


src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -244,7 +244,7 @@ public override object VisitSqlRoot(SqlRootExpression expression, SearchOptions
 
                 StringBuilder.Append(VLatest.Resource.RawResource, resourceTableAlias);
 
-                if (IsSortValueNeeded(context))
+                if (IsSortValueNeeded(context) && !context.IsIncludesOperation)
```
mikaelweave, line 282: I'm a fan of using the property vs the helper method nice work

src/Microsoft.Health.Fhir.SqlServer/Features/Search/IncludesContinuationToken.cs
```diff
@@ -90,18 +108,35 @@ private void Initialize()
...
+                MatchResourceSurrogateIdMin = sid0 < sid1 ? sid0 : sid1;
+                MatchResourceSurrogateIdMax = sid0 < sid1 ? sid1 : sid0;
                 initialized = true;
 
                 if (_tokens.Length > 3)
                 {
-                    if (_tokens.Length == 5
-                        && short.TryParse(_tokens[3]?.ToString(), out tid)
-                        && long.TryParse(_tokens[4]?.ToString(), out sid0))
+                    if (_tokens.Length >= 5)
```
mikaelweave, line 136: I know a lot this code was existing but I'm seeing a lot of magic numbers in this block. Any chance you can add some comments to improve readability for those new to this area?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -147,6 +147,43 @@ public override async Task<SearchResult> SearchAsync(SearchOptions searchOptions
         {
             SqlSearchOptions sqlSearchOptions = new SqlSearchOptions(searchOptions);
 
+            if (sqlSearchOptions.IsIncludesOperation)
+            {
+                var includesContinuationToken = IncludesContinuationToken.FromString(sqlSearchOptions.IncludesContinuationToken);
+                if (includesContinuationToken == null)
+                {
+                    _logger.LogWarning("Bad Request (InvalidIncludesContinuationToken)");
+                    throw new BadRequestException(Resources.InvalidIncludesContinuationToken);
```
mikaelweave, line 156: Is a `BadRequestException` correct here? If the token is malformed is it an issue from the client or in our code?

mikaelweave (commented):
A couple questions

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
LTA-Thinking: I think the safer expectation is that the client passed in the wrong value for the CT. It is possible for us to have a bug in the CT generation code, but I don't think that is the right assumption to take with a user provided value.

LTA-Thinking:
There are still more edge cases to fix here. The behavior when using a second phase includes continuation token after the second page of first phase includes results isn't working correctly.

jestradaMS (dismissed):

jestradaMS (approved):

PR #5266, opened 2025-12-09
PUT/DELETE spec defined search params not allowed

paorodma-ms:
Updated responses when deleting nonexistent search params and spec defined search params. Avoid updating spec defined search params.

## Description

## Related issues
Addresses issue https://microsofthealth.visualstudio.com/Health/_workitems/edit/177538

## Testing
Describe how this change was tested. Using unit tests and REST API Calls.

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/SearchParameters/SearchParameterBehaviorTests.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/SearchParameters/SearchParameterBehaviorTests.cs
@@ -165,8 +165,6 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Search
         public async Task GivenADeleteResourceRequest_WhenDeletingASpecDefinedSearchParameterResource_ThenDeleteSearchParameterShouldNotBeCalled()
         {
             var searchParameter = new SearchParameter() { Id = "code", Url = "http://hl7.org/fhir/SearchParameter/code" };
-            var resource = searchParameter.ToTypedElement().ToResourceElement();
-
             var key = new ResourceKey("SearchParameter", "code");
             var request = new DeleteResourceRequest(key, DeleteOperation.SoftDelete);
 


src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Parameters/SearchParameterValidator.cs
```diff
@@ -171,6 +176,42 @@ public async Task ValidateSearchParameterInput(SearchParameter searchParam, stri
...
+            if (existingResource != null)
+            {
+                // Resource exists in Resource table, so it's a custom SearchParameter
+                // Allow the PUT operation and validate for code conflicts
+                CheckForConflictingCodeValue(searchParam, validationFailures);
+            }
+            else
+            {
+                // No resource in Resource table - this is a spec-defined SearchParameter
+                // Block PUT operations on spec-defined parameters with MethodNotAllowedException (405)
```
feordin, line 207: PUT can be an upsert operation where it creates a new resource.  Are we still allowing upsert via PUT?

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/DeleteSearchParameterBehavior.cs
jestradaMS, line 1: Why was modifying at this point also required? Was the validator not the only initial entry point?

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Parameters/SearchParameterValidator.cs
paorodma-ms: We are. This method is only when it finds a matching search parameter, and it wants to update it. Before calling this method, it will validate this case. Line 154.

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/DeleteSearchParameterBehavior.cs
paorodma-ms: No, the validator only checks the body of a search parameter in the request.

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/DeleteSearchParameterBehavior.cs
```diff
@@ -17,33 +20,63 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Parameters
...
         public async Task<TDeleteResourceResponse> Handle(TDeleteResourceRequest request, RequestHandlerDelegate<TDeleteResourceResponse> next, CancellationToken cancellationToken)
         {
             var deleteRequest = request as DeleteResourceRequest;
-            ResourceWrapper searchParamResource = null;
+            ResourceWrapper customSearchParamResource = null;
 
             if (deleteRequest.ResourceKey.ResourceType.Equals(KnownResourceTypes.SearchParameter, StringComparison.Ordinal))
             {
-                searchParamResource = await _fhirDataStore.GetAsync(deleteRequest.ResourceKey, cancellationToken);
+                customSearchParamResource = await _fhirDataStore.GetAsync(deleteRequest.ResourceKey, cancellationToken);
```
jestradaMS, line 48: As discussed, I am concerned with this approach as users could in theory create a param as "http://hl7.org/fhir/SearchParameter/" then perform a hard delete which removes from the table, but leaves it in PendingDelete and therefore it would have remained in def manager and not in resource table. 

To be more accurate, it might be better to use our JSON file that provides or Spec Defined parameters and if found in that file, we throw method not allowed.

paorodma-ms: Will abandon this PR and create a new one with these changes.

PR #5433, opened 2026-03-10
Atomic Search Param operations

jestradaMS:
## Description
This PR makes SearchParameter CRUD operations fully atomic and more reliable, especially for SQL Server-backed FHIR servers. It removes partial commit risks, reduces lock contention in transaction bundles, and makes the background refresh service the only owner of cache updates to prevent cache inconsistency.

It also improves diagnostics by adding clearer lifecycle logging, adds a dedicated retry policy for reindex operations to better handle transient SQL and Cosmos DB failures, and expands test coverage to verify that status updates no longer mutate the in-memory cache or publish mediator notifications.

## Related issues
Addresses [AB#183136](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/183136).

## Testing

## FHIR Team Checklist


docs/arch/adr-2603-atomic-searchparameter-crud-and-cache-refresh-ownership.md
```diff
@@ -0,0 +1,60 @@
...
+  A[SearchParameter Operation] -->|Single operation| B[Behavior queues pending status in request context];
+  B --> C[SqlServerFhirDataStore calls MergeSearchParams];
+  C --> D[MergeSearchParams: status + MergeResources in one transaction];
+  A -->|Transaction bundle| E[Per-entry: Behavior queues status, UpsertAsync calls MergeResources only];
+  E --> F[BundleHandler drains pending statuses after each entry];
+  F --> G[After all entries: single MergeSearchParams flush within outer transaction];
+  G --> H[Transaction commits atomically];
+```
+
+### 3. Cache update removal from CRUD path
```
SergeyGaluzo, line 37: Update search param workflow calls validate logic that triggers cache update, so, currently, cache update is run 2 times for an update workflow. How will proposed change work in regards of validate?

src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterStatusManager.cs
```diff
@@ -144,60 +144,32 @@ public async Task UpdateSearchParameterStatusAsync(IReadOnlyCollection<string> s
...
                     {
-                        throw;
-                    }
+                        Status = status,
+                        Uri = new Uri(uri),
+                    });
                 }
             }
 
             await _searchParameterStatusDataStore.UpsertStatuses(searchParameterStatusList, cancellationToken);
```
SergeyGaluzo, line 174: Can we use terminology consistent with SQL Server and call it Merge?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/MergeSearchParams.sql
SergeyGaluzo, line 57: We should not need this logic in new design as we are not writing to local cache anymore using these results. 
To make it clean I suggest to create new stored procedure MergeSearchParameters that is free from any old logic leftovers.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/108.diff.sql
```diff
@@ -0,0 +1,134 @@
+CREATE OR ALTER PROCEDURE dbo.MergeSearchParams @SearchParams dbo.SearchParamList READONLY
```
SergeyGaluzo, line 1: CREATE is not needed.

SergeyGaluzo, line 1: Please remove not used stored procedure UpsertSearchParamsWithOptimisticConcurrency

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/MergeSearchParams.sql
```diff
@@ -1,4 +1,23 @@
 ﻿CREATE PROCEDURE dbo.MergeSearchParams @SearchParams dbo.SearchParamList READONLY
+           ,@IsResourceChangeCaptureEnabled bit = 0
+           ,@TransactionId bigint = NULL
+           ,@SingleTransaction bit = 1
```
SergeyGaluzo: We cannot use eventual consistency in search param writes workflow, so having @SingleTransaction as input parameter is misleading. Please remove @SingleTransaction input and set its value to 1 on MergeResources call.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/MergeSearchParams.sql
```diff
@@ -21,11 +41,14 @@ BEGIN
 END
 
 DECLARE @SummaryOfChanges TABLE (Uri varchar(128) COLLATE Latin1_General_100_CS_AS NOT NULL, Operation varchar(20) NOT NULL)
+DECLARE @InitialTranCount int = @@trancount
 
 BEGIN TRY
-  SET TRANSACTION ISOLATION LEVEL SERIALIZABLE
-
-  BEGIN TRANSACTION
+  IF @InitialTranCount = 0
```
SergeyGaluzo: I assume that initial tran count here is present because you want to handle sequential bundles with multiple search param writes. We still need to set serializable transaction level to enable correct concurrency check. Please review and change.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
```diff
@@ -1165,6 +1293,7 @@ private async Task VerifySearchParameterIsWorkingAsync(
             // Report on any failures
             if (failedIds.Any())
             {
+                _output.WriteLine($"WARNING: {failedIds.Count} resources failed to create after {maxCreateRetries} retries");
```
SergeyGaluzo, line 1437: I like writing to output. Please consider removing System.Diagnostics.Debug.WriteLine calls.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
```diff
@@ -59,18 +168,20 @@ public async Task GivenReindexJobWithConcurrentUpdates_ThenReportedCountsAreLess
                 {
                     Parameter =
                     [
-                        new Parameters.ParameterComponent { Name = "maximumNumberOfResourcesPerQuery", Value = new Integer(1) },
+                        new Parameters.ParameterComponent { Name = "maximumNumberOfResourcesPerQuery", Value = new Integer(2) },
                         new Parameters.ParameterComponent { Name = "maximumNumberOfResourcesPerWrite", Value = new Integer(1) },
                     ],
                 };
 
+                await Task.Delay(5000);
```
SergeyGaluzo, line 322: We should be able to start reindex without any delays right after search param creates. It is normal workflow. Is this wait necessary?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
```diff
@@ -59,18 +168,20 @@ public async Task GivenReindexJobWithConcurrentUpdates_ThenReportedCountsAreLess
...
+
                 value = await _fixture.TestFhirClient.PostReindexJobAsync(parameters);
                 Assert.Equal(HttpStatusCode.Created, value.response.Response.StatusCode);
 
                 var tasks = new[]
                 {
                     WaitForJobCompletionAsync(value.jobUri, TimeSpan.FromSeconds(300)),
-                    RandomPersonUpdate(testResources),
+                    RandomPersonUpdate(testResources.Take(6).ToList()),
```
SergeyGaluzo, line 330: The goal for this test is to ensure that number of reported by reindex resources is less than original due to parallel updates. This condition should be true regardless of a particular number (as long as it is >=1) of updated resources. If my understanding is correct, please remove limiting logic.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
jestradaMS: This delay is not for reindex, it is to ensure the cache is updated before calling RandomPersonUpdate, otherwise it is racing and the person updates would happen before hashmap update, making them be reindexable still since they are on the old hashmap. This ensures that when we start reindex, then jump to parallel wait and update, we have correct hashmap for the Random updates to work as expected.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
jestradaMS: This provides a more accurate test result as I was seeing all updates happen and nothing reindexed due to the wait in reindex and fast update of 20 entries. This still produces the expected results consistently now.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/108.diff.sql
jestradaMS: This should be done in a different PR.

src/Microsoft.Health.Fhir.Core/Features/Search/Registry/SearchParameterStatusManager.cs
jestradaMS: This class has not been changed in this PR at this point and I feel this is clearer intent at application layer. Additionally, this is not specific to SQL as it has implementations in Cosmos still. Will leave as is for this PR and we can address seperately if necessary.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
SergeyGaluzo: The chance of "nothing reindexed due to the wait in reindex and fast update" can only increase with reduction of updated resources. I am not sure that introduced reduction helps this case. Please reevaluate.

SergeyGaluzo: The chance of not desired "nothing reindexed due to the wait in reindex and fast update" cannot decrease with reduction of updated resources. I am not sure that introduced reduction helps this case. Please reevaluate.

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkDelete/BulkDeleteProcessingJob.cs
```diff
@@ -78,6 +83,13 @@ public async Task<string> ExecuteAsync(JobInfo jobInfo, CancellationToken cancel
                 Exception exception = null;
                 List<string> types = definition.Type.SplitByOrSeparator().ToList();
 
+                if (types.Count > 0
+                    && string.Equals(types[0], KnownResourceTypes.SearchParameter, StringComparison.OrdinalIgnoreCase)
+                    && !(definition.ExcludedResourceTypes?.Any(x => string.Equals(x, KnownResourceTypes.SearchParameter, StringComparison.OrdinalIgnoreCase)) ?? false))
+                {
+                    await _searchParameterOperations.EnsureNoActiveReindexJobAsync(cancellationToken);
```
SergeyGaluzo, line 88: There are several problems with this implementation of "running reindex" check:

1. This check is not transacted with search param write, so there are always chances for racing - successful check, reindex started, search param write, resources are indexed incorrectly.
2. It is up to each workflow that allow search param writes to call this check. This increases chances that it is missed in code, and reindex is not protected.
 
Implementations that is free from the above problems is to move this check to MergeSearchParams stored procedure in the same SQL transaction that runs write, and rollback this transaction if check does not pass. For reindex orchestrator to bypass this check we can add its job id as input.

MergeSearchParam is already rewritten in this PR significantly, so it is right time to add above logic and get robust processing.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/108.diff.sql
SergeyGaluzo: @jestradaMS Yes, it can be done in a different PR. I just advocate for normal housekeeping practice that removes garbage in the area being affected. If we do separate PRs just for cleanup we will run out of versions :-)

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
jestradaMS: I think you should re-evaluate this test in main first. It is not working as you think it is.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/MergeSearchParams.sql
jestradaMS: Please take this up in a different PR

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/108.diff.sql
jestradaMS: Removed

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/MergeSearchParams.sql
jestradaMS: Updated

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/MergeSearchParams.sql
jestradaMS: updated

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
jestradaMS: Updated

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/MergeSearchParams.sql
SergeyGaluzo: @jestradaMS You are changing the code to stop using this output. Does it not make sense to clean it up right away?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
```diff
@@ -31,11 +35,133 @@ public class ReindexTests : IClassFixture<HttpIntegrationTestFixture>
...
 
-        public ReindexTests(HttpIntegrationTestFixture fixture)
+        public ReindexTests(HttpIntegrationTestFixture fixture, ITestOutputHelper output)
         {
             _fixture = fixture;
             _isSql = _fixture.DataStore == DataStore.SqlServer;
+            _output = output;
+        }
+
+        [SkippableFact(Skip = "Test is being flaky, needs investigation.")]
```
SergeyGaluzo: If bundle contains 10 search params this test consistently passes using the version from https://github.com/microsoft/fhir-server/pull/5435 PR for both Cosmos and SQL.
Instead of allowing to by-pass it, I would prefer to remove this test from this PR, and deal with this later.

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkDelete/BulkDeleteProcessingJob.cs
jestradaMS: This change closes the BulkDelete gap and ensures this path now checks for active reindex instead of bypassing the rule entirely. Moving the enforcement further down the stack is a larger follow-up: we would need to handle Cosmos correctly and define an internal override mechanism for flows like reindex, which must be able to update search parameter status while reindex is running. I think those extra considerations warrant a separate follow-up item rather than adding that additional improvement now as the PR is already pretty large. It is at least centralized in SearchParameterOperations which clarifies the intent and creates a pattern of centralized check.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
```diff
@@ -31,11 +35,133 @@ public class ReindexTests : IClassFixture<HttpIntegrationTestFixture>
...
+        public async Task Given500SearchParams_WhenReindexCompletes_ThenSearchParamsAreEnabled()
+        {
+            // Skip.IfNot(_isSql, "SQL Server only test. Cosmos DB is being deprecated and this scenario is validated on the SQL path.");
+
+            await CancelAnyRunningReindexJobsAsync();
+
+            int numberOfSearchParams = _isSql ? 500 : 100;
+            int searchParameterBatchSize = _isSql ? 100 : 10;
```
SergeyGaluzo: Original problem this test is supposed to simulate is to create burst of search parameters to be synchronized by incremental cache sync logic.
If we batch creates across several bundles this "burst" will be spread across multiple syncs, so it defeats the purpose of test.
Please remove batching.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
jestradaMS: removed

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
jestradaMS: removed the test

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
jestradaMS: removed it

docs/arch/adr-2603-atomic-searchparameter-crud-and-cache-refresh-ownership.md
jestradaMS: Currently these checks are still required due to the limitations in Cosmos. I did look to removing them and handle with Sql Deadlock / OptimisticConcurrency error handling and throw a 409, but Cosmos would still require these upper-level checks. For now, it's acceptable to ensure consistency at this layer until we are able to remove the Cosmos code path and move this all to SQL layer.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/MergeSearchParams.sql
jestradaMS: I did not remove what is consuming this output. It is used by the [FhirModel](https://github.com/microsoft/fhir-server/blob/46df45c24c093617e4ba131ac70eb1e2e265ca49/src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SqlServerSearchParameterStatusDataStore.cs#L225) which is not currently managed by SearchParameter background job. Changing this would have broader implications, hence if there is a desire to address FhirModel, this should be handled separately. This PR is addressing SearchParameterDefinitionManager updates and Atomic CUD.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/108.diff.sql
jestradaMS: I hear you, perhaps an option is to have a quarterly clean-up item which we use to track all clean-up from changes made within a quarter and at the start of the new quarter, we run a single schema update to remove old items? This way it's planned and more consistently followed?

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkDelete/BulkDeleteProcessingJob.cs
SergeyGaluzo: I agree that moving down the stack is more work, but it is **not** that more. 
For example, we do not need to deal with Cosmos in the same way as with SQL. Just add a separate check call in the corresponding Cosmos Data Store. Passing extra nullable job id parameter is not big either, especially if AI generates tests.
Please point me to the test added to check bulk delete and reindex coordination.

docs/arch/adr-2603-atomic-searchparameter-crud-and-cache-refresh-ownership.md
SergeyGaluzo: I think situation is **worse** than you describe :-) as any validation (does not matter with Cosmos or with SQL) requires cache being in sync. I don't think sync will go away after gen 1 is retired. 
This means, we need to add work item to ensure that sync via background worker and on demand sync required by validation can coexist without cache pollution.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/MergeSearchParams.sql
SergeyGaluzo: AFAIK the only thing that consumes this output is this line
_fhirModel.TryAddSearchParamIdToUriMapping(searchParamUri, searchParamId) 
in the UpsertStatusesInternal method.
This is **direct** local cache update which, I thought, you had the goal to avoid. 
Did I miss anything?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/108.diff.sql
SergeyGaluzo: :-) I can predict what will happen. We will never do it because we are always busy with something else. I was thinking about mentality to keep things clean. Especially when diff changes are already present, and it is a single line

IF object_id('UpsertSearchParamsWithOptimisticConcurrency') IS NOT NULL DROP PROCEDURE UpsertSearchParamsWithOptimisticConcurrency

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/MergeSearchParams.sql
```diff
@@ -21,11 +40,15 @@ BEGIN
 END
 
 DECLARE @SummaryOfChanges TABLE (Uri varchar(128) COLLATE Latin1_General_100_CS_AS NOT NULL, Operation varchar(20) NOT NULL)
+DECLARE @InitialTranCount int = @@trancount
 
 BEGIN TRY
   SET TRANSACTION ISOLATION LEVEL SERIALIZABLE
```
SergeyGaluzo, line 26: Execution of this line will throw exception if outer transaction is present. Please change,

SergeyGaluzo, line 26: Do we have tests that are testing outer transaction rollback in case of sequential parallel bundles with multiple search params?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
SergeyGaluzo: Good suggestion. One finding is that partitioning logic in orchestrator does not work as expected for max resources per query = 1. We will need to fix...

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/MergeSearchParams.sql
jestradaMS: Again, only cache dictionaries in scope were SearchParameterDefinitionManager dictionaries. Feel free to create a follow-up item.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/108.diff.sql
jestradaMS: I'll add the line

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkDelete/BulkDeleteProcessingJob.cs
jestradaMS: Feel free to create a follow-up item for this work. Below is the test for bulk delete reindex coordination.

src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/BulkDelete/BulkDeleteProcessingJobTests.cs

SergeyGaluzo: I will add repair item.
As far as tests, I expected e2e test that tests this functionality. Do we have one?

jestradaMS: Not e2e, I will add an e2e test to cover this.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/MergeSearchParams.sql
jestradaMS: No, will add integration tests to cover this.

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkDelete/BulkDeleteProcessingJob.cs
SergeyGaluzo: I added this task https://microsofthealth.visualstudio.com/Health/_workitems/edit/186796/

BTW Would we absolutely need e2e tests if above task was implemented? Could not we just rely on current exception trapping?

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexOrchestratorJob.cs
```diff
@@ -189,18 +187,14 @@ public async Task<string> ExecuteAsync(JobInfo jobInfo, CancellationToken cancel
...
         private async Task RefreshSearchParameterCache(bool isReindexStart)
         {
-            // before starting anything wait for natural cache refresh. this will also make sure that all processing pods have latest search param definitions.
+            // Wait for the background cache refresh service to complete N successful refresh cycles.
+            // This ensures all instances (including processing pods) have the latest search parameter definitions.
             var suffix = isReindexStart ? "Start" : "End";
             _logger.LogJobInformation(_jobInfo, $"Reindex orchestrator job started cache refresh at the {suffix}.");
             await TryLogEvent($"ReindexOrchestratorJob={_jobInfo.Id}.ExecuteAsync.{suffix}", "Warn", "Started", null, _cancellationToken);
-            await WaitForRefresh(); // wait for M * cache refresh intervals
+            await _searchParameterOperations.WaitForRefreshCyclesAsync(_operationsConfiguration.Reindex.CacheRefreshWaitMultiplier, _cancellationToken);
```
SergeyGaluzo: I am guessing the reason for replacing blind wait with this one is an attempt to account for load on the machines, when cache refresh cycles can run longer than expected. 

Please consider this scenario:
Orchestrator VM is not overloaded. Processes all 3 wait cycles, gets to creating jobs and completes this.
Processing VM is overloaded, cache is stale, but VM still writes resources.
When reindex item gets picked up by that "slow" VM, it is discovered that local hash/timestamp does not match requested by orchestrator.
At this moment attempt to wait for local cache refresh will help with processing resources registered by orchestrator, but it will miss resources written in the interval after registering jobs and moment when cache is fixed.

I mentioned my attempt to implement similar waits on one of scrums 1.5 weeks ago. My attempt was triggered by the changes I made to reduce RAM on FHIR server PR VMs to match PROD. This change made multiple tests flaky because of lack of CPU. I implemented similar to your changes but realized (see above) that it does not work for above scenario, so I removed the logic. To fix the tests I doubled search cache refresh interval. 

Please note that your waiting logic might help reindex tests, but it cannot guarantee missed reindexing caused by API calls.

In short, orchestrator cannot make decisions based on its own cache sync. The better solution is to wait for syncing of caches across **all** VMs before starting id range logic. This is also described here https://github.com/microsoft/fhir-server/pull/5449

jestradaMS: I agree, this implementation is not bullet proof. I have reviewed your ADR and agree with you; however, it does not align with Cosmos. Also, keep in mind that VerifyCountsAndUpdateReadySearchParameters is the stop gap for the scenario when a pod might write using the old definitions after range generation has started. 

I will evaluate what it would look like to implement your approach but leave this intact for Cosmos as it is still better than a non-deterministic wait.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexProcessingJob.cs
```diff
@@ -134,6 +134,25 @@ private async Task CheckDiscrepancies(CancellationToken cancellationToken)
             {
                 _logger.LogJobWarning(_jobInfo, msg);
                 await TryLogEvent($"ReindexProcessingJob={_jobInfo.Id}.GetResourcesToReindexAsync", "Error", msg, null, cancellationToken); // elevate in SQL to log w/o extra settings
+
+                // Wait for background cache refresh cycles to bring this host's cache up to date
+                _logger.LogJobInformation(_jobInfo, "Waiting for background cache refresh cycles to resolve hash mismatch...");
+                await _searchParameterOperations.WaitForRefreshCyclesAsync(3, cancellationToken);
```
SergeyGaluzo: If cache is not in sync at this point, writes via API are not be correctly indexed. Waiting at this point just hides the problem. I suggest removing wait.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexProcessingJob.cs
```diff
@@ -150,11 +169,13 @@ private async Task CheckDiscrepancies(CancellationToken cancellationToken)
             isBad = _reindexProcessingJobDefinition.SearchParamLastUpdated > currentDate;
             msg = $"SearchParamLastUpdated: Requested={requested} {(isBad ? ">" : "<=")} Current={current}";
             //// If timestamp from definition (requested by orchestrator) is more recent, then cache on processing VM is stale.
-            //// Cannot just refresh here because we might be missing resources updated via API.
             if (isBad)
             {
                 _logger.LogJobWarning(_jobInfo, msg);
                 await TryLogEvent($"ReindexProcessingJob={_jobInfo.Id}.ExecuteAsync", "Error", msg, null, cancellationToken); // elevate in SQL to log w/o extra settings
+
+                // Cache already refreshed via hash wait above, but timestamp may still lag.
```
SergeyGaluzo: Order of checks "hash then timestamp" is historical. I suggest changing to "timestamp then hash". This should be more in line with overall cache sync strategy we plan to have.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexOrchestratorJob.cs
SergeyGaluzo: I suggest leaving Cosmos as is with blind retries. We can increase wait cycles to 10-20. For SQL it should not matter because we will be waiting for the first cross VM sync event only. 
Also, I see that you started implementation. To avoid confusion, I outlined details which I had in mind in this user story https://microsofthealth.visualstudio.com/Health/_workitems/edit/186971

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/MergeSearchParams.sql
SergeyGaluzo: We discussed C# transaction in the search param workflow in scrum. Current thinking was to remove the need for that (see https://microsofthealth.visualstudio.com/Health/_workitems/edit/186795)
This PR removes C# transaction https://github.com/microsoft/fhir-server/pull/5455. 
This logic eliminates the need to handle outer transaction in the stored procedure and need for extra test cases. 
I suggest to base your PR on this logic.

SergeyGaluzo: You can remove outer transaction handling if we check in this PR https://github.com/microsoft/fhir-server/pull/5455 first.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -786,16 +793,28 @@ await MergeInternalAsync(
...
             var sw = Stopwatch.StartNew();
             using var cmd = new SqlCommand();
             //// Do not use auto generated tvp generator as it does not allow to skip compartment tvp and paramters with default values
             cmd.CommandType = CommandType.StoredProcedure;
-            cmd.CommandText = "dbo.MergeResources";
+            bool hasPendingStatuses = pendingStatuses?.Count > 0;
+
+            if (hasPendingStatuses)
+            {
+                cmd.CommandText = "dbo.MergeSearchParams";
```
SergeyGaluzo: AFAIK When bundle contains other than search param resources code will still call MergeSearchParams. It will not only make transactional changes to search param statuses and their resources but also apply changes to other resources. This makes MergeSearchParams name misleading, as it suggests that scope is limited to search params only. If we still want to use current logic, i.e. keep mixed (search param with other resource types) bundles, we should use different stored procedure name, something like MergeResourcesWithSearchParamStatuses.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/MergeSearchParams.sql
SergeyGaluzo: Fixed in https://github.com/microsoft/fhir-server/pull/5485

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/MergeSearchParams.sql
SergeyGaluzo: Fixed in https://github.com/microsoft/fhir-server/pull/5485

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SqlServerSearchParameterStatusDataStore.cs
```diff
@@ -188,6 +190,29 @@ private async Task UpsertStatusesInternal(IReadOnlyCollection<ResourceSearchPara
             cmd.CommandText = "dbo.MergeSearchParams";
             new SearchParamListTableValuedParameterDefinition("@SearchParams").AddParameter(cmd.Parameters, new SearchParamListRowGenerator().GenerateRows(statuses.ToList()));
 
+            if (_schemaInformation.Current >= 109)
+            {
+                cmd.Parameters.AddWithValue("@IsResourceChangeCaptureEnabled", false);
+                cmd.Parameters.Add(new SqlParameter("@TransactionId", SqlDbType.BigInt) { Value = DBNull.Value });
+
+                new ResourceListTableValuedParameterDefinition("@Resources").AddParameter(cmd.Parameters, Array.Empty<ResourceListRow>());
```
SergeyGaluzo: All entries with empty lists are not needed. Removed in https://github.com/microsoft/fhir-server/pull/5485

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SqlServerSearchParameterStatusDataStore.cs
```diff
@@ -221,5 +246,85 @@ public void SyncStatuses(IReadOnlyCollection<ResourceSearchParameterStatus> stat
                 _fhirModel.TryAddSearchParamIdToUriMapping(status.Uri.OriginalString, status.Id);
             }
         }
+
+        public async Task<CacheConsistencyResult> CheckCacheConsistencyAsync(string targetSearchParamLastUpdated, DateTime syncStartDate, DateTime activeHostsSince, CancellationToken cancellationToken)
```
SergeyGaluzo: Rewritten in https://github.com/microsoft/fhir-server/pull/5485

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
SergeyGaluzo: Added new MergeResourcesAndSearchParams stored procedure in https://github.com/microsoft/fhir-server/pull/5485

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -551,6 +560,44 @@ private async Task ExecuteTransactionForAllRequestsAsync(Hl7.Fhir.Model.Bundle r
...
+                statuses.Count == 0)
+            {
+                return;
+            }
+
+            _accumulatedPendingStatuses.AddRange(statuses);
+            context.Properties.Remove(SearchParameterRequestContextPropertyNames.PendingStatusUpdates);
+        }
+
+        private async Task FlushPendingSearchParameterStatusesAsync(CancellationToken cancellationToken)
```
fhibf, line 591: I would add a Boolean flag to the context of bundle handler, and just set it to true when a search param is present. 

then many some code don't event need to be executed. 

Per example, adding this validation on top of 'FlushPendingSearchParameterStatusesAsync' would not even call 'DrainPendingSearchParameterStatuses'.

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/SearchParameterState/SearchParameterStateUpdateHandler.cs
```diff
@@ -66,22 +61,14 @@ public async Task<SearchParameterStateUpdateResponse> Handle(SearchParameterStat
                 throw new UnauthorizedFhirActionException();
             }
 
-            if (await IsReindexRunningAsync(cancellationToken))
-            {
-                throw new PreconditionFailedException("A Reindex Job is currently running. Wait till it has completed before trying again.");
-            }
+            await _searchParameterOperations.EnsureNoActiveReindexJobAsync(cancellationToken);
```
fhibf, line 64: 'EnsureNoActiveReindexJobAsync' method is already called inside _searchParameterOperations.UpdateSearchParameterStatusAsync.

Are we keeping both to fail fast?

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/CreateOrUpdateSearchParameterBehavior.cs
```diff
@@ -19,25 +26,41 @@ namespace Microsoft.Health.Fhir.Core.Features.Search.Parameters
...
 
         public async Task<UpsertResourceResponse> Handle(CreateResourceRequest request, RequestHandlerDelegate<UpsertResourceResponse> next, CancellationToken cancellationToken)
         {
             if (request.Resource.InstanceType.Equals(KnownResourceTypes.SearchParameter, StringComparison.Ordinal))
             {
-                // Before committing the SearchParameter resource to the data store, add it to the SearchParameterDefinitionManager
-                // and parse the fhirPath, as well as validate the parameter type
-                await _searchParameterOperations.AddSearchParameterAsync(request.Resource.Instance, cancellationToken);
+                // Before committing the SearchParameter resource to the data store, validate the parameter type
+                await _searchParameterOperations.ValidateSearchParameterAsync(request.Resource.Instance, cancellationToken);
```
feordin, line 60: I do understand the goal of having only the background job update the in-memory cache of SearchParameters.  I want to double check a possible scenario of conflicting SearchParameters:
1) One instance processes SearchParamA with code=foo, and expression = resource._id
2) Second instance process SearchParamB with code=foo, and expression = resource.identifier, at almost the same time, inside the eventual consistency window

I believe the outcome is that when the background job runs to update in-memory cache, SearchParamB will win and we will not index SearchParamA,  A customer could search using foo, believing they are getting results for _id, but in reality they are getting resutls for identifier.

This is a very unlikely scenario, and could be characterized as a customer error, but is still bad.  

Is my analysis correct? Or do we have a protection against this that I missed?

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Reindex/ReindexSingleResourceRequestHandler.cs
```diff
@@ -74,8 +74,6 @@ public async Task<ReindexSingleResourceResponse> Handle(ReindexSingleResourceReq
                 throw new ResourceNotFoundException(string.Format(Core.Resources.ResourceNotFoundById, request.ResourceType, request.ResourceId));
             }
 
-            await _searchParameterOperations.GetAndApplySearchParameterUpdates(cancellationToken);
```
fhibf, line 77: Following up on my question on Teams: This method was ensuring the search params were in sync/memory, correct? 
 
Show we keep it to ensure consistency before executing the write below?

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/CreateOrUpdateSearchParameterBehavior.cs
SergeyGaluzo: We are not protected from this scenario today, and I don't think this PR has a goal to address it. I believe it can be solved only with eventual consistency based on max(LastUpdated).

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/SearchParameterState/SearchParameterStateUpdateHandler.cs
jestradaMS: correct, no sense in even doing any work if reindex is running at this point.

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/CreateOrUpdateSearchParameterBehavior.cs
feordin: I do agree that we are already exposed to this issue today, and it is an unlikely scenario.  I am fine addressing this in a future PR.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
jestradaMS: added!

src/Microsoft.Health.Fhir.Shared.Core/Features/Operations/Reindex/ReindexSingleResourceRequestHandler.cs
jestradaMS: You are correct. This was incorrectly removed. Added the check back.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -721,6 +788,8 @@ private async Task<EntryComponent> ExecuteRequestsWithSingleHttpVerbInSequenceAs
 
                         await resourceContext.Context.Handler.Invoke(httpContext);
 
+                        DrainPendingSearchParameterStatuses();
```
fhibf, line 791: We should probably add to "DrainPendingSearchParameterStatuses" a check if _containsSearchParameter if true or not. And avoid looping through the FHIR context properties if there is no need for.

SergeyGaluzo:
Creation of new search parameter in this PR should result in a single call to new MergeResourcesAndSearchParams stored procedure, which should write both definition and status. 

It is not what I see in some tests in ReindexJobTests class. I see separate calls to the MergeResources (writes search param definition) and calls to MergeResourcesAndSearchParams (writes status). This looks like an old behavior. 

2 questions:
1. Should we be testing with new behavior?
2. If answer on question 1 is no, which integration tests are testing new behavior?

SergeyGaluzo:
Do we use single SQL transaction for search param delete?

src/Microsoft.Health.Fhir.Core/Features/Context/BundleSqlTransactionContext.cs
```diff
@@ -0,0 +1,58 @@
...
+
+using System;
+using System.Threading;
+
+namespace Microsoft.Health.Fhir.Core.Features.Context
+{
+    /// <summary>
+    /// Default ambient bundle SQL transaction context implementation.
+    /// </summary>
+    public sealed class BundleSqlTransactionContext : IBundleSqlTransactionContext
```
SergeyGaluzo: What is the reason for this class if all transactional operations with search params in bundles are parallel, i.e. submitted to SQL in a single command, so no C# transactions needed?

SergeyGaluzo (changes requested):
I see an attempt to revive C# transactions. This contradicts with single SQL command restart ability approach implemented in the SQL retry class. As C# transactions approach has its drawbacks, I suggest discussing this prior to releasing. If we want to release this PR soon, I suggest moving C# transactions in a separate PR.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -529,8 +544,10 @@ private async Task ExecuteTransactionForAllRequestsAsync(Hl7.Fhir.Model.Bundle r
                 if (processingLogic == BundleProcessingLogic.Sequential)
                 {
                     using (var transaction = _transactionHandler.BeginTransaction())
```
fhibf, line 534: As we discussed yesterday, this logic can be simplified as the new logic forces Transactional Bundles with Search Params to be executed in parallel. 

I've shared the code for a test that ensures this logic is followed.

src/Microsoft.Health.Fhir.Core/Features/Context/BundleSqlTransactionContext.cs
fhibf: Just saw this comment. 
Yesterday John and I had a quick meeting around the same area.
Some components will be removed, and a test to ensure that transactional bundles with search parameters are set to be executed in parallel is being added.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SqlServerSearchParameterStatusDataStore.cs
SergeyGaluzo, line 145: I am confused. Maybe I don't understand the code. Looks like when optimistic conflict occurs, we just update input data to the latest last updated value and try to write again. If so, would not it work same way as "last write wins"? If it is desired, then should we just assign last updated in the stored proc without any concurrency control?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
jestradaMS: Removed this logic as we are no longer processing search parameters transactionally per a recent change.

src/Microsoft.Health.Fhir.Core/Features/Context/BundleSqlTransactionContext.cs
jestradaMS: Removed

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SqlServerSearchParameterStatusDataStore.cs
jestradaMS: This is pre-existing code. Please address separately than this PR as it is unmodified code.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
jestradaMS: Code was removed

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SqlServerSearchParameterStatusDataStore.cs
SergeyGaluzo: I am not asking to change in this PR. I just noticed this code and, as you added it, just wanted to ask.

jestradaMS:
> I see an attempt to revive C# transactions. This contradicts with single SQL command restart ability approach implemented in the SQL retry class. As C# transactions approach has its drawbacks, I suggest discussing this prior to releasing. If we want to release this PR soon, I suggest moving C# transactions in a separate PR.

Code was necessary to work properly when we were allowing search parameters be processed as transaction. Since we have recently updated to run parallel for all search parameters we no longer require this.

SergeyGaluzo (commented):
Thanks for removing C# transactions

SergeyGaluzo (approved):

build/jobs/provision-deploy.yml
```diff
@@ -68,7 +68,9 @@ jobs:
 
         $additionalProperties = $flattenedTestConfig
         $additionalProperties["TaskHosting__PollingFrequencyInSeconds"] = 1
-        $additionalProperties["FhirServer__CoreFeatures__SearchParameterCacheRefreshIntervalSeconds"] = 2
+        $additionalProperties["TaskHosting__MaxRunningTaskCount"] = "4"
+        $additionalProperties["FhirServer__CoreFeatures__SearchParameterCacheRefreshIntervalSeconds"] = 1
```
fhibf, line 72: Just double checking if the change on "SearchParameterCacheRefreshIntervalSeconds" from 2 to 1 is really planned, or if it was something done for testing purpose.

fhibf (approved):

build/jobs/provision-deploy.yml
jestradaMS: @SergeyGaluzo made this change to be more aggressive. I have another PR making this configurable so we can adjust CI so it runs like production vs. w/these more aggressive settings from PR builds.

SergeyGaluzo: I did that intentionally to intensify work on VMs in PR pipeline.

PR #5490, opened 2026-04-09
Fix cascading BundleOrch Exception log on parallel bundle failure

apurvabhaleMS:
## Description
In parallel bundle processing, if one worker fails (e.g., SQL timeout) the bundle operation transitions to Failed (or Canceled). The remaining parallel workers later call ReleaseResourceAsync().
Those calls could attempt further state transitions and end up throwing BundleOrchestratorException repeatedly—creating noisy cascading logs.

What the PR changes
Core fix: In BundleOrchestratorOperation.ReleaseResourceAsync(...), it now early-returns if the operation is already in a terminal state (Failed or Canceled).
This suppresses additional transitions/exceptions once the operation is already done.

## Related issues
Addresses [AB188320](https://microsofthealth.visualstudio.com/Health/_workitems/edit/188320)

## Testing

## FHIR Team Checklist


src/Microsoft.Health.Fhir.Core/Features/Persistence/Orchestration/BundleOrchestratorOperation.cs
```diff
@@ -444,6 +456,10 @@ private void SetStatusSafe(BundleOrchestratorOperationStatus suggestedStatus)
                 {
                     Status = suggestedStatus;
                 }
+                else if (Status == BundleOrchestratorOperationStatus.Failed || Status == BundleOrchestratorOperationStatus.Canceled)
```
fhibf: I would make this condition very specific to the case we saw in production. 
I would change this condition to: 

```
else if (suggestedStatus == BundleOrchestratorOperationStatus.WaitingForResources && (Status == BundleOrchestratorOperationStatus.Failed || Status == BundleOrchestratorOperationStatus.Canceled)
```

fhibf: That would help us catching the known scenarios, and avoid raising the BundleOrchestratorOperationCanceledException in more broad situations.

src/Microsoft.Health.Fhir.Core/Features/Persistence/Orchestration/BundleOrchestratorOperation.cs
```diff
@@ -240,6 +246,12 @@ public async Task ReleaseResourceAsync(string reason, CancellationToken cancella
 
                 InitializeMergeTaskSafe(dataStore: null, cancellationToken);
             }
+            catch (BundleOrchestratorOperationCanceledException)
```
fhibf, line 249: I believe that just returning (in case of error) makes sense, as in BundleHandler there is already a condition that says: "if (bundleOperation.Status != BundleOrchestratorOperationStatus.Completed)"

Bundle handler when calling ReleaseResourceAsync is already aware that something happened.

src/Microsoft.Health.Fhir.Core/Features/Persistence/Orchestration/BundleOrchestratorOperation.cs
```diff
@@ -135,6 +135,12 @@ public async Task<UpsertOutcome> AppendResourceAsync(ResourceWrapperOperation re
 
                 InitializeMergeTaskSafe(dataStore, cancellationToken);
             }
+            catch (BundleOrchestratorOperationCanceledException)
+            {
```
fhibf, line 139: Perfect!

fhibf (dismissed):

fhibf (approved):

PR #5491, opened 2026-04-10
Fix data store key comparison

fhibf:
## Description
SQL Data Store that can leave resources in a “full history mode” where the highest version available is set as “IsHistory: true”.

This behavior will lead to two problems:

Resources cannot be updated: as the service will try to recreate the combination of Resource Type, Resource ID and Version 1, and it’ll collide with the existing constraints.
* The resource does not have an active version, that will return HTTP404 when queried, unless the resource is queried explicitly for one of its historical versions.
* The following screenshot shows how the version system will look when this bug is present.

In this case, all versions of the record with ResourceTypeID 96 are set as “IsHistory: 1”, which indicates that those are all historical versions of the same record, not leaving an active version of the record.

The correct for this scenario would be the following, where the highest version is set as “IsHistory:0”.

Cases where this problem can happen:
* Import, when a blob has resources with different resource types sharing the same resource ID.
* Parallel bundles, when the bundle has resources with different resource types sharing the same resource ID.

---

The current plan includes:

Hotfix release to fix and stop the problem from happening with more resources.
An update to active the sleeping active version of the affected resources.

## Related issues
Addresses [AB#188729](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/188729)

## Testing

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
@@ -230,19 +230,19 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Storage
 
             var index = 0;
             var mergeWrappersWithVersions = new List<(MergeResourceWrapper Wrapper, bool KeepVersion, int ResourceVersion, int? ExistingVersion)>();
-            var prevResourceId = string.Empty;
+            ResourceKey prevResourceId = null;
             foreach (var resourceExt in resources) // if list contains more that one version per resource it must be sorted by id and last updated DESC.
             {
                 var metaHistory = true;
                 var resource = resourceExt.Wrapper;
-                var setAsHistory = prevResourceId == resource.ResourceId; // this assumes that first resource version is the latest one
+                var setAsHistory = prevResourceId == resource.ToResourceKey(true); // this assumes that first resource version is the latest one
                 //// negative versions are historical by definition
                 if (resourceExt.KeepVersion && int.Parse(resource.Version) < 0)
                 {
                     setAsHistory = true;
                 }
 
-                prevResourceId = resource.ResourceId;
+                prevResourceId = resource.ToResourceKey(true);
                 var weakETag = resourceExt.WeakETag;
                 int? eTag = weakETag == null
                     ? null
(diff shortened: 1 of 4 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -230,19 +230,19 @@ private async Task<MergeOutcome> MergeInternalAsync(IReadOnlyList<ResourceWrappe
...
-                var setAsHistory = prevResourceId == resource.ResourceId; // this assumes that first resource version is the latest one
+                var setAsHistory = prevResourceId == resource.ToResourceKey(true); // this assumes that first resource version is the latest one
                 //// negative versions are historical by definition
                 if (resourceExt.KeepVersion && int.Parse(resource.Version) < 0)
                 {
                     setAsHistory = true;
                 }
 
-                prevResourceId = resource.ResourceId;
+                prevResourceId = resource.ToResourceKey(true);
```
SergeyGaluzo: should be prevResourceKey

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -402,16 +402,16 @@ private async Task<MergeOutcome> MergeInternalAsync(IReadOnlyList<ResourceWrappe
             // Resources with keepVersion=true must be in separate call, and not mixed with keepVersion=false ones.
             // Sort them in groups by resource id and order by version.
             // In each group find the smallest version higher then existing
-            prevResourceId = string.Empty;
+            prevResourceId = null;
             var notSetInResoureGroup = false;
             foreach (var mergeWrapper in mergeWrappersWithVersions.Where(x => x.KeepVersion && x.ExistingVersion != 0).OrderBy(x => x.Wrapper.ResourceWrapper.ResourceId).ThenBy(x => x.ResourceVersion))
```
SergeyGaluzo: Please resolve order by

SergeyGaluzo (changes requested):
Please resolve order by

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
fhibf: Resolved.

LTA-Thinking (dismissed):

LTA-Thinking (approved):

SergeyGaluzo (approved):

Issue #5176, opened 2025-10-03
Fix AmbiguousMatchException between PatchJson and PatchFhir methods


brendankowitz:
## Problem

PATCH requests **without a Content-Type header or with an empty body** cause an `AmbiguousMatchException` and return HTTP 500.

### Error Details
```
Microsoft.AspNetCore.Routing.Matching.AmbiguousMatchException: 
The request matched multiple endpoints. Matches: 
  FhirController.PatchJson (Microsoft.Health.Fhir.Api)
  FhirController.PatchFhir (Microsoft.Health.Fhir.Api)
```

## Root Cause

`FhirController` has two PATCH methods with identical routes but different `Consumes` attributes:

**Method 1: PatchJson** (~line 484)
```csharp
[HttpPatch]
[Route(KnownRoutes.ResourceTypeById)]
[Consumes("application/json-patch+json")]
public async Task<IActionResult> PatchJson(
    [FromBody] JsonPatchDocument<Resource> patch,
    ...)
```

**Method 2: PatchFhir** (~line 531)
```csharp
[HttpPatch]
[Route(KnownRoutes.ResourceTypeById)]
[Consumes("application/fhir+json")]
public async Task<IActionResult> PatchFhir(
    [FromBody] Parameters parameters,
    ...)
```

ASP.NET Core evaluates route matching **before** applying `Consumes` constraints and before model binding. When a PATCH request arrives without a `Content-Type` header or with an empty body, the router cannot distinguish between these methods.

## Reproduction

```http
PATCH https://{{hostname}}/Patient/example
(No Content-Type header, no body)
```

**Result:** HTTP 500 with AmbiguousMatchException

## Impact

- Users receive HTTP 500 instead of proper validation error
- Occurs with testing tools (PostmanRuntime) or malformed requests
- Low frequency but poor user experience

## Recommended Solution

Mark `[FromBody]` parameters as required to fail early with a clear error instead of ambiguous match:

### Option 1: Use `EmptyBodyBehavior.Disallow` (Preferred)

```csharp
public async Task<IActionResult> PatchJson(
    [FromBody(EmptyBodyBehavior = EmptyBodyBehavior.Disallow)] JsonPatchDocument<Resource> patch,
    ...)

public async Task<IActionResult> PatchFhir(
    [FromBody(EmptyBodyBehavior = EmptyBodyBehavior.Disallow)] Parameters parameters,
    ...)
```

This approach:
[...]

PR #5503, opened 2026-04-16
Fix versioned canonical profile resolution

rohithkg-code:
## Description
Fix versioned canonical profile resolution for validation-backed profile lookup.

This change updates `ServerProvideProfileValidation.ResolveByCanonicalUriAsync` so that when a profile is referenced using a versioned canonical URI such as `http://example.org/fhir/StructureDefinition/custom-patient|3.0.0`, the resolver can match it against the stored `StructureDefinition` by canonical URL plus version.

Without this fallback, versioned profile references could fail resolution even when the corresponding `StructureDefinition` was present on the server.

This PR also adds focused unit tests covering:
- returning versioned supported profile URLs
- resolving a versioned canonical URI to the correct `StructureDefinition`

## Related issues
Addresses #5176.

## Testing
Code changes were validated by adding targeted unit tests in `ServerProvideProfileValidationTests` for:
- `GivenVersionedStructureDefinition_WhenGettingSupportedProfiles_ThenVersionedCanonicalIsReturned`
- `GivenVersionedStructureDefinition_WhenResolvingByCanonicalUriWithVersion_ThenMatchingProfileIsReturned`

Planned test command:
```powershell
dotnet test src\Microsoft.Health.Fhir.R4.Core.UnitTests\Microsoft.Health.Fhir.R4.Core.UnitTests.csproj --filter ServerProvideProfileValidationTests

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Validation/ServerProvideProfileValidationTests.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Validation/ServerProvideProfileValidationTests.cs
@@ -112,6 +112,38 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Validation
         }
 
         [Fact]
+        public async Task GivenVersionedStructureDefinition_WhenGettingSupportedProfiles_ThenVersionedCanonicalIsReturned()
+        {
+            // Arrange
+            var patientProfile = CreateStructureDefinition("http://example.org/fhir/StructureDefinition/custom-patient", "Patient", "3.0.0");
+            SetupSearchServiceWithResults("StructureDefinition", patientProfile);
+
+            // Act
+            var profiles = await _serverProvideProfileValidation.GetSupportedProfilesAsync("Patient", CancellationToken.None);
+
+            // Assert
+            Assert.Single(profiles);
+            Assert.Contains("http://example.org/fhir/StructureDefinition/custom-patient|3.0.0", profiles);
+        }
+
+        [Fact]
+        public async Task GivenVersionedStructureDefinition_WhenResolvingByCanonicalUriWithVersion_ThenMatchingProfileIsReturned()
+        {
+            // Arrange
+            var patientProfile = CreateStructureDefinition("http://example.org/fhir/StructureDefinition/custom-patient", "Patient", "3.0.0");
+            SetupSearchServiceWithResults("StructureDefinition", patientProfile);
+
+            // Act
+            var profile = await _serverProvideProfileValidation.ResolveByCanonicalUriAsync("http://example.org/fhir/StructureDefinition/custom-patient|3.0.0");
+
+            // Assert
+            Assert.NotNull(profile);
+            var structureDefinition = Assert.IsType<StructureDefinition>(profile);
+            Assert.Equal("http://example.org/fhir/StructureDefinition/custom-patient", structureDefinition.Url);
+            Assert.Equal("3.0.0", structureDefinition.Version);
+        }
+
+        [Fact]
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/Microsoft.Health.Fhir.Shared.Core/Features/Validation/ServerProvideProfileValidation.cs [hunk 1]
```diff
@@ -138,7 +138,18 @@ public void Dispose()
 
         public async Task<Resource> ResolveByCanonicalUriAsync(string uri)
         {
-            var summary = (await ListSummariesAsync(CancellationToken.None)).ResolveByCanonicalUri(uri);
+            var summaries = (await ListSummariesAsync(CancellationToken.None)).ToList();
```
feordin, line 141: Thanks for the contribution!

I have a question on this method: 
It appears that this fallback only works correctly if at most one version of each StructureDefinition exists on the server.

_summaries is built in GetSummariesAsync using result[artifact.ResourceUri] = artifact — keyed on the base canonical URL with no version. When a server holds both v2.0.0 and v3.0.0 of the same SD, only the last one processed survives in _summaries, so url|2.0.0 will silently fail to resolve if v3.0.0 was stored last.

I think to fully support multi-version scenarios, GetSummariesAsync needs to key the dictionary by url + version (e.g. $"{artifact.ResourceUri}|{ver}"), and LoadBySummary's cache key needs the same treatment (see separate comment).

src/Microsoft.Health.Fhir.Shared.Core/Features/Validation/ServerProvideProfileValidation.cs
```diff
@@ -138,7 +138,18 @@ public void Dispose()
...
+            if (summary == null &&
+                TrySplitVersionedCanonicalUri(uri, out string canonicalUri, out string version))
+            {
+                summary = summaries.FirstOrDefault(x =>
+                    string.Equals(x.ResourceUri, canonicalUri, StringComparison.OrdinalIgnoreCase) &&
+                    x.TryGetValue(_structureDefinitionVersionKey, out object summaryVersion) &&
+                    string.Equals(summaryVersion?.ToString(), version, StringComparison.OrdinalIgnoreCase));
+            }
+
             return LoadBySummary(summary);
```
feordin, line 153: I think, if we fix the first comment I made, we will have a new issue: 
The _resourcesByUri cache key doesn't include version, so it will return the wrong resource once multi-version support is added.

Steps to reproduce (after fixing the deduplication bug above):

ResolveByCanonicalUriAsync("url|v2") → loads v2 from JSON → stores in cache under key url
ResolveByCanonicalUriAsync("url|v3") → cache hit on url → returns v2 incorrectly

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Validation/ServerProvideProfileValidationTests.cs
```diff
@@ -111,6 +111,38 @@ public async Task GivenStructureDefinitionsExist_WhenGettingSupportedProfiles_Th
             Assert.Contains("http://example.org/fhir/StructureDefinition/custom-patient", profiles);
         }
 
+        [Fact]
```
feordin, line 114: More possible test cases:
Version mismatch returns null — call ResolveByCanonicalUriAsync("url|9.9.9") when only url|3.0.0 exists; assert the result is null. This ensures the fallback doesn't match on URL alone.

Unversioned URI still resolves — regression guard that an unversioned canonical URI resolves correctly after the restructured code path in ResolveByCanonicalUriAsync.

Two versions of the same SD, each resolves correctly — a test with both v2.0.0 and v3.0.0 of the same SD on the server, verifying each versioned URI resolves to the right resource. This will also expose the deduplication/cache bugs noted above.

PR #5522, opened 2026-04-24
Add SQL Elastic Pool and tweak CI env vars

jestradaMS:
## Description
This pull request introduces support for deploying Azure SQL Elastic Pools in the CI pipeline and enhances the configuration and reliability of deployment scripts, and tests. The changes improve resource management for SQL-backed environments, make deployment parameters more flexible, and address an issue with orphaned test databases during integration testing.

Adds log analytics for ACAs for historical telemetry for troubleshooting runtime issues.

It also ensures that application settings are appropriate for CI environment with flexibility of more aggressive settings in a PR and combines all e2e tests back into a single job for SQL.

## Related issues
Addresses [AB#182377](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/182377).

## Testing

## FHIR Team Checklist


build/ci-pipeline.yml
LTA-Thinking, line 177: Should the tests wait on the scale up?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
LTA-Thinking, line 395: I don't like tests that rely so heavily on good timing, but I guess it is working already.

build/ci-pipeline.yml
jestradaMS: Scale up happens at the very start of the pipeline and is done within the first few minutes. By the time we get to first set of test we should be scaled up. Didn't want to make it a hard dependency. 

Here's an example run. 

https://microsofthealthoss.visualstudio.com/FhirServer/_build/results?buildId=49097&view=results

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
jestradaMS: Me either and this particular test is still flakey. It usually succeeds on a rerun. I'm not sure it's really needed anymore. We can reevaluate it.

build/jobs/provision-deploy.yml
```diff
@@ -31,6 +31,36 @@ parameters:
 - name: minReplicas
   type: number
   default: 0
+- name: maxReplicas
```
mikaelweave: NIT - would these be better served in a shared variable file vs passed as parameters and repeated?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
```diff
@@ -321,18 +334,44 @@ public async Task GivenReindexJobWithConcurrentUpdates_ThenReportedCountsAreLess
...
-                var tasks = new[]
+                // Drive concurrent updates as a continuous loop instead of a single one-shot batch. With the
+                // orchestrator's SearchParameter cache convergence wait set to several seconds (and longer in CI),
+                // a one-shot batch of 6 PUTs completes well before reindex starts scanning, so the updated rows'
+                // new surrogate IDs are still inside the reindex snapshot's upper-bound and nothing gets skipped.
+                // The loop guarantees at least some updates land while the scan is active.
+                using var updatesCts = new CancellationTokenSource();
+                var updateLoop = RandomPersonUpdateLoop(testResources.Take(6).ToList(), updatesCts.Token);
+
+                try
```
mikaelweave, line 441: Should we add a catch and a basic log of the exception just in case?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/BulkDeleteTests.cs
```diff
@@ -489,48 +490,40 @@ public async Task GivenBulkDeleteRequestWithMultipleExcludedResourceTypes_WhenCo
         [Trait(Traits.Category, Categories.IndexAndReindex)] // this moves tests to reindex group to avoid racing failures
```
mikaelweave: Should we keep this trait on the test now?

build/ci-pipeline.yml
```diff
@@ -288,6 +302,25 @@ stages:
...
+  dependsOn:
+  - testStu3Cosmos
+  - testStu3Sql
+  - testR4Cosmos
+  - testR4Sql
+  - testR5Sql
+  # Only scale down on success. On failure we leave the apps at min=3 so developers can investigate
+  # via the live endpoints and ADO "Rerun failed stages" (which does NOT re-execute scaleUpContainerApps)
+  # doesn't hit cold-started apps mid-test.
+  condition: succeeded()
```
mikaelweave, line 316: Do we have a cleanup pipeline to scale these down on failure?

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlServerFhirStorageTestHelper.cs
```diff
@@ -221,27 +221,37 @@ public async Task DeleteDatabase(string databaseName, CancellationToken cancella
         {
             try
             {
-                await DbSetupSemaphore.WaitAsync(cancellationToken);
-                try
+                await _dbSetupRetryPolicy.ExecuteAsync(async () =>
```
mikaelweave, line 225: Optimization suggestion

`_dbSetupRetryPolicy` catches `OperationCanceledException` then re-throws.

The retry policy is `Policy.Handle<Exception>()` — it catches `OperationCanceledException` and retries 5× at 3s each before re-throwing. Now that `DeleteDatabase` actually re-throws (instead of swallowing), a timeout-triggered cancellation during test teardown adds ~15s of overhead and surfaces a secondary exception in xUnit that may obscure the real test failure. 

Consider `Policy.Handle<SqlException>().Or<TimeoutException>()` so `OperationCanceledException` flows through.

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Features/Operations/Reindex/ReindexJobTests.cs
```diff
@@ -1411,7 +1411,9 @@ await _searchParameterStatusManager.UpdateSearchParameterStatusAsync(
                 SearchParameterStatus.Supported,
                 CancellationToken.None);
 
-            await _searchParameterOperations.GetAndApplySearchParameterUpdates(CancellationToken.None, true);
+            // These tests start a background cache refresh loop, so a zero-wait refresh can
+            // skip the one apply that populates the SQL search-parameter URI->ID mapping.
+            await _searchParameterOperations.GetAndApplySearchParameterUpdates(CancellationToken.None);
```
mikaelweave, line 1416: Changing this means this method can throw an exception now - will this cause the whole test process to crash?

build/jobs/run-sql-tests.yml
```diff
@@ -93,38 +93,7 @@ jobs:
...
-    parameters:
-      version: ${{ parameters.version }}
-      containerAppName: '${{ parameters.containerAppName }}'
-      appServiceType: 'SqlServer'
-      categoryFilter: 'Category=IndexAndReindex'
-      testRunTitleSuffix: ' Reindex'
-
-- job: 'sqlE2eTests_BulkUpdate'
-  timeoutInMinutes: 75
+  timeoutInMinutes: 120
```
mikaelweave, line 96: Should we be waiting 120 minutes for these tests? They used to run in 1 hours now it looks like 2 hours. Will this slow down our hotfix speed / release speed?

mikaelweave, line 96: SQL Tests now take 2 hours vs 1 in CI!?

jestradaMS: They still run in ~1-1.25 hours. The increase is more of a cushion than an expectation, but the biggest factor is that reindex tests will creep up in time as the db size grows, and as we have combined them in the CI in this PR, we should assume CI will take longer over time. 

This should not impact general release regardless as long as everything is running as expected. As for hotfixes, we shouldn't / wouldn't want to run CI for hotfix.

jestradaMS: See response in previous comment..

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Features/Operations/Reindex/ReindexJobTests.cs
jestradaMS: It will not cause the entire test process to crash. In fact, this fix is needed to prevent that from happening in this test case.

build/ci-pipeline.yml
jestradaMS: I considered this as well as having it run regardless of succeeding or failing. I felt since expectation is CI does not fail, this was the lessor of the two evils as it will ensure we never accidently run with less than 3 and don't need another pipeline to track if we were to break it out.

build/jobs/provision-deploy.yml
jestradaMS: Moved them into pr-variables and ci-variables.

build/ci-pipeline.yml
mikaelweave: makes sese for ci as is, thanks

build/jobs/run-sql-tests.yml
mikaelweave: sure - we can observe and optimize if needed

mikaelweave (approved):

PR #5525, opened 2026-04-27
Disable query plan reuse based on stats

LTA-Thinking:
## Description
When a search parameter has skewed stats in the SQL DB the query plan generated for one value of the search parameter can be a bad plan to use for a different value of the same search parameter. For this reason query plans should not be reused on search parameters like this. 

This PR introduces a class that checks the skew of a search parameter and disables query plan reuse if it detects the skew is high.

This also adds a FHIR SQL configuration to control query plan reuse settings. This moves the existing setting from the DB's parameters table to the appsettings file for easier control.

## Related issues
Addresses [User Story 188612](https://microsofthealth.visualstudio.com/Health/_workitems/edit/188612)

## Testing
New E2E and unit tests

## FHIR Team Checklist


src/Microsoft.Health.Fhir.SqlServer/Features/Search/QueryPlanReuseChecker.cs
```diff
@@ -0,0 +1,175 @@
...
+                _skewedParameters = new List<IGrouping<string, (string Uri, string ResourceTypeId)>>();
+                _isInitialized = true;
+                return;
+            }
+
+            var searchParamIds = results.Where(r => !string.IsNullOrEmpty(r.ResourceTypeId) && !string.IsNullOrEmpty(r.SearchParamId)).Select(r => r.SearchParamId).Distinct().Aggregate((a, b) => a + "," + b);
+
+            using var cmd2 = new SqlCommand();
+            cmd2.CommandType = CommandType.Text;
+            cmd2.CommandText = "SELECT SearchParamId, Uri FROM SearchParam WHERE SearchParamId IN (@searchParamIds)";
```
mikaelweave: Should we be using a TVP here for the parameter? Or build the multiple parameters `p0, p1` dynamically?

Not sure SQL will expand multiple parameters.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/112.diff.sql
```diff
@@ -0,0 +1,73 @@
...
+
+	PRINT @stat_name + ' ' + @table_name;
+
+	INSERT @stats_histogram
+	EXEC('DBCC SHOW_STATISTICS(''' + @table_name + ''',' + @stat_name + ') WITH HISTOGRAM');
+
+	SELECT @min_rows = MIN(eq_rows), @max_rows = MAX(eq_rows)
+	FROM @stats_histogram;
+
+	INSERT @results VALUES(@stat_name, @table_name, @max_rows/@min_rows);
```
mikaelweave: Is it possible that `@min_rows` is zero here? Even if remotely possible we should consider handling this.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/QueryPlanReuseChecker.cs
```diff
@@ -0,0 +1,175 @@
...
+        private readonly double _skewThreshold = 30;
+
+        // Holds a list of urls for skewed search parameters. If the search parameters are skewed, the query plan should not be reused.
+        private List<IGrouping<string, (string Uri, string ResourceTypeId)>> _skewedParameters = new List<IGrouping<string, (string Uri, string ResourceTypeId)>>();
+        private readonly ISqlRetryService _sqlRetryService;
+        private readonly FhirTimer _timer;
+        private readonly ILogger<QueryPlanReuseChecker> _logger;
+
+        private bool _isInitialized = false;
+        private bool _storageReady = false;
```
mikaelweave, line 43: Do we need thread safety on these since they are written by the background timer and read by the request-handling threads?

Same for `_skewedParameters`

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/QueryPlanReuseCheckerTests.cs
```diff
@@ -0,0 +1,405 @@
...
+            // Assert
+            Assert.True(result);
+            _output.WriteLine("CanReuseQueryPlan returned true for empty search parameters.");
+        }
+
+        /* These tests are currently not working because the test fixture isn't populating search parameters in the database, which causes the statistics creation to fail.
+         * We need to either update the fixture to include search parameter population or mock the database interactions to properly test the cache loading logic.
+         *
+        [Fact]
+        public async Task GivenSkewedStatisticsInDatabase_WhenCacheIsRefreshed_ThenSkewedParametersAreIdentified()
```
mikaelweave, line 124: Should we Skip these and add a work item vs commenting out and committing?

src/Microsoft.Health.Fhir.SqlServer/Features/Search/QueryPlanReuseChecker.cs
```diff
@@ -0,0 +1,175 @@
...
+using System.Data;
+using System.Linq;
+using System.Security.Policy;
+using System.Text;
+using System.Text.RegularExpressions;
+using System.Threading;
+using System.Threading.Tasks;
+using EnsureThat;
+using MediatR;
+using Microsoft.Build.Framework;
```
mikaelweave: NIT Do we need `Microsoft.Build.Framework` and `System.Security.Policy`?

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/SearchOptionsFactory.cs
```diff
@@ -12,6 +12,7 @@
 using Hl7.Fhir.Model;
 using Hl7.Fhir.Rest;
 using Hl7.Fhir.Utility;
+using Microsoft.Extensions.Azure;
```
mikaelweave: nit - do we need this new import

src/Microsoft.Health.Fhir.SqlServer/Features/Search/QueryPlanReuseChecker.cs
```diff
@@ -0,0 +1,175 @@
...
+using Microsoft.Health.Fhir.SqlServer.Features.Search.Expressions.Visitors.QueryGenerators;
+using Microsoft.Health.Fhir.SqlServer.Features.Storage;
+using Microsoft.Health.Fhir.SqlServer.Features.Watchdogs;
+
+namespace Microsoft.Health.Fhir.SqlServer.Features.Search
+{
+    internal class QueryPlanReuseChecker : IQueryPlanReuseChecker, INotificationHandler<SearchParametersInitializedNotification>
+    {
+        private readonly Regex _skewedParameterRegex = new Regex(@"^ST_.*_WHERE_ResourceTypeId_(\d*)_SearchParamId_(\d*)$");
+        private readonly double _refreshPeriod = 3600;
```
mikaelweave: Is there any reason we should make these numbers configurable like `ReuseQueryPlans.IsEnabled`?

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/112.diff.sql
```diff
@@ -0,0 +1,73 @@
...
+
+WHILE @i < @count
+BEGIN
+	SELECT @stat_name = stat_name, @table_name = table_name
+	FROM @stat_names
+	ORDER BY table_name, stat_name
+	OFFSET @i ROWS
+	FETCH NEXT 1 ROWS ONLY;
+
+	PRINT @stat_name + ' ' + @table_name;
```
mikaelweave: Is this a debug statement that was left in? If not close comment

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/FhirStorageTestsFixture.cs
```diff
@@ -97,7 +97,7 @@ internal FhirStorageTestsFixture(IServiceProvider fixture)
                 case CosmosDbFhirStorageTestsFixture _:
                     _fhirRuntimeConfiguration = new AzureApiForFhirRuntimeConfiguration();
                     break;
-                case SqlServerFhirStorageTestsFixture _:
+                case SqlServerFhirStorageTestsFixture sqlFixture:
```
mikaelweave: +1 on copilot here

src/Microsoft.Health.Fhir.SqlServer/Features/Search/QueryPlanReuseChecker.cs
```diff
@@ -0,0 +1,175 @@
...
+        {
+            if (!_isInitialized)
+            {
+                return true;
+            }
+
+            // Check the skew of the search parameters. If the search parameters are skewed, the query plan should not be reused.
+            var parameters = searchOptions.SearchParameters;
+
+            foreach (var parameter in parameters.Where(p => _skewedParameters.Any(skew => skew.Key == p.Url.OriginalString)))
```
mikaelweave: nit - does the `for` on the `.Any(skew => skew.Key == ...)` mean `O(n×m)` per search? If so, consider building a `HashSet<string>` during refresh and doing `O(1)` lookups.

src/Microsoft.Health.Fhir.Core/Features/Search/SearchOptions.cs
```diff
@@ -140,6 +141,8 @@ internal set
         /// </summary>
         public Expression Expression { get; internal set; }
 
+        public IReadOnlyList<SearchParameterInfo> SearchParameters { get; internal set; } = new List<SearchParameterInfo>();
```
mikaelweave, line 147: nit - add xml comment

src/Microsoft.Health.Fhir.SqlServer/Features/Search/QueryPlanReuseChecker.cs
LTA-Thinking: The boolean's should be safe because they are atomic operations, their reads and writes can't be interrupted.
For _skewedParameters there should be no risk as it is never updated, the list is just replaced. So even if the old version is being read when a new version is made the old version should still be used because the new version doesn't alter the memory locations of the old version.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/QueryPlanReuseChecker.cs
LTA-Thinking: Replaced with inline parameter.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Migrations/112.diff.sql
LTA-Thinking: Added check

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/QueryPlanReuseCheckerTests.cs
LTA-Thinking: The issue is that they don't even compile right now. I don't want to put more work into them and hold up the story, but I also don't want to lose the work I've already done.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/QueryPlanReuseChecker.cs
LTA-Thinking: Added config file

mikaelweave (dismissed):

mikaelweave (dismissed):

mikaelweave (approved):

PR #5530, opened 2026-04-29
Return 404 for mis-cased FHIR resource type segments

mikaelweave:
## Description

This PR fixes two related case-sensitivity bugs surfaced by requests like
`GET /patient/{id}` (lowercase resource type):

1. **405 instead of 404 for mis-cased resource types.** - not fhir spec
2. **Capability cache poisoning.**  - once you go 405, you can't run a valid query for the resource
3. 
### Bug 1: mis-cased resource type returns 405
1. Start the server.
2. Issue `GET https://localhost:44348/patient/123` (lowercase) — or `PUT`, `DELETE`, `POST`.
4. Current behavior: **405**.
5. Expected behavior: **404**.

### Bug 2: lowercase request poisons later correctly-cased requests
1. Start the server.
2. Issue `DELETE https://localhost:44348/patient/123`.
3. Issue `DELETE https://localhost:44348/Patient/123`.
4. Current behavior: the second request can return **405** from
   `ValidateCapabilityPreProcessor` and continues to do so for the lifetime of the process.
6. Expected behavior: only the lowercase request should fail.

## Related Work Items

[AB#190027](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/190027)

### DIFF
--- a/src/Microsoft.Health.Fhir.Api/Features/Routing/ResourceTypesRouteConstraint.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/Routing/ResourceTypesRouteConstraint.cs
@@ -21,7 +21,7 @@ namespace Microsoft.Health.Fhir.Api.Features.Routing
 
             if (values.TryGetValue(KnownActionParameterNames.ResourceType, out var resourceTypeObj) && resourceTypeObj is string resourceType && !string.IsNullOrEmpty(resourceType))
             {
-                return ModelInfoProvider.IsKnownResource(resourceType);
+                return ModelInfoProvider.IsKnownResource(resourceType, ignoreCase: false);
             }
 
             return false;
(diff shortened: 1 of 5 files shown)


src/Microsoft.Health.Fhir.Api/Features/Routing/ResourceTypesRouteConstraint.cs
```diff
@@ -21,7 +21,7 @@ public bool Match(HttpContext httpContext, IRouter route, string routeKey, Route
 
             if (values.TryGetValue(KnownActionParameterNames.ResourceType, out var resourceTypeObj) && resourceTypeObj is string resourceType && !string.IsNullOrEmpty(resourceType))
             {
-                return ModelInfoProvider.IsKnownResource(resourceType);
+                return ModelInfoProvider.IsKnownResource(resourceType, ignoreCase: false);
```
feordin, line 24: Just to confirm, we don't support mis-cased resources right now, and won't after your PR, we're just changing the type of error response?

mikaelweave: Correct! We just want to return a 404 vs a 405. While returning a 405 vs a 404 is a minor change, the cache poisioning is the larger issue.

`Resource and Element names are case-sensitive (though duplicates that differ only in case are never defined)`

https://build.fhir.org/formats.html

feordin (approved):

PR #5550, opened 2026-05-05
Fix startup memory pressure

LTA-Thinking:
## Description
Removes memory pressure on startup happening when the database has a large number of systems or quantity codes.

## Related issues
Addresses [Bug 191287](https://microsofthealth.visualstudio.com/Health/_workitems/edit/191287)

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
@@ -242,6 +242,11 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Storage
             string commaSeparatedClaimTypes = string.Join(',', _securityConfiguration.PrincipalClaims);
             string commaSeparatedCompartmentTypes = string.Join(',', ModelInfoProvider.GetCompartmentTypeNames());
 
+            _systemToId = new FhirMemoryCache<int>("systemToId", _logger, ignoreCase: true);
+            _quantityCodeToId = new FhirMemoryCache<int>("quantityCodeToId", _logger, ignoreCase: true);
+            bool systemWarningLogged = false;
+            bool quantityCodeWarningLogged = false;
+
             using var cmd = new SqlCommand();
             cmd.CommandType = CommandType.StoredProcedure;
             cmd.CommandText = "dbo.InitializeBase";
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
fhibf, line 382: Could you add a logic to clean up resultsList at the end of this method? 

My impression is that, as this list is becoming big in memory, GC is having a bad time deciding if it has to clean it up.

The reason is: from what we see in logs, the 5.5GB are elements hanging on resultsList, as Memory Cache objects are limited to 50MB.

LTA-Thinking: I don't know if we can do it cleanly. The result list is a read only list of read only lists, there isn't a way to clear it.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
```diff
@@ -273,11 +278,41 @@ private async Task InitializeBase(CancellationToken cancellationToken)
...
+
+                        var (value, quantityCodeId) = reader.ReadRow(VLatest.QuantityCode.Value, VLatest.QuantityCode.QuantityCodeId);
+
+                        if (!_quantityCodeToId.TryAdd(value, quantityCodeId) && !quantityCodeWarningLogged)
+                        {
+                            _logger.LogWarning($"Cache '{_quantityCodeToId.Name}' reached the limit of {_quantityCodeToId.CacheMemoryLimit} bytes (with {_quantityCodeToId.Count} cached elements).");
+                            quantityCodeWarningLogged = true;
+                        }
+
+                        return null;
```
fhibf, line 315: I think that, as we'll return null for all the cases (in System and QuantityCode), we should change _SqlRetryService.ExecuteMultiResultReaderAsync_ to ignore null values and do not add them to the list, otherwise it'll cause memory fragmentation, as the List will continue to grow to huge sizes (500k+ elements), even with null values.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirModel.cs
fhibf: That should be fine with the current changes.

fhibf (approved):

PR #5560, opened 2026-05-11
Use DB time for background job time limit

LTA-Thinking:
## Description
Some background jobs have a clause to prevent them working on data created after they were queued. This PR changes the source of the time to be the DB set create time instead of it being set by the server when the job is queued. This prevents mismatches between the two sources that could lead to resources being missed if they were created very soon before a job was enqueued.

## Related issues
Addresses [Bug 191061](https://microsofthealth.visualstudio.com/Health/_workitems/edit/191061)

## Testing
Existing E2E tests

## FHIR Team Checklist


src/Microsoft.Health.Fhir.Core/Features/Operations/BulkDelete/BulkDeleteOrchestratorJob.cs
```diff
@@ -65,21 +66,25 @@ public async Task<string> ExecuteAsync(JobInfo jobInfo, CancellationToken cancel
...
         // Creates a bulk delete processing job.
         // Each processing job only deletes one resource type, but it contains a comma seperated list of all resource types to be deleted. Once one type is deleted it will start a new job to delete the next one.
-        internal static async Task<BulkDeleteDefinition> CreateProcessingDefinition(BulkDeleteDefinition baseDefinition, ISearchService searchService, IList<string> resourceTypes, CancellationToken cancellationToken)
+        internal static async Task<BulkDeleteDefinition> CreateProcessingDefinition(JobInfo jobInfo, BulkDeleteDefinition baseDefinition, ISearchService searchService, IList<string> resourceTypes, CancellationToken cancellationToken)
         {
+            var createDate = new PartialDateTime(new DateTimeOffset(jobInfo.CreateDate, TimeSpan.Zero));
             var searchParameters = new List<Tuple<string, string>>()
                 {
-                    new Tuple<string, string>(KnownQueryParameterNames.Summary, "count"),
+                    new Tuple<string, string>(KnownQueryParameterNames.LastUpdated, $"lt{createDate}"),
```
rbans96, line 74: This uses `lt` but BulkUpdate's equivalent uses `le`. The original code used `lt` for both. Was switching BulkUpdate to `le` intentional? As-is, a resource written at the exact same instant as `CreateDate` gets skipped by delete but included by update.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/Models/ExportJobRecord.cs
```diff
@@ -97,7 +97,8 @@ public ExportJobRecord(
             Status = OperationStatus.Queued;
 
             QueuedTime = Clock.UtcNow;
-            Till = till ?? new PartialDateTime(Clock.UtcNow);
+            Till = till;
```
rbans96: `Till` can now be null between job creation and orchestrator execution (previously it always had a value via the `?? Clock.UtcNow` fallback). Any code that reads `Till` before the orchestrator runs would hit a null. Worth checking nothing does, or keeping a default.

src/Microsoft.Health.Fhir.CosmosDb/Features/Operations/Export/CosmosExportOrchestratorJob.cs
```diff
@@ -43,6 +44,12 @@ public async Task<string> ExecuteAsync(JobInfo jobInfo, CancellationToken cancel
             var record = jobInfo.DeserializeDefinition<ExportJobRecord>();
             record.QueuedTime = jobInfo.CreateDate; // get record of truth
 
+            // If Till was not explicitly set by the user, use the job's CreateDate
+            if (!record.IsTillExplicit)
```
rbans96: nit: Same block exists in `SqlExportOrchestratorJob.cs`. Maybe extract to a shared helper to avoid drift.

rbans96 (commented):
Nice fix ? using DB time eliminates the clock skew issue cleanly. The `Task.Delay` removals in E2E tests confirm it works. A couple things to consider:

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkDelete/BulkDeleteOrchestratorJob.cs
LTA-Thinking: This was part of investigating an issue, I'll revert it.

src/Microsoft.Health.Fhir.Core/Features/Operations/Export/Models/ExportJobRecord.cs
LTA-Thinking: The issue with keeping a default is we still want the user to be able to manually set it, and if there is a default there isn't a way to know if the value set is default or user set. Nothing reads till until it reaches the processing job, so it shouldn't be an issue.

LTA-Thinking: What I said made no sense, I had forgotten I had already added the IsTillExplicit flag. Default returned.

rbans96:
**Potential Bug: Inconsistent DateTime → DateTimeOffset conversion**

In BulkUpdateOrchestratorJob.cs around line 178:
`csharp
var createDate = new PartialDateTime(jobInfo.CreateDate);
`

But in CreateProcessingDefinition (same file, line ~279):
`csharp
var createDate = new PartialDateTime(new DateTimeOffset(jobCreateDate, TimeSpan.Zero));
`

JobInfo.CreateDate is DateTime (not DateTimeOffset). The first form relies on C#'s implicit conversion, which uses the **local timezone offset** when DateTime.Kind == Unspecified. The second form explicitly sets UTC offset. If the DB returns Kind == Unspecified, these two paths will produce different timestamps on non-UTC servers.

Suggest using 
ew PartialDateTime(new DateTimeOffset(jobInfo.CreateDate, TimeSpan.Zero)) consistently in both places to avoid subtle timezone bugs.

rbans96 (approved):

PR #5563, opened 2026-05-12
Reverted unnecessary reindex test changes

SergeyGaluzo:
Reverts back all recent changes related to CI pipeline not working in the e2e ReindexTests class.
Reverts back parallel update test in e2e ReindexTests class to spread the update load.  
Add deletes of resources before reindex for count sensitive test.
Changes resource deletes on cleanup to hard deletes.
Adjusted CI settings to match PR ones (exception CPU and RAM on replicas)

### DIFF
--- a/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
+++ b/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
@@ -1192,9 +1192,6 @@ namespace Microsoft.Health.Fhir.Tests.E2E.Rest.Reindex
                     _output.WriteLine($"Failed to delete SearchParameter/{param?.Id}: {ex.Message}");
                 }
             }
-
-            // Allow time for soft deletes to be processed
-            await Task.Delay(500);
         }
 
         private async Task<OperationStatus> WaitForJobCompletionAsync(Uri jobUri, TimeSpan timeout)
--- a/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
+++ b/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
@@ -396,6 +396,9 @@ namespace Microsoft.Health.Fhir.Tests.E2E.Rest.Reindex
                 ],
             };
 
+            // This test is count sensitive.
+            // It will work only if there is expected number of persons and there are no other resource types to reindex.
+            // Hence - delete and reindex.
             await DeleteResourcesAsync("Person");
             value = await _fixture.TestFhirClient.PostReindexJobAsync(parameters);
             await WaitForJobCompletionAsync(value.jobUri, TimeSpan.FromSeconds(300));
--- a/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
+++ b/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
@@ -1219,10 +1219,6 @@ namespace Microsoft.Health.Fhir.Tests.E2E.Rest.Reindex
 
             // Allow time for soft deletes to be processed
             await Task.Delay(500);
-
-            // Note: Final reindex is handled by ReindexTestFixture.OnDisposedAsync() which runs
-            // once at the very end after all tests complete. This eliminates the need to reindex
-            // after each individual test cleanup.
         }
 
         private async Task<OperationStatus> WaitForJobCompletionAsync(Uri jobUri, TimeSpan timeout)
--- a/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
+++ b/test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
@@ -48,16 +48,20 @@ namespace Microsoft.Health.Fhir.Tests.E2E.Rest.Reindex
 
         public async Task InitializeAsync()
         {
-            // Delete leftover resources from previous test runs
-            _output.WriteLine("Initializing ReindexTests - cleaning up leftover resources...");
+            // delete leftover resources from previous test runs
+            // this guarantees that we start from clean state even on previous hard failures when we cannot rely on finally
+            // we can consider removing finally altogether
+            var sw = Stopwatch.StartNew();
+            _output.WriteLine("ReindexTests.InitializeAsync: Starting...");
 
+            await CancelAnyRunningReindexJobsAsync();
             await DeleteResourcesAsync("SearchParameter", false); // hard delete does not work right
             await DeleteResourcesAsync("Person", true);
             await DeleteResourcesAsync("Specimen", true);
             await DeleteResourcesAsync("SupplyDelivery", true);
             await DeleteResourcesAsync("Immunization", true);
 
-            _output.WriteLine("ReindexTests initialization completed.");
+            _output.WriteLine($"ReindexTests.InitializeAsync: Completed. Elapsed={(int)sw.Elapsed.TotalMilliseconds} msec.");
         }
 
         public Task DisposeAsync()
@@ -68,8 +72,6 @@ namespace Microsoft.Health.Fhir.Tests.E2E.Rest.Reindex
         [Fact]
         public async Task Given500SearchParams_WhenReindexCompletes_ThenSearchParamsAreEnabled()
         {
-            await CancelAnyRunningReindexJobsAsync();
-
             const int numberOfSearchParams = 10; // increase to 500 when cache is not updated by API calls and status is saved with resources in a single SQL transaction
             const string urlPrefix = "http://my.org/";
             var codes = new List<string>();
@@ -153,7 +155,6 @@ namespace Microsoft.Health.Fhir.Tests.E2E.Rest.Reindex
         [Fact]
         public async Task GivenTwoSearchParamsWithCodeConflictOnDerived_ThenBadRequestIsReturned()
         {
-            await CancelAnyRunningReindexJobsAsync();
 #if R5
             var personTypes = new List<VersionIndependentResourceTypesAll?>() { VersionIndependentResourceTypesAll.Person };
             var resourceTypes = new List<VersionIndependentResourceTypesAll?>() { VersionIndependentResourceTypesAll.Resource };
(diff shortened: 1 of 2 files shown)


fhibf (dismissed):

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
```diff
@@ -35,12 +35,6 @@ namespace Microsoft.Health.Fhir.Tests.E2E.Rest.Reindex
     [HttpIntegrationFixtureArgumentSets(DataStore.All, Format.Json)]
     public class ReindexTests : IClassFixture<HttpIntegrationTestFixture>
     {
-        // Maximum time to wait for a reindex job to reach a terminal state. Set high enough to accommodate
-        // multi-replica search-parameter cache convergence in CI (poll interval up to 30s, conformance refresh
-        // up to 60s, plus reindex worker queue scheduling and retry backoffs).
-        private static readonly TimeSpan ReindexJobCompletionTimeout = TimeSpan.FromMinutes(20);
```
jestradaMS, line 41: removing this will cause issues for in PaaS as well as CI as we have seen cases where reindex tests take longer than 5 minutes. IF you have already solved this, great, otherwise you will recreate this failure of the failing as it expects completed but got running......

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
```diff
@@ -537,7 +501,7 @@ public async Task GivenReindexJobWithMixedZeroAndNonZeroCountResources_WhenReind
                 Assert.NotNull(value.jobUri);
 
                 // Wait for job to complete (this will wait for all sub-jobs to complete)
-                var jobStatus = await WaitForJobCompletionAsync(value.jobUri, ReindexJobCompletionTimeout);
+                var jobStatus = await WaitForJobCompletionAsync(value.jobUri, TimeSpan.FromSeconds(300));
```
jestradaMS, line 516: suggest using the common const for all of these as well as allow for more time for larger dbs that also use these tests in PaaS

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
SergeyGaluzo: All reindex tests are passing in CI

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
```diff
@@ -1226,16 +1216,13 @@ private async Task CleanupSearchParametersAsync(params SearchParameter[] searchP
...
-        }
+            // Allow time for soft deletes to be processed
+            await Task.Delay(500);
 
-        private static string BuildTestSearchParameterUrl(string suffix)
-        {
-            return $"{TestSearchParameterUrlPrefix}{suffix}";
+            // Note: Final reindex is handled by ReindexTestFixture.OnDisposedAsync() which runs
+            // once at the very end after all tests complete. This eliminates the need to reindex
+            // after each individual test cleanup.
```
jestradaMS: ReindexTestFixture does not exist anymore. I believe you deleted this several months ago.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
jestradaMS: For example in your current run, it passed next team as it got lucky on retry with no other load

https://microsofthealthoss.visualstudio.com/FhirServer/_build/results?buildId=49380&view=ms.vss-test-web.build-test-results-tab&runId=2228876&resultId=100848&paneView=debug

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
jestradaMS: https://microsofthealthoss.visualstudio.com/FhirServer/_build/results?buildId=49380&view=ms.vss-test-web.build-test-results-tab&runId=2228876&resultId=100848&paneView=debug

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
SergeyGaluzo: Good point. Old comment removed.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
```diff
@@ -396,69 +387,42 @@ public async Task GivenReindexJobWithConcurrentUpdates_ThenReportedCountsAreLess
...
+            // Hence - delete and reindex.
+            await DeleteResourcesAsync("Person");
+            value = await _fixture.TestFhirClient.PostReindexJobAsync(parameters);
+            await WaitForJobCompletionAsync(value.jobUri, TimeSpan.FromSeconds(300));
 
             try
             {
                 var randomSuffix = Guid.NewGuid().ToString("N").Substring(0, 8);
-                var resources = await SetupTestDataAsync("Person", 20, randomSuffix, CreatePersonResourceAsync);
+                var resources = await SetupTestDataAsync("Person", 100, randomSuffix, CreatePersonResourceAsync);
```
jestradaMS, line 424: This test is also still very flaky. It is also an issue PaaS, but tends to pass on retries. 

https://microsofthealthoss.visualstudio.com/FhirServer/_build/results?buildId=49380&view=ms.vss-test-web.build-test-results-tab&runId=2228826&resultId=100701&paneView=debug

SergeyGaluzo: My goal was to revert to the state where **processing** jobs are reporting less records than orchestrator originally saw. This makes this test highly dependent on the cache refresh timings, which are environment dependent and not accessible in e2e. It is possible that we might disable this test in any environments other than PR. If you can suggest how I can put in code.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
SergeyGaluzo: I don't see the reason why reindex tests which are dealing with resource types not used by other e2e tests would run longer even in bigger databases. Can you provide examples?

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
SergeyGaluzo: There should not any good reason for single resource test to run longer than 5 minutes. We need to look at why it is happening. Increasing time is not a correct approach because it might hide the root cause.

jestradaMS: Based on log analytics it was a phantom host that had scaled down but fell into the lookback period set in the convergence logic. You can take a look at this job in CI log analytics. 


  Finding: Of 11 reindex orchestrator jobs in this build, only Job 1571 (the first) exceeded 5 minutes — it ran 330 seconds (17:33:39 → 17:39:09 UTC) before being cancelled by the test client. Jobs 1572–1644 all completed in 81–94 seconds.

  Root cause: ReindexOrchestratorJob.WaitForAllInstancesCacheSyncAsync got stuck waiting for 3/4 hosts synced. The 4th "active" host was a phantom — a replica (xhgdw or lqlp8) that ACA had scaled down ~2 min before the test started but whose last heartbeat still
  fell inside the orchestrator's 180 s active-hosts window (ActiveHostsEventsMultiplier=9 × SearchParameterCacheRefreshIntervalSeconds=20s). The dead replica could never refresh its search-parameter cache hash, so the orchestrator polled until the E2E client
  cancelled it first.

SergeyGaluzo: Good info. If we waited a little bit longer, orchestrator would have failed anyway. Therefore, increase in total reindex wait time is not a solution for this problem. I think it is acceptable to have intermittent tests failures because of this,

jestradaMS: This is something that could happen in production as well during scaling. If nothing else we should at least have a follow up work item to harden the convergence for self healing this case e.g. for hosts that haven't converged, check if they are still active in last x time vs just relying on the start lookback.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
jestradaMS: In ci-pipeline.yml update the mainCategoryFilter for the sql tests to exclude that specific test e.g. DisplayName!=[test name]

This will need to be added to PaaS AKS E2E tests as well.

SergeyGaluzo: Thanks for suggestion. I will keep it in mind as last resort. For now, it works in CI.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
SergeyGaluzo: I am not following. First, reindex will fail with "unable to update cache please retry" or such. When customer retries, old pods should not be considered as there are no messages in the interval orchestrator looks at. Looks that we do not need to do anything. Am I missing sometging?

jestradaMS: Yes, it's a bad customer experience for something we can solve.

SergeyGaluzo: We are discussing current functionality that exists in PROD, and it is not related to this PR.
I don't see indications that we have PROD problems in this functionality, and therefore I am not comfortable to add any work items. If you think it is justified, please go ahead.

fhibf (approved):

PR #5576, opened 2026-05-15
Search param concurrency based on max last updated

SergeyGaluzo:
Fixes 
https://microsofthealth.visualstudio.com/Health/_workitems/edit/188506/ 
by implementing
https://github.com/microsoft/fhir-server/blob/main/docs/arch/adr-2603-search-params-concurrency.md
This significantly simplified processing, so special concurrency classes were removed.

It also addresses:
https://microsofthealth.visualstudio.com/Health/_workitems/edit/191077/
https://microsofthealth.visualstudio.com/Health/_workitems/edit/188488/
https://microsofthealth.visualstudio.com/Health/_workitems/edit/192230/
https://microsofthealth.visualstudio.com/Health/_workitems/edit/192256/

**max(last updated) handling rules:**

**Common flow**
- If request context does not contain LastUpdated, validation refreshes cache, gets it, and saves to the context.
- Create/update/delete behavior uses that value for pending status updates.
- LastUpdated is cleared from the context when the request completes.

**By execution mode**
- **Non-bundle**: follows the common flow.
- **Sequential bundle**: each SearchParameter entry follows the common flow independently.
- **Parallel bundle**: bundle handler preloads LastUpdated once and propagates it to inner requests.

**Retries**
- **Non-bundle** and **sequential bundle** search param API calls are retried on concurrency conflicts. This significantly reduces probability of concurrency conflicts returned to the clients for calls executed in parallel.


src/Microsoft.Health.Fhir.SqlServer/Features/ExceptionExtension.cs
```diff
@@ -25,6 +27,12 @@ public static bool IsExecutionTimeout(this Exception e)
             return str.Contains("execution timeout expired", StringComparison.OrdinalIgnoreCase);
         }
 
+        public static bool IsSearchParameterConcurrencyConflict(this Exception e)
+        {
+            var sqlEx = e as SqlException;
+            return sqlEx != null && sqlEx.Number == 50001 && e.Message.StartsWith("optimistic concurrency conflict", StringComparison.OrdinalIgnoreCase);
```
fhibf: nit: I believe we should add this error code to FhirSqlErrorCodes. 
Just to keep it documented as the "search parameter optimistic concurrency error code".

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/SearchParameterOperations.cs
```diff
@@ -98,78 +96,75 @@ public async Task EnsureNoActiveReindexJobAsync(CancellationToken cancellationTo
...
-                },
-                _logger,
-                cancellationToken);
+            return lastUpdated.Value;
         }
 
         /// <summary>
         /// Marks the Search Parameter as PendingDelete. This is only used by DeletionService.cs and will be removed when refactoring is done
         /// to allow deletion service to properly handle Hard deletions for Search Parameters (e.g. allow reindex prior to removing resource from DB).
+        /// !!! This method has incorrect name. It does not delete search parameter, it just updates its status.
```
fhibf, line 178: Is there any blocked that would avoid updating the method name?

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Parameters/SearchParameterValidator.cs
```diff
@@ -108,7 +109,11 @@ public async Task ValidateSearchParameterInput(SearchParameter searchParam, stri
                 else
                 {
                     // Refresh the search parameter cache in the search parameter definition manager before starting the validation.
-                    await _searchParameterOperations.GetAndApplySearchParameterUpdates(cancellationToken);
+                    if (!lastUpdated.HasValue)
+                    {
+                        await _searchParameterOperations.GetAndApplySearchParameterUpdates(cancellationToken);
+                        lastUpdated = _searchParameterOperations.SearchParamLastUpdated.Value;
```
fhibf: Just double-checking something I've noticed: I was investigating possible conditions that would lead __searchParameterOperations.SearchParamLastUpdated_ to be null. 

Based on what I see in the code, the stored procedure "_dbo.GetSearchParamStatuses_" will always return a list of search parameters (even if customer does not have custom search parameters, the service has the default search parameters).

With that, I don't see a reason for SearchParamLastUpdated to be from a nullable type. 

What I mean with that is: every time there is a variable of a nullable type, we expect to have some code checking if that value is null or not. But, from the existing state of the service, I don't see conditions where would ne nullable. 

Can we change SearchParamLastUpdated from 'DateTimeOffset?' to 'DateTimeOffset' (non-nullable)? - at least from the SearchParameterOperations's perspective.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -332,9 +332,17 @@ private async Task CheckSearchParamInputConflictsAndUpdateCache(Hl7.Fhir.Model.B
                 throw new RequestNotValidException(string.Format(Api.Resources.DuplicateSearchParamCodesAndUrlsInBundle, string.Join(", ", dupCodes), string.Join(", ", dupUrls)));
             }
 
+            // for deletes Entry.Resource is null. need to check in other way
+            if (!searchParamsInBundle && bundle.Entry.Any(e => e.Request.Method == HTTPVerb.DELETE && e.Request.Url.StartsWith(KnownResourceTypes.SearchParameter, StringComparison.OrdinalIgnoreCase) == true))
```
LTA-Thinking: Nit: StartsWith already returns a boolean, it doesn't need the `== true`

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -332,9 +332,17 @@ private async Task CheckSearchParamInputConflictsAndUpdateCache(Hl7.Fhir.Model.B
...
+            {
+                searchParamsInBundle = true;
+            }
+
             if (searchParamsInBundle)
             {
                 await _searchParameterOperations.GetAndApplySearchParameterUpdates(cancellationToken); // refresh search param cache
+
+                _fhirRequestContextAccessor.RequestContext.Properties[SearchParameterRequestContextPropertyNames.LastUpdated] = _searchParameterOperations.SearchParamLastUpdated.Value; // capture last updated
```
LTA-Thinking: Could this use the extension method to set the last updated value?

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Delete/DeletionService.cs
```diff
@@ -459,19 +459,33 @@ await CreateAuditLog(
...
-                await Parallel.ForEachAsync(includedResources, cancellationToken, async (item, innerCt) =>
+                await Parallel.ForEachAsync(includedResources.Where(_ => _.Resource.ResourceTypeName != KnownResourceTypes.SearchParameter), cancellationToken, async (item, innerCt) =>
                 {
-                    await DeleteSearchParameterAsync(item.Resource, cancellationToken);
                     await _retryPolicy.ExecuteAsync(async () => await fhirDataStore.HardDeleteAsync(new ResourceKey(item.Resource.ResourceTypeName, item.Resource.ResourceId), request.DeleteOperation == DeleteOperation.PurgeHistory, request.AllowPartialSuccess, innerCt));
                     parallelBag.Add((item.Resource.ResourceTypeName, item.Resource.ResourceId, item.SearchEntryMode == ValueSets.SearchEntryMode.Include));
                 });
 
-                await Parallel.ForEachAsync(matchedResources, cancellationToken, async (item, innerCt) =>
+                // with concurrency based on max last updated search params must be deleteted one-by-one
```
LTA-Thinking: The deletion service can have multiple groups being deleted at once. On line 217, this method is called inside a loop that can have up to 64 threads running at once. I don't think this alone will guaranty that Search Parameters aren't deleted in parallel.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SqlServerSearchParameterStatusDataStore.cs
```diff
@@ -139,64 +141,21 @@ public async Task UpsertStatuses(IReadOnlyCollection<ResourceSearchParameterStat
...
-                }
-            }
-        }
-
-        private async Task UpsertStatusesInternal(IReadOnlyCollection<ResourceSearchParameterStatus> statuses, CancellationToken cancellationToken, long? reindexId = null)
-        {
             using var cmd = new SqlCommand();
             cmd.CommandType = CommandType.StoredProcedure;
             cmd.CommandText = "dbo.MergeSearchParams";
-            if (_schemaInformation.Current >= 112 && reindexId.HasValue) // remove value check to invoke new max(LastUpdated) logic
```
LTA-Thinking, line 192: We need to wait until after the cutoff for the next release since we had to rollback the release last week, so not all production databases are on 112 yet.

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/CreateOrUpdateSearchParameterBehavior.cs
```diff
@@ -90,38 +88,32 @@ public async Task<UpsertResourceResponse> Handle(UpsertResourceRequest request,
...
+                    QueueStatusAsync(request.Resource.Instance.GetStringScalar("url"), SearchParameterStatus.Supported, lastUpdated, cancellationToken);
                 }
             }
 
             // Now allow the resource to updated per the normal behavior
             return await next(cancellationToken);
         }
 
-        private async Task QueueStatusAsync(string url, SearchParameterStatus status, CancellationToken cancellationToken)
+        private void QueueStatusAsync(string url, SearchParameterStatus status, DateTimeOffset lastUpdated, CancellationToken cancellationToken)
```
fhibf: nit: rename method to "QueueStatus" and remove the cancellation token, as this method is not async anymore.

This change would be very well appreciated: every time I read the code an see the calls to QueueStatusAsync I think "this statement is missing an 'await'", and then I remember the method is not async.

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/SearchParameterOperations.cs
fhibf: While testing, I've found a bug on the format of the error message in this line: the message "ChangesToSearchParametersNotAllowedWhileReindexing" contains an argument which is not being passed. 

The error message mentions a Job, but the ID is not provided. 

Image

I know it's not related to your changes, but could you change the code on like 95 to:

```
throw new JobConflictException(string.Format(Core.Resources.ChangesToSearchParametersNotAllowedWhileReindexing, activeReindexJob.id));
```

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/CreateOrUpdateSearchParameterBehavior.cs
```diff
@@ -141,14 +133,13 @@ private async Task QueueStatusAsync(string url, SearchParameterStatus status, Ca
                 context.Properties[SearchParameterRequestContextPropertyNames.PendingStatusUpdates] = pendingStatuses;
```
fhibf, line 139: We could add to FhirRequestContextExtension two new methods to handle 'SearchParameterRequestContextPropertyNames.PendingStatusUpdates' the same way how 'SearchParameterRequestContextPropertyNames.LastUpdated' is handled. 

I see some places where that could be reused.

src/Microsoft.Health.Fhir.SqlServer/Features/ExceptionExtension.cs
SergeyGaluzo: This is current functionality this PR does not change. It existed when Uri-based concurrency was first introduced. I don't have strong opinion on error codes versus error strings, because strings are as reliable as codes. If you believe that code is required, please create a separate work item

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/SearchParameterOperations.cs
SergeyGaluzo: I put a note here for future because we need to clean this area as far as minimizing truly needed methods. This will be a good time to change. I don't want to do it in this PR as it will increase number of changed files significantly.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Parameters/SearchParameterValidator.cs
SergeyGaluzo: The only reason is that before the first run on GetAndAply method SearchParamLastUpdated is **not** set. NULL is a natural choice here. I do not agree that for all nullable properties we should always check if there are null. We can rely on workflow that updates them. This is the case for SearchParamLastUpdated. In the new concurrency code, I always use **.Value** without any checks and/or defaults. I prefer code to fail rather than change its behavior **silently**.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
SergeyGaluzo: It could. I just did not see the need for that, as it is a simple single liner.

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
SergeyGaluzo: You are right. Fixed.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/Registry/SqlServerSearchParameterStatusDataStore.cs
SergeyGaluzo: That is true. This PR takes dependency on 112 deployed in PROD. I made this explicit in schema versions. I will check in only after confirming.

src/Microsoft.Health.Fhir.Shared.Core/Features/Search/Parameters/SearchParameterValidator.cs
SergeyGaluzo: On the second thought... I figured a way to make it more explicit. Fixed.

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/CreateOrUpdateSearchParameterBehavior.cs
SergeyGaluzo: Fixed

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Delete/DeletionService.cs
SergeyGaluzo: That is a strange logic. Looks like we have double parallelism. This can easily get out of hand... Let's discuss what options we have,

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/SearchParameterOperations.cs
SergeyGaluzo: Fixed as sugested

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
SergeyGaluzo: Changed to use SetSearchParameterLastUpdated extension method

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/CreateOrUpdateSearchParameterBehavior.cs
SergeyGaluzo: I added this work item to handle this in separate PR https://microsofthealth.visualstudio.com/Health/_workitems/edit/194526

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Delete/DeletionService.cs
SergeyGaluzo: I added serialization via semaphore.

fhibf (dismissed):

LTA-Thinking (dismissed):

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Search/SearchParameters/SearchParameterRetryIntegrationTests.cs
```diff
@@ -0,0 +1,251 @@
...
+    {
+        private readonly ISearchParameterOperations _searchParameterOperations;
+        private readonly IFhirDataStore _fhirDataStore;
+        private readonly ISearchParameterDefinitionManager _searchParameterDefinitionManager;
+        private readonly ISearchParameterStatusManager _searchParameterStatusManager;
+        private readonly RequestContextAccessor<IFhirRequestContext> _requestContextAccessor;
+        private readonly IFhirRequestContext _fhirRequestContext;
+        private readonly IModelInfoProvider _modelInfoProvider;
+
+        public SearchParameterRetryIntegrationTests()
```
LTA-Thinking: Nit: I don't like a unit test having Integration in its name. I think it will cause confusion with actual integration tests.

SergeyGaluzo: Fixed

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -264,14 +264,36 @@ public async Task<BundleResponse> Handle(BundleRequest request, CancellationToke
...
 
-                    await ExecuteTransactionForAllRequestsAsync(responseBundle, bundleProcessingLogic, cancellationToken);
+                    // Check if bundle contains SearchParameter resources to determine if retry logic is needed
+                    bool containsSearchParameters = bundleResource.Entry.Any(e => string.Equals(e.Resource?.TypeName, KnownResourceTypes.SearchParameter, StringComparison.Ordinal));
+
+                    if (containsSearchParameters && bundleProcessingLogic == BundleProcessingLogic.Parallel)
+                    {
+                        // Apply retry logic for SearchParameter bundles that use parallel processing
+                        // Bundle handler is always the entry point that sets lastUpdated, so it should handle retries
+                        // This includes both the cache refresh and the transaction execution
```
LTA-Thinking: This statement feels weird. Why are we including the retry policy here if the retries are being handled elsewhere?

SergeyGaluzo: Retry policy is skipped depending on whether last updated is set in context. This allows to apply policy once. It is handy if different pieces of http pipeline are called on different code paths.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/FhirController.cs
LTA-Thinking: Should the SearchParameter retry also be added to the conditional operations?

SergeyGaluzo: @fhibf identified that parallel calls are happening for simple APIs. That is what was protected. I believe we do not need more.

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SearchParameterOptimisticConcurrencyTests.cs
```diff
@@ -0,0 +1,175 @@
...
+                {
+                    Uri = createdStatus.Uri,
+                    Status = SearchParameterStatus.Supported,
+                    IsPartiallySupported = true,
+                    LastUpdated = createdStatus.LastUpdated, // This is now stale
+                };
+
+                try
+                {
+                    await _fixture.SearchParameterStatusDataStore.UpsertStatuses([staleUpdate], CancellationToken.None);
```
LTA-Thinking, line 162: There should be an assert fail after the upsert call to make sure the exception is thrown.

SergeyGaluzo: Fixed

src/Microsoft.Health.Fhir.Shared.Api/Controllers/FhirController.cs
SergeyGaluzo: Actually... added retries on conditional

LTA-Thinking (dismissed):

LTA-Thinking (approved):

fhibf (approved):

PR #5578, opened 2026-05-19
Reject requests with JWT access tokens in URL path

rbans96:
## Problem
Customer sent an access token as a URL path parameter (e.g. /Encounter/{jwt}). While the request returned 401, it still reached the FHIR pipeline and the full URL (containing the token) was logged by the HTTP middleware.

## Workitem
[192527](https://microsofthealth.visualstudio.com/Health/_workitems/edit/192527)

## Fix
Add AccessTokenUrlValidationMiddleware that runs **before** FhirRequestContextMiddleware in the pipeline. It detects JWT patterns in the URL path or query string and immediately returns **400 Bad Request**, preventing the token from ever being captured in audit logs or telemetry.

## Changes
- AccessTokenUrlValidationMiddleware.cs — Detects JWTs via regex, rejects with 400
- AccessTokenUrlValidationMiddlewareExtensions.cs — Pipeline registration extension
- FhirServerServiceCollectionExtensions.cs — Register middleware before FhirRequestContext
- AccessTokenUrlValidationMiddlewareTests.cs — Unit tests verifying rejection, passthrough, and that tokens are never logged downstream

### DIFF
--- a/src/Microsoft.Health.Fhir.Api/Features/Security/AccessTokenUrlValidationMiddleware.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/Security/AccessTokenUrlValidationMiddleware.cs
@@ -1,0 +1,79 @@
+// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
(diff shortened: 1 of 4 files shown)


feordin:
Manually tested with AadProxy enabled, and the flow still works.

feordin (approved):

PR #4942, opened 2025-04-23
***DO NOT REVIEW*** Reindex job refactor to Queue Client

jestradaMS:
## Description
Describe the changes in this PR.

Rebasing main on previous branch for reindex job refactoring to Queue Client
Changes required post rebase to support dependency or impacted changes since

## Related issues
Addresses [issue #].

## Testing
Describe how this change was tested.

## FHIR Team Checklist
- **Update the title** of the PR to be succinct and less than 65 characters
- **Add a milestone** to the PR for the sprint that it is merged (i.e. add S47)
- Tag the PR with the type of update: **Bug**, **Build**, **Dependencies**, **Enhancement**, **New-Feature** or **Documentation**
- Tag the PR with **Open source**, **Azure API for FHIR** (CosmosDB or common code) or **Azure Healthcare APIs** (SQL or common code) to specify where this change is intended to be released.
- Tag the PR with **Schema Version backward compatible** or **Schema Version backward incompatible** or **Schema Version unchanged** if this adds or updates Sql script which is/is not backward compatible with the code.
- When changing or adding behavior, if your code modifies the system design or changes design assumptions, please create and include an [ADR](https://github.com/microsoft/fhir-server/blob/main/docs/arch).
- [ ] CI is green before merge [Build Status](https://microsofthealthoss.visualstudio.com/FhirServer/_build/latest?definitionId=27&branchName=main) 
- Review [squash-merge requirements](https://github.com/microsoft/fhir-server/blob/main/SquashMergeRequirements.md)

### Semver Change ([docs](https://github.com/microsoft/fhir-server/blob/main/docs/Versioning.md))
Patch|Skip|Feature|Breaking (reason)

PR #5579, opened 2026-05-19
Remove validation on TargetSearchParameterTypes

LTA-Thinking:
## Description
The reindex parameter TargetSearchParameterTypes isn't used anymore but is still being validated, which can cause a reindex job to be refused. This PR removes the validation on this field.

## Related issues
Addresses [Bug 192567](https://microsofthealth.visualstudio.com/Health/_workitems/edit/192567)

## Testing
Manual and existing tests

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/CreateReindexRequestHandler.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/CreateReindexRequestHandler.cs
@@ -67,27 +67,8 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Reindex
                 return new CreateReindexResponse(existingJob);
             }
 
-            // What this handles is the scenario where a user is effectively forcing a reindex to run by passing
-            // in a parameter of targetSearchParameterTypes. From those we can identify the base resource types.
-            var searchParameterResourceTypes = new HashSet<string>();
-            if (request.TargetSearchParameterTypes.Any())
-            {
-                foreach (var searchParameterUrl in request.TargetSearchParameterTypes)
-                {
-                    SearchParameterInfo searchParameterInfo = _searchParameterDefinitionManager.GetSearchParameter(searchParameterUrl);
-                    if (searchParameterInfo == null)
-                    {
-                        throw new RequestNotValidException(Core.Resources.SearchParameterByDefinitionUriNotSupported, searchParameterUrl);
-                    }
-
-                    searchParameterResourceTypes.UnionWith(searchParameterInfo.BaseResourceTypes);
-                }
-            }
-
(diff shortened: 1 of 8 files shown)


src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/CreateReindexRequestHandler.cs
```diff
@@ -67,27 +67,8 @@ public async Task<CreateReindexResponse> Handle(CreateReindexRequest request, Ca
                 return new CreateReindexResponse(existingJob);
             }
 
-            // What this handles is the scenario where a user is effectively forcing a reindex to run by passing
```
fhibf, line 70: Is this a regressions? Are we stopping to support this scenario?

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/Models/ReindexJobRecord.cs
```diff
@@ -32,17 +32,13 @@ public class ReindexJobRecord : JobRecord, IJobData
 
         public ReindexJobRecord(
             IReadOnlyCollection<string> targetResourceTypes,
-            IReadOnlyCollection<string> targetSearchParameterTypes,
-            IReadOnlyCollection<string> searchParameterResourceTypes,
```
fhibf, line 36: Is this deserialized from the job queue? Any risk of regressions in case we release these changes and then rollback?

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/CreateReindexRequestHandler.cs
PTaladay: I'm wondering the same thing. I don't believe this scenario was used all that often to begin with as if I recall it was more of a stop gap for when a full reindex was failing and kind of force it to be chunked up.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/Models/ReindexJobRecord.cs
LTA-Thinking: The deserializer uses the empty constructor, so this wouldn't affect it rolling forward or backward.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/CreateReindexRequestHandler.cs
LTA-Thinking: We removed support for this in #4942

fhibf (approved):

PR #5582, opened 2026-05-22
Fix transaction bundle returning HTTP 408 instead of 4xx for validation errors

rbans96:
Transaction bundle returns HTTP 408 instead of 400 when entry fails validation during parallel execution. In `TransactionExceptionHandler.ThrowTransactionException`, when `isCancelled` is true, the code unconditionally throws `FhirTransactionCancelledException` (hardcoded to 408), ignoring the actual entry error status. During parallel bundle execution, when one entry fails validation, the catch handler cancels the linked `CancellationTokenSource`. Other entries with validation errors that complete after the token is cancelled see `IsBundleCancelledByClient() == true` and get misclassified as client cancellations (408) instead of validation failures (400). Since `await Task.WhenAll` propagates the first faulted task by array position (not chronological order), the 408 is returned to the caller. Callers correctly treat 408 as transient/retryable per HTTP semantics, causing infinite retry loops and massive request amplification (6-12x) for permanently failing bundles.

**Fix:** When an entry has a concrete client error (4xx, excluding 408), that status code takes priority over the cancellation signal. HTTP 408 is now reserved for genuine timeout/cancellation scenarios only. This aligns with the FHIR specification: 'return an HTTP 400 or 500 type response' for failed transactions.

Addresses : https://microsofthealth.visualstudio.com/Health/_workitems/edit/192974

### DIFF
--- a/src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Resources/Bundle/BundleHandlerTests.cs
+++ b/src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Resources/Bundle/BundleHandlerTests.cs
@@ -501,14 +501,90 @@ namespace Microsoft.Health.Fhir.Api.UnitTests.Features.Resources.Bundle
 
                 var bundleRequest = new BundleRequest(bundle.ToResourceElement());
 
-                // As the cancellation is requested during the bundle execution and the before the max transaction time, a FhirTransactionCancelledException is expected.
-                // Resulting in a HTTP408 error.
+                // The cancellation prevents the router from setting a handler, resulting in a 404 entry.
+                // Since isCancelled is true, FhirTransactionCancelledException (408) is thrown.
+                // The correct error prioritization for multi-entry scenarios happens at the
+                // AggregateException level using IsErrorCausedDueClientFailure().
                 FhirTransactionCancelledException fhirTce = await Assert.ThrowsAsync<FhirTransactionCancelledException>(async () => await _bundleHandler.Handle(bundleRequest, cancellationToken));
                 Assert.True(fhirTce.ResponseStatusCode == System.Net.HttpStatusCode.RequestTimeout);
 
                 // Ensures failure sign is emitted.
                 _bundleMetricHandler.Received(1).EmitFailure(Arg.Any<string>());
             }
+        }
+
+        [Fact]
+        public async Task GivenATransaction_WithTwoEntriesAndOneFailsWithClientError_ReturnClientErrorNotTimeout()
+        {
+            // When a transaction bundle has multiple entries processed in parallel, and one entry
+            // fails with a client error (400), the transaction should return that client error
+            // instead of 408 (Request Timeout) from entries cancelled as a side effect.
+            int callCount = 0;
+
+            var bundle = new Hl7.Fhir.Model.Bundle
+            {
+                Type = BundleType.Transaction,
+                Entry = new List<EntryComponent>
+                {
+                    new EntryComponent
+                    {
+                        Request = new RequestComponent
+                        {
+                            Method = HTTPVerb.POST,
+                            Url = "/Observation",
+                        },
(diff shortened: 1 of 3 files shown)


src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/TransactionExceptionHandler.cs
```diff
@@ -17,7 +17,16 @@ public static void ThrowTransactionException(string errorMessage, HttpStatusCode
         {
             List<OperationOutcomeIssue> operationOutcomeIssues = GetOperationOutcomeIssues(operationOutcome.Issue);
 
-            if (isCancelled)
+            // When an entry has a client error (4xx, excluding 408 itself), that error takes
+            // priority over the cancellation signal. The cancellation token may have been triggered
+            // by a previous entry's failure in parallel execution, not by the client disconnecting.
+            // Returning 408 for a validation error (e.g., 400) misleads callers into retrying
+            // a permanently failing request.
+            bool isClientError = statusCode >= HttpStatusCode.BadRequest
```
fhibf: As discussed offline, the suggestion is to use a flag to detect when a FhirTransactionFailedException happened due a customer caused issue. And then when detecting if the error was caused by a client cancellation, then mark it as a customer issue. 

That should remove the need to compare the precedence between BadRequest, InternalServerError and RequestTimeout. 

Image

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Resources/Bundle/BundleHandlerTests.cs
```diff
@@ -501,13 +501,14 @@ public async Task GivenATransaction_WithACancellationHappens_ReturnAProperError(
...
-                FhirTransactionCancelledException fhirTce = await Assert.ThrowsAsync<FhirTransactionCancelledException>(async () => await _bundleHandler.Handle(bundleRequest, cancellationToken));
-                Assert.True(fhirTce.ResponseStatusCode == System.Net.HttpStatusCode.RequestTimeout);
-
-                // Ensures failure sign is emitted.
-                _bundleMetricHandler.Received(1).EmitFailure(Arg.Any<string>());
+                // The cancellation causes the router to not set a handler, resulting in a 404 entry.
+                // Even though the cancellation token is triggered, the 4xx status from the entry
+                // takes priority over 408 to avoid misleading callers into retrying permanently failing requests.
+                FhirTransactionFailedException fhirTfe = await Assert.ThrowsAsync<FhirTransactionFailedException>(async () => await _bundleHandler.Handle(bundleRequest, cancellationToken));
+                Assert.True(fhirTfe.ResponseStatusCode == System.Net.HttpStatusCode.NotFound);
```
fhibf: I was not expecting this operation to return HTTP404. If that's the case, then we need to change this test to proper submit an Observation that would be ingested if the cancellation did not happen.

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Features/Resources/Bundle/TransactionExceptionHandlerTests.cs
```diff
@@ -49,6 +49,29 @@ public void GivenAnOperationOutcome_WhenCancelled_ThenACorrectExceptionIsThrown(
...
+            string message = "Error Message";
+            var operationOutcome = GetOperationOutcome();
+
+            FhirTransactionCancelledException tce = Assert.Throws<FhirTransactionCancelledException>(() => TransactionExceptionHandler.ThrowTransactionException(
+                message,
+                statusCode,
+                operationOutcome,
+                isCancelled: true));
+
+            Assert.Equal(HttpStatusCode.RequestTimeout, tce.ResponseStatusCode);
```
fhibf, line 72: I had the impression that the results should be different than Timeout.

PR #5583, opened 2026-05-22
Return 400 Bad Request for malformed Cosmos DB ct

rbans96:
## Summary

When the Cosmos DB SDK throws `MalformedContinuationTokenException` during client-side continuation token parsing, the exception was previously unhandled and resulted in **HTTP 500 Internal Server Error**.

This is a **client error** - the continuation token provided by the caller is invalid (e.g., stale, expired, or corrupted). Returning 500 incorrectly triggers ICM alerts (FHIR001 HTTP Internal Error threshold) and misrepresents the error to clients.

Addresses - https://microsofthealth.visualstudio.com/Health/_workitems/edit/192998

## Root Cause (ICM 791551287)

Clients were sending continuation tokens with `null` token fields and `resumeValues` timestamps from one year prior. The Cosmos DB SDK's internal `MalformedContinuationTokenException` escaped without being wrapped in a `CosmosException`, bypassing the existing handling in `CosmosResponseProcessor`.

## Testing

- 3 new unit tests covering:
  - MalformedContinuationToken exception -> 400 Bad Request
  - CosmosException still calls ProcessErrorResponse (existing behavior preserved)
  - Generic exceptions still propagate unchanged

### DIFF
--- a/src/Microsoft.Health.Fhir.CosmosDb/Features/Queries/CosmosQuery.cs
+++ b/src/Microsoft.Health.Fhir.CosmosDb/Features/Queries/CosmosQuery.cs
@@ -92,6 +92,13 @@ namespace Microsoft.Health.Fhir.CosmosDb.Features.Queries
 
                 throw;
             }
+            catch (Exception ex) when (ex.GetType().FullName?.Contains("MalformedContinuationToken", StringComparison.OrdinalIgnoreCase) == true)
+            {
+                // The Cosmos DB SDK may throw MalformedContinuationTokenException (an internal SDK type)
+                // during client-side continuation token parsing without wrapping it in a CosmosException.
+                // This is a client error (invalid/stale token), not a server error.
+                throw new RequestNotValidException(Microsoft.Health.Fhir.Core.Resources.InvalidContinuationToken);
+            }
         }
     }
 }
--- a/src/Microsoft.Health.Fhir.CosmosDb/Features/Queries/CosmosQuery.cs
+++ b/src/Microsoft.Health.Fhir.CosmosDb/Features/Queries/CosmosQuery.cs
@@ -97,6 +97,7 @@ namespace Microsoft.Health.Fhir.CosmosDb.Features.Queries
                 // The Cosmos DB SDK may throw MalformedContinuationTokenException (an internal SDK type)
                 // during client-side continuation token parsing without wrapping it in a CosmosException.
                 // This is a client error (invalid/stale token), not a server error.
+                _logger.LogMalformedContinuationToken(ex);
                 throw new RequestNotValidException(Microsoft.Health.Fhir.Core.Resources.InvalidContinuationToken);
             }
         }
--- a/src/Microsoft.Health.Fhir.CosmosDb.UnitTests/Features/Queries/CosmosQueryTests.cs
+++ b/src/Microsoft.Health.Fhir.CosmosDb.UnitTests/Features/Queries/CosmosQueryTests.cs
@@ -1,0 +1,115 @@
+// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Azure.Cosmos;
+using Microsoft.Health.Fhir.Core.Exceptions;
+using Microsoft.Health.Fhir.CosmosDb.Features.Queries;
+using Microsoft.Health.Fhir.CosmosDb.Features.Storage;
+using Microsoft.Health.Fhir.Tests.Common;
+using Microsoft.Health.Test.Utilities;
+using NSubstitute;
+using NSubstitute.ExceptionExtensions;
+using Xunit;
+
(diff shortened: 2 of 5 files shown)


src/Microsoft.Health.Fhir.CosmosDb/Features/Queries/CosmosQuery.cs
```diff
@@ -92,6 +92,13 @@ public async Task<FeedResponse<T>> ExecuteNextAsync(CancellationToken token = de
 
                 throw;
             }
+            catch (Exception ex) when (ex.GetType().FullName?.Contains("MalformedContinuationToken", StringComparison.OrdinalIgnoreCase) == true)
+            {
+                // The Cosmos DB SDK may throw MalformedContinuationTokenException (an internal SDK type)
+                // during client-side continuation token parsing without wrapping it in a CosmosException.
+                // This is a client error (invalid/stale token), not a server error.
+                throw new RequestNotValidException(Microsoft.Health.Fhir.Core.Resources.InvalidContinuationToken);
```
mikaelweave, line 101: Should we add logging when this fires to help livesite debugging?

`_logger.LogWarning(ex, "Cosmos SDK threw MalformedContinuationTokenException");`

src/Microsoft.Health.Fhir.CosmosDb/Features/Queries/CosmosQuery.cs
```diff
@@ -92,6 +92,13 @@ public async Task<FeedResponse<T>> ExecuteNextAsync(CancellationToken token = de
 
                 throw;
             }
+            catch (Exception ex) when (ex.GetType().FullName?.Contains("MalformedContinuationToken", StringComparison.OrdinalIgnoreCase) == true)
```
mikaelweave, line 95: A couple questions for this approach - don't need changes curious.

1. Could we de-duplicate the logic and create a helper with this logic that already checks? https://github.com/microsoft/fhir-server/blob/ee554e92a107e3f6dfd7b9e00a6438b47f563092/src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosResponseProcessor.cs#L82 Or are these too far apart? Idea would be a helper called something like `IsMalformedContinuationToken()`

2.
Are we using a contains to catch multiple exceptions or make the code less brittle to changes?

(vs looking for the entire exception name `Microsoft.Azure.Cosmos.Query.Core.Exceptions.MalformedContinuationTokenException` or just using `GetType().Name == `MalformedContinuationTokenException`)

mikaelweave (dismissed):
Just a couple questions. Mergable.

src/Microsoft.Health.Fhir.CosmosDb/Features/Queries/CosmosQuery.cs
rbans96: Good points-

1. kept them separate because the two checks handle different exception shapes:

CosmosResponseProcessor.cs (line 82) — Checks the error message string inside a [CosmosException](vscode-file://vscode-app/c:/Users/ribans/AppData/Local/Programs/Microsoft%20VS%20Code/f6cfa2ea24/resources/app/out/vs/code/electron-browser/workbench/workbench.html) (the SDK wraps some continuation token errors in a [CosmosException](vscode-file://vscode-app/c:/Users/ribans/AppData/Local/Programs/Microsoft%20VS%20Code/f6cfa2ea24/resources/app/out/vs/code/electron-browser/workbench/workbench.html) with status code 400).
[CosmosQuery.cs](vscode-file://vscode-app/c:/Users/ribans/AppData/Local/Programs/Microsoft%20VS%20Code/f6cfa2ea24/resources/app/out/vs/code/electron-browser/workbench/workbench.html) (this PR) — Checks the exception type name for cases where the SDK throws MalformedContinuationTokenException directly without wrapping it in a [CosmosException](vscode-file://vscode-app/c:/Users/ribans/AppData/Local/Programs/Microsoft%20VS%20Code/f6cfa2ea24/resources/app/out/vs/code/electron-browser/workbench/workbench.html).

2. The exception type ([Microsoft.Azure.Cosmos.Query.Core.Exceptions.MalformedContinuationTokenException](vscode-file://vscode-app/c:/Users/ribans/AppData/Local/Programs/Microsoft%20VS%20Code/f6cfa2ea24/resources/app/out/vs/code/electron-browser/workbench/workbench.html)) is internal to the SDK, so we can't reference it directly. Using [Contains](vscode-file://vscode-app/c:/Users/ribans/AppData/Local/Programs/Microsoft%20VS%20Code/f6cfa2ea24/resources/app/out/vs/code/electron-browser/workbench/workbench.html) rather than an exact [GetType().Name ==](vscode-file://vscode-app/c:/Users/ribans/AppData/Local/Programs/Microsoft%20VS%20Code/f6cfa2ea24/resources/app/out/vs/code/electron-browser/workbench/workbench.html) match makes this more resilient to the SDK moving the type to a different namespace in future versions — the class name substring is the stable identifier.

mikaelweave (dismissed):

src/Microsoft.Health.Fhir.CosmosDb/Features/Queries/CosmosQuery.cs
mikaelweave: `GetType().Name` should always return `MalformedContinuationTokenException` making it resistant to class changes.

`GetType().FullName` returns the class name per my understanding.

That being said - I think this is splitting hairs and I trust your judgement :P

mikaelweave (approved):

PR #5611, opened 2026-06-15
Enlist transactions enabled only for Sequential Transactions

fhibf:
## Description
In this PR, I'm setting "_enlistTransaction_" to **false** for most all operations, except by Sequential Bundle Transactions. 

During last week, it was noticed that deadlocks were not retried in standalone operations as they were set to enlist in transaction. By setting that to false, it'll now: (1) be included as part of SQL transaction in MergeResources; (2) retry deadlock errors if any. 

With these changes we should make standalone operations more stable and resilient to sporadic deadlocks (considered as retriable). 

## Related issues
Addresses [AB#195581](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/195581).

## Testing
- Tested through local validations. 

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
@@ -168,7 +168,7 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Storage
                         mergeOptions.EnlistInTransaction,
                         retries == 0,
                         eventualConsistency: false,
-                        ensureAtomicOperations: mergeOptions.EnsureAtomicOperations,
+                        isBundleTransaction: mergeOptions.IsBundleTransaction,
                         cancellationToken); // TODO: Pass correct retries value once we start supporting retries
                     return results;
                 }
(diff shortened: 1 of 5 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -894,8 +902,9 @@ public async Task<UpsertOutcome> UpsertAsync(ResourceWrapperOperation resource,
                     SetAndClearPendingSearchParameterStatus(resource);
                 }
 
-                // For regular upserts and sequential bundle operations, enlistTransaction is set to true.
-                MergeOptions mergeOptions = new MergeOptions(enlistTransaction: true, ensureAtomicOperations: isBundleTransaction);
+                MergeOptions mergeOptions = new MergeOptions(
+                    enlistTransaction: enlistTransaction,
+                    ensureAtomicOperations: isBundleTransaction);
```
SergeyGaluzo: ensureAtomicOperations always equals to isBundleTransaction. Why not to rename ensureAtomicOperations  to match?

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -878,13 +878,21 @@ public async Task<UpsertOutcome> UpsertAsync(ResourceWrapperOperation resource,
 
             if (isBundleParallelOperation)
             {
+                // Parallel operations:
+                // - EnlistTransaction: should be always false, and rely on SQL transactions.
+
                 IBundleOrchestratorOperation bundleOperation = _bundleOrchestrator.GetOperation(resource.BundleResourceContext.BundleOperationId);
                 SetAndClearPendingSearchParameterStatus(resource);
```
SergeyGaluzo, line 894: It looks like SetAndClearPendingSearchParameterStatus(resource); can be moved outside any ifs to line 878. Do you mind moving it there?

fhibf: let's do it!

LTA-Thinking (dismissed):

LTA-Thinking (approved):

PR #5617, opened 2026-06-17
Fix CapabilityStatement conformance for resources

v-rachitsh:
## Summary

Fixes #5394. External PR #5473

In STU3, `CapabilityStatement.rest.resource.interaction` has a **minimum cardinality of 1** ([spec](https://www.hl7.org/fhir/STU3/capabilitystatement.html)). Resources like `Parameters` that are explicitly excluded from interaction population (line 303-307) were still appearing in the resource list with an empty interaction array, causing FHIR-conformant parsers (e.g., Firely SDK 6.x with strict validation) to reject the CapabilityStatement.

## Changes

- In `CapabilityStatementBuilder.Build()`, added a cleanup step that removes resource entries with no interactions before serialization
- This ensures the output CapabilityStatement conforms to the FHIR standard across all versions

## Why this approach

Rather than modifying how `Parameters` is handled specifically, this cleanup catches *any* resource that ends up with zero interactions — making it robust against future edge cases. The comment explains the STU3 cardinality requirement for maintainability.

## Related issues
[AB#191808](https://dev.azure.com/microsofthealth/Health/_workitems/edit/191808)

## Testing

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Conformance/CapabilityStatementBuilder.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Conformance/CapabilityStatementBuilder.cs
@@ -407,15 +407,18 @@ namespace Microsoft.Health.Fhir.Core.Features.Conformance
             // Remove resource entries with no interactions to ensure FHIR conformance.
             // In STU3, CapabilityStatement.rest.resource.interaction has min cardinality 1,
             // so resources without interactions (e.g. Parameters) must be excluded.
-            foreach (var restComponent in _statement.Rest)
+            if (_modelInfoProvider.Version == FhirSpecification.Stu3)
             {
-                var emptyResources = restComponent.Resource
-                    .Where(r => r.Interaction == null || r.Interaction.Count == 0)
-                    .ToList();
-
-                foreach (var resource in emptyResources)
+                foreach (var restComponent in _statement.Rest)
                 {
-                    restComponent.Resource.Remove(resource);
+                    var emptyResources = restComponent.Resource
+                        .Where(r => r.Interaction == null || r.Interaction.Count == 0)
+                        .ToList();
+
+                    foreach (var resource in emptyResources)
+                    {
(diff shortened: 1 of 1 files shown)


src/Microsoft.Health.Fhir.Core/Features/Conformance/CapabilityStatementBuilder.cs
```diff
@@ -404,6 +404,21 @@ public async Task<ICapabilityStatementBuilder> SyncProfilesAsync(CancellationTok
 
         public ITypedElement Build()
         {
+            // Remove resource entries with no interactions to ensure FHIR conformance.
+            // In STU3, CapabilityStatement.rest.resource.interaction has min cardinality 1,
+            // so resources without interactions (e.g. Parameters) must be excluded.
```
LTA-Thinking, line 409: @EXPEkesheth this could technically be called a breaking change. I think it is a correct change, and it should be non-impactful as anything removed in this filter would have no request types. What are your thoughts?

src/Microsoft.Health.Fhir.Core/Features/Conformance/CapabilityStatementBuilder.cs
```diff
@@ -404,6 +404,21 @@ public async Task<ICapabilityStatementBuilder> SyncProfilesAsync(CancellationTok
...
+            foreach (var restComponent in _statement.Rest)
```
LTA-Thinking: @v-rachitsh I know this is an external PR, but could you add an if statement here so this only runs if the _modelInfoProvider.Version is Stu3? This isn't an issue in other FHIR versions.

v-rachitsh: @LTA-Thinking Wrapped up in if statement as you have suggested

LTA-Thinking (approved):

PR #5624, opened 2026-06-18
Fix temporal rewriter chained union SQL generation

mikaelweave:
## Summary

Fixes issuse with the Date of Birth query optimization mainly with observed regressions around chaining and search.

This fixes the root SQL generator issuse vs skipping using the rewriter for certain searces.

The majority of the work here is tracking when a union is happening from the DOB rewriter and ensuring we apply the logic correctly when generating SQL.

## Test Plan
- Added unit and E2E Tests


src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -1678,45 +1678,31 @@ private void AppendIntersectionWithPredecessorUsingInnerJoin(IndentedStringBuild
             }
         }
 
-        private int FindRestrictingPredecessorTableExpressionIndex()
+        private int FindRestrictingPredecessorTableExpressionIndex(SearchParamTableExpressionKind currentKind)
```
mikaelweave: The brittle recursive index-walker was replaced with simple, signature-driven counter arithmetic.

Concatenation was in issue in the past with Union CTEs.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/SortRewriter.cs
```diff
@@ -156,6 +156,22 @@ public override Expression VisitSqlRoot(SqlRootExpression expression, SqlSearchO
...
+        /// Propagates the sort-parameter "found" signal through a <see cref="UnionExpression"/>.
+        /// Returns <c>null</c> (the found signal) only when <b>every</b> branch of the union
+        /// guarantees the sort parameter is present — i.e. every resource that matches the union
+        /// is guaranteed to have a sort value and the missing-value sort phases can be skipped.
+        /// If any branch does not contain the sort parameter the original expression is returned
+        /// unchanged. The original reference is always returned rather than reconstructing the
+        /// union, because the <see cref="UnionExpression"/> constructor rejects null children and
+        /// would throw if any branch visit returned null.
+        /// </summary>
+        public override Expression VisitUnion(UnionExpression expression, SqlSearchOptions context)
```
mikaelweave: The ScalarTemporalEqualityRewriter now emits a UnionExpression, a shape SortRewriter never saw before. We need this so the base visitor can handle correctly.

PR #5629, opened 2026-06-20
Correct delete search param workflow

SergeyGaluzo:
Fixes https://microsofthealth.visualstudio.com/Health/_workitems/edit/190743
User story has good explanation. 
As far as implementation, I decided to keep as much of current pipeline logic as possible and remove resources from being deleted at the last stage in merge. This approach allows to keep parallelism for search param deletes in bundles. Now, merge does not include resources for search params with pending delete and pending hard delete statuses.
I originally planned to temporarily keep current logic to find resources for URLs based on full search param scan (handles current logic when resource deletes happen in the same SQL transaction with status updated). This should cover cases when current pending deletes are not yet processed by reindex. After discussing with Jared, fallback logic was removed.


src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Delete/DeletionService.cs
```diff
@@ -118,6 +118,19 @@ public async Task<ResourceKey> DeleteAsync(DeleteResourceRequest request, Cancel
                     version = result?.Wrapper.Version;
                     break;
                 case DeleteOperation.HardDelete:
+                    if (key.ResourceType == KnownResourceTypes.SearchParameter)
+                    {
+                        var resourceWrapper = await fhirDataStore.GetAsync(key, cancellationToken);
+                        if (resourceWrapper != null && !resourceWrapper.IsDeleted)
```
feordin, line 124: Does this need a check for soft deleted resources?  If someone wishes to hard delete after a soft delete has already been done?

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Delete/DeletionService.cs
```diff
@@ -118,6 +118,19 @@ public async Task<ResourceKey> DeleteAsync(DeleteResourceRequest request, Cancel
...
+                            await _retryPolicy.ExecuteAsync(async () => await _searchParameterOperations.DeleteSearchParameterAsync(resourceWrapper.RawResource, cancellationToken, ignoreSearchParameterNotSupportedException: true, isHardDelete: true));
+                        }
+
+                        break;
+                    }
+
+                    await _retryPolicy.ExecuteAsync(async () => await fhirDataStore.HardDeleteAsync(key, false, request.AllowPartialSuccess, cancellationToken));
+                    break;
                 case DeleteOperation.PurgeHistory:
                     await _retryPolicy.ExecuteAsync(async () => await fhirDataStore.HardDeleteAsync(key, request.DeleteOperation == DeleteOperation.PurgeHistory, request.AllowPartialSuccess, cancellationToken));
```
LTA-Thinking: Nit: the purge history argument in this call can just be true now.

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Delete/DeletionService.cs
SergeyGaluzo: If customers change their mind, fine. The only thing matters is what the status is when reindex completes. If it finds hard delete, it will do that, if soft it will do soft.

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/SearchParameterOperations.cs
```diff
@@ -354,70 +363,173 @@ private void DeleteSearchParameter(string url)
...
-                cancellationToken.ThrowIfCancellationRequested();
+                ["resourceType"] = resourceKey.ResourceType,
+                ["id"] = resourceKey.Id,
+                ["meta"] = new JObject { ["lastUpdated"] = lastModified.UtcDateTime },
+            };
+
+            var rawResource = new RawResource(json.ToString(Newtonsoft.Json.Formatting.None), FhirResourceFormat.Json, isMetaSet: false);
+            var searchParamHash = _searchParameterDefinitionManager.GetSearchParameterHashForResourceType(resourceKey.ResourceType);
+
+            var wrapper = new ResourceWrapper(
```
LTA-Thinking, line 421: Why do this manually instead of using the CreateResourceWrapper method in ResourceWrapperFactoryExtensions?

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlServerSchemaUpgradeTests.cs
```diff
@@ -67,64 +67,36 @@ public async Task GivenTwoSchemaInitializationMethods_WhenCreatingTwoDatabases_B
...
             {
-                await (testHelper1?.DeleteDatabase(snapshotDatabaseName) ?? Task.CompletedTask);
-                await (testHelper2?.DeleteDatabase(diffDatabaseName) ?? Task.CompletedTask);
+                await (snapshotHelper?.DeleteDatabase(snapshotDatabaseName) ?? Task.CompletedTask);
+                await (diffHelper?.DeleteDatabase(diffDatabaseName) ?? Task.CompletedTask);
             }
         }
 
-        [RetryFact]
-        public void GivenASchemaVersion_WhenApplyingDiffTwice_ShouldSucceed()
```
LTA-Thinking, line 97: Why should this test be deleted?

LTA-Thinking (commented):
A few small comments

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/SqlServerSchemaUpgradeTests.cs
SergeyGaluzo: This test requires multiple full versions, which we do not have, so I moved its functionality to the above test.

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Delete/DeletionService.cs
SergeyGaluzo: Fixed

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/SearchParameterOperations.cs
SergeyGaluzo: Fixed.

LTA-Thinking (dismissed):

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/SearchParameterOperations.cs
SergeyGaluzo: I tried fixing but factort.Create() uses request context, which is not properly set in this case. I reverted.

LTA-Thinking (dismissed):

src/Microsoft.Health.Fhir.Shared.Core/Features/Resources/Delete/DeletionService.cs
feordin: I was thinking of the the check: if (resourceWrapper != null && !resourceWrapper.IsDeleted)    If a resource is Soft Deleted, then resoourceWrapper.IsDeleted will be true.  So we won't enter the if block.  If a customer attempts to hard delete an already soft deleted resource, will it fail?  Because of this if statement?

SergeyGaluzo: In old code we were just calling hard delete and returning **success**. 
In new code we are checking if resource is there and call delete logic. If resource is not there, NOOP, we return **success**. If resource is there, invoke delete, return **success**.
No failures...
The reason for getting resource first is kinda historical.
Do you believe we should hard delete regardless of the resource state?

feordin (approved):

PR #5641, opened 2026-06-25
Fix temporal rewriter chained union SQL generation

mikaelweave:
## Summary

Fixes issuse with the Date of Birth query optimization mainly with observed regressions around chaining and search.

This fixes the root SQL generator issuse vs skipping using the rewriter for certain searces.

The majority of the work here is tracking when a union is happening from the DOB rewriter and ensuring we apply the logic correctly when generating SQL.

## Test Plan
- Added unit and E2E Tests


src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/ScalarTemporalEqualityRewriter.cs
```diff
@@ -50,24 +50,8 @@ private enum Precision
             ExactDay,
         }
 
-        public override Expression VisitChained(ChainedExpression expression, bool context)
```
mikaelweave, line 53: We don't need to skip chained / reverse chained expressions anymore as the root issue was fixed.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/SearchParamTableExpressionExtensions.cs
```diff
@@ -20,8 +20,9 @@ internal static class SearchParamTableExpressionExtensions
         /// <param name="expression">Instance of <see cref="SearchParamTableExpression"/> under evaluation.</param>
         public static bool HasUnionAllExpression(this SearchParamTableExpression expression)
         {
-            IExpressionsContainer expressionContainer = expression.Predicate as IExpressionsContainer;
-            return expressionContainer?.Expressions.Any(e => e is UnionExpression) ?? false;
+            return expression.Predicate is UnionExpression ||
```
mikaelweave, line 23: tiny change to check for `Predicate is UnionExpression`. This is a new path for chained expressions with unions from the rewriter.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/SearchParamTableExpressionExtensions.cs
```diff
@@ -36,6 +37,12 @@ public static bool SplitExpressions(this SearchParamTableExpression expression,
             unionExpression = null;
             allOtherRemainingExpressions = null;
 
+            if (expression.Predicate is UnionExpression bareUnion)
```
mikaelweave, line 40: If expression Predicate is a `UnionExpression` itself from the new rewriter w/ chained search, report true.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/SearchParamTableExpressionExtensions.cs
```diff
@@ -86,7 +93,7 @@ public static IReadOnlyList<SearchParamTableExpression> SortExpressionsByQueryLo
 
             foreach (SearchParamTableExpression tableExpression in expressions)
             {
-                if (tableExpression.HasUnionAllExpression())
+                if (tableExpression.HasUnionAllExpression() && tableExpression.ChainLevel == 0)
```
mikaelweave, line 96: We should not sort expressions that are chained unions (from the new rewriter) as a chain-nested union must stay where it is so it can be lowered after its chain link.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/SortRewriter.cs
```diff
@@ -156,6 +156,9 @@ public override Expression VisitSqlRoot(SqlRootExpression expression, SqlSearchO
             return new SqlRootExpression(newTableExpressions, expression.ResourceTableExpressions);
         }
 
+        public override Expression VisitUnion(UnionExpression expression, SqlSearchOptions context) =>
```
mikaelweave, line 159: A new `VisitUnion` override. A union is "sortable" (returns null, the visitor's signal to drop the node) only when every branch is itself sortable; otherwise the union is returned unchanged so a `SortWithFilter` is emitted.

Without this, a union node would fall through to default handling and mis-classify the query's sortability.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -171,7 +172,7 @@ public override object VisitSqlRoot(SqlRootExpression expression, SearchOptions
                         }
                         else
                         {
-                            AppendNewSetOfUnionAllTableExpressions(context, unionExpression, tableExpression.QueryGenerator);
+                            AppendNewSetOfUnionAllTableExpressions(context, unionExpression, tableExpression.QueryGenerator, tableExpression.ChainLevel);
```
mikaelweave, line 183: We need to track chain level for Union All now. 

`chainLevel == 0` → the union sits at the top level of the query (the original, always-true assumption).
`chainLevel > 0` → the union is nested inside a chained search (e.g. `MedicationDispense?patient:Patient.birthdate=…`). This is the case the old code never handled.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -1422,31 +1437,100 @@ private SearchParameterQueryGeneratorContext GetContext(string tableAlias = null
...
+        }
+
+        // Emits one Normal CTE per branch of a chain-nested union. Each branch must restrict against the
+        // chain link emitted immediately before the union rather than against the previous branch, so we
+        // pin that predecessor for the duration of the loop (FindRestrictingPredecessorTableExpressionIndex
+        // reads _unionBranchPredecessorIndex) and restore it in finally so nested unions unwind cleanly.
+        private void AppendChainNestedUnionBranches(SearchOptions context, UnionExpression unionExpression, SearchParamTableExpressionQueryGenerator defaultQueryGenerator, int chainLevel, int firstInclusiveTableExpressionId)
+        {
+            int previousUnionBranchPredecessorIndex = _unionBranchPredecessorIndex;
+            _unionBranchPredecessorIndex = firstInclusiveTableExpressionId - 1;
```
mikaelweave, line 1486: If we're inside a chain-nested union, each branch must restrict against the chain link emitted just *before* the union — not against the previous branch. So we pin that predecessor (`_unionBranchPredecessorIndex`) for the duration of the loop and restore it in `finally` so nested unions unwind cleanly.

(`FindRestrictingPredecessorTableExpressionIndex` reads this pinned value.)

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -1422,31 +1437,100 @@ private SearchParameterQueryGeneratorContext GetContext(string tableAlias = null
...
                 throw new ArgumentOutOfRangeException(unionExpression.Operator.ToString());
             }
 
-            // Iterate through all expressions and create a unique CTE for each one.
             int firstInclusiveTableExpressionId = _tableExpressionCounter + 1;
+            bool isChainNestedUnion = chainLevel > 0;
+
+            // Emit one CTE per union branch. Chain-nested unions emit Normal CTEs restricted by their
+            // pinned chain-link predecessor; top-level unions emit Union CTEs via the normal visitor path.
+            if (isChainNestedUnion)
```
mikaelweave, line 1452: Two emission strategies, now split into named helpers:

- `AppendTopLevelUnionBranches` — free-floating union members go through the normal `Union` visitor path.
- `AppendChainNestedUnionBranches` — each branch is emitted directly as an ordinary restricting (`Normal`) CTE that chains off the pinned predecessor, exactly how a normal chain link behaves.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -1422,31 +1437,100 @@ private SearchParameterQueryGeneratorContext GetContext(string tableAlias = null
...
+                AppendTopLevelUnionBranches(context, unionExpression, defaultQueryGenerator, chainLevel);
+            }
+
+            // Emit the CTE that UNION ALLs every branch CTE together.
+            AppendUnionAggregateCte(firstInclusiveTableExpressionId);
+
+            // Chain-nested unions are restricted by their pinned chain-link predecessor, not by the
+            // top-level union state machine, so they stop here without joining a previous union or
+            // mutating the shared union state.
+            if (isChainNestedUnion)
```
mikaelweave, line 1467: `JoinWithPreviousUnionAggregate` and the union flags below assume the aggregate is the final thing the outer query joins on. In a chained query the union sits in the *middle*, so running those mutated shared state and produced the broken CTE. Chain-nested unions are pinned to their predecessor instead, so we return early here.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/QueryGenerators/SqlQueryGenerator.cs
```diff
@@ -1671,52 +1754,33 @@ private void AppendIntersectionWithPredecessorUsingInnerJoin(IndentedStringBuild
...
-                    case SearchParamTableExpressionKind.IncludeLimit:
-                    case SearchParamTableExpressionKind.Union:
-                    case SearchParamTableExpressionKind.IncludeUnionAll:
-                        return currentIndex - 1;
-                    default:
-                        throw new ArgumentOutOfRangeException(currentSearchParamTableExpression.Kind.ToString());
-                }
+            // Restrict against the immediately previous CTE (_tableExpressionCounter - 1). We track against
+            // _tableExpressionCounter rather than root-expression indices because unions inject extra
+            // aggregate CTEs that those logical indices don't account for.
```
mikaelweave, line 1773: This used to be much more complex. The old recursive `FindImpl` indexed into `_rootExpression.SearchParamTableExpressions`, but CTE numbering uses `_tableExpressionCounter`, which diverges once a union injects extra aggregate CTEs — two coordinate systems, one bug. This version resolves purely from emitted CTE order, with a pinned override (`_unionBranchPredecessorIndex`) for chain-nested branches and the `-2` `Concatenation` case documented just above.

mikaelweave:
Closing in favor of #5642, the smaller alternative that achieves the same fix and results. Both PRs fix the same bug (chained/reverse-chained exact-day birthdate emitting a UNION ALL that broke SQL generation) with no correctness difference. We chose #5642 because it restores the chained-search guard in ScalarTemporalEqualityRewriter rather than teaching SqlQueryGenerator to lower chain-nested unions, which removes the stateful chain-aware machinery (SqlQueryGenerator 62 vs 178 changed lines; 10 files / +802 vs 12 files / +1016). Data-driven justification: the only behavior #5642 concedes is the IsLongerThanADay index-split on the chained birthdate leaf, which is functionally equivalent for single-day birthdate (the range fallback selects the identical rows), and the split is still preserved on the high-volume top-level birthdate path. See #5642.

PR #5642, opened 2026-06-25
Fix temporal rewriter ignoring chained searches

mikaelweave:
## Summary

Fixes issuse with the Date of Birth query optimization mainly with observed regressions around chaining and sort in search.

This is a **smaller alternative to #5641**. It fixes the same bug — exact-day birthdate equality (e.g. `birthdate=1980-01-01`) being rewritten by `ScalarTemporalEqualityRewriter` into a `UNION ALL` that broke **chained / reverse-chained** searches (they returned 0 rows) — but with far less production code.

### The two approaches

- **#5641** keeps the rewriter active inside chained expressions and teaches `SqlQueryGenerator` how to correctly lower a *chain-nested* union (chain-aware union branches, predecessor pinning on the chain target, etc.). That's ~178 changed lines in `SqlQueryGenerator.cs` alone.
- **This PR** instead **restores the original guard** that simply *skips* the rewriter inside chained expressions. If chained searches never produce a union, the chain-specific union machinery becomes dead code and is removed. Chained exact-day birthdate falls back to the normal datetime-range predicate — **correct results**, just without the `IsLongerThanADay` union perf optimization in the chained context (exactly how it behaved before the rewriter existed).

## Test Plan
- Added unit and E2E Tests

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/ResourceColumnPredicatePushdownRewriter.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/ResourceColumnPredicatePushdownRewriter.cs
@@ -86,15 +86,49 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Search.Expressions.Visitors
                 }
                 else
                 {
-                    Expression predicate = tableExpression.Predicate == null
-                        ? extractedCommonResourceExpressions
-                        : Expression.And(tableExpression.Predicate, extractedCommonResourceExpressions);
+                    Expression predicate;
+
+                    if (extractedCommonResourceExpressions != null
+                        && tableExpression.Predicate is UnionExpression unionPredicate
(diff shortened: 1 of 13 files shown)


src/Microsoft.Health.Fhir.SqlServer/Features/Search/Expressions/Visitors/ResourceColumnPredicatePushdownRewriter.cs
```diff
@@ -86,15 +86,49 @@ public override Expression VisitSqlRoot(SqlRootExpression expression, object con
                 }
                 else
                 {
-                    Expression predicate = tableExpression.Predicate == null
-                        ? extractedCommonResourceExpressions
-                        : Expression.And(tableExpression.Predicate, extractedCommonResourceExpressions);
+                    Expression predicate;
+
+                    if (extractedCommonResourceExpressions != null
```
mikaelweave, line 91: This logic is not 100% needed but creates the correct query shape we want.

Image

PR #5648, opened 2026-06-28
Move check for "running reindex" in the data store

SergeyGaluzo:
Fixes https://microsofthealth.visualstudio.com/Health/_workitems/edit/186796/


PTaladay (dismissed):

rajithaalurims:
**Backward- and forward-compatible schema changes**
The old code works with the new schema changes as expected, which covers our rollback scenario.
However, new code with the old schema changes will silently skip the Reindex check.
Since our rolling release leaves us in a state where pods running new code need to work with the old schema, I think we need version-gated behavior for this change.

**Change in customer visible error message:**
Before PR: Changes to search parameters is not allowed while a reindex job is ongoing.  Wait for the reindex job with Id: {jobId} to finish, or cancel it.
After PR : Reindex job is in progress. Job Id=12345
Could we replace the SQL message with the resource string so the error message stays the same?

**Change in when customer receives the error:**
For example when customer runs Bulk delete, it used to fail with 409 at the controller level and return 409 right away. Now it starts the job (returns 202 Accepted), and the job itself fails later with 409 status and a conflict issue in the OperationOutcome. This changes when the error surfaces. Do we need to announce the behavior change? What if the customer was only listening to the bulk update API response and not depending on the actual operation outcome?

SergeyGaluzo:
> **Backward- and forward-compatible schema changes** The old code works with the new schema changes as expected, which covers our rollback scenario. However, new code with the old schema changes will silently skip the Reindex check. Since our rolling release leaves us in a state where pods running new code need to work with the old schema, I think we need version-gated behavior for this change.
> 
> **Change in customer visible error message:** Before PR: Changes to search parameters is not allowed while a reindex job is ongoing.  Wait for the reindex job with Id: {jobId} to finish, or cancel it. After PR : Reindex job is in progress. Job Id=12345 Could we replace the SQL message with the resource string so the error message stays the same?
> 
> **Change in when customer receives the error:** For example when customer runs Bulk delete, it used to fail with 409 at the controller level and return 409 right away. Now it starts the job (returns 202 Accepted), and the job itself fails later with 409 status and a conflict issue in the OperationOutcome. This changes when the error surfaces. Do we need to announce the behavior change? What if the customer was only listening to the bulk update API response and not depending on the actual operation outcome?

1. Compatibility. Old schema is also checking for reindex conflicts, so we should be covered. New schema is accepting 0 reindex id and does the checks too.
2. I will replace the message with old one.
3. I doubt that there are real customers, who use bulk delete to delete search parameters. If there are, we should announce before, if not - we can just update in release notes. I might need some help in determining that.

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkDelete/BulkDeleteProcessingJob.cs
```diff
@@ -106,8 +98,36 @@ public async Task<string> ExecuteAsync(JobInfo jobInfo, CancellationToken cancel
                 catch (IncompleteOperationException<IDictionary<string, long>> ex)
                 {
                     resourcesDeleted = ex.PartialResults;
-                    result.Issues.Add(ex.Message);
-                    exception = ex;
+                    bool conflictFound = false;
```
LTA-Thinking, line 102: This same change needed in the Bulk Update processing job. We didn't add the old SP protection to it (which was a miss), but this catch logic would be good to have there.

LTA-Thinking (dismissed):

src/Microsoft.Health.Fhir.Core/Features/Operations/BulkDelete/BulkDeleteProcessingJob.cs
SergeyGaluzo: AFAIK SPs are explicitly excluded from bulk update, so reindex conflicts are not possible.

LTA-Thinking (dismissed):

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandlerParallelOperations.cs
```diff
@@ -132,6 +133,22 @@ private async Task<EntryComponent> ExecuteRequestsInParallelAsync(
                         // If the exception raised is a OperationCanceledException, then either client cancelled the request or httprequest timed out.
                         _logger.LogWarning(ex, "Bundle request timed out. Elapsed time {TotalElapsedMilliseconds}ms. Error: {ErrorMessage}", watch.Elapsed.TotalMilliseconds, ex.Message);
                     }
+                    catch (JobConflictException ex) when (bundleExecutionContext.BundleType == BundleType.Batch)
```
fhibf, line 136: I need to read the rest of the PR, but I'm curious why this is limited to batches at this moment. 
Probably reading the rest of the PR I'll find the answer. Leaving this comment until then.

SergeyGaluzo: For batch we cannot raise exception on batch level, we need to convert into resource level problem. For non-batch we just send exception to the client.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Registry/CosmosDbSearchParameterStatusDataStore.cs
```diff
@@ -96,6 +102,17 @@ public async Task UpsertStatuses(IReadOnlyList<ResourceSearchParameterStatus> st
...
 
+            // Check for active reindex jobs unless this call is from a reindex job itself
+            if (!reindexId.HasValue || reindexId.Value <= 0)
+            {
+                var activeJobs = await _queueClient.GetActiveJobsByQueueTypeAsync((byte)QueueType.Reindex, returnParentOnly: true, cancellationToken);
+                if (activeJobs.Count > 0)
+                {
+                    var jobId = activeJobs[0].Id;
+                    throw new Fhir.Core.Features.Operations.JobConflictException(string.Format(Fhir.Core.Resources.ChangesToSearchParametersNotAllowedWhileReindexing, jobId));
+                }
```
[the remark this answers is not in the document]
SergeyGaluzo: This is not correct. Full names space is added to clearly state which JobConflictException is raised, because there are 2 options.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Registry/CosmosDbSearchParameterStatusDataStore.cs
```diff
@@ -13,28 +13,34 @@
...
         private readonly ICosmosQueryFactory _queryFactory;
+        private readonly IQueueClient _queueClient;
         private readonly SemaphoreSlim _statusListSemaphore = new(1, 1);
 
         public CosmosDbSearchParameterStatusDataStore(
             Func<IScoped<Container>> containerScopeFactory,
             CosmosDataStoreConfiguration cosmosDataStoreConfiguration,
-            ICosmosQueryFactory queryFactory)
+            ICosmosQueryFactory queryFactory,
+            IQueueClient queueClient)
```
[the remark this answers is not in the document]
SergeyGaluzo: Fixed.

LTA-Thinking (dismissed):

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Persistence/CosmosDbFhirStorageTestsFixture.cs
```diff
@@ -248,10 +249,15 @@ public virtual async Task InitializeAsync()
...
+            var mockQueueClientForSearchParams = Substitute.For<IQueueClient>();
+            mockQueueClientForSearchParams.GetActiveJobsByQueueTypeAsync(Arg.Any<byte>(), Arg.Any<bool>(), Arg.Any<CancellationToken>())
+                .Returns(Task.FromResult((IReadOnlyList<JobInfo>)Array.Empty<JobInfo>()));
+
             _searchParameterStatusDataStore = new CosmosDbSearchParameterStatusDataStore(
                 () => documentClient,
                 _cosmosDataStoreConfiguration,
-                cosmosDocumentQueryFactory);
+                cosmosDocumentQueryFactory,
+                mockQueueClientForSearchParams);
```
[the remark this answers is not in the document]
SergeyGaluzo: Fixed

LTA-Thinking (dismissed):

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -43,6 +43,7 @@
 using Microsoft.Health.Fhir.Core.Features;
 using Microsoft.Health.Fhir.Core.Features.Context;
 using Microsoft.Health.Fhir.Core.Features.Definition;
+using Microsoft.Health.Fhir.Core.Features.Operations;
```
fhibf: nit: this using statement can be removed.

SergeyGaluzo: Fixed.

fhibf (approved):

rajithaalurims (approved):

PR #5659, opened 2026-07-07
Add feature flag gating 3-column reference-type filtered stats

apurvabhaleMS:
Related work item: [AB#197291](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/197291)

## What this does

- Puts the new 3-column filtered stat on `dbo.ReferenceSearchParam` (`ResourceTypeId` + `SearchParamId` + `ReferenceResourceTypeId`) behind a new flag `Search.ReferenceResourceTypeFilteredStats.IsEnabled`, defaulting to **off**. When off, typed reference searches use the existing 2-column stat.
- Adds a concurrency limit and a circuit breaker around the slow-query diagnostic logger (`FireAndForgetQueryStoreLookup`) so it can't overload the database.

## Why

OSS-CI R4 SQL E2E started timing out at 120 min with SQL CPU at 100%. It was a feedback loop:

new 3-column stat → plan cache invalidated → giant compartment/chained queries recompile → compile time (35-41s) pushes the query over the 5s "long-running" threshold → diagnostic logger fires → the logger's `LIKE` scan over `sys.query_store_query_text` is expensive → CPU saturates → more queries cross 5s → more logger calls → snowball.

The logger query alone used ~2.27 hours of CPU (8,166 executions), which is the 2-hour timeout.

## Numbers (both runs had the logger flag on)

| Metric | 4.0.804 (passed) | 4.0.810 (timed out) |
|---|---|---|
| Long-running SQL events | 3 (whole run) | 2,552 (first 88 min) |
| p50 | 5,789 ms | 17,579 ms |
| p95 | 9,767 ms | 36,035 ms |
| max | 9,767 ms | 102,891 ms |
| queries > 10s | 0 | 2,489 |
| queries > 20s | 0 | 1,037 |
| compile time (compartment/chained) | ~47 ms | 35,000-41,000 ms |

## Testing

- Unit tests for the circuit breaker (open / cooldown / probe transitions).
- Integration test covering the 2-column fallback (flag off) and 3-column path (flag on).

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
+++ b/src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
@@ -1269,7 +1269,9 @@ namespace Microsoft.Health.Fhir.SqlServer.Features.Search
             // Circuit breaker: if too many consecutive lookups have failed/timed out, the database is
             // likely overloaded. Suspend Query Store enrichment for a cooldown window so diagnostics
             // don't compound the problem. The slow query is still logged below — only the enrichment
-            // is skipped. When the cooldown elapses, exactly one probe lookup is allowed through.
+            // is skipped. When the cooldown elapses, the breaker closes and lookups resume (bounded by
+            // the concurrency gate below); the failure counter stays elevated, so the next failure
+            // re-opens the breaker while any success fully resets it.
             if (!TryEnterQueryStoreCircuit())
             {
                 _logger.LogWarning(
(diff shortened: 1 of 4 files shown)


apurvabhaleMS:
Design note for reviewers: the diagnostic Query Store enrichment is now protected by two independent mechanisms. The `SemaphoreSlim(MaxConcurrentQueryStoreLookups)` try-or-skip gate is the essential fix for the CI CPU-saturation incident — it hard-caps concurrent diagnostic connections so a request burst can never open more than N extra SQL connections. The circuit breaker on top is intentional defense-in-depth: after `QueryStoreCircuitBreakerFailureThreshold` consecutive failures it stops attempting lookups for a short cooldown, so a persistently unhealthy DB doesn't keep paying the diagnostic cost. Both are best-effort and only skip the *enrichment*; the long-running-query warning itself is always logged. A couple of things to keep in mind: (1) the breaker state is process-global (static, per-pod) by design; (2) `QueryStoreCircuitBreakerFailureThreshold` must stay a `const` because the unit-test `[Theory]` references it in `[InlineData]`.

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -58,6 +58,16 @@ internal class SqlServerSearchService : SearchService
...
+
+        /// <summary>
+        /// Feature flag that gates creation of the 3-column reference-type filtered statistic on the
+        /// ReferenceSearchParam table (the stat that additionally filters on ReferenceResourceTypeId,
+        /// i.e. <c>WHERE ResourceTypeId = .. AND SearchParamId = .. AND ReferenceResourceTypeId = ..</c>).
+        /// Off by default: unless a row exists in the Parameters table with this Id set to an enabled
+        /// value, these stats are not created and reference searches fall back to the broader 2-column
+        /// filtered statistic (ResourceTypeId + SearchParamId).
+        /// </summary>
+        internal const string ReferenceResourceTypeFilteredStatsParameterId = "Search.ReferenceResourceTypeFilteredStats.IsEnabled";
```
rajithaalurims, line 70: nit: It looks like Search.ReferenceResourceTypeFilteredStats.IsEnabled is currently being read from the db.Parameters table. Could account overrides be a better way to avoid DB access when enabling the feature?

rajithaalurims (approved):

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
apurvabhaleMS: Ideally yes. In this particular case while toggling the flag we might have to drop the stats too which would anyways require dB access. Keeping it as is for now. Also, the Geneva action for override update is not in Prod yet, can move flag to override once it can be easily toggled or set

PR #5660, opened 2026-07-07
Pin Az 5.5.0/6.5.0 and prune newer agent-shipped versions

apurvabhaleMS:
[AB#197379](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/197379)

The `Setup AAD Test Environment` step in the PR pipeline started failing on `Connect-MgGraph`. This turned out to be two separate assembly conflicts, both triggered by new Az PowerShell module versions published on 2026-07-07. Verified fixed in build 50321 (Setup AAD Test Environment: succeeded).

## Symptom

```
Installing Az modules...
Connecting to Microsoft Graph...
Connect-MgGraph : Entry point was not found.
```

Build, unit tests, and Docker images all pass - only the AAD environment setup breaks, which then cascades so the downstream Create Resource Group / Key Vault / test-provisioning steps are marked failed or skipped.

## Root cause (two conflicts)

`build/jobs/add-aad-test-environment.yml` loads Microsoft Graph modules and Az modules (`Az.Accounts`, `Az.KeyVault`) into the same PowerShell process, then later calls into a shared script (`Add-AadTestAuthEnvironment.ps1`) that uses `Register-SecretVault` with the `Az.KeyVault` SecretManagement extension. On 2026-07-07 new Az builds shipped and exposed two independent shared-assembly clashes:

1. **Connect-MgGraph vs Az.Accounts 5.5.1.** `Az.Accounts 5.5.1` (published 2026-07-07) ships a shared assembly that is incompatible with `Microsoft.Graph.Authentication`, so `Connect-MgGraph` throws `EntryPointNotFoundException`. `Az.Accounts 5.5.0` (yesterday's version) does not.

2. **Register-SecretVault vs highest installed Az.KeyVault.** SecretManagement resolves the `Az.KeyVault` extension to the **highest version installed**, not the version we imported. The hosted agent image itself now ships `Az.KeyVault 6.5.1 / Az.Accounts 5.5.1`, so even after pinning to 5.5.0/6.5.0, `Register-SecretVault` picked up 6.5.1 and failed with:

```
Register-SecretVault : Could not load ... Az.KeyVault ...
Assembly with same name is already loaded.
```

So pinning `Az.Accounts` down to 5.5.0 fixes conflict #1 but exposes conflict #2, because the newer build is still present on the agent for SecretManagement to auto-resolve to.

### Version reference

| Module | Yesterday (worked) | Today (broke) | Published |
| --- | --- | --- | --- |
| Az.Accounts | 5.5.0 | 5.5.1 | 5.5.1 on 2026-07-07 |
| Az.KeyVault | 6.5.0 | 6.5.1 | 6.5.1 on 2026-07-07 |
| Microsoft.Graph.Authentication | 2.38.0 | 2.38.0 | unchanged (2026-06-16) |

Graph 2.38.0 was already "latest" yesterday and worked. It is not the regression; the Az modules are.

## Change

`build/jobs/add-aad-test-environment.yml`:
- Pin `Az.Accounts` to 5.5.0 and `Az.KeyVault` to 6.5.0 (fixes conflict #1).
- After install, remove any `Az.Accounts` / `Az.KeyVault` versions higher than the pin from the module path, so `Register-SecretVault` can only resolve to the pinned 6.5.0 (fixes conflict #2).
- Keep Graph pinned at 2.38.0 (proven compatible with the 5.5.0 Az set).
- Version log lines use `Get-Module` (imported version) instead of `Get-Module -ListAvailable`.

`build/cleanup-aad-resources.yml` and `build/jobs/cleanup-aad.yml`:
- Graph-only scripts (they intentionally avoid Az). Graph pinned to 2.38.0 for consistency; they never hit either conflict.

## Verification

Both conflicts were reproduced and the fix confirmed locally before pushing (both Az version sets were installed to mirror the agent):
- Reproduced conflict #2: with 5.5.0 imported but 6.5.1 also present, the unqualified `Register-SecretVault` fails with "Assembly with same name is already loaded".
- Confirmed the fix: after removing the higher versions, the same `Register-SecretVault` call succeeds.
- Ran the full install -> remove-higher -> import -> Register-SecretVault sequence end to end: PASS.

Pipeline build 50321: `Setup AAD Test Environment` completed successfully.

Note: this pins to a known-good set and prunes newer builds, so the Az versions will need an occasional manual bump. That is the correct trade-off for CI - deterministic builds that do not break when a new module release ships.

### DIFF
--- a/build/cleanup-aad-resources.yml
+++ b/build/cleanup-aad-resources.yml
@@ -64,11 +64,10 @@ stages:
             "Microsoft.Graph.Identity.DirectoryManagement"
           )
 
-          # Pin all Graph submodules to a single version. Without a pinned version each submodule
-          # resolves to "latest" independently (or a stale copy is imported), producing a mismatched
-          # set that fails Connect-MgGraph with EntryPointNotFoundException. 2.33.0 is the last version
-          # proven compatible with the Az/agent assemblies (newer releases such as 2.38.0 conflict).
-          $graphVersion = '2.33.0'
+          # Pin all Graph submodules to a single version so the installed and imported set always
+          # match (floating to "latest" independently, or importing a stale copy pre-baked on the
+          # agent, can produce a mismatched set that fails Connect-MgGraph with EntryPointNotFoundException).
+          $graphVersion = '2.38.0'
 
           foreach ($module in $modules) {
             Install-Module -Name $module -RequiredVersion $graphVersion -Force -Scope CurrentUser -ErrorAction Stop
--- a/build/cleanup-aad-resources.yml
+++ b/build/cleanup-aad-resources.yml
@@ -66,8 +66,9 @@ stages:
 
           # Pin all Graph submodules to a single version. Without a pinned version each submodule
           # resolves to "latest" independently (or a stale copy is imported), producing a mismatched
-          # set that fails Connect-MgGraph with EntryPointNotFoundException.
-          $graphVersion = '2.38.0'
+          # set that fails Connect-MgGraph with EntryPointNotFoundException. 2.33.0 is the last version
+          # proven compatible with the Az/agent assemblies (newer releases such as 2.38.0 conflict).
+          $graphVersion = '2.33.0'
 
           foreach ($module in $modules) {
             Install-Module -Name $module -RequiredVersion $graphVersion -Force -Scope CurrentUser -ErrorAction Stop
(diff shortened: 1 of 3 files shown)


feordin (approved):

feordin:
If we pin these, how do we track when to update them in the future?

PR #5662, opened 2026-07-08
Parallel bulk update ignores_lastUpdated filter

apurvabhaleMS:
## Description
When `PATCH /{resourceType}/$bulk-update` was called with `_lastUpdated` as
the only search parameter (e.g. `_lastUpdated=gt2026-07-06`), the operation
updated **all** resources instead of only the filtered subset.

**Root cause:** `_lastUpdated` was included in `allowedParams` in
`BulkUpdateOrchestratorJob.ExecuteAsync`. This caused `noOtherSearchParameters`
to evaluate as `true`, sending the job down the fast surrogateId-scan path.
The stored procedure `dbo.GetResourcesByTypeAndSurrogateIdRange` accepts only
surrogate-id range bounds and ignores all other search parameters — so the
`_lastUpdated` filter was silently dropped and every resource of that type
was updated.

## Related issues
Addresses [[AB197545](https://microsofthealth.visualstudio.com/Health/_workitems/edit/197545)].

## Testing

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Operations/BulkUpdate/BulkUpdateOrchestratorJob.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Operations/BulkUpdate/BulkUpdateOrchestratorJob.cs
@@ -91,7 +91,6 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.BulkUpdate
                     // Collect allowed parameter names (case-insensitive)
                     var allowedParams = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
                     {
-                        KnownQueryParameterNames.LastUpdated,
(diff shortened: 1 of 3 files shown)


src/Microsoft.Health.Fhir.Core/Features/Operations/BulkUpdate/BulkUpdateOrchestratorJob.cs
LTA-Thinking, line 232: Can you add the create date filter to this branch?

apurvabhaleMS: It is handled in the CreateProcessingDefinition method

LTA-Thinking (approved):

PR #5674, opened 2026-07-15
Enforce system-only SMART export access

mikaelweave:
## Description

- authorize SMART Bulk Data `$export` create, status, and cancel only through unconstrained `system/` scopes while preserving RBAC and non-SMART behavior
- separate population selection from output authorization: `Patient/$export` requires Patient access, and `Group/{id}/$export` requires Group and Patient access, in addition to every output type
- preserve explicit `_type` subsets and reject scope overreach; when a partial system caller omits `_type`, infer and persist the authorized resource-type subset instead of requiring wildcard access
- revalidate persisted export level, effective `_type`, and completed output types for status/cancel while preserving not-found masking and legacy no-`_type` wildcard behavior
- document the decision and add focused unit, API, integration, and E2E coverage

## Related issues
[AB#197149](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/197149)

## Testing
- Export validator/status/cancel tests on .NET 8 and .NET 9
- API RBAC ordering tests
- SQL create-handler persistence tests
- R4 E2E assembly compilation; focused E2E execution requires local `testconfiguration.json`


src/Microsoft.Health.Fhir.Core/Features/Operations/Security/ExportSmartScopeAuthorizer.cs
```diff
@@ -0,0 +1,222 @@
...
+            var explicitResourceTypes = new List<string>();
+            if (string.IsNullOrWhiteSpace(resourceType))
+            {
+                return explicitResourceTypes;
+            }
+
+            var seen = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+            foreach (string type in resourceType
+                .Split(',', StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries)
+                .Where(type => seen.Add(type)))
```
LTA-Thinking: This feels like a weird way to make a dedupped list. I feel like just using the set and using toList would be easier to understand.

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Controllers/ExportControllerTests.cs
```diff
@@ -155,10 +155,32 @@ await Assert.ThrowsAsync<RequestNotValidException>(() => _exportEnabledControlle
                 anonymizationConfigCollectionReference: null,
                 anonymizationConfigLocation: null,
                 anonymizationConfigFileETag: null,
-                typeParameter: ResourceType.Patient.ToString(),
+                typeParameter: ResourceType.Observation.ToString(),
                 idParameter: "id"));
         }
 
+        [Fact]
+        public async Task GivenPatientInstanceExport_WhenRequested_ThenRequestNotValidExceptionShouldBeThrown()
```
LTA-Thinking: Why was this test added?

mikaelweave: I wasted time during this implementation assuming `Patient/123/$export` was supported. It is not.

QQ - does this cause an issue with the type filter? Is this a valid export? `/$export?_typeFilter=Patient?_id=123`. If so I should re-visit this.

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Features/Operations/Export/CreateExportRequestHandlerTests.cs
```diff
@@ -356,6 +360,48 @@ public async Task GivenDifferentRequestor_WhenCreatingAnExportJob_ThenNewJobShou
...
+            CreateExportResponse response = await _createExportRequestHandler.HandleAsync(request, _cancellationToken);
+
+            Assert.NotNull(response);
+            Assert.NotNull(actualRecord);
+            Assert.Equal("Observation,Patient", actualRecord.ResourceType);
+            _exportSmartScopeAuthorizer.Received(1).AuthorizeCreateAndResolveResourceType(request);
+        }
+
+        [Fact]
+        public async Task GivenNonSmartRequest_WhenCreatingAnExportJob_ThenRequestedResourceTypeIsPersistedWithoutSmartAuthorization()
```
LTA-Thinking: Should we have a test to show that a non smart export without a type filter don't get one added?

src/Microsoft.Health.Fhir.Shared.Api.UnitTests/Controllers/ExportControllerTests.cs
LTA-Thinking: The format would be `/$export?_type=Patient&_typeFilter=Patient%3F_id%3D123`. When using `_typeFilter` a `_type` parameter must be included that matches the `_typeFilter`. That is FHIRs restriction, not ours. Also, the URL encoding in `_typeFilter` is mandatory.

LTA-Thinking (dismissed):

src/Microsoft.Health.Fhir.Core/Features/Operations/Security/ExportSmartScopeAuthorizer.cs
```diff
@@ -0,0 +1,204 @@
...
+        }
+
+        /// <summary>
+        /// Infers the eligible, deterministic output resource types from every resource-specific
+        /// (non-wildcard) unconstrained system scope whose aggregated matching actions provide complete
+        /// export-read access.
+        /// </summary>
+        private static List<string> InferEligibleOutputResourceTypes(ScopeRestriction[] systemScopes)
+        {
+            return systemScopes
```
feordin, line 169: This will return scope values that match resource type names regardless of casing.  But it seems that downstream, in the SQL server data layer, we use a case sensitive match to resources.  Perhaps an E2E test with lowercase scopes to repro?  system.observeration.read

mikaelweave: You are right, this should be a case sensitive check! Great catch.

src/Microsoft.Health.Fhir.Api/Features/SMART/SmartClinicalScopesMiddleware.cs
```diff
@@ -335,8 +335,8 @@ public async Task Invoke(
                             // silently stored as a ScopeRestriction that matches nothing, masking client
                             // misconfiguration. The "all" wildcard is not a concrete resource type, so it is
                             // excluded from the check.
-                            if (!resource.Equals(KnownResourceTypes.All, StringComparison.OrdinalIgnoreCase)
-                                && !ModelInfoProvider.IsKnownResource(resource))
+                            if (!resource.Equals(KnownResourceTypes.All, StringComparison.Ordinal)
```
mikaelweave, line 338: @feordin -- since scopes are case sensitive should we make this change?

mikaelweave: talked with the orig dev -- this change is correct.

feordin (approved):

PR #5685, opened 2026-07-17
Ignixa import phase 0: opt-in $import parsing behind flag

brendankowitz:
## Description
Phase 0 of the incremental Firely → Ignixa SDK migration (see [ADR 2607](docs/arch/adr-2607-ignixa-import-phase0.md) for the full decision record and migration ladder). Adds an opt-in Ignixa implementation of `$import` parsing behind a new `CoreFeatureConfiguration.FhirSdkProvider` setting (`Firely` default, `Ignixa` opt-in on net10 only), while keeping the existing import, indexing, SQL, and Cosmos boundaries completely unchanged.

Deliberately scoped small: only `$import` parsing is touched, the Ignixa parser converts to the existing Firely-shaped `ResourceElement` rather than preserving the native node end-to-end, and conditional-reference/soft-delete detection is intentionally schema- and FHIRPath-driven using genuine native Ignixa APIs (`IFhirSchemaProvider.ReferenceMetadataProvider`, `IElement.Predicate()`, `SourceNodeExtensions.RemoveExtension`) rather than hand-rolled JSON traversal.

> **Performance note — please read before citing this PR's rationale.** Phase 0 delivers **no** `$import` speedup, and is likely a slight regression in Ignixa mode. `RawResourceFactory` still rebuilds a full Firely POCO and serializes through Firely's `FhirJsonSerializer` regardless of which parser produced the resource, so Ignixa mode pays an Ignixa parse *on top of* the same Firely cost. Recovering the win is Phase 2a (the write-side persistence codec), which is scheduled immediately after this PR for exactly that reason. The flip point is marked in code at `IgnixaImportResourceParser.cs`.

Key pieces:
- `FhirSdkProvider` two-state enum + `CoreFeatureConfiguration` property, defaulting to `Firely`.
- Four version-specific Firely provider projects (STU3/R4/R4B/R5) sharing one `FirelyImportResourceParser` source file.
- One `Microsoft.Health.Fhir.Ignixa` project with `IgnixaImportResourceParser` and `IgnixaSchemaContext`. It is functionally net10-only: it declares `net10.0;net8.0`, but net8.0 is a deliberate no-op target (sources behind `#if NET10_0_OR_GREATER`, no Ignixa package references on net8.0) so that solution-wide builds pinning a single framework (`-f net8.0`, used by CI) don't fail with NETSDK1005.
- `OperationsModule` registers exactly one `IImportResourceParser` from the configured provider at startup; Ignixa on net8 fails fast with a clear error.
- Conditional-reference and soft-delete checks are scoped to the resource's own schema-declared fields — matching `TypedElementSearchIndexer`'s existing scope (no recursion into `contained` resources; Bundle entries are out of scope since `$import` NDJSON carries individual resources, not transactional Bundles).

### Migration ladder update
ADR 2607 documents the full ladder (export serialization → persistence codecs → FHIRPath/indexing → HTTP ingress/egress → validation → complex write semantics → Firely removal). This PR also amends it in two ways:

- **Phase 2 is split into 2a (writes) and 2b (reads), and the order is swapped.** The ladder previously listed persistence codecs as reads-then-writes. That ordering is wrong for the `$import` performance goal: import is a write path, so `ResourceWrapperFactory` calls `IRawResourceFactory` on every imported resource, while `IResourceDeserializer` is not on the import path at all.
- **A new "Execution order" section** records that the ladder is a *dependency* order, not a schedule. The first delivery target is a demonstrable `$import` win reachable through a supported configuration change — a vertical slice of Phase 0 + Phase 2a — gated on a benchmark reproducing the delta measured on the throwaway `feature/ignixa-sdk` branch. Phase 1 and Phase 2b follow that gate. Phase numbers are identifiers, not a schedule, since the backlog now references them.

## Related issues
[AB#198706](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/198706)

Backlog realigned under Feature [180726 — FHIR vNext SDK migration](https://microsofthealth.visualstudio.com/Health/_workitems/edit/180726). Follow-on work for the performance goal is tracked by [AB#198707](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/198707) (Phase 2a — provider-selected `IRawResourceFactory` with native Ignixa serialization) and [AB#198708](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/198708) (import performance validation against the `feature/ignixa-sdk` baseline).

## Testing
- Shared parity test suite (`ImportResourceParserParityTests`) runs the same behavior corpus against both the Firely and Ignixa parsers for every supported FHIR version (STU3, R4, R4B, R5) — valid/invalid IDs, meta normalization, initial vs. incremental load, soft-delete detection and extension removal (including polymorphic `value[x]` shapes), and conditional-reference handling (including the documented Firely/Ignixa divergence for `contained` resources and Bundle entries).
- `dotnet test` (not just build) run and green across all four FHIR version projects; CI covers both `Windows_dotnet10` and `Linux_dotnet8` unit-test legs.
- Existing Firely-mode import tests remain green and unmodified in behavior; `Firely` stays the default provider, so no deployment is affected by this PR.

## FHIR Team Checklist
- [x] ADR added: `docs/arch/adr-2607-ignixa-import-phase0.md`

Feature — adds a new opt-in `FhirSdkProvider` configuration option; default behavior (Firely) is unchanged.


mikaelweave (approved):

brendankowitz:
Adding the breaking change label since this adds new dependencies that may need signing etc downstream

PR #5693, opened 2026-07-22
Fix to $bulk-update and $bulk-delete

NRamirez01:
## Description
Updates the value used for _lastUpdated (when no parameters are supplied) to one compatible with the FHIR API allowing $bulk-delete endpoints to work as per the [documentation](https://learn.microsoft.com/en-us/azure/healthcare-apis/fhir/fhir-bulk-delete).

Applies same update to $bulk-update job creation.

## Related issues
Addresses [Bug 195219](https://microsofthealth.visualstudio.com/Health/_workitems/edit/195219): $bulk-delete does not work without search parameters present
Builds on code added in [PR #5609](https://github.com/microsoft/fhir-server/pull/5609)

## Testing
Comparing the API response when populating "_lastUpdated" with the value produced by original code versus the proposed code.

image

Before:
image

After:
image

image

This approach was taken instead of changing PartialDateTime directly as that class is referenced in quite a few places.

## FHIR Team Checklist


NRamirez01:
suggested fix in https://github.com/microsoft/fhir-server/pull/5695; closed PR due to redundancy

PR #5695, opened 2026-07-23
Bulk delete date filter fix

LTA-Thinking:
## Description
Fixes the default last updated filter being added to bulk delete.

## Related issues
Addresses [issue #].

## Testing
Enhanced unit test

## FHIR Team Checklist


rajithaalurims (approved):

NRamirez01:
I was experimenting with trying to get bulk-delete working and I believe I have found an issue with the value used for _lastUpdated. 

I believe the value provided by PartialDateTime is an invalid format for the $bulk-delete endpoint.

https://github.com/microsoft/fhir-server/pull/5693/changes

Looking again at this today it seems like I was mistaken as to where the fix belongs, I had put it in CreateBulkDeleteHandler as opposed to the BulkDeleteOrchestratorJob but I believe the fix should still apply. 

My pull request will be closed as I am not going to bother my company's legal dept. with the CLA agreement at this time. Good luck!

LTA-Thinking:
@NRamirez01 thanks for the contribution. The service can handle timestamps in either format, the issue you saw was caused because the + sign wasn't URL encoded so it got turned into a space before it reached the service. Since the date created within the code never needs to leave the service that issue won't happen.

NRamirez01:
> @NRamirez01 thanks for the contribution. The service can handle timestamps in either format, the issue you saw was caused because the + sign wasn't URL encoded so it got turned into a space before it reached the service. Since the date created within the code never needs to leave the service that issue won't happen.

That makes sense! Sorry. 

What prompted me to dig into this code a bit was basically that I saw the patch notes from Microsoft saying bulk-delete was fixed in July, but as of last week we still had no luck using it without a search param.

Out of curiosity do you know by any chance what the release schedule is like to get patches from this repo to Azure Health Data Services?

Did just check and as of today 7-29-2026 it does work!

PR #5701, opened 2026-07-29
Added transaction bundle test with reindex conflict

SergeyGaluzo:


test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
```diff
@@ -1178,10 +1178,17 @@ public async Task GivenSearchParamDelete_ThenConflictWhenReindexAndSuccessOnNext
             Assert.Equal(hardDelete ? HttpStatusCode.NotFound : HttpStatusCode.Gone, notFoundEx.StatusCode);
         }
 
-        [Fact]
-        public async Task GivenSearchParamBundleDelete_ThenConflictWhenReindexAndSuccessOnNext()
+        [Theory]
```
LTA-Thinking, line 1181: Can you make this a SkipableTheory and use a skip statement instead of having it return true for the untestable case? There is an example in BundleBatchTests line 52.

SergeyGaluzo: I prefer not. We will be removing cosmos support anyway soon. Also, this pattern is used in other places in the tests.

LTA-Thinking (approved):

PR #5711, opened 2026-08-03
Reindex refactoring

SergeyGaluzo:
Scope of changes:

Removes 3 unnecessary resource table scans to get counts.
Simplifies job models by eliminating derived fields and consistently using job info instead of job record.
Reduces 10x memory footprint required for processing.
Simplifies OOM processing. 
Introduces fail-fast approach. Removes unnecessary and incorrect code to derive job status for not completed reindex. 
Removes support for old GUID based job ids and old storages.
Simplifies logic by removing "early exists" based on zero counts and empty lists when possible.
Introduces "transient" dictionaries to facilitate reindex completion management.
Simplifies logic to derive final search param statuses.
Eliminates redundant calls to the database to get search param statuses.
Improves scalability of getting job status calls.
Adds logic to deal with orphan search params. 
Extends throws to cases when cache is not in sync with storage.
Fixes incremental cache update bug.

Also fixes:
https://microsofthealth.visualstudio.com/Health/_workitems/edit/204790/
https://microsofthealth.visualstudio.com/Health/_workitems/edit/180432/
https://microsofthealth.visualstudio.com/Health/_workitems/edit/180601/
https://microsofthealth.visualstudio.com/Health/_workitems/edit/186787/
https://microsofthealth.visualstudio.com/Health/_workitems/edit/203947/
https://microsofthealth.visualstudio.com/Health/_workitems/edit/204877
https://microsofthealth.visualstudio.com/Health/_workitems/edit/204896
https://microsofthealth.visualstudio.com/Health/_workitems/edit/205616


src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexProcessingJob.cs
```diff
@@ -305,399 +227,117 @@ private async Task ProcessQueryAsync(CancellationToken cancellationToken)
...
-                    _reindexProcessingJobDefinition.ResourceType,
-                    failedRange.StartId,
-                    failedRange.EndId,
-                    rangeSize,
-                    numberOfRanges,
-                    true,
-                    cancellationToken,
-                    true);
-            }
+                await ComputeAndWrite(resources, null, store.Value, _cancellationToken);
```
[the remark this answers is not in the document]
SergeyGaluzo: Fixed

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexOrchestratorJob.cs
```diff
@@ -9,11 +9,11 @@
...
 using System.Linq;
 using System.Security.Cryptography;
+using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 using EnsureThat;
 using Hl7.Fhir.Model;
-using Hl7.Fhir.Utility;
 using Microsoft.AspNetCore.JsonPatch.Internal;
 using Microsoft.Data.SqlClient;
```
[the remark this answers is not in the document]
SergeyGaluzo: Fixed

src/Microsoft.Health.Fhir.SqlServer/Features/Search/SqlServerSearchService.cs
```diff
@@ -922,7 +922,8 @@ private void PopulateSqlCommandFromQueryHints(SqlSearchOptions options, SqlComma
             var resourceTypeId = _model.GetResourceTypeId(hints.First(x => x.Param == KnownQueryParameterNames.Type).Value);
             var startId = long.Parse(hints.First(x => x.Param == KnownQueryParameterNames.StartSurrogateId).Value);
             var endId = long.Parse(hints.First(x => x.Param == KnownQueryParameterNames.EndSurrogateId).Value);
-            var globalEndId = long.Parse(hints.First(x => x.Param == KnownQueryParameterNames.GlobalEndSurrogateId).Value);
+            var globalStr = hints.First(x => x.Param == KnownQueryParameterNames.GlobalEndSurrogateId).Value;
+            var globalEndId = globalStr == null ? null : (long?)long.Parse(globalStr);
```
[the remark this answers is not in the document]
SergeyGaluzo: Fixed

src/Microsoft.Health.Fhir.Core/Features/Definition/SearchableSearchParameterDefinitionManager.cs
```diff
@@ -183,5 +183,10 @@ private bool UsePartialSearchParams(SearchParameterInfo parameter)
                    _fhirReqeustContextAccessor.RequestContext.IncludePartiallyIndexedSearchParams &&
                    parameter.IsSupported;
         }
+
+        public ICollection<string> GetDerivedResourceTypes(IReadOnlyCollection<string> resourceTypes)
+        {
+            throw new NotImplementedException();
+        }
```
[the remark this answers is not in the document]
SergeyGaluzo: Fixed

tools/Microsoft.Health.Fhir.R4.ResourceParser/Code/MinimalSearchParameterDefinitionManager.cs
```diff
@@ -222,5 +222,10 @@ public bool TryGetSearchParameter(string definitionUri, bool excludePendingDelet
         {
             throw new NotImplementedException();
         }
+
+        public ICollection<string> GetDerivedResourceTypes(IReadOnlyCollection<string> resourceTypes)
+        {
+            throw new NotImplementedException();
+        }
```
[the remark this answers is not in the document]
SergeyGaluzo: Will not fix

src/Microsoft.Health.Fhir.Core/Features/Definition/SupportedSearchParameterDefinitionManager.cs
```diff
@@ -146,5 +146,10 @@ public void UpdateSearchParameterStatus(string url, SearchParameterStatus desire
         {
             _inner.UpdateSearchParameterStatus(url, desiredStatus);
         }
+
+        public ICollection<string> GetDerivedResourceTypes(IReadOnlyCollection<string> resourceTypes)
+        {
+            throw new NotImplementedException();
+        }
```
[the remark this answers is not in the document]
SergeyGaluzo: Fixed

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexOrchestratorJob.cs
```diff
@@ -968,572 +634,103 @@ private string GetSearchParameterHash(string resourceType)
...
+                        if (result == null)
                         {
-                            _currentResult.FailedResources += definition?.ResourceCount?.Count ?? 0;
-                            AddErrorResult(
-                                OperationOutcomeConstants.IssueSeverity.Error,
-                                OperationOutcomeConstants.IssueType.Exception,
-                                $"Processing job {jobInfo.Id} failed but returned no result payload.");
+                            _logger.LogJobWarning(_jobInfo, "Processing job {ProcessingJobId} had an empty result payload. Status={Status}.", job.Id, job.Status);
+                            AddErrorResult(OperationOutcomeConstants.IssueSeverity.Error, OperationOutcomeConstants.IssueType.Exception, $"Processing job {job.Id} failed but returned no result payload.");
                         }
```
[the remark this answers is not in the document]
SergeyGaluzo: Fixed

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexProcessingJob.cs
```diff
@@ -50,654 +50,283 @@ public class ReindexProcessingJob : IJob
...
-                    await ProcessWithContinuationTokensAsync(_searchParameterHash, cancellationToken);
-                }
-
-                if (!cancellationToken.IsCancellationRequested)
-                {
-                    _logger.LogJobInformation(_jobInfo, "Reindex processing job complete. Total number of resources indexed by this job: {Progress}.", _reindexProcessingJobResult.SucceededResourceCount);
-                }
+                    if (_isSql)
+                    {
+                        await ProcessWithSurrogateIdRangeAsync();
```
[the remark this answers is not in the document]
SergeyGaluzo: Fixed by moving _result initialization inside OOM retries.

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
```diff
@@ -25,96 +25,16 @@
...
         public override async Task<ReindexJobWrapper> GetReindexJobByIdAsync(string jobId, CancellationToken cancellationToken)
         {
-            if (IsLegacyJob(jobId))
-            {
-                // try old job records
-                var oldJobs = (ILegacyReindexOperationDataStore)this;
-                return await oldJobs.GetLegacyReindexJobByIdAsync(jobId, cancellationToken);
-            }
-
             return await base.GetReindexJobByIdAsync(jobId, cancellationToken);
```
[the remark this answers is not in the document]
SergeyGaluzo: This is intentional. Old stores/formats were deprecated >6 months ago.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Operations/CosmosFhirOperationDataStore.cs
```diff
@@ -125,47 +125,7 @@ async Task<ExportJobOutcome> ILegacyExportOperationDataStore.GetLegacyExportJobB
...
         public override async Task<ReindexJobWrapper> GetReindexJobByIdAsync(string jobId, CancellationToken cancellationToken)
         {
-            if (IsLegacyJob(jobId))
-            {
-                // try old job records
-                var oldJobs = (ILegacyReindexOperationDataStore)this;
-                return await oldJobs.GetLegacyReindexJobByIdAsync(jobId, cancellationToken);
-            }
-
             return await base.GetReindexJobByIdAsync(jobId, cancellationToken);
```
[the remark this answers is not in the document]
SergeyGaluzo: This is intentional. Old stores/formats were deprecated >6 months ago.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/CreateReindexRequestHandler.cs
```diff
@@ -55,10 +55,8 @@ public async Task<CreateReindexResponse> HandleAsync(CreateReindexRequest reques
             }
 
             var jobRecord = new ReindexJobRecord(
-            request.TargetResourceTypes,
-            request.MaximumResourcesPerQuery ?? _reindexJobConfiguration.MaximumNumberOfResourcesPerQuery,
-            request.MaximumResourcesPerWrite ?? _reindexJobConfiguration.MaximumNumberOfResourcesPerWrite,
-            request.QueryDelayIntervalInMilliseconds ?? _reindexJobConfiguration.QueryDelayIntervalInMilliseconds);
+                request.MaximumResourcesPerQuery ?? _reindexJobConfiguration.MaximumNumberOfResourcesPerQuery,
+                request.MaximumResourcesPerWrite ?? _reindexJobConfiguration.MaximumNumberOfResourcesPerWrite);
```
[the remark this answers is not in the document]
SergeyGaluzo: This is intentional, as these values were actually deprecated >6 months ago. Exceptions are not raised for backward compatibility. Old values were removed from documentation, so new customers cannot use them.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexOrchestratorJob.cs
```diff
@@ -266,399 +249,219 @@ private async Task<List<string>> CleanupMissingSearchParameterResourcesAsync(IRe
...
-
-                // Get search parameters that are valid for this specific resource type
-                var validSearchParameterUrls = GetValidSearchParameterUrlsForResourceType(resourceType);
+            var existingCosmosJobs = !_isSql
+                                   ? (await _timeoutRetries.ExecuteAsync(async () => await _queueClient.GetJobByGroupIdAsync((byte)QueueType.Reindex, _jobInfo.GroupId, true, _cancellationToken)))
+                                        .Where(_ => _.Id != _jobInfo.GroupId)
+                                        .Select(_ => new { _.Id, Def = JsonConvert.DeserializeObject<ReindexProcessingJobDefinition>(_.Definition) })
+                                        .GroupBy(_ => _.Def.ResourceType)
+                                        .ToDictionary(_ => _.Key, _ => _.Select(item => item.Id).ToList())
+                                   : null;
```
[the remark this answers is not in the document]
SergeyGaluzo: This is incorrect. 
During the whole period reindex is running, search param updates are prohibited. During restart process reindex is still considered running, so not search param data can change.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexProcessingJob.cs
```diff
@@ -50,654 +50,283 @@ public class ReindexProcessingJob : IJob
...
-            if (_reindexProcessingJobDefinition == null)
-            {
-                throw new InvalidOperationException("_reindexProcessingJobDefinition cannot be null during processing.");
-            }
-
-            if (cancellationToken.IsCancellationRequested)
+            if (_cancellationToken.IsCancellationRequested)
             {
                 return;
             }
```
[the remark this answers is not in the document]
SergeyGaluzo: Fixed

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
```diff
@@ -806,62 +721,54 @@
...
-                immunizationSearchParam = await CreateCustomSearchParameterAsync($"custom-parameter-before-imm-{randomSuffix}", ["Immunization"], "Immunization.vaccineCode", SearchParamType.Token);
+            // Create custom search parameters - one with valid FHIRPath expression
+            specimenSearchParam = await CreateCustomSearchParameterAsync($"custom-parameter-before-specimen-{randomSuffix}", ["Specimen"], "Specimen.type", SearchParamType.Token);
+            immunizationSearchParam = await CreateCustomSearchParameterAsync($"custom-parameter-before-imm-{randomSuffix}", ["Immunization"], "Immunization.vaccineCode", SearchParamType.Token);
 
-                // wait for several cache refresh cycles to let all VMs get new search param definition. Assume that refresh interval is 1 sec.
-                await Task.Delay(3000);
+            // wait for several cache refresh cycles to let all VMs get new search param definition. Assume that refresh interval is 1 sec.
+            await Task.Delay(3000);
```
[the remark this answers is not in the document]
SergeyGaluzo: Fixed.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexOrchestratorJob.cs
```diff
@@ -266,399 +258,219 @@ private async Task<List<string>> CleanupMissingSearchParameterResourcesAsync(IRe
...
 
-                // Check if surrogate ID ranging hasn't been determined yet or is supported
-                if (_isSurrogateIdRangingSupported)
+                if (_isSql)
                 {
                     // Use batched calls to GetSurrogateIdRanges to avoid timeout on large tables
-                    // Following the same pattern as Export job
-                    // Stream and enqueue each batch immediately so workers can start processing sooner
+                    // Enqueue each batch immediately so workers can start processing sooner
                     var numberOfRangesPerBatch = _operationsConfiguration.Reindex.NumberOfRecordRanges;
```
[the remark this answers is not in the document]
SergeyGaluzo: This is incorrect. 
On restart jobs with the same definitions are generated. When enqueued they will get old job ids. No duplication. Enqueue is designed as idempotent call.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexProcessingJob.cs
```diff
@@ -50,654 +50,275 @@ public class ReindexProcessingJob : IJob
...
+                var resources = result.Results?.Select(_ => _.Resource).ToList();
+                if (resources?.Count > 0)
                 {
-                    _logger.LogJobInformation(_jobInfo, "Continuation token found. Fetching next batch of resources for reindexing.");
+                    await ComputeAndWrite(resources, store.Value, _cancellationToken);
 
-                    // Clear the previous continuation token first to avoid conflicts
-                    queryState.ContinuationToken = null;
+                    _result.SucceededResourceCount += resources.Count;
+                    _jobInfo.Data = _result.SucceededResourceCount;
```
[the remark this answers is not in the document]
SergeyGaluzo: Will not fix, due to complexity. low probability, and Cosmos deprecation soon.

src/Microsoft.Health.Fhir.Core/Features/Operations/FhirOperationDataStoreBase.cs
```diff
@@ -253,50 +253,37 @@ public virtual async Task<ReindexJobWrapper> CreateReindexJobAsync(ReindexJobRec
...
-            var jobWithGroupId = await _queueClient.GetJobByIdAsync((byte)QueueType.Reindex, long.Parse(jobRecord.Id), false, cancellationToken);
+            var jobWithGroupId = await _queueClient.GetJobByIdAsync((byte)QueueType.Reindex, id, false, cancellationToken);
             await _queueClient.CancelJobByGroupIdAsync((byte)QueueType.Reindex, jobWithGroupId.GroupId, cancellationToken);
+
+            var jobRecord = JsonConvert.DeserializeObject<ReindexJobRecord>(jobWithGroupId.Definition);
+            var now = Clock.UtcNow;
+            jobRecord.Status = OperationStatus.Canceled;
+            jobRecord.CanceledTime = now;
+            jobRecord.LastModified = now;
+            PopulateReindexJobRecordTimestampsFromJobInfo(jobWithGroupId, jobRecord);
```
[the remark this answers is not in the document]
SergeyGaluzo: Fixed

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Reindex/ReindexTests.cs
```diff
@@ -801,67 +712,57 @@
...
-                immunizationSearchParam = await CreateCustomSearchParameterAsync($"custom-parameter-before-imm-{randomSuffix}", ["Immunization"], "Immunization.vaccineCode", SearchParamType.Token);
+            // Create custom search parameters - one with valid FHIRPath expression
+            var specimenSearchParam = await CreateCustomSearchParameterAsync($"custom-parameter-before-specimen-{randomSuffix}", ["Specimen"], "Specimen.type", SearchParamType.Token);
+            var immunizationSearchParam = await CreateCustomSearchParameterAsync($"custom-parameter-before-imm-{randomSuffix}", ["Immunization"], "Immunization.vaccineCode", SearchParamType.Token);
 
-                // wait for several cache refresh cycles to let all VMs get new search param definition. Assume that refresh interval is 1 sec.
-                await Task.Delay(3000);
+            // wait for several cache refresh cycles to let all VMs get new search param definition. Assume that refresh interval is 1 sec.
+            await Task.Delay(3000);
```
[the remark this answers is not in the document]
SergeyGaluzo: Fixed

src/Microsoft.Health.Fhir.Core/Features/Search/SearchService.cs
```diff
@@ -286,9 +286,14 @@ public abstract Task<SearchResult> SearchAsync(
...
             CancellationToken cancellationToken);
 
-        protected abstract Task<SearchResult> SearchForReindexInternalAsync(
-            SearchOptions searchOptions,
-            string searchParameterHash,
-            CancellationToken cancellationToken);
+        public virtual Task<SearchResult> SearchForReindexInternalAsync(SearchOptions searchOptions, string searchParameterHash, CancellationToken cancellationToken)
+        {
+            throw new NotImplementedException();
+        }
```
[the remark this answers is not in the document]
SergeyGaluzo: Adding implementation to SQL is wrong, as we want to keep SQL clean. This will be fixed in near future when Cosmos is removed.

src/Microsoft.Health.Fhir.Core/Configs/ReindexJobConfiguration.cs
```diff
@@ -60,6 +43,16 @@ public ReindexJobConfiguration()
         /// </summary>
         public int ActiveHostsEventsMultiplier { get; set; } = 9;
 
+        /// <summary>
+        /// Controls orchetrator job info polling interval
```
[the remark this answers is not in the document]
SergeyGaluzo: Fixed.

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Features/Operations/Reindex/ReindexSearchTests.cs
```diff
@@ -35,19 +35,24 @@ public class ReindexSearchTests : IClassFixture<FhirStorageTestsFixture>
...
             _scopedDataStore = fixture.DataStore.CreateMockScope();
             _searchService = fixture.SearchService.CreateMockScope();
             _searchParameterDefinitionManager = fixture.SearchParameterDefinitionManager;
+            _isSQL = fixture.DataStore is SqlServer.Features.Storage.SqlServerFhirDataStore;
         }
 
         [Fact]
         public async Task GivenResourceWithMatchingHash_WhenPerformingReindexSearch_ThenResourceShouldNotBeReturned()
         {
+            if (_isSQL) // search param hash matching is supprted ony in cosmos.
```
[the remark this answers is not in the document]
SergeyGaluzo: Fixed

test/Microsoft.Health.Fhir.Shared.Tests.Integration/Features/Operations/Reindex/ReindexSearchTests.cs
```diff
@@ -34,20 +34,23 @@ public class ReindexSearchTests : IClassFixture<FhirStorageTestsFixture>
...
             _scopedDataStore = fixture.DataStore.CreateMockScope();
             _searchService = fixture.SearchService.CreateMockScope();
-            _searchParameterDefinitionManager = fixture.SearchParameterDefinitionManager;
+            _isSQL = fixture.DataStore is SqlServer.Features.Storage.SqlServerFhirDataStore;
         }
 
         [Fact]
         public async Task GivenResourceWithMatchingHash_WhenPerformingReindexSearch_ThenResourceShouldNotBeReturned()
         {
+            if (_isSQL) // search param hash matching is supported only in cosmos.
```
LTA-Thinking: Instead of this check you can add the attribute [FhirStorageTestsFixtureArgumentSets(DataStore.CosmosDb)] to the test.

SergeyGaluzo: Sure

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexProcessingJob.cs
```diff
@@ -50,654 +51,267 @@ public class ReindexProcessingJob : IJob
...
             EnsureArg.IsNotNull(jobInfo, nameof(jobInfo));
-
+            _cancellationToken = cancellationToken;
             _jobInfo = jobInfo;
-            _reindexProcessingJobDefinition = DeserializeJobDefinition(_jobInfo);
+            _definition = JsonConvert.DeserializeObject<ReindexProcessingJobDefinition>(jobInfo.Definition);
 
-            await CheckDiscrepancies(cancellationToken);
+            // code will change batch size on OOM retries. Do not reference MaximumNumberOfResourcesPerWrite down below.
+            _batchSize = _definition.MaximumNumberOfResourcesPerWrite;
```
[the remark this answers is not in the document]
SergeyGaluzo: per query value is not applicable to cosmos and is rightfully ignored.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexOrchestratorJob.cs
```diff
@@ -968,572 +635,98 @@ private string GetSearchParameterHash(string resourceType)
...
-                                OperationOutcomeConstants.IssueType.Exception,
-                                $"Processing job {jobInfo.Id} failed but returned no result payload.");
-                        }
-
-                        continue;
-                    }
-
-                    _currentResult.SucceededResources += result.SucceededResourceCount;
+                        _result.SucceededResources += succeededResourceCount;
+                        _result.FailedResources += def.ResourceCount.Count - succeededResourceCount;
```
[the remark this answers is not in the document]
SergeyGaluzo: This is not correct. Reindex logic ensures that all writes while reindex is running write new indexes, and such then can only reduce processed count because they use new search param hash.

src/Microsoft.Health.Fhir.Shared.Core/Extensions/ReindexJobRecordExtensions.cs
```diff
@@ -48,25 +48,26 @@ public static ResourceElement ToParametersResourceElement(this ReindexJobWrapper
             totalResourceCount = job.Status == OperationStatus.Completed
                 ? job.Count
                 : job.ResourceCounts?.Sum(entry => entry.Value.Count) ?? 0;
+            long succeededResourceCount = Math.Max(0, job.Count - job.FailureCount);
```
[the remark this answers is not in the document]
SergeyGaluzo: This is not correct. See above.

src/Microsoft.Health.Fhir.Core/Features/Operations/FhirOperationDataStoreBase.cs
```diff
@@ -454,102 +394,60 @@ record = JsonConvert.DeserializeObject<ReindexJobRecord>(jobInfo.Definition);
...
             {
-                if (job.Status == JobStatus.Completed)
-                {
-                    record.Progress += jobResult.SucceededResourceCount;
-                }
-
-                record.Count += jobResult.SucceededResourceCount + jobResult.FailedResourceCount;
+                var result = string.IsNullOrEmpty(job.Result) ? null : JsonConvert.DeserializeObject<ReindexProcessingJobResult>(job.Result);
+                record.Count += definition.ResourceCount.Count;
+                record.FailureCount += definition.ResourceCount.Count - (result?.SucceededResourceCount ?? 0);
```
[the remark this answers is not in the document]
SergeyGaluzo: This is not correct. See above.,

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexOrchestratorJob.cs
```diff
@@ -80,16 +78,22 @@ public sealed class ReindexOrchestratorJob : IJob
...
 
-        private HashSet<long> _processedJobIds = new HashSet<long>();
-        private HashSet<string> _processedSearchParameters = new HashSet<string>();
-        private List<JobInfo> _jobsToProcess;
+        private readonly HashSet<string> _processedSearchParameters = []; // to prevent multiple status updates
+
+        // Transient dictionaries below are populated on processing job creates. After a job is in the terminal state
+        // it is removed from _transientResourceTypeJobs. When all jobs removed then resource type is completed.
+        // Similar concept is used for _transientSearchParamResouceTypes
+        private readonly Dictionary<string, HashSet<long>> _transientResourceTypeJobs = [];
```
mikaelweave, line 86: If the orchestrator job stops mid-way through and restarts (compute upgrade, rollout, or a crash) will the orchestrator queue up duplicate jobs (jobs from before the restart + restarted orchestrator job)?

If so -- is this an acceptable edge condition?

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexOrchestratorJob.cs
```diff
@@ -113,86 +116,76 @@ public ReindexOrchestratorJob(
...
+            }
+            catch (JobExecutionException ex) when (ex.Error is ReindexOrchestratorJobResult)
+            {
+                throw;
             }
             catch (Exception ex)
             {
-                HandleException(ex);
+                AddErrorResult(OperationOutcomeConstants.IssueSeverity.Error, OperationOutcomeConstants.IssueType.Exception, ex.Message);
+                _logger.LogJobError(ex, _jobInfo, $"The reindex failed. Id={_jobInfo.Id}");
```
mikaelweave: Do we need to throw here to signal an error to `JobHosting`?

something like `throw new JobExecutionException(ex.Message, _result, ex)` after the log error

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexOrchestratorJob.cs
mikaelweave: Is this code still needed with the refactor? 

`if (searchResultReindex != null)` block I mean

`_resourceCounts` is only set for Cosmos and this block says `this should never be for cosmos`

CalculateAndSetTotalAndResourceCounts looks to be only called for Cosmos which sets this.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexOrchestratorJob.cs
SergeyGaluzo: First, this is not different from the old reindex code. Second, if there is no write activity on affected resource type, new processing jobs registration will yield the same jobs. If there is write activity, orchestrator might schedule different jobs. Orchestrator cares only about one current set, so whether others work or not, it will not matter for the end result.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexOrchestratorJob.cs
SergeyGaluzo: I am not sure. I left current behavior. To make fail-fast consistent we probably should.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexOrchestratorJob.cs
SergeyGaluzo: Good. This is indeed dead code. I will remove,

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/Models/ReindexJobRecord.cs
```diff
@@ -81,25 +63,12 @@ protected ReindexJobRecord()
         {
         }
 
-        /// <summary>
-        /// Use Concurrent dictionary to allow access to specific items in the list
-        /// Ignore the byte value field, effective using the dictionary as a hashset
-        /// </summary>
-        [JsonProperty(JobRecordProperties.QueryList)]
-        [JsonConverter(typeof(ReindexJobQueryStatusConverter))]
```
mikaelweave, line 89: Given we are removing this field, do we still needed `ReindexJobQueryStatusConverter.cs`

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/Models/ReindexJobRecord.cs
```diff
@@ -24,55 +24,37 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Reindex.Models
...
+            int maxResourcesPerQuery = MaxMaximumNumberOfResourcesPerQuery,
+            int maxResourcesPerWrite = MaxMaximumNumberOfResourcesPerWrite,
+            int typeId = (int)JobType.ReindexOrchestrator)
         {
-            TargetResourceTypes = EnsureArg.IsNotNull(targetResourceTypes, nameof(targetResourceTypes));
             TypeId = typeId;
 
-            // Default values
-            SchemaVersion = 1;
             Id = Guid.NewGuid().ToString();
```
mikaelweave, line 40: Do we still need the GUID Id?

src/Microsoft.Health.Fhir.Core/Features/Operations/FhirOperationDataStoreBase.cs
```diff
@@ -253,50 +253,34 @@ public virtual async Task<ReindexJobWrapper> CreateReindexJobAsync(ReindexJobRec
...
         {
-            var jobWithGroupId = await _queueClient.GetJobByIdAsync((byte)QueueType.Reindex, long.Parse(jobRecord.Id), false, cancellationToken);
+            var jobWithGroupId = await _queueClient.GetJobByIdAsync((byte)QueueType.Reindex, id, true, cancellationToken);
             await _queueClient.CancelJobByGroupIdAsync((byte)QueueType.Reindex, jobWithGroupId.GroupId, cancellationToken);
+
+            var jobRecord = JsonConvert.DeserializeObject<ReindexJobRecord>(jobWithGroupId.Definition);
+            jobRecord.Status = OperationStatus.Canceled;
+            PopulateReindexJobRecordTimestampsFromJobInfo(jobWithGroupId, jobRecord);
+
+            return new ReindexJobWrapper(jobRecord, WeakETag.FromVersionId(jobWithGroupId.Version.ToString()));
```
mikaelweave, line 278: Will this return the job id as a GUID? Do we need something like `GetReindexJobByIdAsync` has that overrides the default guid? `record.Id = jobInfo.Id.ToString()`

src/Microsoft.Health.Fhir.Core/Features/Definition/ISearchParameterDefinitionManager.cs
```diff
@@ -127,5 +127,7 @@ public interface ISearchParameterDefinitionManager
         /// <param name="url">The url identifying the custom search parameter to update.</param>
         /// <param name="desiredStatus">The desired status for the custom search parameter to update.</param>
         void UpdateSearchParameterStatus(string url, SearchParameterStatus desiredStatus);
+
+        ICollection<string> GetDerivedResourceTypes(IReadOnlyCollection<string> resourceTypes);
```
mikaelweave, line 131: nit - xml comments missing

src/Microsoft.Health.Fhir.Core/Features/Search/ISearchService.cs
```diff
@@ -110,6 +110,8 @@ Task<SearchResult> SearchForReindexAsync(
             CancellationToken cancellationToken,
             bool activeOnly = false);
 
+        Task<SearchResult> SearchBySurrogateIdRange(string resourceType, long startId, long endId, CancellationToken cancellationToken);
```
mikaelweave, line 113: nit - xml comments missing. not a big deal as others are missing in this file

src/Microsoft.Health.Fhir.Core/Features/Search/Registry/ISearchParameterStatusManager.cs
```diff
@@ -24,5 +24,7 @@ public interface ISearchParameterStatusManager
         Task UpdateSearchParameterStatusAsync(IReadOnlyCollection<string> searchParameterUris, SearchParameterStatus status, CancellationToken cancellationToken, bool ignoreSearchParameterNotSupportedException = false, long? reindexId = null, DateTimeOffset? lastUpdated = null);
 
         Task<CacheConsistencyResult> CheckCacheConsistencyAsync(DateTime updateEventsSince, DateTime activeHostsSince, CancellationToken cancellationToken);
+
+        Task TryLogEvent(string process, string status, string text, DateTime? startDate, CancellationToken cancellationToken);
```
mikaelweave, line 28: nit - xml comments missing. not a big deal as others are missing in this file

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Operations/Reindex/ReindexProcessingJobTests.cs
```diff
@@ -855,463 +560,150 @@ public async Task ExecuteAsync_WithLargeSurrogateIdRange_ProcessesInMemorySafeBa
...
-        [Fact]
-        public async Task ExecuteAsync_WithOutOfMemoryException_UsesReducedRangeParametersForSplit()
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public async Task WhenOutOfMemoryException_ThenRetriesWorkCorrectly(bool throwOnRead)
         {
-            const string expectedResourceType = "DiagnosticReport";
-            int callCount = 0;
+            ReindexProcessingJob.OomRetryDelayBaseSec = 1;
```
mikaelweave, line 620: If this is being set during a test, we should consider making  `ReindexProcessingJob.OomRetryDelayBaseSec` not static across multiple parallel tests.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
```diff
@@ -647,8 +647,8 @@ public async Task<ResourceWrapper> UpdateSearchParameterIndicesAsync(ResourceWra
                 switch (exception.GetSubStatusCode())
                 {
                     case HttpStatusCode.PreconditionFailed:
-                        _logger.LogError(string.Format(Microsoft.Health.Fhir.Core.Resources.ResourceVersionConflict, WeakETag.FromVersionId(resourceWrapper.Version)));
-                        throw new PreconditionFailedException(string.Format(Microsoft.Health.Fhir.Core.Resources.ResourceVersionConflict, WeakETag.FromVersionId(resourceWrapper.Version)));
+                        _logger.LogWarning(string.Format(Microsoft.Health.Fhir.Core.Resources.ResourceVersionConflict, WeakETag.FromVersionId(resourceWrapper.Version)));
```
mikaelweave, line 650: Are version conflicts no longer an issue with this new approach? Can you add a small comment here to document this?

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/CreateReindexRequestHandler.cs
```diff
@@ -55,10 +55,8 @@ public async Task<CreateReindexResponse> HandleAsync(CreateReindexRequest reques
             }
 
             var jobRecord = new ReindexJobRecord(
-            request.TargetResourceTypes,
```
mikaelweave, line 58: Should we keep target resource types for requests that only are trying to reindex one type (like `Patient`)?

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexOrchestratorJob.cs
```diff
@@ -968,572 +635,93 @@ private string GetSearchParameterHash(string resourceType)
...
-                                _logger.LogJobWarning(ex, _jobInfo, "Heartbeat conflict - another worker updated the job");
-                            }
-                        }
-                    }
-                    catch (OperationCanceledException) when (!cancellationToken.IsCancellationRequested)
-                    {
-                        // Poll interval timeout occurred - log and continue with next iteration
-                        _logger.LogDebug("Poll interval timeout occurred for job {JobId}. Continuing with next iteration.", _jobInfo.Id);
-                    }
+                var finishedJobs = batch.Where(j => j.Status == JobStatus.Completed || j.Status == JobStatus.Failed).ToList();
```
mikaelweave, line 593: Do we need to account for cancelled jobs?

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexOrchestratorJob.cs
```diff
@@ -246,419 +240,229 @@ async Task<bool> WaitForAllInstancesCacheSyncAsync(DateTime updateEventsSince, C
...
-            _logger.LogJobInformation(_jobInfo, "Found {Count} search parameter(s) in pending delete states. Checking if resources exist.", pending.Count);
+            var all = (await _searchParameterStatusManager.GetAllSearchParameterStatus(_cancellationToken)).Select(_ => _.Uri.OriginalString).ToList();
+            var systemDefined = new HashSet<string>(_searchParameterDefinitionManager.AllSearchParameters.Where(p => p.IsSystemDefined).Select(p => p.Url.OriginalString));
+            var custom = all.Where(_ => !systemDefined.Contains(_)).ToList();
+            _logger.LogJobInformation(_jobInfo, $"DeleteOrphans: Custom params total={custom.Count}.");
 
-            var searchParameters = await _retries.ExecuteAsync(async () => await _searchParameterOperations.GetSearchParametersByUrlsAsync(pending, cancellationToken));
+            var searchParamsWithResources = await _retries.ExecuteAsync(async () => await _searchParameterOperations.GetSearchParametersByUrlsAsync(custom, _cancellationToken));
```
mikaelweave, line 254: Are we sure the `GetSearchParametersByUrlsAsync` call will get all search params? 

I am worried about the Cosmos path. Sometimes `SearchAsync` for cosmos does not return every record and can return a continuation toke. 

Can we double check here? I worry search parameters can be removed.

If this code is risky, we could feature flag the actual deletion of orphans to `LogOnly` for now and change to `Delete` later (or via override if we see correct orphan removals)

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexProcessingJob.cs
```diff
@@ -50,654 +51,267 @@ public class ReindexProcessingJob : IJob
...
-                    _logger.LogJobInformation(_jobInfo, "Reindex processing job complete. Total number of resources indexed by this job: {Progress}.", _reindexProcessingJobResult.SucceededResourceCount);
-                }
+                    if (_isSql)
+                    {
+                        _result = new ReindexProcessingJobResult(); // sql reruns all, so result has to be initialized on every oom retry
+                        await ProcessWithSurrogateIdRangeAsync();
+                    }
+                    else
+                    {
+                        await ProcessWithContinuationTokensAsync();
```
mikaelweave, line 188: Can this flow result in inaccurate resource counts (like count of resources reindexed) for Cosmos?

I think SQL starts again and resets `_result` -- so this would only be cosmos.

Does this flow mean on OOM the cosmos work restarts at the continuation token location but the counters are not taking this "reset factor" into the calculation and can over report resources reindexed?

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/CreateReindexRequestHandler.cs
mikaelweave: Looks like target resource types was removed per below comment. This can be ignored.

https://github.com/microsoft/fhir-server/pull/5711/changes#r3799512321

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/Models/ReindexJobRecord.cs
```diff
@@ -24,55 +24,37 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Reindex.Models
...
     public class ReindexJobRecord : JobRecord, IJobData
     {
-        public const uint MaxMaximumNumberOfResourcesPerQuery = 10000;
-        public const uint MinMaximumNumberOfResourcesPerQuery = 1;
+        public const int MaxMaximumNumberOfResourcesPerQuery = 10000;
+        public const int MinMaximumNumberOfResourcesPerQuery = 1;
 
-        public const uint MaxMaximumNumberOfResourcesPerWrite = 10000;
-        public const uint MinMaximumNumberOfResourcesPerWrite = 1;
+        public const int MaxMaximumNumberOfResourcesPerWrite = 1000;
```
mikaelweave: This will result in requests with values between 1000 and 10000 returning 400 for customer calls while previously they did not. I think this is an acceptable outcome.

Do we need any notification as part of the release? Or is the error flow self-evident?

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexProcessingJob.cs
```diff
@@ -50,654 +51,267 @@ public class ReindexProcessingJob : IJob
...
-        /// </summary>
         private static readonly AsyncPolicy _bulkUpdateRetries = Policy.WrapAsync(_requestRateRetries, _timeoutRetries);
 
         private readonly Func<IScoped<ISearchService>> _searchServiceFactory;
         private readonly IResourceWrapperFactory _resourceWrapperFactory;
         private readonly Func<IScoped<IFhirDataStore>> _fhirDataStoreFactory;
         private readonly ILogger<ReindexProcessingJob> _logger;
-        private readonly ISearchParameterStatusManager _searchParameterStatusManager;
         private readonly ISearchParameterOperations _searchParameterOperations;
+        private readonly ISearchParameterDefinitionManager _searchParameterDefinitionManager;
```
mikaelweave: nit. Looks like this is never used, could be removed.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexOrchestratorJobResult.cs
mikaelweave: Looks like this isn't used anymore. If we want to keep if for backwards compatibility, we should mark it obsolete.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/Models/ReindexJobRecord.cs
SergeyGaluzo: Yes. This ensures that new reindex definition does not match previous one.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/Models/ReindexJobRecord.cs
SergeyGaluzo: Removed

src/Microsoft.Health.Fhir.Core/Features/Operations/FhirOperationDataStoreBase.cs
SergeyGaluzo: Not today. But I plan to remove reindex job record projection completely, and replace it by job info some day soon.

src/Microsoft.Health.Fhir.Core/Features/Definition/ISearchParameterDefinitionManager.cs
SergeyGaluzo: Intentional. I hate meaningless comments.

src/Microsoft.Health.Fhir.Core/Features/Search/ISearchService.cs
SergeyGaluzo: same thing.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/CosmosFhirDataStore.cs
SergeyGaluzo: They were never an issue. Old code was throwing, then catching, and then doing nothing,

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexOrchestratorJob.cs
SergeyGaluzo: I do not think so. User cancellation is propagated to all jobs in a group including orchestrator, so orchestrator will be cancelled out, and its opinion on cancelled jobs does not matter.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexOrchestratorJob.cs
SergeyGaluzo: Get by URLs runs in batches, each with 100 URLs. This guarantees getting full page from SQL. For Cosmos I added continuation token handling,

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexProcessingJob.cs
SergeyGaluzo: It does not restart at continuation token, but uses search param hash filter, so it only goes to remaining rows even on OOM restart.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/Models/ReindexJobRecord.cs
SergeyGaluzo: The less notifications the better. I will change back to 10K. By default we still do 1K. If customer want to shoot itself in a foot, be my guest.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexProcessingJob.cs
SergeyGaluzo: Fixed

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexOrchestratorJobResult.cs
SergeyGaluzo: This is not used by current code either, Removed,

src/Microsoft.Health.Fhir.Shared.Core.UnitTests/Features/Operations/Reindex/ReindexProcessingJobTests.cs
SergeyGaluzo: It should not matter because tests do not hit OOM exception in real.

src/Microsoft.Health.Fhir.Core/Features/Search/Registry/ISearchParameterStatusManager.cs
SergeyGaluzo: Same...

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirOperationDataStore.cs
```diff
@@ -25,96 +25,16 @@
...
-            {
-                Converters = new List<JsonConverter>
-                {
-                    new EnumLiteralJsonConverter(),
-                },
-            };
-         }
+        }
 
         public override async Task<ReindexJobWrapper> GetReindexJobByIdAsync(string jobId, CancellationToken cancellationToken)
```
LTA-Thinking: Nit: This method isn't needed anymore. It just points to the base class method.

SergeyGaluzo: I will remove it. Once done there is nothing left in this class. I put TODO comment for cosmos cleanup

rajithaalurims (dismissed):

rajithaalurims (dismissed):

src/Microsoft.Health.Fhir.Core/Features/Search/Parameters/SearchParameterOperations.cs
```diff
@@ -255,8 +255,12 @@ public async Task<bool> GetAndApplySearchParameterUpdates(CancellationToken canc
                 // Identify all System Defined Search Parameters and filter them from statuses
                 var systemDefinedSearchParameterUris = new HashSet<string>(_searchParameterDefinitionManager.AllSearchParameters.Where(p => p.IsSystemDefined).Select(p => p.Url.OriginalString));
 
+                // Fetch statuses that require definition materialization during incremental refresh.
                 var statusesToFetch = statuses
-                                        .Where(p => p.Status == SearchParameterStatus.Enabled || p.Status == SearchParameterStatus.Supported)
+                                        .Where(p => p.Status == SearchParameterStatus.Enabled
```
SergeyGaluzo, line 260: This is the place for https://microsofthealth.visualstudio.com/Health/_workitems/edit/205616 fix

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexProcessingJob.cs
```diff
@@ -50,654 +50,263 @@ public class ReindexProcessingJob : IJob
...
+            catch (SqlException ex)
             {
-                // For non-timeout SQL errors
-                _logger.LogJobError(sqlEx, _jobInfo, "SQL error occurred during reindex processing.");
-                SetJobError($"SQL Error: {sqlEx.Message}");
-
-                throw new JobExecutionSoftFailureException($"SQL error occurred during reindex processing: {sqlEx.Message}", _reindexProcessingJobResult, sqlEx, isCustomerCaused: false);
+                _logger.LogJobError(ex, _jobInfo, $"Reindex processing job error occurred. SqlException: {ex.Message}.");
+                _result.Error = ex.Message;
+                throw new JobExecutionSoftFailureException(_result.Error, _result, ex, false);
```
feordin, line 196: All of the JobExecutionSoftFailureException will result in failed job with no retry?

SergeyGaluzo: Correct. This is not a change though. Processing jobs even today have this behavior.

The only difference is that new code is more graceful in raising exceptions on cache discrepancies (search param hash mismatch). New code, unlike old, will, not cancel the group (including orchestrator), but will let orchestrator to record/fail job from there. This eliminates current bug about incorrect message in final reindex result.

The change is that entire reindex follows fail-fast approach. It is very important that processing jobs are reliable, so everything processing job does goes via retry policies.

According to our current logs processing jobs are reliable enough to participate in fail-fast logic.

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexOrchestratorJob.cs
SergeyGaluzo: I changed it. It throws now.

feordin (approved):

rajithaalurims:
@SergeyGaluzo  for some reason I do not see option to approve. I reviewed and providing my approval to merge.

PR #5714, opened 2026-08-04
Fix three long-lived-database search parameter defects (Location?near= 500, Unsupported status repair, dead ignoreSearchParameterNotSupportedException flag)

feordin:
Fixes three defects that are only reachable on a **long-lived** SQL database. fhir-server's own E2E CI deploys a brand-new environment with a fresh database on every run (`samples/templates/default-azuredeploy-docker.json`), so none of these are reachable there. The Azure FHIR PaaS (Gen2) prod E2E pipeline runs the same test binaries against databases that are never recreated, which is where all three show up.

Each defect is a separate commit so they can be reviewed and reverted independently.

---

## Defect 1a — `KeyNotFoundException` for search parameter types with no value parser

**Symptom:** `KeyNotFoundException: The given key 'Special' was not present in the dictionary.` escaping as HTTP 500 (60 occurrences over 14 days on the eastus2euap R4 test account, first 2026-07-26T20:27:10Z, persisting across three service versions).

```
at SearchParameterExpressionParser.Parse(...) SearchParameterExpressionParser.cs:line 52
at ExpressionParser.ParseImpl(...) ExpressionParser.cs:line 295
at SearchOptionsFactory.Create(...) SearchOptionsFactory.cs:line 413
```

**Root cause:** `SearchParameterExpressionParser` builds `_parserDictionary` with exactly 7 entries — Date, Number, Quantity, Reference, String, Token, Uri (`SearchParameterExpressionParser.cs:34-44`). There is no entry for `SearchParamType.Special` (Composite is handled upstream). `Build(...)` looked the parser up with a raw indexer, so anything else threw `KeyNotFoundException`, which the API layer reports as a 500.

**Fix:** `TryGetValue`, and throw `SearchParameterNotSupportedException` when there is no parser. `SearchOptionsFactory` already catches that exception (`SearchOptionsFactory.cs:430`) and records the parameter in `unsupportedSearchParameters`, which becomes a `NotSupported` warning under `handling=lenient` and a 400 under `handling=strict`.

This is defence in depth — on its own it does **not** make the E2E test pass, because that test asserts the exact diagnostics string produced by the status path in 1b.

**Why CI never caught it:** on a fresh database `Location-near` is correctly stored as `Unsupported`, `IsSearchable` is false, and the parameter never reaches the parser.

---

## Defect 1b — `Unsupported` status is never repaired on existing databases (the real cause of the `Location?near=` 500)

**Symptom:** `Location?near=...` returns HTTP 500 instead of a bundle with a `NotSupported` warning. Acceptance test: `BasicSearchTests.GivenASearchRequestOnASearchParameterMarkedUnsupportedAtStartup_WhenHandled_ReturnsBundleWithNotSupportedWarning`.

**Root cause chain:**

1. `http://hl7.org/fhir/SearchParameter/Location-near` is listed in `src/Microsoft.Health.Fhir.Core/Data/{R4,R4B,Stu3,R5}/unsupported-search-parameters.json`, and `FilebasedSearchParameterStatusDataStore` returns it with `Status = Unsupported`.
2. **PR #5403** (2026-03-06) added `fs.Status = hasResources ? Supported : Enabled;` unconditionally in `SqlServerFhirModel.InitializeSearchParameterStatuses`, clobbering the file's `Unsupported` status and persisting `Location-near` as Enabled/Supported into customer databases.
3. **PR #5684** (2026-07-23) guarded that with `if (fs.Status != SearchParameterStatus.Unsupported)`. **The guard is unreachable for already-seeded rows**, because earlier in the same method
   ```csharp
   fileStatuses.RemoveAll((fs) => existingParams.Any((param) => param.Uri == fs.Uri && (param.Status != SearchParameterStatus.Initialized)));
   ```
   drops every file status whose DB row already exists with any status other than `Initialized`. So #5684 only helps green-field databases.
4. `SearchParameterStatusManager.EnsureInitializedAsync` trusts the DB status — it only re-evaluates support when the stored status *is* `Unsupported`. There is no path that demotes a stored `Enabled` back to `Unsupported`. `IsSearchable` stays true, the parameter reaches the expression parser, and every request 500s.

> ⚠️ **Databases seeded between 2026-03-06 (PR #5403) and 2026-07-23 (PR #5684) are affected and are repaired automatically on the next startup by this change.** No manual remediation or migration script is required.

**Fix:** keep file statuses in the merge list when the file says `Unsupported` but the stored status is `Enabled` or `Supported`, so `dbo.MergeSearchParams` writes them back to `Unsupported`.

*Why the repair is limited to `Enabled`/`Supported`:* `SearchParameterStateUpdateHandler.ParseRequestForUpdate` refuses `$status` updates on parameters whose stored status is `Unsupported` or `Deleted`, and only allows `Supported`/`Disabled` as a target. So `Enabled`/`Supported` on a file-`Unsupported` parameter can only have come from the #5403 defect. `Disabled`, `Deleted` and the `Pending*` states are left untouched so customer-driven and reindex-driven states are never clobbered. `PartialSupport` entries come back from the file as `Enabled` with `IsPartiallySupported = true` (not `Unsupported`), so they are never matched by the repair rule and keep their existing behaviour.

*On `LastUpdated`:* the old code did `fs.LastUpdated = existingParams.FirstOrDefault(...)?.LastUpdated ?? fs.LastUpdated;`. Reading `MergeSearchParams.sql`, the highest `LastUpdated` in the input TVP is used purely as an **optimistic concurrency token** — the sproc computes `@ExpectedLastUpdated = (SELECT max(LastUpdated) FROM @SearchParams)` and raises error 50001 unless it equals `(SELECT max(LastUpdated) FROM dbo.SearchParam)`. It then stamps **every merged row** with a fresh server-side `@LastUpdated = sysUTCdatetime()`, and that is what other replicas poll on to refresh their caches. So we do not need to (and must not) hand-pick a per-row value: carrying an *old* repaired row's `LastUpdated` into the input would fail the concurrency check, retry 3× and then throw at startup. The value is now normalized to the current max in `dbo.SearchParam` for every row in the merge, and cache refresh on other replicas happens automatically via the sproc's fresh stamp. (The previous code only worked because `InitializeBase.sql` inserts brand-new rows with `SYSDATETIMEOFFSET()`, so those rows were always the DB max.)

*Idempotency:* after a repaired startup the file status and stored status agree again, so the row is dropped by `RemoveAll` as before, `fileStatuses` is empty and no merge is issued at all. Covered by a dedicated test that asserts `LastUpdated` is unchanged across two startups.

**Why CI never caught it:** on a fresh database every `SearchParam` row starts at `Initialized`, so `RemoveAll` keeps it and the #5684 guard applies. The corruption requires a database that was seeded by a build in the affected window.

---

## Defect 2 — `ignoreSearchParameterNotSupportedException` was a dead parameter → HTTP 400 on hard-delete

**Symptom:** `DELETE /SearchParameter/{id}?hardDelete=true` returns 400 instead of 204. On the STU3 eastus2euap test account `DELETE /SearchParameter/us-core-patient-gender-identity?hardDelete=true` returned 400 on 70/70 attempts since 2026-07-27T01:49:25Z, while its two siblings from the same fixture (`us-core-ethnicity`, `us-core-race`) returned 204 every time. Its first two hard-deletes succeeded on 2026-07-26T21:42; the permanent 400 began ~4 minutes after `ReindexOrchestratorJob ... completed cache refresh at the End: SearchParamLastUpdated 07/27/2026 01:45:19`. Logged exception is `SearchParameterNotSupportedException` from logger `SearchParameterOperations` with body `Error deleting search parameter.` — **not** the SQL 50001 / `SearchParameterConcurrencyConflict` error.

Acceptance test: `BulkDeleteTests.GivenBulkDeleteRequest_WhenSearchParametersDeleted_ThenSearchParameterStatusShouldBeUpdated`, which fails in its first step, the `CleanupAsync()` local function.

**Root cause:**

| Step | Location | Behaviour |
|---|---|---|
| 1 | `DeletionService.DeleteSearchParameterWithLockAsync` (~712) | passes `ignoreSearchParameterNotSupportedException: true` |
| 2 | `SearchParameterOperations.DeleteSearchParameterAsync` (~195) | declared the flag, **never read it**; line ~208 passed `lastUpdated:` as a named argument, so the flag fell back to its default of `false` |
| 3 | `SearchParameterStatusManager.UpdateSearchParameterStatusAsync` (~139) | declared the flag, **never read it**; called `GetSearchParameter(uri)` for any status other than `Deleted` |
| 4 | `SearchParameterDefinitionManager.GetSearchParameter(string)` (~187) | throws `SearchParameterNotSupportedException` when the url is absent from `UrlLookup` |
| 5 | `SearchParameterOperations.DeleteSearchParameterAsync` catch block | logs `Error deleting search parameter.`, appends `CustomSearchDeleteError` and rethrows → HTTP 400 |

The url is missing from `UrlLookup` because `GetAndApplySearchParameterUpdates` (~270-273) calls `DeleteSearchParameter(uri)` for statuses that reached `Deleted`, removing the entry from the definition manager — while the `SearchParameter` **resource row still exists** in the store. A hard delete of that resource then tries to set `PendingHardDelete` on a url the registry no longer knows.

**History:** the flag *was* honoured when **PR #5026** (2025-07-08, commit `1a9c4a0da`) added it — it wrapped the `GetSearchParameter` validation in a `try/catch (SearchParameterNotSupportedException)` inside `UpdateSearchParameterStatusAsync`. That handling was removed by the **#5433 "Atomic Search Param operations" refactor** (2026-04-28, commit `fc4753958`), leaving the parameter dead in both methods ever since.

**Fix:** thread the flag end to end and restore #5026's intent.
- `SearchParameterOperations.DeleteSearchParameterAsync` now passes the flag explicitly (avoiding the named-argument trap).
- `SearchParameterStatusManager.UpdateSearchParameterStatusAsync` catches `SearchParameterNotSupportedException` around the validation with an exception filter (`when (ignoreSearchParameterNotSupportedException)`) and logs + skips the uri.
- **A skipped uri gets no status row**, matching #5026. Writing an orphaned `PendingHardDelete` row would poison every reindex cycle: `ReindexOrchestratorJob.CreateReindexProcessingJobsAsync` includes `PendingHardDelete` in its `validStatus` set (~279) and then logs `status=null uri=...` as an Error for urls it cannot resolve in the registry (~304-310), with a `TODO` to throw in future.
- Added an early return when every uri was skipped, so we don't call `UpsertStatuses` with an empty list.

**Net effect:** hard-deleting a `SearchParameter` resource whose registry entry is already gone returns 204, not 400.

**Why CI never caught it:** it requires a `SearchParameter` that previously reached `Deleted` status (removing it from the in-memory registry) while its resource row survives — i.e. accumulated state from earlier runs against the same database.

---

## Tests

No E2E test was modified — the two named in the brief are the acceptance criteria.

**New unit tests**
- `SearchValueExpressionBuilderTests` — 2 tests: a `Special`-typed parameter with a url, and one without a url, both must throw `SearchParameterNotSupportedException` rather than `KeyNotFoundException`.
- `SearchParameterStatusManagerTests` — 4 tests: without the flag it still throws; with the flag the uri is skipped and `UpsertStatuses` is not called; a mix of known and missing uris persists only the known ones; `Deleted` still bypasses registry validation entirely.
- `SearchParameterOperationsTests` (new file) — exercises the **real** `SearchParameterOperations` wired to a **real** `SearchParameterStatusManager`, so the named-argument trap in step 2 cannot be silently reintroduced. Covers hard and soft delete with the flag (no exception, no upsert) and without the flag (throws).

**New integration tests** (`SqlServerSearchParameterInitializationTests`)
- `GivenAnAgedDatabaseWhereAnUnsupportedSearchParameterIsStoredAsEnabled_WhenInitializing_ThenItIsRepairedToUnsupported` — seeds `Location-near` as `Enabled` (simulating a database seeded in the affected window), plus a second file-`Unsupported` parameter as `Disabled`, then restarts. Asserts the first is repaired to `Unsupported` and the second is left alone.
- `GivenADatabaseWhereSearchParameterStatusesAreAlreadyCorrect_WhenInitializing_ThenNothingIsRewritten` — asserts every status *and* `LastUpdated` is byte-for-byte unchanged across a second startup, which proves no merge was issued.

Both use a **fresh `SqlServerFhirModel` instance** rather than the fixture's. `SqlServerFhirModel.Initialize` short-circuits on `_highestInitializedVersion == version`, and the fixture already initializes to `SchemaVersionConstants.Max`, so calling `Initialize` again on the fixture's instance is a no-op and cannot exercise the startup path. A new instance is what makes the test behave like a service starting up against an existing database.

The repair test was verified to **fail** (`Expected: Unsupported / Actual: Enabled`) with the `SqlServerFhirModel` fix reverted, and pass with it applied.

**Results** (SDK 10.0.302, local SQL Server):

| Suite | Result |
|---|---|
| `{R4,R4B,Stu3,R5}.Core.UnitTests` — `SearchValueExpressionBuilderTests`, `SearchParameterOperationsTests`, `SearchParameterBehaviorTests`, `SearchParameterRetryTests`, `DeletionServiceTests` | 216 / 216 / 215 / 216 passed |
| `Core.UnitTests` — `SearchParameterStatusManagerTests` | 9 passed |
| `R4.Core.UnitTests` — full targeted sweep incl. `SearchParameterDefinitionManagerTests`, `SearchOptionsFactoryTests` | 315 passed |
| `R4.Tests.Integration` — `SqlServerSearchParameterInitializationTests` | 5 passed |

All touched projects build with 0 warnings, 0 errors.

### DIFF
--- a/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Search/Registry/SearchParameterStatusManagerTests.cs
+++ b/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Search/Registry/SearchParameterStatusManagerTests.cs
@@ -13,6 +13,7 @@ using Microsoft.Extensions.Logging.Abstractions;
 using Microsoft.Health.Core;
 using Microsoft.Health.Fhir.Core.Extensions;
 using Microsoft.Health.Fhir.Core.Features.Definition;
+using Microsoft.Health.Fhir.Core.Features.Search;
 using Microsoft.Health.Fhir.Core.Features.Search.Parameters;
 using Microsoft.Health.Fhir.Core.Features.Search.Registry;
 using Microsoft.Health.Fhir.Core.Messages.Search;
@@ -35,6 +36,7 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Search.Registry
         private static readonly string ResourceSecurity = "http://hl7.org/fhir/SearchParameter/Resource-security";
         private static readonly string ResourceQuery = "http://hl7.org/fhir/SearchParameter/Resource-query";
         private static readonly string ResourceSource = "http://hl7.org/fhir/SearchParameter/Resource-source";
+        private static readonly string MissingUri = "http://hl7.org/fhir/SearchParameter/us-core-patient-gender-identity";
 
         private readonly SearchParameterStatusManager _manager;
         private readonly ISearchParameterStatusDataStore _searchParameterStatusDataStore;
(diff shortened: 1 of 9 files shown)


feordin:
Closed automatically by GitHub when the head branch was renamed to users/jaerwin/search-param-status-fixes to follow our team's branch naming convention. Superseded by #5715, which contains the identical three commits.

PR #5734, opened 2026-08-18
[CosmosDb] Add FHIR runtime state configuration

fhibf:
Introduction of a runtime state configuration.
Currently, two values are going to be supported: active and deprecated. 

Future changes should leverage the use of both states. 

## Related issues
Addresses [AB#204631](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/204631).

## Testing
Tested using unit tests and manual validations. 

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Configs/CoreFeatureConfiguration.cs
+++ b/src/Microsoft.Health.Fhir.Core/Configs/CoreFeatureConfiguration.cs
@@ -180,5 +180,10 @@ namespace Microsoft.Health.Fhir.Core.Configs
         /// Gets or sets a value indicating whether SMART system scope authorization is enforced for Bulk Export.
         /// </summary>
         public bool EnableSmartExportScopeAuthorization { get; set; } = true;
+
+        /// <summary>
+        /// Gets or sets a value indicating the runtime state of the FHIR server.
+        /// </summary>
+        public string RuntimeState { get; set; }
     }
 }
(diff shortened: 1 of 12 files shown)


src/Microsoft.Health.Fhir.Core/Configs/CoreFeatureConfiguration.cs
```diff
@@ -180,5 +180,10 @@ public VersioningConfiguration Versioning
         /// Gets or sets a value indicating whether SMART system scope authorization is enforced for Bulk Export.
         /// </summary>
         public bool EnableSmartExportScopeAuthorization { get; set; } = true;
+
+        /// <summary>
+        /// Gets or sets a value indicating the runtime state of the FHIR server.
+        /// </summary>
+        public string RuntimeState { get; set; }
```
fhibf, line 187: Not using this property at the moment. 
I added it, as the new "RuntimeState" is added to appsettings.json, and this is a direct representation in code. 

When working on future user stories I'll probably use this value to log the raw value in the configurations, or remove it.

feordin (approved):

src/Microsoft.Health.Fhir.Core/Configs/CoreFeatureConfiguration.cs
mikaelweave, line 187: Should this be an enum vs a string like `TotalType` above? We could use the native .NET config parster.

src/Microsoft.Health.Fhir.Core/Registration/FhirRuntimeState.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        /// <summary>
+        /// The FHIR service is active.
+        /// </summary>
+        Active,
+
+        /// <summary>
+        /// The FHIR service is deprecated.
+        /// </summary>
+        Deprecated,
```
mikaelweave, line 21: Does this belong in OSS vs PaaS?

mikaelweave (commented):
looks good a couple questions

PR #5742, opened 2026-08-21
Protection from incorrect search param writes

SergeyGaluzo:


mikaelweave (approved):

mikaelweave:
Approved -- would be nice to have a test but looks good.

PR #5748, opened 2026-08-21
Fix 500 from $import status endpoint

rbans96:
## Description

Polling the `$import` status endpoint with a processing job id (rather than the orchestrator's id) returns a 500 forever:

```
System.ArgumentNullException: Value cannot be null. (Parameter 'uriString')
```

`GetJobByGroupIdAsync` hands back every job in the import group, orchestrator included, and the handler only filtered out `coord.Id`. When you poll by the orchestrator's own id that's enough, because the orchestrator's `Id` equals its `GroupId`. Poll by any other id in the group and the orchestrator record survives the filter, gets deserialized as an `ImportProcessingJobDefinition`, and comes back with a null `ResourceLocation` — `ImportOrchestratorJobDefinition` simply has no such property. `new Uri(null)` then throws, and since nothing about the stored state changes between polls, the caller is stuck on 500 with no way out.

Changes:

- Skip job records that don't carry an input file url, or whose result was never persisted, instead of letting one of them fail the whole status request. The url check is what distinguishes a processing job from the orchestrator, and it's deliberately used in place of a job-type check so that records written before job types were persisted still get reported exactly as before.
- Fall back to the `OperationFailed` message when a failed job has no input url, rather than throwing while building the `OperationFailedWithErrorFile` message.
- Route the three `JsonConvert.DeserializeObject` calls through a `DeserializeOrDefault<T>` helper. Newtonsoft returns null for absent, empty, or literal-`null` state, and every one of those call sites dereferenced it immediately.

Every input the new code skips is one the old code threw an unhandled exception on, so no request that previously succeeded returns anything different.

Worth calling out: this stops the 500, but polling by a processing job id still isn't semantically right — you get the group's status back with a null `request`. Arguably that should be a 404. That's a behavior change, so it's left for separate work.

## Related issues

Addresses [AB#204980](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/204980).

## Testing

Five regression tests added to `GetImportRequestHandlerTests`, covering status requested by a processing job id, a completed job with no input location, one with neither definition nor result, one with no result, and a failed job with no input location. All five fail on `main` with the original `ArgumentNullException` and pass with the fix.

Existing assertions on the group-status path were tightened to pin exact counts and input urls so the behavior is actually held in place rather than just not crashing.

Full `Microsoft.Health.Fhir.Core.UnitTests` suite compared against baseline: 1045 passed / 0 failed on `main`, 1050 passed / 0 failed on this branch. No new failures, clean build with no new warnings.

## FHIR Team Checklist

Patch

### DIFF
--- a/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Import/GetImportRequestHandlerTests.cs
+++ b/src/Microsoft.Health.Fhir.Core.UnitTests/Features/Operations/Import/GetImportRequestHandlerTests.cs
@@ -160,6 +160,109 @@ namespace Microsoft.Health.Fhir.Core.UnitTests.Features.Operations.BulkImport
             await Assert.ThrowsAsync<ResourceNotFoundException>(async () => await _mediator.GetImportStatusAsync(1, CancellationToken.None));
         }
 
+        [Fact]
+        public async Task WhenStatusIsRequestedByProcessingJobId_ThenOrchestratorJobIsSkippedAndResponseCodeShouldBeOk()
+        {
+            var orchestratorResult = new ImportOrchestratorJobResult() { Request = "Request" };
+            var orchestrator = new JobInfo()
+            {
+                Id = 0,
+                Status = JobStatus.Completed,
+                Result = JsonConvert.SerializeObject(orchestratorResult),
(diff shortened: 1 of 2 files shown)


mikaelweave (approved):
Comments could be fuller -- besides that approved

PR #5755, opened 2026-08-24
Extend middleware with Inbound Request Logs

fhibf:
Extend middleware with Inbound Request Logs

## Related issues
Addresses [AB#204637](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/204637).

## Testing
Tested these changes with unit and ad-hoc tests. 

## FHIR Team Checklist

### DIFF
--- a/src/Microsoft.Health.Fhir.Api/Features/Logging/HttpInboundRequestLogger.cs
+++ b/src/Microsoft.Health.Fhir.Api/Features/Logging/HttpInboundRequestLogger.cs
@@ -10,6 +10,7 @@ using EnsureThat;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Features;
 using Microsoft.Extensions.Logging;
+using Microsoft.Health.Fhir.Core.Features;
 
 namespace Microsoft.Health.Fhir.Api.Features.Logging
 {
@@ -23,6 +24,13 @@ namespace Microsoft.Health.Fhir.Api.Features.Logging
     /// </remarks>
     public sealed class HttpInboundRequestLogger : IHttpInboundRequestLogger
     {
+        internal const string DurationColumnName = "duration";
+        internal const string HttpHostColumnName = "httpHost";
+        internal const string HttpMethodColumnName = "httpMethod";
+        internal const string HttpPathColumnName = "httpPath";
+        internal const string HttpStatusCodeColumnName = "httpStatusCode";
+        internal const string XCorrelationIdColumnName = "XCorrelationId";
+
         private static readonly EventId RequestCompletedEventId = new EventId(1, "RequestCompleted");
         private static readonly EventId RequestProcessingErrorEventId = new EventId(2, "RequestProcessingError");
 
@@ -56,11 +64,12 @@ namespace Microsoft.Health.Fhir.Api.Features.Logging
 
                 IReadOnlyList<KeyValuePair<string, object>> state =
                 [
-                    new KeyValuePair<string, object>("duration", GetElapsedMilliseconds(context)),
-                    new KeyValuePair<string, object>("httpHost", context.Request?.Host.Value ?? string.Empty),
-                    new KeyValuePair<string, object>("httpMethod", context.Request?.Method ?? string.Empty),
-                    new KeyValuePair<string, object>("httpPath", context.Request?.Path.Value ?? string.Empty),
-                    new KeyValuePair<string, object>("httpStatusCode", httpStatusCode),
+                    new KeyValuePair<string, object>(DurationColumnName, GetElapsedMilliseconds(context)),
+                    new KeyValuePair<string, object>(HttpHostColumnName, context.Request?.Host.Value ?? string.Empty),
+                    new KeyValuePair<string, object>(HttpMethodColumnName, context.Request?.Method ?? string.Empty),
+                    new KeyValuePair<string, object>(HttpPathColumnName, context.Request?.Path.Value ?? string.Empty),
+                    new KeyValuePair<string, object>(HttpStatusCodeColumnName, httpStatusCode),
+                    new KeyValuePair<string, object>(XCorrelationIdColumnName, GetCorrelationId(context)),
                 ];
 
                 _logger.Log(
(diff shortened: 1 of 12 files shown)


src/Microsoft.Health.Fhir.Api/Features/Logging/HttpInboundRequestLogger.cs
```diff
@@ -0,0 +1,102 @@
...
+                    exception != null ? RequestProcessingErrorEventId : RequestCompletedEventId,
+                    state,
+                    exception,
+                    FormatLogMessage);
+            }
+            catch (Exception e)
+            {
+                // Generic catch to intercept any logging errors and avoid breaking the request pipeline. The logger should not throw exceptions.
+                _logger.LogError(e, "An error occurred while logging the HTTP Inbound request.");
+            }
```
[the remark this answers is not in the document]
fhibf: I'm keeping a generic catch to avoid errors in this part of the code to affect the entire request.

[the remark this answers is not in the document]
fhibf: As this part of the code encapsulates an external component, our team has no control over the possible errors that can be raised.

src/Microsoft.Health.Fhir.Api/Features/Logging/HttpInboundRequestLogger.cs
```diff
@@ -0,0 +1,102 @@
...
+                    new KeyValuePair<string, object>("httpHost", context.Request?.Host.Value ?? string.Empty),
+                    new KeyValuePair<string, object>("httpMethod", context.Request?.Method ?? string.Empty),
+                    new KeyValuePair<string, object>("httpPath", context.Request?.Path.Value ?? string.Empty),
+                    new KeyValuePair<string, object>("httpStatusCode", httpStatusCode),
+                ];
+
+                _logger.Log(
+                    httpStatusCode >= 500 && exception != null ? LogLevel.Error : LogLevel.Information,
+                    exception != null ? RequestProcessingErrorEventId : RequestCompletedEventId,
+                    state,
```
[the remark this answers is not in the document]
fhibf: @copilot - can you provide more details why it depends on an user-provided value?

src/Microsoft.Health.Fhir.Api/Features/Throttling/ThrottlingMiddleware.cs
```diff
@@ -309,6 +314,8 @@ private async Task Return429FromRequestRateExceededException(RequestRateExceeded
             context.Response.ContentLength = body.Length;
             context.Response.ContentType = ThrottledContentType;
 
+            _inboundRequestLogger.LogRequest(context, exception: exception);
```
rbans96: Careful with passing the exception here. Today this path logs a plain warning with nothing attached; now every throttled request produces a log record with `Exception != null`.

OTel processors often key off `LogRecord.Exception` to emit failure metrics regardless of log level, so a throttling storm would also spike exception metrics — right when things are already bad. A 429 is backpressure, not a failure.

Could you pass the type as a dimension instead? `new KeyValuePair<string, object>("exceptionType", exception.GetType().Name)` keeps the detail without reclassifying it.

src/Microsoft.Health.Fhir.Shared.Web/Startup.cs
```diff
@@ -355,6 +356,8 @@ private void AddTelemetryProvider(IServiceCollection services)
                     Startup.AddAzureMonitorOpenTelemetry(services, configuration);
                     break;
             }
+
+            services.AddSingleton<IHttpInboundRequestLogger, HttpInboundRequestLogger>();
```
rbans96: Would this be better in `AddFhirServer`?

Throttling and RuntimeState middleware now hard-require this logger, and both come from `FhirServerStartupFilter` in `AddFhirServer` — but the registration lives here. Any host that calls `AddFhirServer` without this Startup will blow up on the first request.

A `TryAddSingleton` in `FhirServerServiceCollectionExtensions` keeps it next to the code that needs it and still lets hosts override. It's also sitting inside `AddTelemetryProvider`, which it has nothing to do with.

src/Microsoft.Health.Fhir.Api/Features/Logging/HttpInboundRequestLogger.cs
```diff
@@ -0,0 +1,114 @@
...
+        {
+            EnsureArg.IsNotNull(context, nameof(context));
+
+            try
+            {
+                int httpStatusCode = context.Response?.StatusCode ?? 0;
+
+                IReadOnlyList<KeyValuePair<string, object>> state =
+                [
+                    new KeyValuePair<string, object>("duration", GetElapsedMilliseconds(context)),
```
rbans96: These names are a contract with the exporter's column mapping, but nothing checks them. Typo one and the export still succeeds — the column just shows up empty. No build break, no failing test, easy to miss in review.

Worth pulling them out as `internal const string` so tests and downstream code can reference them instead of repeating literals.

src/Microsoft.Health.Fhir.Api/Features/Logging/HttpInboundRequestLogger.cs
```diff
@@ -0,0 +1,114 @@
...
+                    new KeyValuePair<string, object>("httpHost", SanitizeForLog(context.Request?.Host.Value)),
+                    new KeyValuePair<string, object>("httpMethod", SanitizeForLog(context.Request?.Method)),
+                    new KeyValuePair<string, object>("httpPath", SanitizeForLog(context.Request?.Path.Value)),
+                    new KeyValuePair<string, object>("httpStatusCode", httpStatusCode),
```
rbans96: Could you add the correlation id here? Without it these rows can't be joined to anything else for a given request, which is most of the value.

It's already available — `FhirServerStartupFilter` runs `UseFhirRequestContext()` (which generates the correlation id) before both `UseFhirRuntimeState()` and `UseThrottling()`, so `RequestContextAccessor<IFhirRequestContext>` is populated by the time this is called.

rbans96 (commented):
A few notes from reading through this.

src/Microsoft.Health.Fhir.Api/Features/Throttling/ThrottlingMiddleware.cs
fhibf: Removed the exception logging from this context.
I believe current monitors are looking for exceptions in FhirLog. 
As these logs should go to another table, then it should be fine.

src/Microsoft.Health.Fhir.Api/Features/Logging/HttpInboundRequestLogger.cs
fhibf: Addressed as suggested.

src/Microsoft.Health.Fhir.Shared.Web/Startup.cs
fhibf: Updated as suggested.

rbans96 (approved):

PR #5756, opened 2026-08-24
Extract shared collection definitions and add unit test manifest guard

mikaelweave:
## Description

Preparatory refactor ahead of the xUnit v2 → v3 migration. **Everything here is inert on xUnit v2** — it is split out so the migration itself reviews as a single mechanical change.

This is the **bottom layer of a 3-PR stack**:

| Layer | Contents | Status |
|---|---|---|
| **1 (this PR)** | Shared collection definitions + unit test project manifest guard | ← you are here |
| 2 | Core migration: `Extensions.Xunit` rewrite, CPM/props switch, ~109 test sources, 27 csproj, Tests.Common guards, **CI pipeline YAML** | #5758 |
| 3 | `Extensions.Xunit.UnitTests` harness + `.TestAssets` | follows |

Layer 2 is irreducible: `Directory.Packages.props` deletes the xUnit v2 package versions globally under Central Package Management, `Directory.Build.props` fans the MTP references out to every project matching `Contains('Test')`, and `Microsoft.Health.Extensions.Xunit` binds a single xUnit major that every test project references. Those three jointly force one atomic flip, which is why this prep layer contains only changes that are already correct on v2. The CI pipeline YAML rides with layer 2 for a related reason: `main` already runs xUnit v2 under Microsoft Testing Platform via the `YTest.MTP.XUnit2` shim, which accepts VSTest-style `--filter`, whereas xUnit v3's MTP requires `--filter-query`. Swapping the shim without flipping the YAML would leave every integration and E2E leg selecting nothing.

### Collection definitions

Collection definitions are moved out of the test classes that declared them, onto dedicated trait-free definition classes:

- **`ModelInfoProviderCollection`** — serialises the SqlServer unit test classes that read or write the process-global `ModelInfoProvider`. A class fixture only orders the classes that share it, so it cannot stop a third class overwriting the same static; one shared collection can.
- **`CustomQueriesTestsCollection`** — serialises the CustomQueries tests, which share static state.
- **`IndexAndReindexCollection`** — replaces the duplicate `[CollectionDefinition]` that `ReindexTests` and `CustomSearchParamTests` *each* declared for the same collection name. The definition deliberately carries **no traits**: a trait on a shared definition is inherited by every member of the collection, and a CI leg filtering that category out would silently drop members that only joined to be serialised.

### Manifest guard

Adds `build/jobs/scripts/Assert-UnitTestProjectsDiscovered.ps1` and its manifest. The unit test leg runs whatever `**/*UnitTests/*.csproj` matches, so a project that leaves the glob — renamed, moved, or its directory suffix quietly changed — stops being tested without anything failing. The script checks the glob against the manifest before the tests start.

It is **not wired into the pipeline yet**; the YAML that calls it lands with the migration. Its automated tests live in the harness project in layer 3, so this script ships here without them.

## Related issues

Addresses [AB#189958](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/189958).

## Testing

Verified by execution on the unmodified v2 tree:

- `Microsoft.Health.Fhir.SqlServer.UnitTests` — **1053 passed / 0 failed, both with and without this change**, confirming the collection extraction is a true no-op (identical totals rule out silently-dropped tests).
- `Assert-UnitTestProjectsDiscovered.ps1` — prints `Matched 19 unit test project(s) against 19 expected`, exit code 0. The manifest reads 19 here and goes to 20 in layer 3, when the harness project exists.
- `Microsoft.Health.Fhir.R4.Tests.E2E` — compiles clean (0 warnings, 0 errors) with `IndexAndReindexCollection.cs` explicitly registered in `Microsoft.Health.Fhir.Shared.Tests.E2E.projitems`, so the new shared file genuinely compiles into consumers rather than being silently skipped.

Note: `dotnet build` of the E2E projects trips the repo's `VerifyExactSdkVersion` target locally because the installed SDK is 10.0.303 while `global.json` pins 10.0.302. That is a pre-existing local-environment condition unrelated to this change, so `global.json` was left untouched and the E2E verification was run via the `Compile` target instead.

## FHIR Team Checklist

Skip (test-only preparatory refactor, no product code or behavior change)

### DIFF
--- a/build/jobs/scripts/Assert-UnitTestProjectsDiscovered.ps1
+++ b/build/jobs/scripts/Assert-UnitTestProjectsDiscovered.ps1
@@ -1,0 +1,102 @@
+<#
+.SYNOPSIS
+    Fails when the unit test leg's project glob matches fewer projects than expected.
+
+.DESCRIPTION
+    The unit test leg does not name its projects: it runs whatever '**/*UnitTests/*.csproj'
+    matches. Every other leg guards against running a fraction of its tests by counting the tests
+    a filter discovers, but that shape does not fit here, because this leg's failure is a whole
+    project leaving the glob - renamed, moved, or a directory suffix quietly changed - rather than
+    a filter selecting too little. Nothing notices: the remaining projects run, every one of them
+    passes, and the leg reports success with an entire assembly's tests missing.
+
+    --minimum-expected-tests cannot cover this either. It is applied per assembly, so it says
+    nothing about an assembly that was never run, and it has to stay at 1 because the runner
+    re-applies it to the much smaller pass that --retry-failed-tests starts.
+
+    Each expected project is checked by path rather than by counting them, because a count only
+    notices projects leaving while none arrive: renaming one project and adding another in the same
+    change leaves the count where it was, and the renamed project stops being tested silently.
+
+    Adding projects is fine; only losing one that is listed fails.
+
+.PARAMETER SourcesDirectory
+    The repository root to search.
+
+.PARAMETER ManifestPath
+    The file listing the repository-relative path of every project this leg is expected to run.
+#>
+#requires -Version 7.0
+
+[CmdletBinding()]
+param(
+    [Parameter(Mandatory = $true)]
+    [string] $SourcesDirectory,
+
+    [Parameter(Mandatory = $true)]
+    [string] $ManifestPath
+)
+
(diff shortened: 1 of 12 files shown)


mikaelweave:
Superseded by #5761.

PR #5507, opened 2026-04-17
Xunit2 -> Xunit3 Migration

mikaelweave:
## Description

- Migrates the entire test suite from xUnit v2 to xUnit v3 running on Microsoft Testing Platform, replacing VSTest as the runner across every CI leg.
- Rewrites this repo's custom xUnit extensions — fixture argument sets, `[RetryFact]`/`[RetryTheory]`, and skippable tests — against v3's new discovery and execution APIs, which have no v2 equivalents.
- Fixes several ways the v3 move could have silently lost tests: collection traits not propagating to fixture argument set variants, a missing framework attribute disabling discovery for a whole assembly, and retried failures being reported as skips.
- Adds CI guards so a leg that discovers or runs nothing fails loudly instead of exiting green: per-leg discovery floors, a checked-in manifest of expected unit test projects, and reconciliation of every retry attempt against the final TRX.
- Adds a test harness for the extensions themselves (150 tests over 34 purpose-built scenario programs), since none of this behaviour was previously covered by anything.

## Changes overview

The PR touches **293 files (+12,738 / −968)**, but roughly two-thirds of that is new test code that adds no production behaviour:

| Area | Files | Lines | Review priority |
|---|---|---|---|
| `src/Microsoft.Health.Extensions.Xunit/` — the custom framework | 21 | +3,067 / −631 | **This is the PR.** |
| `build/jobs/` — pipeline YAML + guard scripts | 8 | +702 / −28 | **Read this second.** |
| `src/Microsoft.Health.Extensions.Xunit.UnitTests/` — new harness | 48 | +5,014 | New tests, additive |
| `src/Microsoft.Health.Extensions.Xunit.TestAssets/` — new scenarios | 68 | +2,966 | New fixtures, additive |
| Test sources across the repo | 108 | +903 / −195 | Mechanical |
| `.csproj` / props / solution files | 33 | +56 / −57 | Mechanical |

**Where to actually look:**

- `CustomXunitTestFrameworkDiscoverer.cs` (+1,399), `RetryTestCase.cs` (+698) and `CustomXunitTestFrameworkExecutor.cs` (+341) hold nearly all the real logic — v2's discoverer/executor surface no longer exists, so these were rewritten rather than edited.
- `build/jobs/scripts/Assert-RetriedFailuresPassed.ps1` (+155) and `Assert-UnitTestProjectsDiscovered.ps1` (+102) are the new safety nets; the five YAML files mostly wire them in and switch each leg to the MTP runner.

[...]

PR #5759, opened 2026-08-24
Add the xUnit extensions test harness and test assets (3/3)

mikaelweave:
**Layer 3 of 3** in the stack splitting #5507 (xUnit v2 -> v3 + Microsoft Testing Platform migration). Targets `mikaelweave-xunit3-core-migration` (L2, #5758), which targets `mikaelweave-xunit3-prep-collection-definitions` (L1, #5756).

Merge L1, then L2, then this.

## What this adds

The test harness for the xUnit extensions themselves, held back from L2 so that the v3 flip could be reviewed on its own.

- `src/Microsoft.Health.Extensions.Xunit.UnitTests/` (48 files) - the harness covering the extensions.
- `src/Microsoft.Health.Extensions.Xunit.TestAssets/` (68 files) - 34 deliberately-misbehaving mini xUnit programs used as fixture data. `TestAssetRunner` launches one per scenario as a subprocess and asserts on the published TRX plus the process exit code.

Plus the six coupling edits L2 could not carry because the harness did not exist yet: the solution project entry, its four build-configuration lines and its `NestedProjects` mapping; the `R4.slnf`/`R5.slnf` entries; the unit test project manifest going from 19 back to 20; and the `InternalsVisibleTo` entry in `Microsoft.Health.Extensions.Xunit`.

## Why `TestAssets` is not in the solution

Deliberate, and please keep it that way. The assets are programs that are *supposed* to fail. The project sets `IsTestProject=false` before the repo props import and is absent from the solution and both solution filters, so nothing discovers them or runs their failing tests directly. They are built solely through a `ReferenceOutputAssembly="false"` `ProjectReference` from the harness.

## Verification

| Check | Result |
|---|---|
| Stack reconstructs the reviewed tree exactly | `git diff HEAD b7513423c5e40ab9dda4e338d66dea41a6c8cfda` is **empty** |
| Solution build | 0 errors attributable to this change |
| Harness suite | **150/150 passed, 0 failed, 0 skipped** |
| Unit test manifest guard | 20/20 matched, exit 0 |
| Solution integrity | Both `.slnf` valid JSON, harness present, `TestAssets` absent |
| Tree | Clean, single commit |

The solution build reports 4 errors from the repo's pre-existing `VerifyExactSdkVersion` target (installed SDK 10.0.303 vs `global.json` 10.0.302), all in E2E projects and unrelated to this change. `global.json` was deliberately not touched. Verified around it the same way L2 did, by compiling the four E2E projects individually with `-t:Compile -p:TargetFramework=net10.0`: all four returned 0 errors, 0 warnings.

## Reviewer note

`TestAssetRunner` uses `--filter-query`, which is assembly-wide rather than namespace-scoped. Adding further concurrent query-filtered runs will destabilise sibling tests and surface as MTP exit code 8 / "zero tests ran" in tests you did not touch.


mikaelweave:
Superseded by #5761.

PR #5789, opened 2026-09-03
Adds logging for search params and normalizes reindex log messages.

SergeyGaluzo:
Fixes https://microsofthealth.visualstudio.com/Health/_workitems/edit/204978/
by adding logging in SqlServerFhirDataStore for pending search param statuses

Normalizes reindex messages to facilitate debugging (reindex word is part of every message)

Also corrects (makes plural) method names GetAllSearchParameterStatus**es** and ApplySearchParameterStatus**es**

### DIFF
--- a/src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexOrchestratorJob.cs
+++ b/src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexOrchestratorJob.cs
@@ -141,7 +141,7 @@ namespace Microsoft.Health.Fhir.Core.Features.Operations.Reindex
 
                 await RefreshSearchParameterCache(true);
 
-                _logger.LogInformation("Reindex job with Id: {Id} has been started. Status: {Status}.", _jobInfo.Id, _jobInfo.Status);
+                _logger.LogJobInformation(_jobInfo, $"The reindex job started. Id={_jobInfo.Id} isSql={_isSql}");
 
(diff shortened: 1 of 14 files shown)


mikaelweave (approved):

rajithaalurims (approved):

src/Microsoft.Health.Fhir.Core/Features/Operations/Reindex/ReindexOrchestratorJob.cs
```diff
@@ -141,7 +141,7 @@ public async Task<string> ExecuteAsync(JobInfo jobInfo, CancellationToken cancel
 
                 await RefreshSearchParameterCache(true);
 
-                _logger.LogInformation("Reindex job with Id: {Id} has been started. Status: {Status}.", _jobInfo.Id, _jobInfo.Status);
+                _logger.LogJobInformation(_jobInfo, $"The reindex job started. Id={_jobInfo.Id} isSql={_isSql}");
```
brendankowitz, line 144: @mikaelweave - Doesn't this break structured logging? are we omitting the parameters to a single formatted string on purpose?

SergeyGaluzo: In this particular case I removed **Id** from analytics on "main" string, but I added it back via Log**Job**Info. Status field is completely removed because it was completely meaningless as it always has same value Started.

PR #5793, opened 2026-09-05
Import CPU by wall clock

SergeyGaluzo:
Added ability to measure import CPU consumption based on walk-clock method:
1. Defined curated sample dataset of 1000-resources based on a large baseline corpus on synthetic hospital data used in all prior $import testing. Algorithm was using per-resource-type proportional allocation plus a fixed seed. This set is part of solution and is lazily loaded in memory once. This eliminates latency in reading input resources from storage.
2. Introduced inmemorytest:// input support with synthetic processing-job fan-out (InMemoryTestProcessingJobs).
3. Added per-job wall-clock timing (ClockMilliseconds) and aggregated database calls latency (DatabaseMilliseconds).
4. Added in-memory execution stats to standard import output (only if InMemoryTestProcessingJobs > 0).
5. Added aggregate cpu_msec_per_resource (CPU normalized by actual processed resources).

Sample $import registration

POST {{fhir}}/$import
Content-Type: application/fhir+json
Prefer: respond-async

{
  "resourceType": "Parameters",
  "parameter": [
    { "name": "inputFormat", "valueString": "application/fhir+ndjson" },
    { "name": "input", "part": [ { "name": "url", "valueUri": "inmemorytest://whatever" } ] },
    { "name": "mode", "valueString": "IncrementalLoad" },
    { "name": "inMemoryTestProcessingJobs", "valueInteger": 1000 }
  ]
}


src/Microsoft.Health.Fhir.Core/Features/Operations/Import/GetImportRequestHandler.cs
```diff
@@ -58,7 +65,7 @@ public async Task<GetImportResponse> HandleAsync(GetImportRequest request, Cance
             }
             else if (coord.Status == JobStatus.Failed)
             {
-                var errorResult = DeserializeOrDefault<ImportJobErrorResult>(coord.Result);
+                var errorResult = JsonConvert.DeserializeObject<ImportJobErrorResult>(coord.Result);
                 if (errorResult.HttpStatusCode == 0)
```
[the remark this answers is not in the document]
SergeyGaluzo: Current workflow guarantees that jobs with Failed status always have Result set.

src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/ImportOrchestratorJob.cs
```diff
@@ -237,32 +237,57 @@ internal static async Task SendNotification<T>(JobStatus status, JobInfo info, l
...
+                for (var i = 0; i < coordDefinition.InMemoryTestProcessingJobs; i++)
                 {
-                    var newInput = input.Clone();
-                    newInput.Offset = offset;
-                    newInput.BytesToRead = bytesToRead;
-                    lock (inputs)
+                    result.TotalBytes += blobLength;
+                    var syntheticInput = template.Clone();
+                    syntheticInput.Url = new Uri($"{IntegrationDataStoreClientConstants.InMemoryTestSourceScheme}://{template.Url.Host}-{i}");
+                    AddNewInput(inputs, syntheticInput, 0, bytesToRead);
```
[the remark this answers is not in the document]
SergeyGaluzo: This is never true for curated in-memory data.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ImportController.cs
```diff
@@ -249,8 +250,16 @@ private void ValidateImportRequestConfiguration(ImportRequest importData)
             }
         }
 
-        private bool IsConfiguredStorageAccountEndpoint(Uri inputUri)
+        private bool IsConfiguredStorageAccountEndpoint(Uri inputUri, bool allowInMemoryTestSource)
         {
+            // Only recognized when the request itself opts into in-memory test processing via
+            // InMemoryTestProcessingJobs; this never applies to ordinary import requests.
+            if (allowInMemoryTestSource
+                && string.Equals(inputUri.Scheme, IntegrationDataStoreClientConstants.InMemoryTestSourceScheme, StringComparison.OrdinalIgnoreCase))
```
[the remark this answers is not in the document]
SergeyGaluzo: Fixed

src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/ImportOrchestratorJob.cs
```diff
@@ -237,32 +237,57 @@ internal static async Task SendNotification<T>(JobStatus status, JobInfo info, l
...
-                var blobLength = (long)(await _integrationDataStoreClient.GetPropertiesAsync(input.Url, cancellationToken))[IntegrationDataStoreClientConstants.BlobPropertyLength];
-                result.TotalBytes += blobLength;
-                var bytesToRead = coordDefinition.ProcessingUnitBytesToRead == 0
-                                ? BytesToReadDefault
-                                : coordDefinition.ProcessingUnitBytesToRead;
-                foreach (var offset in GetOffsets(blobLength, bytesToRead))
+                // Test-only: use the first provided input as a template and generate the requested number of
+                // in-memory processing jobs from it (each with its own unique inmemorytest:// URL), so tests can
+                // exercise orchestrator scaling without needing to construct that many distinct input files.
+                var template = coordDefinition.Input[0];
```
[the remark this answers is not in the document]
SergeyGaluzo: Fixed

src/Microsoft.Health.Fhir.SqlServer/Features/Storage/SqlServerFhirDataStore.cs
```diff
@@ -244,12 +265,12 @@ private async Task<MergeOutcome> MergeInternalAsync(IReadOnlyList<ResourceWrappe
...
-            var existingResources = (await GetAsync(resources.Select(r => r.Wrapper.ToResourceKey(true)).Distinct().ToList(), true, useReplicasForReads, cancellationToken)).ToDictionary(r => r.ToResourceKey(true), r => r);
+            var existingResources = (await ExecTime(() => GetAsync(resources.Select(r => r.Wrapper.ToResourceKey(true)).Distinct().ToList(), true, useReplicasForReads, cancellationToken), addDatabaseMilliseconds)).ToDictionary(r => r.ToResourceKey(true), r => r);
 
             // Assume that most likely case is that all resources should be updated.
             // TODO: MergeResourcesBeginTransaction accepts new parameter allowing to throw exception on overload.
             // TODO: Set this parameter to true when 429 instead of intenal waits is desired. Make sure that exception is NOT thrown only for API calls.
-            (var transactionId, var minSequenceId) = await StoreClient.MergeResourcesBeginTransactionAsync(resources.Count, cancellationToken);
+            (var transactionId, var minSequenceId) = await ExecTime(() => StoreClient.MergeResourcesBeginTransactionAsync(resources.Count, cancellationToken), addDatabaseMilliseconds);
```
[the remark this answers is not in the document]
SergeyGaluzo: I don’t think this internal retry should null DatabaseMilliseconds. In these stats, “retried” effectively means “exclude from CPU computation,” not “any SQL call performed more than one attempt.” The earlier nulling was needed because retry logic crossed multiple blocks and could leave partial DB timings that made clock_msec - database_msec unreliable. MergeResourcesBeginTransactionAsync is different: the retry/backoff/success path is fully contained within one ExecTime block, so any added wait increases both clock and DB elapsed time by the same amount. CPU remains computable. Nulling DB ms here would exclude a job whose CPU math is still valid.

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/GetImportRequestHandler.cs
```diff
@@ -105,7 +108,50 @@ public async Task<GetImportResponse> HandleAsync(GetImportRequest request, Cance
...
+                            var elapsedmsec = (jobLines.Max(_ => _.EndDate) - jobLines.Min(_ => _.StartDate)).TotalMilliseconds;
+                            var parallelism = elapsedmsec > 0 ? Math.Round(jobmsec / elapsedmsec, 2) : 0;
+                            var jobsWithReliableDatabaseTiming = jobLines.Where(_ => _.DatabaseMilliseconds.HasValue).ToList();
+                            var resourceCount = jobsWithReliableDatabaseTiming.Sum(_ => _.ResourceCount);
+                            var cpuMillisecondsPerResource = resourceCount > 0
+                                ? Math.Round((double)jobsWithReliableDatabaseTiming.Sum(_ => _.CpuMilliseconds.Value) / resourceCount, 2)
+                                : (double?)null;
+                            var executionStats = new List<string> { $"jobs={jobLines.Count} cpu_msec_per_resource={cpuMillisecondsPerResource:F2} clock_msec={jobLines.Sum(_ => _.ClockMilliseconds)} database_msec={jobLines.Sum(_ => _.DatabaseMilliseconds)} retried_jobs={retriedJobs} parallelism={parallelism:F2}" };
+                            executionStats.AddRange(jobLines.Take(50).Select(x => x.Line));
```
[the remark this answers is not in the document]
SergeyGaluzo: Fixed

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/GetImportRequestHandler.cs
```diff
@@ -85,26 +93,66 @@ public async Task<GetImportResponse> HandleAsync(GetImportRequest request, Cance
...
+                    // Include per-job execution stats only for in-memory test imports (for CPU profiling)
+                    if (coordDefinition.InMemoryTestProcessingJobs > 0)
+                    {
+                        var jobLines = jobs.Select(job => jobResultsById.TryGetValue(job.Id, out var result) ? new { Job = job, Result = result } : null).Where(_ => _ != null)
+                            .OrderByDescending(_ => _.Job.StartDate.Value)
+                            .Select(_ =>
+                            {
+                                var clockMilliseconds = _.Result.ClockMilliseconds;
+                                var databaseMilliseconds = _.Result.DatabaseMilliseconds;
+                                var cpuMilliseconds = clockMilliseconds - databaseMilliseconds; // x - null = null
```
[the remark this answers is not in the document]
SergeyGaluzo: For a single 1000-record batch, parser CPU does not overlap with database work. The importer waits until all 1000 resources are read and parsed before calling ImportResourcesInBuffer.
Parsing can overlap DB work only when the job contains multiple batches: while batch N is persisted, the loader may parse batch N+1 into the channel.
Therefore, the general limitation of clock - database remains, but the parser-overlap argument does not apply to this single-batch case.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ImportController.cs
```diff
@@ -211,6 +212,16 @@ private void ValidateImportRequestConfiguration(ImportRequest importData)
                 throw new RequestNotValidException(string.Format(Resources.ImportRequestValueNotValid, nameof(input)));
             }
 
+            if (importData.InMemoryTestProcessingJobs > 0
+                && (!_importConfig.InMemoryTestEnabled
+                    || importData.InMemoryTestProcessingJobs > 1_000_000
+                    || input.Count != 1
+                    || input[0].Url == null
+                    || !string.Equals(input[0].Url.Scheme, IntegrationDataStoreClientConstants.InMemoryTestSourceScheme, StringComparison.OrdinalIgnoreCase)))
```
brendankowitz, line 221: **[P2] Validate absolute URLs before reading `Scheme`**

With `InMemoryTestEnabled=true`, a positive `InMemoryTestProcessingJobs` value, and one input, a relative URL such as `relative.ndjson` reaches this expression. `ParametersExtensions.TryGetUriValue` accepts `UriKind.RelativeOrAbsolute`, but accessing `Scheme` on that URI throws `InvalidOperationException`. The later `IsAbsoluteUri` check is never reached, so malformed input returns HTTP 500 instead of the intended 400.

Please reject relative URLs before accessing `Scheme` and add a regression test with the synthetic-source flag enabled and a positive job count.

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/GetImportRequestHandler.cs
```diff
@@ -105,7 +108,50 @@ public async Task<GetImportResponse> HandleAsync(GetImportRequest request, Cance
...
+                        if (jobLines.Count > 0)
+                        {
+                            var retriedJobs = jobLines.Count(x => x.DatabaseMilliseconds is null);
+                            var jobMsec = jobLines.Sum(_ => (_.EndDate - _.StartDate).TotalMilliseconds);
+                            var elapsedMsec = (jobLines.Max(_ => _.EndDate) - jobLines.Min(_ => _.StartDate)).TotalMilliseconds;
+                            var parallelism = elapsedMsec > 0 ? jobMsec / elapsedMsec : -1;
+                            var jobsToCount = jobLines.Where(_ => _.DatabaseMilliseconds.HasValue).OrderBy(_ => _.ClockMilliseconds).Take((int)(jobLines.Count * 0.3)).ToList(); // take first 30% to make std lower
+                            var resCnt = jobsToCount.Sum(_ => _.ResourceCount);
+                            var cpu = (double?)jobsToCount.Sum(_ => _.CpuMilliseconds.Value) / resCnt;
+                            var std = Math.Sqrt(jobsToCount.Sum(_ => _.ResourceCount * Math.Pow(((double)_.CpuMilliseconds.Value / _.ResourceCount) - cpu.Value, 2)) / resCnt);
```
brendankowitz: **[P2] Handle an empty timing sample before calculating statistics**

For 1–3 completed, nonempty workers, `(int)(jobLines.Count * 0.3)` is zero, so `Take(0)` discards every measurement even when all timings are valid. `resCnt` then becomes zero, and both the mean and standard deviation render as `NaN`. The same problem occurs when every worker has a null database timing. This affects valid small benchmark requests and intermediate polls while a larger import is running.

Please define the small-sample policy and handle an empty eligible population explicitly before dividing—either use a nonempty sample when valid measurements exist or report that statistics are unavailable. Add focused cases for 1–3 workers and all-null database timings; the current E2E field-presence assertions do not catch these invalid values.

brendankowitz (commented):
Two correctness issues in the enabled synthetic-import path, with details inline.

src/Microsoft.Health.Fhir.Shared.Api/Controllers/ImportController.cs
SergeyGaluzo: Fixed

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/GetImportRequestHandler.cs
SergeyGaluzo: Fixed

SergeyGaluzo:
> Two correctness issues in the enabled synthetic-import path, with details inline.

@brendankowitz Fixed both.

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/GetImportRequestHandler.cs
```diff
@@ -105,7 +108,56 @@
...
+                            var jobMsec = jobLines.Sum(_ => (_.EndDate - _.StartDate).TotalMilliseconds);
+                            var elapsedMsec = (jobLines.Max(_ => _.EndDate) - jobLines.Min(_ => _.StartDate)).TotalMilliseconds;
+                            var parallelism = elapsedMsec > 0 ? jobMsec / elapsedMsec : -1;
+                            var validJobs = jobLines.Where(_ => _.DatabaseMilliseconds.HasValue).OrderBy(_ => _.ClockMilliseconds).Take((int)(jobLines.Count * 0.3)).ToList(); // take first 30% to make std lower
+                            var resCnt = validJobs.Sum(_ => _.ResourceCount);
+                            string cpuStr = null;
+                            if (validJobs.Count >= 3) // it does not make sense to compute statistical values for low counts.
+                            {
+                                var cpu = (double?)validJobs.Sum(_ => _.CpuMilliseconds.Value) / resCnt;
+                                var std = Math.Sqrt(validJobs.Sum(_ => _.ResourceCount * Math.Pow(((double)_.CpuMilliseconds.Value / _.ResourceCount) - cpu.Value, 2)) / resCnt);
```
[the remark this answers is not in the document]
SergeyGaluzo: fixed

src/Microsoft.Health.Fhir.Azure/IntegrationDataStore/AzureBlobIntegrationDataStoreClient.cs
```diff
@@ -49,6 +47,11 @@ public Stream DownloadResource(Uri resourceUri, long startOffset, CancellationTo
         {
             EnsureArg.IsNotNull(resourceUri, nameof(resourceUri));
 
+            if (InMemoryTestDataSource.IsTestSourceUri(resourceUri))
```
mikaelweave, line 50: Should we also gate this on the feature flag `InMemoryTestEnabled` from `ImportJobConfiguration`?

src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/ImportOrchestratorJob.cs
```diff
@@ -237,32 +237,57 @@ internal static async Task SendNotification<T>(JobStatus status, JobInfo info, l
...
+                var blobLength = (long)(await _integrationDataStoreClient.GetPropertiesAsync(template.Url, cancellationToken))[IntegrationDataStoreClientConstants.BlobPropertyLength];
+                for (var i = 0; i < coordDefinition.InMemoryTestProcessingJobs; i++)
                 {
-                    var newInput = input.Clone();
-                    newInput.Offset = offset;
-                    newInput.BytesToRead = bytesToRead;
-                    lock (inputs)
+                    result.TotalBytes += blobLength;
+                    var syntheticInput = template.Clone();
```
mikaelweave, line 255: Do we need clone twice? First on 255, then again on `AddNewInput` call?

src/Microsoft.Health.Fhir.Azure/IntegrationDataStore/InMemoryTestDataSource.cs
```diff
@@ -0,0 +1,92 @@
...
+        /// <summary>
+        /// Determines whether the given resource URI refers to the in-memory test source.
+        /// Any URI with the <see cref="IntegrationDataStoreClientConstants.InMemoryTestSourceScheme"/> scheme
+        /// (e.g., "inmemorytest://representative-import-1000", "inmemorytest://batch-1", etc.) is accepted
+        /// and serves the same embedded payload. This allows tests to use multiple distinct URLs without
+        /// requiring separate real storage blobs.
+        /// </summary>
+        internal static bool IsTestSourceUri(Uri resourceUri)
+        {
+            return resourceUri != null && string.Equals(resourceUri.Scheme, IntegrationDataStoreClientConstants.InMemoryTestSourceScheme, StringComparison.OrdinalIgnoreCase);
```
mikaelweave, line 40: Null check redundant / extra check as all callers already ensure not null. 

Not needed but safety if other callers will be added later.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Import/ImportTests.cs
```diff
@@ -802,6 +807,38 @@ public async Task GivenIncrementalLoad_MultipleResoureTypesInSingleFile_Success(
...
+            _testOutputHelper.WriteLine(statsJson);
+            Console.WriteLine("========== ExecutionStats ==========");
+            Console.WriteLine(statsJson);
+            Console.WriteLine("====================================");
+
+            Assert.StartsWith($"jobs_total={jobs} ", result.ExecutionStats.First());
+            Assert.Contains("cpu_msec_per_resource=", result.ExecutionStats.First()); // only emitted when timings were actually collected
+            var jobLines = result.ExecutionStats.Where(l => l.StartsWith("job=")).ToList();
+            Assert.Equal(50, jobLines.Count); // max output is 50
+            foreach (var jobLine in jobLines)
```
mikaelweave: You could use Assert.All vs foreach over the list

```
Assert.All(jobLines, jobLine =>
{
    Assert.Contains("succeeded=1000", jobLine);
    Assert.Contains("failed=0", jobLine);
    Assert.Contains("cpu_msec=", jobLine);
    Assert.Contains("clock_msec=", jobLine);
    Assert.Contains("database_msec=", jobLine);
});
```

https://xunit.net/releases/v2/2.5.0-assertions#assertall

mikaelweave (commented):
A couple minor items, PR is in good shape.

src/Microsoft.Health.Fhir.Azure/IntegrationDataStore/AzureBlobIntegrationDataStoreClient.cs
SergeyGaluzo: No. There is no way to get special resourceUri other than for input controller. And it checks the flag.

src/Microsoft.Health.Fhir.Azure/IntegrationDataStore/InMemoryTestDataSource.cs
SergeyGaluzo: Well. This is a difficult one. This class is static and there is no protection inside it. I would leave as is.

test/Microsoft.Health.Fhir.Shared.Tests.E2E/Rest/Import/ImportTests.cs
SergeyGaluzo: Fixed.

src/Microsoft.Health.Fhir.SqlServer/Features/Operations/Import/ImportOrchestratorJob.cs
SergeyGaluzo: For in-memory - no. But AddNewInput is shared with normal processing for which we do need to clone. I suggest to leave as is for soimplicity.

src/Microsoft.Health.Fhir.Core/Features/Operations/Import/GetImportRequestHandler.cs
```diff
@@ -105,7 +107,60 @@ public async Task<GetImportResponse> HandleAsync(GetImportRequest request, Cance
...
+                                cpuStr = $" cpu_msec_per_resource={cpu:F2} std_cpu_msec_per_resource={std:F2}";
+                            }
+
+                            // database_msec skips retried jobs by design; jobs_retried reports how many.
+                            var executionStats = new List<string> { $"jobs_total={jobLines.Count} jobs={validJobs.Count}{cpuStr} clock_msec={jobLines.Sum(_ => _.ClockMilliseconds)} database_msec={jobLines.Sum(_ => _.DatabaseMilliseconds)} jobs_retried={retriedJobs} parallelism={parallelism:F2}" };
+                            executionStats.AddRange(jobLines.Take(50).Select(x => x.Line));
+
+                            result.ExecutionStats = executionStats;
+                        }
+                    }
```
brendankowitz, line 162: For readability I'd probably move this to its own method and call from here.
nit: still never been a an of using discards as variables 😃

brendankowitz (approved):

mikaelweave (approved):

PR #2813, opened 2022-09-20
Updates to Central Package Management

brendankowitz:
## Description
Updates nuget versions to Central Package Management

## Testing
It shall build with harmonious package version simplicity. 

## FHIR Team Checklist
- [x] **Update the title** of the PR to be succinct and less than 50 characters
- [x] **Add a milestone** to the PR for the sprint that it is merged (i.e. add S47)
- [x] Tag the PR with the type of update: **Bug**, **Dependencies**, **Enhancement**, or **New-Feature**
- [x] Tag the PR with Azure API for FHIR if this will release to the Azure API for FHIR managed service (CosmosDB or common code related to service)
- [x] Tag the PR with Azure Healthcare APIs if this will release to the Azure Healthcare APIs managed service (Sql server or common code related to service)
- [x] CI is green before merge
- Review [squash-merge requirements](https://github.com/microsoft/fhir-server/blob/master/SquashMergeRequirements.md)

### Semver Change ([docs](https://github.com/microsoft/fhir-server/blob/master/docs/Versioning.md))
Skip

PR #5821, opened 2026-09-15
Fix Dependabot job failure and 55 minute timeout

mikaelweave:
## Summary

Dependabot's NuGet updates were failing, then timing out, and routine package updates had quietly stopped appearing for months. This fixes both, and makes the fix hard to break by accident.

[AB#207007](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/207007)

## The two problems

**1. The job crashed.** `Sql.targets` was imported unconditionally. Dependabot evaluates project files without restoring, which leaves `NuGetPackageRoot` empty, so the import threw `MissingFileException` and took the whole update job down.

CleanShot 2026-09-15 at 16 14 47@2x

**2. Once it stopped crashing, it timed out.** Dependabot re-runs a *full* dependency discovery pass before **every** group in `dependabot.yml`. Pointed at `/`, each pass expanded `Microsoft.Health.Fhir.sln` and all 78 projects — about four minutes a group, against a fixed 55 minute limit GitHub puts on Dependabot jobs. The job died partway through, so the later groups, including the catch-all `minor-and-patch`, never ran at all. That is why e.g. NSubstitute sat on 6.0.0 while 6.2.0 was available.

## The fix

`dependabot.yml` now points the nuget ecosystem at `build/DependabotDiscovery`, a stub project listing every package in `Directory.Packages.props`. Discovery resolves one project instead of 78. Central Package Management still supplies the versions, and `Directory.Packages.props` is still the file Dependabot edits, so the PRs are unchanged.

Verified end to end on a fork:

| | Before | After |
|---|---|---|
| Outcome | 55m00s timeout, failed | **39m40s, success** |
| Projects per discovery pass | 78 | **1** |
| Time per pass | ~4 min | **~2 sec** |
| Groups completed | 5 of 8 | **8 of 8** |
| Job log | 925,243 lines | 76,458 lines |

## Keeping it fixed

Every way this can silently stop working is now a build failure rather than a comment, because each one was previously only a comment. `ValidateDependabotCoverage` fails when:

- the package list drifts from `Directory.Packages.props` (a dropped package is one Dependabot silently stops updating);
- any package resolves to **no version** — the signal that `TargetFramework` no longer matches the `<Choose>` in `Directory.Packages.props`;
- `Microsoft.Health.Fhir.sln` claims the project, which would send discovery back to scanning everything;
- `dependabot.yml` stops pointing at this folder, which would stop NuGet updates outright.

All four are evaluation-only — no restore, no feeds, about a second — so the new PR stage just runs the target, and you can run it locally:

```bash
dotnet msbuild build/DependabotDiscovery -t:ValidateDependabotCoverage
```

## Also included

- **`dotnet/*` docker group.** The `mcr.microsoft.com/dotnet/*` pattern matched nothing: Dependabot names these dependencies by image path, not registry. Confirmed on the fork, which produced two ungrouped PRs.
- **Dead `System.Data.SqlClient` reference removed from `FHIRDataSynth`.** It lost its version in the Central Package Management migration (#2813) and has been unrestorable with `NU1010` ever since. Nothing in that tool references it — it is in no solution, so nothing built it and nobody noticed.


tools/FHIRDataSynth/FHIRDataSynth.csproj
```diff
@@ -6,7 +6,6 @@
 
   <ItemGroup>
     <PackageReference Include="Azure.Storage.Blobs" />
-    <PackageReference Include="System.Data.SqlClient" />
```
mikaelweave, line 9: Removing hanging package

src/Microsoft.Health.Fhir.SqlServer/Microsoft.Health.Fhir.SqlServer.csproj
```diff
@@ -88,8 +88,16 @@
...
+  <!--
+    Target contains sql build tasks.
+
+    The condition exists for evaluation-only tools. Dependabot reads project files without
+    restoring, which leaves NuGetPackageRoot empty and made this import throw MissingFileException,
+    failing the entire update job. It does not open a silent-failure window in a real build: a build
+    without a restore already fails on the missing assets file (NETSDK1004), and after a restore
+    NuGetPackageRoot is always set by nuget.g.props, so the import is never skipped.
+  -->
+  <Import Project="$([MSBuild]::NormalizePath($(NuGetPackageRoot), 'microsoft.health.tools.sql.tasks', $(HealthcareSharedPackageVersion), 'build', 'Sql.targets'))" Condition="'$(NuGetPackageRoot)' != ''" />
```
mikaelweave, line 100: This was the initial condition that was causing dependabot to crash. I think we should keep this to add more dependability to our remote builds.

PTaladay (approved):

PR #5826, opened 2026-09-17
High latency bundles

LTA-Thinking:
## Description
Adds a new header to allow for extra large bundles by having a user acknowledge that it will lead to higher latency.

## Related issues
Addresses [User Story 195349](https://microsofthealth.visualstudio.com/Health/_workitems/edit/195349)

## Testing
New tests added

## FHIR Team Checklist


docs/superpowers/specs/2026-09-17-high-latency-bundles-design.md
```diff
@@ -0,0 +1,49 @@
+# High-Latency Bundle Processing
```
fhibf, line 1: I think high-latency is the wrong term. 
I've been running tests with bundles with more than 1k conditional-operations, and they are taking 6 seconds to execute. That's not high latency based on the complexity of this type of requests. 

I suggest we follow a different perspective: if customer uses a header like "x-ms-bundle-large-operations" we always handle them as parallel. That will save us time and not cause high latency. And we should probably, with the presence of this flag, assume some behaviors to optimize the execution time. 

With this flag, we can also report this execution as part of a different SLI/SLO, and that should not affect our existing limits.

src/Microsoft.Health.Fhir.Core/Configs/BundleConfiguration.cs
```diff
@@ -11,6 +11,11 @@ public class BundleConfiguration
     {
         public int EntryLimit { get; set; } = 500;
 
+        /// <summary>
+        /// Gets or sets the maximum number of entries allowed when high-latency bundle processing is enabled.
+        /// </summary>
+        public int EntryLimitHighLatency { get; set; } = 1000;
```
fhibf, line 17: Could we rename this property to "EntryExpandedLimit"?

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
```diff
@@ -586,11 +586,23 @@ private async Task ExecuteTransactionForAllRequestsAsync(Hl7.Fhir.Model.Bundle r
             }
         }
 
+        private int GetEntryLimit()
```
fhibf, line 589: Maybe GetEntryExtendedLimit()?

src/Microsoft.Health.Fhir.Shared.Web/appsettings.json
```diff
@@ -122,6 +122,7 @@
         },
         "Bundle": {
             "EntryLimit": 500,
+            "EntryLimitHighLatency": 1000,
```
fhibf, line 125: EntryExtendedLimit?

fhibf (changes requested):

src/Microsoft.Health.Fhir.Shared.Api/Features/Resources/Bundle/BundleHandler.cs
LTA-Thinking: I would like to keep the method name just GetEntryLimit since this is what decides on if the normal or extended limit is used.

docs/superpowers/specs/2026-09-17-high-latency-bundles-design.md
LTA-Thinking: My concern is that 'large-operations' doesn't tell the user what to expect. It is saying what the user wants (support for larger bundles), but it isn't letting us know they acknowledge the trade offs that larger bundles come with. 

I agree this header can be used to trigger automatic performance enhancements, but if a customer starts sending these extra large bundles in volume I don't want to get complaints from people that they are being slow.

fhibf (approved):

PR #5828, opened 2026-09-17
Search throttling ADR

SergeyGaluzo:


docs/arch/adr-2609-limit-concurrent-searches.md
```diff
@@ -0,0 +1,55 @@
...
+```
+
+`C` is read live rather than cached, so the threshold tracks serverless autoscale and pool reconfiguration with no invalidation logic. Expressing the limit as a multiplier of allocated cores makes it SKU-independent; the floor prevents small SKUs from throttling on noise. `F = 2.0` is not arbitrary: under processor sharing, overload equals the latency multiplier, so the threshold is stated in the units of the problem — the point where responses take twice as long as on an idle database. The resulting SQL error maps to `RequestRateExceededException`, which the existing middleware already converts to a 429 with `Retry-After`.
+
+## Consequences
+
+- Overload becomes a visible, actionable 429 instead of unexplained latency — the primary goal.
+- Latency degradation is bounded at ~2x, and the limit holds regardless of instance count.
+- Clients that do not honour 429 see hard failures where they previously saw slow successes.
+- Adds ~0.6 ms CPU per request (~15% of a 4 ms request), self-limiting since shed requests skip the work.
```
LTA-Thinking, line 48: How is this time affected by the database being under high load? Since the point of this is to quickly tell the user the DB is overloaded, since this uses a DB call I'm concerned it will get stuck in the queue it is meant to measure.

SergeyGaluzo: AFAIK CPU scheduling does not work as queues. They consider all requests in round-robin fashion giving each of them a quantum of CPU time. So, in the worst scenario, of 2x overload, this time will double. We should always add network latency on the top, but that one does not change.

LTA-Thinking (approved):

PR #5830, opened 2026-09-18
Expired Resource Watchdog Rework

LTA-Thinking:
## Description
This PR changes the way the Expired Resource Watchdog works so that it runs more frequently.
It also adds more logs to the watchdogs to make debugging easier, but avoids sending too many unimportant logs by grouping together identical logs into periodic batches.

## Related issues
Addresses [Bug 207283](https://microsofthealth.visualstudio.com/Health/_workitems/edit/207283)

## Testing

## FHIR Team Checklist


src/Microsoft.Health.Fhir.SqlServer/Features/Watchdogs/ExpiredResourceCleanupWatchdog.cs
```diff
@@ -88,45 +91,69 @@ private async Task EnqueueBulkDeleteJobAsync(CancellationToken cancellationToken
...
             {
                 _logger.LogError(ex, "ExpiredResourceCleanupWatchdog: Error while enqueuing bulk delete job.");
             }
         }
+
+        private async Task<bool> ShouldCreateNewCleanupJob(CancellationToken cancellationToken)
+        {
+            // Checks the database for the last bulk delete job definition that was created by this watchdog. If the last job was created before the retention period, returns true to indicate a new job should be created.
+            IReadOnlyList<JobInfo> jobs = await _queueClient.GetJobsByQueueTypeAsync((byte)QueueType.BulkDelete, true, cancellationToken, Clock.UtcNow.AddMinutes(-_configuration.ExecutionIntervalInMinutes));
+
```
rajithaalurims, line 142: Can we add if (jobs == null || jobs.Count == 0) return true here?

rajithaalurims (dismissed):

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/GetAllJobs.sql
```diff
@@ -0,0 +1,54 @@
+--DROP PROCEDURE dbo.GetAllJobs
+GO
+CREATE PROCEDURE dbo.GetAllJobs @QueueType tinyint, @ReturnParentOnly bit = 0, @Since datetime = NULL
```
SergeyGaluzo, line 3: Logic in this new stored procedure is **not** supported by indexes. Adding indexes is not an option because it will reduce write perf of the queue. 
There are 2 alternatives I can think of:
1. Consider asking for active jobs, and if there is not active schedule one. This logic will schedule jobs more often, but because they are incremental by nature, you will have more smaller jobs, so over all work will be the same.
2. Another is to ask about last completed job (job with the highest id). It should be fast with existing indexes.
3. You can try thinking not about filtering by ids instead of dates.

src/Microsoft.Health.Fhir.SqlServer/Features/Watchdogs/Watchdog.cs
```diff
@@ -54,19 +54,21 @@ protected Watchdog()
 
         public abstract double PeriodSec { get; internal set; }
 
+        public bool AllowDbPeriodOverride { get; internal set; } = true;
+
         public bool IsInitialized { get; private set; }
 
-        public async Task ExecuteAsync(CancellationToken cancellationToken)
+        public async Task ExecuteAsync(Guid guid, CancellationToken cancellationToken)
```
SergeyGaluzo, line 61: Please explain why you need guid

SergeyGaluzo (changes requested):
I have performance concerns about this. New stored procedure is very expensive because job queue can contain millions of items for a given queue type.

In general job queue is a queue, and, by design, it guarantees efficient access to "current" items only. Date fields are not designed for any filtering (no indexes by design). They can be used within a group, when access is defined by group id. They are also useful for debugging purposes but should never be used for filtering access in production.

Please keep in mind that current SQL queue in FHIR is a state-of-the-art implementation which provides performance on par with well-known Azure storage implementations. If indexing changes are required, then they should be preceded by through performance testing. 

I suggest removing this new method and changing logic to avoid the need for this type of call.

src/Microsoft.Health.Fhir.CosmosDb/Features/Storage/Queues/CosmosQueueClient.cs
```diff
@@ -526,6 +526,11 @@ public async Task<IReadOnlyList<JobInfo>> GetActiveJobsByQueueTypeAsync(byte que
         return await GetActiveJobsByQueueTypeInternalAsync(queueType, returnParentOnly, cancellationToken);
     }
 
+    public async Task<IReadOnlyList<JobInfo>> GetJobsByQueueTypeAsync(byte queueType, bool returnParentOnly, CancellationToken cancellationToken, DateTimeOffset? cutoffDate = null)
```
SergeyGaluzo: I have performance concerns about this call. This call is very expensive because job queue can contain millions of items for a given queue type. I suggest changing logic to avoid this call.

src/Microsoft.Health.Fhir.SqlServer/Features/Schema/Sql/Sprocs/GetAllJobs.sql
LTA-Thinking: The stored procedure has been changed to only get the most recent job for a given queue type, sorted by group id. So this should make use of our existing indexes.

src/Microsoft.Health.Fhir.SqlServer/Features/Watchdogs/ExpiredResourceCleanupWatchdog.cs
LTA-Thinking: Added

rajithaalurims (approved):

